PR #375, opened 2024-03-20
Sort quorum IDs before aggregation

ian-shim:
Quorum numbers should be sorted to make sure APKs are in the right order as quorum numbers specified in [signature checker](https://github.com/Layr-Labs/eigenda/blob/master/core/eth/tx.go#L436)

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/batcher/batcher.go
+++ b/disperser/batcher/batcher.go
@@ -7,6 +7,7 @@ import (
 	"fmt"
 	"math"
 	"math/big"
+	"slices"
 	"time"
 
 	"github.com/Layr-Labs/eigenda/common"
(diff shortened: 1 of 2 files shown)


disperser/batcher/batcher.go
```diff
@@ -426,6 +427,7 @@ func (b *Batcher) HandleSingleBatch(ctx context.Context) error {
 	for quorumID := range batch.State.Operators {
 		quorumIDs = append(quorumIDs, quorumID)
 	}
+	slices.Sort(quorumIDs)
```
mooselumph, line 430: Out of curiosity, what happens if they aren't sorted?

mooselumph (approved):

disperser/batcher/batcher.go
ian-shim: [QuorumApks](https://github.com/Layr-Labs/eigenda/blob/master/core/eth/tx.go#L492) can be out of order and [signature validation](https://github.com/Layr-Labs/eigenlayer-middleware/blob/dev/src/BLSSignatureChecker.sol#L207) can fail

PR #376, opened 2024-03-21
Timeout disperser grpc streams

mooselumph:

Prevents callers from indefinitely using up GRPC stream resources. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/apiserver/server.go
+++ b/disperser/apiserver/server.go
@@ -98,6 +98,10 @@ func NewDispersalServer(
 }
 
 func (s *DispersalServer) DisperseBlobAuthenticated(stream pb.Disperser_DisperseBlobAuthenticatedServer) error {
+
+	ctx, cancel := context.WithTimeout(stream.Context(), s.serverConfig.GrpcTimeout)
+	defer cancel()
+
 	// Process disperse_request
 	in, err := stream.Recv()
 	if err != nil {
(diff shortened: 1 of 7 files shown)


disperser/apiserver/server.go
```diff
@@ -98,6 +98,10 @@ func NewDispersalServer(
 }
 
 func (s *DispersalServer) DisperseBlobAuthenticated(stream pb.Disperser_DisperseBlobAuthenticatedServer) error {
+
+	ctx, cancel := context.WithTimeout(stream.Context(), s.serverConfig.GrpcTimeout)
```
jianoaix, line 102: should this create a brand new context, with `deadline = min{stream.Context().Deadline(), s.serverConfig.GrpcTimeout)`?

jianoaix (commented):
lg

disperser/apiserver/server.go
mooselumph: It seems that `WithTimeout` already performs this check: 

```
func WithDeadlineCause(parent Context, d time.Time, cause error) (Context, CancelFunc) {
	if parent == nil {
		panic("cannot create context from nil parent")
	}
	if cur, ok := parent.Deadline(); ok && cur.Before(d) {
		// The current deadline is already sooner than the new one.
		return WithCancel(parent)
	}
```

jianoaix (approved):

PR #377, opened 2024-03-21
Fireblock with aws secret manager

bxue-l2:

Allow fireblock to fetch from secret manager

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/cmd/batcher/main.go
+++ b/disperser/cmd/batcher/main.go
@@ -180,13 +180,13 @@ func RunBatcher(ctx *cli.Context) error {
 		if err != nil {
 			return fmt.Errorf("cannot read fireblocks api key %s from secret manager: %w", config.FireblocksConfig.APIKeyName, err)
 		}
-		secretKey, err := secretmanager.ReadBytesFromSecretManager(context.Background(), config.FireblocksConfig.SecretKeyName, config.FireblocksConfig.Region)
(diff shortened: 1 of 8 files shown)


disperser/cmd/batcher/main.go
```diff
@@ -169,18 +170,23 @@ func RunBatcher(ctx *cli.Context) error {
...
-	if len(config.FireblocksConfig.APIKey) > 0 &&
-		len(config.FireblocksConfig.SecretKeyPath) > 0 &&
+	if len(config.FireblocksConfig.APIKeyName) > 0 &&
+		len(config.FireblocksConfig.SecretKeyName) > 0 &&
 		len(config.FireblocksConfig.BaseURL) > 0 &&
 		len(config.FireblocksConfig.VaultAccountName) > 0 &&
-		len(config.FireblocksConfig.WalletAddress) > 0 {
-		secretKey, err := os.ReadFile(config.FireblocksConfig.SecretKeyPath)
+		len(config.FireblocksConfig.WalletAddress) > 0 &&
+		len(config.FireblocksConfig.Region) > 0 {
```
pschork, line 178: In terms of fallback to EOA wallet, looks like we would need to *delete* one of these FireblocksConfig secrets - which seems not ideal. A cleaner alternative might be to add check on env var like `FIREBLOCKS_DISABLED=1` that can be easily updated & deployed via helm chart to trigger fallback to EOA wallet.

pschork (approved):
Thanks for adding disable logic

PR #379, opened 2024-03-21
Separate validation functions into new file

mooselumph:

Just moves code around for readablity.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


disperser/apiserver/server.go
```diff
@@ -97,6 +96,32 @@ func NewDispersalServer(
 	}
 }
 
+func (s *DispersalServer) Start(ctx context.Context) error {
```
ian-shim, line 99: not related to this PR, but i don't think `ctx` is used here. Do we need it?

ian-shim (approved):

disperser/apiserver/validation.go
```diff
@@ -0,0 +1,239 @@
...
+
+	blob := &core.Blob{
+		RequestHeader: header,
+		Data:          data,
+	}
+
+	return blob, nil
+}
+
+func (s *DispersalServer) getAccountRate(origin, authenticatedAddress string, quorumID core.QuorumID) (*PerUserRateInfo, string, error) {
```
jianoaix, line 102: The rate limiting is generally not in the category of request validation, but in access control category (authentication, authorization, rate limiting etc)

jianoaix (approved):

disperser/apiserver/validation.go
mooselumph: Do you want me to move this to an `authentication` file?

jianoaix: Or access control? To cover both auth and rate limiting

PR #380, opened 2024-03-21
Add retrieval ratelimit

mooselumph:

Adds a ratelimit to the retrieval endpoint. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


disperser/apiserver/server.go
```diff
@@ -605,6 +605,46 @@ func (s *DispersalServer) RetrieveBlob(ctx context.Context, req *pb.RetrieveBlob
 		return nil, api.NewInternalError("failed to get blob metadata, please retry")
 	}
 
+	// Check Ratelimit
```
jianoaix, line 608: Move this before the L600, so it can protect invalid requests from hitting DynamoDB

jianoaix (approved):
what's the plan for GetBlobStatus?

disperser/apiserver/server.go
mooselumph: I moved the blob rate limit before the dynamo call. The throughput one comes after because we need the blob size.

mooselumph:
> what's the plan for GetBlobStatus?

Thinking that we should just be able to rely on cloudfare for this.

disperser/apiserver/server.go
```diff
@@ -605,6 +643,64 @@ func (s *DispersalServer) RetrieveBlob(ctx context.Context, req *pb.RetrieveBlob
...
+			errorString := "request ratelimited"
+			info, ok := param.Info.(string)
+			if ok {
+				errorString += ": " + info
+			}
+			return nil, api.NewResourceExhaustedError(errorString)
+		}
+	}
+
+	if s.ratelimiter != nil {
```
jianoaix, line 673: This part no long needed?

jianoaix (approved):

PR #381, opened 2024-03-22
Revamp the operator CLI for opt-out

jianoaix:
We settle on the operator CLI semantics like this:
- opt-in quorum-list: add quorum-list to the operator (quorum-list must be not yet opted in)
- opt-out quorum-list: remove quorum-list from the operator (quorum-list must be already opted in)

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/eth/tx.go
+++ b/core/eth/tx.go
@@ -3,6 +3,7 @@ package eth
 import (
 	"context"
 	"crypto/ecdsa"
+	"encoding/hex"
 	"encoding/json"
 	"fmt"
 	"math/big"
@@ -307,7 +308,7 @@ func (t *Transactor) DeregisterOperator(ctx context.Context, pubkeyG1 *core.G1Po
 			}
 		}
 		if !found {
-			return fmt.Errorf("operator is not registered in quorum %d at block %d", quorumToDereg, blockNumber)
+			return fmt.Errorf("operatorId %s is not registered in quorum %d at block %d", hex.EncodeToString(operatorId[:]), quorumToDereg, blockNumber)
 		}
 	}
 
--- a/core/eth/tx.go
+++ b/core/eth/tx.go
@@ -288,6 +288,9 @@ func (t *Transactor) RegisterOperatorWithChurn(
 // If the operator isn't registered with any of the specified quorums, this function will return error, and
 // no quorum will be deregistered.
 func (t *Transactor) DeregisterOperator(ctx context.Context, pubkeyG1 *core.G1Point, blockNumber uint32, quorumIds []core.QuorumID) error {
+	if len(quorumIds) == 0 {
+		return errors.New("no quorum is specified to deregister from")
+	}
 	// Make sure the operator is registered in all the quorums it tries to deregister.
 	operatorId := HashPubKeyG1(pubkeyG1)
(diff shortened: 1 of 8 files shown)


core/eth/tx.go
```diff
@@ -280,26 +282,35 @@ func (t *Transactor) RegisterOperatorWithChurn(
...
 }
 
-// DeregisterOperator deregisters an operator with the given public key from the all the quorums that it is
+// DeregisterOperator deregisters an operator with the given public key from the specified the quorums that it is
 // registered with at the supplied block number. To fully deregister an operator, this function should be called
 // with the current block number.
-func (t *Transactor) DeregisterOperator(ctx context.Context, pubkeyG1 *core.G1Point, blockNumber uint32) error {
+// If the operator isn't registered with any of the specified quorums, this function will return error, and
+// no quorum will be deregistered.
+func (t *Transactor) DeregisterOperator(ctx context.Context, pubkeyG1 *core.G1Point, blockNumber uint32, quorumIds []core.QuorumID) error {
```
ian-shim, line 290: validate the length of `quorumIDs`?

node/plugin/config.go
```diff
@@ -66,7 +66,7 @@ var (
 	}
 	QuorumIDListFlag = cli.StringFlag{
 		Name:     "quorum-id-list",
-		Usage:    "Comma separated list of quorum IDs that the node will participate in. There should be at least one quorum ID. This list can contain quorums node is already registered with. If the node opts in to quorums already registered with, it will be a no-op.",
+		Usage:    "Comma separated list of quorum IDs that the node will opt-in or opt-out, depending on the OperationFlag. If OperationFlag is opt-in, all quorums should not have been registered already; if it's opt-out, all quorums should have been registered already",
```
ian-shim, line 69: `all quorums should not have been registered already` sounds like they can't provide quorums that they're already registered with, which isn't true. They _can_ include quorum IDs they've already opted in

jianoaix: I'm inclined to align with smart contract API which fails the call if "register quorums that are already registered" or "deregister quorums that are not currently registered" (@0x0aa0 please correct if this isn't the case).

jianoaix: We should actually also make opt-in/opt-out consistent w.r.t. this semantics

core/eth/tx.go
```diff
@@ -280,26 +282,38 @@ func (t *Transactor) RegisterOperatorWithChurn(
...
+	if len(quorumIds) == 0 {
```
ian-shim, line 291: Can we dedup them as well? Just wanted to make sure we don't loop over a massive slice on :305

jianoaix: That doesn't seem to help much, no matter it's dup or no dup, it's provided by users who are going to run this cmd

ian-shim: Sure, how about just check `len(quorumIDs) < 256`? It would prevent them from passing a huge number of elems

ian-shim: still to be done

ian-shim (approved):
lgtm. one remaining comment on slice length validation

core/eth/tx.go
jianoaix: https://github.com/Layr-Labs/eigenda/pull/381/commits/1629ea6e15a843a9bd3eb58e38bb7652a6da81e6

PR #383, opened 2024-03-22
Handle txn failure when retrieving txn receipt

ian-shim:
If the transaction has been submitted by Fireblocks, `wallet.GetTransactionReceipt` [returns](https://github.com/Layr-Labs/eigensdk-go/pull/158/files) `ErrTransactionFailed` if the transaction has failed permanently. 
Check this condition and disregard the transaction for receipt retrieval. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/batcher/txn_manager.go
+++ b/disperser/batcher/txn_manager.go
@@ -176,9 +176,15 @@ func (t *txnManager) ensureAnyTransactionEvaled(ctx context.Context, txs []*tran
 	defer queryTicker.Stop()
 	var receipt *types.Receipt
 	var err error
+	// transactions that need to be queried. Some transactions will be removed from this map depending on their status.
+	txnsToQuery := make(map[walletsdk.TxID]*types.Transaction, len(txs))
+	for _, tx := range txs {
+		txnsToQuery[tx.TxID] = tx.Transaction
+	}
+
 	for {
-		for _, tx := range txs {
-			receipt, err = t.wallet.GetTransactionReceipt(ctx, tx.TxID)
+		for txID, tx := range txnsToQuery {
+			receipt, err = t.wallet.GetTransactionReceipt(ctx, txID)
 			if err == nil {
 				chainTip, err := t.ethClient.BlockNumber(ctx)
 				if err == nil {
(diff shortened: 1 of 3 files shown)


disperser/batcher/txn_manager.go
```diff
@@ -176,9 +176,15 @@ func (t *txnManager) ensureAnyTransactionEvaled(ctx context.Context, txs []*tran
 	defer queryTicker.Stop()
 	var receipt *types.Receipt
 	var err error
+	// transactions that need to be queried. Some transactions will be removed from this map depending on their status.
+	txnsToQuery := make(map[walletsdk.TxID]*types.Transaction, len(txs))
+	for _, tx := range txs {
```
mooselumph, line 181: Not this PR, but do these "gas-upping" transactions have different hashes? Why do we need to separately watch for all of them?

disperser/batcher/txn_manager.go
```diff
@@ -193,8 +199,11 @@ func (t *txnManager) ensureAnyTransactionEvaled(ctx context.Context, txs []*tran
 				}
 			}
 
-			if errors.Is(err, ethereum.NotFound) {
-				t.logger.Debug("Transaction not yet mined", "component", "TxnManager", "method", "ensureAnyTransactionEvaled", "txID", tx.TxID, "txHash", tx.Hash().Hex())
+			if errors.Is(err, ethereum.NotFound) || errors.Is(err, walletsdk.ErrReceiptNotYetAvailable) {
+				t.logger.Debug("Transaction not yet mined", "component", "TxnManager", "method", "ensureAnyTransactionEvaled", "txID", txID, "txHash", tx.Hash().Hex(), "err", err)
+			} else if errors.Is(err, walletsdk.ErrTransactionFailed) {
+				t.logger.Debug("Transaction failed", "component", "TxnManager", "method", "ensureAnyTransactionEvaled", "txID", txID, "txHash", tx.Hash().Hex(), "err", err)
+				delete(txnsToQuery, txID)
```
mooselumph, line 206: I guess an optimization would be to immediately return if the map is empty, but doesn't seem important.

mooselumph (approved):

disperser/batcher/txn_manager.go
ian-shim: Each replacing transaction will have a new transaction hash even if it replaces an existing transaction, and not all nodes necessarily are sync'd with the new transaction at the same time. Which means that the old transaction being replaced can be confirmed before the new transaction propagates to the network. Therefore, we'd need to look at all of these transactions to check if any one succeeded.

PR #388, opened 2024-03-24
refactor: make the security param validation a shared function

jianoaix:
This will reduce code duplication, and make the security param validation logic consistent across the system.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/data.go
+++ b/core/data.go
@@ -75,17 +75,21 @@ type BlobRequestHeader struct {
 	SecurityParams []*SecurityParam `json:"security_params"`
 }
 
-func (sp *SecurityParam) Validate() error {
-	if sp.ConfirmationThreshold < sp.AdversaryThreshold || sp.ConfirmationThreshold-sp.AdversaryThreshold < 10 {
-		return errors.New("invalid request: quorum threshold must be >= 10 + adversary threshold")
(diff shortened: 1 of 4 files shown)


core/data.go
```diff
@@ -75,19 +75,23 @@ type BlobRequestHeader struct {
 	SecurityParams []*SecurityParam `json:"security_params"`
 }
 
-func (sp *SecurityParam) Validate() error {
-	if sp.ConfirmationThreshold < sp.AdversaryThreshold || sp.ConfirmationThreshold-sp.AdversaryThreshold < 10 {
-		return errors.New("invalid request: quorum threshold must be >= 10 + adversary threshold")
+func ValidateSecurityParam(confirmationThreshold, adversaryThreshold uint32) error {
```
mooselumph, line 78: It seems like maybe we should be checking that these parameters will be valid `uint8`s?

jianoaix: The range is checked below: they cannot even be greater than 100, so that's even stricter than unit8

mooselumph: lol, good point

mooselumph (approved):

PR #399, opened 2024-03-27
Index operator state by quorum id not quorum index

mooselumph:

Uses the `quorumID` instead of the quorum index in the `operatorState` returned by the `chainState` implementation.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/eth/tx.go
+++ b/core/eth/tx.go
@@ -370,6 +370,7 @@ func (t *Transactor) GetOperatorStakes(ctx context.Context, operator core.Operat
 		return nil, nil, err
 	}
 
+	// BitmapToQuorumIds returns an ordered list of quorums in ascending order, which is the same order as the state_ returned by the contract
 	quorumIds := BitmapToQuorumIds(quorumBitmap)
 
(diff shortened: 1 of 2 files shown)


jianoaix (approved):

core/eth/tx.go
```diff
@@ -370,9 +370,11 @@ func (t *Transactor) GetOperatorStakes(ctx context.Context, operator core.Operat
 		return nil, nil, err
 	}
 
+	quorumIds := BitmapToQuorumIds(quorumBitmap)
+
 	state := make(core.OperatorStakes, len(state_))
 	for i := range state_ {
-		quorumID := core.QuorumID(i)
+		quorumID := quorumIds[i]
```
ian-shim, line 377: So we're assuming `BitmapToQuorumIds` returned an ordered list of quorums in ascending order, and the quorums in `state_` is in the same order, which is true in the [current implementation](https://github.com/Layr-Labs/eigenlayer-middleware/blob/dev/src/OperatorStateRetriever.sol#L40).

PR #406, opened 2024-03-28
Add blob quorum metrics at node

jianoaix:
Provide visibility to multi quorum blobs at node

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


node/metrics.go
```diff
@@ -31,6 +33,8 @@ type Metrics struct {
 	AccuBatches *prometheus.CounterVec
```
ian-shim, line 33: ok to remove `AccuBatchesDeprecated`?

ian-shim (approved):

node/metrics.go
jianoaix: Yep

PR #407, opened 2024-03-28
Add checks for g1 and g2 point deserialization

mooselumph:

Ensures that DA nodes cannot panic the disperser by sending an empty signature. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/attestation.go
+++ b/core/attestation.go
@@ -59,7 +59,7 @@ func (p *G1Point) Deserialize(data []byte) (*G1Point, error) {
 	return &G1Point{point}, nil
 }
 
-func (p *G1Point) Copy() *G1Point {
+func (p *G1Point) Clone() *G1Point {
 	return &G1Point{&bn254.G1Affine{
 		X: newFpElement(p.X.BigInt(new(big.Int))),
 		Y: newFpElement(p.Y.BigInt(new(big.Int))),
@@ -96,7 +96,7 @@ func (p *G2Point) Deserialize(data []byte) (*G2Point, error) {
 	return &G2Point{point}, nil
 }
 
-func (p *G2Point) Copy() *G2Point {
+func (p *G2Point) Clone() *G2Point {
 	return &G2Point{&bn254.G2Affine{
 		X: struct {
 			A0, A1 fp.Element
--- a/core/attestation.go
+++ b/core/attestation.go
@@ -51,8 +51,19 @@ func (p *G1Point) Serialize() []byte {
 	return bn254utils.SerializeG1(p.G1Affine)
 }
 
-func (p *G1Point) Deserialize(data []byte) *G1Point {
-	return &G1Point{bn254utils.DeserializeG1(data)}
+func (p *G1Point) Deserialize(data []byte) (*G1Point, error) {
+	point, err := bn254utils.DeserializeG1(data)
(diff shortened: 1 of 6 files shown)


operators/churner/server.go
```diff
@@ -74,7 +74,11 @@ func (s *Server) Churn(ctx context.Context, req *pb.ChurnRequest) (*pb.ChurnRepl
 		return nil, api.NewResourceExhaustedError(fmt.Sprintf("previous approval not expired, retry in %d", s.latestExpiry-now.Unix()))
 	}
 
-	request := createChurnRequest(req)
+	request, err := createChurnRequest(req)
+	if err != nil {
+		s.metrics.IncrementFailedRequestNum("Churn", FailReasonInvalidRequest)
+		return nil, api.NewInvalidArgError(fmt.Sprintf("invalid request: %s", err.Error()))
```
jianoaix, line 80: We may not need to prefix it with "invalid request" now with the standard error code, otherwise the end result will look like "InvalidRequest invalid request ...", which is slightly redundant

core/attestation.go
```diff
@@ -51,8 +51,19 @@ func (p *G1Point) Serialize() []byte {
...
-	return &G1Point{bn254utils.DeserializeG1(data)}
+func (p *G1Point) Deserialize(data []byte) (*G1Point, error) {
+	point, err := bn254utils.DeserializeG1(data)
+	if err != nil {
+		return nil, err
+	}
+	return &G1Point{point}, nil
+}
+
+func (p *G1Point) Copy() *G1Point {
```
jianoaix, line 62: Will it make more sense to name it "Clone"? This is deep copying (creating a new instance)

jianoaix (approved):

core/attestation.go
mooselumph: Yes, this is a deep copy. Is `Clone` a standard usage for this? Would `DeepCopy` be clearest?

core/bn254/attestation.go
```diff
@@ -106,11 +107,14 @@ func SerializeG1(p *bn254.G1Affine) []byte {
 	return b
 }
 
-func DeserializeG1(b []byte) *bn254.G1Affine {
+func DeserializeG1(b []byte) (*bn254.G1Affine, error) {
+	if len(b) != 64 {
+		return nil, errors.New("could not deserialize point, invalid length")
```
ian-shim, line 112: nit: `could not deserialize G1 point: length must be exactly 64`

core/bn254/attestation.go
```diff
@@ -134,13 +138,16 @@ func SerializeG2(p *bn254.G2Affine) []byte {
 	return b
 }
 
-func DeserializeG2(b []byte) *bn254.G2Affine {
+func DeserializeG2(b []byte) (*bn254.G2Affine, error) {
+	if len(b) != 128 {
+		return nil, errors.New("could not deserialize point, invalid length")
```
ian-shim, line 143: nit: `could not deserialize G2 point: length must be exactly 128`

core/attestation.go
jianoaix: Yea, clone is usually deep copy

PR #413, opened 2024-03-29
Fix most critical integer casting issues

mooselumph:

Addresses unsafe integer casting (1 of 2)

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/node/grpc/utils.go
+++ b/node/grpc/utils.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 	"reflect"
 
+	"github.com/Layr-Labs/eigenda/api"
 	pb "github.com/Layr-Labs/eigenda/api/grpc/node"
 	"github.com/Layr-Labs/eigenda/core"
 	"github.com/Layr-Labs/eigenda/encoding"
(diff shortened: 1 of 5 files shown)


node/grpc/utils.go
```diff
@@ -88,6 +89,13 @@ func GetBlobHeaderFromProto(h *pb.BlobHeader) (*core.BlobHeader, error) {
 
 	quorumHeaders := make([]*core.BlobQuorumInfo, len(h.GetQuorumHeaders()))
 	for i, header := range h.GetQuorumHeaders() {
+		if header.GetQuorumId() > core.MaxQuorumID {
```
jianoaix, line 92: This validation is redundant as we validated it as soon as the request entered the gRPC server (in server.go). But maybe fine since it's low cost.

jianoaix (approved):

node/grpc/utils.go
mooselumph: I think I noted at least one other place where this function was called without validation!

PR #421, opened 2024-03-30
Update accounting logic for total encoded blob size 

ian-shim:
- use existing `Bundle.Size()` method to keep track of the total size of encoded blobs
- update the bundle size to be in `uint64` instead of `int64`
- added proof size in `Frame.Size()`

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/data.go
+++ b/core/data.go
@@ -157,10 +157,10 @@ type BlobMessage struct {
 	Bundles    Bundles
 }
 
-func (b Bundle) Size() int64 {
-	size := int64(0)
+func (b Bundle) Size() uint64 {
+	size := uint64(0)
 	for _, chunk := range b {
-		size += int64(chunk.Size())
+		size += chunk.Size()
 	}
 	return size
 }
@@ -181,8 +181,8 @@ func (cb Bundles) Serialize() (map[uint32][][]byte, error) {
 }
 
 // Returns the size of the bundles in bytes.
-func (cb Bundles) Size() int64 {
-	size := int64(0)
+func (cb Bundles) Size() uint64 {
+	size := uint64(0)
 	for _, bundle := range cb {
 		size += bundle.Size()
 	}
--- a/disperser/batcher/encoded_blob_store.go
+++ b/disperser/batcher/encoded_blob_store.go
@@ -35,7 +35,7 @@ type EncodingResult struct {
 	ReferenceBlockNumber uint
 	BlobQuorumInfo       *core.BlobQuorumInfo
 	Commitment           *encoding.BlobCommitments
-	Chunks               core.Bundle
+	Chunks               []*encoding.Frame
 	Assignments          map[core.OperatorID]core.Assignment
 	Status               status
 }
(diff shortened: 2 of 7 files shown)


encoding/data.go
```diff
@@ -41,8 +41,9 @@ func (f *Frame) Length() int {
 }
 
 // Returns the size of chunk in bytes.
-func (f *Frame) Size() int {
-	return f.Length() * BYTES_PER_COEFFICIENT
+func (f *Frame) Size() uint64 {
+	proofSize := uint64(2 * 32)
```
jianoaix, line 45: This may make the blobs stored and removed not the same at Node.
When storing blobs, it's in this entry: https://github.com/Layr-Labs/eigenda/blob/4db2480f472fc6958786ffc95eea96b39fee2d20/node/store.go#L246
When removing blobs, it's deducting this number: https://github.com/Layr-Labs/eigenda/blob/4db2480f472fc6958786ffc95eea96b39fee2d20/node/store.go#L142

ian-shim: ah ok i'll exclude the proof size and comment that this method only returns the size of the chunks

disperser/batcher/encoded_blob_store.go
```diff
@@ -35,7 +35,7 @@ type EncodingResult struct {
 	ReferenceBlockNumber uint
 	BlobQuorumInfo       *core.BlobQuorumInfo
 	Commitment           *encoding.BlobCommitments
-	Chunks               []*encoding.Frame
+	Chunks               core.Bundle
```
bxue-l2, line 38: In the comment above Bundle struct,  it is intended as the collection of chunks associated with a single blob, for a single operator and a single quorum. EncodingResult contains multiple operators, we can pack frames into bundles later

core/data.go
```diff
@@ -157,10 +157,10 @@ type BlobMessage struct {
 	Bundles    Bundles
 }
 
-func (b Bundle) Size() int64 {
-	size := int64(0)
+func (b Bundle) Size() uint64 {
+	size := uint64(0)
 	for _, chunk := range b {
-		size += int64(chunk.Size())
+		size += chunk.Size()
```
bxue-l2, line 163: Just for your information, referring to yesterdays discussion, all chunk will now contains 32bytes. Nothing wrong here

disperser/batcher/encoded_blob_store.go
```diff
@@ -191,11 +191,7 @@ func getRequestID(key disperser.BlobKey, quorumID core.QuorumID) requestID {
...
 
+// getChunksSize returns the total size of all the chunks in the encoded result in bytes
 func getChunksSize(result *EncodingResult) uint64 {
-	var size uint64
-
-	for _, chunk := range result.Chunks {
-		size += uint64(len(chunk.Coeffs) * 256) // 256 bytes per symbol
-	}
-	return size + 256*2 // + 256 * 2 bytes for proof
+	return result.Chunks.Size()
```
bxue-l2, line 196: good catch, 256 byte is 8 times over

bxue-l2 (commented):
Bundles is the part that concerns me

disperser/batcher/encoded_blob_store.go
ian-shim: Hmm I see your concern. There is really no concept of operator at this point, but wanted to use `Bundle` so that we can reuse existing `Size` method. I'll just leave it as a slice here and convert it when it actually uses `Size` method

bxue-l2 (approved):

PR #422, opened 2024-03-31
add checks on curve, and serialize using gnark

bxue-l2:

Add more checks on g1, g2 checks, as suggested by Sigp. It contains some overlap with previous PR #407.

The new code uses Gnark to check if a point(G1, G2) is on the subgroup, and which is eventually called if a point is in the curve inside the [gnark](https://github.com/Consensys/gnark-crypto/blob/v0.12.1/ecc/bn254/g1.go#L415) library. 

Since G1Commitment serde are only used in communication offchain, replacing with gnark setBytes and Bytes() is safe.

ONE Important issue is that: 

the previous serde code assigns wrong value for G2Point,  see https://github.com/Layr-Labs/eigenda/pull/407/files#diff-5480a750aef8baded2523848b6b296388101bc41127b739c7c25d232e3358abaL139 
and see https://github.com/consensys/gnark-crypto/blob/v0.12.1/ecc/bn254/marshal.go#L1092

But since everything is internal, between disperser to node, and between node to churner.

However, since this part is so pervasively used, it might be a good idea to keep the original, but adding the checks to the wrong point. But it did a silly mark to the core

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/bn254/attestation.go
+++ b/core/bn254/attestation.go
@@ -93,34 +93,6 @@ func MulByGeneratorG2(a *fr.Element) *bn254.G2Affine {
 	return new(bn254.G2Affine).ScalarMultiplication(g2Gen, a.BigInt(new(big.Int)))
 }
 
-func SerializeG1(p *bn254.G1Affine) []byte {
-	res := p.RawBytes()
-	return res[:]
-}
-
-func DeserializeG1(b []byte) (*bn254.G1Affine, error) {
-	var p bn254.G1Affine
-	_, err := p.SetBytes(b)
-	if err != nil {
-		return nil, err
-	}
-	return &p, nil
-}
-
-func SerializeG2(p *bn254.G2Affine) []byte {
-	res := p.RawBytes()
-	return res[:]
-}
-
-func DeserializeG2(b []byte) (*bn254.G2Affine, error) {
-	var p bn254.G2Affine
-	_, err := p.SetBytes(b)
-	if err != nil {
-		return nil, err
-	}
-	return &p, nil
-}
-
 func MakePubkeyRegistrationData(privKey *fr.Element, operatorAddress common.Address) *bn254.G1Affine {
 	toHash := make([]byte, 0)
 	toHash = append(toHash, crypto.Keccak256([]byte("BN254PubkeyRegistration(address operator)"))...)
--- a/core/bn254/attestation.go
+++ b/core/bn254/attestation.go
@@ -94,7 +94,7 @@ func MulByGeneratorG2(a *fr.Element) *bn254.G2Affine {
 }
 
 func SerializeG1(p *bn254.G1Affine) []byte {
-	res := p.Bytes()
+	res := p.RawBytes()
 	return res[:]
 }
 
@@ -108,7 +108,7 @@ func DeserializeG1(b []byte) (*bn254.G1Affine, error) {
 }
 
 func SerializeG2(p *bn254.G2Affine) []byte {
-	res := p.Bytes()
+	res := p.RawBytes()
 	return res[:]
 }
 
--- a/core/bn254/attestation.go
+++ b/core/bn254/attestation.go
@@ -94,7 +94,7 @@ func MulByGeneratorG2(a *fr.Element) *bn254.G2Affine {
 }
 
 func SerializeG1(p *bn254.G1Affine) []byte {
-	res := p.RawBytes()
+	res := p.Bytes()
 	return res[:]
 }
 
@@ -108,7 +108,7 @@ func DeserializeG1(b []byte) (*bn254.G1Affine, error) {
 }
 
 func SerializeG2(p *bn254.G2Affine) []byte {
-	res := p.RawBytes()
+	res := p.Bytes()
 	return res[:]
 }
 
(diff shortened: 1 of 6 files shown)


core/bn254/attestation.go
```diff
@@ -95,59 +94,31 @@ func MulByGeneratorG2(a *fr.Element) *bn254.G2Affine {
 }
 
 func SerializeG1(p *bn254.G1Affine) []byte {
```
mooselumph, line 96: Wouldn't it be better to just discard these altogether, and just make use of the gnark methods whenever the serialization methods are called?

encoding/serialization.go
```diff
@@ -55,15 +64,25 @@ func (c *G1Commitment) UnmarshalJSON(data []byte) error {
 	}
 	c.X = g1Point.X
 	c.Y = g1Point.Y
+
```
mooselumph, line 67: `UnmarshalJSON` seems to be unused. Remove it?

encoding/serialization.go
```diff
@@ -75,6 +94,10 @@ func (c *G2Commitment) UnmarshalJSON(data []byte) error {
 	}
 	c.X = g2Point.X
 	c.Y = g2Point.Y
+
```
mooselumph, line 97: Same here, let's remove this function if unused.

core/bn254/attestation.go
anupsv: +1

encoding/serialization.go
```diff
@@ -39,11 +44,15 @@ func Decode(b []byte) (Frame, error) {
...
 }
 
 func (c *G1Commitment) Deserialize(data []byte) (*G1Commitment, error) {
-	err := decode(data, c)
+	_, err := (*bn254.G1Affine)(c).SetBytes(data)
+	if err != nil {
+		return nil, err
+	}
 	return c, err
 }
```
mooselumph, line 58: These changes seem unnecessary. However, since these methods only appear to be used for communication between batcher and encoder, I don't think there is a big impact.

mooselumph (approved):
Looks good, but I think we can clean it up a bit.

core/serialization.go
```diff
@@ -362,6 +363,19 @@ func (h *BlobHeader) Serialize() ([]byte, error) {
 
 func (h *BlobHeader) Deserialize(data []byte) (*BlobHeader, error) {
 	err := decode(data, h)
+
+	if !(*bn254.G1Affine)(h.BlobCommitments.Commitment).IsInSubGroup() {
+		return nil, fmt.Errorf("in BlobHeader Commitment is in the subgroup")
```
jianoaix, line 368: is `not` in subgroup

encoding/serialization.go
```diff
@@ -14,6 +15,10 @@ func (c *Frame) Serialize() ([]byte, error) {
 
 func (c *Frame) Deserialize(data []byte) (*Frame, error) {
 	err := decode(data, c)
+	if !c.Proof.IsInSubGroup() {
+		return nil, fmt.Errorf("proof is in not the subgroup")
```
jianoaix, line 19: is not `in` the subgroup

node/grpc/utils.go
```diff
@@ -73,20 +75,34 @@ func GetBlobHeaderFromProto(h *pb.BlobHeader) (*core.BlobHeader, error) {
 		X: *commitX,
 		Y: *commitY,
 	}
+
+	if !(*bn254.G1Affine)(commitment).IsInSubGroup() {
```
jianoaix, line 79: should these go into the request validation so it can be rejected as soon as possible?

anupsv (approved):

core/serialization.go
bxue-l2: we want to reject points that is not part of subgroup, which has undefined operation for (G1, G2)point operation

encoding/serialization.go
bxue-l2: right, we reject it if it is not in the subgroup

jianoaix: I'm pointing out a typo:)

core/serialization.go
jianoaix: typo as well

bxue-l2: I see, thanks!

core/bn254/attestation.go
bxue-l2: fixed

encoding/serialization.go
bxue-l2: I think it is used by dataapi, https://github.com/Layr-Labs/eigenda/blob/master/disperser/dataapi/server.go#L50

encoding/serialization.go
bxue-l2: dataapi?

encoding/serialization.go
bxue-l2: it is pointed out by the audit. Unless there is performance downgrade, it doesn't hurt to make it safe.

node/grpc/utils.go
bxue-l2: i can make the call to "GetBlobHeaderFromProto" in the validation phase

bxue-l2: fixed, please double check

PR #424, opened 2024-04-01
Measure attestation latency

ian-shim:
Exports metrics for attestation latency

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/batcher/metrics.go
+++ b/disperser/batcher/metrics.go
@@ -49,10 +49,15 @@ type FinalizerMetrics struct {
 	Latency                *prometheus.SummaryVec
 }
 
+type DispatcherMetrics struct {
+	Latency *prometheus.SummaryVec
+}
+
 type Metrics struct {
 	*EncodingStreamerMetrics
 	*TxnManagerMetrics
 	*FinalizerMetrics
+	*DispatcherMetrics
 
 	registry *prometheus.Registry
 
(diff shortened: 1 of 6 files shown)


disperser/batcher/metrics.go
```diff
@@ -149,10 +154,23 @@ func NewMetrics(httpPort string, logger logging.Logger) *Metrics {
 		),
 	}
 
+	dispatcherMatrics := DispatcherMetrics{
+		Latency: promauto.With(reg).NewSummaryVec(
+			prometheus.SummaryOpts{
+				Namespace:  namespace,
+				Name:       "attestation_latency_s",
```
dmanc, line 161: Could we do milliseconds? It will convert on the dashboard anyway.

ian-shim: done

disperser/batcher/metrics.go
```diff
@@ -149,10 +154,23 @@ func NewMetrics(httpPort string, logger logging.Logger) *Metrics {
...
 
+	dispatcherMatrics := DispatcherMetrics{
+		Latency: promauto.With(reg).NewSummaryVec(
+			prometheus.SummaryOpts{
+				Namespace:  namespace,
+				Name:       "attestation_latency_ms",
+				Help:       "attestation latency summary in milliseconds",
+				Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.95: 0.01, 0.99: 0.001},
+			},
+			[]string{"type"},
```
jianoaix, line 165: We'll standardizing the Node API as the Disperser. So this label can be `status`.

PR #425, opened 2024-04-01
[Encoder][Node] Add nil checks

siddimore:

Add nil checks for incoming requests in Disperser and Node

Additional Protobuf checks in this PR: https://github.com/Layr-Labs/eigenda/pull/427

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -271,10 +271,6 @@ func (n *Node) ProcessBatch(ctx context.Context, header *core.BatchHeader, blobs
 		return nil, errors.New("ProcessBatch: number of blobs must be greater than zero")
 	}
 
-	if len(rawBlobs) == 0 {
-		return nil, errors.New("ProcessBatch: number of raw blobs must be greater than zero")
-	}
-
 	if len(blobs) != len(rawBlobs) {
 		return nil, errors.New("number of parsed blobs must be the same as number of blobs from protobuf request")
 	}


ian-shim (commented):
can any of these proto requests themselves be nil?

node/node.go
```diff
@@ -268,10 +268,15 @@ func (n *Node) ProcessBatch(ctx context.Context, header *core.BatchHeader, blobs
 	log.Debug("Processing batch", "num of blobs", len(blobs))
 
 	if len(blobs) == 0 {
-		return nil, errors.New("the number of blobs must be greater than zero")
+		return nil, errors.New("ProcessBatch: number of blobs must be greater than zero")
 	}
+
+	if len(rawBlobs) == 0 {
```
mooselumph, line 274: This check is unneeded; if `len(blobs) != 0` and `len(rawBlobs) == 0`, then the next check will fail.

mooselumph (approved):
lg! just one comment

node/grpc/utils.go
```diff
@@ -67,6 +67,12 @@ func GetBlobMessages(in *pb.StoreChunksRequest) ([]*core.BlobMessage, error) {
 
 // GetBlobHeaderFromProto constructs a core.BlobHeader from a proto of pb.BlobHeader.
 func GetBlobHeaderFromProto(h *pb.BlobHeader) (*core.BlobHeader, error) {
+
+	if h == nil {
+		return nil, api.NewInvalidArgError("GetBlobHeaderFromProto: blob header is nil")
```
ian-shim, line 72: Is this only used to handle API requests at the top level? 
Otherwise, i don't think it should return the api error

PR #426, opened 2024-04-01
RPC requests accounting

jianoaix:
The current `NumBlobRequests` is counting the same RPC requests multiple times (one for each quorum). This is not right for API accounting.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/apiserver/server.go
+++ b/disperser/apiserver/server.go
@@ -102,12 +102,14 @@ func (s *DispersalServer) DisperseBlobAuthenticated(stream pb.Disperser_Disperse
 	// Process disperse_request
 	in, err := stream.Recv()
 	if err != nil {
+		s.metrics.HandleInvalidArgRpcRequest("DisperseBlobAuthenticated")
 		s.metrics.HandleInvalidArgRequest("DisperseBlobAuthenticated")
 		return api.NewInvalidArgError(fmt.Sprintf("error receiving next message: %v", err))
 	}
 
 	request, ok := in.GetPayload().(*pb.AuthenticatedRequest_DisperseRequest)
 	if !ok {
+		s.metrics.HandleInvalidArgRpcRequest("DisperseBlobAuthenticated")
 		s.metrics.HandleInvalidArgRequest("DisperseBlobAuthenticated")
 		return api.NewInvalidArgError("missing DisperseBlobRequest")
 	}
@@ -117,6 +119,7 @@ func (s *DispersalServer) DisperseBlobAuthenticated(stream pb.Disperser_Disperse
 		for _, quorumID := range request.DisperseRequest.CustomQuorumNumbers {
 			s.metrics.HandleFailedRequest(codes.InvalidArgument.String(), fmt.Sprint(quorumID), len(request.DisperseRequest.GetData()), "DisperseBlobAuthenticated")
 		}
+		s.metrics.HandleInvalidArgRpcRequest("DisperseBlobAuthenticated")
 		return api.NewInvalidArgError(err.Error())
 	}
 
@@ -124,12 +127,14 @@ func (s *DispersalServer) DisperseBlobAuthenticated(stream pb.Disperser_Disperse
 	// Decode public key
 	publicKeyBytes, err := hexutil.Decode(blob.RequestHeader.AccountID)
 	if err != nil {
+		s.metrics.HandleInvalidArgRpcRequest("DisperseBlobAuthenticated")
 		s.metrics.HandleInvalidArgRequest("DisperseBlobAuthenticated")
 		return api.NewInvalidArgError(fmt.Sprintf("failed to decode public key (%v): %v", blob.RequestHeader.AccountID, err))
 	}
 
 	pubKey, err := crypto.UnmarshalPubkey(publicKeyBytes)
 	if err != nil {
+		s.metrics.HandleInvalidArgRpcRequest("DisperseBlobAuthenticated")
 		s.metrics.HandleInvalidArgRequest("DisperseBlobAuthenticated")
 		return api.NewInvalidArgError(fmt.Sprintf("failed to decode public key (%v): %v", blob.RequestHeader.AccountID, err))
 	}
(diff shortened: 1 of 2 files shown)


disperser/apiserver/server.go
```diff
@@ -507,13 +532,15 @@ func (s *DispersalServer) GetBlobStatus(ctx context.Context, req *pb.BlobStatusR
 
 	requestID := req.GetRequestId()
 	if len(requestID) == 0 {
+		s.metrics.HandleInvalidArgRpcRequest("GetBlobStatus")
 		s.metrics.HandleInvalidArgRequest("GetBlobStatus")
```
jianoaix, line 536: These should be removed later: quorum tracking doesnt' apply to GetBlobStatus

disperser/apiserver/server.go
```diff
@@ -614,6 +646,7 @@ func (s *DispersalServer) RetrieveBlob(ctx context.Context, req *pb.RetrieveBlob
 
 	origin, err := common.GetClientAddress(ctx, s.rateConfig.ClientIPHeader, 2, true)
 	if err != nil {
+		s.metrics.HandleInvalidArgRpcRequest("RetrieveBlob")
 		s.metrics.HandleInvalidArgRequest("RetrieveBlob")
```
jianoaix, line 650: These should be removed later: quorum tracking doesnt' apply to RetrieveBlob

disperser/apiserver/server.go
```diff
@@ -413,11 +433,13 @@ func (s *DispersalServer) checkRateLimitsAndAddRatesToHeader(ctx context.Context
 
 		globalRates, ok := s.rateConfig.QuorumRateInfos[param.QuorumID]
 		if !ok {
+			s.metrics.HandleInternalFailureRpcRequest("DisperseBlobAuthenticated")
```
mooselumph, line 436: `checkRateLimits...` is called by both "DisperseBlob" and "DispserseBlobAuthenticated". Maybe we need to pass in the `apiMethod` argument here as elsewhere.

disperser/apiserver/server.go
mooselumph: Why can't we remove it now?

disperser/apiserver/server.go
```diff
@@ -522,17 +549,22 @@ func (s *DispersalServer) GetBlobStatus(ctx context.Context, req *pb.BlobStatusR
...
 
 	isConfirmed, err := metadata.IsConfirmed()
 	if err != nil {
+		s.metrics.HandleInternalFailureRpcRequest("GetBlobStatus")
 		return nil, api.NewInternalError(fmt.Sprintf("missing confirmation information: %s", err.Error()))
 	}
 
+	s.metrics.HandleSuccessfulRpcRequest("GetBlobStatus")
+
 	s.logger.Debug("isConfirmed", "metadata", metadata, "isConfirmed", isConfirmed)
```
mooselumph, line 568: Unrelated, but should we remove this debug call? Seems like `metadata` could be large.

disperser/apiserver/server.go
```diff
@@ -675,9 +712,11 @@ func (s *DispersalServer) RetrieveBlob(ctx context.Context, req *pb.RetrieveBlob
 			},
 		})
 		if err != nil {
+			s.metrics.HandleInternalFailureRpcRequest("RetrieveBlob")
 			return nil, api.NewInternalError(fmt.Sprintf("ratelimiter error: %v", err))
 		}
 		if !allowed {
+			s.metrics.HandleRateLimitedRpcRequest("RetrieveBlob")
 			s.metrics.HandleFailedRequest(codes.ResourceExhausted.String(), "", 0, "RetrieveBlob")
```
mooselumph, line 720: Would we also remove this quorum specific metrics?

mooselumph (approved):
Looks good. Does this PR address the overcounting issue? Can you describe how it resolves it?

disperser/apiserver/server.go
jianoaix: Good catch, done

disperser/apiserver/server.go
jianoaix: Switched to just log metadataKey

disperser/apiserver/server.go
jianoaix: It can be a sanity check for the new metric with new v.s. old comparison

disperser/apiserver/server.go
jianoaix: I'd do it in followup, making the goal of this PR focused

jianoaix:
> Looks good. Does this PR address the overcounting issue? Can you describe how it resolves it?

It's straightforward and just avoid updating the metric for each quorum

mooselumph:
A little late, but I'm confused about why some of the non-RPC metrics loop over the quorums and others do not: 

[s.metrics.HandleFailedRequest(codes.InvalidArgument.String(), fmt.Sprint(quorumID), len(request.DisperseRequest.GetData()), "DisperseBlobAuthenticated")](https://github.com/Layr-Labs/eigenda/blob/b94604e9edc82a499700f6876e32cf1fef9a5d3a/disperser/apiserver/server.go#L120-L121)

vs

[s.metrics.HandleInvalidArgRequest("DisperseBlobAuthenticated")](https://github.com/Layr-Labs/eigenda/blob/b94604e9edc82a499700f6876e32cf1fef9a5d3a/disperser/apiserver/server.go#L131)

How are the blob-specific metrics meant to be used here?

jianoaix:
> A little late, but I'm confused about why some of the non-RPC metrics loop over the quorums and others do not:
> 
> [s.metrics.HandleFailedRequest(codes.InvalidArgument.String(), fmt.Sprint(quorumID), len(request.DisperseRequest.GetData()), "DisperseBlobAuthenticated")](https://github.com/Layr-Labs/eigenda/blob/b94604e9edc82a499700f6876e32cf1fef9a5d3a/disperser/apiserver/server.go#L120-L121)
> 
> vs
> 
> [s.metrics.HandleInvalidArgRequest("DisperseBlobAuthenticated")](https://github.com/Layr-Labs/eigenda/blob/b94604e9edc82a499700f6876e32cf1fef9a5d3a/disperser/apiserver/server.go#L131)
> 
> How are the blob-specific metrics meant to be used here?

Yea, this is another issue of current metric: it's not consistent. As a result, the number shows on this metric is not well defined.

PR #428, opened 2024-04-02
eigenda only accepts 32 bytes less than field modulo

bxue-l2:

This PR changes data acceptance rules for EigenDA. All inputs have to satisfy a new constraint that every Big endian 32bytes has to be less than 21888242871839275222246405745257275088548364400416034343698204186575808495617, which is a modulo for bn254 field element.

The PR also contains a simple coded that converts a byte array to a new array where every 32-bytes is a valid field element. It is implemented by adding an empty byte in front of every 31 bytes of the original data. The PR also includes a function to convert data data back.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proto/disperser/disperser.proto
jianoaix, line 1: Please add documentation to DisperseBlobRequest.data about how it's going to be interpreted

encoding/utils/codec/codec.go
```diff
@@ -0,0 +1,66 @@
...
+import (
+	"github.com/Layr-Labs/eigenda/encoding"
+)
+
+// ConvertByPaddingEmptyByte takes bytes and insert an empty at the front of every 31 byte
+// This ensure every 32 bytes are within the valid range of a field element for bn254 curve
+// If the input data is not a multiple of 31, the reminder is added to the output by
+// inserting a 0 and the reminder. The output does not necessarily be a multipler of 32
+func ConvertByPaddingEmptyByte(data []byte) []byte {
+	dataSize := len(data)
```
jianoaix, line 12: The len() will return int for dataSize, which indicates the following may not need uint64

encoding/utils/codec/codec.go
```diff
@@ -0,0 +1,66 @@
+package codec
+
+import (
+	"github.com/Layr-Labs/eigenda/encoding"
+)
+
+// ConvertByPaddingEmptyByte takes bytes and insert an empty at the front of every 31 byte
+// This ensure every 32 bytes are within the valid range of a field element for bn254 curve
```
jianoaix, line 8: 2^248 is only 2% of the field modulo, i suppose it doesn't pose a concern to use 2% of the space?

encoding/utils/codec/codec.go
```diff
@@ -0,0 +1,66 @@
...
+		end := (i + 1) * uint64(parseSize)
+		if end > uint64(len(data)) {
+			end = uint64(len(data))
+			// 1 is the empty byte
+			validEnd = end - start + 1 + i*putSize
+		}
+
+		// with big endian, set first byte is always 0 to ensure data is within valid range of
+		validData[i*encoding.BYTES_PER_COEFFICIENT] = 0x00
+		copy(validData[i*encoding.BYTES_PER_COEFFICIENT+1:(i+1)*encoding.BYTES_PER_COEFFICIENT], data[start:end])
```
jianoaix, line 32: Do we not need to pad leading zeros for the last coeff if it has less than 31 bytes?

encoding/utils/codec/codec.go
```diff
@@ -0,0 +1,66 @@
+package codec
+
+import (
+	"github.com/Layr-Labs/eigenda/encoding"
+)
+
+// ConvertByPaddingEmptyByte takes bytes and insert an empty at the front of every 31 byte
```
jianoaix, line 7: nit: an empty `byte` at...

disperser/apiserver/server.go
```diff
@@ -255,6 +256,13 @@ func (s *DispersalServer) disperseBlob(ctx context.Context, blob *core.Blob, aut
 		}
 	}
 
+	// validate every 32 bytes is a valid field element
+	_, err = rs.ToFrArray(blob.Data)
+	if err != nil {
+		s.logger.Error("failed to convert a 32bytes as a field element", "err", err)
+		return nil, api.NewInternalError("failed to convert bytes into a valid field element, please use a correct format which every 32bytes(big-endian) is less than 21888242871839275222246405745257275088548364400416034343698204186575808495617")
```
jianoaix, line 263: This looks an invalid arg error. Also please update the metrics before return

mooselumph, line 263: This should be `api.NewInvalidArgError`

encoding/constants.go
```diff
@@ -6,7 +6,7 @@ import (
 	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
 )
 
-const BYTES_PER_COEFFICIENT = 31
+const BYTES_PER_COEFFICIENT = 32
```
mooselumph, line 9: Maybe we should rename this to `BYTES_PER_SYMBOL`?

encoding/kzg/verifier/length_test.go
```diff
@@ -21,9 +21,10 @@ func TestLengthProof(t *testing.T) {
 
 	numBlob := 5
 	for z := 0; z < numBlob; z++ {
-		extra := make([]byte, z*31*2)
+		extra := make([]byte, z*32*2)
```
mooselumph, line 24: Use `BYTES_PER_SYMBOL` here?

encoding/rs/utils.go
```diff
@@ -20,15 +19,20 @@ func ToFrArray(data []byte) []fr.Element {
 		end := (i + 1) * uint64(encoding.BYTES_PER_COEFFICIENT)
 		if end >= uint64(len(data)) {
 			var padded [32]byte
```
mooselumph, line 21: Looks like padded can just be a slice here instead of an array. Then you can drop the `[:]`

encoding/utils/codec/codec.go
bxue-l2: It is a simple implementation, I think we can add more sophisticated methods later, or let others contribute.

encoding/utils/codec/codec.go
bxue-l2: no, we need to pad the leading zero, because a field element is interpreting with big endian where the low address contains the most significant bit. Padding the leading zero makes it consistent

api/proto/disperser/disperser.proto
```diff
@@ -71,7 +71,12 @@ message AuthenticationData {
 
 message DisperseBlobRequest {
 	// The data to be dispersed.
-	// The size of data must be <= 2MiB.
+	// The size of data must be <= 2MiB. Every 32 bytes of data chunk is interpreted as an integer in big endian format
+	// where the lower address has more significant bits. The integer must stays stay in the valid range to be interpreted
```
jianoaix, line 75: typo "stays stay"

disperser/apiserver/server.go
```diff
@@ -255,6 +256,13 @@ func (s *DispersalServer) disperseBlob(ctx context.Context, blob *core.Blob, aut
 		}
 	}
 
+	// validate every 32 bytes is a valid field element
+	_, err = rs.ToFrArray(blob.Data)
+	if err != nil {
+		s.logger.Error("failed to convert a 32bytes as a field element", "err", err)
```
jianoaix, line 262: Please call `s.metrics.HandleInvalidArgRpcRequest(apiMethodName)` here

disperser/apiserver/server.go
```diff
@@ -255,6 +256,13 @@ func (s *DispersalServer) disperseBlob(ctx context.Context, blob *core.Blob, aut
 		}
 	}
 
+	// validate every 32 bytes is a valid field element
+	_, err = rs.ToFrArray(blob.Data)
```
jianoaix, line 260: It's better to make this validation early so we can reject it early if it's invalid

disperser/batcher/encoding_streamer_test.go
```diff
@@ -152,7 +152,7 @@ func TestBatchTrigger(t *testing.T) {
 	assert.Nil(t, err)
 	count, size := encodingStreamer.EncodedBlobstore.GetEncodedResultSize()
 	assert.Equal(t, count, 1)
-	assert.Equal(t, size, uint64(15872))
+	assert.Equal(t, size, uint64(16384)) //Robert checks it
```
mooselumph, line 155: These are all fine.

disperser/apiserver/server.go
bxue-l2: moved to "validateRequestAndGetBlob" after rate limit, @mooselumph

tools/traffic/generator.go
```diff
@@ -62,6 +63,8 @@ func (g *TrafficGenerator) StartTraffic(ctx context.Context) error {
 		return err
 	}
 
+	data = codec.ConvertByPaddingEmptyByte(data)
```
mooselumph, line 66: Instead of putting this here, should we put this line in the client? 

https://github.com/bxue-l2/eigenda/blob/ca825a65d97f99fc92376bee78687c0431208782/clients/disperser_client.go#L85

and 

https://github.com/bxue-l2/eigenda/blob/ca825a65d97f99fc92376bee78687c0431208782/clients/disperser_client.go#L129-L130

mooselumph: On second thought, hardcoding a specific implementation into the client would be bad. We should allow more flexible usage of the client. Please disregard this ^ comment.

disperser/batcher/batcher_test.go
```diff
@@ -213,7 +214,7 @@ func TestBatcherIterations(t *testing.T) {
 	assert.NoError(t, err)
 	count, size := components.encodingStreamer.EncodedBlobstore.GetEncodedResultSize()
 	assert.Equal(t, 2, count)
-	assert.Equal(t, uint64(23808), size)
+	assert.Equal(t, uint64(24576), size) // Robert checks it
```
mooselumph, line 217: This is fine, please remove comment.

mooselumph (approved):
Looks good

disperser/apiserver/server.go
```diff
@@ -824,6 +825,14 @@ func (s *DispersalServer) validateRequestAndGetBlob(ctx context.Context, req *pb
 		return nil, errors.New("number of custom_quorum_numbers must not exceed 256")
 	}
 
+	// validate every 32 bytes is a valid field element
+	_, err := rs.ToFrArray(data)
+	if err != nil {
+		s.logger.Error("failed to convert a 32bytes as a field element", "err", err)
+		s.metrics.HandleInvalidArgRequest("DisperseBlob")
```
jianoaix, line 832: Since this validation is now called inside `validateRequestAndGetBlob`, this update to metric is not needed. The caller is already taking care of the errors in metric.

jianoaix (approved):

PR #434, opened 2024-04-02
Update metrics not to double count blob failures

ian-shim:
Right now, it's reporting retryable failures as `Failed` blobs. It should only be reported if it permanently failed dispersal.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/disperser.go
+++ b/disperser/disperser.go
@@ -165,7 +165,8 @@ type BlobStore interface {
 	// GetBlobMetadata returns a blob metadata given a metadata key
 	GetBlobMetadata(ctx context.Context, blobKey BlobKey) (*BlobMetadata, error)
 	// HandleBlobFailure handles a blob failure by either incrementing the retry count or marking the blob as failed
-	HandleBlobFailure(ctx context.Context, metadata *BlobMetadata, maxRetry uint) error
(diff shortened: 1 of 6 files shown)


disperser/disperser.go
```diff
@@ -165,7 +165,8 @@ type BlobStore interface {
 	// GetBlobMetadata returns a blob metadata given a metadata key
 	GetBlobMetadata(ctx context.Context, blobKey BlobKey) (*BlobMetadata, error)
 	// HandleBlobFailure handles a blob failure by either incrementing the retry count or marking the blob as failed
-	HandleBlobFailure(ctx context.Context, metadata *BlobMetadata, maxRetry uint) error
+	// Returns a boolean indicating whether the blob should be retried and an error
```
jianoaix, line 168: nit: It may be fine to omit subject but may need to keep it consistent with methods above

jianoaix (approved):

disperser/disperser.go
ian-shim: Can you clarify the suggestion here? This is consistent with methods above L141, L144

jianoaix: I see, there is "HandleBlobFailure handles" at L167

PR #435, opened 2024-04-02
mainnet deploy

0x0aa0:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


contracts/script/deploy/mainnet/Mainnet_Deploy.s.sol
```diff
@@ -0,0 +1,547 @@
...
+
+        __verifyImplementations();
+        __verifyInitalizations(config_data);
+
+        //write output
+        _writeOutput(config_data);
+    }
+
+    function xtest() external {
+        // get info on all the already-deployed contracts
```
ChaoticWalrus, line 235: can you make this call `run()` instead of copying the code? is the `require(deployer == tx.origin, ...)` check what's in the way right now?
I don't want to be too militant about code duplication but I find things to be much clearer + easier to review without copy-pasting 😄

ChaoticWalrus (approved):
This LGTM.
I don't think we're missing any config params, but am expecting we may need to make a late / post-deploy tweak or two.

ian-shim (commented):
I think it's still a good idea to regenerate bindings (even though none of the existing methods are affected)
wdyt?

0x0aa0:
> I think it's still a good idea to regenerate bindings (even though none of the existing methods are affected) wdyt?

agree

PR #438, opened 2024-04-03
feat(node): support execution without ecdsa key

jeffhubCB:

It is desirable to run an EigenDA Node without providing access to the Eigenlayer Node Operator ECDSA key. If EigenDA Node is started in a configuration that does not use this key, it should not require the configuration to still specify a key and attempt to decrypt it.

There are configuration options which drive a runtime-need for the ECDSA key.
1. opting-in at startup
2. automatic IP checking and updating

When started in a configuration that requires the ECDSA key but no key is provided, a startup error message results.
Screenshot 2024-04-03 at 9 26 32 AM
Screenshot 2024-04-03 at 9 27 46 AM

**NOTE** This only impacts node, not node-plugin. The ECDSA key is still required when running node-plugin commands.

## Checks

- [x] I've made sure the lint is passing in this PR. (golangci-lint github action command)
- Testing Strategy
   - [x] Unit tests (test-contracts and unit-tests github actions commands)
   - [ ] Integration tests

### DIFF
--- a/node/flags/flags.go
+++ b/node/flags/flags.go
@@ -100,7 +100,7 @@ var (
 	}
 	EcdsaKeyFileFlag = cli.StringFlag{
 		Name:     common.PrefixFlag(FlagPrefix, "ecdsa-key-file"),
-		Required: true,
+		Required: false,
 		Usage:    "Path to the encrypted ecdsa private key",
 		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "ECDSA_KEY_FILE"),
 	}
@@ -113,7 +113,7 @@ var (
 	}
 	EcdsaKeyPasswordFlag = cli.StringFlag{
 		Name:     common.PrefixFlag(FlagPrefix, "ecdsa-key-password"),
-		Required: true,
+		Required: false,
 		Usage:    "Password to decrypt ecdsa private key",
 		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "ECDSA_KEY_PASSWORD"),
 	}
@@ -244,9 +244,7 @@ var requiredFlags = []cli.Flag{
 	QuorumIDListFlag,
 	DbPathFlag,
 	BlsKeyFileFlag,
-	EcdsaKeyFileFlag,
 	BlsKeyPasswordFlag,
-	EcdsaKeyPasswordFlag,
 	BlsOperatorStateRetrieverFlag,
 	EigenDAServiceManagerFlag,
 	PubIPProviderFlag,
@@ -266,6 +264,8 @@ var optionalFlags = []cli.Flag{
 	InternalRetrievalPortFlag,
 	ClientIPHeaderFlag,
 	ChurnerUseSecureGRPC,
+	EcdsaKeyFileFlag,
+	EcdsaKeyPasswordFlag,
 }
 
 func init() {
--- a/node/config.go
+++ b/node/config.go
@@ -101,19 +101,32 @@ func NewConfig(ctx *cli.Context) (*Config, error) {
 
 	testMode := ctx.GlobalBool(flags.EnableTestModeFlag.Name)
 
-	// Decrypt ECDSA key
+	// Configuration options that require the Node Operator ECDSA key at runtime
+	registerNodeAtStart := ctx.GlobalBool(flags.RegisterAtNodeStartFlag.Name)
+	pubIPCheckInterval := ctx.GlobalDuration(flags.PubIPCheckIntervalFlag.Name)
+	needECDSAKey := registerNodeAtStart || pubIPCheckInterval > 0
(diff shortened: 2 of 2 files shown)


node/flags/flags.go
```diff
@@ -100,7 +100,7 @@ var (
 	}
 	EcdsaKeyFileFlag = cli.StringFlag{
 		Name:     common.PrefixFlag(FlagPrefix, "ecdsa-key-file"),
-		Required: true,
+		Required: false,
```
jianoaix, line 103: What if you just set some dummy keys for these two flags on the machine running the node, and then use the real keys on other machine to opt-in/out? That way it doesn't even need to make this change?

The `PubIPCheckIntervalFlag` can be set 0 do disable the IP checker. Then there will be no need of (real) keys for node.

jeffhubCB: Setting it to a dummy key is certainly an option, but I believe supporting this as a valid runtime option, versus expecting users that want to run in this configuration to know they can generate a dummy  ECDSA key  and use that, seems like a better user experience.
Additionally, by not providing a key at all it ensures failure on startup  if the key is needed, versus failure at a later point while running when the provided dummy  key (to get eigenda node to start, because it won't otherwise) tries to use that key.

jeffhubCB: @jianoaix Dummy keys is a confusing concept and continues the "standard" that is being demonstrated to the AVS community by EigenDA that the ECDSA key is _required_ at runtime. Generally, my understanding is that an Operator can associate with any number of AVSs which are run separately - this arrangement should not require privileged access to the privkeys of the Operator at all. More than that though, if it did require access to the Operators private keys then the opt in/out process would be unnecessary. Please clarify if that was not the intent of the Operator AVS relationship.

In the context then that Operators and AVSs are separately owned entities, I would suggest that providing the Operator privkey be an _option_ rather than a default. Even then this is not really a clear separation of ownership nor does it communicate the intended relationships. There are lifecycle operations needed to register/deregister an AVS that is consistent with all AVSes. Those  operations require the Operator private key, but that does  not equate to "the AVS should always have access to the Operator private key in its default running state". The separation and usage of the Operator ECDSA key, versus the AVS BLS key, should be much clearer than the current implementation demonstrates.

I believe `node` requires the private key today because of the way  the codebase was originally structured, as well as the auto-socket-update (which can easily be disabled by default, with docs explaining how to enable it and the implications... gas fees, requirement to provide the ECDSA key to node w/ those security implications, etc). There are shared code paths between node, node-plugin, retriever and disperser. I believe only 2 of these _need_ the ECDSA key... node-plugin (for opt-in/update-socket) and disperser. Because of this  setup, the  code was written to assume the ECDSA key is _always_ required, leading to a situation where additional code can readily be added to leverage it at runtime when that  should not be encouraged.

The scope of this PR wasn't intended to clean all of that up and formalize it more, it was an initial step so that the EigenDA Node Operator can be run without providing the Node Operator ECDSA key (or a dummy key), to increase security posture and ideally encourage further changes,  in both documentation and  implementation, that will enable  a more secure deployment pattern and encourage  the rest of the AVS community to do the same.

jianoaix: These are valid points. Note I'm not against having a proper solution. I'm suggesting a way to workaround to quickly unblock the use case you are looking for.

jianoaix (commented):
The implementation lgtm. Did you get a chance to test?

jeffhubCB:
> The implementation lgtm. Did you get a chance to test?

I tested locally as much as I was able. I did not modify the inabox setup which would be a complete validation. I'm working on getting enough stake to be in the active set in Holesky, but that's unlikely to happen until tomorrow.

The screenshots of the startup error message were taken by using eigenda-operator-setup.
To remove the ECDSA key, the following modifications were made:
Remove 
https://github.com/Layr-Labs/eigenda-operator-setup/blob/master/holesky/docker-compose.yml#L32
https://github.com/Layr-Labs/eigenda-operator-setup/blob/master/holesky/.env.example#L32
https://github.com/Layr-Labs/eigenda-operator-setup/blob/master/holesky/.env.example#L92
https://github.com/Layr-Labs/eigenda-operator-setup/blob/master/holesky/.env.example#L100

In this configuration, startup will fail with the following message:
```
application failed: node.ecdsa-key-file and node.ecdsa-key-password are required if node.public-ip-check-interval is > 0
```

Update this line to 0: https://github.com/Layr-Labs/eigenda-operator-setup/blob/master/holesky/.env.example#L65
In this configuration, startup will succeed.

Add this to .env:
`NODE_REGISTER_AT_NODE_START=true`
In this configuration, startup will fail with the following message:
```
application failed: node.ecdsa-key-file and node.ecdsa-key-password are required if node.register-at-node-start is enabled
```

The gap in testing is that at no point was my test node in the active set.

jianoaix (approved):
Thanks for contribution!

PR #450, opened 2024-04-05
batcher allowlist

0x0aa0:

We need to allowlist a fallback batcher

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


contracts/src/core/EigenDAServiceManagerStorage.sol
```diff
@@ -53,8 +53,8 @@ abstract contract EigenDAServiceManagerStorage is IEigenDAServiceManager {
     /// @notice mapping between the batchId to the hash of the metadata of the corresponding Batch
     mapping(uint32 => bytes32) public batchIdToBatchMetadataHash;
 
-    /// @notice address that is permissioned to confirm batches
-    address public batchConfirmer;
+    /// @notice mapping of addressed that are permissioned to confirm batches
+    mapping(address => bool) public isBatchConfirmer;
```
ChaoticWalrus, line 57: this is changing in the same storage slot. will this break testnet deployment when upgrading?
I guess when upgrading we can just:

1. change the confirmer to the zero address 
2. upgrade
3. set batch confirmer

contracts/src/core/EigenDAServiceManager.sol
```diff
@@ -128,9 +130,8 @@ contract EigenDAServiceManager is EigenDAServiceManagerStorage, ServiceManagerBa
 
     /// @notice changes the batch confirmer
     function _setBatchConfirmer(address _batchConfirmer) internal {
-        address previousBatchConfirmer = batchConfirmer;
-        batchConfirmer = _batchConfirmer;
-        emit BatchConfirmerChanged(previousBatchConfirmer, batchConfirmer);
+        isBatchConfirmer[_batchConfirmer] = !isBatchConfirmer[_batchConfirmer];
+        emit BatchConfirmerStatusChanged(_batchConfirmer, isBatchConfirmer[_batchConfirmer]);
```
ChaoticWalrus, line 134: I'm not a huge fan of this pattern, but it's OK with me

ChaoticWalrus (approved):
Approving this small change.
Adding test coverage for new behavior would be good, but I understand that's not the priority right now.

Noting this one storage slot change which may be relevant for testnet https://github.com/Layr-Labs/eigenda/pull/450/files#r1554183962

PR #452, opened 2024-04-05
fix blob size displayed at sendchunks

bxue-l2:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/batcher/grpc/dispatcher.go
+++ b/disperser/batcher/grpc/dispatcher.go
@@ -144,7 +144,7 @@ func GetStoreChunksRequest(blobMessages []*core.BlobMessage, batchHeader *core.B
 		if err != nil {
 			return nil, 0, err
 		}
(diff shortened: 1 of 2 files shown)


disperser/batcher/grpc/dispatcher.go
```diff
@@ -144,7 +144,7 @@ func GetStoreChunksRequest(blobMessages []*core.BlobMessage, batchHeader *core.B
 		if err != nil {
 			return nil, 0, err
 		}
-		totalSize += blob.BlobHeader.EncodedSizeAllQuorums()
+		totalSize += int64(blob.Bundles.Size())
```
ian-shim, line 147: did you want to update the logging in L122 too?

bxue-l2: I think it is more useful to get workload per node, it is currently printing total blob size.

ian-shim (approved):

PR #457, opened 2024-04-08
update blob index size

0x0aa0:

There may be more than uint8.max blobs in a batch

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


contracts/src/libraries/EigenDARollupUtils.sol
```diff
@@ -18,7 +18,7 @@ library EigenDARollupUtils {
     // STRUCTS
     struct BlobVerificationProof {
         uint32 batchId;
-        uint8 blobIndex;
+        uint256 blobIndex;
```
gpsanant, line 21: uint32 should be fine.

also a regression test would be good as a pending issue

gpsanant (approved):

contracts/src/libraries/EigenDARollupUtils.sol
jianoaix: yeah, offchain is also uint32: https://github.com/Layr-Labs/eigenda/blob/master/api/proto/disperser/disperser.proto#L190

teddyknox (approved):

ian-shim (approved):

PR #460, opened 2024-04-08
Doc: byte data need to be base64 encoded

jianoaix:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/proto/disperser/disperser.proto
+++ b/api/proto/disperser/disperser.proto
@@ -71,12 +71,12 @@ message AuthenticationData {
 
 message DisperseBlobRequest {
 	// The data to be dispersed.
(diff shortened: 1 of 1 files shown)


teddyknox (approved):

bxue-l2 (commented):
We shouldn't need this PR

bxue-l2:
We noticed some issues to query disperser grpc endpoint using `grpcul`. The error states 
`Error invoking method "disperser.Disperser/DisperseBlob": error getting request data: illegal base64 data at input byte 0`. 

This issue happens only because grpcurl encodes "byte" in proto using base64 encoding. See this https://github.com/fullstorydev/grpcurl/issues/141 .

PR #464, opened 2024-04-09
Query non-signers between start and stop times. 

mooselumph:

Allows for investigation of non-signer statistics over arbitrary intervals. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/dataapi/docs/docs.go
+++ b/disperser/dataapi/docs/docs.go
@@ -374,6 +374,18 @@ const docTemplate = `{
                         "description": "Interval to query for operators nonsigning percentage [default: 3600]",
                         "name": "interval",
                         "in": "query"
+                    },
+                    {
+                        "type": "string",
+                        "description": "Start time (UTC) to query for operators nonsigning percentage",
+                        "name": "start",
+                        "in": "query"
+                    },
+                    {
+                        "type": "string",
+                        "description": "End time (UTC) to query for operators nonsigning percentage",
+                        "name": "stop",
+                        "in": "query"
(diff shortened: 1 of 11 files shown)


disperser/dataapi/docs/docs.go
```diff
@@ -645,12 +660,16 @@ const docTemplate = `{
                 },
                 "total_stake": {
                     "description": "deprecated: use TotalStakePerQuorum instead. Remove when the frontend is updated.",
-                    "type": "integer"
+                    "allOf": [
```
mooselumph, line 663: I'm not sure what happened here. Perhaps I'm using a different version of swaggo?

disperser/dataapi/server.go
```diff
@@ -465,7 +465,9 @@ func (s *server) FetchNonSigners(c *gin.Context) {
 //	@Summary	Fetch operators non signing percentage
 //	@Tags		Metrics
 //	@Produce	json
-//	@Param		interval	query		int	false	"Interval to query for operators nonsigning percentage [default: 3600]"
+//	@Param		interval	query		int		false	"Interval to query for operators nonsigning percentage [default: 3600]"
```
jianoaix, line 468: It may be simpler to just have 2 params: endtime and interval. When endtime is missing, set it to time.Now(). Then it'll be also be backward compatible.

mooselumph: That seems like a good suggestion. I'll go ahead and update.

PR #465, opened 2024-04-10
Isolate transaction broadcasting latency from regas logic

ian-shim:
Currently, we send a replacement transaction if an existing transaction doesn't get confirmed within predetermined timeout. This duration is measured from the time `TxnManager` sends the transaction. If the transaction is published directly on eth network, this would have been a good measure to detect if the transaction is stuck on mempool and if a replacement transaction with higher gas is necessary.

However, when a transaction is submitted via Fireblocks, this is not a good measure to determine if the transaction is stuck on mempool because it takes non-trivial amount of time for Fireblocks to publish the txn to mempool due to multiparty computation. 

This PR breaks down this part of latency (pre-broadcast), and excludes this duration from the timeout watched to create a replacement transaction. For example, if it takes a long time to broadcast the transaction to mempool (due to MPC operation from Fireblocks), no replacement transactions should be created. Only if it takes a long time to confirm transaction onchain after it's been published in mempool, it creates a replacement transaction with higher gas. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/batcher/txn_manager.go
+++ b/disperser/batcher/txn_manager.go
@@ -169,11 +169,11 @@ func (t *txnManager) ProcessTransaction(ctx context.Context, req *TxnRequest) er
 			didTimeout = urlErr.Timeout()
 		}
 		if didTimeout || errors.Is(err, context.DeadlineExceeded) {
-			t.logger.Warn("failed to send txn due to timeout", "tag", req.Tag, "hash", req.Tx.Hash().Hex(), "numRetries", retryFromFailure, "maxRetry", maxSendTransactionRetry, "err", err)
+			t.logger.Warn("failed to send txn due to timeout", "tag", req.Tag, "hash", txn.Hash().Hex(), "numRetries", retryFromFailure, "maxRetry", maxSendTransactionRetry, "err", err)
 			retryFromFailure++
 			continue
 		} else if err != nil {
-			return fmt.Errorf("failed to send txn (%s) %s: %w", req.Tag, req.Tx.Hash().Hex(), err)
+			return fmt.Errorf("failed to send txn (%s) %s: %w", req.Tag, txn.Hash().Hex(), err)
 		} else {
 			t.logger.Debug("successfully sent txn", "tag", req.Tag, "txID", txID, "txHash", txn.Hash().Hex())
 			break
@@ -314,6 +314,9 @@ func (t *txnManager) monitorTransaction(ctx context.Context, req *TxnRequest) (*
 				}
 			}
 			return ErrTransactionNotBroadcasted
+		} else if err != nil {
+			t.logger.Error("unexpected error while waiting for Fireblocks transaction to broadcast", "txHash", req.Tx.Hash().Hex(), "err", err)
+			return err
 		}
 
 		ctxWithTimeout, cancelEvaluationTimeout := context.WithTimeout(ctx, t.txnRefreshInterval)
--- a/disperser/batcher/txn_manager.go
+++ b/disperser/batcher/txn_manager.go
@@ -22,6 +22,8 @@ var (
 	gasPricePercentageMultiplier = big.NewInt(10)
 	hundred                      = big.NewInt(100)
 	maxSendTransactionRetry      = 3
+	queryTickerDuration          = 3 * time.Second
+	ErrTransactionNotBroadcasted = errors.New("transaction not broadcasted")
 )
 
 // TxnManager receives transactions from the caller, sends them to the chain, and monitors their status.
@@ -36,7 +38,8 @@ type TxnManager interface {
 
 type transaction struct {
 	*types.Transaction
-	TxID walletsdk.TxID
+	TxID        walletsdk.TxID
+	requestedAt time.Time
 }
 
 type TxnRequest struct {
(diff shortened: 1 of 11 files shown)


disperser/batcher/txn_manager.go
```diff
@@ -192,8 +198,31 @@ func (t *txnManager) ReceiptChan() chan *ReceiptOrErr {
...
 
+// ensureAnyTransactionBroadcasted waits until all given transactions are broadcasted to the network.
+func (t *txnManager) ensureAnyTransactionBroadcasted(ctx context.Context, txs []*transaction) error {
+	queryTicker := time.NewTicker(queryTickerDuration)
+	defer queryTicker.Stop()
+
+	for {
+		for _, tx := range txs {
+			_, err := t.wallet.GetTransactionReceipt(ctx, tx.TxID)
+			if err == nil || errors.Is(err, walletsdk.ErrReceiptNotYetAvailable) {
```
mooselumph, line 209: Checking error types seems like it's a bit brittle. I guess there's no other option here?

disperser/batcher/txn_manager.go
```diff
@@ -249,10 +285,37 @@ func (t *txnManager) monitorTransaction(ctx context.Context, req *TxnRequest) (*
...
+
+		// Ensure transactions are broadcasted to the network before querying the receipt.
+		// This is to avoid querying the receipt of a transaction that hasn't been broadcasted yet.
+		// For example, when Fireblocks wallet is used, there may be delays in broadcasting the transaction due to latency from cosigning and MPC operations.
+		err = t.ensureAnyTransactionBroadcasted(ctxWithTimeout, req.txAttempts)
+		if err != nil && errors.Is(err, context.DeadlineExceeded) {
+			t.logger.Warn("transaction not broadcasted within timeout", "tag", req.Tag, "txHash", req.Tx.Hash().Hex(), "nonce", req.Tx.Nonce())
+			fireblocksWallet, ok := t.wallet.(interface {
+				CancelTransactionBroadcast(ctx context.Context, txID walletsdk.TxID) (bool, error)
+			})
```
mooselumph, line 301: Why doesn't the interface have this to begin with?

mooselumph (commented):
Should we being retrying the fireblocks request if that times out? Or would we just set a long timeout for this? It looks like the current behavior is just to fail?

disperser/batcher/txn_manager.go
ian-shim: We can change the implementation in eigensdk, but I think this signature should be kept. 
One thing we can do is make the error type more expressive and return Fireblocks status as part of the error, but not sure if that will be less brittle

ian-shim:
> Should we being retrying the fireblocks request if that times out? Or would we just set a long timeout for this? It looks like the current behavior is just to fail?

I think we should just set a long timeout for this. If the mpc operation latency is extended, it's probably not transient and the retry will likely encounter similar issue. The current behavior is that it will "cancel" the transaction and fail.

disperser/batcher/txn_manager.go
mooselumph: Maybe add a comment: "Before the transaction is broadcast, this wallet method will return error of type `walletsdk.Whatever`. Once the transaction has been broadcast, it will either successfully return the receipt or return an error of type `walletsdk.ErrReceiptNotYetAvailable`."

disperser/batcher/txn_manager.go
```diff
@@ -251,10 +287,37 @@ func (t *txnManager) monitorTransaction(ctx context.Context, req *TxnRequest) (*
...
-		t.logger.Debug("monitoring transaction", "component", "TxnManager", "method", "monitorTransaction", "txHash", req.Tx.Hash().Hex(), "tag", req.Tag, "nonce", req.Tx.Nonce())
+		t.logger.Debug("monitoring transaction", "txHash", req.Tx.Hash().Hex(), "tag", req.Tag, "nonce", req.Tx.Nonce())
+
+		ctxWithTimeout, cancelBroadcastTimeout := context.WithTimeout(ctx, t.txnBroadcastTimeout)
+		defer cancelBroadcastTimeout()
+
+		// Ensure transactions are broadcasted to the network before querying the receipt.
+		// This is to avoid querying the receipt of a transaction that hasn't been broadcasted yet.
+		// For example, when Fireblocks wallet is used, there may be delays in broadcasting the transaction due to latency from cosigning and MPC operations.
+		err = t.ensureAnyTransactionBroadcasted(ctxWithTimeout, req.txAttempts)
```
mooselumph, line 298: We don't retry here because there wouldn't be any advantage over just waiting longer?

disperser/batcher/txn_manager.go
```diff
@@ -251,10 +287,37 @@ func (t *txnManager) monitorTransaction(ctx context.Context, req *TxnRequest) (*
...
+		if err != nil && errors.Is(err, context.DeadlineExceeded) {
```
mooselumph, line 299: It looks like the error is unhandled if it isn't equal to `context.DeadlineExceeded`?

mooselumph (commented):
A couple more comments

disperser/batcher/txn_manager.go
ian-shim: Yeah as explained in [this comment](https://github.com/Layr-Labs/eigenda/pull/465#issuecomment-2050923521)

disperser/batcher/txn_manager.go
ian-shim: Yeah I think this is the only error that can be returned, but agree it should handle other errors too

mooselumph (approved):

PR #466, opened 2024-04-10
Add `Confirming` state for blobs

ian-shim:
At the high level, this PR adds a new blob status called `Confirming` to indicate that it's awaiting `ConfirmBatch` transaction to be confirmed onchain. 
To make this update not a breaking change for rollups, this status will remain internal, and blobs in this status will show as `Processing` status as before. 

The main motivations for this PR are to 
- simplify and reduce complexity for in-memory accounting for blob status logic
- make memory usage more efficient by flushing dispersed encoded blobs immediately after transaction has been created
- fix the accounting issue for total encoded blob size and enables the `EncodedSizeNotifier` to properly trigger new batch iteration.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests


disperser/batcher/finalizer.go
```diff
@@ -179,7 +179,8 @@ func (f *finalizer) updateBlobs(ctx context.Context, metadatas []*disperser.Blob
 		confirmationBlockNumber, err := f.getTransactionBlockNumber(ctx, confirmationMetadata.ConfirmationInfo.ConfirmationTxnHash)
 		if errors.Is(err, ethereum.NotFound) {
 			// The confirmed block is finalized, but the transaction is not found. It means the transaction should be considered forked/invalid and the blob should be considered as failed.
-			_, err := f.blobStore.HandleBlobFailure(ctx, m, f.maxNumRetriesPerBlob)
+			f.logger.Warn("confirmed transaction not found", "blobKey", blobKey.String(), "confirmationTxnHash", confirmationMetadata.ConfirmationInfo.ConfirmationTxnHash.Hex(), "confirmationBlockNumber", confirmationMetadata.ConfirmationInfo.ConfirmationBlockNumber)
+			err := f.blobStore.MarkBlobFailed(ctx, m.GetBlobKey())
```
ian-shim, line 183: From conversation with @teddyknox , it seems like marking a blob as failed in case of reorg like this is more appropriate than retrying dispersal

disperser/apiserver/server.go
```diff
@@ -839,7 +839,7 @@ func (s *DispersalServer) updateQuorumConfig(ctx context.Context) (QuorumConfig,
 
 func getResponseStatus(status disperser.BlobStatus) pb.BlobStatus {
 	switch status {
-	case disperser.Processing:
+	case disperser.Confirming, disperser.Processing:
```
mooselumph, line 842: Why not return the CONFIRMING status?

disperser/batcher/batcher.go
```diff
@@ -493,9 +489,12 @@ func (b *Batcher) HandleSingleBatch(ctx context.Context) error {
 		return fmt.Errorf("HandleSingleBatch: error sending confirmBatch transaction: %w", err)
 	} else {
 		for _, metadata := range batch.BlobMetadata {
-			err = b.EncodingStreamer.MarkBlobPendingConfirmation(metadata)
+			err = b.Queue.MarkBlobConfirming(ctx, metadata.GetBlobKey())
```
mooselumph, line 492: I'm confused about the error handling. Here we check if marking the blobs in blobstore is successful before removing the encoded blobs. In `handleFailure` (e.g. L488), we remove the encoded blobs first and then update the blobstore.  Have we thought through the different states that the system can be in depending on these different failure scenarios?

ian-shim: This is a good point. The encoded blob should probably be removed only if the update is successful.
The difference is whether the blob gets encoded again or not, which isn't super critical but has important performance implications.

disperser/apiserver/server.go
ian-shim: That would be a breaking change for the existing rollup integrations

mooselumph (approved):
looks good, a few more thoughts

disperser/batcher/finalizer.go
ian-shim: I think there are few edge cases to consider here.
1. Reorg unconfirms the blob and is marked as `Failed` before the sequencer retrieves the blob status: The sequencer will see the blob as `Failed` and redisperse the blob. 
2. Reorg unconfirms the blob. The sequencer has already retrieved the confirmed blob status. The sequencer submits inbox transaction.
  a. The inbox transaction gets reorg'd as well. Rollup redisperses the same data. 
  b. The inbox transaction successfully makes it to the chain. Now it doesn't matter if the old blob gets retried or is marked as `Failed`. Need DA cert verification.

I think the finalizer is still necessary to be the source of truth though.

disperser/batcher/batcher.go
ian-shim: I think that's a good idea. 
I made `transitionBlobToConfirming` but kept the transitions to `Failed` and `Confirmed` as they were because those transitions only update blob store.

PR #471, opened 2024-04-11
Add retry for graph node

mooselumph:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/thegraph/config.go
+++ b/core/thegraph/config.go
@@ -1,0 +1,53 @@
+package thegraph
+
+import (
+	"time"
+
+	"github.com/Layr-Labs/eigenda/common"
+	"github.com/urfave/cli"
+)
+
+const (
+	EndpointFlagName     = "thegraph.endpoint"
+	PullIntervalFlagName = "thegraph.pull_interval"
+	MaxRetriesFlagName   = "thegraph.max_retries"
+)
+
+type Config struct {
+	Endpoint     string        // The Graph endpoint
+	PullInterval time.Duration // The interval to pull data from The Graph
+	MaxRetries   int           // The maximum number of retries to pull data from The Graph
+}
+
+func CLIFlags(envPrefix string) []cli.Flag {
+	return []cli.Flag{
+		cli.StringFlag{
+			Name:     EndpointFlagName,
+			Usage:    "The Graph endpoint",
+			Required: true,
+			EnvVar:   common.PrefixEnvVar(envPrefix, "GRAPH_URL"),
+		},
+		cli.DurationFlag{
+			Name:   PullIntervalFlagName,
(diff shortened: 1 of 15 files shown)


core/thegraph/querier.go
```diff
@@ -0,0 +1,46 @@
...
+				return errors.New("max retries exceeded")
+			}
+			retryCount++
+
+			err := q.GraphQLQuerier.Query(ctx, query, variables)
+			if err == nil {
+				return nil
+			}
+			// Optionally, add a delay before retrying
+			time.Sleep(q.PullInterval)
```
jianoaix, line 43: This may need exponential backoff

core/thegraph/config.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"time"
+
+	"github.com/Layr-Labs/eigenda/common"
+	"github.com/urfave/cli"
+)
+
+const (
+	EndpointFlagName     = "thegraph.endpoint"
+	PullIntervalFlagName = "thegraph.pull_interval"
```
ian-shim, line 12: nit: `thegraph.backoff`?

core/thegraph/querier.go
```diff
@@ -0,0 +1,46 @@
...
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		default:
+			if retryCount > q.MaxRetries {
+				return errors.New("max retries exceeded")
+			}
+			retryCount++
+
+			err := q.GraphQLQuerier.Query(ctx, query, variables)
```
ian-shim, line 38: Maybe in a future PR, but it should check `err` and only retry on transient errors (429/5xx). If the request itself is invalid, retries shouldn't be made.

bxue-l2 (approved):

core/thegraph/querier.go
mooselumph: I switched it to an exponential backoff

core/thegraph/querier.go
mooselumph: I'm not really sure how to tell in this case. Going to leave this for a future PR.

core/thegraph/querier_test.go
```diff
@@ -0,0 +1,75 @@
...
+type MockGraphQLQuerier struct {
+	mock.Mock
+}
+
+func (m *MockGraphQLQuerier) Query(ctx context.Context, query any, variables map[string]any) error {
+	args := m.Called(ctx, query, variables)
+	return args.Error(0)
+}
+
+func TestRetryQuerier_Query(t *testing.T) {
```
jianoaix, line 23: Is it conventional to use underscore in golang naming?

jianoaix (approved):

PR #473, opened 2024-04-11
Detect wrapped url error timeout

ian-shim:
Current way of detecting url error with type assertion doesn't work if the `err` being matched is wrapping url error. 
This PR updates it so that it matches a url error from the error tree and then calling `Timeout` from the matched instance.

playground: https://goplay.tools/snippet/reDY2KH7bUD

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/batcher/encoding_streamer.go
+++ b/disperser/batcher/encoding_streamer.go
@@ -137,7 +137,11 @@ func (e *EncodingStreamer) Start(ctx context.Context) error {
 						// ignore canceled errors because canceled encoding requests are normal
 						continue
 					}
-					e.logger.Error("error processing encoded blobs", "err", err)
+					if strings.Contains(err.Error(), "too many requests") {
+						e.logger.Warn("encoding request ratelimited", "err", err)
+					} else {
+						e.logger.Error("error processing encoded blobs", "err", err)
+					}
 				}
 			}
 		}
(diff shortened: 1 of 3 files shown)


disperser/batcher/encoding_streamer.go
```diff
@@ -137,7 +137,11 @@ func (e *EncodingStreamer) Start(ctx context.Context) error {
 						// ignore canceled errors because canceled encoding requests are normal
 						continue
 					}
-					e.logger.Error("error processing encoded blobs", "err", err)
+					if strings.Contains(err.Error(), "too many requests") {
+						e.logger.Warn("error processing encoded blobs", "err", err)
```
ian-shim, line 141: downgraded this case as it happens quite often and it's safe to ignore

bxue-l2 (approved):

disperser/batcher/encoding_streamer.go
pschork: Maybe remove `error` from the Warning message. So what is the source of `too many requests`? Are we being rate limited when this occurs?

disperser/batcher/encoding_streamer.go
```diff
@@ -137,7 +137,11 @@ func (e *EncodingStreamer) Start(ctx context.Context) error {
...
+					} else {
+						e.logger.Error("error processing encoded blobs", "err", err)
```
pschork, line 143: I feel like it could be confusing to use the same message for a Warning (which is safe to ignore) vs Error which are not safe to ignore. Can we change the error message in a way that is more provocative call to action.

ian-shim: Updated the message for warning

PR #481, opened 2024-04-13
increase churner expiry

bxue-l2:
## Checks

- [ ] I've made sure the lint is passing in this PR.
- [ ] I've made sure the tests are passing. Note that there might be a few flaky tests, in that case, please comment that they are not relevant.
- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests
   - [ ] This PR is not tested :(

PR #482, opened 2024-04-15
Backward compatibility with using entire g2

bxue-l2:

This is a correction for PR #472 which checks g2powerof2 file before starting node. However, #472 prevents a legacy usage that the operator might only have the entire g2 file.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/encoding/kzg/verifier/verifier.go
+++ b/encoding/kzg/verifier/verifier.go
@@ -70,21 +70,24 @@ func NewVerifier(config *kzg.KzgConfig, loadG2Points bool) (*Verifier, error) {
 			return nil, err
 		}
 	} else {
-		if len(config.G2PowerOf2Path) == 0 {
-			return nil, errors.New("G2PowerOf2Path is empty. However, object needs to load G2Points")
+		if len(config.G2PowerOf2Path) == 0 && len(config.G2Path) == 0 {
+			return nil, errors.New("both G2Path and G2PowerOf2Path are empty. However, object needs to load G2Points")
 		}
 
-		if config.SRSOrder == 0 {
-			return nil, errors.New("SRS order cannot be 0")
-		}
+		if len(config.G2PowerOf2Path) != 0 {
+			if config.SRSOrder == 0 {
+				return nil, errors.New("SRS order cannot be 0")
+			}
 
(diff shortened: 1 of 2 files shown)


encoding/kzg/verifier/verifier.go
```diff
@@ -70,21 +70,24 @@ func NewVerifier(config *kzg.KzgConfig, loadG2Points bool) (*Verifier, error) {
 			return nil, err
 		}
 	} else {
-		if len(config.G2PowerOf2Path) == 0 {
-			return nil, errors.New("G2PowerOf2Path is empty. However, object needs to load G2Points")
+		if len(config.G2PowerOf2Path) == 0 && len(config.G2Path) == 0 {
```
jianoaix, line 73: The config.G2Path isn't used below, although it's checking it here.
What if the control flow is `if loadG2Points || len(config.G2PowerOf2Path) == 0 { ... }` above? This entire block won't need to change

operators/churner/churner.go
```diff
@@ -21,7 +21,7 @@ import (
 
 var (
 	bipMultiplier     = big.NewInt(10000)
-	secondsTillExpiry = 90 * time.Second
+	secondsTillExpiry = 3600 * time.Second
```
jianoaix, line 24: Why extend to 3600s?

bxue-l2: 90sec is too stringent for some operators to opt in, especially smart contract operator. though it is already merged from #481

encoding/kzg/verifier/verifier.go
bxue-l2: It is used to ensure backward compatibility for operators who uses the entire G2 points, though very unlikely. In the last PR #472 , node is required to read the powerOf2 file at start. However, it prevents the legacy usecase when the node is accessing the entire points.

bxue-l2: node sets loadG2Points to be false.

jianoaix: I think my suggestion has the same logic, but simpler code?

jianoaix (approved):

PR #501, opened 2024-04-19
Optimize the quorum bitmap lookup when there is just one operator

jianoaix:
We are going to add operator registration metric at Node. When operator is querying their own metrics from the chain, we'd want to reduce the number of RPC calls needed.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/eth/tx.go
+++ b/core/eth/tx.go
@@ -588,6 +588,27 @@ func (t *Transactor) GetCurrentQuorumBitmapByOperatorId(ctx context.Context, ope
 }
 
 func (t *Transactor) GetQuorumBitmapForOperatorsAtBlockNumber(ctx context.Context, operatorIds []core.OperatorID, blockNumber uint32) ([]*big.Int, error) {
+	if len(operatorIds) == 0 {
+		return []*big.Int{}, nil
+	}
+	// When there is just one operator, we can get result by a single RPC with
+	// getQuorumBitmapsAtBlockNumber() in OperatorStateRetrievercontract (v.s. 2
+	// RPCs in the general case)
+	if len(operatorIds) == 1 {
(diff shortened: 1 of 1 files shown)


0x0aa0 (approved):

core/eth/tx.go
```diff
@@ -588,6 +588,27 @@ func (t *Transactor) GetCurrentQuorumBitmapByOperatorId(ctx context.Context, ope
...
+	// When there is just one operator, we can get result by a single RPC with
+	// getQuorumBitmapsAtBlockNumber() in OperatorStateRetrievercontract (v.s. 2
+	// RPCs in the general case)
+	if len(operatorIds) == 1 {
+		byteId := [32]byte(operatorIds[0])
+		bitmap, err := t.Bindings.OpStateRetriever.GetQuorumBitmapsAtBlockNumber(&bind.CallOpts{
+			Context: ctx,
+		}, t.Bindings.RegCoordinatorAddr, [][32]byte{byteId}, blockNumber)
+		if err != nil {
+			if err.Error() == "RegistryCoordinator.getQuorumBitmapIndexAtBlockNumber: no bitmap update found for operatorId at block number" {
```
ian-shim, line 603: does this equality match the err properly? have we checked this is the exact string that gets returned for this error?

jianoaix: I think @0x0aa0 is helping check the details, but haven't tested yet

jianoaix: Ok, I'm sure it's "execution reverted: RegistryCoordinator.getQuorumBitmapIndexAtBlockNumber: no bitmap update found for operatorId at block number"

ian-shim (approved):

PR #502, opened 2024-04-19
Add operator registration metric at node

jianoaix:
Provide metrics to help address major operator pain points

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/node/metrics.go
+++ b/node/metrics.go
@@ -58,8 +58,8 @@ func NewMetrics(eigenMetrics eigenmetrics.Metrics, reg *prometheus.Registry, log
 	reg.MustRegister(collectors.NewGoCollector())
 
 	metrics := &Metrics{
-		// The "type" label have values: stake, rank. The "stake" is stake share (in basis point),
-		// and the "rank" is operator's ranking by stake share in the quorum.
+		// The "type" label have values: stake_share, rank. The "stake_share" is stake share (in basis point),
+		// and the "rank" is operator's ranking (the operator with highest amount of stake ranked as 1) by stake share in the quorum.
 		Registered: promauto.With(reg).NewGaugeVec(
 			prometheus.GaugeOpts{
 				Namespace: Namespace,
@@ -184,17 +184,17 @@ func (g *Metrics) collectOnchainMetrics() {
 		}
 		bitmaps, err := g.tx.GetQuorumBitmapForOperatorsAtBlockNumber(ctx, []core.OperatorID{g.operatorId}, blockNum)
 		if err != nil {
-			g.logger.Error("Failed to query chain RPC for quorum bitmap", "err", err)
+			g.logger.Error("Failed to query chain RPC for quorum bitmap", "blockNumber", blockNum, "err", err)
 			continue
 		}
-		quorums := eth.BitmapToQuorumIds(bitmaps[0])
-		if len(quorums) == 0 {
-			g.logger.Info("Warning: this node is no longer in any quorum", "blockNumber", blockNum, "operatorId", g.operatorId.Hex())
+		quorumIds := eth.BitmapToQuorumIds(bitmaps[0])
+		if len(quorumIds) == 0 {
+			g.logger.Warn("This node is currently not in any quorum", "blockNumber", blockNum, "operatorId", g.operatorId.Hex())
 			continue
 		}
-		state, err := g.chainState.GetOperatorState(ctx, uint(blockNum), quorums)
+		state, err := g.chainState.GetOperatorState(ctx, uint(blockNum), quorumIds)
 		if err != nil {
-			g.logger.Error("Failed to query chain RPC for operator state", "blockNumber", blockNum, "quorumIds", quorums, "err", err)
+			g.logger.Error("Failed to query chain RPC for operator state", "blockNumber", blockNum, "quorumIds", quorumIds, "err", err)
 			continue
 		}
 		type OperatorStakeShare struct {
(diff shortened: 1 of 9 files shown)


node/metrics.go
```diff
@@ -150,3 +169,59 @@ func (g *Metrics) AcceptBatches(status string, batchSize uint64) {
...
+			// Descending order by stake share in the quorum.
+			sort.Slice(operatorStakeShares, func(i, j int) bool {
+				if operatorStakeShares[i].stakeShare == operatorStakeShares[j].stakeShare {
+					return operatorStakeShares[i].operatorId.Hex() < operatorStakeShares[j].operatorId.Hex()
+				}
+				return operatorStakeShares[i].stakeShare > operatorStakeShares[j].stakeShare
+			})
+			for i, op := range operatorStakeShares {
+				if op.operatorId == g.operatorId {
+					g.Registered.WithLabelValues(string(q), "stake_share").Set(op.stakeShare)
```
shrimalmadhur, line 219: what is the point of sending 2 different metrics? instead of sending 1 with both label (metric will also need to be updated the two labels)

node/metrics.go
```diff
@@ -150,3 +169,59 @@ func (g *Metrics) AcceptBatches(status string, batchSize uint64) {
 	g.AccuBatches.WithLabelValues("number", status).Inc()
 	g.AccuBatches.WithLabelValues("size", status).Add(float64(batchSize))
 }
+
+func (g *Metrics) collectOnchainMetrics() {
+	ticker := time.NewTicker(time.Duration(uint64(g.onchainMetricsInterval)))
+	defer ticker.Stop()
+
+	// 3 chain RPC calls in each cycle.
```
shrimalmadhur, line 177: why is this 3 ? curious to know!

jianoaix: It's all in the "chain RPC" log below (if that call fails), basically it needs to: 1) get current block number; 2) get the operator's quorums; 3) get the operator states.

We could by default set it to 3mins per cycle, so this metric is just adding 1 chain PRC per min.
Operators could adjust it via flag per their need.

node/metrics.go
jianoaix: We need both stake share and the rank of the operator's share across operators.

node/metrics.go
```diff
@@ -40,22 +44,29 @@ type Metrics struct {
...
+		// The "type" label have values: stake_share, rank. The "stake_share" is stake share (in basis point),
+		// and the "rank" is operator's ranking (the operator with highest amount of stake ranked as 1) by stake share in the quorum.
+		Registered: promauto.With(reg).NewGaugeVec(
 			prometheus.GaugeOpts{
 				Namespace: Namespace,
 				Name:      "registered",
-				Help:      "indicator about if DA node is registered",
+				Help:      "the quorums the DA node is registered",
 			},
+			[]string{"quorum", "type"},
```
shrimalmadhur, line 69: wait I am very confused with this metrics. 
This says `Registered` which seems, operator is registered or not. So binary 0 and 1

but we are setting this metrics to stake_share and rank. I think it will create confusion and probably not a right use of this?

shrimalmadhur: instead I think the better way can be - have 3 diff metrics
1) stake_share and for that label would be `quorum` 
2) rank and for that label would be `quorum`
3) registered  for that label would be `quorum`

so basically you just slice by quorum for all 3 metrics.

jianoaix: It means the stake registered for each quorum: one for the precentage info, the other for the rank by the stake
I'm not sure why it's confusing

shrimalmadhur: coz the metrics name is `registered` so this means- we are emitting the value of registered. but here the value is actually of a label and that's confusing.

shrimalmadhur: like for example
```
node_eigenda_blobs_total{quorum="0",type="number"} 97605
node_eigenda_blobs_total{quorum="0",type="size"} 1.16567530656e+11
```
the metrics value is of `node_eigenda_blobs_total` - and then you can slice by quorum and type to identify more params.

shrimalmadhur: now that I think of the above - even above is a bit wrong

shrimalmadhur: but if you think it's not confusing and will be clear to people directly using this metrics - we are okay. In case they use our dashboard, they would not have to worry about it. so I am okay

node/metrics.go
shrimalmadhur: oh I see. makes sesne.

node/metrics.go
jianoaix: You just have 2 time series per quorum: the stake share and rank. The same for node_eigenda_blobs_total.
There is no confusion point, no inefficiency as well.

shrimalmadhur (approved):

node/metrics.go
dmanc: image
from https://prometheus.io/docs/practices/naming/

Madhur makes good points and I feel it would be clearer if we move these labels into their own metrics time series.

jianoaix: interesting, this may affect a lot of metrics we built

shrimalmadhur: yea I know. but I think we should at least start making sure new metrics are correct and eventually migrate other faulty ones.

jianoaix: If you feel strongly about it, feel free to open a PR for it, I'll approve it

PR #505, opened 2024-04-20
Read disperser allowlist from json file

ian-shim:
Instead of configuring allowlist from cli flag, read from json file.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/apiserver/config.go
+++ b/disperser/apiserver/config.go
@@ -1,8 +1,11 @@
 package apiserver
 
 import (
+	"encoding/json"
 	"errors"
+	"io"
 	"log"
+	"os"
 	"strconv"
 	"strings"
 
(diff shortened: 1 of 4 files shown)


disperser/apiserver/config.go
```diff
@@ -188,6 +181,88 @@ func ReadCLIConfig(c *cli.Context) (RateConfig, error) {
...
+
+		quorumRateInfos[core.QuorumID(quorumID)] = QuorumRateInfo{
+			TotalUnauthThroughput:   common.RateParam(c.IntSlice(TotalUnauthThroughputFlagName)[ind]),
+			PerUserUnauthThroughput: common.RateParam(c.IntSlice(PerUserUnauthThroughputFlagName)[ind]),
+			TotalUnauthBlobRate:     common.RateParam(totalBlobRate * blobRateMultiplier),
+			PerUserUnauthBlobRate:   common.RateParam(accountBlobRate * blobRateMultiplier),
+		}
+	}
+
+	allowlist := parseAllowlistEntry(c)
```
dmanc, line 264: This only reads the entries of the file once right? Don't you want to periodically check the file and update the parameters accordingly?

ian-shim: I'll add that in a separate PR

dmanc (approved):
lgtm, we should make sure it can use both flags but will prioritize the json file so we can upgrade safely.

PR #510, opened 2024-04-22
Periodically reload allowlist from file

ian-shim:
Deprecated `auth.allowlist` and periodically reloads allowlist from the json file specified in `auth.allowlist-file` 

- [ ]  Merge only when `auth.allowlist` is no longer in use

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/apiserver/server.go
+++ b/disperser/apiserver/server.go
@@ -77,6 +77,7 @@ func NewDispersalServer(
 			logger.Info("[Allowlist]", "account", account, "name", rateInfo.Name, "quorumID", quorumID, "throughput", rateInfo.Throughput, "blobRate", rateInfo.BlobRate)
 		}
 	}
+	logger.Info("allowlist config", "file", rateConfig.AllowlistFile, "refreshInterval", rateConfig.AllowlistRefreshInterval.String())
 
 	authenticator := auth.NewAuthenticator(auth.AuthConfig{})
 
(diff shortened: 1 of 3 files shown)


disperser/apiserver/server.go
```diff
@@ -778,6 +795,15 @@ func (s *DispersalServer) Start(ctx context.Context) error {
 	return nil
 }
 
+func (s *DispersalServer) LoadAllowlist() {
+	al, err := ReadAllowlistFromFile(s.rateConfig.AllowlistFile)
+	if err != nil {
+		s.logger.Error("failed to load allowlist", "err", err)
```
dmanc, line 801: So if it fails to load allowlist, it will just keep the old values?

ian-shim: Yes!

disperser/apiserver/server.go
```diff
@@ -778,6 +795,15 @@ func (s *DispersalServer) Start(ctx context.Context) error {
...
+		return
+	}
+	s.rateConfig.Allowlist = al
```
mooselumph, line 804: Does this need any locks?

mooselumph (approved):

disperser/apiserver/server.go
ian-shim: I don't think this really needs to be synchronized. 
The worst case is that a potentially disallowed request goes through at the moment new allowlist is loaded (or vice versa)

PR #516, opened 2024-04-24
Validate onchain operator ID against local operator ID

ian-shim:
Suspend node if the registered operator ID onchain is different from the operator ID locally loaded from BLS key.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


node/node.go
```diff
@@ -216,6 +221,22 @@ func (n *Node) Start(ctx context.Context) error {
...
+	if operator != nil && operator.Address != "" {
+		operatorID, err := n.Transactor.OperatorAddressToID(ctx, gethcommon.HexToAddress(operator.Address))
+		if err != nil {
+			return fmt.Errorf("failed to get operator ID: %w", err)
+		}
+		if operatorID != operator.OperatorId {
+			return fmt.Errorf("operator ID mismatch: expected %s, got %s", operator.OperatorId.Hex(), operatorID.Hex())
+		}
+	} else {
+		// Doesn't have operator address. Just check that operator ID is registered onchain
```
jianoaix, line 233: This part may need to be dropped, we are moving towards a model where operators first bring up the node (without registration)

ian-shim: What happens when operators first bring up the node without registration? It just sits and does nothing?

mooselumph: Yes. They can check aspects of their setup (reachability, etc), without immediately being accountable for data. This should help address problem of operators joining and being down for a few hours.

ian-shim: Got it. Makes sense

jianoaix (approved):

PR #520, opened 2024-04-25
[Node] fix package path

siddimore:

For ldflags to work need to give  full package import path

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/node/cmd/Dockerfile
+++ b/node/cmd/Dockerfile
@@ -1,8 +1,8 @@
 FROM golang:1.21.1-alpine3.18 as builder
 
-ARG SEMVER="v0.x.y.z"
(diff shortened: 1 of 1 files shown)


shrimalmadhur (approved):

node/cmd/Dockerfile
```diff
@@ -1,8 +1,8 @@
 FROM golang:1.21.1-alpine3.18 as builder
 
-ARG SEMVER
-ARG GITCOMMIT
-ARG GITDATE
+ARG SEMVER="v0.x.y.z"
```
dmanc, line 3: maybe better to do empty string? Feels confusing hardcoding values

dmanc (approved):

PR #522, opened 2024-04-25
make ecdsa optional for certain operations

shrimalmadhur:

- This makes ECDSA optional for `list-quorums`
- It re-orders some codebase but ecdsa keys are too deep tied to make it pretty. I am also thinking a bit about future where folks wants this util just to generate payload without the need of ecdsa key and then they can use smart contract or cold key to sign some other way.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/node/plugin/cmd/main.go
+++ b/node/plugin/cmd/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+
 	"log"
 	"os"
 	"strings"
@@ -15,6 +16,8 @@ import (
 	"github.com/Layr-Labs/eigenda/node"
 	"github.com/Layr-Labs/eigenda/node/plugin"
 	"github.com/Layr-Labs/eigensdk-go/crypto/bls"
+
+	"github.com/ethereum/go-ethereum/accounts/keystore"
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/urfave/cli"
 )
@@ -69,12 +72,16 @@ func pluginOps(ctx *cli.Context) {
 
 	operatorID := keyPair.GetPubKeyG1().GetOperatorID()
 
-	sk, privateKey, err := plugin.GetECDSAPrivateKey(config.EcdsaKeyFile, config.EcdsaKeyPassword)
-	if err != nil {
-		log.Printf("Error: failed to read or decrypt the ECDSA private key: %v", err)
-		return
+	var sk *keystore.Key
+	var privateKey *string
+	if config.Operation != plugin.OperationListQuorums {
+		sk, privateKey, err = plugin.GetECDSAPrivateKey(config.EcdsaKeyFile, config.EcdsaKeyPassword)
+		if err != nil {
+			log.Printf("Error: failed to read or decrypt the ECDSA private key: %v", err)
+			return
+		}
+		log.Printf("Info: ECDSA key read and decrypted from %s", config.EcdsaKeyFile)
 	}
-	log.Printf("Info: ECDSA key read and decrypted from %s", config.EcdsaKeyFile)
 
 	loggerConfig := common.DefaultLoggerConfig()
 	logger, err := common.NewLogger(loggerConfig)
@@ -117,6 +124,16 @@ func pluginOps(ctx *cli.Context) {
 		}
 	}
 
+	if config.Operation == plugin.OperationListQuorums {
+		quorumIds, err := tx.GetRegisteredQuorumIdsForOperator(context.Background(), operatorID)
+		if err != nil {
+			log.Printf("Error: failed to get quorum(s) for operatorID: %x, operator address: %x, error: %v", operatorID, sk.Address, err)
+			return
+		}
+		log.Printf("Info: operator ID: %x, operator address: %x, current quorums: %v", operatorID, sk.Address, quorumIds)
+		return
+	}
+
 	operator := &node.Operator{
 		Address:             sk.Address.Hex(),
 		Socket:              socket,
@@ -152,13 +169,6 @@ func pluginOps(ctx *cli.Context) {
 			return
 		}
 		log.Printf("Info: successfully updated socket, for operator ID: %x, operator address: %x, socket: %s", operatorID, sk.Address, config.Socket)
-	} else if config.Operation == plugin.OperationListQuorums {
-		quorumIds, err := tx.GetRegisteredQuorumIdsForOperator(context.Background(), operatorID)
-		if err != nil {
-			log.Printf("Error: failed to get quorum(s) for operatorID: %x, operator address: %x, error: %v", operatorID, sk.Address, err)
-			return
-		}
-		log.Printf("Info: operator ID: %x, operator address: %x, current quorums: %v", operatorID, sk.Address, quorumIds)
 	} else {
 		log.Fatalf("Fatal: unsupported operation: %s", config.Operation)
 	}
--- a/node/plugin/cmd/main.go
+++ b/node/plugin/cmd/main.go
@@ -125,12 +125,17 @@ func pluginOps(ctx *cli.Context) {
 	}
 
 	if config.Operation == plugin.OperationListQuorums {
+		operatorAddress, err := tx.OperatorIDToAddress(context.Background(), operatorID)
+		if err != nil {
+			log.Printf("Error: failed to get operator address for operatorID: %x, error: %v", operatorID, err)
+			return
+		}
 		quorumIds, err := tx.GetRegisteredQuorumIdsForOperator(context.Background(), operatorID)
 		if err != nil {
 			log.Printf("Error: failed to get quorum(s) for operatorID: %x, operator address: %x, error: %v", operatorID, sk.Address, err)
 			return
 		}
-		log.Printf("Info: operator ID: %x, operator address: %x, current quorums: %v", operatorID, sk.Address, quorumIds)
+		log.Printf("Info: operator ID: %x, operator address: %x, current quorums: %v", operatorID, operatorAddress, quorumIds)
 		return
 	}
 
--- a/node/plugin/cmd/main.go
+++ b/node/plugin/cmd/main.go
@@ -132,10 +132,10 @@ func pluginOps(ctx *cli.Context) {
 		}
 		quorumIds, err := tx.GetRegisteredQuorumIdsForOperator(context.Background(), operatorID)
 		if err != nil {
-			log.Printf("Error: failed to get quorum(s) for operatorID: %x, operator address: %x, error: %v", operatorID, sk.Address, err)
+			log.Printf("Error: failed to get quorum(s) for operatorID: %x, operator address: %x, error: %v", operatorID, operatorAddress.Hex(), err)
 			return
 		}
-		log.Printf("Info: operator ID: %x, operator address: %x, current quorums: %v", operatorID, operatorAddress, quorumIds)
+		log.Printf("Info: operator ID: %x, operator address: %x, current quorums: %v", operatorID, operatorAddress.Hex(), quorumIds)
 		return
 	}
 
--- a/node/plugin/config.go
+++ b/node/plugin/config.go
@@ -150,7 +150,10 @@ func NewConfig(ctx *cli.Context) (*Config, error) {
 		return nil, errors.New("unsupported operation type")
 	}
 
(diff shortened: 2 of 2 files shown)


node/plugin/cmd/main.go
```diff
@@ -117,6 +124,16 @@ func pluginOps(ctx *cli.Context) {
 		}
 	}
 
+	if config.Operation == plugin.OperationListQuorums {
```
shrimalmadhur, line 127: putting this operation before `operator` is initialized since that require us to have private key. I am making minimal changes to make sure we make ecdsa optional here too.

node/plugin/cmd/main.go
```diff
@@ -117,6 +124,16 @@ func pluginOps(ctx *cli.Context) {
...
+		quorumIds, err := tx.GetRegisteredQuorumIdsForOperator(context.Background(), operatorID)
+		if err != nil {
+			log.Printf("Error: failed to get quorum(s) for operatorID: %x, operator address: %x, error: %v", operatorID, sk.Address, err)
+			return
+		}
+		log.Printf("Info: operator ID: %x, operator address: %x, current quorums: %v", operatorID, sk.Address, quorumIds)
```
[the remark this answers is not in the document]
shrimalmadhur: good catch.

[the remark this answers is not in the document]
shrimalmadhur: I think I can explicitly get that using smart contract call.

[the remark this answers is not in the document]
shrimalmadhur: ok fixed.

node/plugin/cmd/main.go
```diff
@@ -117,6 +124,21 @@ func pluginOps(ctx *cli.Context) {
 		}
 	}
 
+	if config.Operation == plugin.OperationListQuorums {
+		operatorAddress, err := tx.OperatorIDToAddress(context.Background(), operatorID)
```
shrimalmadhur, line 128: maybe we should just use this to get address for all cases and not rely on private key?

node/plugin/cmd/main.go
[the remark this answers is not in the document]
shrimalmadhur: 602fe252af9847bf56ecc4d5e412cbb146ae3070

node/plugin/cmd/main.go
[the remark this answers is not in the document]
shrimalmadhur: yea could be done. will need some refactor which I can do in another PR. I think this does need some refactor overall

node/plugin/cmd/main.go
```diff
@@ -117,6 +124,21 @@ func pluginOps(ctx *cli.Context) {
 		}
 	}
 
+	if config.Operation == plugin.OperationListQuorums {
```
ian-shim, line 127: why don't we put this block before declaring `sk` and `privateKey`? Then we don't need the block in L77, right?

node/plugin/cmd/main.go
```diff
@@ -117,6 +124,21 @@ func pluginOps(ctx *cli.Context) {
...
+		operatorAddress, err := tx.OperatorIDToAddress(context.Background(), operatorID)
+		if err != nil {
+			log.Printf("Error: failed to get operator address for operatorID: %x, error: %v", operatorID, err)
+			return
+		}
+		quorumIds, err := tx.GetRegisteredQuorumIdsForOperator(context.Background(), operatorID)
+		if err != nil {
+			log.Printf("Error: failed to get quorum(s) for operatorID: %x, operator address: %x, error: %v", operatorID, sk.Address, err)
```
ian-shim, line 135: isn't `sk` nil here?

node/plugin/config.go
```diff
@@ -150,6 +150,10 @@ func NewConfig(ctx *cli.Context) (*Config, error) {
 		return nil, errors.New("unsupported operation type")
 	}
 
+	if op != OperationListQuorums && len(ctx.GlobalString(EcdsaKeyFileFlag.Name)) == 0 && len(ctx.GlobalString(EcdsaKeyPasswordFlag.Name)) == 0 {
```
ian-shim, line 153: nit: I'd prefer matching `op == { opt-in, opt-out, update-socket }` to `op != OperationListQuorums`. I think that's more explicit and future compatible when we add new operation?

jianoaix (commented):
Does this separation actually make sense? like this is going to be the same binary for both mutation and read-only operations, unless there is use case setting up for only list-quorums

node/plugin/cmd/main.go
shrimalmadhur: I thought about that but the reason is - this still require `transactor` which requires `ethConfig` which requires `privateKey` - well `privateKey` is just passed as config. One way is I create `ethConfig` without privateKey and if we get to a point where `privateKey` is needed then I can add that config. would that be better?

node/plugin/cmd/main.go
shrimalmadhur: ah yes. I fixed the other logs but missed this one

node/plugin/config.go
shrimalmadhur: that's a fair point

node/plugin/cmd/main.go
shrimalmadhur: cca159bf8fd31d7356f648cc6908bcdae59a2ca7

node/plugin/config.go
shrimalmadhur: cca159bf8fd31d7356f648cc6908bcdae59a2ca7

node/plugin/cmd/main.go
```diff
@@ -117,6 +124,21 @@ func pluginOps(ctx *cli.Context) {
...
+			return
+		}
+		log.Printf("Info: operator ID: %x, operator address: %x, current quorums: %v", operatorID, operatorAddress, quorumIds)
```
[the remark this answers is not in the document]
shrimalmadhur: cca159bf8fd31d7356f648cc6908bcdae59a2ca7

node/plugin/cmd/main.go
shrimalmadhur: oh but that won't work since `transactor` will already be created by then. I don't think we want to create `transactor` again.

jianoaix:
> > Does this separation actually make sense? like this is going to be the same binary for both mutation and read-only operations, unless there is use case setting up for only list-quorums
> 
> yea why not? I am assuming there are multiple types of operators - one who is probably okay using ecdsa key in this, other might want to just create payloads/ list quorums - which doesn't require ecdsa keys. I don't think so we want diff binary for these use cases as that will probably be more work on their end and having multiple binaries will only create more confusion

The `list-quorums` is intended to be used closely with opt-in/out: the opt-in / out are essentially "adding quorum" and "removing quorum", and `list-quorums` is like a GPS telling them where they are right now, so they can be informed to add/remove quorums.

jianoaix:
The other way to look at it: if there are more read-only operations to add, the code here will be bloated with a lot such if-else catches to support them.

So my view on this is:
- Read-only operations closely tied to mutation operations as supporting tools: In this case they are intended to be used together, so no separation needed
- Read-only operations as a separate toolkit which may not be just used to support mutation operations: in this case, it'll make more sense to separate a binary/plugin (with its own flags, which will be less than mutation), so code will be cleaner and more maintainable

shrimalmadhur:
> The other way to look at it: if there are more read-only operations to add, the code here will be bloated with a lot such if-else catches to support them.
> 
> So my view on this is:
> 
> * Read-only operations closely tied to mutation operations as supporting tools: In this case they are intended to be used together, so no separation needed
> * Read-only operations as a separate toolkit which may not be just used to support mutation operations: in this case, it'll make more sense to separate a binary/plugin (with its own flags, which will be less than mutation), so code will be cleaner and more maintainable

I still don't like the idea of two binaries per AVS - more binaries will make more effort on us to maintain plus on operators to use, plus other AVS might follow that and it might not be a good idea.
But I understand that we do have bloated code and we need to make clear separation. This could de solved by single binary but having different sub commands which are related. This is what standard CLI does. 

Also as a probably a longer term - we should probably re-think nodeplugin - as in it can be just a binary which operators can download instead of docker image.

jianoaix:
> > The other way to look at it: if there are more read-only operations to add, the code here will be bloated with a lot such if-else catches to support them.
> > So my view on this is:
> > 
> > * Read-only operations closely tied to mutation operations as supporting tools: In this case they are intended to be used together, so no separation needed
> > * Read-only operations as a separate toolkit which may not be just used to support mutation operations: in this case, it'll make more sense to separate a binary/plugin (with its own flags, which will be less than mutation), so code will be cleaner and more maintainable
> 
> I still don't like the idea of two binaries per AVS - more binaries will make more effort on us to maintain plus on operators to use, plus other AVS might follow that and it might not be a good idea. But I understand that we do have bloated code and we need to make clear separation. This could de solved by single binary but having different sub commands which are related. This is what standard CLI does.
> 
> Also as a probably a longer term - we should probably re-think nodeplugin - as in it can be just a binary which operators can download instead of docker image.

What is the subcommand approach? "list-quorums" is separate already from "opt-in"?

jianoaix:
> > > > The other way to look at it: if there are more read-only operations to add, the code here will be bloated with a lot such if-else catches to support them.
> > > > So my view on this is:
> > > > 
> > > > * Read-only operations closely tied to mutation operations as supporting tools: In this case they are intended to be used together, so no separation needed
> > > > * Read-only operations as a separate toolkit which may not be just used to support mutation operations: in this case, it'll make more sense to separate a binary/plugin (with its own flags, which will be less than mutation), so code will be cleaner and more maintainable
> > > 
> > > 
> > > I still don't like the idea of two binaries per AVS - more binaries will make more effort on us to maintain plus on operators to use, plus other AVS might follow that and it might not be a good idea. But I understand that we do have bloated code and we need to make clear separation. This could de solved by single binary but having different sub commands which are related. This is what standard CLI does.
> > > Also as a probably a longer term - we should probably re-think nodeplugin - as in it can be just a binary which operators can download instead of docker image.
> > 
> > 
> > What is the subcommand approach? "list-quorums" is separate already from "opt-in"?
> 
> I mean right now it's not needed. but eventually if we add more commands, we might be able to group commands. Right now they are fine.

shrimalmadhur:
> Is using "list-quorums" on a separate machine than the one using opt-in/out needed right now?

It could be - what if operator opts-in using smart contract operator (or other ways where ecdsa key is not in same machine?) and want to list quorums? They should be able to run this without ecdsa key and password

shrimalmadhur:
> > > Is using "list-quorums" on a separate machine than the one using opt-in/out needed right now?
> > 
> > 
> > It could be - what if operator opts-in using smart contract operator (or other ways where ecdsa key is not in same machine?) and want to list quorums? They should be able to run this without ecdsa key and password
> 
> I don't even see there is a solution for this. If you are also based on speculation, I'd not jump into it before I see this is actually needed.

wdym by `I don't even see there is a solution for this`? 
We have all the feedback that we need to avoid giving access to ecdsa keys to node software as much as possible for any operations we do. This also applies to the nodeplugin cli where we should only require ecsda key for certain operations. Why would we want to force operators to provide ecdsa key path and pass for `list-quorums` if it's not needed? We need to set best practices to AVSs to avoid needing ecdsa keys as much as possible for things which doesn't require it. I think that's a reason enough to do that - whatever we set right now becomes the way AVSs follow it. Why do you think we should not do that?

jianoaix:
> > > > Is using "list-quorums" on a separate machine than the one using opt-in/out needed right now?
> > > 
> > > 
> > > It could be - what if operator opts-in using smart contract operator (or other ways where ecdsa key is not in same machine?) and want to list quorums? They should be able to run this without ecdsa key and password
> > 
> > 
> > I don't even see there is a solution for this. If you are also based on speculation, I'd not jump into it before I see this is actually needed.
> 
> wdym by `I don't even see there is a solution for this`? We have all the feedback that we need to avoid giving access to ecdsa keys to node software as much as possible for any operations we do. This also applies to the nodeplugin cli where we should only require ecsda key for certain operations. Why would we want to force operators to provide ecdsa key path and password for `list-quorums` if it's not needed? We need to set best practices to AVSs to avoid needing ecdsa keys as much as possible for things which doesn't require it. I think that's a reason enough to do that - whatever we set right now becomes the way AVSs follow it. Why do you think we should not do that?

You are way over extraplating the point with principles as arguments. That doesn't help the discussion.
You may want to explain:
1) using "list-quorums" as a standalone tool has real use cases (not "they could use it separately...")
2) if the point is smart contract account, then the proper tooling solution for that, how does it work with this nodeplugin etc

PR #527, opened 2024-04-27
Integrate fireblocks wallet into ejector

ian-shim:
Integrate Fireblocks wallet into ejector

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/cmd/dataapi/main.go
+++ b/disperser/cmd/dataapi/main.go
@@ -2,6 +2,7 @@ package main
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"log"
 	"os"
@@ -11,6 +12,7 @@ import (
 	"github.com/Layr-Labs/eigenda/common"
 	"github.com/Layr-Labs/eigenda/common/aws/dynamodb"
 	"github.com/Layr-Labs/eigenda/common/aws/s3"
+	"github.com/Layr-Labs/eigenda/common/aws/secretmanager"
 	"github.com/Layr-Labs/eigenda/common/geth"
 	coreeth "github.com/Layr-Labs/eigenda/core/eth"
 	"github.com/Layr-Labs/eigenda/disperser/cmd/dataapi/flags"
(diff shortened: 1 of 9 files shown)


disperser/dataapi/ejector.go
```diff
@@ -86,11 +106,73 @@ func (e *ejector) eject(ctx context.Context, nonsigningRate *OperatorsNonsigning
...
+		txID, err = e.wallet.SendTransaction(ctx, txn)
+		var urlErr *url.Error
+		didTimeout := false
+		if errors.As(err, &urlErr) {
+			didTimeout = urlErr.Timeout()
+		}
+		if didTimeout || errors.Is(err, context.DeadlineExceeded) {
+			e.logger.Warn("failed to send txn due to timeout", "hash", txn.Hash().Hex(), "numRetries", retryFromFailure, "maxRetry", maxSendTransactionRetry, "err", err)
+			retryFromFailure++
+			continue
```
jianoaix, line 136: No need "continue"

disperser/dataapi/ejector.go
```diff
@@ -86,11 +106,73 @@ func (e *ejector) eject(ctx context.Context, nonsigningRate *OperatorsNonsigning
...
+		if err != nil {
+			return fmt.Errorf("failed to update gas price: %w", err)
+		}
+		txID, err = e.wallet.SendTransaction(ctx, txn)
+		var urlErr *url.Error
+		didTimeout := false
+		if errors.As(err, &urlErr) {
+			didTimeout = urlErr.Timeout()
+		}
+		if didTimeout || errors.Is(err, context.DeadlineExceeded) {
```
jianoaix, line 133: So one timeout is from connecting to fireblocks endpoint, the other the fireblocks itself couldn't send txn internally?

disperser/cmd/dataapi/main.go
```diff
@@ -147,3 +160,70 @@ func RunDataApi(ctx *cli.Context) error {
...
+		sender, err := wallet.SenderAddress(context.Background())
+		if err != nil {
+			return nil, err
+		}
+		if sender.Cmp(gethcommon.HexToAddress(config.FireblocksConfig.WalletAddress)) != 0 {
+			return nil, fmt.Errorf("configured wallet address %s does not match derived address %s", config.FireblocksConfig.WalletAddress, sender.Hex())
+		}
+		logger.Info("Initialized Fireblocks wallet", "vaultAccountName", config.FireblocksConfig.VaultAccountName, "address", sender.Hex())
+	} else if len(config.EthClientConfig.PrivateKeyString) > 0 {
+		privateKey, err := crypto.HexToECDSA(config.EthClientConfig.PrivateKeyString)
```
jianoaix, line 207: wonder if there is any benefit if we put raw private key into the secret manager like the api keys above?

jianoaix (commented):
Lg, thanks for adding this!

disperser/dataapi/ejector.go
ian-shim: `didTimeout` detects whether it timed out from the http client, which can happen with fireblocks API call. `context.DeadlineExceeded` detects whether it timed out from from rpc calls, which can happen if wallet is using ethClient to send transactions

disperser/cmd/dataapi/main.go
ian-shim: Yes. That would be arguably more secure, but it adds a dependency on AWS secret manager so neither wallet will be functional if secret manager is down

jianoaix (approved):

PR #530, opened 2024-04-30
Use POST instead of GET for ejection requests

jianoaix:
Ejection request is to ask server to perform an action, so the appropriate method will be POST, not GET or PUT.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


ian-shim (approved):

dmanc (approved):

jianoaix:
Tested it work with `curl -X POST ....`

PR #541, opened 2024-05-03
[DataApi] Get BlobCount By AccountID

siddimore:
These changes enable querying blob count by AccountId

## Changes
1. Add GSI AccountId
2. Add method to GetBlobMetadataCount by AccountId
3. Assign accountKey to `accountId` in BlobAuthHeader.
4. Update DataApiHandler to getBlobMetadataCount by AccountId
5. Define DataApi Endpt for getting BlobCount by AccountId

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/apiserver/server.go
+++ b/disperser/apiserver/server.go
@@ -450,6 +450,11 @@ func (s *DispersalServer) checkRateLimitsAndAddRatesToHeader(ctx context.Context
 		// Update the quorum rate
 		blob.RequestHeader.SecurityParams[i].QuorumRate = accountRates.Throughput
 
+		// Update AccountID to accountKey
+		// This is a combination of origin and authenticatedAddress
+		// AccountId is later used to track blobs sent by the same account
+		blob.RequestHeader.BlobAuthHeader.AccountID = accountKey
+
 		// Get the encoded blob size from the blob header. Calculation is done in a way that nodes can replicate
 		encodedLength := encoding.GetEncodedBlobLength(length, uint8(param.ConfirmationThreshold), uint8(param.AdversaryThreshold))
 		encodedSize := encoding.GetBlobSize(encodedLength)
--- a/disperser/common/blobstore/blob_metadata_store.go
+++ b/disperser/common/blobstore/blob_metadata_store.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	commondynamodb "github.com/Layr-Labs/eigenda/common/aws/dynamodb"
+	"github.com/Layr-Labs/eigenda/core"
 	"github.com/Layr-Labs/eigenda/disperser"
 	"github.com/Layr-Labs/eigensdk-go/logging"
 	"github.com/aws/aws-sdk-go-v2/aws"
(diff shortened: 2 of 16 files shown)


ian-shim (commented):
Overall looks good to me. 
Is the primary purpose of adding the new field and index to support count query? 
I feel like getting the actual metadata might be more useful. Maybe we can support that later?

disperser/dataapi/server.go
```diff
@@ -244,6 +250,7 @@ func (s *server) Start() error {
 		{
 			feed.GET("/blobs", s.FetchBlobsHandler)
 			feed.GET("/blobs/:blob_key", s.FetchBlobHandler)
+			feed.GET("/blobs/:accountId", s.FetchBlobCountByAccountIdHandler)
```
dmanc, line 253: Can we make it more explicit that we're returning a count in the API name? Maybe `/blobs/:accountId/count`?

disperser/common/inmem/store.go
```diff
@@ -270,6 +271,18 @@ func (q *BlobStore) GetBlobMetadata(ctx context.Context, blobKey disperser.BlobK
 	return nil, disperser.ErrBlobNotFound
 }
 
+func (q *BlobStore) GetBlobMetadataCountByAccountID(ctx context.Context, accountID core.AccountID) (int32, error) {
+	q.mu.RLock()
+	defer q.mu.RUnlock()
+	count := int32(0)
+	for _, meta := range q.Metadata {
```
dmanc, line 278: This for loop can potentially be really large right?

disperser/common/blobstore/blob_metadata_store.go
```diff
@@ -218,6 +220,21 @@ func (s *BlobMetadataStore) GetBlobMetadataInBatch(ctx context.Context, batchHea
 	return metadata, nil
 }
 
+// GetBlobMetadataByAccount Count returns the count of all the metadata with the given status
+// Because this function scans the entire index, it should only be used for status with a limited number of items.
+// It should only be used to filter "Processing" status. To support other status, a streaming version should be implemented.
+func (s *BlobMetadataStore) GetBlobMetadataCountByAccountID(ctx context.Context, accountID core.AccountID) (int32, error) {
```
mooselumph, line 226: We probably want the total amount of data rather than the total number of blobs. Or perhaps both. 

Let's do another sync on the exact use case for this work!

disperser/apiserver/server.go
```diff
@@ -450,6 +450,11 @@ func (s *DispersalServer) checkRateLimitsAndAddRatesToHeader(ctx context.Context
 		// Update the quorum rate
 		blob.RequestHeader.SecurityParams[i].QuorumRate = accountRates.Throughput
 
+		// Update AccountID to accountKey
+		// This is a combination of origin and authenticatedAddress
+		// AccountId is later used to track blobs sent by the same account
+		blob.RequestHeader.BlobAuthHeader.AccountID = accountKey
```
anupsv, line 456: This might need to change depending on if the authenticated endpoint is being used right ?

disperser/dataapi/server.go
```diff
@@ -244,6 +250,7 @@ func (s *server) Start() error {
 		{
 			feed.GET("/blobs", s.FetchBlobsHandler)
 			feed.GET("/blobs/:blob_key", s.FetchBlobHandler)
+			feed.GET("/blobs/count/:accountId", s.FetchBlobCountByAccountIdHandler)
```
anupsv, line 253: Should definitely cache these endpoints.

PR #542, opened 2024-05-03
Bucket level metric

dmanc:

Adds bucket level metrics to the rate limiter library. Currently I'm making an assumption that if the RequesterName is not empty then we've authenticated the request is some way.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


common/ratelimit/limiter.go
```diff
@@ -109,7 +119,14 @@ func (d *rateLimiter) checkAllowed(ctx context.Context, params common.RequestPar
 		bucketParams.BucketLevels[i] = getBucketLevel(bucketParams.BucketLevels[i], size, interval, deduction)
 		allowed = allowed && bucketParams.BucketLevels[i] > 0
 
-		d.logger.Debug("Bucket level", "key", params.RequesterID, "prevLevel", prevLevel, "level", bucketParams.BucketLevels[i], "size", size, "interval", interval, "deduction", deduction, "allowed", allowed)
+		// Check if the user is authenticated before adding to metrics
```
mooselumph, line 122: I think it would be useful to also capture the system-level ratelimits. 

I guess this could be done by adding another field to the params struct indicating if it is a system or account level rate limit. Alternatively, we could replace `IsAuthenticated` with a different semantic, like `ShouldRecordMetrics`.

mooselumph (commented):
Looks good. Just one comment.

common/ratelimit/limiter.go
dmanc: Added system level ratelimits + parsed the string to get more labels.
```
# HELP rate_limiter_bucket_levels Current level of each bucket for rate limiting
# TYPE rate_limiter_bucket_levels gauge
rate_limiter_bucket_levels{account_key="18.208.69.153",account_type="ip",bucket_index="0",quorum="0",type="account_blob_rate"} 6e+11
rate_limiter_bucket_levels{account_key="18.208.69.153",account_type="ip",bucket_index="0",quorum="0",type="account_throughput"} 6e+11
rate_limiter_bucket_levels{account_key="18.208.69.153",account_type="ip",bucket_index="0",quorum="1",type="account_blob_rate"} 6e+11
rate_limiter_bucket_levels{account_key="18.208.69.153",account_type="ip",bucket_index="0",quorum="1",type="account_throughput"} 6e+11
# HELP rate_limiter_system_bucket_levels Current level of each bucket for system rate limiting
# TYPE rate_limiter_system_bucket_levels gauge
rate_limiter_system_bucket_levels{bucket_index="0",quorum="0",type="system_blob_rate"} 6e+11
rate_limiter_system_bucket_levels{bucket_index="0",quorum="0",type="system_throughput"} 6e+11
rate_limiter_system_bucket_levels{bucket_index="0",quorum="1",type="system_blob_rate"} 6e+11
rate_limiter_system_bucket_levels{bucket_index="0",quorum="1",type="system_throughput"} 6e+11
```

common/ratelimit/limiter.go
```diff
@@ -109,7 +125,44 @@ func (d *rateLimiter) checkAllowed(ctx context.Context, params common.RequestPar
 		bucketParams.BucketLevels[i] = getBucketLevel(bucketParams.BucketLevels[i], size, interval, deduction)
 		allowed = allowed && bucketParams.BucketLevels[i] > 0
 
-		d.logger.Debug("Bucket level", "key", params.RequesterID, "prevLevel", prevLevel, "level", bucketParams.BucketLevels[i], "size", size, "interval", interval, "deduction", deduction, "allowed", allowed)
+		if strings.HasPrefix(params.RequesterID, "system:") {
```
mooselumph, line 128: This logic seems pretty highly coupled with the format of the keys that are passed into the library methods, which someone could change. I think a simpler interface--even if it results in less intricate metrics--is probably preferable.

A suggestion would be to replace `params.IsAuthenticated` with a string parameter `params.MetricsKey`. If the string is blank, no metrics are recorded. If not, the metrics are recorded with this key. What do you think?

dmanc: So are you thinking we would just add a single label called "metrics_key" and pass in the requesterId?

mooselumph: I could see having two levels. Maybe something like
- `MetricsPrimaryKey`
- `MetricsSecondaryKey`

or 
- `MetricsKey`
- `MetricsTags`

Then the user of the library could populate these as desired.

disperser/apiserver/server.go
```diff
@@ -457,9 +463,10 @@ func (s *DispersalServer) checkRateLimitsAndAddRatesToHeader(ctx context.Context
 		// System Level
 		key := fmt.Sprintf("%s:%d-%s", systemAccountKey, param.QuorumID, SystemThroughputType.Plug())
 		requestParams = append(requestParams, common.RequestParams{
-			RequesterID: key,
-			BlobSize:    encodedSize,
-			Rate:        globalRates.TotalUnauthThroughput,
+			RequesterID:   key,
+			RequesterName: systemAccountKey,
```
ian-shim, line 467: shouldn't these be `requesterName`?

ian-shim (commented):
lgtm overall, left one comment

disperser/apiserver/server.go
dmanc: Hmm I see how it's confusing but in this case we are calculating the system rate limit metric so we want to make sure the metric that is being updated corresponds to the name "system".

mooselumph: I think to Ian's point, this might be a good reason to name these something like `MetricsId` and `MetricsFlag`, so that the intended usage is clear to the library user. Not super critical though.

mooselumph (approved):
Lgtm

PR #544, opened 2024-05-05
[SynthTest] Move RetrieveClient Step before blob confirmation

siddimore:

Blob Retrieval using Retrieve Client 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


ian-shim (approved):

ian-shim:
This doesn't make the synthetic client pass yet, right?

ian-shim:
> > This doesn't make the synthetic client pass yet, right?
> 
> that is correct......i am seeing random execution errors only on testnet

Is it random? I think it's consistently failing

PR #547, opened 2024-05-08
Optimize the node store writing perf

jianoaix:
About half of the node store writing time was spent on unnecessary serialization of chunks.

See the "chunk serialization duration" v.s. "write batch duration", which are the two dominant components of db writing and are roughly the same. With this PR, the "chunk serialization duration" will become negligible.

Screenshot 2024-05-07 at 8 55 40 PM

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests

### DIFF
--- a/node/store.go
+++ b/node/store.go
@@ -218,7 +218,6 @@ func (s *Store) StoreBatch(ctx context.Context, header *core.BatchHeader, blobs 
 
 	// Generate key/value pairs for all blob headers and blob chunks .
 	size := int64(0)
-	var serializationDuration, encodingDuration time.Duration
 	for idx, blob := range blobs {
 		// blob header
 		blobHeaderKey, err := EncodeBlobHeaderKey(batchHeaderHash, idx)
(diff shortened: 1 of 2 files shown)


ian-shim (approved):

pschork:
Nice optimization @jianoaix 

Wrt Fiona, do you think that their problematic node had degraded IO that was amplified with redundant serialization?

Also, I feel like some of your profiling code that was removed in https://github.com/Layr-Labs/eigenda/pull/547/commits/a65c410467deb8d1fa8fb7aa30c460c7ec38d2c4 (ie encode duration, serialization duration) could be useful as prometheus gauge in the operator's grafana dashboard for future debugging needs.

bxue-l2 (approved):

jianoaix:
> Nice optimization @jianoaix
> 
> Wrt Fiona, do you think that their problematic node had degraded IO that was amplified with redundant serialization?
> 
> Also, I feel like some of your profiling code that was removed in [a65c410](https://github.com/Layr-Labs/eigenda/commit/a65c410467deb8d1fa8fb7aa30c460c7ec38d2c4) (ie encode duration, serialization duration) could be useful as prometheus gauge in the operator's grafana dashboard for future debugging needs.

It is possible, but not certain. I think we could run another test after v0.7.0 (I'm waiting for the reachability check as an additional tool for this).
The metric was added separately so they were removed here to avoid conflict.

PR #550, opened 2024-05-09
[batcher] Replace Fireblocks wallet with KMS wallet

ian-shim:
We're no longer using Fireblocks as hot wallet to send transactions. We're using an EOA with a private key managed by KMS instead. 
This PR replaces the Fireblocks wallet to a private key wallet using KMS signer.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


common/logger_config.go
```diff
@@ -95,10 +95,10 @@ func ReadLoggerCLIConfig(ctx *cli.Context, flagPrefix string) (*LoggerConfig, er
 
 func NewLogger(cfg LoggerConfig) (logging.Logger, error) {
 	if cfg.Format == JSONLogFormat {
-		return logging.NewSlogJsonLogger(cfg.OutputWriter, &cfg.HandlerOpts), nil
```
ian-shim, line 98: This constructor has been [deprecated](https://github.com/Layr-Labs/eigensdk-go/pull/218)

disperser/cmd/batcher/main.go
```diff
@@ -102,11 +103,73 @@ func RunBatcher(ctx *cli.Context) error {
...
-	if err != nil {
-		logger.Error("Cannot create chain.Client", "err", err)
-		return err
+	var wallet walletsdk.Wallet
+	var client *geth.MultiHomingClient
+	if !config.KMSKeyConfig.Disable {
+		if config.KMSKeyConfig.KeyID == "" || config.KMSKeyConfig.Region == "" {
+			return errors.New("KMS key ID and region must be specified unless KMS wallet is disabled")
+		}
+		kmsClient, err := kms.NewKMSClient(context.Background(), config.KMSKeyConfig.Region)
```
dmanc, line 112: Is it ok to use context.Background()?

dmanc (approved):
lgtm

disperser/cmd/batcher/main.go
ian-shim: We can add a timeout there, but I think it's ok during the initialization

common/kms_wallet_config.go
```diff
@@ -0,0 +1,42 @@
...
+	Disable bool
+}
+
+func KMSWalletCLIFlags(envPrefix string, flagPrefix string) []cli.Flag {
+	return []cli.Flag{
+		cli.StringFlag{
+			Name:     PrefixFlag(flagPrefix, "kms-key-id"),
+			Usage:    "KMS key ID that stores the private key",
+			Required: false,
+			EnvVar:   PrefixEnvVar(envPrefix, "KMS_KEY_ID"),
```
dmanc, line 19: Should we use a [KMS alias](https://docs.aws.amazon.com/kms/latest/developerguide/kms-alias.html) rather than the key id? Think it would help when rotating the key since we wouldn't need to do a deployment.

anupsv (approved):

common/kms_wallet_config.go
ian-shim: Sounds good. This has to happen in sdk first, so I'll merge this and implement that as next iteration

ian-shim: Apparently it just works: https://docs.aws.amazon.com/kms/latest/APIReference/API_Sign.html#API_Sign_RequestSyntax

PR #551, opened 2024-05-09
Add EigenDA client using handshake based authentication

teddyknox:

To provide a client for accessing EigenDA with the following features:
1. BN254 31-byte encoding (dispersal and retrieval)
2. Length prefix encoding to return exact blob length that was dispersed (dispersal and retrieval)
3. IFFT encoding to enable location specific point opening for fraud/validity proofs. (dispersal and retrieval)
4. Automatically waiting for finalization (abstracting over `DisperseBlob()` and `GetBlobStatus()`).
5. Implement ECDSA signature based authentication.

This client will unify logic across many integrations and make EigenDA integrations easier to maintain.

This change also unifies the `api/` package into the root module, which can be done because the original reason for separating api into its own module no longer applies; we have upgraded certain dependencies within the root module that were conflicting with Optimism's set of dependencies before.

My testing strategy here has been based on testnet integration tests, since otherwise testing this client would require either mocking out the disperser, spinning up an entire inabox environment, or some middle ground between the two. For now, given time constraints, it is fine to verify that the client works against testnet.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Integration tests


core/auth/signer.go
```diff
@@ -12,24 +12,26 @@ import (
...
 
-type signer struct {
+type LocalBlobRequestSigner struct {
 	PrivateKey *ecdsa.PrivateKey
 }
 
-func NewSigner(privateKeyHex string) core.BlobRequestSigner {
+var _ core.BlobRequestSigner = &LocalBlobRequestSigner{}
+
+func NewLocalBlobRequestSigner(privateKeyHex string) *LocalBlobRequestSigner {
```
anupsv, line 21: Definitely good to have but would be good to recommend integrators to use something like remote signers while informing them.

teddyknox: This would not be a signer with any account balance so probably ok right?

api/client/client.go
```diff
@@ -0,0 +1,249 @@
...
+}
+
+func (m EigenDAClient) PutBlob(ctx context.Context, data []byte) (*Cert, error) {
+	m.Log.Info("Attempting to disperse blob to EigenDA")
+
+	// encode current blob encoding version byte
+	data = append([]byte{byte(NoIFFT)}, data...)
+
+	// encode data length
+	data = append(ConvertIntToVarUInt(len(data)), data...)
```
anupsv, line 118: How about adding a version as well ? To start with, version 0, or something similar, so that in the future, if we add more context like this, we can bump the version to know which deserialization would need to happen.

mooselumph:
What is the intended relationship between this client and the client in the `clients` folder? This is creating a lot of duplication. Moreover, the client there is integrated into several unit tests throughout the codebase. 

> This change also unifies the api/ package into the root module, which can be done because the original reason for separating api into its own module no longer applies; we have upgraded certain dependencies within the root module that were conflicting with Optimism's set of dependencies before.

Do we anticipate this remaining to be the case? 

If we are using the same module, then why should this code be contained in the `api` folder instead of the prexisting `clients` folder?

api/client/client.go
```diff
@@ -0,0 +1,249 @@
...
+	GetBlob(ctx context.Context, BatchHeaderHash []byte, BlobIndex uint32) ([]byte, error)
+	PutBlob(ctx context.Context, txData []byte) (*Cert, error)
+}
+
+type EigenDAClient struct {
+	Config
+
+	Log log.Logger
+
+	client disperser.DisperserClient
```
epociask, line 37: any reason why we aren't mocking the client and running assertions on the put loop failure cases? so far there's a test to verify correct dispersal/retrieval flows with testnet but nothing pertaining to other interactions

epociask (commented):
just some small comments - otherwise looks really good

api/client/client.go
```diff
@@ -0,0 +1,249 @@
...
+		Config: config,
+		client: client,
+		signer: auth.NewLocalBlobRequestSigner(config.SignerPrivateKeyHex),
+	}, nil
+}
+
+func NewDisperserClient(rpc string) (disperser.DisperserClient, error) {
+	config := &tls.Config{}
+	credential := credentials.NewTLS(config)
+	dialOptions := []grpc.DialOption{grpc.WithTransportCredentials(credential)}
```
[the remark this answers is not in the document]
anupsv: Making it default true, and setting it explicitly to false for local is good 👍

core/auth/signer.go
anupsv: Definitely thought of that. But since it could result in DOS or other issues in case it's leaked, better to still suggest a more safer practice.

api/client/client.go
```diff
@@ -0,0 +1,249 @@
...
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/credentials"
+)
+
+type BlobEncodingVersion byte
+
+var NoIFFT BlobEncodingVersion = 0x00
+
+type IEigenDAClient interface {
+	GetBlob(ctx context.Context, BatchHeaderHash []byte, BlobIndex uint32) ([]byte, error)
```
ian-shim, line 28: Will this only support retrieval through dispersal endpoint or also support decentralized retrieval from DA nodes?
Retrieval from DA nodes will require additional parameters

api/client/client.go
```diff
@@ -0,0 +1,249 @@
...
+	PutBlob(ctx context.Context, txData []byte) (*Cert, error)
```
ian-shim, line 29: Should it also take extra quorums as parameter?

core/auth/signer.go
ian-shim: If the key is compromised, anyone can sign a blob request and exhaust the rate limit

core/auth/signer.go
```diff
@@ -12,24 +12,26 @@ import (
 	"github.com/ethereum/go-ethereum/crypto"
 )
 
-type signer struct {
+type LocalBlobRequestSigner struct {
```
ian-shim, line 15: What other kinds of signer do we want to add in the future?

ian-shim (commented):
This kind of client implementation makes sense, and code lgtm. But I feel like it should belong on the integrator's codebase. 
Is this client tailored for a specific integration or all integrations?
If it's tailored for a specific integration, it should belong in the integrator's codebase utilizing the generic [disperser client](https://github.com/Layr-Labs/eigenda/blob/master/clients/disperser_client.go#L36-L40).
If your judgement is that all integrations will use this client, we should update [the existing client](https://github.com/Layr-Labs/eigenda/blob/master/clients/disperser_client.go#L36-L40).

What do you think?

core/auth/signer.go
teddyknox: True, there is always a consequence of a key leaking, but the severity of the consequence can vary. I've been operating under the assumption that we can treat this auth key the way many other web services tend to treat (non-monetary) auth keys, which is as parameters passed in via env vars and CLI params, and managed via K8s and whatever other DevOps infra. 

I think it would delay our goal of having customers use ECDSA handshake auth at launch if we were to ask rollups to set up remote signers infra (or at least configure it with additional new keys), and if we were to build/integrate a remote signer client.

api/client/client.go
teddyknox: It was my intention that the version byte written on line 115 would describe the encoding spec for the entire blob. Is there a benefit to having distinct version bytes for different subsections of the blob's encoding?

api/client/client.go
```diff
@@ -0,0 +1,249 @@
...
+	}
+
+	authHeaderReply, ok := reply.Payload.(*disperser.AuthenticatedReply_BlobAuthHeader)
+	if !ok {
+		return nil, fmt.Errorf("expected blob auth header message in response to initial disperse blob authenticated request: %w", err)
+	}
+
+	authHeader := core.BlobAuthHeader{
+		BlobCommitments: encoding.BlobCommitments{},
+		AccountID:       "",
```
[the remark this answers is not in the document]
teddyknox: I'm not sure, I copied this logic from [the test for the authentication logic](https://github.com/teddyknox/eigenda/blob/79613a97a964b4dccc278fd31aaca119ee7e2897/disperser/apiserver/ratelimit_test.go/#L209), so either the test is incorrect or this is faulty.

api/client/client.go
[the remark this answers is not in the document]
teddyknox: Sounds good.

api/client/client.go
```diff
@@ -0,0 +1,249 @@
...
+			if err != nil {
+				m.Log.Warn("Unable to retrieve blob dispersal status, will retry", "requestID", base64RequestID, "err", err)
+				continue
+			}
+
+			switch statusRes.Status {
+			case disperser.BlobStatus_PROCESSING:
+				m.Log.Info("Waiting for confirmation from EigenDA", "requestID", base64RequestID)
+			case disperser.BlobStatus_FAILED:
+				m.Log.Error("EigenDA blob dispersal failed in processing", "requestID", base64RequestID, "err", err)
```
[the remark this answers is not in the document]
teddyknox: Nope, will update.

api/client/client.go
```diff
@@ -0,0 +1,249 @@
...
+		return nil, fmt.Errorf("EigenDA client failed to copy unpadded data into final buffer")
+	}
+	if uint64(n) != length {
+		return nil, fmt.Errorf("EigenDA client failed, data length does not match length prefix")
+	}
+
+	return data, nil
+}
+
+func (m EigenDAClient) PutBlob(ctx context.Context, data []byte) (*Cert, error) {
```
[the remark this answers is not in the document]
teddyknox: I think we should expose both. I don't know of a use case for an await policy right now. In most rollup calling contexts the request should block.

api/client/client.go
```diff
@@ -0,0 +1,249 @@
...
+	"encoding/hex"
+	"fmt"
+	"time"
+
+	"github.com/Layr-Labs/eigenda/api/grpc/disperser"
+	"github.com/Layr-Labs/eigenda/core"
+	"github.com/Layr-Labs/eigenda/core/auth"
+	"github.com/Layr-Labs/eigenda/encoding"
+	"github.com/Layr-Labs/eigenda/encoding/utils/codec"
+	"github.com/ethereum/go-ethereum/log"
```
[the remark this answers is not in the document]
teddyknox: I would say we should go with industry standard for rollup frameworks, but can safely ignore whatever decision EigenDA made since this component will live on the client side. OP uses Ethereum's logging library, I don't know about the others.

api/client/client.go
teddyknox: No reason except that mocking out the entire API server seems beyond the scope of this PR.

api/client/client.go
teddyknox: The goal of this PR is to unify logic that exists today in several rollup integrations. Adding support for retriever sidecars is a more involved effort that will require operator nodes to stop whitelisting the disperser IP.

api/client/client.go
teddyknox: Extra quorums are configured at initialization time within the config.

core/auth/signer.go
teddyknox: A remote signer? Beyond the scope of this PR though.

api/client/client.go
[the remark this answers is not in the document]
teddyknox: Updated to add.

api/client/client.go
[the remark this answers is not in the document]
teddyknox: I'm updating the config setting to be named `DisableTLS` so the default value (false) refers to TLS being enabled.

teddyknox:
Thanks for all the feedback on this PR. I wasn't aware of the existing EigenDA client in the `clients/` directory as pointed out by @mooselumph 
but I'm open to combining them. They seem to be operating on two different levels, where this one is higher-level and handles encoding and status polling, and the existing client seems lower-level, and offering polish over the GRPC endpoints and handshake-based authentication.

Since the existing client is being used for traffic generation and integration tests which depend on its lower level semantics, it doesn't seem viable to fold that client into this one.

Conversely, it does seem possible to factor out aspects of this client which could be handled by the lower-level client, but would require some duplication of supporting code, e.g. separate config objects etc. This is the path I will take in pending follow ups to this PR.

Based on @epociask requests I think it does make sense to mock out the apiserver for the purpose of testing failure cases of this client, that work is also pending. For now though I've updated the branch with another draft of refactors that respond to the comments above.

teddyknox:
Ok I've updated this client to depend on the existing low level EigenDA client. The only thing I'd change about the low level client at this point is that the low level client reconnects with the server with every request. This seems like overkill, since GRPC connections are virtual and can abstract over whether there is a real TCP connection going at any point in time.

teddyknox:
In my latest push I've further updated this PR to add tests.

api/client/client.go
anupsv: I think we can bump the main version incase something changes in the subsections as well, so should be fine.

core/auth/signer.go
anupsv: Yep, definitely agree. It's not the best to write integrations to many of the cloud options or to pick anyone specific remote signer service. We can make some examples with AWS, GCP and things as fast follow as well.

api/clients/cert.go
```diff
@@ -0,0 +1,13 @@
+package clients
+
+import "github.com/Layr-Labs/eigenda/api/grpc/common"
+
+type Cert struct {
```
mooselumph, line 5: Shouldn't this struct contain all fields of the blob status reply? It looks like the inclusion proof has been dropped (not sure if there's anything else.)

api/clients/eigenda_client.go
```diff
@@ -0,0 +1,187 @@
...
+				m.Log.Info("Blob submitted, waiting for dispersal from EigenDA", "requestID", base64RequestID)
+			case grpcdisperser.BlobStatus_FAILED:
+				m.Log.Error("EigenDA blob dispersal failed in processing", "requestID", base64RequestID, "err", err)
+				errChan <- fmt.Errorf("EigenDA blob dispersal failed in processing, requestID=%s: %w", base64RequestID, err)
+				return
+			case grpcdisperser.BlobStatus_INSUFFICIENT_SIGNATURES:
+				m.Log.Error("EigenDA blob dispersal failed in processing with insufficient signatures", "requestID", base64RequestID, "err", err)
+				errChan <- fmt.Errorf("EigenDA blob dispersal failed in processing with insufficient signatures, requestID=%s: %w", base64RequestID, err)
+				return
+			case grpcdisperser.BlobStatus_CONFIRMED:
```
mooselumph, line 162: Couldn't there be a situation where a user wants to make their own determination of finalization status and to access the confirmation data in the meantime? Would they use the low-level client for this?

teddyknox: Yes, the low level client is a way to do this for clients who don't want to wait for finalization, and PutAsync() exists on the high-level client if the client wants to avoid a long block, but still wants to wait for finalization.

In general though until we have batch confirmation retries we want clients to wait until batches are finalized, otherwise reorgs can break rollups. The easiest way to do this is to this is to encode the opinion in the client.

api/clients/cert.go
teddyknox: The immediate need was to generalize across stage 0 integrations, and then add support for a stage 1 cert next. But will generalize across both stage 0, 1 of integrations. Stage 2 integrations require an aggregated signature in addition to the existing fields, so I've been operating under the assumption that the different stages will require different sets of fields.

api/clients/eigenda_client.go
teddyknox: > Couldn't there be a situation where a user wants to make their own determination of finalization status and to access the confirmation data in the meantime?

The only situation where the user would want to wait for confirmation but not finalization is if they're using some separate system for waiting for DA cert finalization, is there a specific use case you're thinking of?

api/clients/config.go
```diff
@@ -0,0 +1,49 @@
...
+	// Whether to disable TLS for an insecure connection when connecting to a local EigenDA disperser instance.
+	DisableTLS bool
+
+	// The blob encoding version to use when writing blobs from the high level interface.
+	PutBlobEncodingVersion BlobEncodingVersion
+}
+
+var DefaultQuorums = map[uint]bool{0: true, 1: true}
+
+func (c *EigenDAClientConfig) Check() error {
```
ian-shim, line 36: These validations and default parameter setup are usually done in constructors.
Maybe we can turn this into a constructor?
`func NewEigenDAClientConfig(...) (*EigenDAClientConfig, error)`

api/clients/config.go
```diff
@@ -0,0 +1,49 @@
...
+	SignerPrivateKeyHex string
+
+	// Whether to disable TLS for an insecure connection when connecting to a local EigenDA disperser instance.
+	DisableTLS bool
+
+	// The blob encoding version to use when writing blobs from the high level interface.
+	PutBlobEncodingVersion BlobEncodingVersion
+}
+
+var DefaultQuorums = map[uint]bool{0: true, 1: true}
```
ian-shim, line 34: Not sure hardcoding this is a good idea. We may omit the custom quorum validation altogether since the same validation is already done by the disperser.

ian-shim (commented):
lgtm, one comment on validation

api/clients/config.go
teddyknox: The idea is to fail on startup vs when the first dispersal takes place. If this is invalid config then I think it's ok to encode into the client that it's invalid config.

ian-shim: If we want to pull this off, I think it should read these from the source of truth (chain)

api/clients/config.go
ian-shim: Having the `Check()` method shares the same disadvantages as you'd need to pass in all fields to construct the config + calling `Check` method is completely optional and easy to miss by the user. 
By making the fields private, callers are forced to run the validation/set the default values.

teddyknox: The client calls Check() in case the user forgets. The client can't call NewEigenDAClientConfig() if the user forgets.

api/clients/config.go
teddyknox: Do you envision the default quorum IDs changing? What is the need for increased modularity here?

api/clients/config.go
```diff
@@ -0,0 +1,41 @@
...
+	// Whether to disable TLS for an insecure connection when connecting to a local EigenDA disperser instance.
+	DisableTLS bool
+
+	// The blob encoding version to use when writing blobs from the high level interface.
+	PutBlobEncodingVersion BlobEncodingVersion
+}
+
+func (c *EigenDAClientConfig) Check() error {
+	if c.StatusQueryRetryInterval == 0 {
+		c.StatusQueryRetryInterval = 5 * time.Second
```
ian-shim, line 35: It would be better to rename the method name to make it clear to the caller that this method modifies the content of the receiver (or return a new config instead of modifying the receiver in place).

ian-shim (approved):

PR #556, opened 2024-05-11
Add reachability check loop to expose reachability failures in node logs

pschork:
Default check interval is 5min
Minimum check interval is 10sec
Operators can disable by setting `NODE_REACHABILITY_POLL_INTERVAL` to 0

For backwards compatibility, the reachability check loop is disabled if the `NODE_DATAAPI_URL` is undefined (ie operator did not update their .env)

See also https://github.com/Layr-Labs/eigenda-operator-setup/pull/123

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [x] Integration tests

### DIFF
--- a/node/flags/flags.go
+++ b/node/flags/flags.go
@@ -181,6 +181,21 @@ var (
 		Value:    "180",
 		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "EXPIRATION_POLL_INTERVAL"),
 	}
+	ReachabilityPollIntervalSecFlag = cli.StringFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "reachability-poll-interval"),
+		Usage:    "How often (in second) to check if node is reachabile from DA backend",
+		Required: false,
+		Value:    "300",
+		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "REACHABILITY_POLL_INTERVAL"),
+	}
+	// Optional DataAPI URL. If not set, reachability checks are disabled
+	DataApiUrlFlag = cli.StringFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "dataapi-url"),
+		Usage:    "URL of the DataAPI",
+		Required: false,
+		Value:    "",
+		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "DATAAPI_URL"),
+	}
 	// NumBatchValidators is the maximum number of parallel workers used to
 	// validate a batch (defaults to 128).
 	NumBatchValidatorsFlag = cli.IntFlag{
@@ -263,6 +278,7 @@ var requiredFlags = []cli.Flag{
 var optionalFlags = []cli.Flag{
 	RegisterAtNodeStartFlag,
 	ExpirationPollIntervalSecFlag,
+	ReachabilityPollIntervalSecFlag,
 	EnableTestModeFlag,
 	OverrideBlockStaleMeasureFlag,
 	OverrideStoreDurationBlocksFlag,
@@ -274,6 +290,7 @@ var optionalFlags = []cli.Flag{
 	ChurnerUseSecureGRPC,
 	EcdsaKeyFileFlag,
 	EcdsaKeyPasswordFlag,
+	DataApiUrlFlag,
 }
 
 func init() {
--- a/node/flags/flags.go
+++ b/node/flags/flags.go
@@ -183,9 +183,9 @@ var (
 	}
 	ReachabilityPollIntervalSecFlag = cli.StringFlag{
 		Name:     common.PrefixFlag(FlagPrefix, "reachability-poll-interval"),
-		Usage:    "How often (in second) to check if node is reachabile from DA backend",
+		Usage:    "How often (in second) to check if node is reachabile from Disperser",
 		Required: false,
-		Value:    "300",
+		Value:    "60",
 		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "REACHABILITY_POLL_INTERVAL"),
 	}
 	// Optional DataAPI URL. If not set, reachability checks are disabled
--- a/node/node.go
+++ b/node/node.go
@@ -459,8 +459,9 @@ func (n *Node) checkCurrentNodeIp(ctx context.Context) {
 	}
 }
 
+// OperatorReachabilityResponse is the response object for the reachability check
 type OperatorReachabilityResponse struct {
-	OperatorId      string `json:"operator_id"`
+	OperatorID      string `json:"operator_id"`
 	DispersalSocket string `json:"dispersal_socket"`
 	RetrievalSocket string `json:"retrieval_socket"`
 	DispersalOnline bool   `json:"dispersal_online"`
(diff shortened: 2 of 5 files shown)


pschork:
Adds auto semantic versioning to node makefile.

```
> make semver
0.7.1-pschork-node-port-check.1
```

GitVersion knows that the latest tag release (currently) is `v0.7.0` and automatically bumps a patch release (ie `0.7.1`) and because this is a feature branch it includes branch name `pschork/node-port-check` plus build metadata `.1`

```
> make docker-node
> docker images opr-node
REPOSITORY   TAG                               IMAGE ID       CREATED          SIZE
opr-node     0.7.1-pschork-node-port-check.1   86d1ddd34fee   30 minutes ago   33.8MB
opr-node     latest                            86d1ddd34fee   30 minutes ago   33.8MB
```

```
> docker run -it opr-node:latest --version
da-node version 0.7.1-pschork-node-port-check.1-073ba2e-1715392025
```

node/node.go
```diff
@@ -453,3 +458,76 @@ func (n *Node) checkCurrentNodeIp(ctx context.Context) {
...
+// OperatorReachabilityResponse is the response object for the reachability check
+type OperatorReachabilityResponse struct {
+	OperatorID      string `json:"operator_id"`
+	DispersalSocket string `json:"dispersal_socket"`
+	RetrievalSocket string `json:"retrieval_socket"`
+	DispersalOnline bool   `json:"dispersal_online"`
+	RetrievalOnline bool   `json:"retrieval_online"`
+}
+
+func (n *Node) checkNodeReachability() {
```
jianoaix, line 471: I'd build this into a metric (additionally log it as well)

node/node.go
```diff
@@ -453,3 +458,76 @@ func (n *Node) checkCurrentNodeIp(ctx context.Context) {
...
+	}
+
+	n.Logger.Info("Start nodeReachabilityCheck goroutine in background to check the reachability of the operator node")
+	ticker := time.NewTicker(time.Duration(n.Config.ReachabilityPollIntervalSec) * time.Second)
+	defer ticker.Stop()
+
+	for {
+		<-ticker.C
+
+		if n.Config.DataApiUrl == "" {
```
ian-shim, line 484: can we early exit this method instead of spinning up a new goroutine if dataapi url is empty?

node/node.go
```diff
@@ -453,3 +458,76 @@ func (n *Node) checkCurrentNodeIp(ctx context.Context) {
...
+			n.Logger.Error("Unable to perform reachability check - NODE_DATAAPI_URL is not defined in .env")
+			continue
+		}
+
+		checkUrl, err := url.Parse(fmt.Sprintf("%s/api/v1/operators-info/port-check?operator_id=%s", n.Config.DataApiUrl, n.Config.ID.Hex()))
+		if err != nil {
+			n.Logger.Error("Reachability check failed - invalid check url", err, "checkUrl", checkUrl.String())
```
ian-shim, line 491: Do we want to return in this case or log error and continue checking reachability?

node/node.go
```diff
@@ -453,3 +458,76 @@ func (n *Node) checkCurrentNodeIp(ctx context.Context) {
...
+	for {
+		<-ticker.C
+
+		if n.Config.DataApiUrl == "" {
+			n.Logger.Error("Unable to perform reachability check - NODE_DATAAPI_URL is not defined in .env")
+			continue
+		}
+
+		checkUrl, err := url.Parse(fmt.Sprintf("%s/api/v1/operators-info/port-check?operator_id=%s", n.Config.DataApiUrl, n.Config.ID.Hex()))
+		if err != nil {
```
ian-shim, line 490: Which condition does it fall into if the dataapi itself was down?

node/node.go
pschork: My original implementation did that, but I changed the implementation to continuously throw errors to increase the likelihood of operators seeing the error. This error condition will only occur if the operator updates their docker image without manually adding the `NODE_DATAAPI_URL`.

node/node.go
pschork: My original implementation did that, but I changed the implementation to continuously throw errors to increase the likelihood of operators seeing the error.

node/node.go
pschork: Not this one. This is just ensuring that the constructed URL parses as valid.
If DataAPI is [down/unresolvable
](https://github.com/Layr-Labs/eigenda/pull/556/files#diff-375d57e386f20eaa5f09f02bb9d28bfc48ac3dca18d0325f59492208219e5618R498)
If DataAPI is [up but throwing non-200  ](https://github.com/Layr-Labs/eigenda/pull/556/files#diff-375d57e386f20eaa5f09f02bb9d28bfc48ac3dca18d0325f59492208219e5618R504)

node/node.go
ian-shim: I think this exits instead of continuously throwing error

pschork: Good catch.

pschork: Reverted to original disabled logic

node/node.go
pschork: We early exit now

node/flags/flags.go
```diff
@@ -181,6 +181,21 @@ var (
 		Value:    "180",
 		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "EXPIRATION_POLL_INTERVAL"),
 	}
+	ReachabilityPollIntervalSecFlag = cli.StringFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "reachability-poll-interval"),
+		Usage:    "How often (in second) to check if node is reachabile from DA backend",
```
jianoaix, line 186: nit: "DA backend" -> "Disperser"

node/flags/flags.go
```diff
@@ -181,6 +181,21 @@ var (
...
+		Required: false,
+		Value:    "300",
```
jianoaix, line 188: should we do it a bit more frequent, like once per min? It needs to clearly more frequent than batch interval to be more useful

node/node.go
```diff
@@ -453,3 +458,80 @@ func (n *Node) checkCurrentNodeIp(ctx context.Context) {
...
+	ticker := time.NewTicker(time.Duration(n.Config.ReachabilityPollIntervalSec) * time.Second)
+	defer ticker.Stop()
+
+	for {
+		<-ticker.C
+
+		n.Logger.Debug("Calling reachability check", "url", checkUrl.String())
+
+		resp, err := http.Get(checkUrl.String())
+		if err != nil {
```
jianoaix, line 498: Should we set a value in the gauge in case it gets error? @shrimalmadhur

jianoaix (approved):
LGTM, just one comment on if we need to update the gauge, when the call to dataapi fails to get response. @shrimalmadhur may have dealt with an issue for the onchain metric before, please chime in.

ian-shim (approved):

pschork:
> LGTM, just one comment on if we need to update the gauge, when the call to dataapi fails to get response. @shrimalmadhur may have dealt with an issue for the onchain metric before, please chime in.

Thanks @jianoaix. If Dataapi is down, this isn't really the operators issue and does not mean they are unreachable. We will log this condition as error in log every interval, so I feel like it wont go unnoticed.

node/node.go
shrimalmadhur: maybe we should, otherwise it will show last value.

shrimalmadhur:
> > LGTM, just one comment on if we need to update the gauge, when the call to dataapi fails to get response. @shrimalmadhur may have dealt with an issue for the onchain metric before, please chime in.
> 
> Thanks @jianoaix. If Dataapi is down, this isn't really the operators issue and does not mean they are unreachable. We will log this condition as error in log every interval, so I feel like it wont go unnoticed.

so if the dataapi is down - what's the indicator for operators? like they will still see reachable from their side since the gauge value has not changed. How do they know if data api is down and reachability check is not working fine?

PR #573, opened 2024-05-21
Optimize the batch deserialization perf

jianoaix:
The batch request deserialization is CPU intensive (especially at scale), however it's parallelizable.

This makes it about 5x faster in preprod test.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


ian-shim (approved):
🚀

node/grpc/utils.go
```diff
@@ -13,11 +13,16 @@ import (
...
 	"github.com/consensys/gnark-crypto/ecc/bn254"
 	"github.com/consensys/gnark-crypto/ecc/bn254/fp"
+	"github.com/gammazero/workerpool"
 	"github.com/wealdtech/go-merkletree"
 	"github.com/wealdtech/go-merkletree/keccak256"
 	"google.golang.org/protobuf/proto"
 )
 
+var (
+	maxDeserializationWorkers = 32
```
pschork, line 23: would be nice to expose this in operator config for tuning purposes

dmanc (approved):

pschork (approved):
🙏

node/grpc/utils.go
jianoaix: sounds good, I didn't make it so as I was thinking the validator was not using flag yet and wanted to do it together separately later (validation is actually using flag)

PR #581, opened 2024-05-24
Filter empty quorums when confirming batch

ian-shim:
With additional (custom) quorums, blobs may fail to get attested by some quorums and the current implementation would still post the batch including all the non-signers in the failed quorums. 
This PR makes batcher inspect the quorum results and filter out empty quorums when submitting the batch confirmation transaction onchain. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/aggregation.go
+++ b/core/aggregation.go
@@ -34,13 +34,12 @@ type SigningMessage struct {
 }
 
 // QuorumAttestation contains the results of aggregating signatures from a set of operators by quorums
+// It also returns map of all signers across all quorums
 type QuorumAttestation struct {
 	// QuorumAggPubKeys contains the aggregated public keys for all of the operators each quorum,
 	// including those that did not sign
 	QuorumAggPubKey map[QuorumID]*G1Point
-	// SignersAggPubKey is the aggregated public key for all of the operators that signed the message for each quorum,
-	// further aggregated across the quorums; operators signing for multiple quorums will be included in
-	// the aggregation multiple times
+	// SignersAggPubKey is the aggregated public key for all of the operators that signed the message by each quorum
 	SignersAggPubKey map[QuorumID]*G2Point
 	// AggSignature is the aggregated signature for all of the operators that signed the message for each quorum, mirroring the
 	// SignersAggPubKey.
--- a/core/aggregation.go
+++ b/core/aggregation.go
@@ -7,6 +7,7 @@ import (
 	"errors"
 	"fmt"
 	"math/big"
+	"slices"
 	"sort"
 
 	"github.com/Layr-Labs/eigensdk-go/logging"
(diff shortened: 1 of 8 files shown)


disperser/batcher/batcher.go
```diff
@@ -455,23 +445,41 @@ func (b *Batcher) HandleSingleBatch(ctx context.Context) error {
...
 	// TODO(mooselumph): Determine whether to confirm the batch based on the number of successes
 	if numPassed == 0 {
 		_ = b.handleFailure(ctx, batch.BlobMetadata, FailNoSignatures)
 		return errors.New("HandleSingleBatch: no blobs received sufficient signatures")
 	}
 
+	nonEmptyQuorums := []core.QuorumID{}
+	for quorumID, count := range numPassedByQuorum {
+		log.Info("Blobs attested by quorum", "quorumID", quorumID, "count", count)
+		if count > 0 {
```
jianoaix, line 468: Should this filter out quorums that have a signing rate below confirmation threshold?

core/aggregation.go
```diff
@@ -32,13 +33,31 @@ type SigningMessage struct {
...
 }
 
-// SignatureAggregation contains the results of aggregating signatures from a set of operators
+// QuorumAttestation contains the results of aggregating signatures from a set of operators by quorums
+type QuorumAttestation struct {
+	// QuorumAggPubKeys contains the aggregated public keys for all of the operators each quorum,
+	// including those that did not sign
+	QuorumAggPubKey map[QuorumID]*G1Point
+	// SignersAggPubKey is the aggregated public key for all of the operators that signed the message for each quorum,
+	// further aggregated across the quorums; operators signing for multiple quorums will be included in
```
jianoaix, line 42: nit: this doesn't aggregate across quorums now (it's per quorum)

core/aggregation.go
```diff
@@ -32,13 +33,31 @@ type SigningMessage struct {
...
+	// the aggregation multiple times
+	SignersAggPubKey map[QuorumID]*G2Point
```
jianoaix, line 44: Why pubkey is G2 here?

core/aggregation.go
```diff
@@ -32,13 +33,31 @@ type SigningMessage struct {
...
+	// AggSignature is the aggregated signature for all of the operators that signed the message for each quorum, mirroring the
+	// SignersAggPubKey.
+	AggSignature map[QuorumID]*Signature
+	// QuorumResults contains the quorum ID and the amount signed for each quorum
+	QuorumResults map[QuorumID]*QuorumResult
+	// SignerMap contains the operator IDs that signed the message
+	SignerMap map[OperatorID]bool
```
jianoaix, line 51: This is not per quorum, should it be left at in SignatureAggregation?

core/aggregation.go
ian-shim: It aggregates operator's `PubKeyG2`

core/aggregation.go
ian-shim: Yeah it's a bit awkward because it's the only field not divided into quorums. But it's easier to track this map at the stage of receiving signatures. I added a comment in the struct to clarify what this field is

disperser/batcher/batcher.go
ian-shim: This should be already filtering out quorums that have a signing rate below the confirmation threshold. 
`count` here in `numPassedByQuorum` is the number of blobs that have signing rate above the confirmation threshold, and we're filtering out quorums with count == 0

jianoaix: Since all blobs are in the same batch, they have the same signing rate for a given quorum. It looks this doesn't need to count num blobs (it's going to be either 0 or Num-Blobs), just whether the quorum collected sufficient signing rate

jianoaix: Does this make sense?

ian-shim: Sure, what you're suggesting is that we can make this logic simpler by simply comparing the quorum result vs. the confirmation threshold required at each quorum which is hardcoded in contracts. Correct?
But then the batcher needs to be aware of those thresholds and make an assumption that all blobs will be dispersed using the same thresholds for each quorum. 
Comparing the security params for each blobs with the quorums results allows the batcher to just use the security params set at each blob and filter quorums more generally.

jianoaix: I think numBlobsAttestedByQuorum may just return a set of quorumIDs that have collected sufficient signatures. It's semantically more clear but cutting away the count since it's 0 or all.

ian-shim: Sure, updated

jianoaix (approved):

PR #583, opened 2024-05-28
Add IFFT Encoding to the High Level Client

afkbyte:

High level client now always IFFT/FFTs when dispersing and retrieving data to/from eigenDA.

integration test with eigenDA testnet
Screenshot 2024-05-29 at 1 38 35 AM

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests


api/clients/codecs/blob_codec.go
```diff
@@ -0,0 +1,29 @@
+package codecs
```
teddyknox, line 1: No change

api/clients/codecs/default_blob_encoding_codec.go
```diff
@@ -8,32 +8,11 @@ import (
 	"github.com/Layr-Labs/eigenda/encoding/utils/codec"
 )
 
-type BlobEncodingVersion byte
```
teddyknox, line 11: Moved to blob_codec.go

api/clients/codecs/encodings_test.go
```diff
@@ -0,0 +1,88 @@
...
+		}
+
+		// Compare the original data with the decoded data
+		if !bytes.Equal(originalData, decodedData) {
+			t.Fatalf("Iteration %d: original and decoded data do not match\nOriginal: %v\nDecoded: %v", i, originalData, decodedData)
+		}
+	}
+}
+
+func TestIFFTBlobEncodingCodec(t *testing.T) {
```
teddyknox, line 57: New test

api/clients/codecs/encodings_test.go
```diff
@@ -0,0 +1,88 @@
...
+	b := make([]byte, length)
+	_, err := rand.Read(b)
+	if err != nil {
+		panic(err)
+	}
+	return b
+}
+
+// TestDefaultBlobEncodingCodec tests the encoding and decoding of random byte streams
+func TestDefaultBlobEncodingCodec(t *testing.T) {
```
teddyknox, line 23: Existing test, no change

api/clients/codecs/ifft_blob_encoding_codec.go
```diff
@@ -0,0 +1,125 @@
+package codecs
```
teddyknox, line 1: New file

teddyknox (approved):
LGTM

api/clients/eigenda_client.go
```diff
@@ -66,8 +67,8 @@ func (m EigenDAClient) GetBlob(ctx context.Context, BatchHeaderHash []byte, Blob
 		return nil, fmt.Errorf("blob has length zero")
 	}
 
-	version := BlobEncodingVersion(data[0])
-	codec, err := BlobEncodingVersionToCodec(version)
+	version := codecs.BlobEncodingVersion(data[0])
```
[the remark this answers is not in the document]
afkbyte: the codecBlobHeader has now changed, it's a 32 byte value that isn't transformed, the first byte is a 0, second byte is a version byte, and 3rd to 10th bytes are the length encoded as a uint64 to avoid this.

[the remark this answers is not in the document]
afkbyte: 26a95912051997b8972d18664c7b76777970fff0

api/clients/codecs/ifft_blob_encoding_codec.go
```diff
@@ -0,0 +1,125 @@
...
+	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
+)
+
+type IFFTBlobEncodingCodec struct{}
+
+var _ BlobCodec = IFFTBlobEncodingCodec{}
+
+func (v IFFTBlobEncodingCodec) EncodeBlob(rawData []byte) ([]byte, error) {
+	// encode current blob encoding version byte
+	encodedData := make([]byte, 0, 1+8+len(rawData))
```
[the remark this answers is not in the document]
afkbyte: ad57fc5ece87a98bdde564e58f83b08b0a328a1b

api/clients/codecs/blob_codec.go
```diff
@@ -0,0 +1,29 @@
...
+
+import (
+	"fmt"
+)
+
+type BlobEncodingVersion byte
+
+const (
+	// This minimal blob encoding includes a version byte, a length varuint, and 31 byte field element mapping. It does not include IFFT padding + IFFT.
```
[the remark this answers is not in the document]
afkbyte: 98f948c9745f7af866f0416ded42a2ea571b42e7

api/clients/eigenda_client.go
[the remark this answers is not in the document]
bxue-l2: following my question, even though the first several bytes are 0, but after taking IFFT, those are very likely non-zero. Mostly likely it will produce an error, because EncodingVersion is wrong. So my second response just says they probably have to know the encodingVersion at first, or you try both encodingVersion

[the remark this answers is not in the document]
afkbyte: so now we don't do an IFFT on the first 32 bytes, only the rawData itself

https://github.com/Layr-Labs/eigenda/blob/9c004ddbb0129f766ea6809915327cbf044ad6eb/api/clients/codecs/ifft_blob_encoding_codec.go#L26

api/clients/eigenda_client.go
```diff
@@ -57,17 +63,44 @@ func NewEigenDAClient(log log.Logger, config EigenDAClientConfig) (*EigenDAClien
...
+
+	maxScale := uint8(math.Log2(float64(dataFrIFFTLenPow2)))
+
+	fs := fft.NewFFTSettings(maxScale)
+	dataFr, err := fs.FFT(dataFrIFFT, false)
+	if err != nil {
+		return nil, fmt.Errorf("failed to perform FFT: %w", err)
+	}
+
+	data := rs.ToByteArray(dataFr, dataFrIFFTLenPow2*encoding.BYTES_PER_SYMBOL)
```
teddyknox, line 95: Can we encapsulate this in a function?

api/clients/eigenda_client.go
```diff
@@ -57,17 +63,44 @@ func NewEigenDAClient(log log.Logger, config EigenDAClientConfig) (*EigenDAClien
...
+
+	// get version and length from codec blob header
+	version, _, err := codecs.DecodeCodecBlobHeader(data[:5])
```
teddyknox, line 98: Don't we need to use the length here?

afkbyte: don't need length, it's handled by the Decode function in the Codec this version byte points to

api/clients/codecs/default_blob_encoding_codec.go
```diff
@@ -0,0 +1,52 @@
...
+func (v DefaultBlobEncodingCodec) DecodeBlob(encodedData []byte) ([]byte, error) {
+	// decode raw data modulo bn254
+	decodedData := codec.RemoveEmptyByteFromPaddedBytes(encodedData)
+
+	versionByte, length, err := DecodeCodecBlobHeader(decodedData[:31])
+	if err != nil {
+		return nil, err
+	}
+
+	if DefaultBlobEncoding != BlobEncodingVersion(versionByte) {
```
teddyknox, line 36: I'm wondering what this version byte tells us that we don't already know from the above steps.

api/clients/codecs/default_blob_encoding_codec.go
```diff
@@ -0,0 +1,52 @@
...
+
+	// encode all data modulo bn254
+	encodedData = codec.ConvertByPaddingEmptyByte(encodedData)
+
+	return encodedData, nil
+}
+
+func (v DefaultBlobEncodingCodec) DecodeBlob(encodedData []byte) ([]byte, error) {
+	// decode raw data modulo bn254
+	decodedData := codec.RemoveEmptyByteFromPaddedBytes(encodedData)
```
teddyknox, line 29: I think we should only apply this to the header first, then read the header, then use the version byte in the header to determine what kind of padding decoding to apply to the rest of the blob. Otherwise the version byte doesn't really give any information on the encoding that's not determined optimistically.

teddyknox: This all makes sense 👍

teddyknox (approved):
LGTM

api/clients/codecs/default_blob_encoding_codec.go
afkbyte: addressed here https://github.com/Layr-Labs/eigenda/pull/583#discussion_r1619430577

PR #584, opened 2024-05-28
[node] Log operator state hash on validation failure

ian-shim:
- adds a method to hash operator state
- on batch validation failure, log operator state hash 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/state.go
+++ b/core/state.go
@@ -2,8 +2,11 @@ package core
 
 import (
 	"context"
+	"crypto/md5"
+	"encoding/json"
 	"fmt"
 	"math/big"
+	"slices"
 	"strings"
 )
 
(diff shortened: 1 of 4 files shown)


pschork (approved):

core/state.go
```diff
@@ -60,6 +62,38 @@ type OperatorState struct {
 	BlockNumber uint
 }
 
+func (s *OperatorState) Hash() ([16]byte, error) {
+	marshalable := struct {
+		Operators   map[QuorumID]map[string]OperatorInfo
```
jianoaix, line 67: Hashing on hashmap may be nondeterministic

node/node.go
```diff
@@ -395,7 +395,15 @@ func (n *Node) ValidateBatch(ctx context.Context, header *core.BatchHeader, blob
 	}
 
 	pool := workerpool.New(n.Config.NumBatchValidators)
-	return n.Validator.ValidateBatch(header, blobs, operatorState, pool)
+	err = n.Validator.ValidateBatch(header, blobs, operatorState, pool)
+	if err != nil {
+		h, hashErr := operatorState.Hash()
```
jianoaix, line 400: How to use this hash returned from node?

core/state.go
ian-shim: good point, I converted it to sorted slice

node/node.go
ian-shim: This only gets embedded in the error that gets returned. 
Batcher also logs the hash so that it can be manually compared for debugging

disperser/batcher/batcher.go
```diff
@@ -420,6 +420,15 @@ func (b *Batcher) HandleSingleBatch(ctx context.Context) error {
 	stageTimer = time.Now()
 	update := b.Dispatcher.DisperseBatch(ctx, batch.State, batch.EncodedBlobs, batch.BatchHeader)
 	log.Debug("DisperseBatch took", "duration", time.Since(stageTimer))
+	h, err := batch.State.OperatorState.Hash()
```
jianoaix, line 423: The state here is for all operators, but the one used at node is for the specific operator, they may not be comparable

ian-shim: The state here has operator states for all quorums, so it will produce hashes for operators states in all quorums (note that one operator state hash is calculated for each quorum)
The operator state at the node will contain operator states for quorums the operator is part of. So it will only return hashes for operator states the operator is part of. 
Each hash for each quorum should be comparable.

jianoaix (approved):

PR #592, opened 2024-05-31
Add decode flag for GetBlob() by reworking codec layer

teddyknox:
## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests


api/clients/codecs/default_blob_codec.go
```diff
@@ -0,0 +1,55 @@
...
+	// encode raw data modulo bn254
+	rawDataPadded := codec.ConvertByPaddingEmptyByte(rawData)
+
+	// append raw data
+	encodedData := append(codecBlobHeader, rawDataPadded...)
+
+	return encodedData, nil
+}
+
+func (v DefaultBlobCodec) DecodeBlob(data []byte) ([]byte, error) {
```
epociask, line 37: Any reason why we aren't doing a length invariant on the data input? If not mistaken this could index out of bounds

api/clients/eigenda_client.go
```diff
@@ -44,21 +44,34 @@ func NewEigenDAClient(log log.Logger, config EigenDAClientConfig) (*EigenDAClien
...
 	if err != nil {
 		return nil, fmt.Errorf("error initializing EigenDA client: %w", err)
 	}
 
+	var codec codecs.BlobCodec
+	if config.DisablePointVerificationMode {
+		codec = codecs.NewNoIFFTCodec(lowLevelCodec)
+	} else {
+		codec = codecs.NewIFFTCodec(lowLevelCodec)
+	}
```
epociask, line 57: Knit - the same effect could be achieved via some encoding option type versus having to manage multiple codec types in different files (i.e, ifft, no_ifft)

teddyknox: I'm going to leave this as it is for now on the rationale that this option pattern is most useful when you're in a situation where you would otherwise define multiple constructor functions (with different goals). Since we are happy with one constructor function right now it seems like overkill to decompose the initialization of this into separate option functions.

PR #593, opened 2024-06-01
Automatic semantic versioning

pschork:
Enables automatic semantic versioning to released docker images

To release tagged docker images
1. Create and push a release branch
```
git checkout -b release/0.0.1
git push origin release/0.0.1
```
2.  Create a new [eigenda release](https://github.com/Layr-Labs/eigenda/releases) using a semver compliant tag ie `v0.0.1` or `v0.0.1-rc.1` 
3. You're done. The workflow is automatically triggered and image published.

You can also create a tag locally and push to origin to automatically create a new release
```
git checkout -b release/0.0.1
git tag v0.0.1-rc.1 -m v0.0.1-rc.1
git push origin release/0.0.1
git push origin tag v0.0.1-rc.1
```

If you need/want to manually push an untagged release you can trigger the workflow manually and specify `force` flag. If you don't force, then the image is not published.
image

Previously the workflow required you to enter a bunch of redundant information. This is no longer necissary.
image

The Makefile is a bit ugly, but it is handling 2 error conditions
1. someone downloaded a release zip archive and is trying to build from src. The archive does not include `.git` repo, so calculating semver is not possible. We fallback to using the parent dir name as the version (ie `eigenda-0.7.1`)
3. If docker is not available, thus gitversion is not available, so we use fall back version `0.0.0`

If a user clones `eigenda` where HEAD is tagged `v0.8.0` and then commits some changes on `master`, then build docker image, their `node --version` will resolve semver to `v0.8.1-pre.1`. In that case gitversion knows that the last tagged release was `v0.8.0`, it assume a minor point release bump and append `-pre.1` metadata aka `prerelease.1`.


.github/workflows/docker-publish-release.yaml
```diff
@@ -0,0 +1,80 @@
+name: docker-publish-release
+
+on:
+  push:
+    branches:
+      - pschork/semver
```
dmanc, line 6: Do we need to change this?

.github/workflows/docker-publish-release.yaml
```diff
@@ -0,0 +1,80 @@
...
+    tags:
+      - v*
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      force:
+        description: "Force untagged release (expert mode)"
```
dmanc, line 13: What's the usecase for pushing an untagged release?

dmanc (commented):
lgtm, want to meet to understand how the gitversion.yml and then I'll approve

.github/workflows/docker-publish-release.yaml
pschork: Allows someone to release a feature branch for debugging. Running the workflow manually should be rare.

dmanc (approved):

PR #600, opened 2024-06-05
[finalizer] Update confirmation block number in finalizer

ian-shim:
The `confirmationBlockNumber` [retrieved](https://github.com/ian-shim/eigenda/blob/master/disperser/batcher/finalizer.go#L93) from the confirmation transaction may have been updated due to chain reorg. 
This PR makes the finalizer update the `confirmationBlockNumber` to keep the metadata in sync. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/common/blobstore/blob_metadata_store.go
+++ b/disperser/common/blobstore/blob_metadata_store.go
@@ -235,6 +235,30 @@ func (s *BlobMetadataStore) IncrementNumRetries(ctx context.Context, existingMet
 	return err
 }
 
+func (s *BlobMetadataStore) UpdateConfirmationBlockNumber(ctx context.Context, existingMetadata *disperser.BlobMetadata, confirmationBlockNumber uint32) error {
+	updated := *existingMetadata
+	if updated.ConfirmationInfo == nil {
+		return fmt.Errorf("failed to update confirmation block number because confirmation info is missing for blob key %s", existingMetadata.GetBlobKey().String())
+	}
+
+	updated.ConfirmationInfo.ConfirmationBlockNumber = confirmationBlockNumber
+	item, err := MarshalBlobMetadata(&updated)
+	if err != nil {
+		return err
+	}
+
+	_, err = s.dynamoDBClient.UpdateItem(ctx, s.tableName, map[string]types.AttributeValue{
+		"BlobHash": &types.AttributeValueMemberS{
+			Value: existingMetadata.BlobHash,
+		},
+		"MetadataHash": &types.AttributeValueMemberS{
+			Value: existingMetadata.MetadataHash,
+		},
+	}, item)
+
+	return err
+}
+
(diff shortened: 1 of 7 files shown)


disperser/common/blobstore/shared_storage.go
```diff
@@ -161,14 +161,28 @@ func (s *SharedBlobStore) MarkBlobDispersing(ctx context.Context, metadataKey di
...
-	newMetadata.ConfirmationInfo = confirmationInfo
+	if confirmationInfo != nil {
+		newMetadata.ConfirmationInfo = confirmationInfo
+	}
 	return &newMetadata, s.blobMetadataStore.UpdateBlobMetadata(ctx, existingMetadata.GetBlobKey(), &newMetadata)
 }
 
-func (s *SharedBlobStore) MarkBlobFinalized(ctx context.Context, metadataKey disperser.BlobKey) error {
-	return s.blobMetadataStore.SetBlobStatus(ctx, metadataKey, disperser.Finalized)
+func (s *SharedBlobStore) MarkBlobFinalized(ctx context.Context, existingMetadata *disperser.BlobMetadata, confirmationBlockNumber uint64) (*disperser.BlobMetadata, error) {
```
jianoaix, line 175: It looks a more robust and simpler interface if the entire read-validate-update status-write back flow is in one function, as these steps are all tightly coupled.

ian-shim: what do you mean by read-validate-update status-write back flow?

jianoaix: Instead of taking a "existingMetadata" from the caller, this method can just take the potentially new confirmation block number: and internally it reads the metadata, compare the confirmation number with the given one, and if they are different, update it and write back. It's cleaner as it doesn't need to assume a "existingMetadata".

ian-shim: I reverted this method back so it just updates the blob status. 
I added a separate method `UpdateConfirmationBlockNumber` and update the confirmation block number regardless of whether the block is finalized or not as we need to update the confirmation block even when it's in CONFIRMED state.

disperser/common/blobstore/blob_metadata_store.go
```diff
@@ -235,6 +235,30 @@ func (s *BlobMetadataStore) IncrementNumRetries(ctx context.Context, existingMet
 	return err
 }
 
+func (s *BlobMetadataStore) UpdateConfirmationBlockNumber(ctx context.Context, existingMetadata *disperser.BlobMetadata, confirmationBlockNumber uint32) error {
+	updated := *existingMetadata
+	if updated.ConfirmationInfo == nil {
+		return fmt.Errorf("failed to update confirmation block number because confirmation info is missing for metadata %s", existingMetadata.GetBlobKey().String())
```
jianoaix, line 241: nit: "... is missing for blobkey ..."

jianoaix (approved):

PR #601, opened 2024-06-06
Add nonsigning rate threshold control to ejectorEjectorfilter

jianoaix:
Allow gradual rollout of ejection.

Tested: locally

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/cmd/dataapi/flags/flags.go
+++ b/disperser/cmd/dataapi/flags/flags.go
@@ -146,6 +146,13 @@ var (
 		Value:    6 * time.Minute,
 		EnvVar:   common.PrefixEnvVar(envVarPrefix, "TRANSACTION_TIMEOUT"),
 	}
+	NonsigningRateThresholdFlag = cli.IntFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "nonsigning-rate-threshold"),
+		Usage:    "only operators with nonsigning rate >= this threshold will be ejected, this value must be in range [10, 100], any value not in this range means disabling this flag",
+		Required: false,
+		Value:    -1,
+		EnvVar:   common.PrefixEnvVar(envVarPrefix, "NONSIGNING_RATE_THRESHOLD"),
+	}
 )
 
 var requiredFlags = []cli.Flag{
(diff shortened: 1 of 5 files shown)


bxue-l2 (approved):
lgtm

disperser/cmd/dataapi/flags/flags.go
```diff
@@ -146,6 +146,13 @@ var (
 		Value:    6 * time.Minute,
 		EnvVar:   common.PrefixEnvVar(envVarPrefix, "TRANSACTION_TIMEOUT"),
 	}
+	NonsigningRateThresholdFlag = cli.IntFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "nonsigning-rate-threshold"),
+		Usage:    "only operators with nonsigning rate >= this threshold are eligible for ejection, this value must be in range [10, 100], any value not in this range means disabling this flag",
```
bxue-l2, line 151: disabling the flag means, eject based on regular SLA?

disperser/dataapi/ejector.go
```diff
@@ -82,6 +84,11 @@ func (e *Ejector) Eject(ctx context.Context, nonsigningRate *OperatorsNonsigning
 
 	nonsigners := make([]*OperatorNonsigningPercentageMetrics, 0)
 	for _, metric := range nonsigningRate.Data {
+		// If nonsigningRateThreshold is set and valid, we will only eject operators with
+		// nonsigning rate >= nonsigningRateThreshold.
+		if e.nonsigningRateThreshold >= 10 && e.nonsigningRateThreshold <= 100 && metric.Percentage < float64(e.nonsigningRateThreshold) {
```
bxue-l2, line 89: what if e.nonsigningRateThreshold is 9, then regular SLA is enforced.
then if e.nonsigningRateThreshold is 50, then we enable the relaxed ejection.

disperser/cmd/dataapi/flags/flags.go
jianoaix: Yes, no additional conditions to affect the standard SLA and ejection flow.

disperser/dataapi/ejector.go
jianoaix: In first case: e.nonsigningRateThreshold will have no effect, it will just perform regular ejection

In second case: it'll eject an operator if 1) the operator violates SLA; **AND** 2) the operator's nonsigning rate >= 50%

Note: when it's regular ejection, only condition 1) is used

PR #602, opened 2024-06-07
Use accountKey as AccountID

dmanc:

Breaking this PR into separate components https://github.com/Layr-Labs/eigenda/pull/541. Want to get this in so I can add the secondary index on DynamoDB

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/apiserver/server.go
+++ b/disperser/apiserver/server.go
@@ -267,7 +267,7 @@ func (s *DispersalServer) disperseBlob(ctx context.Context, blob *core.Blob, aut
 		return nil, api.NewInvalidArgError(err.Error())
 	}
 
-	s.logger.Debug("received a new blob dispersal request", "origin", origin, "securityParams", strings.Join(securityParamsStrings, ", "))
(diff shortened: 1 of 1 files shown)


disperser/apiserver/server.go
```diff
@@ -457,6 +457,11 @@ func (s *DispersalServer) checkRateLimitsAndAddRatesToHeader(ctx context.Context
 		// Update the quorum rate
 		blob.RequestHeader.SecurityParams[i].QuorumRate = accountRates.Throughput
 
+		// Update AccountID to accountKey.
+		// This will use the origin as the account key if the user does not provide
+		// an authenticated address.
+		blob.RequestHeader.BlobAuthHeader.AccountID = accountKey
```
ian-shim, line 463: check if `blob.RequestHeader.BlobAuthHeader` is nil?

ian-shim (approved):

disperser/apiserver/server.go
dmanc: validateRequestAndGetBlob already makes sure that the blob is defined properly.

PR #603, opened 2024-06-09
Adds support for multi-platform linux/amd64 & linux/arm64 docker images

pschork:
This change will allow operators to run released docker images on ARM based [AWS Gravitron](https://aws.amazon.com/ec2/graviton/) instances as well as M1/2/3 Macs

Multi-platform docker compose builds need to be built + pushed within the same context, so this change consolidates the previous 2-staged _built + push_ flow into a single _build + potentially push_ flow via a `PUSH_FLAG` env variable that gets set if a push is intended.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


.github/workflows/docker-publish.yaml
```diff
@@ -1,8 +1,9 @@
-name: docker-publish
+name: docker-publish-internal
 on:
   push:
     branches:
       - master
+      - pschork/multi-platform-releases
```
dmanc, line 6: nit: remove branch

docker-compose-internal.yaml
```diff
@@ -24,6 +24,9 @@ services:
     build:
       context: .
       dockerfile: node/cmd/Dockerfile
+      platforms:
+        - "linux/amd64"
+        - "linux/arm64"
```
dmanc, line 29: Should we add arm images to the other components too?

dmanc (approved):

docker-compose-internal.yaml
pschork: Dropping this from this change. Building with docker compose is kludgy and we might be better off using buildx directly. There is also a problem with gitversion that needs more debugging.

PR #615, opened 2024-06-23
Use auto rate interval for blob thoughput metrics

pschork:
DA internal dashboard blob throughput does not match blob explorer throughput graphs because dataapi query uses `120s` rate interval and DA dashboard uses auto `$__rateInterval`. This change converts dataAPI metrics to use the auto `$__rateInterval`.

See https://grafana.com/blog/2020/09/28/new-in-grafana-7.2-__rate_interval-for-prometheus-rate-queries-that-just-work/

Comparison of using different rate intervals 
image

The time series gets really funky at 24hrs without auto `$__rateInterval`
image

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


pschork:
Abandoning. The `$__rate_interval` is Grafana only feature and is not supported for raw prometheus queries.

PR #617, opened 2024-06-24
[1/N] Chunk encoding optimization: Support encoding format with gnark

jianoaix:
There are inefficiency in chunk encoding, which causes more than data to transmit over network and store to db than necessary.

This PR adds the encoding with gnark. The followup PR will integrate it (compatibly).

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/encoding/serialization.go
+++ b/encoding/serialization.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"encoding/gob"
 	"encoding/json"
+	"errors"
 	"fmt"
 
 	"github.com/consensys/gnark-crypto/ecc/bn254"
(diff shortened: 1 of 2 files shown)


encoding/serialization.go
```diff
@@ -22,6 +23,41 @@ func (c *Frame) Deserialize(data []byte) (*Frame, error) {
 	return c, err
 }
 
+func (c *Frame) SerializeGnark() ([]byte, error) {
+	coded := make([]byte, 0, 32*(1+len(c.Coeffs)))
```
bxue-l2, line 27: can replace 32 with constants.go inside encoding/*

bxue-l2: for point size you can use SizeOfG1AffineUncompressed, see https://pkg.go.dev/github.com/consensys/gnark-crypto/ecc/bn254#G1Affine

bxue-l2 (approved):
lgtm

encoding/serialization.go
jianoaix: Done

encoding/serialization.go
```diff
@@ -22,6 +23,41 @@ func (c *Frame) Deserialize(data []byte) (*Frame, error) {
 	return c, err
 }
 
+func (c *Frame) SerializeGnark() ([]byte, error) {
+	coded := make([]byte, 0, bn254.SizeOfG1AffineCompressed+BYTES_PER_SYMBOL*len(c.Coeffs))
```
bxue-l2, line 27: Do you want to consider to use SizeOfG1AffineUncompressed?  Marshal uses uncompressed format. https://github.com/consensys/gnark-crypto/blob/v0.6.1/ecc/bn254/marshal.go#L607

In current EigenDA, we use uncompressed format with 64 bytes

jianoaix: We could settle on a 32 bytes point for this new encoding format

PR #619, opened 2024-06-26
Log the batch validation latency breakdown

jianoaix:
To gain visibility into latency breakdown between chain RPC and actual kzg validation.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -389,10 +389,12 @@ func (n *Node) ProcessBatch(ctx context.Context, header *core.BatchHeader, blobs
 }
 
 func (n *Node) ValidateBatch(ctx context.Context, header *core.BatchHeader, blobs []*core.BlobMessage) error {
+	start := time.Now()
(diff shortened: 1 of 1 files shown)


node/node.go
```diff
@@ -389,10 +389,12 @@ func (n *Node) ProcessBatch(ctx context.Context, header *core.BatchHeader, blobs
 }
 
 func (n *Node) ValidateBatch(ctx context.Context, header *core.BatchHeader, blobs []*core.BlobMessage) error {
+	start := time.Now()
```
bxue-l2, line 392: Would it be ok to take "operatorState, err := n.ChainState.GetOperatorStateByOperator(ctx, header.ReferenceBlockNumber, n.Config.ID)" out of the function. 

then in the caller context, use grafana to emit the metrics

jianoaix: It's too much to have a metric to just track a RPC call, I think logging is appropriate.

bxue-l2 (approved):

PR #620, opened 2024-06-26
Add configuration setting for high-level EigenDA client for returning as soon as the blob has confirmed

teddyknox:
Needed for the OP Stack integration


ian-shim (approved):
lgtm, please make sure the unit tests pass

PR #621, opened 2024-06-26
Use variable rate interval for blob thoughput metrics based on time interval

pschork:
DA internal dashboard blob throughput does not match blob explorer throughput graphs because dataapi query uses 2m rate interval and DA dashboard uses auto `$__rate_interval`

Blob explorer show 5MiB spikes, whereas DA dashboard show 1.5MiB throughput.
image
image
Comparison overlay of 5MiB `2m` rate vs 1.5MiB `4m` rate aka `$__rate_interval`
image

Blob explorer throughput breaks when you expand to 24h & 7d because of the 2m rate used.
image
image

See https://grafana.com/blog/2020/09/28/new-in-grafana-7.2-__rate_interval-for-prometheus-rate-queries-that-just-work/

The problem is that `$__rate_interval` is a grafana only feature and can not be used for raw prometheus queries. This change converts dataAPI metrics to use the auto variable rate based on query interval length and changes the default rate from 2m to 4m to match what `$__rate_interval` does under the covers.

Use `4m` for < 7d time interval
Use `11m` for >= 7d time interval

10m (note `$__rate_interval` matches 4m perfectly)
image
30m (note `$__rate_interval` matches 4m perfectly)
image
1h (note `$__rate_interval` matches 4m perfectly)
image
12h (note `$__rate_interval` matches 4m perfectly)
image
24h (note `$__rate_interval` matches 4m perfectly)
image
7d (note `$__rate_interval` matches 11m perfectly)
image

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/dataapi/metrics_handlers.go
+++ b/disperser/dataapi/metrics_handlers.go
@@ -11,8 +11,9 @@ import (
 )
 
 const (
-	avgThroughputWindowSize    = 120 // The time window (in seconds) to calculate the data throughput.
 	maxWorkersGetOperatorState = 10  // The maximum number of workers to use when querying operator state.
+	defaultThroughputRateSecs  = 240 // 4m rate is used for < 7d window to match $__rate_interval
+	sevenDayThroughputRateSecs = 660 // 11m rate is used for >= 7d window to match $__rate_interval
 )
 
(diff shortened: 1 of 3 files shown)


jianoaix (approved):
LGTM!

PR #626, opened 2024-07-05
[TxnManager] Make pk wallet compatible with broadcast check

ian-shim:
The broadcasting latency check introduced in https://github.com/Layr-Labs/eigenda/pull/465 only works for Fireblocks wallet. This PR makes it work with pk wallet. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/batcher/txn_manager.go
+++ b/disperser/batcher/txn_manager.go
@@ -208,7 +208,7 @@ func (t *txnManager) ensureAnyTransactionBroadcasted(ctx context.Context, txs []
 	for {
 		for _, tx := range txs {
 			_, err := t.wallet.GetTransactionReceipt(ctx, tx.TxID)
-			if err == nil || errors.Is(err, walletsdk.ErrReceiptNotYetAvailable) {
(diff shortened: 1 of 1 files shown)


samlaf (approved):
Hi this is @bxue-l2. Good catch. Nah, this is @samlaf

samlaf:
Wait #465 was merged in April, and since then this is the first time a sk wallet tx needs to get gasPrice bumped?

pschork:
> Wait #465 was merged in April, and since then this is the first time a sk wallet tx needs to get gasPrice bumped?

This might be true on mainnet. I remember us doing lots of gas bumps on holesky pre-launch, but those might have been limited to fireblocks as opposed to eoa wallet.

ian-shim:
> Wait #465 was merged in April, and since then this is the first time a sk wallet tx needs to get gasPrice bumped?

We only enabled the pk wallet in late May, and this was probably the first time it needed bumps since then

PR #627, opened 2024-07-07
Gpu encoder no blob batching

bxue-l2:

This PR adds GPU support for the encoder. The latency gains is approximately 33 and cost saving is 15 times better.

In implementation, the PR separates keys functions from the parameterized_prover by adding new interfaces called 

- ProofComputeDevice for kzg prover

- RsComputeDevice for RSEncoder

The PR contains both CPU and GPU implementation for the devices, so that an encoder can choose either one depending on the actual hardware. (Note, many functions in the GPU implementation of ProofComputeDevice relies on CPUs)

Further, this PR optimized encoding performance by adding parallel threads for non-interfering tasks like 
- lengthproof, lengthcommit, kzgCommit, proofs and chunks

The GPU support is possible with Ingonyama GPU acceleration technology. GPU has best performance when computation is batched. Currently within a single blob, there are already plenty batching. However, batching over blobs give additional benefits. 

This PR has not yet tested against the end-2-end test, the report and future works can be found in the [link](https://www.notion.so/274d3060dd24442cbd8d87c13cf5c46f?pvs=25). 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


encoding/test/main.go
```diff
@@ -49,30 +50,55 @@ func readpoints() {
...
+		numNode = 4
+		numSys = 2
+		numPar = numNode - numSys
+		kzgConfig = &kzg.KzgConfig{
+			G1Path:          "../../inabox/resources/kzg/g1.point.300000",
+			G2Path:          "../../inabox/resources/kzg/g2.point.300000",
+			CacheDir:        "SRSTables",
+			SRSOrder:        3000,
+			SRSNumberToLoad: 3000,
+			NumWorker:       uint64(runtime.GOMAXPROCS(0)),
```
samlaf, line 85:
```suggestion
			NumWorker:       uint64(runtime.NumCPU()),
```
Think you're just trying to get the number of CPUs here right? Since GOMAXPROCS(0) doesn't have any effect except returning the current set value. A more explicit way to do this is runtime.NumCPU()

encoding/test/main.go
```diff
@@ -49,30 +50,55 @@ func readpoints() {
...
+	numSys := uint64(512)   // 180
 	numPar := numNode - numSys
-	// Prepare data
-	fmt.Printf("* Task Starts\n")
-	fmt.Printf("    Num Sys: %v\n", numSys)
-	fmt.Printf("    Num Par: %v\n", numPar)
-	//fmt.Printf("    Data size(byte): %v\n", len(inputBytes))
+
+	numDevices, _ := cr.GetDeviceCount()
+	fmt.Println("num device ", numDevices)
```
samlaf, line 62: This function is just a scratch pad right? you're not using numDevices anywhere and you even have `p.UseGpu = false` below. What are you testing here exactly?

encoding/rs/cpu/extend_poly.go
```diff
@@ -0,0 +1,22 @@
...
+}
+
+// Encoding Reed Solomon using FFT
+func (g *CpuComputer) ExtendPolyEval(coeffs []fr.Element) ([]fr.Element, error) {
+	evals, err := g.Fs.FFT(coeffs, false)
+	if err != nil {
+		return nil, err
+	}
+
+	return evals, nil
```
samlaf, line 21:
```suggestion
	return g.Fs.FFT(coeffs, false)
```

encoding/rs/cpu/extend_poly.go
```diff
@@ -0,0 +1,22 @@
...
+
+import (
+	"github.com/Layr-Labs/eigenda/encoding"
+	"github.com/Layr-Labs/eigenda/encoding/fft"
+	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
+)
+
+type CpuComputer struct {
+	Fs *fft.FFTSettings
+	encoding.EncodingParams
```
samlaf, line 11: WHy is the encoding here? You don't seem to use it in ExtendPolyEval

encoding/rs/encode.go
```diff
@@ -122,27 +111,31 @@ func (g *Encoder) MakeFrames(
...
+func (g *Encoder) ExtendPolyEval(coeffs []fr.Element) ([]fr.Element, error) {
+
+	evals, err := g.Fs.FFT(coeffs, false)
 	if err != nil {
-		return nil, nil, err
+		return nil, err
 	}
 
-	return evals, pdCoeffs, nil
+	return evals, nil
```
samlaf, line 138: Think you can remove this method here since it's been moved to `{cpu,gpu}/extend_poly.go`?

encoding/rs/gpu/extend_poly.go
```diff
@@ -0,0 +1,39 @@
...
+	encoding.EncodingParams
+	NttCfg  core.NTTConfig[[bn254_icicle.SCALAR_LIMBS]uint32]
+	GpuLock *sync.Mutex
+}
+
+// Encoding Reed Solomon using FFT
+func (g *GpuComputeDevice) ExtendPolyEval(coeffs []fr.Element) ([]fr.Element, error) {
+
+	g.GpuLock.Lock()
+	defer g.GpuLock.Unlock()
```
samlaf, line 26: why is this needed? Are there going to be multiple encoders sharing a GPU?

encoding/rs/gpu/extend_poly.go
```diff
@@ -0,0 +1,39 @@
...
+
+	scalarsSF := gpu_utils.ConvertFrToScalarFieldsBytes(coeffs)
```
samlaf, line 28: is it important that coeffs is not tampered with? Otherwise you could potentially convert it inplace to save memory. Not sure how much of an improvement it would be but could be interesting to benchmark. You're doing a lot of memory allocations + copies everywhere.

encoding/rs/gpu/extend_poly.go
samlaf: oh is it because both the rs encoder and the prover share the gpu? You might want to add that as a comment.

encoding/kzg/prover/parametrized_prover.go
```diff
@@ -46,228 +69,118 @@ func (g *ParametrizedProver) EncodeBytes(inputBytes []byte) (*bn254.G1Affine, *b
...
+	proofsResult := <-proofChan
 
-	// outputs is out of order - buttefly
-	proofs, err := p.Fs.FFTG1(sumVecInv[:dimE], false)
-	if err != nil {
-		return nil, err
+	if lengthProofResult.Err != nil || lengthCommitmentResult.Err != nil ||
+		commitmentResult.Err != nil || rsResult.Err != nil ||
+		proofsResult.Err != nil {
+		return nil, nil, nil, nil, nil, multierror.Append(lengthProofResult.Err, lengthCommitmentResult.Err, commitmentResult.Err, rsResult.Err, proofsResult.Err)
```
samlaf, line 158: I think using https://pkg.go.dev/golang.org/x/sync/errgroup could be very idiomatic here.
I've never used errgroup myself but I think the idea is that here if one of the computation errors early, your current code will still wait for everything to finish, and then return the joint error.
An errgroup would short-circuit the other computations as soon as one goroutine errors out, and return that. Unless you really want to know all errors that happened (and not just the first one), then this could be useful to save resources?

samlaf (commented):
First round of review. I don't understand most of the code here so only left a bunch of style/nit comments. Still getting familiar with the codebase, will need to ask you some questions.

PR #628, opened 2024-07-08
[3/N] Chunk encoding optimization: Add support of new encoding format…

jianoaix:
… in API

To handle a different format of chunk encoding.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proto/node/node.proto
```diff
@@ -54,6 +54,14 @@ message RetrieveChunksRequest {
 message RetrieveChunksReply {
 	// All chunks the Node is storing for the requested blob per RetrieveChunksRequest.
 	repeated bytes chunks = 1;
+	// Describes how the chunks above are encoded.
+	// If not specified, it’s treated as GOB.
```
bxue-l2, line 58: if unspecified, wouldn't it be the first definition in enum, which is unknown

bxue-l2 (approved):

api/proto/node/node.proto
jianoaix: We need to default to GOB in order to be backward compatible

jianoaix: Actually it's fine to default o UNKNOWN -- we just need to make sure the GOB is set when we return GOB encoded chunks

PR #629, opened 2024-07-09
Refactor cpu prover

bxue-l2:

This PR refactors GPU provers by 
- adding new interface responsible for computation, and it is useful for gpu implementation in the future
- adding parallelization among different tasks including (proof), (lengthCommitment) and so on
- adding better log and remove verbosity flag

The encode task has 5 smaller tasks, and the time spent is the sum of all small task. With the second optimization, the time spent would be the max of all tasks. Empricially speaking, it would reduce the encoding tie from 8.15sec to 5.18 second, which is about 37.5% improvement.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


encoding/kzg/prover/compute_proof.go
```diff
@@ -0,0 +1,206 @@
...
+//
+// phi ^ (coset size ) = 1
+//
+// implicitly pad slices to power of 2
+func (p *CpuComputer) GetSlicesCoeff(polyFr []fr.Element, dimE, j, l uint64) ([]fr.Element, error) {
+	// there is a constant term
+	m := uint64(len(polyFr)) - 1
+	dim := (m - j) / l
+
+	toeV := make([]fr.Element, 2*dimE-1)
```
dmanc, line 186: nit: Make 2*dimE-1 a constant

encoding/kzg/prover/compute_proof.go
```diff
@@ -0,0 +1,206 @@
...
+
+	"github.com/Layr-Labs/eigenda/encoding/fft"
+	"github.com/Layr-Labs/eigenda/encoding/kzg"
+	"github.com/Layr-Labs/eigenda/encoding/utils/toeplitz"
+	"github.com/consensys/gnark-crypto/ecc"
+	"github.com/consensys/gnark-crypto/ecc/bn254"
+	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
+)
+
+type CpuComputer struct {
```
dmanc, line 16: nit: Rename to CPUComputer or CPUProofComputer

encoding/kzg/prover/encode.go
```diff
@@ -0,0 +1,14 @@
+package prover
+
+import (
+	"github.com/consensys/gnark-crypto/ecc/bn254"
+	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
+)
+
+type ProofComputeDevice interface {
```
dmanc, line 8: How about just `ProofComputer`

dmanc (approved):
lgtm

bxue-l2:
Actually, I made a mistake, this optimization would not save us from 8.15 sec to 5.18sec. Probably only going to save us 100-200ms. This is because the bottleneck proving part isn't parallelized, and that total time is 8.15.

The parallelization is only saving 100ms for encoding.

encoding/kzg/prover/encode.go
bxue-l2: that is its old name

encoding/kzg/prover/compute_proof.go
jianoaix, line 1: The file name could be proof_computer.go

encoding/kzg/prover/compute_proof.go
jianoaix: +1 to CPUProofComputer

encoding/kzg/prover/encode.go
jianoaix, line 1: Can the file name match closer to the content?
If file name is encoder (encode as a verb isn't fit well for file name), maybe the interface below should be `ProofEncoder`

encoding/kzg/prover/compute_proof.go
bxue-l2: I dont think we should add cpu into the name, since in the future, it could have been a gpu. Either cpu or gpu, I don't think it makes sense to have both

encoding/kzg/prover/encode.go
bxue-l2: how about proof_device.go

jianoaix (approved):

PR #630, opened 2024-07-09
Adds NodeInfo endpoint to retriever and dispersal services for observability

pschork:
After deploying this change, we can have dataapi (via reachability check) hit this endpoint and log semver per operator.

```
grpcurl --plaintext localhost:32005 node.Dispersal.VersionInfo
{
  "semver": "0.7.5"
}
```

```
grpcurl --plaintext localhost:32005 node.Dispersal.VersionInfo
{
  "semver": "0.8.0-rc.1"
}
```

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proto/node/node.proto
```diff
@@ -14,13 +14,17 @@ service Dispersal {
 	// for the protocol-defined length of custody. It will return a signature at the
 	// end to attest to the data in this request it has processed.
 	rpc StoreChunks(StoreChunksRequest) returns (StoreChunksReply) {}
+    // Retrieve version into metadata
```
ian-shim, line 17: nit: indent

node/config.go
```diff
@@ -33,7 +33,7 @@ var (
 		0: "eth_quorum",
 		1: "permissioned_quorum",
 	}
-	SemVer    = "v0.0.0"
```
ian-shim, line 36: How is this used today and what difference does this make?

node/config.go
```diff
@@ -33,7 +33,7 @@ var (
 		0: "eth_quorum",
 		1: "permissioned_quorum",
```
ian-shim, line 34: Not sure how this is used. Do we need to update this to `eigen_quorum`?

ian-shim (commented):
lgtm 
Does this work in parallel with the existing http endpoint? `xxx.xxx.xxx.xxx/eigen/node` 
What's the advantage over the previous version check?

node/config.go
jianoaix: +1

api/proto/node/node.proto
```diff
@@ -14,13 +14,17 @@ service Dispersal {
...
+	rpc VersionInfo(VersionInfoRequest) returns (VersionInfoReply) {}
 }
 
 service Retrieval {
 	// RetrieveChunks retrieves the chunks for a blob custodied at the Node.
 	rpc RetrieveChunks(RetrieveChunksRequest) returns (RetrieveChunksReply) {}
 	// GetBlobHeader is similar to RetrieveChunks, this just returns the header of the blob.
 	rpc GetBlobHeader(GetBlobHeaderRequest) returns (GetBlobHeaderReply) {}
+    // Retrieve version into metadata
```
jianoaix, line 26: `into` -> `info`

api/proto/node/node.proto
```diff
@@ -153,3 +157,12 @@ message BatchHeader {
 	// The Ethereum block number at which the batch is dispersed.
 	uint32 reference_block_number = 3;
 }
+
+// Version info request
```
jianoaix, line 161: It may be helpful to think a bit longer term. Beyond software version, maybe other node info will be desired, e.g. how many os, num CPUs etc. In that case, this API would be `GetNodeInfo`?

node/config.go
pschork: This is a fallback. Only gets triggered if you do a `go build` without passing `-ldflags="-X 'github.com/Layr-Labs/eigenda/node.SemVer=${SEMVER}'` so pretty much just our unit tests will see `0.0.0`

I'm updating it to be semver spec compliant which does not use `v` prefix.

pschork:
> lgtm Does this work in parallel with the existing http endpoint? `xxx.xxx.xxx.xxx/eigen/node` What's the advantage over the previous version check?

The advantage is that we have a guarantee that it is reachable because the IP:PORT is registered. The node api runs on port `9093` by default, but operator is free to change that port, and there is no guarantee that port is publicly exposed.

pschork:
Adds node info metadata

```
grpcurl --plaintext localhost:32005 node.Dispersal.NodeInfo
{
  "semver": "0.8.0-pschork-version-info.1",
  "arch": "arm64",
  "os": "linux",
  "numCpu": 8,
  "memBytes": "8222203904"
}
```

api/proto/node/node.proto
```diff
@@ -153,3 +157,16 @@ message BatchHeader {
...
+message NodeInfoRequest {
+}
+
+// Node info reply
+message NodeInfoReply {
+    string semver = 1;
+    string arch = 2;
+    string os = 3;
+    uint32 num_cpu = 4;
+	uint64 mem_bytes = 5;
```
jianoaix, line 171: nit: indent

jianoaix (approved):

jianoaix:
> > lgtm Does this work in parallel with the existing http endpoint? `xxx.xxx.xxx.xxx/eigen/node` What's the advantage over the previous version check?
> 
> The advantage is that we have a guarantee that it is reachable because the IP:PORT is registered. The node api runs on port `9093` by default, but operator is free to change that port, and there is no guarantee that port is publicly exposed.

If we want to rely on this API operationally (not just informationally), it has to have a granrantee.
However if there is any concern of forcing this API on each operator

node/grpc/server.go
```diff
@@ -130,6 +132,16 @@ func (s *Server) serveRetrieval() error {
...
 }
 
+func (s *Server) NodeInfo(ctx context.Context, in *pb.NodeInfoRequest) (*pb.NodeInfoReply, error) {
+	memBytes := uint64(0)
+	v, err := mem.VirtualMemory()
+	if err == nil {
+		memBytes = v.Total
+	}
+
+	return &pb.NodeInfoReply{Semver: node.SemVer, Os: runtime.GOOS, Arch: runtime.GOARCH, NumCpu: uint32(runtime.NumCPU()), MemBytes: memBytes}, nil
```
ian-shim, line 142: I don't think `runtime.NumCPU` is aware of the CPU limits set on the container. 
I think https://github.com/uber-go/automaxprocs can help set the right number of CPUs in `runtime.GOMAXPROCS`.

pschork:
> However I wonder if there is any concern of forcing this API on each operator

Exposing `semver` seems reasonable and defensible for purpose of managing breaking migration rollout. System level info is less defensible and operators could raise security/privacy concerns to enumerating internal specs.

cc @anupsv

pschork:
Added a flag to disable resource information (os, arch, cpus, mem) from the HostInfo API.
This will allow operator to OPT-OUT of exposing resource info. 
Semver is always exposed.

```
# NODE_DISABLE_HOST_INFO_RESOURCES=true
grpcurl --plaintext localhost:32005 node.Dispersal.NodeInfo
{
  "semver": "0.8.0-pschork-version-info.1"
}
```

jianoaix (approved):

pschork:
> Added a flag to disable resource information (os, arch, cpus, mem) from the NodeInfo API. This will allow operator to OPT-OUT of exposing resource info. Semver is always exposed.
> 
> ```
> # NODE_DISABLE_NODE_INFO_RESOURCES=true
> grpcurl --plaintext localhost:32005 node.Dispersal.NodeInfo
> {
>   "semver": "0.8.0-pschork-version-info.1"
> }
> ```

If we want to be less aggressive, perhaps this should be OPT-IN. @jianoaix @ian-shim @anupsv @nivethapu

ian-shim (approved):
I think opt out is totally reasonable

PR #632, opened 2024-07-09
[4/N] Chunk encoding optimization: Add bundle encoding

jianoaix:
Support the encoding of the entire bundle into bytes.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/data.go
+++ b/core/data.go
@@ -1,11 +1,13 @@
 package core
 
 import (
+	"encoding/binary"
 	"errors"
 	"fmt"
 
 	"github.com/Layr-Labs/eigenda/common"
 	"github.com/Layr-Labs/eigenda/encoding"
+	"github.com/consensys/gnark-crypto/ecc/bn254"
 )
 
 type AccountID = string
@@ -33,6 +35,8 @@ const (
 	// which means the max ID can not be larger than 254 (from 0 to 254, there are 255
 	// different IDs).
 	MaxQuorumID = 254
+
+	GnarkBundleEncodingFormat = 1
 )
 
 func (s *SecurityParam) String() string {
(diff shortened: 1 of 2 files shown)


core/data.go
```diff
@@ -165,6 +169,67 @@ func (b Bundle) Size() uint64 {
...
+	size := 0
+	for _, f := range b {
+		if len(f.Coeffs) != len(b[0].Coeffs) {
+			return nil, errors.New("invalid bundle: all chunks should have the same length")
+		}
+		size += bn254.SizeOfG1AffineCompressed + encoding.BYTES_PER_SYMBOL*len(f.Coeffs)
+	}
+	result := make([]byte, size+8)
+	buf := result
+	metadata := uint64(GnarkBundleEncodingFormat) | (uint64(len(b[0].Coeffs)) << 8)
```
bxue-l2, line 188: just informational. The first 64 bytes belong to the header, the least significant byte has the length, the next 63 bytes are for length? I think the maximal chunk length we have is 2^28.

core/data.go
```diff
@@ -165,6 +169,67 @@ func (b Bundle) Size() uint64 {
...
+	}
+	return result, nil
+}
+
+func (b Bundle) Deserialize(data []byte) (Bundle, error) {
+	if len(data) < 8 {
+		return nil, errors.New("bundle data must have at least 8 bytes")
+	}
+	// Parse metadata
+	meta := binary.LittleEndian.Uint64(data)
```
bxue-l2, line 207: So there is only one serializatoin format based on bundle. The old format is based on chunk, I guess?

core/data.go
```diff
@@ -165,6 +169,67 @@ func (b Bundle) Size() uint64 {
...
+	if (meta & 0xFF) != GnarkBundleEncodingFormat {
+		return nil, errors.New("invalid bundle data encoding format")
+	}
+	chunkLen := meta >> 8
```
bxue-l2, line 211: maybe add a check on the size of chunk size?

bxue-l2 (approved):

core/data.go
jianoaix: Yes, let me add some documentation to make it clear.

The first 8 bytes are header (not 64).
Good callout about the max chunk length, then just this 8 bytes will provide a lot of space for encoding information. Right now it'll just be: 1) what's the encoding format is this? 2) what's the chunk len?

core/data.go
jianoaix: That's right. We can put the Gob chunk into this bundle bytes as well, but since we'll move to Gnark which is strictly better (more compressed and much faster to encode/decode), then we don't need to bother old chunk any more.

PR #635, opened 2024-07-12
Minibatch store interface

ian-shim:
Adding interface for minibatch store. 
It's missing many methods to be filled later. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/batcher/minibatch_store.go
+++ b/disperser/batcher/minibatch_store.go
@@ -1,0 +1,54 @@
+package batcher
+
+import (
+	"time"
+
+	"github.com/Layr-Labs/eigenda/core"
+	gcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/google/uuid"
+)
+
+type BatchRecord struct {
+	ID                   uuid.UUID
+	CreatedAt            time.Time
+	ReferenceBlockNumber uint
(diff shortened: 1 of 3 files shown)


disperser/batcher/minibatch_store.go
```diff
@@ -0,0 +1,55 @@
+package batcher
+
+import (
+	"time"
+
+	"github.com/Layr-Labs/eigenda/core"
+	gcommon "github.com/ethereum/go-ethereum/common"
+)
+
+type BatchID time.Time
```
pschork, line 10: I think the better option would be to go with UUIDv7 https://pkg.go.dev/github.com/google/uuid#NewV7

UUIDv7 gives us sorting and entropy to avoid collisions. Time.Now() works, but the lack of entropy makes collisions a real possibility if we ever introduce multi-threading or distributed minibatchers

pschork: I guess one benefit of using nanoseconds is that the primary key field is smaller vs UUID. After reading dynamo docs some more, using timestamp as primary key is pretty common pattern. So now I think time.Time is fine.

pschork (approved):

Issue #87, opened 2023-11-29
[Bug]: Division by Zero in InplaceFFT Function


estensen:
### What happened + What you expected to happen

A runtime panic due to division by zero occurs in the `InplaceFFT` function of the `kzg` package when the input slice `vals` is empty.

```go
func (fs *FFTSettings) InplaceFFT(vals []bls.Fr, out []bls.Fr, inv bool) error {
	n := uint64(len(vals))
	...
	if !bls.IsPowerOfTwo(n) {
		return fmt.Errorf("got %d values but not a power of two", n)
	}
	...
	// Potential division by zero
	stride := fs.MaxWidth / n
	...
}

func IsPowerOfTwo(v uint64) bool {
	return v&(v-1) == 0
}
```

Testing `IsPowerOfTwo` I found that it's incorrectly implemented. `0` is a power of 2, but shouldn't be.

```go
func TestIsPowerOfTwo(t *testing.T) {
	testCases := []struct {
		input    uint64
		expected bool
	}{
		{0, false}, // 0 is not a power of 2
		{1, true},  // 2^0
		{2, true},  // 2^1
		{3, false}, // Not a power of 2
		{4, true},  // 2^2
	}

	for _, tc := range testCases {
		t.Run(fmt.Sprintf("Input%d", tc.input), func(t *testing.T) {
			result := IsPowerOfTwo(tc.input)
			if result != tc.expected {
				t.Fatalf("IsPowerOfTwo(%d) = %v; want %v", tc.input, result, tc.expected)
			}
		})
	}
}

--- FAIL: TestIsPowerOfTwo (0.00s)
    --- FAIL: TestIsPowerOfTwo/Input0 (0.00s)
        /Users/estensen/Developer/eigenda/pkg/kzg/bn254/globals_test.go:49: IsPowerOfTwo(0) = true; want false
```

The reason is that `v-1` will underflow to `18446744073709551615`
Then `0000...0000 & 1111...1111 == 0`

The correct implementation of `IsPowerOfTwo`
```go
func IsPowerOfTwo(v uint64) bool {
    return v != 0 && v&(v-1) == 0
}
```

While there might be validation other places so `vals []bls.Fr` is never an empty slice I think there should be multiple layers of validation.

### Versions / Dependencies

cefcf3eed64439350e30fda35a46fb9093475035 (latest main)
`go version go1.21.4 darwin/amd64`

### How to reproduce

```go
$ cd pkg/encoding/encoder
[...]

PR #638, opened 2024-07-15
Fix fft_fr.IsPowerOfTwo(). (#83)

cody-littley:

This change fixes https://github.com/Layr-Labs/eigenda/issues/87

The function `fft_fr.IsPowerOfTwo()` did not return the correct response for `0`.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests

### DIFF
--- a/encoding/fft/fft_fr.go
+++ b/encoding/fft/fft_fr.go
@@ -132,7 +132,7 @@ func (fs *FFTSettings) InplaceFFT(vals []fr.Element, out []fr.Element, inv bool)
 	}
 }
 
-// Returns true if the provided integer v is a power of 2.
+// IsPowerOfTwo Returns true if the provided integer v is a power of 2.
 func IsPowerOfTwo(v uint64) bool {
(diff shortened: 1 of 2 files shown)


encoding/fft/fft_fr.go
```diff
@@ -132,6 +132,7 @@ func (fs *FFTSettings) InplaceFFT(vals []fr.Element, out []fr.Element, inv bool)
 	}
 }
 
+// IsPowerOfTwo Returns true if the provided integer v is a power of 2.
```
jianoaix, line 135: nit: **r**eturns

jianoaix (approved):

encoding/fft/fft_fr.go
cody-littley: Replaced `Returns` with `returns`

dmanc (approved):
lgtm as long as nothing depends on this specific behavior (I doubt it).

PR #642, opened 2024-07-16
Gpu encoder no blob batching

bxue-l2:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


dmanc (commented):
Can probably use build tags in order to only compile the cpu code in the docker / unit tests.

https://www.digitalocean.com/community/tutorials/customizing-go-binaries-with-build-tags

PR #649, opened 2024-07-19
Generate inclusion proof with blob index instead of hash

ian-shim:
Previously, it was using the blob header hash to identify the index of the leaf, which is problematic when there are duplicate leaves.
Generating proofs with the leaf index solves this issue.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


bxue-l2 (approved):

disperser/batcher/batcher.go
```diff
@@ -262,15 +262,8 @@ func (b *Batcher) updateConfirmationInfo(
...
-			blobHeader := batchData.blobHeaders[blobIndex]
 
-			blobHeaderHash, err := blobHeader.GetBlobHeaderHash()
-			if err != nil {
-				b.logger.Error("HandleSingleBatch: failed to get blob header hash", "err", err)
-				blobsToRetry = append(blobsToRetry, batchData.blobs[blobIndex])
-				continue
-			}
-			merkleProof, err := batchData.merkleTree.GenerateProof(blobHeaderHash[:], 0)
+			merkleProof, err := batchData.merkleTree.GenerateProofWithIndex(uint64(blobIndex), 0)
```
jianoaix, line 266: Any compatibility issue with this change?

ian-shim: Shouldn't be. It won't affect blobs that had been dispersed already. It will only affect blobs to be dispersed after the deploy by populating correct inclusion proofs in the confirmation info

jianoaix (approved):

disperser/batcher/batcher.go
jianoaix: What about the consumer of this inclusion proof? Will this break anything on that side?

ian-shim: No the consumer of the inclusion proof reads it from DB without re-generating it

PR #654, opened 2024-07-22
[7/N] Chunk encoding optimization: Node support for mixed encoding

jianoaix:
To make the Node able to serve both Gob and Gnark chunks, and return only Gob chunks during migration -- guarded by a flag.

This also added more test coverage for the mixed encoding which can be the case during migration.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests


node/grpc/server.go
```diff
@@ -307,6 +307,26 @@ func (s *Server) RetrieveChunks(ctx context.Context, in *pb.RetrieveChunksReques
 		s.node.Metrics.RecordRPCRequest("RetrieveChunks", "failure", time.Since(start))
 		return nil, fmt.Errorf("could not find chunks for batchHeaderHash %v, blob index: %v, quorumID: %v", batchHeaderHash, in.GetBlobIndex(), in.GetQuorumId())
 	}
+	if !s.config.EnableGnarkBundleEncoding && format == pb.ChunkEncoding_GNARK {
```
bxue-l2, line 310: there would be a case when gnark blob is enabled but, the client forgot to set the flag. why not make two interfaces altogether, and remove the flag?

node/grpc/server_test.go
```diff
@@ -372,7 +410,79 @@ func TestAttestBatch(t *testing.T) {
...
+	assert.NoError(t, err)
+	assert.Equal(t, recovered, chunk)
+
+	retrievalReply, err = server.RetrieveChunks(ctx, &pb.RetrieveChunksRequest{
+		BatchHeaderHash: batchHeaderHash[:],
+		BlobIndex:       0,
+		QuorumId:        1,
+	})
+	assert.NoError(t, err)
+	assert.Equal(t, pb.ChunkEncoding_UNKNOWN, retrievalReply.Encoding)
```
bxue-l2, line 442: nice, so you handled the case when bundle is empty when operator is not in that quorum. But you also return the unknown when blob cannot be retrieved, but since you return the error there,so it would fine I think

node/grpc/server_test.go
```diff
@@ -388,6 +498,7 @@ func TestRetrieveChunks(t *testing.T) {
 		QuorumId:        0,
 	})
 	assert.NoError(t, err)
+	assert.Equal(t, pb.ChunkEncoding_GOB, retrievalReply.Encoding)
```
bxue-l2, line 501: why is it gob, since the storeChunkfunction is true?

bxue-l2: I think in the node/store.go func DecodeChunks, you regard format 0 as ChunkEncoding_GOB. It is confusing since it is sometimes 0 sometimes 2.

bxue-l2: but anyway, shouldn't the format be 1, since it is requested?

node/grpc/server.go
jianoaix: Having a separate interface (from `RetrieveChunks`) will still have the issue of "old-new" communication: if a new client talks to an old server (which doesn't have the new API), it will fail.

node/grpc/server_test.go
jianoaix: It should be fine, empty bundle has no header to tell what's the encoding format anyway.

node/grpc/server_test.go
jianoaix: By default, `EnableGnarkBundleEncoding` flag is false, so the node will continue to return GOB, even though the underlying data stored in DB is in GNARK.

The ID for protobuf and the enum in golang are different for a reason, added the documentation: https://github.com/Layr-Labs/eigenda/pull/654/commits/797838931b2674ff5784431d2a305ae56035233a

node/grpc/utils.go
```diff
@@ -59,12 +59,16 @@ func GetBlobMessages(in *pb.StoreChunksRequest, numWorkers int) ([]*core.BlobMes
...
-						return
+					if len(bundle.GetBundle()) > 0 {
+						bundleMsg, err := new(core.Bundle).Deserialize(bundle.GetBundle())
+						if err != nil {
+							resultChan <- err
+							return
+						}
+						bundles[uint8(quorumID)] = bundleMsg
+					} else {
+						bundles[uint8(quorumID)] = make([]*encoding.Frame, 0)
```
bxue-l2, line 70: good catch. I think it is handling the case, when there is quorum which the operator does not belong to. But I thought you already catch it here, https://github.com/Layr-Labs/eigenda/blob/master/disperser/batcher/grpc/dispatcher.go#L376

bxue-l2: I guess the question is if the grpc would return a nil, or empty array.

bxue-l2 (commented):
lgtm

bxue-l2 (approved):

node/grpc/utils.go
jianoaix: If the operator is not in this quorum, then the bundle is an empty []byte (not nil), so len(bundle.GetBundle()) is 0

PR #658, opened 2024-07-23
Fix the operator stake share basis point calculation

jianoaix:
The current `100000` is not providing basis point, but 0.1 basis point. It needs to be `10000`.

Otherwise, we will see `17351.099508447942` basis point for a large operator, which means more than 100% of stake (it should be just 17.35%)

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


jianoaix:
ping

jianoaix:
ping again

jianoaix:
This basically reverts https://github.com/Layr-Labs/eigenda/pull/590, which wasn't right fix to *100000

dmanc (approved):

PR #670, opened 2024-07-29
Add BatchConfirmer

ian-shim:
BatchConfirmer will take the `MinibatchResponse`s and confirm the full batch by getting & aggregating the nodes' signatures for the full batch. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/batcher/batch_confirmer.go
+++ b/disperser/batcher/batch_confirmer.go
@@ -1,0 +1,532 @@
+package batcher
+
+import (
+	"bytes"
+	"context"
+	"errors"
+	"fmt"
+	"math"
+	"math/big"
+	"time"
+
+	"github.com/Layr-Labs/eigenda/common"
+	"github.com/Layr-Labs/eigenda/core"
+	"github.com/Layr-Labs/eigenda/disperser"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	"github.com/ethereum/go-ethereum/accounts/abi"
+	"github.com/ethereum/go-ethereum/core/types"
+	"github.com/hashicorp/go-multierror"
+)
+
+type BatchConfirmerConfig struct {
+	PullInterval                 time.Duration
+	DispersalTimeout             time.Duration
+	DispersalStatusCheckInterval time.Duration
+	AttestationTimeout           time.Duration
+	SRSOrder                     int
+	NumConnections               int
+	MaxNumRetriesPerBlob         uint
+}
+
+type BatchConfirmer struct {
+	BatchConfirmerConfig
+
+	BlobStore      disperser.BlobStore
+	MinibatchStore MinibatchStore
+	Dispatcher     disperser.Dispatcher
+	EncoderClient  disperser.EncoderClient
+
+	ChainState         core.IndexedChainState
+	EncodingStreamer   *EncodingStreamer
+	Aggregator         core.SignatureAggregator
+	Transactor         core.Transactor
+	TransactionManager TxnManager
+	Minibatcher        *Minibatcher
+
+	ethClient common.EthClient
+	logger    logging.Logger
+}
+
+func NewBatchConfirmer(
+	config BatchConfirmerConfig,
+	blobStore disperser.BlobStore,
+	minibatchStore MinibatchStore,
+	dispatcher disperser.Dispatcher,
+	chainState core.IndexedChainState,
+	assignmentCoordinator core.AssignmentCoordinator,
+	encodingStreamer *EncodingStreamer,
+	aggregator core.SignatureAggregator,
+	ethClient common.EthClient,
+	transactor core.Transactor,
+	txnManager TxnManager,
+	minibatcher *Minibatcher,
+	logger logging.Logger,
+) (*BatchConfirmer, error) {
+	return &BatchConfirmer{
+		BatchConfirmerConfig: config,
+
+		BlobStore:        blobStore,
+		MinibatchStore:   minibatchStore,
+		Dispatcher:       dispatcher,
+		EncodingStreamer: encodingStreamer,
+
+		ChainState:         chainState,
+		Aggregator:         aggregator,
+		Transactor:         transactor,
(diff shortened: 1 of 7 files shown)


disperser/batcher/batch_confirmer.go
```diff
@@ -0,0 +1,536 @@
...
+
+	for blobIndex, metadata := range batchData.blobs {
+		// Mark the blob failed if it didn't get enough signatures.
+		status := disperser.InsufficientSignatures
+
+		var proof []byte
+		if isBlobAttested(batchData.aggSig.QuorumResults, batchData.blobHeaders[blobIndex]) {
+			status = disperser.Confirmed
+			// generate inclusion proof
+			if blobIndex >= len(batchData.blobHeaders) {
```
mooselumph, line 173: It's unclear to me when this would ever happen. Isn't in at invariant that the lengths should be the same?

disperser/batcher/batch_confirmer.go
```diff
@@ -0,0 +1,536 @@
...
+	return result.ErrorOrNil()
+}
+
+func (b *BatchConfirmer) HandleSingleBatch(ctx context.Context) error {
+	currentBlock, err := b.ethClient.BlockNumber(ctx)
+	if err != nil {
+		return fmt.Errorf("error getting current block number: %w", err)
+	}
+
+	err = b.EncodingStreamer.UpdateReferenecBlock(uint(currentBlock))
```
mooselumph, line 301: nit: there seems to be a typo in the name of this function

disperser/batcher/batch_confirmer.go
```diff
@@ -0,0 +1,536 @@
...
+	batchDispersed := false
+	stateUpdateTicker.Reset(b.DispersalStatusCheckInterval)
+	ctxWithTimeout, cancel := context.WithTimeout(ctx, b.DispersalTimeout)
+	defer cancel()
+	for !batchDispersed {
+		select {
+		case <-ctxWithTimeout.Done():
+			return ctxWithTimeout.Err()
+		case <-stateUpdateTicker.C:
+			batchDispersed, err = b.MinibatchStore.BatchDispersed(ctx, batch.ID)
```
mooselumph, line 337: Note: There doesn't seem to be any check to predict whether AttestBatch will deterministically fail based on the dispersal status of the minibatches. This is probably fine for now; the more important optimization would be to check which set of minibatches to include in the batch, which we have agreed can follow later.

disperser/batcher/batch_confirmer.go
ian-shim: Yeah, not sure why this was added before. 
If this condition were to ever happen, `batchData.blobHeaders[blobIndex]` would have resulted in out of bounds exception in L170. 
I don't see any errors logs from this block as well.
I'll remove this block.

disperser/batcher/batch_confirmer.go
ian-shim: nice catch

disperser/batcher/batch_confirmer.go
```diff
@@ -0,0 +1,531 @@
...
+}
+
+func (b *BatchConfirmer) Start(ctx context.Context) error {
+	err := b.ChainState.Start(ctx)
+	if err != nil {
+		return err
+	}
+	// Wait for few seconds for indexer to index blockchain
+	// This won't be needed when we switch to using Graph node
+	time.Sleep(indexerWarmupDelay)
```
mooselumph, line 91: Is this still used?

ian-shim: It _could_ still be used if we wanted to (i.e. testing)

disperser/batcher/batch_confirmer.go
```diff
@@ -0,0 +1,531 @@
...
+	}
+
+	// Make sure all minibatches in the batch have been dispersed
+	batchDispersed := false
+	stateUpdateTicker.Reset(b.DispersalStatusCheckInterval)
+	ctxWithTimeout, cancel := context.WithTimeout(ctx, b.DispersalTimeout)
+	defer cancel()
+	for !batchDispersed {
+		select {
+		case <-ctxWithTimeout.Done():
```
mooselumph, line 329: What happens if this timeout is hit? What happens to the minibatches?

It seems possible that it will try to process this batch again the next round...

disperser/batcher/batch_confirmer.go
```diff
@@ -0,0 +1,531 @@
...
+	// Aggregate the signatures
+	b.logger.Debug("Aggregating signatures...")
+
+	quorumAttestation, err := b.Aggregator.ReceiveSignatures(ctx, batchState.OperatorState, batchHeaderHash, replyChan)
+	if err != nil {
+		_ = b.handleFailure(ctx, batchState.BlobMetadata, FailAggregateSignatures)
+		return fmt.Errorf("error receiving and validating signatures: %w", err)
+	}
+	operatorCount := make(map[core.QuorumID]int)
+	signerCount := make(map[core.QuorumID]int)
```
mooselumph, line 399: Are the `signerCount` and `operatorCount` actually used?

mooselumph (approved):

disperser/batcher/batch_confirmer.go
ian-shim: Yes, that's the current behavior now. 
This timeout should be longer than the [dispersal timeout](https://github.com/Layr-Labs/eigenda/blob/master/disperser/batcher/minibatcher.go#L292), then all requests will have succeeded/failed before this timeout. 
In case batcher shutdown mid-dispersal, the minibatcher recovery logic should clear the dispersal requests with no matching responses. 
Is this too brittle?

disperser/batcher/batch_confirmer.go
ian-shim: These are meant for reporting metrics, but there are no metrics on minibatching right now. 
I added a log line that prints these numbers for now.

PR #671, opened 2024-07-30
[node] Add `StoreBlobs` endpoint for minibatch streaming

ian-shim:
This PR adds `StoreBlobs` endpoint to DA node. 
This method can be used to disperse blobs and have signatures for each blob returned, instead of a batch containing those blobs. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


node/grpc/server.go
```diff
@@ -251,7 +253,77 @@ func (s *Server) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (*p
 }
 
 func (s *Server) StoreBlobs(ctx context.Context, in *pb.StoreBlobsRequest) (*pb.StoreBlobsReply, error) {
-	return nil, errors.New("StoreBlobs is not implemented")
+	start := time.Now()
+
+	// Validate the request
+	if in.GetReferenceBlockNumber() == 0 {
```
mooselumph, line 259: Should we put some of this in a validation function?

mooselumph: Can we generalize the `validateStoreChunksRequest` and reuse in both places?

node/store.go
```diff
@@ -66,31 +66,108 @@ func NewLevelDBStore(path string, logger logging.Logger, metrics *Metrics, block
...
+			}
+			numBlobsDeleted += numDeleted
+
+			numDeleted, err = s.deleteNBatches(currentTimeUnixSec, numBatchesToDeleteAtomically)
+			if err != nil {
+				return numBatchesDeleted, numBlobsDeleted, err
 			}
 			// When there is no error and we didn't delete any batch, it means we have
 			// no obsolete batches to delete, so we can return.
 			if numDeleted == 0 {
```
mooselumph, line 92: This seems like it will return if there are no more batches to delete even if there are still blobs to delete.

node/grpc/server.go
ian-shim: If this method is getting too long, I'd prefer creating another method `validateStoreBlobsRequest` instead of generalizing `validateStoreChunksRequest` and using it in both places. 
After the migration, StoreChunks method will be deprecated. I think it's better to make two codepaths somewhat independent and clear for each use case, so that eventually when one codepath is removed, what's left is clear for those without context.

node/store.go
ian-shim: Thanks for catching this!

node/store.go
```diff
@@ -318,6 +398,147 @@ func (s *Store) StoreBatch(ctx context.Context, header *core.BatchHeader, blobs
...
+		expirationValue = append(expirationValue, blobHeaderHash[:]...)
+
+		// Get raw chunks
+		start := time.Now()
+		format := GetBundleEncodingFormat(rawBlob)
+		rawBundles := make(map[core.QuorumID][]byte)
+		rawChunks := make(map[core.QuorumID][][]byte)
+		for i, bundle := range rawBlob.GetBundles() {
+			quorumID := uint8(rawBlob.GetHeader().GetQuorumHeaders()[i].GetQuorumId())
+			if format == core.GnarkBundleEncodingFormat {
```
mooselumph, line 472: It seems unideal that we are processing the chunk format both here at when we process the raw blobs in `GetBlobMessages` of `utils.go`. I'd prefer it if `GetBlobMessages` or some other utility function did all of this processing and then returned all formats of the chunks needed by validation and storage, so that the storage functions can be format agnostic.

mooselumph (commented):
A few comments

node/store.go
ian-shim: Agree it's a bit agree. I was mostly leaving the implementation same as `StoreBatch` to minimize risk
On the bright side, this branching based on encoding format will also go away soon when all blobs are encoded with gnark

mooselumph (approved):

PR #666, opened 2024-07-25
771 read traffic generator

cody-littley:
## Why are these changes needed?

The current traffic generation utilty writes blobs to eigenDA, but does not read them back. This PR augments the generator utility so that it is also capable of reading the blobs it has previously written.

## Checks

- [x] I've made sure the lint is passing in this PR.
- [ ] I've made sure the tests are passing. Note that there might be a few flaky tests, in that case, please comment that they are not relevant.
- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests
   - [ ] This PR is not tested :(

PR #674, opened 2024-07-30
Configuration for new generator functionality.

cody-littley:

This code was broken off of another PR that got too large: https://github.com/Layr-Labs/eigenda/pull/666

This PR contains configuration for the new generator functionality. It is not utilized in this branch, but it will be utilized once the rest of the code merges.

## Checks

- Testing Strategy
   - [x] Unit tests (in the other PR)
   - [ ] Integration tests

### DIFF
--- a/tools/traffic/config/config.go
+++ b/tools/traffic/config/config.go
@@ -1,0 +1,137 @@
+package config
+
+import (
+	"errors"
+	"time"
+
+	"github.com/Layr-Labs/eigenda/common"
+	"github.com/urfave/cli"
+)
+
+// Config configures a traffic generator.
+type Config struct {
+	// Logging configuration.
+	LoggingConfig common.LoggerConfig
+	// The hostname of the disperser.
+	DisperserHostname string
+	// The port of the disperser.
+	DisperserPort string
+	// The timeout for the disperser.
+	DisperserTimeout time.Duration
+	// Whether to use a secure gRPC connection to the disperser.
+	DisperserUseSecureGrpcFlag bool
+	// The private key to use for signing requests.
+	SignerPrivateKey string
+	// Custom quorum numbers to use for the traffic generator.
+	CustomQuorums []uint8
+	// Whether to disable TLS for an insecure connection.
+	DisableTlS bool
+	// The port at which the metrics server listens for HTTP requests.
+	MetricsHTTPPort string
+	// The hostname of the Ethereum client.
+	EthClientHostname string
+	// The port of the Ethereum client.
+	EthClientPort string
+	// The address of the BLS operator state retriever smart contract, in hex.
+	BlsOperatorStateRetriever string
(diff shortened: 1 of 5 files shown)


tools/traffic/config/flags.go
```diff
@@ -0,0 +1,340 @@
...
+	"github.com/Layr-Labs/eigenda/common"
+	"github.com/urfave/cli"
+)
+
+const (
+	FlagPrefix = "traffic-generator"
+	envPrefix  = "TRAFFIC_GENERATOR"
+)
+
+var (
```
ian-shim, line 15: Instead of listing flags for dependent packages again, we could import the flags directly like [here](https://github.com/Layr-Labs/eigenda/blob/master/retriever/flags/flags.go#L99-L103)

tools/traffic/config/config.go
```diff
@@ -0,0 +1,137 @@
...
+import (
+	"errors"
+	"time"
+
+	"github.com/Layr-Labs/eigenda/common"
+	"github.com/urfave/cli"
+)
+
+// Config configures a traffic generator.
+type Config struct {
```
ian-shim, line 12: We could reuse the configs from dependencies directly like we do [here](https://github.com/Layr-Labs/eigenda/blob/master/retriever/config.go#L16-L21)

tools/traffic/config/flags.go
cody-littley: change made

tools/traffic/config/config.go
cody-littley: change made

tools/traffic/config/config.go
```diff
@@ -0,0 +1,119 @@
...
+	DisperserClientConfig *clients.Config
+
+	// Configuration for the retriever client.
+	RetrievalClientConfig *retriever.Config
+
+	// Configuration for the graph.
+	TheGraphConfig *thegraph.Config
+
+	// Configuration for the EigenDA client.
+	EigenDaClientConfig *clients.EigenDAClientConfig
```
ian-shim, line 31: nit: EigenDAClientConfig

tools/traffic/config/flags.go
```diff
@@ -0,0 +1,262 @@
+package config
```
ian-shim, line 1: are these moved from `/tools/traffic/flags.go`? or are these being added in addition to the existing ones?

cody-littley: In the short term, the new `flags.go` file will duplicate the original file in the `master` branch, but will not be used for anything. When the rest of the code from [this branch](https://github.com/Layr-Labs/eigenda/pull/666) merges, the new `flags.go` will entirely replace the old one.

I'm merging this way so that I can split off the config changes into a stand alone PR. It can't replace the old flags because it's not fully backwards compatible.

ian-shim (approved):

PR #675, opened 2024-07-31
add configurable max blob size

bxue-l2:

This PR makes max blob size a configurable parameters, validate if the blob is power of 2.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/cmd/apiserver/flags/flags.go
+++ b/disperser/cmd/apiserver/flags/flags.go
@@ -94,6 +94,13 @@ var (
 		EnvVar:   common.PrefixEnvVar(envVarPrefix, "RATE_BUCKET_STORE_SIZE"),
 		Required: false,
 	}
+	MaxBlobSize = cli.IntFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "max-blob-size"),
+		Usage:    "max blob size disperser is accepting",
+		Value:    2_097_152,
+		EnvVar:   common.PrefixEnvVar(envVarPrefix, "MAX_BLOB_SIZE"),
+		Required: false,
+	}
 )
 
 var requiredFlags = []cli.Flag{
@@ -112,6 +119,7 @@ var optionalFlags = []cli.Flag{
 	BucketStoreSize,
 	GrpcTimeoutFlag,
 	ShadowTableNameFlag,
+	MaxBlobSize,
 }
 
 // Flags contains the list of configuration options available to the binary.
--- a/disperser/cmd/apiserver/main.go
+++ b/disperser/cmd/apiserver/main.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"log"
+	"math"
 	"os"
 	"time"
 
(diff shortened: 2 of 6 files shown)


disperser/cmd/apiserver/main.go
```diff
@@ -114,6 +115,14 @@ func RunDisperserServer(ctx *cli.Context) error {
 		ratelimiter = ratelimit.NewRateLimiter(reg, globalParams, bucketStore, logger)
 	}
 
+	if config.MaxBlobSize < 0 || config.MaxBlobSize > 64*1024*1024 {
+		return fmt.Errorf("configured max blob size is invalid %v", config.MaxBlobSize)
+	}
+
+	if int64(math.Log2(float64(config.MaxBlobSize))) == int64(math.Log2(float64(config.MaxBlobSize-1))) {
```
jianoaix, line 122: It's faster to use bit operation, `config.MaxBlobSize > 0 && (config.MaxBlobSize&(config.MaxBlobSize-1)) == 0` should work

disperser/cmd/apiserver/main.go
```diff
@@ -114,6 +115,14 @@ func RunDisperserServer(ctx *cli.Context) error {
 		ratelimiter = ratelimit.NewRateLimiter(reg, globalParams, bucketStore, logger)
 	}
 
+	if config.MaxBlobSize < 0 || config.MaxBlobSize > 64*1024*1024 {
```
jianoaix, line 118: Should it fail if config.MaxBlobSize == 0?

disperser/cmd/apiserver/flags/flags.go
```diff
@@ -94,6 +94,13 @@ var (
 		EnvVar:   common.PrefixEnvVar(envVarPrefix, "RATE_BUCKET_STORE_SIZE"),
 		Required: false,
 	}
+	MaxBlobSize = cli.IntFlag{
```
jianoaix, line 97: It looks no issue if different disperser replicas are using different MaxBlobSize.
Should we add validation to the Node to check if the blob size is under a limit?

disperser/cmd/apiserver/main.go
bxue-l2: yeah, I will exclude 0

disperser/cmd/apiserver/flags/flags.go
bxue-l2: I don't think there is issue. But we want it, it has to be different deployment, which the current devops repo has yet offered that. 

I don't think we need to artificially constraint on blob size on the DA node size now, as far as the disperser isn't decentralized, but yes, once we got to the step. We need to have some rate limit or blob constraint. For now, I think we are good.

Empirically, there is no impact on validation speed with respect to the blob size.

disperser/cmd/apiserver/main.go
bxue-l2: this only runs once. I think it is more readable this way

bxue-l2: but your solution works

jianoaix: I think there are utility functions to check if a number is power of 2 (in encoding or common package). Reusing that has no impact on readabilty.

disperser/cmd/apiserver/main.go
```diff
@@ -114,6 +115,14 @@ func RunDisperserServer(ctx *cli.Context) error {
 		ratelimiter = ratelimit.NewRateLimiter(reg, globalParams, bucketStore, logger)
 	}
 
+	if config.MaxBlobSize <= 0 || config.MaxBlobSize > 64*1024*1024 {
```
jianoaix, line 118: So we want to support 64MiB?

disperser/cmd/apiserver/main.go
bxue-l2: that function is in "github.com/Layr-Labs/eigenda/encoding/fft". I don't want to add a new strange dependancy, what do you think

disperser/cmd/apiserver/main.go
bxue-l2: It is just a check, we might experiment with it

jianoaix (approved):

disperser/cmd/apiserver/main.go
jianoaix: ideally we can pull it out to `common/` and share across the codebase

PR #676, opened 2024-07-31
[node] `AttestBatch` endpoint

ian-shim:
This PR adds `AttestBatch` endpoint to DA node.
This endpoint can be used to attest to a batch of blobs represented by a list of blob header hashes that the node had previously received. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


core/validator.go
```diff
@@ -87,12 +88,20 @@ func (v *shardValidator) UpdateOperatorID(operatorID OperatorID) {
...
+	headers := make([]*BlobHeader, len(blobs))
+	for i, blob := range blobs {
+		headers[i] = blob.BlobHeader
+	}
+	err := ValidateBatchHeaderRoot(batchHeader, headers)
 	if err != nil {
 		return err
 	}
 
+	return v.ValidateBlobs(blobs, operatorState, pool)
```
anupsv, line 100: This is a all or none check ? I.e either all blobs are valid or error is returned ?

core/validator.go
```diff
@@ -87,12 +88,20 @@ func (v *shardValidator) UpdateOperatorID(operatorID OperatorID) {
 }
 
 func (v *shardValidator) ValidateBatch(batchHeader *BatchHeader, blobs []*BlobMessage, operatorState *OperatorState, pool common.WorkerPool) error {
-
-	err := validateBatchHeaderRoot(batchHeader, blobs)
+	headers := make([]*BlobHeader, len(blobs))
```
anupsv, line 91: Are there limits to the number of blobs that can be requested ?

node/utils.go
```diff
@@ -1,88 +1,216 @@
...
-// EncodeBlobKey returns an encoded key as blob identification.
-func EncodeBlobKey(batchHeaderHash [32]byte, blobIndex int, quorumID core.QuorumID) ([]byte, error) {
-	buf := bytes.NewBuffer(batchHeaderHash[:])
-	err := binary.Write(buf, binary.LittleEndian, int32(blobIndex))
-	if err != nil {
-		return nil, err
+// GetBatchHeader constructs a core.BatchHeader from a proto of pb.StoreChunksRequest.
+// Note the StoreChunksRequest is validated as soon as it enters the node gRPC
+// interface, see grpc.Server.validateStoreChunkRequest.
+func GetBatchHeader(in *pb.BatchHeader) (*core.BatchHeader, error) {
```
ian-shim, line 21: moving a bunch of util methods from `node/grpc/utils.go` to prevent circular dependency

core/validator.go
ian-shim: Yes, it's all or none check for now. 
Down the road we may filter out invalid blobs from the set and only sign on valid blobs

core/validator.go
ian-shim: There is a limit on the size of the payload but not on the number of blobs

node/utils.go
```diff
@@ -1,155 +1,216 @@
...
 	pb "github.com/Layr-Labs/eigenda/api/grpc/node"
 	"github.com/Layr-Labs/eigenda/common/pubip"
 	"github.com/Layr-Labs/eigenda/core"
+	"github.com/Layr-Labs/eigenda/encoding"
+	"github.com/consensys/gnark-crypto/ecc/bn254"
+	"github.com/consensys/gnark-crypto/ecc/bn254/fp"
+	"github.com/gammazero/workerpool"
 )
 
-const (
```
ian-shim, line 15: store related utils are moved to `store_utils.go`

node/node.go
```diff
@@ -589,6 +593,60 @@ func (n *Node) SignBlobs(blobs []*core.BlobMessage, referenceBlockNumber uint) (
 	return signatures, nil
 }
```
mooselumph, line 595: Should this be part of core?

node/node.go
```diff
@@ -589,6 +593,60 @@ func (n *Node) SignBlobs(blobs []*core.BlobMessage, referenceBlockNumber uint) (
...
 
+// ValidateBlobHeadersRoot validates the blob headers root hash
+// by comparing it with the merkle tree root hash of the blob headers.
+// It also checks if all blob headers have the same reference block number
+func (n *Node) ValidateBlobHeadersRoot(ctx context.Context, blobHeaderHashes [][32]byte, root [32]byte) error {
```
mooselumph, line 599: Naming seems a little misleading. It's doing the following:
1. Check that all blobs have been previously validated and stored. 
2. Check reference block numbers
3. Check root

`ValidateBatchContents`?

node/node.go
```diff
@@ -589,6 +593,60 @@ func (n *Node) SignBlobs(blobs []*core.BlobMessage, referenceBlockNumber uint) (
...
+	leafs := make([][]byte, 0)
+	referenceBlockNumber := uint32(0)
```
mooselumph, line 601: We should be comparing the `referenceBlockNumber`s againts the one in the batch being attested. What if we just passed in the entire `BatchHeader` and used its reference block number here instead of just passing in the merkle root?

ian-shim: That makes sense!

node/store.go
```diff
@@ -223,6 +223,13 @@ func (s *Store) deleteNBatches(currentTimeUnixSec int64, numBatches int) (int, e
 			size += int64(len(blobIter.Value()))
 		}
 		blobIter.Release()
+
+		// Blob index.
+		blobIndexIter := s.db.NewIterator(EncodeBlobIndexKeyPrefix(batchHeaderHash))
```
mooselumph, line 228: If the intent is to mostly deprecate this function, should we put this code somewhere else?

node/store.go
```diff
@@ -539,6 +513,50 @@ func (s *Store) StoreBlobs(ctx context.Context, blobs []*core.BlobMessage, blobs
...
 }
 
+func (s *Store) StoreBatchBlobMapping(ctx context.Context, batchHeaderHash [32]byte, blobHeaderHashes [][32]byte) error {
+	start := time.Now()
+	// The key/value pairs that need to be written to the local database.
+	keys := make([][]byte, 0)
+	values := make([][]byte, 0)
+
+	expirationTime := s.expirationTime()
+	expirationKey := EncodeBatchExpirationKey(expirationTime)
```
mooselumph, line 523: We allow multiple blobs to have the same expiration time, but not for batches. Is this correct? Esp. if we were to allow multiple dispersers?

node/store.go
ian-shim: Which function are you referring to? `deleteNBatches` or `EncodeBlobIndexKeyPrefix`?

node/store.go
ian-shim: Yes, a blob have its [own expiration time](https://github.com/Layr-Labs/eigenda/blob/master/node/store.go#L424). This would just expire the batch <-> blob mapping.

mooselumph: What if there happen to be two batches that get created with the same expiration time? It seems like we're worrying about that case for blobs but not for batch mappings.

node/store.go
mooselumph: Referring to `deleteNBatches`. It's currently handling both "Batches" which will be deprecated and "Batch Mappings" which won't be?

node/store.go
ian-shim: I see what you mean. That's a valid edge case. 
In fact, the existing `StoreChunks` endpoint would have this issue as well. 

I also see an issue with how we're adding expirations for blobs - there could be a concurrency issue that can void a write without a lock. 
I'll refactor the expiration key such that the expiration key is in format: 
- `_EXPIRATION_<EXPIRATION_TIME>_<BATCH_HEADER_HASH>` -> batch header hash
- `_BLOBEXPIRATION_<EXPIRATION_TIME>_<BLOB_HEADER_HASH>` -> blob header hash

This way we can still query expirations by expiration time, but each batch/blob gets its own record. Wdyt? 

cc @jianoaix

node/store.go
ian-shim: Sure, agree having a separate method would make it easier to deprecate the functionality in the future

PR #677, opened 2024-07-31
Added blob table data structure.

cody-littley:

This is code split off from another PR that got too big: https://github.com/Layr-Labs/eigenda/pull/666

This PR adds a data structure called a `BlobTable`, which is used by the traffic generator to track blobs.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests


tools/traffic/table/blob_metadata.go
```diff
@@ -0,0 +1,77 @@
...
+
+// BlobMetadata encapsulates various information about a blob written by the traffic generator.
+type BlobMetadata struct {
+	// key of the blob, set when the blob is initially uploaded.
+	key *[]byte
+
+	// checksum of the blob.
+	checksum *[16]byte
+
+	// batchHeaderHash of the blob in bytes.
```
ian-shim, line 11: this comment should be updated

tools/traffic/table/blob_table.go
```diff
@@ -0,0 +1,125 @@
...
+	return table.blobs[index]
+}
+
+// Add a blob to the requiredReads.
+func (table *BlobTable) Add(blob *BlobMetadata) {
+	table.lock.Lock()
+	defer table.lock.Unlock()
+
+	blob.index = table.size
+	table.blobs[table.size] = blob
```
ian-shim, line 57: what if the `table.size` is out of bound?

tools/traffic/table/blob_table.go
```diff
@@ -0,0 +1,125 @@
...
+	size uint
+
+	// lock is used to synchronize access to the requiredReads.
+	lock sync.Mutex
+}
+
+// NewBlobTable creates a new BlobTable instance.
+func NewBlobTable() BlobTable {
+	return BlobTable{
+		blobs: make([]*BlobMetadata, 1024),
```
ian-shim, line 26: why don't we make the size of this slice variable?

tools/traffic/table/blob_table.go
```diff
@@ -0,0 +1,125 @@
...
+
+	// blobs contains all blobs currently tracked by the requiredReads.
+	blobs []*BlobMetadata
+
+	// size describes the total number of blobs currently tracked by the requiredReads.
+	// size may be smaller than the capacity of the blobs slice.
+	size uint
+
+	// lock is used to synchronize access to the requiredReads.
+	lock sync.Mutex
```
ian-shim, line 20: we can use `RWMutex` for more granular control

tools/traffic/table/blob_table.go
```diff
@@ -0,0 +1,125 @@
...
+func (table *BlobTable) Add(blob *BlobMetadata) {
+	table.lock.Lock()
+	defer table.lock.Unlock()
+
+	blob.index = table.size
+	table.blobs[table.size] = blob
+	table.size++
+}
+
+// AddOrReplace adds a blob to the requiredReads if there is capacity or replaces an existing blob at random
```
ian-shim, line 61: What's the `requiredReads`?

tools/traffic/table/blob_table.go
```diff
@@ -0,0 +1,125 @@
...
+func (table *BlobTable) AddOrReplace(blob *BlobMetadata, maximumCapacity uint) {
+	if maximumCapacity == 0 {
+		return
+	}
+
+	table.lock.Lock()
+	defer table.lock.Unlock()
+
+	if table.size >= maximumCapacity {
+		// replace random existing blob
```
ian-shim, line 72: what if the existing blob hasn't been retrieved the required number of times?

tools/traffic/table/blob_table.go
```diff
@@ -0,0 +1,125 @@
...
+	defer table.lock.Unlock()
+
+	if table.size == 0 {
+		return nil, false
+	}
+
+	blob := table.blobs[rand.Int31n(int32(table.size))]
+
+	removed := false
+	if decrement && blob.remainingReadPermits != -1 {
```
ian-shim, line 99: shouldn't the condition be `blob.remainingReadPermits != -1`?
Say `blob.remainingReadPermits` is 0 before this call. Then it gets decremented to -1 inside this block and is never removed

tools/traffic/table/blob_table.go
```diff
@@ -0,0 +1,125 @@
...
+			table.remove(blob)
+			removed = true
+		}
+	}
+
+	return blob, removed
+}
+
+// remove a blob from the requiredReads.
+func (table *BlobTable) remove(blob *BlobMetadata) {
```
ian-shim, line 111: Is there a particular reason why we need table to be a slice?
I feel like removing/adding elements would be a lot simpler and less brittle if we used a map

cody-littley: Primary reason why I use a slice is to give an `O(1)` implementation of `GetRandom()`. Can you think of a good way to do this backed by a map?

ian-shim: Do we need to access a random element from the slice? 
For sampling, could we just use the first element from the map?

tools/traffic/table/blob_metadata.go
cody-littley: oops, fixed

tools/traffic/table/blob_table.go
cody-littley: Fixed.

Good catch. It just so happened that the original capacity was 1024, which exceeded the number of blobs I was testing with.

I also made the initial capacity 0, thus making existing unit tests sensitive to this problem.

tools/traffic/table/blob_table.go
cody-littley: That would work as long as we only add blobs. When blobs are removed from the table, the size of the table may actually be smaller than the size of this slice.

tools/traffic/table/blob_table.go
cody-littley: done

tools/traffic/table/blob_table.go
cody-littley: Documentation was out of date. Fixed.

```
// AddOrReplace is equivalent to Add if there is capacity, or replaces an existing blob at random
// if the is no remaining capacity. This method is a no-op if maximumCapacity is 0.
```

tools/traffic/table/blob_table.go
cody-littley: Then it is removed.

This is not a problem for our use case though, since we keep two blob tables: one for required blobs, and another for optional blobs. The code never calls `AddOrReplace()` on the table that contains the required blobs.

tools/traffic/table/blob_table.go
cody-littley: Read permits for blobs in the table are never 0. They are always -1 or greater than 0, and removed the moment the count reaches 0. 

Just in case, I added an assertion in the `NewBlobMetadata()` method to validate that this invariant is not violated.

```
	if readPermits == 0 {
		panic("readPermits must be greater than 0, or -1 for unlimited reads")
	}
```

tools/traffic/table/blob_table.go
cody-littley: The entire reason why this complex data structure exists in the first place is to facilitate random access. 😜 

For the blobs with a number of required reads: yes, we could get away without random access. But I'm under the impression a random access pattern is preferable to a fixed one when simulating workloads like this. 

For the pool of optional blobs to read, I think random access is necessary. Otherwise, we'd just be reading the same blob over and over until we get a new blob to start reading.

I'm open to discussing this more in depth if you are not convinced by my reasoning.

tools/traffic/table/blob_table.go
```diff
@@ -0,0 +1,125 @@
...
+func NewBlobTable() BlobTable {
+	return BlobTable{
+		blobs: make([]*BlobMetadata, 1024),
+		size:  0,
+	}
+}
+
+// Size returns the total number of blobs currently tracked by the requiredReads.
+func (table *BlobTable) Size() uint {
+	table.lock.Lock()
```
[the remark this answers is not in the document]
cody-littley: Yup, I've switched to a read/write lock pattern.

tools/traffic/table/blob_metadata.go
```diff
@@ -0,0 +1,77 @@
+package table
+
+// BlobMetadata encapsulates various information about a blob written by the traffic generator.
+type BlobMetadata struct {
+	// key of the blob, set when the blob is initially uploaded.
+	key *[]byte
```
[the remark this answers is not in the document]
cody-littley: Interesting, TIL. Will simplify code to not pass slice pointers.

tools/traffic/table/blob_metadata.go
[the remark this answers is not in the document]
cody-littley: Field removed.

tools/traffic/table/blob_table.go
```diff
@@ -0,0 +1,140 @@
...
+		}
+	}
+
+	return blob, removed
+}
+
+// remove a blob from the requiredReads.
+func (table *BlobTable) remove(blob *BlobMetadata) {
+	if table.blobs[blob.index] != blob {
+		panic(fmt.Sprintf("blob %x is not not present in the requiredReads at index %d", blob.Key(), blob.index))
```
ian-shim, line 128: Should we log at error level and handle this case gracefully vs. crashing the whole program?

tools/traffic/table/blob_table.go
ian-shim: > For the blobs with a number of required reads: yes, we could get away without random access. But I'm under the impression a random access pattern is preferable to a fixed one when simulating workloads like this.

We _could_ make get semi random access with a map in constant time (just generate random number n < 10 and pick _n_th element from map) if the access doesn't require sampling from uniform distribution. 

> For the pool of optional blobs to read, I think random access is necessary. Otherwise, we'd just be reading the same blob over and over until we get a new blob to start reading.

I don't think optional blob reads were ever part of the spec. The primary goal for this observability tool is to monitor if the network can handle a given retrieval traffic. Are there benefits of saturating the network with optional reads?

Not a big deal, but would bias toward simplicity vs. optimization

tools/traffic/table/blob_table.go
ian-shim: ohh got it

tools/traffic/table/blob_metadata.go
```diff
@@ -0,0 +1,71 @@
...
+// remaining reads permitted against this blob. If -1 then an unlimited number of reads are permitted.
+func NewBlobMetadata(
+	key []byte,
+	checksum *[16]byte,
+	size uint,
+	blobIndex uint,
+	readPermits int) *BlobMetadata {
+
+	if readPermits == 0 {
+		panic("readPermits must be greater than 0, or -1 for unlimited reads")
```
ian-shim, line 35: We could return error instead of crashing when validation fails

ian-shim (approved):
lgtm! Can we address [this comment](https://github.com/Layr-Labs/eigenda/pull/677/files#r1701018662) from last review?

tools/traffic/table/blob_table.go
cody-littley: This code was replaced with the simpler implementation.

tools/traffic/table/blob_metadata.go
cody-littley: change made

tools/traffic/table/blob_metadata.go
```diff
@@ -0,0 +1,69 @@
...
+
+import "errors"
+
+// BlobMetadata encapsulates various information about a blob written by the traffic generator.
+type BlobMetadata struct {
+	// key of the blob, set when the blob is initially uploaded.
+	key []byte
+
+	// checksum of the blob.
+	checksum *[16]byte
```
jianoaix, line 11: Does this need to be a pointer?

tools/traffic/table/blob_metadata.go
```diff
@@ -0,0 +1,69 @@
...
+
+	// size of the blob, in bytes.
+	size uint
+
+	// blobIndex of the blob.
+	blobIndex uint
```
jianoaix, line 17: What is blob index?

tools/traffic/table/blob_metadata.go
```diff
@@ -0,0 +1,69 @@
+package table
+
+import "errors"
+
+// BlobMetadata encapsulates various information about a blob written by the traffic generator.
+type BlobMetadata struct {
+	// key of the blob, set when the blob is initially uploaded.
+	key []byte
```
jianoaix, line 8: Is this the blobKey: https://github.com/Layr-Labs/eigenda/blob/2566c2135b5174c23560fcbe425b1aec01b8df73/disperser/disperser.go#L52 ?

tools/traffic/table/blob_metadata.go
```diff
@@ -0,0 +1,69 @@
...
+	return &BlobMetadata{
+		key:                  key,
+		checksum:             checksum,
+		size:                 size,
+		blobIndex:            blobIndex,
+		remainingReadPermits: readPermits,
+	}, nil
+}
+
+// Key returns the key of the blob.
```
jianoaix, line 46: What about making the member variables public? These getters are quite simple so seem not worth the verbosity.

tools/traffic/table/blob_table.go
cody-littley: Refactored to use a map based implementation.

tools/traffic/table/blob_metadata.go
cody-littley: Intention was to make it possible to read these values without the capability of updating them. But maybe that's more of a java design pattern. I've made the member variables public as you suggest.

tools/traffic/table/blob_metadata.go
cody-littley: The key, in this context, is the `[]byte` return value of this method:

```
func (c *disperserClient) DisperseBlob(ctx context.Context, data []byte, quorums []uint8) (*disperser.BlobStatus, []byte, error) {
```

The `BlobKey` seems like it is holding the same data, although it is in string form, not byte array form. I originally had the `MetadataHash` field as well, although that is currently removed in this PR.

Should I be using the `BlobKey` struct here?

tools/traffic/table/blob_metadata.go
jianoaix: Yea, this index is the the position of the blob in the batch (the batch is a list of blobs)

tools/traffic/table/blob_metadata.go
jianoaix: Yea it feels quite Java-ish

jianoaix (approved):

PR #685, opened 2024-08-06
Split blob writer code out of larger PR.

cody-littley:

This is code broken off from the larger PR https://github.com/Layr-Labs/eigenda/pull/666

This branch contains the "BlobWriter", which manages the writing of blobs for the traffic generator.

```
	┌------------┐                                       ┌------------┐
	|   writer   |-┐             ┌------------┐          |   reader   |-┐
	└------------┘ |-┐  -------> |  verifier  | -------> └------------┘ |-┐
	  └------------┘ |           └------------┘            └------------┘ |
	    └------------┘                                       └------------┘
```

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests


tools/traffic/test/blob_writer_test.go
```diff
@@ -0,0 +1,143 @@
...
+		// Wait until the writer finishes its work.
+		tu.AssertEventuallyTrue(t, func() bool {
+			lock.Lock()
+			defer lock.Unlock()
+			return int(disperserClient.DisperseCount) > i && int(unconfirmedKeyHandler.Count)+errorCount > i
+		}, time.Second)
+
+		// These methods should be called exactly once per tick if there are no errors.
+		// In the presence of errors, nothing should be passed to the unconfirmed key handler.
+		assert.Equal(t, uint(i+1), disperserClient.DisperseCount)
```
ian-shim, line 101: You can also directly match if `disperserClient.DisperseBlob` method has been called by `disperserClient.AssertCalled(...)` or `disperserClient.AssertNumberOfCalls(...)`

ian-shim, line 101: Timing/synchronizing with another goroutine might be tricky just using time. Does this reliably pass? How do you know if L101 will execute after disperser request has been made? 
Maybe one thing we could do is make `run` method public and run test against this method? That way we can assert expectations after the method returns.

tools/traffic/test/mock_disperser.go
```diff
@@ -0,0 +1,96 @@
+package test
```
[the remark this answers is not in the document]
cody-littley: Neat trick! I actually noticed a few similar lines in other files and was confused by them.

tools/traffic/test/blob_writer_test.go
cody-littley: Empirically, this was reliably passing (ran it in a loop a few thousand times). Primary reason why I was going through all the trouble of doing it in an attempt to put the tests in another directory without making methods public that shouldn't be public.

I understand your general concern though. I refactored this so that I don't rely on timing. As a result, the test files will end up being in the same package as the other files. Ugly to me, but it seems like a fairly normal practice in go.

tools/traffic/test/blob_writer_test.go
```diff
@@ -0,0 +1,143 @@
...
+		generatorMetrics)
+	writer.Start()
+
+	errorProbability := 0.1
+	errorCount := 0
+
+	var previousData []byte
+
+	for i := 0; i < 100; i++ {
+		if rand.Float64() < errorProbability {
```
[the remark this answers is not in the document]
cody-littley: I changed this to non-random, it now generates an error every tenth cycle: `if i%10 == 0`. Would you still like to se a stand alone unit test for the error case, or is this sufficient?

tools/traffic/test/blob_writer_test.go
cody-littley: I've switched over to your recommended style for mock objects.

tools/traffic/workers/blob_writer.go
```diff
@@ -0,0 +1,164 @@
...
+	logger logging.Logger
+
+	// Config contains the configuration for the generator.
+	config *config2.WorkerConfig
+
+	// disperser is the client used to send blobs to the disperser.
+	disperser clients.DisperserClient
+
+	// Unconfirmed keys are sent here.
+	unconfirmedKeyHandler KeyHandler
```
ian-shim, line 35: have you considered passing unconfirmed keys to a channel instead and letting the caller handle it? 
I think that's a common pattern in Go

tools/traffic/workers/blob_writer.go
```diff
@@ -0,0 +1,164 @@
...
+// getRandomData returns a slice of random data to be used for a blob.
+func (writer *BlobWriter) getRandomData() *[]byte {
+	if *writer.fixedRandomData != nil {
+		return writer.fixedRandomData
+	}
+
+	data := make([]byte, writer.config.DataSize)
+	_, err := rand.Read(data)
+	if err != nil {
+		panic(fmt.Sprintf("unable to read random data: %s", err))
```
ian-shim, line 134: return an error instead of panicking?

tools/traffic/workers/blob_writer.go
```diff
@@ -0,0 +1,164 @@
...
+
+// writeNextBlob attempts to send a random blob to the disperser.
+func (writer *BlobWriter) writeNextBlob() {
+	data := writer.getRandomData()
+	key, err := metrics.InvokeAndReportLatency(writer.writeLatencyMetric, func() ([]byte, error) {
+		return writer.sendRequest(*data)
+	})
+	if err != nil {
+		writer.writeFailureMetric.Increment()
+		writer.logger.Error("failed to send blob request", "err:", err)
```
ian-shim, line 115: nit: no `:` postfix on labels

tools/traffic/workers/blob_writer.go
```diff
@@ -0,0 +1,164 @@
...
+	if err != nil {
+		panic(fmt.Sprintf("unable to read random data: %s", err))
+	}
+	data = codec.ConvertByPaddingEmptyByte(data)
+
+	return &data
+}
+
+// sendRequest sends a blob to a disperser.
+func (writer *BlobWriter) sendRequest(data []byte) ([]byte /* key */, error) {
```
ian-shim, line 142: you could use a named return values if the returns values look ambiguous
`...sendRequest(data []byte) (key []byte, err error)`

ian-shim (approved):
lgtm, left few more comments for you to consider before merging

tools/traffic/workers/blob_writer.go
cody-littley: I'm willing to make this change, but since this modifies the API, would it be ok with you if I made this change in the next PR that will add the code that consumes the unconfirmed keys? It will cause some merge conflicts if I change this API in this PR.

tools/traffic/workers/blob_writer.go
cody-littley: done

tools/traffic/workers/blob_writer.go
cody-littley: Interesting, today I learned that syntax exists. :) Change made.

tools/traffic/workers/blob_writer.go
```diff
@@ -0,0 +1,162 @@
...
+		writeLatencyMetric:    generatorMetrics.NewLatencyMetric("write"),
+		writeSuccessMetric:    generatorMetrics.NewCountMetric("write_success"),
+		writeFailureMetric:    generatorMetrics.NewCountMetric("write_failure"),
+	}
+}
+
+// Start begins the blob writer goroutine.
+func (writer *BlobWriter) Start() {
+	writer.waitGroup.Add(1)
+	ticker := time.NewTicker(writer.config.WriteRequestInterval)
```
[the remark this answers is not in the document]
cody-littley: Yes. This is the period, aka the time in-between writes. If you set a period of 5 seconds, on average it will write 1/5 or 0.2 blobs per second.

tools/traffic/workers/blob_writer.go
```diff
@@ -0,0 +1,162 @@
+package workers
+
+import (
+	"context"
+	"crypto/md5"
+	"crypto/rand"
+	"fmt"
+	"github.com/Layr-Labs/eigenda/api/clients"
+	"github.com/Layr-Labs/eigenda/encoding/utils/codec"
+	config2 "github.com/Layr-Labs/eigenda/tools/traffic/config"
```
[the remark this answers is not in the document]
cody-littley: I blame my IDE, it automatically aliased this when I did an automatic refactor. 🙃 

Fixed.

tools/traffic/workers/key_handler.go
```diff
@@ -0,0 +1,7 @@
+package workers
+
+// KeyHandler is an interface describing an object that can accept unconfirmed keys.
+type KeyHandler interface {
```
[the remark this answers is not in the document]
cody-littley: This will go away in the next PR when I implement a request from @ian-shim to use a channel to pass this data. I'm intentionally not changing it in this PR to reduce the number of merge conflicts in my feature branch.

PR #686, opened 2024-08-06
increase node client grpc buffer to 32MiB

bxue-l2:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/node_client.go
+++ b/api/clients/node_client.go
@@ -89,6 +89,7 @@ func (c client) GetChunks(
 	conn, err := grpc.Dial(
 		core.OperatorSocket(opInfo.Socket).GetRetrievalSocket(),
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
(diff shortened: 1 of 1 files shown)


api/clients/node_client.go
```diff
@@ -89,6 +89,7 @@ func (c client) GetChunks(
 	conn, err := grpc.Dial(
 		core.OperatorSocket(opInfo.Socket).GetRetrievalSocket(),
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
+		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(32*1024*1024)), // 32 MiB buffer for the worst case
```
jianoaix, line 92: After erasure coding, if the operator has more than 25% stake, it may exceed 32MiB for 32MiB blob

bxue-l2: yeah, but I don't intend the buffer to be too big, maybe we should just use a flag with some reasonable default number, what do you think

PR #690, opened 2024-08-08
[1/N][zero serialization] Create data structure for passing encode chunks

jianoaix:
Create a `ChunksData` data structure for the encoded chunks data. This will be used to pass chunks around Encoder/Batcher/Dispatcher and eventually used to create dispersal requests to operator nodes.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/data.go
+++ b/core/data.go
@@ -63,39 +63,39 @@ const (
 
 type ChunksData struct {
 	// Chunks is the encoded bytes of the chunks.
-	// Each chunk are of same size.
 	Chunks [][]byte
 	// Format describes how the bytes of the chunks are encoded.
 	Format ChunkEncodingFormat
+	// The number of symbols in each chunk.
+	// Note each chunk of the same blob will always have the same number of symbols.
+	ChunkLen int
 }
 
 func (cd *ChunksData) Size() uint64 {
 	if len(cd.Chunks) == 0 {
 		return 0
 	}
-	return uint64(len(cd.Chunks)) * uint64(len(cd.Chunks[0]))
-}
-
-func (cd *ChunksData) ChunkLength() int {
-	if len(cd.Chunks) == 0 {
-		return 0
+	// GnarkChunkEncoding will create chunks of equal size.
+	if cd.Format == GnarkChunkEncodingFormat {
+		return uint64(len(cd.Chunks)) * uint64(len(cd.Chunks[0]))
 	}
-	return len(cd.Chunks[0]) / encoding.BYTES_PER_SYMBOL
+	// GobChunkEncoding can create chunks of different sizes.
+	size := uint64(0)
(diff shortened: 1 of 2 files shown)


core/data.go
```diff
@@ -52,8 +53,111 @@ const (
...
+		if len(c) != len(cd.Chunks[0]) {
+			return nil, errors.New("all chunks must be of same size")
+		}
+		copy(buf, c)
+		buf = buf[len(c):]
+	}
+	return result, nil
+}
+
+func (cd *ChunksData) ToGobFormat() (*ChunksData, error) {
```
jianoaix, line 111: The `ToGobForamt` and `ToGnarkFormat` conversions are used to handle compatibility when we enable the Gnark at Encoder server: during the transition, some Encoder servers may return new format (Gnark) and some may still return the old (Gob), and the EncoderClient/Batcher will need to work correctly during transition.

core/data.go
```diff
@@ -52,8 +53,111 @@ const (
...
+	}
+	// GobChunkEncoding can create chunks of different sizes.
+	size := uint64(0)
+	for _, c := range cd.Chunks {
+		size += uint64(len(c))
+	}
+	return size
+}
+
+func (cd *ChunksData) FlattenToBundle() ([]byte, error) {
```
jianoaix, line 90: This is the part that will replace the serialization of chunks when dispersing them to nodes.

core/data.go
bxue-l2: We could just first drain all the encoder pods to 0. Then quickly spin up new pods. But it could results in some increased latency

core/data.go
```diff
@@ -52,8 +53,111 @@ const (
...
+}
+
+func (cd *ChunksData) ToGobFormat() (*ChunksData, error) {
+	if cd.Format == GobChunkEncodingFormat {
+		return cd, nil
+	}
+	if cd.Format != GnarkChunkEncodingFormat {
+		return nil, fmt.Errorf("unsupported chunk encoding format: %d", cd.Format)
+	}
+	gobChunks := make([][]byte, 0, len(cd.Chunks))
```
bxue-l2, line 118: So for a while, serialization will still be done in batcher?

bxue-l2: I guess the plan is to gob encode in the encoder.

bxue-l2 (approved):

core/data.go
jianoaix: Yes, the serialization will happen only once, which will be at Encoder. That'll in a separate PR.

core/data.go
jianoaix: Yep. This is really a choice between software complexity v.s. operation complexity with a short/temporary latency spike. I'm just to make operation simpler here with some added software complexity.

PR #692, opened 2024-08-09
[blobstore] Filter out expired blobs from query

ian-shim:
Updates `GetBlobMetadataByStatusXXX` methods so that it doesn't fetch any expired blobs (determined by filtering on `RequestedAt > Now - TTL`). 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/common/blobstore/blob_metadata_store.go
+++ b/disperser/common/blobstore/blob_metadata_store.go
@@ -19,6 +19,7 @@ import (
 const (
 	statusIndexName = "StatusIndex"
 	batchIndexName  = "BatchIndex"
+	expiryIndexName = "Status-Expiry-Index"
 )
 
 // BlobMetadataStore is a blob metadata storage backed by DynamoDB
@@ -121,7 +122,7 @@ func (s *BlobMetadataStore) GetBulkBlobMetadata(ctx context.Context, blobKeys []
 // Because this function scans the entire index, it should only be used for status with a limited number of items.
 // It should only be used to filter "Processing" status. To support other status, a streaming version should be implemented.
 func (s *BlobMetadataStore) GetBlobMetadataByStatus(ctx context.Context, status disperser.BlobStatus) ([]*disperser.BlobMetadata, error) {
-	items, err := s.dynamoDBClient.QueryIndex(ctx, s.tableName, statusIndexName, "BlobStatus = :status AND Expiry > :expiry", commondynamodb.ExpresseionValues{
+	items, err := s.dynamoDBClient.QueryIndex(ctx, s.tableName, expiryIndexName, "BlobStatus = :status AND Expiry > :expiry", commondynamodb.ExpresseionValues{
 		":status": &types.AttributeValueMemberN{
 			Value: strconv.Itoa(int(status)),
 		},
@@ -143,11 +144,11 @@ func (s *BlobMetadataStore) GetBlobMetadataByStatus(ctx context.Context, status 
 	return metadata, nil
 }
 
-// GetBlobMetadataByStatusCount returns the count of all the metadata with the given status
+// GetBlobMetadataCountByStatus returns the count of all the metadata with the given status
 // Because this function scans the entire index, it should only be used for status with a limited number of items.
 // It should only be used to filter "Processing" status. To support other status, a streaming version should be implemented.
-func (s *BlobMetadataStore) GetBlobMetadataByStatusCount(ctx context.Context, status disperser.BlobStatus) (int32, error) {
-	count, err := s.dynamoDBClient.QueryIndexCount(ctx, s.tableName, statusIndexName, "BlobStatus = :status AND Expiry > :expiry", commondynamodb.ExpresseionValues{
+func (s *BlobMetadataStore) GetBlobMetadataCountByStatus(ctx context.Context, status disperser.BlobStatus) (int32, error) {
+	count, err := s.dynamoDBClient.QueryIndexCount(ctx, s.tableName, expiryIndexName, "BlobStatus = :status AND Expiry > :expiry", commondynamodb.ExpresseionValues{
 		":status": &types.AttributeValueMemberN{
 			Value: strconv.Itoa(int(status)),
 		},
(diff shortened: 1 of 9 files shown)


pschork (approved):

ian-shim:
This works as long as we don't change the TTL, but this would break if TTL changes - which is defined onchain. 
I think the better solution is to either create an index based on `Expiry` attribute or delete the expired blobs. Any thoughts @dmanc ?

dmanc:
> This works as long as we don't change the TTL, but this would break if TTL changes - which is defined onchain. I think the better solution is to either create an index based on `Expiry` attribute or delete the expired blobs. Any thoughts @dmanc ?

Why would we change the TTL? Also I don't like anything to do with assuming the metadata store doesn't have expired blobs. Blobs usually aren't deleted immediately when they get expired on DynamoDB for example, it goes through them at some rate.

["DynamoDB automatically deletes expired items within a few days of their expiration time, without consuming write throughput."](https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/TTL.html)

ian-shim:
> Why would we change the TTL?

Unclear for now, but we may want to extend the TTL in the future. It's an assumption we have to keep in mind, so I'd rather handle it the right way.

> Also I don't like anything to do with assuming the metadata store doesn't have expired blobs.

I think thats fair. Should we create an index on `Expiry` then?

dmanc:
> > Why would we change the TTL?
> 
> Unclear for now, but we may want to extend the TTL in the future. It's an assumption we have to keep in mind, so I'd rather handle it the right way.
> 
> > Also I don't like anything to do with assuming the metadata store doesn't have expired blobs.
> 
> I think thats fair. Should we create an index on `Expiry` then?

Yeah let's create an index

> > Why would we change the TTL?
> 
> Unclear for now, but we may want to extend the TTL in the future. It's an assumption we have to keep in mind, so I'd rather handle it the right way.
> 
> > Also I don't like anything to do with assuming the metadata store doesn't have expired blobs.
> 
> I think thats fair. Should we create an index on `Expiry` then?

Yeah I can add it tomorrow.

dmanc:
Deployed the index, should be unblocked now

disperser/common/blobstore/blob_metadata_store.go
```diff
@@ -117,10 +124,14 @@ func (s *BlobMetadataStore) GetBlobMetadataByStatus(ctx context.Context, status
 // Because this function scans the entire index, it should only be used for status with a limited number of items.
 // It should only be used to filter "Processing" status. To support other status, a streaming version should be implemented.
 func (s *BlobMetadataStore) GetBlobMetadataByStatusCount(ctx context.Context, status disperser.BlobStatus) (int32, error) {
```
jianoaix, line 126: This function name is a bit odd. It should probably be `GetBlobCountByStatus` or `GetBlobMetadatCountByStatus`.

disperser/common/blobstore/blob_metadata_store.go
```diff
@@ -94,9 +98,12 @@ func (s *BlobMetadataStore) GetBlobMetadata(ctx context.Context, metadataKey dis
 // Because this function scans the entire index, it should only be used for status with a limited number of items.
 // It should only be used to filter "Processing" status. To support other status, a streaming version should be implemented.
 func (s *BlobMetadataStore) GetBlobMetadataByStatus(ctx context.Context, status disperser.BlobStatus) ([]*disperser.BlobMetadata, error) {
-	items, err := s.dynamoDBClient.QueryIndex(ctx, s.tableName, statusIndexName, "BlobStatus = :status", commondynamodb.ExpresseionValues{
+	items, err := s.dynamoDBClient.QueryIndex(ctx, s.tableName, statusIndexName, "BlobStatus = :status AND RequestedAt > :requestedAt", commondynamodb.ExpresseionValues{
 		":status": &types.AttributeValueMemberN{
 			Value: strconv.Itoa(int(status)),
+		},
+		":requestedAt": &types.AttributeValueMemberN{
```
jianoaix, line 105: `liveBlobBoundary` or something? More readable than two `requestedAt`s

jianoaix (commented):
There seems an issue where Dispersing blobs that are expired will remain stuck in Dispersing after this change

disperser/common/blobstore/blob_metadata_store.go
ian-shim: Yeah updated to `GetBlobMetadatCountByStatus`

ian-shim:
> There seems an issue where Dispersing blobs that are expired will remain stuck in Dispersing after this change

That's true. Is it a concern?

jianoaix:
> > There seems an issue where Dispersing blobs that are expired will remain stuck in Dispersing after this change
> 
> That's true. Is it a concern?

Not sure if there is unexpected access (and use) of these "zombie" blobs before they get removed, if so that can cause issue

ian-shim:
> > > There seems an issue where Dispersing blobs that are expired will remain stuck in Dispersing after this change
> > 
> > 
> > That's true. Is it a concern?
> 
> Not sure if there is unexpected access (and use) of these "zombie" blobs before they get removed, if so that can cause issue

Added a check on `Expiry` and moves to `Failed` [if it has passed](https://github.com/Layr-Labs/eigenda/pull/692/files#diff-b3512a4819965cc0824e31e98c1add01712302a4740156af5b0b487106b59ab9R160). Wdyt?

jianoaix:
> > > > There seems an issue where Dispersing blobs that are expired will remain stuck in Dispersing after this change
> > > 
> > > 
> > > That's true. Is it a concern?
> > 
> > 
> > Not sure if there is unexpected access (and use) of these "zombie" blobs before they get removed, if so that can cause issue
> 
> Added a check on `Expiry` and moves to `Failed` [if it has passed](https://github.com/Layr-Labs/eigenda/pull/692/files#diff-b3512a4819965cc0824e31e98c1add01712302a4740156af5b0b487106b59ab9R160). Wdyt?

Sg, thanks

jianoaix (approved):

PR #698, opened 2024-08-13
[node] Make `GetBlobHeader` compatible with minibatches

ian-shim:
`GetBlobHeader` attempts to find blob header using the existing schema. If it doesn't find one, it uses the new schema (`blobIndexKey` to fetch the blob header hash and then `blobHeaderKey` to fetch the blob header).

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/node/store.go
+++ b/node/store.go
@@ -654,13 +654,27 @@ func (s *Store) GetBlobHeader(ctx context.Context, batchHeaderHash [32]byte, blo
 		return nil, err
 	}
 	data, err := s.db.Get(blobHeaderKey)
+	if err == nil {
+		return data, nil
+	}
+
+	if !errors.Is(err, leveldb.ErrNotFound) {
+		return nil, err
+	}
+
+	// error is leveldb.ErrNotFound
+	// try to get blob header by blobIndexPrefix
+	blobIndexKey := EncodeBlobIndexKey(batchHeaderHash, blobIndex)
+	blobHeaderHashBytes, err := s.db.Get(blobIndexKey)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


node/store.go
```diff
@@ -654,13 +654,27 @@ func (s *Store) GetBlobHeader(ctx context.Context, batchHeaderHash [32]byte, blo
...
 	if err != nil {
 		if errors.Is(err, leveldb.ErrNotFound) {
 			return nil, ErrKeyNotFound
 		}
 		return nil, err
 	}
-	return data, nil
+	var blobHeaderHash [32]byte
+	copy(blobHeaderHash[:], blobHeaderHashBytes)
+	return s.GetBlobHeaderByHeaderHash(ctx, blobHeaderHash)
```
mooselumph, line 677: Where does this mapping get created? All I see is the first mapping: https://github.com/Layr-Labs/eigenda/blob/dbeaf12c3a59c1d5c3b49ad7a0d59d07576fe9ca/node/store.go#L493

ian-shim: Yes, blobHeaderHash -> blobHeader mapping is getting created there in `StoreBlobs` method. 
(batchHeader, blobIndex) -> blobHeaderHash mapping is getting created from `AttestBatch` method calling [StoreBatchBlobMapping](https://github.com/Layr-Labs/eigenda/blob/master/node/store.go#L569)

mooselumph: Wow, the thing I hyperlinked makes no sense, lol. I think I was looking at `StoreBatch`.  Ok, this makes sense.

PR #702, opened 2024-08-14
Keep chunk length less than blob length

mooselumph:

Updates chunk length proposal logic so that chunk lengths are never greater than blob lengths. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/assignment.go
+++ b/core/assignment.go
@@ -204,21 +204,19 @@ func (c *StdAssignmentCoordinator) ValidateChunkLength(state *OperatorState, blo
 		}
 		num := new(big.Int).Mul(big.NewInt(2*int64(blobLength*percentMultiplier)), minStake)
 		denom := new(big.Int).Mul(big.NewInt(int64(info.ConfirmationThreshold-info.AdversaryThreshold)), totalStake)
-		maxChunkLength := uint(roundUpDivideBig(num, denom).Uint64())
+		maxChunkLength := nextPowerOf2(uint(roundUpDivideBig(num, denom).Uint64()))
 
 		// Ensure that the max chunk length is not greater than the blob length
 		if maxChunkLength > blobLength {
 			maxChunkLength = blobLength
 		}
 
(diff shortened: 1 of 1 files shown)


core/assignment.go
```diff
@@ -234,6 +235,10 @@ func (c *StdAssignmentCoordinator) CalculateChunkLength(state *OperatorState, bl
 
 	for {
 
+		if chunkLength > blobLength {
```
ian-shim, line 238: would it make sense to place this check as a condition in `ValidateChunkLength`?

mooselumph: I don't think it's essential, but it probably makes sense. I'll add it.

core/assignment.go
```diff
@@ -230,10 +236,18 @@ func (c *StdAssignmentCoordinator) ValidateChunkLength(state *OperatorState, blo
 // too large for the constraint in ValidateChunkLength
 func (c *StdAssignmentCoordinator) CalculateChunkLength(state *OperatorState, blobLength, targetNumChunks uint, param *SecurityParam) (uint, error) {
 
+	if targetNumChunks != 0 {
```
jianoaix, line 239: Why target num chunks has to be zero? Actually it's not used below.

mooselumph: I think we should deprecate this feature:
- It's currently not used.
- It's insufficient for the intended purpose (DAS)

New design here: https://www.notion.so/eigen-labs/Data-Allocation-v3-a75080bd7d334365809bbb2cfcaae538

core/assignment.go
ian-shim: oh I was thinking if we moved this to `ValidateChunkLength`, we wouldn't even need to validate it here as its done in L256

mooselumph: Makes sense. Changed it.

jianoaix (approved):

PR #704, opened 2024-08-16
Adds sampling rate flag to limit shadow writes

pschork:

Allows us to control the rate of blobs that are written to the shadow table and thus processed by minibatcher.

This will allow us to control ramp up of minibatches and ensure that we do not overwhelm the (preprod) node network with forced 2x load (legacy batch + minibatch)

Sampling could also be useful when we deploy to testnet allowing us to perform a light smoke test without doubling traffic to testnet nodes.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


pschork:
Abandoned. RIP Minibatcher

PR #707, opened 2024-08-19
Simplify minibatch data model

ian-shim:
Two changes in minibatch data model:
- removed `Minibatch` record itself. This model was used for identifying all blobs when constructing a full batch. Instead of this model, we can use `BlobMinibatchMapping` for this purpose
- unified `DispersalRequest` and `DispersalResponse` into one model `Dispersal`. This ensures request and response are 1:1

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


pschork (approved):
Nice cleanup.

PR #710, opened 2024-08-19
DAS protobuffs

cody-littley:

Introduces protobuf definitions for DAS.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests


api/proto/disperser/disperser.proto
```diff
@@ -115,17 +127,39 @@ message BlobStatusReply {
 	BlobInfo info = 2;
 }
 
-// RetrieveBlobRequest contains parameters to retrieve the blob.
-message RetrieveBlobRequest {
+// GetBlobRequest contains parameters to retrieve the blob.
+message GetBlobRequest {
```
mooselumph, line 131: I think we should allow two lookup methods, as specified in this proposal: https://www.notion.so/eigen-labs/Unified-blob-addressal-in-EigenDA-ae426c610f9e4e61b2e46eec76f7ca4d?d=f09f902f4d0743a7abef61faae0e2045#d392986bccbc43f9b6f8bcf6ace29255

1. BlobAuthHeaderHash
2. BatchHeaderHash + BlobIndex

api/proto/disperser/disperser.proto
```diff
@@ -115,17 +127,39 @@ message BlobStatusReply {
...
 	bytes batch_header_hash = 1;
 	uint32 blob_index = 2;
 }
 
-// RetrieveBlobReply contains the retrieved blob data
-message RetrieveBlobReply {
+// GetBlobReply contains the retrieved blob data
+message GetBlobReply {
```
mooselumph, line 137: I wonder if we should update this interface to also return the header.

api/proto/disperser/disperser.proto
```diff
@@ -115,17 +127,39 @@ message BlobStatusReply {
...
 	bytes data = 1;
 }
 
+// TODO this is identical to the node.proto definition. Should we share this definition?
+// A request to retrieve a specific chunk.
+message GetChunkRequest {
+	// The hash of the blob's ReducedBatchHeader defined on-chain.
+	bytes batch_header_hash = 1;
```
mooselumph, line 145: As above, should we support both lookup schemes?

mooselumph: Would it make sense to support requesting multiple chunks at once?

api/proto/lightnode/lightnode.proto
```diff
@@ -0,0 +1,65 @@
...
+	// For use by a DA node for monitoring the availability of chunks through its constellation of agent light nodes.
+	rpc StreamAvailabilityStatus(StreamAvailabilityStatusRequest) returns (stream SampledChunk) {}
+}
+
+// TODO these message types are duplicated from node.proto, perhaps they should share a common definition
+
+// A request to retrieve a specific chunk.
+message GetChunkRequest {
+	// The hash of the blob's ReducedBatchHeader defined on-chain.
+	bytes batch_header_hash = 1;
```
mooselumph, line 19: same as above

api/proto/node/node.proto
```diff
@@ -27,12 +27,32 @@ service Dispersal {
...
+	// GetChunk retrieves a specific chunk for a blob custodied at the Node.
+	rpc GetChunk(GetChunkRequest) returns (GetChunkReply) {}
+	// StreamChunks requests a stream of chunks with a specific index.
+	rpc StreamChunks(stream StreamChunksRequest) returns (stream GetChunkReply) {}
 	// GetBlobHeader is similar to RetrieveChunks, this just returns the header of the blob.
 	rpc GetBlobHeader(GetBlobHeaderRequest) returns (GetBlobHeaderReply) {}
-	// Retrieve node info metadata
-	rpc NodeInfo(NodeInfoRequest) returns (NodeInfoReply) {}
+	// StreamBlobHeaders streams the blob headers to the client.
+	rpc StreamBlobHeaders(StreamBlobHeadersRequest) returns (stream GetBlobHeaderReply) {}
```
mooselumph, line 39: I think we can just include the blob headers in the `StreamChunks` stream.

api/proto/disperser/disperser.proto
cody-littley: change made

cody-littley: Are we planning on deprecating lookup via batch header hash? The answer to this question will determine the best way to add the secondary lookup strategy.

# If we will always support both lookup mechanisms

Make `GetBlobRequest` contain a `oneof` argument where the caller can either provide the auth header hash or the batch header hash + blob index.

# If we will eventually stop supporting lookup by batch header hash

Leave the old and deprecated "RetrieveChunkRequest" using the batch header hash, and the have the new "GetChunkRequest" use the auth header hash.

api/proto/disperser/disperser.proto
cody-littley: Exactly how this is implemented depends a lot on whether we will eventually stop supporting lookup by blob header hash. Will make this change once I know the answer to this question.

Supporting multiple requests at once makes a lot of sense to me. All also add that in when I make the change to support lookup by both mechanisms.

api/proto/node/node.proto
cody-littley: removed

api/proto/disperser/disperser.proto
mooselumph: Hmm, this is a good question which is difficult to immediately answer. I think there's a possibility that we would deprecate the old one, but probably not before we wanted to start also using the new one. Which option would you go with in that world?

api/proto/disperser/disperser.proto
mooselumph: It would be this: https://github.com/Layr-Labs/eigenda/blob/dbeaf12c3a59c1d5c3b49ad7a0d59d07576fe9ca/api/proto/disperser/disperser.proto#L173

But there's some standardization we should do here. I'd like to get started on this next sprint. https://www.notion.so/eigen-labs/Unified-blob-addressal-in-EigenDA-ae426c610f9e4e61b2e46eec76f7ca4d?d=f09f902f4d0743a7abef61faae0e2045#d392986bccbc43f9b6f8bcf6ace29255

cody-littley: updated to support lookup via both mechanisms

api/proto/lightnode/lightnode.proto
cody-littley: updated to support lookup via both mechanisms

api/proto/common/common.proto
```diff
@@ -3,8 +3,47 @@ package common;
...
+// Uniquely identifies a blob based on its auth header hash.
+message AuthHeaderBlobKey {
+  bytes auth_header_hash = 1;
+}
+
+// Uniquely identifies a blob based on its batch header hash and blob index.
+message BatchHeaderBlobKey {
+  bytes batch_header_hash = 1;
+  uint32 blob_index = 2;
+}
```
cody-littley, line 33: Is there any missing information here? Do we need the quorum ID?

api/proto/disperser/disperser.proto
```diff
@@ -32,6 +32,18 @@ service Disperser {
...
+	// These RPCs are a work in progress and are not yet fully functional. API may not be stable. //
+	////////////////////////////////////////////////////////////////////////////////////////////////
+
+	// Retrieves the requested blob from the disperser's backend.
+	// This RPC is is a work in progress with undefined behavior, use with caution.
+	rpc GetBlob(common.BlobKey) returns (common.BlobData) {}
+
+	// Retrieves the requested chunk from the Disperser's backend.
+	// This RPC is is a work in progress with undefined behavior, use with caution.
+	rpc GetChunk(common.ChunkKey) returns (common.ChunkData) {}
```
cody-littley, line 46: Will the addition of new RPCs cause problems if I don't actually register methods to handle them? Or do I need to register simple no-op methods for safety?

api/proto/node/node.proto
cody-littley: This is added back in per our most recent discussions.

api/proto/common/common.proto
```diff
@@ -3,8 +3,47 @@ package common;
...
 }
+
+/////////////////////////////////////////////////////////////////////////////////////////////////
+// Messages for WIP RPCs. These are subject to change, do not consider these to be stable API. //
+/////////////////////////////////////////////////////////////////////////////////////////////////
+
+// A unique identifier for a blob.
+message BlobKey {
+  oneof key {
+    AuthHeaderBlobKey auth_header_blob_key = 1;
```
mooselumph, line 19: I'm proposing renaming `BlobAutherHeader` to `BlobHeader` and `BLobHeader` to `BlobCertificate`. If we do this, we could call this field `BlobHeaderBlobKey`. 

https://www.notion.so/eigen-labs/Unified-blob-addressal-in-EigenDA-ae426c610f9e4e61b2e46eec76f7ca4d?pvs=4#bde89f03197d45b89188d541b8345547

api/proto/common/common.proto
mooselumph: These are sufficient to identify the blob. We probably want to include the quorum ID in the chunk key.

api/proto/common/common.proto
```diff
@@ -3,8 +3,47 @@ package common;
...
+}
+
+// Uniquely identifies a blob based on its batch header hash and blob index.
+message BatchHeaderBlobKey {
+  bytes batch_header_hash = 1;
+  uint32 blob_index = 2;
+}
+
+// A unique identifier for a chunk.
+message ChunkKey {
```
mooselumph, line 36: This should probably include the quorum ID.

api/proto/disperser/disperser.proto
mooselumph: I don't know the answer to this off-hand.

mooselumph: But my guess is that we would need to implement no-op methods in order for the server to even compile?

api/proto/node/node.proto
```diff
@@ -33,6 +33,24 @@ service Retrieval {
 	rpc GetBlobHeader(GetBlobHeaderRequest) returns (GetBlobHeaderReply) {}
 	// Retrieve node info metadata
 	rpc NodeInfo(NodeInfoRequest) returns (NodeInfoReply) {}
+
+	////////////////////////////////////////////////////////////////////////////////////////////////
+	// These RPCs are a work in progress and are not yet fully functional. API may not be stable. //
+	////////////////////////////////////////////////////////////////////////////////////////////////
+
+	// GetChunks retrieves all of the chunks for a blob held by the node. Eventually this will replace RetrieveChunks.
+	rpc GetChunks(common.BlobKey) returns (GetChunksReply) {}
```
mooselumph, line 42: There's some ambiguity here, since the quorum ID isn't specified in the `BlobKey`. Maybe we should have a `common.BlobKeyAndQuorumID` which wraps the `BlobKey`. 

`RetrieveChunks` can take `BlobKeyandQuorumID`;
`ChunkKey` can contain the `BlobKeyAndQuorumID` and the `ChunkIndex`

mooselumph: ^^Just one idea

api/proto/node/node.proto
```diff
@@ -33,6 +33,24 @@ service Retrieval {
...
+	// GetChunk retrieves a specific chunk for a blob custodied at the Node.
+	rpc GetChunk(common.ChunkKey) returns (common.ChunkData) {}
+	// StreamChunks requests a stream of chunks with a specific index.
+	rpc StreamChunks(stream StreamChunksRequest) returns (stream common.ChunkData) {}
```
mooselumph, line 46: We can drop this one.

mooselumph (commented):
A couple comments.

api/proto/common/common.proto
cody-littley: Added quorum ID to chunk key.

api/proto/disperser/disperser.proto
cody-littley: It compiles without them. In the interest of safety, probably I will write unit tests to verify behavior with these new RPCs.

api/proto/common/common.proto
cody-littley: added

api/proto/node/node.proto
cody-littley: I ended up just calling the argument `GetChunksRequest`, which contains a blob key and a quorum ID. Do you think we will end up reusing this type? If so, I can move it to common and give it a more general name.

api/proto/node/node.proto
cody-littley: removed

api/proto/common/common.proto
```diff
@@ -3,8 +3,48 @@ package common;
...
+  }
+}
+
+// Uniquely identifies a blob based on its blob header hash.
+message HeaderBlobKey {
+  bytes header_hash = 1;
+}
+
+// Uniquely identifies a blob based on its certificate hash and blob index.
+message CertificateBlobKey {
```
cody-littley, line 30: BatchBlobKey

api/proto/common/common.proto
```diff
@@ -3,8 +3,48 @@ package common;
...
+  bytes certificate_hash = 1;
+  uint32 blob_index = 2;
+}
+
+// A unique identifier for a chunk.
+message ChunkKey {
+  BlobKey blob_key = 1;
+  uint32 quorum_id = 2;
```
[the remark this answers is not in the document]
cody-littley: removed

api/proto/common/common.proto
```diff
@@ -3,8 +3,48 @@ package common;
...
 }
+
+/////////////////////////////////////////////////////////////////////////////////////////////////
+// Messages for WIP RPCs. These are subject to change, do not consider these to be stable API. //
+/////////////////////////////////////////////////////////////////////////////////////////////////
+
+// A unique identifier for a blob.
+message BlobKey {
+  oneof key {
+    HeaderBlobKey header_blob_key = 1;
```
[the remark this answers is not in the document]
cody-littley: I'm not sure what you mean here... isn't the current plan to allow lookup by either the blob's header hash (aka `HeaderBlobKey`) or the batch hash + index (aka `BatchBlobKey`)?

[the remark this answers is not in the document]
mooselumph: Oh, this comment was from before you made the most recent changes. It looks good now that you've switched from `CertificateBlobKey` to `BatchBlobKey`.

api/proto/common/common.proto
```diff
@@ -3,8 +3,47 @@ package common;
...
+}
+
+// Uniquely identifies a blob based on its blob header hash.
+message HeaderBlobKey {
+  bytes header_hash = 1;
+}
+
+// Uniquely identifies a blob based on its certificate hash and blob index.
+message BatchBlobKey {
+  bytes certificate_hash = 1;
```
mooselumph, line 31: nit: `batch_header_hash`

api/proto/node/node.proto
```diff
@@ -33,6 +33,23 @@ service Retrieval {
...
+	// GetChunks retrieves all of the chunks from a specified quorum for a blob held by the node.
+	// Eventually this will replace RetrieveChunks.
+	rpc GetChunks(GetChunksRequest) returns (GetChunksReply) {}
+	// GetChunk retrieves a specific chunk for a blob custodied at the Node.
+	rpc GetChunk(common.ChunkKey) returns (common.ChunkData) {}
+	// GetHeader is similar to RetrieveChunks, this just returns the header of the blob.
+	// Eventually this will replace GetBlobHeader.
+	rpc GetHeader(common.BlobKey) returns (GetBlobHeaderReply) {}
+	// StreamHeaders requests a stream all new headers.
+	rpc StreamHeaders(stream StreamHeadersRequest) returns (stream GetBlobHeaderReply) {}
```
mooselumph, line 50: We should discuss whether we also want to have a `StreamBlobKeys` endpoint.

mooselumph (commented):
Looks good to me. @jianoaix, can you please take a look?

api/proto/common/common.proto
cody-littley: fixed

api/proto/node/node.proto
cody-littley: Repeating result of discussion here. It was decided that having a `StreamBlobKeys` endpoint would probably be useful, but that it was not a necessary component for the initial MVP. We plan on circling back on this question after the MVP launches.

api/proto/node/node.proto
```diff
@@ -33,6 +33,23 @@ service Retrieval {
...
+	////////////////////////////////////////////////////////////////////////////////////////////////
+	// These RPCs are a work in progress and are not yet fully functional. API may not be stable. //
+	////////////////////////////////////////////////////////////////////////////////////////////////
+
+	// GetChunks retrieves all of the chunks from a specified quorum for a blob held by the node.
+	// Eventually this will replace RetrieveChunks.
+	rpc GetChunks(GetChunksRequest) returns (GetChunksReply) {}
+	// GetChunk retrieves a specific chunk for a blob custodied at the Node.
+	rpc GetChunk(common.ChunkKey) returns (common.ChunkData) {}
+	// GetHeader is similar to RetrieveChunks, this just returns the header of the blob.
```
jianoaix, line 46: You may not want to refer to `RetrieveChunks` in this new set of apis

api/proto/node/node.proto
```diff
@@ -216,3 +233,38 @@ message NodeInfoReply {
 	uint32 num_cpu = 4;
 	uint64 mem_bytes = 5;
 }
+
+/////////////////////////////////////////////////////////////////////////////////////////////////
+// Messages for WIP RPCs. These are subject to change, do not consider these to be stable API. //
+/////////////////////////////////////////////////////////////////////////////////////////////////
+
+// Request that all chunks from a particular quorum for a blob be sent.
```
jianoaix, line 241: "...all custodied chunks..." and drop "be sent"

api/proto/node/node.proto
```diff
@@ -216,3 +233,38 @@ message NodeInfoReply {
...
+message GetChunksRequest {
+	common.BlobKey blob_key = 1;
+	uint32 quorum_id = 2;
+}
+
+// Reply to GetChunksRequest.
+message GetChunksReply {
+	// The indices of the chunks sent. Each index corresponds to the chunk at the same position in the chunks field.
```
jianoaix, line 249: "indices of the chunks custodied"

api/proto/node/node.proto
```diff
@@ -33,6 +33,23 @@ service Retrieval {
...
+	rpc GetChunks(GetChunksRequest) returns (GetChunksReply) {}
+	// GetChunk retrieves a specific chunk for a blob custodied at the Node.
+	rpc GetChunk(common.ChunkKey) returns (common.ChunkData) {}
+	// GetHeader is similar to RetrieveChunks, this just returns the header of the blob.
+	// Eventually this will replace GetBlobHeader.
+	rpc GetHeader(common.BlobKey) returns (GetBlobHeaderReply) {}
+	// StreamHeaders requests a stream all new headers.
+	rpc StreamHeaders(stream StreamHeadersRequest) returns (stream GetBlobHeaderReply) {}
+	// Retrieve node info metadata. Eventually this will replace NodeInfo.
+	rpc GetNodeInfo(GetNodeInfoRequest) returns (GetNodeInfoReply) {}
```
jianoaix, line 52: Why do we need a new node info API?

api/proto/node/node.proto
jianoaix: Remove it for now if we plan to revisit/decide later

api/proto/common/common.proto
```diff
@@ -3,8 +3,47 @@ package common;
...
+}
+
+// Uniquely identifies a blob based on its blob header hash.
+message HeaderBlobKey {
+  bytes header_hash = 1;
+}
+
+// Uniquely identifies a blob based on its certificate hash and blob index.
+message BatchBlobKey {
+  bytes batch_header_hash = 1;
```
jianoaix, line 31: It'd be more readable if we make the wording consistent with comment: either it's batch_header_hash, or certificate_hash

api/proto/common/common.proto
```diff
@@ -3,8 +3,47 @@ package common;
...
+message BlobKey {
+  oneof key {
+    HeaderBlobKey header_blob_key = 1;
+    BatchBlobKey batch_blob_key = 2;
+  }
+}
+
+// Uniquely identifies a blob based on its blob header hash.
+message HeaderBlobKey {
+  bytes header_hash = 1;
```
jianoaix, line 26: If there is a length expected for this hash, comment it out

api/proto/common/common.proto
```diff
@@ -3,8 +3,47 @@ package common;
...
+  bytes y = 2;
 }
+
+/////////////////////////////////////////////////////////////////////////////////////////////////
+// Messages for WIP RPCs. These are subject to change, do not consider these to be stable API. //
+/////////////////////////////////////////////////////////////////////////////////////////////////
+
+// A unique identifier for a blob.
+message BlobKey {
+  oneof key {
```
jianoaix, line 18: Do we have to support both? It makes the golang code more verbose and less easy to work with (have to check which field is set each time)

api/proto/common/common.proto
```diff
@@ -3,8 +3,47 @@ package common;
...
+    HeaderBlobKey header_blob_key = 1;
```
jianoaix, line 19: It'd be nice to have a bit more intuitive naming for these two formats of keys

node/grpc/server.go
```diff
@@ -154,6 +157,26 @@ func (s *Server) NodeInfo(ctx context.Context, in *pb.NodeInfoRequest) (*pb.Node
...
+
+func (s *Server) GetChunk(context.Context, *commonpb.ChunkKey) (*commonpb.ChunkData, error) {
+	return nil, status.Errorf(codes.Unimplemented, "method GetChunk not implemented")
+}
+
+func (s *Server) GetHeader(context.Context, *commonpb.BlobKey) (*pb.GetBlobHeaderReply, error) {
+	return nil, status.Errorf(codes.Unimplemented, "method GetHeader not implemented")
+}
+
+func (s *Server) StreamHeaders(pb.Retrieval_StreamHeadersServer) error {
```
anupsv, line 172: I assume this means there could be support for long standing GRPC connections ?

api/proto/node/node.proto
cody-littley: @mooselumph's prior comment was about adding an additional RPC. The `StreamHeaders` RPC is one that we are confident we will want to keep.

api/proto/node/node.proto
cody-littley: I removed the reference to `RetrieveChunks`, but left the note that it will eventually replace `GetBlobHeader`.

api/proto/node/node.proto
cody-littley: change made

api/proto/node/node.proto
cody-littley: The main goal is to have better uniformity in the names of the RPCs. This does the same thing as the RPC `NodeInfo`, it just has a different name. Should we remove this new RPC for now?

api/proto/common/common.proto
cody-littley: Comment fixed. This was leftover cruft from a previous iteration where we were calling it something different.

api/proto/common/common.proto
cody-littley: @mooselumph what is the expected length of the header hash and the batch header hash?

api/proto/common/common.proto
cody-littley: This is probably a good discussion topic with @mooselumph, the plan for how to address a blob is something he's been working on.

api/proto/common/common.proto
cody-littley: How about `BatchHeaderHashBlobKey` and `HeaderHashBlobKey`? Or is there something else you think would fit better.

node/grpc/server.go
cody-littley: That was the idea. I've never set up a system with long lived GRPC connections, so I don't know if there are any pitfalls we need to worry about. Is this something you think we should reconsider from a design perspective?

api/proto/common/common.proto
```diff
@@ -3,8 +3,47 @@ package common;
 option go_package = "github.com/Layr-Labs/eigenda/api/grpc/common";
 
 message G1Commitment {
-	// The X coordinate of the KZG commitment. This is the raw byte representation of the field element.
```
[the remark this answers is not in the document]
cody-littley: Whitespace changed back to tabs.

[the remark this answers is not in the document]
cody-littley: I did a search of this directory tree and found a few places that used spaces instead of tabs. I also updated those places to use tabs.

api/proto/common/common.proto
```diff
@@ -8,3 +8,12 @@ message G1Commitment {
...
 	bytes y = 2;
 }
+
+/////////////////////////////////////////////////////////////////////////////////////////////////
+// Messages for WIP RPCs. These are subject to change, do not consider these to be stable API. //
+/////////////////////////////////////////////////////////////////////////////////////////////////
+
+// A chunk of a blob.
+message ChunkData {
+	bytes data = 1;
```
jianoaix, line 18: Users need a way to decode the bytes. We may settle on Gnark, so maybe it's fine to just have bytes and implicitly assume it's Gnark.

api/proto/disperser/disperser.proto
```diff
@@ -250,3 +257,21 @@ message BatchHeader {
...
 }
+
+/////////////////////////////////////////////////////////////////////////////////////////////////
+// Messages for WIP RPCs. These are subject to change, do not consider these to be stable API. //
+/////////////////////////////////////////////////////////////////////////////////////////////////
+
+// Request a specific chunk
+message GetChunkRequest {
+	// The hash of the blob's header.
+	bytes header_hash = 1;
```
jianoaix, line 268: blob_header_hash for clarity?

api/proto/lightnode/lightnode.proto
```diff
@@ -0,0 +1,39 @@
...
+	bytes header_hash = 1;
+}
+
+// A reply to a GetChunk request.
+message GetChunkReply {
+	// The chunk data.
+	common.ChunkData chunk = 1;
+}
+
+// A request from a DA node to an agent light node to stream the availability status of all chunks
```
jianoaix, line 30: Why is this API specific to agent light node? Should it applicable to all light nodes?

api/proto/node/node.proto
```diff
@@ -33,6 +33,15 @@ service Retrieval {
...
 	rpc NodeInfo(NodeInfoRequest) returns (NodeInfoReply) {}
+
+	////////////////////////////////////////////////////////////////////////////////////////////////
+	// These RPCs are a work in progress and are not yet fully functional. API may not be stable. //
+	////////////////////////////////////////////////////////////////////////////////////////////////
+
+	// GetChunk retrieves a specific chunk for a blob custodied at the Node.
+	rpc GetChunk(GetChunkRequest) returns (GetChunkReply) {}
+	// StreamHeaders requests a stream all new headers.
+	rpc StreamHeaders(stream StreamHeadersRequest) returns (stream StreamHeadersReply) {}
```
jianoaix, line 44: `StreamBlobHeaders`?

api/proto/lightnode/lightnode.proto
```diff
@@ -0,0 +1,39 @@
...
+// assigned to the light node.
+message StreamAvailabilityStatusRequest {
+	bytes authentication_token = 1;
+}
+
+// A reply to a StreamAvailabilityStatus request.
+message StreamAvailabilityStatusReply {
+	bytes header_hash = 1;
```
jianoaix, line 38: The comment of the API says "availability status of all chunks", but this is a blob header hash, how should users interpret it? (document it here)

api/proto/common/common.proto
cody-littley: My thinking here for just having bytes is that once we settle on a single encoding, we can document the format. If we end up deciding to support multiple encodings, we can trivially add another field to specify the encoding.

api/proto/disperser/disperser.proto
cody-littley: change made

api/proto/node/node.proto
cody-littley: change made

api/proto/lightnode/lightnode.proto
cody-littley: Updated doc as follows:

```
	// The hash of a blob header corresponding to a chunk the agent received and verified. From the light node's
	// perspective, the blob is available if the chunk is available.
	bytes header_hash = 1;
```

api/proto/lightnode/lightnode.proto
```diff
@@ -0,0 +1,41 @@
...
+// Experimental: the following definitions are experimental and subject to change. //
+/////////////////////////////////////////////////////////////////////////////////////
+
+service LightNode {
+	// GetChunk retrieves a specific chunk held by the light node.
+	rpc GetChunk(GetChunkRequest) returns (GetChunkReply) {}
+
+	// StreamAvailabilityStatus streams the availability status of all chunks assigned to the light node.
+	// For use by a DA node for monitoring the availability of chunks through its constellation of agent light nodes.
+	rpc StreamAvailabilityStatus(StreamAvailabilityStatusRequest) returns (stream StreamAvailabilityStatusReply) {}
```
jianoaix, line 16: Should it be `StreamAvailableChunks` or `StreamAvailableBlobs`?
The returned message is not a status (like available, not available, live, expired etc), it's just the blobs that are available at the light node.

api/proto/lightnode/lightnode.proto
jianoaix: Good to know this, it may be worth documenting the difference of behavior between agent vs. non-agent LNs.

api/proto/lightnode/lightnode.proto
mooselumph: This is a good callout. I think there's probably still room to rename these if needed. Cody, maybe you can make a judgement call and go with it for now.

mooselumph (approved):
Looks good to me!

api/proto/lightnode/lightnode.proto
cody-littley: I've changed this to `StreamChunkAvailability`. Even after this PR merges, not too late to change naming. Only becomes tricky once we have community members running light nodes of their own.

api/proto/relay/relay.proto
```diff
@@ -0,0 +1,30 @@
...
+import "google/protobuf/wrappers.proto";
+import "common/common.proto";
+option go_package = "github.com/Layr-Labs/eigenda/api/grpc/relay";
+
+/////////////////////////////////////////////////////////////////////////////////////
+// Experimental: the following definitions are experimental and subject to change. //
+/////////////////////////////////////////////////////////////////////////////////////
+
+service Relay {
+  // Note: the relay API is still being fleshed out. This definition is not yet complete.
```
jianoaix, line 12: nit: align the indent style with other .proto files

api/proto/lightnode/lightnode.proto
jianoaix: This comment is helpful. Note my point was actually related to the naming of the API method itself, i.e. the return value is "blobs (header hash) that are available at the light node", not "availability statuses of all blobs that the light node received".

jianoaix (commented):
Mostly LG.

api/proto/relay/relay.proto
cody-littley: done

api/proto/lightnode/lightnode.proto
cody-littley: Updated RPC name and comment as follows:

```
	// StreamBlobAvailability streams the availability status blobs from the light node's perspective.
	// A light node considers a blob to be available if all chunks it wants to sample are available.
	// This API is for use by a DA node for monitoring the availability of chunks through its 
	// constellation of agent light nodes.
	rpc StreamBlobAvailability(StreamChunkAvailabilityRequest) returns (stream StreamChunkAvailabilityReply) {}
```

PR #716, opened 2024-08-21
Perf observability: Encoder data throughput and blob size breakdown

jianoaix:
- Better understanding of the encoding throughput
- Better understanding of the latency by blob size buckets

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/encoder/server.go
+++ b/disperser/encoder/server.go
@@ -47,7 +47,7 @@ func (s *Server) EncodeBlob(ctx context.Context, req *pb.EncodeBlobRequest) (*pb
 	select {
 	case s.requestPool <- struct{}{}:
 	default:
-		s.metrics.IncrementRateLimitedBlobRequestNum()
+		s.metrics.IncrementRateLimitedBlobRequestNum(len(req.GetData()))
 		s.logger.Warn("rate limiting as request pool is full", "requestPoolSize", s.config.RequestPoolSize, "maxConcurrentRequests", s.config.MaxConcurrentRequests)
 		return nil, errors.New("too many requests")
 	}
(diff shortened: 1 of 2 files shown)


dmanc (approved):

disperser/encoder/server.go
```diff
@@ -126,7 +126,7 @@ func (s *Server) handleEncoding(ctx context.Context, req *pb.EncodeBlobRequest)
 	}
 
 	totalTime := time.Since(begin)
-	s.metrics.TakeLatency(encodingTime, totalTime)
+	s.metrics.TakeLatency(len(req.GetData()), encodingTime, totalTime)
```
bxue-l2, line 129: since coding time actually depends on the size of encoded blob, that is a function of input blob size and current stake distribution. Should we measure against the coded size.
i.e. ChunkLength*NumChunks @dmanc

dmanc: Why not add chunklength numchunks as labels too?

jianoaix: We are not trying to add all factors that affect latency to a metric.
Also when the operators are given, each blob has the same condition.

bxue-l2: Blobs has the same condition only in the same reference Blocknumber. It could change depending on the stake distribution. I guess the real question is what we want to get out of this metrics. My understanding is that we want to know the encoding speed. Now here is my concern, suppose for one stake distribution, min stake is 0.01, and in another distribution, the stake distribution is 0.001. Although both blob are the same size, the proof generation time would be different, because the encoded size are differnt. We might misinterpret as if there is performance degradation, whereas in reality, it is just some stake distribution change.

jianoaix: Ideally we want to track a variable from user perspective, like "I have a blob of size X, and it takes Y ms to process".
I'm fine to drop this label if blobSize is a poor variable to monitor.

jianoaix: Actually I should have kept it: we are not using blobSize to determine latency here. And this blobSize label will help detect the significant change in stake distribution.
E.g. the same blob size is getting a much larger encoding latency. We will be informed that there is fundamental change in the stake distribution.
Without this label, we cannot tell if it's because stake distribution change or something else, as it's mixing all blob sizes which can have very different latencies.

PR #720, opened 2024-08-22
Minibatch state recovery

ian-shim:
To be merged on top of https://github.com/Layr-Labs/eigenda/pull/717

`RecoverState` recovers minibatch/blob metadata state. It's called before the BatchConfirmer routine starts to ensure the state has been recovered in case of crashes/forced exits. 
It pulls batch state from minibatch/blob metadata store, and identifies minibatches that have been dispersed to all operators to form a full batch. If a minibatch is missing a response from any operators or the response contains an error, it's excluded from the full batch. Then, it sends `AttestBatch` requests and confirmed the batch onchain before it begins the regular routine. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/indexer/state.go
+++ b/core/indexer/state.go
@@ -76,6 +76,31 @@ func (ics *IndexedChainState) GetIndexedOperatorState(ctx context.Context, block
 	return state, nil
 }
 
+func (ics *IndexedChainState) GetIndexedOperators(ctx context.Context, blockNumber uint) (map[core.OperatorID]*core.IndexedOperatorInfo, error) {
+
+	pubkeys, sockets, err := ics.getObjects(blockNumber)
+	if err != nil {
+		return nil, err
+	}
+
+	ops := make(map[core.OperatorID]*core.IndexedOperatorInfo, len(pubkeys.Operators))
+	for id, op := range pubkeys.Operators {
+
+		socket, ok := sockets[id]
+		if !ok {
+			return nil, errors.New("socket for operator not found")
+		}
+
+		ops[id] = &core.IndexedOperatorInfo{
+			PubkeyG1: &core.G1Point{G1Affine: op.PubKeyG1},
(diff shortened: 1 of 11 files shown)


core/indexer/state.go
```diff
@@ -76,6 +76,31 @@ func (ics *IndexedChainState) GetIndexedOperatorState(ctx context.Context, block
...
+	if err != nil {
+		return nil, err
+	}
+
+	ops := make(map[core.OperatorID]*core.IndexedOperatorInfo, len(pubkeys.Operators))
+	for id, op := range pubkeys.Operators {
+
+		socket, ok := sockets[id]
+		if !ok {
+			return nil, errors.New("socket for operator not found")
```
pschork, line 91: Can we add opertator id to error

pschork (approved):

disperser/batcher/batch_confirmer.go
```diff
@@ -127,6 +132,95 @@ func (b *BatchConfirmer) Start(ctx context.Context) error {
 	return nil
 }
 
+func (b *BatchConfirmer) RecoverState(ctx context.Context) error {
```
mooselumph, line 135: What keeps this from being just a call to `HandleSingleBatch` that doesn't update the reference block number?

disperser/batcher/minibatch_utils.go
```diff
@@ -0,0 +1,61 @@
...
+			continue
+		}
+
+		if _, ok := received[d.MinibatchIndex]; !ok {
+			received[d.MinibatchIndex] = make(map[core.OperatorID]struct{})
+		}
+		received[d.MinibatchIndex][d.OperatorID] = struct{}{}
+	}
+
+	for minibatchIndex, opIDs := range received {
```
mooselumph, line 37: The logic here is way more aggressive than what's used in the `MinibatchStore.BatchDispersed` method. Why the difference?

PR #721, opened 2024-08-22
Lightnode kv store

cody-littley:

Creates an abstraction around level DB that allows it to be used by the light node without having too hard of a dependency on level DB.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests

### DIFF
--- a/common/kvstore/ttl/ttl_wrapper.go
+++ b/common/kvstore/ttl/ttl_wrapper.go
@@ -95,10 +95,7 @@ func buildExpiryKey(
 
 // parseExpiryKey extracts the expiry time and base key from the given expiry key.
 func parseExpiryKey(expiryKey []byte) (baseKey []byte, expiryTime time.Time) {
-	expiryUnixNano := int64(0)
-	for i := 0; i < 8; i++ {
-		expiryUnixNano |= int64(expiryKey[1+i]) << (56 - i*8)
-	}
+	expiryUnixNano := int64(binary.BigEndian.Uint64(expiryKey[1:]))
 	expiryTime = time.Unix(0, expiryUnixNano)
 
 	baseKey = expiryKey[9:]


kvstore/store.go
```diff
@@ -0,0 +1,41 @@
+package kvstore
+
+import (
+	"errors"
+	"github.com/syndtr/goleveldb/leveldb/iterator"
+)
+
+// Store implements a key-value store. May be backed by a database like LevelDB.
+type Store interface {
```
jianoaix, line 9: I'm not quite sure if this will be the right level to abstract. It may be a can of worms, e.g. if you want to specify the options in these APIs, then you need to keep adding more structs to hide different DBs.
It's possible the abstraction at application level will be more appropriate. These are very thin wrappers anyway.

lightnode/kvstore/store.go
```diff
@@ -0,0 +1,41 @@
+package kvstore
+
+import (
+	"errors"
+	"github.com/syndtr/goleveldb/leveldb/iterator"
+)
+
+// Store implements a key-value store. May be backed by a database like LevelDB.
+type Store interface {
```
jianoaix, line 9: Note this interface and its levelDB impl exist in DA Node already (https://github.com/Layr-Labs/eigenda/blob/master/node/db.go)

lightnode/kvstore/store.go
```diff
@@ -0,0 +1,41 @@
...
+	// Delete removes the key from the database. Does not return an error if the key does not exist.
+	Delete(key []byte) error
+
+	// DeleteBatch atomically removes a list of keys from the database.
+	DeleteBatch(keys [][]byte) error
+
+	// WriteBatch atomically writes a list of key / value pairs to the database. The key at index i in the keys slice
+	// corresponds to the value at index i in the values slice.
+	WriteBatch(keys, values [][]byte) error
+
```
jianoaix, line 26: What if the client wants to both read and write kvs atomically?

lightnode/kvstore/store.go
jianoaix, line 9: Comment the thread safety of these interfaces

lightnode/kvstore/storeutil/thread_safe_wrapper.go
```diff
@@ -0,0 +1,79 @@
...
+
+import (
+	"github.com/Layr-Labs/eigenda/lightnode/kvstore"
+	"github.com/syndtr/goleveldb/leveldb/iterator"
+	"sync"
+)
+
+var _ kvstore.Store = &threadSafeStore{}
+
+// threadSafeStore is a wrapper around a KVStore that makes store access thread safe.
```
jianoaix, line 11: The KV databases are usually thread safe already

lightnode/kvstore/store.go
```diff
@@ -0,0 +1,41 @@
...
+	// NewIterator returns an iterator that can be used to iterate over a subset of the keys in the database.
+	// Only keys with the given prefix will be iterated. The iterator must be closed by calling Release() when done.
+	// The iterator will return keys in lexicographically sorted order. The iterator walks over a consistent snapshot
+	// of the database, so it will not see any writes that occur after the iterator is created.
+	NewIterator(prefix []byte) (iterator.Iterator, error)
+
+	// Shutdown shuts down the store, flushing any remaining data to disk.
+	Shutdown() error
```
jianoaix, line 34: Not sure why Shutdown and Destroy are needed explicitly, or if you have intended use cases

lightnode/kvstore/storeutil/ttl_wrapper.go
```diff
@@ -0,0 +1,272 @@
...
+		logger: logger,
+	}
+	if gcPeriod > 0 {
+		ttlStore.expireKeysInBackground(gcPeriod)
+	}
+	return ttlStore
+}
+
+var keyPrefix = []byte("k")
+var expiryPrefix = []byte("e")
```
jianoaix, line 53: What if they collide with normal keys?

lightnode/kvstore/storeutil/ttl_wrapper.go
```diff
@@ -0,0 +1,272 @@
...
+	}
+	expiryTime = time.Unix(0, expiryUnixNano)
+
+	baseKey = expiryKey[9:]
+	return
+}
+
+// PutWithExpiration adds a key-value pair to the store that expires at a specified time.
+// Key is eventually deleted after the expiry time.
+func (store *ttlStore) PutWithExpiration(key []byte, value []byte, expiryTime time.Time) error {
```
jianoaix, line 100: The client may want to expire the entire batch that was written atomically, as that may often be a logically meaningful group of KVs.

lightnode/kvstore/store.go
cody-littley: Yes, I copied this interface. The reason why this exists in isolation is to avoid making any changes to existing code as a part of this PR. 

In the future, I think these interfaces should be merged. Do you think they should be merged now, or can that wait for a later time?

lightnode/kvstore/store.go
cody-littley: That's a legitimate feature for a general purpose database, but I don't think it will be necessary for the features we are planning for light nodes. I intentionally avoided adding features that won't be needed for light nodes.

lightnode/kvstore/storeutil/thread_safe_wrapper.go
cody-littley: Most off the shelf implementations probably are, but it's possible to imagine one that isn't. For example, the simple in-memory implementation I wrote for unit testing is not thread safe. We probably don't have any production use cases for the thread safe wrapper in the short term, do you think I should delete it?

lightnode/kvstore/store.go
cody-littley: Shutdown flushes the data out to disk and closes the database. Destroy goes a step further and deletes the on-disk data in addition to shutting things down. Destroy is mostly useful in test environments where you want to clean things up after a test.

lightnode/kvstore/storeutil/ttl_wrapper.go
cody-littley: Short answer, if used properly it's impossible to get key collisions.

If somebody is using the TTL wrapper, then it is unsafe to access the store directly. If somebody only accesses through the TTL wrapper, then every key gets a prefix, meaning it is impossible for there to be collisions. All "normal" keys are prefixed with "k", all expiry keys are prefixed with "e".

I'll add a comment pointing out that it's unsafe to access the store directly if you are also using the TTL wrapper.

cody-littley:
```
// Note: it is unsafe to access the wrapped store directly while the TTLStore is in use. The TTLStore uses special
// key formatting, and direct access to the wrapped store may violate the TTLStore's invariants, resulting in
// undefined behavior.
```

lightnode/kvstore/storeutil/ttl_wrapper.go
cody-littley: I agree that this is a useful general purpose feature. I didn't add something like this because light nodes won't need it. Do you think I should add this feature as a part of this PR?

jianoaix: If we want to unify different KV store wrappers, it may be needed. In the DA Node, it handles all blobs in a batch atomically (store and delete)

lightnode/kvstore/storeutil/thread_safe_wrapper.go
jianoaix: If there is no anticipation of production use, it may be cleaner just make the in-memory KV store itself thread safe. Having this threadSafeStore feels like we are anticipating some prod KV DB to be thread unsafe.

lightnode/kvstore/storeutil/ttl_wrapper.go
jianoaix: Sounds good to me to have the wrapper to handle the namespacing

lightnode/kvstore/store.go
cody-littley: Commend changed to the following:

```
// Implementations of this interface are expected to be thread-safe.
```

lightnode/kvstore/storeutil/thread_safe_wrapper.go
cody-littley: change made

lightnode/kvstore/store.go
cody-littley: > Shutdown: I'd think this store is a durable one, so there is no extra step required to persist when it stops the store?

Even if there is no work required to make data durable, there are still resources that need to be released when a store is shut down. The following is taken from the godocs on level DB:

```
// Close closes the DB. This will also releases any outstanding snapshot,
// abort any in-flight compaction and discard open transaction.
```

> Destroy: can the test create a different store if the purpose is to have DB from scratch?

Yes, this is mostly just syntactic sugar for automatically cleaning up all resources held by the DB (including disk resources). If you don't like this method I can delete it, but it will add some repeated code to unit tests.

kvstore/store.go
cody-littley: I've now moved it to `eigenda/common/kvstore` (from your request to unify the interfaces).

lightnode/kvstore/store.go
cody-littley: Change made. I didn't utilize the new implementation for the TTL feature, since in doing so there may be some very non-trivial changes needed. But I did make it so that they both use the same low level store interface and implementation.

lightnode/kvstore/storeutil/ttl_wrapper.go
cody-littley: I've implemented the requested `PutBatchWithExpiration()` method in anticipation of the future merge of the two implementations, but haven't actually merged the implementations as a part of this PR.

lightnode/kvstore/store.go
```diff
@@ -0,0 +1,45 @@
...
+
+	// Destroy shuts down and permanently deletes all data in the store.
+	Destroy() error
+}
```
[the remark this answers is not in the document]
cody-littley: For the specific case of migrating from one computer to another, the way I'd suggest doing that for light nodes would be to skip copying any data and just start from a fresh DB. The disruption from a single light node losing its state is small. The protocol works just fine as long as most light nodes don't lose state. Individual light nodes have a very small amount of responsibility.

If a light node operator did want to preserve state when migrating, the way to do that would be to call the `Shutdown()` method on the DB, then just copy the DB directory. I don't think there needs to be a special snapshot method.

With respect to adding features like this, my personal preference would be to wait until there is a specific use case in mind prior to going through the effort. Would it be ok with you if we waited to add this feature?

jianoaix (commented):
Looks better!

common/kvstore/leveldb/leveldb_store.go
```diff
@@ -0,0 +1,147 @@
...
+	batch := new(leveldb.Batch)
+	for _, key := range keys {
+		batch.Delete(key)
+	}
+	return store.db.Write(batch, nil)
+}
+
+// WriteBatch adds multiple key-value pairs to the store.
+func (store *levelDBStore) WriteBatch(keys [][]byte, values [][]byte) error {
+	if store.shutdown {
```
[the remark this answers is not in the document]
cody-littley: I've removed these checks.

common/kvstore/store.go
```diff
@@ -0,0 +1,43 @@
...
+// Store implements a key-value store. May be backed by a database like LevelDB.
+//
+// Implementations of this interface are expected to be thread-safe.
+type Store interface {
+
+	// Put stores the given key / value pair in the database, overwriting any existing value for that key.
+	Put(key []byte, value []byte) error
+
+	// Get retrieves the value for the given key. Returns a ErrNotFound error if the key does not exist.
+	Get(key []byte) ([]byte, error)
```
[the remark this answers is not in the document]
cody-littley: Added the following comment:

```
	// The value returned is safe to modify.
```

common/kvstore/ttl/ttl_wrapper.go
```diff
@@ -0,0 +1,305 @@
...
+func buildExpiryKey(
+	baseKey []byte,
+	expiryTime time.Time) []byte {
+
+	expiryKeyLength := 1 /* prefix */ + 8 /* expiry timestamp */ + len(baseKey)
+	expiryKey := make([]byte, expiryKeyLength)
+
+	expiryKey[0] = 'e'
+	expiryUnixNano := expiryTime.UnixNano()
+	for i := 0; i < 8; i++ {
```
jianoaix, line 88: I would just use `PutUint64` from binary library so no bits handling needed

common/kvstore/ttl/ttl_wrapper.go
```diff
@@ -0,0 +1,305 @@
...
+
+// Spawns a background goroutine that periodically checks for expired keys and deletes them.
+func (store *ttlStore) expireKeysInBackground(gcPeriod time.Duration) {
+	ticker := time.NewTicker(gcPeriod)
+	go func() {
+		for {
+			select {
+			case now := <-ticker.C:
+				err := store.expireKeys(now)
+				if err != nil {
```
jianoaix, line 156: Stopping the GC upon an error may be not desired, it may have to continue as an error is not fatal to GC

cody-littley: This will now log an error without shutting things down.

common/kvstore/ttl/ttl_wrapper.go
cody-littley: Didn't know that function existed. Good suggestion. Done.

common/kvstore/ttl/ttl_wrapper.go
```diff
@@ -0,0 +1,303 @@
...
+	copy(expiryKey[9:], baseKey)
+
+	return expiryKey
+}
+
+// parseExpiryKey extracts the expiry time and base key from the given expiry key.
+func parseExpiryKey(expiryKey []byte) (baseKey []byte, expiryTime time.Time) {
+	expiryUnixNano := int64(0)
+	for i := 0; i < 8; i++ {
+		expiryUnixNano |= int64(expiryKey[1+i]) << (56 - i*8)
```
jianoaix, line 100: The parsing needs to change corresponding to `PutUint64` as well.

Actually, it may be good to have test coverage for key building.

cody-littley: One step ahead of you on this one. I already have this test. ;) The byte format is the same between the previous and new implementation, so this old line of code continued passing the test.

I've switched this implementation to instead use `binary.BigEndian.Uint64()` and verified that the unit test continues to pass.

jianoaix (approved):
Nice work

PR #722, opened 2024-08-22
Perf observability: Add blob encoding latency at client side

jianoaix:
This adds blob encoding latency metric on the Batcher (i.e. client) side.

Currently we only have encoding latency on Encoder (i.e. server) side, which has the following problems:
- It doesn't track the TCP connection creation latency
- It doesn't track the request/response data transmission latency between Batcher and Encoder
- It doesn't track the queuing delay (in case of high traffic)

Having this Batcher-side latency metric addresses all of these problems, and provides a more accurate picture of encoding subsystem performance.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/batcher/encoding_streamer.go
+++ b/disperser/batcher/encoding_streamer.go
@@ -386,7 +386,7 @@ func (e *EncodingStreamer) RequestEncodingForBlob(ctx context.Context, metadata 
 					BlobMetadata:   metadata,
 					BlobQuorumInfo: res.BlobQuorumInfo,
 				}}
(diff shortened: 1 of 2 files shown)


dmanc (approved):
Nice, so is the idea that if we subtract server side latency with the client side latency the diff will be the transmission time?

jianoaix:
> Nice, so is the idea that if we subtract server side latency with the client side latency the diff will be the transmission time?

Yep. I think it anyway makes more sense to watch the overall latency from client's perspective, that's what will affect the E2E latency eventually.

PR #725, opened 2024-08-26
Perf observability: Add attestation and encoding throughput

jianoaix:
This will enable us to have visibility into the throughput at 3 key stages: ingestion (at disperser API server, metrics already exist), encoding, and attestation. Such visibility will enable clear picture of system throughput and bottleneck.

This also adds the throughput breakdown by quorum to improve the multi-quorum understanding.

Tested:
On preprod, this is a healthy throughput where there is no bottleneck

Screenshot 2024-08-26 at 5 21 43 PM

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/batcher/batcher.go
+++ b/disperser/batcher/batcher.go
@@ -328,6 +328,9 @@ func (b *Batcher) updateConfirmationInfo(
 		requestTime := time.Unix(0, int64(metadata.RequestMetadata.RequestedAt))
 		b.Metrics.ObserveLatency("E2E", float64(time.Since(requestTime).Milliseconds()))
 		b.Metrics.ObserveBlobAge("confirmed", float64(time.Since(requestTime).Milliseconds()))
+		for _, quorumInfo := range batchData.blobHeaders[blobIndex].QuorumInfos {
+			b.Metrics.IncrementBlobSize("confirmed", quorumInfo.QuorumID, int(metadata.RequestMetadata.BlobSize))
+		}
 	}
 
 	return blobsToRetry, nil
--- a/disperser/batcher/batcher.go
+++ b/disperser/batcher/batcher.go
@@ -419,6 +419,16 @@ func (b *Batcher) observeBlobAge(stage string, batch *batch) {
 	}
 }
 
+func (b *Batcher) observeBlobAgeAndSize(stage string, batch *batch) {
+	for i, m := range batch.BlobMetadata {
+		requestTime := time.Unix(0, int64(m.RequestMetadata.RequestedAt))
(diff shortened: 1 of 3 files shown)


disperser/batcher/batcher.go
```diff
@@ -328,6 +328,9 @@ func (b *Batcher) updateConfirmationInfo(
 		requestTime := time.Unix(0, int64(metadata.RequestMetadata.RequestedAt))
 		b.Metrics.ObserveLatency("E2E", float64(time.Since(requestTime).Milliseconds()))
 		b.Metrics.ObserveBlobAge("confirmed", float64(time.Since(requestTime).Milliseconds()))
+		for _, quorumInfo := range batchData.blobHeaders[blobIndex].QuorumInfos {
+			b.Metrics.IncrementBlobSize("confirmed", quorumInfo.QuorumID, int(metadata.RequestMetadata.BlobSize))
```
bxue-l2, line 332: wouldn't the blobSize be identical for eth and eigen?

disperser/batcher/batcher.go
```diff
@@ -419,6 +422,16 @@ func (b *Batcher) observeBlobAge(stage string, batch *batch) {
 	}
 }
 
+func (b *Batcher) observeBlobAgeAndSize(stage string, batch *batch) {
+	for i, m := range batch.BlobMetadata {
+		requestTime := time.Unix(0, int64(m.RequestMetadata.RequestedAt))
+		b.Metrics.ObserveBlobAge(stage, float64(time.Since(requestTime).Milliseconds()))
+		for _, quorumInfo := range batch.BlobHeaders[i].QuorumInfos {
+			b.Metrics.IncrementBlobSize(stage, quorumInfo.QuorumID, int(m.RequestMetadata.BlobSize))
```
bxue-l2, line 430: do you want to record the original or encoded size?

bxue-l2: https://github.com/Layr-Labs/eigenda/blob/5692cacc062da330439f8e3ae6d8be3046becf98/disperser/common/blobstore/shared_storage.go#L110

bxue-l2 (approved):

bxue-l2 (approved):

disperser/batcher/batcher.go
jianoaix: It's the original size

disperser/batcher/batcher.go
jianoaix: Yes, they will be always the same, since they are by default required. Only custom quorums will have a difference.

PR #726, opened 2024-08-27
Lightnode docker

cody-littley:

Create a docker image suitable for running the light node on a raspberry pi.

## Checks

- Testing Strategy
   - [x] hand tests
   - [ ] Unit tests
   - [ ] Integration tests


lightnode/docker/lnode-base.dockerfile
```diff
@@ -0,0 +1,38 @@
+# syntax=docker/dockerfile:1
+
+# The base libraries required to build and run the light node.
+
+FROM debian
```
pschork, line 5: This is gonna pull in the entire distro. Recommend using [distroless](https://github.com/GoogleContainerTools/distroless) see also [smallest-golang-docker-image](https://klotzandrew.com/blog/smallest-golang-docker-image/)

cody-littley: Thanks! I will look into this.

lightnode/docker/default-args.sh
```diff
@@ -0,0 +1,11 @@
+# Default arguments for building the docker image.
+# To override these values locally, create a file named `args.sh` in the same directory. 'args.sh' is ignored by git.
+
+# The location of the code to clone.
+export GIT_URL=https://github.com/Layr-Labs/eigenda.git
+
+# The name of the branch or the commit sha to clone.
+export BRANCH_OR_COMMIT=master
```
mooselumph, line 8: Would it make sense to use the commit that is currently checked out? Curious about the thought process here

lightnode/docker/lnode-base.dockerfile
cody-littley: I've updated this branch. It now builds from `golang:1.21.12-bookworm`, and then copies the resulting go binary onto the `scratch` image. Total size of the resulting docker file with the "hello world" placeholder is 3.4 MB.

lightnode/docker/default-args.sh
cody-littley: I'm open to that idea, but we should weigh pros and cons.

This build process clones the repo into the docker image, as opposed to using the copy on the host file system. This adds extra overhead. The reason I took this step was that I wanted to avoid uncommitted local changes accidentally making their way into the resulting docker file, as well as requiring an explicit choice when it comes to choosing which version to build. But there is certainly additional friction to the developer in doing it this way. Let's discuss.

anupsv:
In general, are we going to publish these docker builds as part of releases? If so, we should consider signing the containers and GHCR does enable this.

cody-littley:
@anupsv 

> In general, are we going to publish these docker builds as part of releases? If so, we should consider signing the containers and GHCR does enable this.

I'm guessing the answer is "yes", although I haven't spent much time yet thinking about releases. Would it be ok with you if we addressed this as a stand alone task? This docker image will be useful during testing as we do development work.

anupsv:
> @anupsv
> 
> > In general, are we going to publish these docker builds as part of releases? If so, we should consider signing the containers and GHCR does enable this.
> 
> I'm guessing the answer is "yes", although I haven't spent much time yet thinking about releases. Would it be ok with you if we addressed this as a stand alone task? This docker image will be useful during testing as we do development work.

Yep thats fine.

lightnode/docker/Dockerfile
```diff
@@ -0,0 +1,41 @@
...
+RUN rm .bashrc
+RUN rm .bash_logout
+RUN rm .profile
+
+# Copy the eigenda repository
+COPY --chown=lnode . /home/lnode/eigenda
+
+# Download all go dependencies and build the binary.
+WORKDIR /home/lnode/eigenda/lightnode
+RUN go mod download
```
mooselumph, line 21: I think a worthwhile optimization would be the following sequence:
1. copy go.mod
2. Run `go mod download`
3. Copy the remainder of source
4. Run build

This would prevent having to redownload the go modules every time the source changes (as long as modules don't change)

mooselumph (approved):

lightnode/docker/lnode-base.dockerfile
pschork: 🥳

pschork (approved):

lightnode/docker/Dockerfile
cody-littley: Very good suggestion. Change made.

PR #729, opened 2024-08-28
quorum threshold update

0x0aa0:

Updates EigenDA thresholds

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


ian-shim (approved):
Hmm it's missing the thresholds for EIGEN quorum. I guess we deployed the contract without checking in the code before?

0x0aa0:
> Hmm it's missing the thresholds for EIGEN quorum. I guess we deployed the contract without checking in the code before?

Never merged because it breaks the integration test but his is mainly here for a commit to reference the diff

PR #730, opened 2024-08-28
Json config

cody-littley:

Create a simple configuration framework for light nodes. Eventually we may use this in other products too.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests


configuration/parser.go
```diff
@@ -0,0 +1,102 @@
...
+	decoder := json.NewDecoder(reader)
+	decoder.DisallowUnknownFields()
+
+	err := decoder.Decode(t)
+	if err != nil {
+		// Add the line number and column number to the error message.
+		var syntaxErr *json.SyntaxError
+		errors.As(err, &syntaxErr)
+
+		if syntaxErr == nil {
```
mooselumph, line 72: I'm curious if this is the idiomatic way to use `errors.As`. Doesn't `errors.As` return false if the error does not confirm? Should we be checking that return value instead of checking for nil here?

cody-littley: I agree, better to check the return value. Didn't even realize there was a return value here, I was originally just reacting to a nil pointer exception. Updated.

configuration/README.md
jianoaix, line 1: It doesn't seem this need to be a top level directory, maybe `common` can fit

cody-littley: I've moved `eigenda/configuration` to `eigenda/common/configuration`.

mooselumph (approved):

common/configuration/README.md
```diff
@@ -0,0 +1,129 @@
...
+
+```go
+type BasicConfig struct {
+	Foo string
+	Bar int
+	Baz bool
+}
+```
+
+Put your configuration into a json file. For example, `config.json`:
```
jianoaix, line 33: Why do we want to put configs in json? What are the configs (the above mentioned it's not env/flags etc) and use cases

cody-littley: As to "what are the configs", in my mind this is just a general purpose configuration framework. Any input parameter to the light node software should be entered through the configuration framework. Some example configurations:

- what is your light node ID?
- what are the IP addresses of the DA nodes? (MVP will require manual configuration, eventually this will be automatic)
- what is the length of time a chunk should be stored before it is deleted?
- should I attempt to publish metrics to a prometheus server?

As to why I selected json as the input format, here are my reasons:

- There is a json parser built into the standard golang libraries, and it allows me to avoid adding another third party dependency. 
- The built in json parser requires very little boilerplate code (i.e. you just define a simple struct) and then call a single method to parse the json into the struct. Existing configuration patterns in our codebase require several hundred lines of custom go code in order to populate a small struct.
- json allows for complex data types, for example multidimensional arrays, maps, nested structs, etc. CLI flags and environment variables are good at encoding simple and flat configurations, but it gets nontrivial to support more complex ones.
- json is a well known language, and the learning curve will not be high for people to figure out how to use it

As a part of this framework, I intentionally do not support environment variables and CLI flags. My reasons are as follows:

- It adds a lot of boilerplate code and framework complexity.
- A design philosophy that I generally agree with is that it's usually better to make there one way to do things rather than to make multiple equivalent ways of doing the same thing. Having a uniform way of specifying configuration will hopefully make the management of a light node easier to do.
- Although sometimes convenient when first building something, I find environment to be not user friendly for beginners who are first setting up a system. It's really easy to not realize when an environment variable is set (or unset).

jianoaix: "Any input parameter" is a wide target area, all those CLI flags are parameters.

And the examples you listed are also mostly either CLI flags (enable metrics), or not a parameter for the light node (IP of the DA Node).

cody-littley: > "Any input parameter" is a wide target area

Yes, that was my intention. My goal is for all light node configuration that isn't compiled into the go executable to be provided via json configuration files. Even if flags like enabling metrics could function as CLI flags, I'm not convinced that the development overhead justifies the benefits. 

> or not a parameter for the light node (IP of the DA Node).

The initial MVP for light nodes will require them to know the IP addresses of all of the DA nodes. Since we don't have a good on-chain solution for this yet, the idea was that we'd have to provide a list via configuration files.

cody-littley: @jianoaix do you think it would be beneficial to discuss light node configuration strategies on a live call with a small group of people?

PR #732, opened 2024-08-29
Update internal data structures

mooselumph:

Internal renaming of data structures in support of planned interface updates and unified blob lookup scheme. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


core/serialization.go
```diff
@@ -38,10 +38,10 @@ func ComputeSignatoryRecordHash(referenceBlockNumber uint32, nonSignerKeys []*G1
 }
 
 // SetBatchRoot sets the BatchRoot field of the BatchHeader to the Merkle root of the blob headers in the batch (i.e. the root of the Merkle tree whose leaves are the blob headers)
-func (h *BatchHeader) SetBatchRoot(blobHeaders []*BlobHeader) (*merkletree.MerkleTree, error) {
+func (h *BatchHeader) SetBatchRoot(blobHeaders []*BlobCertificate) (*merkletree.MerkleTree, error) {
 	leafs := make([][]byte, len(blobHeaders))
 	for i, header := range blobHeaders {
```
ian-shim, line 43: s/blobHeaders/blobCertificates ?

api/clients/mock/node_client.go
```diff
@@ -19,7 +19,7 @@ func NewNodeClient() *MockNodeClient {
 	return &MockNodeClient{}
 }
 
-func (c *MockNodeClient) GetBlobHeader(ctx context.Context, socket string, batchHeaderHash [32]byte, blobIndex uint32) (*core.BlobHeader, *merkletree.Proof, error) {
+func (c *MockNodeClient) GetBlobHeader(ctx context.Context, socket string, batchHeaderHash [32]byte, blobIndex uint32) (*core.BlobCertificate, *merkletree.Proof, error) {
```
ian-shim, line 22: `GetBlobCertificate`?

api/clients/node_client.go
```diff
@@ -21,7 +21,7 @@ type RetrievedChunks struct {
 }
 
 type NodeClient interface {
-	GetBlobHeader(ctx context.Context, socket string, batchHeaderHash [32]byte, blobIndex uint32) (*core.BlobHeader, *merkletree.Proof, error)
+	GetBlobHeader(ctx context.Context, socket string, batchHeaderHash [32]byte, blobIndex uint32) (*core.BlobCertificate, *merkletree.Proof, error)
```
ian-shim, line 24: `GetBlobCertificate`?

api/clients/node_client.go
```diff
@@ -64,7 +64,7 @@ func (c client) GetBlobHeader(
 		return nil, nil, err
 	}
 
-	blobHeader, err := node.GetBlobHeaderFromProto(reply.GetBlobHeader())
+	blobHeader, err := node.GetBlobCertFromProto(reply.GetBlobHeader())
```
ian-shim, line 67: `certificate, err := ...`

core/assignment.go
```diff
@@ -162,7 +162,7 @@ func (c *StdAssignmentCoordinator) GetAssignments(state *OperatorState, blobLeng
 
 }
 
-func (c *StdAssignmentCoordinator) GetOperatorAssignment(state *OperatorState, header *BlobHeader, quorum QuorumID, id OperatorID) (Assignment, AssignmentInfo, error) {
+func (c *StdAssignmentCoordinator) GetOperatorAssignment(state *OperatorState, header *BlobCertificate, quorum QuorumID, id OperatorID) (Assignment, AssignmentInfo, error) {
```
ian-shim, line 165: `cert *BlobCertificate`?

ian-shim (commented):
lgtm, there are still references calling blob certificates as blob headers, which could be confusing. Do you think there's any issues with updating the references as part of this refactor?

core/data.go
```diff
@@ -220,11 +220,11 @@ type Blob struct {
...
 
-// BlobAuthHeader contains the data that a user must sign to authenticate a blob request.
+// BlobHeader contains the data that a user must sign to authenticate a blob request.
 // Signing the combination of the Nonce and the BlobCommitments prohibits the disperser from
 // using the signature to charge the user for a different blob or for dispersing the same blob
 // multiple times (Replay attack).
-type BlobAuthHeader struct {
+type BlobHeader struct {
 	// Commitments
 	encoding.BlobCommitments `json:"commitments"`
```
jianoaix, line 229: For the purpose of auth, is the kzg commitment itself sufficient (i.e. the 3 length related fields are not needed)?

core/serialization.go
```diff
@@ -184,7 +184,7 @@ func HashBatchHeader(batchHeader binding.IEigenDAServiceManagerBatchHeader) ([32
 }
 
 // GetBlobHeaderHash returns the hash of the BlobHeader that is used to sign the Blob
```
jianoaix, line 186: Stale comment

core/data.go
```diff
@@ -326,15 +324,15 @@ type EncodedBundles map[QuorumID]*ChunksData
 
 // BlobMessage is the message that is sent to DA nodes. It contains the blob header and the associated chunk bundles.
 type BlobMessage struct {
-	BlobHeader *BlobHeader
```
jianoaix, line 329: Do we need the concept of certificate internally? It looked nicer/more intuitive to have blob header + body (chunks/bundle).

core/data.go
mooselumph: I think the length is needed for payment purposes, but the length proof can probably be ommitted. I've made some updates to the document: https://www.notion.so/eigen-labs/Unified-blob-addressal-in-EigenDA-ae426c610f9e4e61b2e46eec76f7ca4d?pvs=4#ed6c05cfb2e34e3e9289a998cb25dc4d

core/data.go
mooselumph: I think it's pretty useful to maintain the distinction between the two since they have different purposes. Here it makes sense that it is the certificate because we are asking the DA node to sign the certificate. 

That being said, I'm open to other naming suggestions!

ian-shim:
Does the way we ended up building this into a breaking v2 release vs. just interface update in v1 change the approach in this PR?
For example, we don't need the concept of blob cert and nodes keep signing the batch header hash in v1.

mooselumph:
> Does the way we ended up building this into a breaking v2 release vs. just interface update in v1 change the approach in this PR? For example, we don't need the concept of blob cert and nodes keep signing the batch header hash in v1.

As we build V2, I think it will be useful to transform the internal data structures to the V2 format as much as possible. Do you have an alternative approach?

hopeyen (approved):
lgtm

ian-shim:
> > Does the way we ended up building this into a breaking v2 release vs. just interface update in v1 change the approach in this PR? For example, we don't need the concept of blob cert and nodes keep signing the batch header hash in v1.
> 
> As we build V2, I think it will be useful to transform the internal data structures to the V2 format as much as possible. Do you have an alternative approach?

I agree in principle, but the blob header or blob certificate in v1 (change proposed in this PR) don't match that of v2, right? How do we ensure that these two core structs interchangeable for both v1 & v2?

PR #736, opened 2024-08-31
V2 chunk assignment

mooselumph:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/assignment_v2.go
+++ b/core/assignment_v2.go
@@ -2,6 +2,7 @@ package core
 
 import (
 	"fmt"
+	"math"
 	"math/big"
 	"sort"
 )
@@ -10,6 +11,12 @@ type BlobParameters struct {
 	CodingRate              uint
 	ReconstructionThreshold float64
 	NumChunks               uint
+}
+
+func (p BlobParameters) MaxNumOperators() uint {
+
+	return uint(math.Floor(float64(p.NumChunks) * (1 - 1/(p.ReconstructionThreshold*float64(p.CodingRate)))))
+
 }
 
 var (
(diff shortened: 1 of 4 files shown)


core/assignment_v2.go
```diff
@@ -0,0 +1,159 @@
...
+	// GetChunkLength determines the length of a chunk given the blob version and blob length
+	GetChunkLength(blobVersion byte, blobLength uint) (uint, error)
+}
+
+type StdAssignmentCoordinatorV2 struct {
+}
+
+var _ AssignmentCoordinatorV2 = (*StdAssignmentCoordinatorV2)(nil)
+
+func (c *StdAssignmentCoordinatorV2) GetAssignments(state *OperatorState, blobVersion byte, quorum QuorumID) (map[OperatorID]Assignment, error) {
```
cody-littley, line 47: Why do we need this struct (as opposed to having methods that don't operate on a struct)? Is the idea that we may end up having multiple algorithms, and we can toggle between which algorithm we use by using a different struct that implements `AssignmentCoordinatorV2`?

core/assignment_v2.go
```diff
@@ -0,0 +1,159 @@
...
+		index += a.chunks
+	}
+
+	return assignments, nil
+
+}
+
+func (c *StdAssignmentCoordinatorV2) GetAssignment(state *OperatorState, blobVersion byte, quorum QuorumID, id OperatorID) (Assignment, error) {
+
+	assignments, err := c.GetAssignments(state, blobVersion, quorum)
```
cody-littley, line 124: I'm not sure on the planned call patterns for these methods, so it's possible what I'm about to suggest is premature optimization. Would it make sense to cache the value of `GetAssignments()` instead of recomputing it each time?

cody-littley (approved):

ian-shim:
@mooselumph Is this PR up to date?

cody-littley:
@mooselumph It's up to date, but we probably want to wait until we are closer to finalizing light node design before merging. I'm ok leaving the PR closed until then, will create a new one if the time comes.

PR #741, opened 2024-09-01
Add dispersing transistion check

bxue-l2:
This PR adds checks in shared_storage.go to ensure, the state of a blob never transit from confirmed to dispersing


disperser/common/blobstore/shared_storage.go
```diff
@@ -158,6 +160,17 @@ func (s *SharedBlobStore) MarkBlobConfirmed(ctx context.Context, existingMetadat
 }
 
 func (s *SharedBlobStore) MarkBlobDispersing(ctx context.Context, metadataKey disperser.BlobKey) error {
+	refreshedMetadata, err := s.GetBlobMetadata(ctx, metadataKey)
+	if err != nil {
+		s.logger.Error("error getting blob metadata while marking blobDispersing", "err", err)
+		return err
+	}
+	alreadyConfirmed, _ := refreshedMetadata.IsConfirmed()
```
jianoaix, line 168: What about just checking if the status is PROCESSING? That's the only valid predecessor for DISPERSING.

bxue-l2: updated

disperser/common/blobstore/shared_storage.go
```diff
@@ -158,6 +162,23 @@ func (s *SharedBlobStore) MarkBlobConfirmed(ctx context.Context, existingMetadat
...
+	if alreadyConfirmed {
```
jianoaix, line 171: This part seems not needed anymore

jianoaix (approved):

disperser/common/blobstore/shared_storage.go
```diff
@@ -158,6 +162,23 @@ func (s *SharedBlobStore) MarkBlobConfirmed(ctx context.Context, existingMetadat
 }
 
 func (s *SharedBlobStore) MarkBlobDispersing(ctx context.Context, metadataKey disperser.BlobKey) error {
+	refreshedMetadata, err := s.GetBlobMetadata(ctx, metadataKey)
```
jianoaix, line 165: Have you checked the conditional update which may save one RPC call: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/Expressions.ConditionExpressions.html#Expressions.ConditionExpressions.SimpleComparisons

mooselumph:
Looks good. Should we add a similar check for the `PROCESSING` state as well?

PR #744, opened 2024-09-04
add check when initializing processing state

bxue-l2:
This PR adds checks in StoreBlob, called by apiserver when first receiving a blob

This PR inserts a s.GetBlobMetadata before uploading the blob on S3 and setting blobMetaData. The expected behaviors is that such metadata should not ever exists. If it does, return err

### DIFF
--- a/disperser/common/blobstore/shared_storage.go
+++ b/disperser/common/blobstore/shared_storage.go
@@ -21,6 +21,8 @@ const (
 
 var errProcessingToDispersing = errors.New("blob transit to dispersing from non processing")
 
+var errProcessingInitialization = errors.New("blob status initialization")
+
 // The shared blob store that the disperser is operating on.
 // The metadata store is backed by DynamoDB and the blob store is backed by S3.
 //
@@ -89,6 +91,12 @@ func (s *SharedBlobStore) StoreBlob(ctx context.Context, blob *core.Blob, reques
 	}
 	metadataKey.BlobHash = blobHash
 	metadataKey.MetadataHash = metadataHash
+
+	refreshedMetadata, err := s.GetBlobMetadata(ctx, metadataKey)
+	if !errors.Is(err, disperser.ErrMetadataNotFound) {
+		s.logger.Error("error intentionally get blob status before creation", "err", err, "status", refreshedMetadata.BlobStatus, "metadataKey", refreshedMetadata.GetBlobKey().String())
+		return metadataKey, errProcessingInitialization
+	}
 
 	err = s.s3Client.UploadObject(ctx, s.bucketName, blobObjectKey(blobHash), blob.Data)
 	if err != nil {


disperser/common/blobstore/shared_storage.go
```diff
@@ -90,6 +92,12 @@ func (s *SharedBlobStore) StoreBlob(ctx context.Context, blob *core.Blob, reques
 	metadataKey.BlobHash = blobHash
 	metadataKey.MetadataHash = metadataHash
 
+	refreshedMetadata, err := s.GetBlobMetadata(ctx, metadataKey)
```
jianoaix, line 95: what if there is a 500 error from dynamo?

disperser/common/blobstore/shared_storage.go
```diff
@@ -21,6 +21,8 @@ const (
 
 var errProcessingToDispersing = errors.New("blob transit to dispersing from non processing")
 
+var errProcessingInitialization = errors.New("blob status initialization")
```
jianoaix, line 24: I'd call this not meeting the precondition

disperser/common/blobstore/shared_storage.go
```diff
@@ -90,6 +92,12 @@ func (s *SharedBlobStore) StoreBlob(ctx context.Context, blob *core.Blob, reques
 	metadataKey.BlobHash = blobHash
 	metadataKey.MetadataHash = metadataHash
 
+	refreshedMetadata, err := s.GetBlobMetadata(ctx, metadataKey)
+	if !errors.Is(err, disperser.ErrMetadataNotFound) {
+		s.logger.Error("error intentionally get blob status before creation", "err", err, "status", refreshedMetadata.BlobStatus, "metadataKey", refreshedMetadata.GetBlobKey().String())
+		return metadataKey, errProcessingInitialization
```
jianoaix, line 98: what is the gRPC error code that will be returned to end user from here?

disperser/common/blobstore/shared_storage.go
bxue-l2: I think I should just check if refreshedMetadata is nil or not

disperser/common/blobstore/shared_storage.go
bxue-l2: this would be this

api.NewInternalError("failed to store blob, please try again later")

disperser/common/blobstore/shared_storage.go
```diff
@@ -21,6 +21,8 @@ const (
 
 var errProcessingToDispersing = errors.New("blob transit to dispersing from non processing")
 
+var errProcessingMeetingPrecondition = errors.New("blob not meeting precondition to processing")
```
jianoaix, line 24: `errProcessingFailedPrecondition("blob metadata already exists....")`?

disperser/common/blobstore/shared_storage.go
```diff
@@ -90,6 +92,24 @@ func (s *SharedBlobStore) StoreBlob(ctx context.Context, blob *core.Blob, reques
 	metadataKey.BlobHash = blobHash
 	metadataKey.MetadataHash = metadataHash
 
+	refreshedMetadata, err := s.GetBlobMetadata(ctx, metadataKey)
+
+	// the only safe condition:
+	// err is disperser.ErrMetadataNotFound && refreshedMetadata is nil
```
jianoaix, line 98: It looks to me there are 2 failure cases:
- Failed to get metadata (due to errors other than Not Found): this is an INTERNAL error
- Got the metadata: this is FAILED_PRECONDITION error

disperser/common/blobstore/shared_storage.go
jianoaix: This seems good point: How would users store the same metadataKey twice? Even if sending the same blob, it'll result in different metadataKey, so there is no overwrite.

PR #745, opened 2024-09-04
feat: bls remote signer

shrimalmadhur:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


node/config.go
```diff
@@ -74,6 +78,11 @@ type Config struct {
 	UseSecureGrpc                  bool
 	ReachabilityPollIntervalSec    uint64
 	DisableNodeInfoResources       bool
+	BLSRemoteSignerUrl             string
+	BLSPublicKeyHex                string
+	BLSKeyPassword                 string
+	BLSSignerTLSCertFilePath       string
+	UseBLSRemoteSigner             bool
```
samlaf, line 85:
```suggestion

	BLSRemoteSignerEnabled      bool
	BLSRemoteSignerUrl             string
	BLSPublicKeyHex                string
	BLSKeyPassword                 string
	BLSSignerTLSCertFilePath       string
```

samlaf: Personally prefer to keep same naming convention for all vars, and to have the `enabled` var at the top.

samlaf: Maybe consider creating subconfigs also? The config for the node is crazy long.

node/config.go
```diff
@@ -143,14 +152,20 @@ func NewConfig(ctx *cli.Context) (*Config, error) {
 		ethClientConfig = geth.ReadEthClientConfig(ctx)
 	}
 
+	// check if BLS remote signer configuration is provided
+	useBLSRemoteSigner := ctx.GlobalString(flags.BLSRemoteSignerUrlFlag.Name) != "" && ctx.GlobalString(flags.BLSPublicKeyHexFlag.Name) != ""
```
samlaf, line 156: Personally prefer to have an explicit flag to turn a feature on/off

node/flags/flags.go
```diff
@@ -272,6 +272,27 @@ var (
...
+		Usage:    "The URL of the BLS remote signer",
+		Required: false,
+		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "BLS_REMOTE_SIGNER_URL"),
+	}
+
+	BLSPublicKeyHexFlag = cli.StringFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "bls-public-key-hex"),
+		Usage:    "The hex-encoded public key of the BLS signer",
+		Required: false,
+		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "BLS_PUBLIC_KEY_HEX"),
```
samlaf, line 287: does bls signer do auth? Or does it just store a bunch of private keys and if I have access to its endpoint then I can just ask it to sign from any private key just by giving it the respective public key?

node/grpc/server.go
```diff
@@ -283,7 +283,10 @@ func (s *Server) AttestBatch(ctx context.Context, in *pb.AttestBatchRequest) (*p
 	if err != nil {
 		return nil, fmt.Errorf("failed to get the batch header hash: %w", err)
 	}
-	sig := s.node.KeyPair.SignMessage(batchHeaderHash)
+	sig, err := s.node.GetSignature(ctx, batchHeaderHash)
```
samlaf, line 286: consider renaming `GetSignature` to just `Sign`?

node/node.go
```diff
@@ -384,12 +428,41 @@ func (n *Node) ProcessBatch(ctx context.Context, header *core.BatchHeader, blobs
...
 	n.Metrics.RecordStoreChunksStage("signed", batchSize, time.Since(stageTimer))
-	log.Debug("Sign batch succeeded", "pubkey", hexutil.Encode(n.KeyPair.GetPubKeyG2().Serialize()), "duration", time.Since(stageTimer))
+	log.Debug("Sign batch succeeded", "pubkey", n.Config.BLSPublicKeyHex, "duration", time.Since(stageTimer))
 
 	log.Debug("Exiting process batch", "duration", time.Since(start))
-	return sig, nil
+	return signature, nil
+}
+
+func (n *Node) GetSignature(ctx context.Context, data [32]byte) (*core.Signature, error) {
```
samlaf, line 443:
```suggestion
func (n *Node) SignMessage(ctx context.Context, data [32]byte) (*core.Signature, error) {
```

samlaf: rename to SignMessage to match the keypair.SignMessage function?

node/node.go
```diff
@@ -114,6 +116,47 @@ func NewNode(reg *prometheus.Registry, config *Config, pubIPProvider pubip.Provi
...
+		nodeLogger.Info("using local keystore private key for BLS signing")
+		// Generate BLS keys
+		keyPair, err = core.MakeKeyPairFromString(config.PrivateBls)
+		if err != nil {
+			return nil, err
+		}
+
+		config.ID = keyPair.GetPubKeyG1().GetOperatorID()
+	} else {
+		pkBytes, err := hex.DecodeString(config.BLSPublicKeyHex)
```
samlaf, line 131: personally find this a bit confusing to read.
Is the intention that:
- test mode: bls private key
- prod mode: bls sk file OR bls remote signer
?

in config.go above you first check if !testmode, then build the bls remote signer, so I think here you should also duplicate the logic (for redundancy and future buggy refactor proofing) - check
```
if !testMode && config.PrivateBls != ""
```
or something. Haven't thought this through, but I personally don't like any implicit stuff (like implicitly turning bls remote signer when some other flag is present. Prefer everything to have an "enabled" switch. Perhaps you can have a `SignerType` flag, and then crash when that thing is set to private-key and testMode is set to disabled?

node/node.go
```diff
@@ -384,12 +428,41 @@ func (n *Node) ProcessBatch(ctx context.Context, header *core.BatchHeader, blobs
...
+		sigResp, err := n.BLSSigner.SignGeneric(
+			ctx,
+			&blssignerV1.SignGenericRequest{
+				PublicKey: n.Config.BLSPublicKeyHex,
+				Password:  n.Config.BLSKeyPassword,
+				Data:      data[:],
+			},
+		)
+		if err != nil {
+			return nil, fmt.Errorf("failed to sign data: %w", err)
```
samlaf, line 454: security check: are you sure that the error returned from your bls remote signer never leaks any sensitive info?

samlaf (commented):
Overall LGTM. Added a bunch of nits

node/node.go
```diff
@@ -114,6 +116,47 @@ func NewNode(reg *prometheus.Registry, config *Config, pubIPProvider pubip.Provi
...
+
+		nodeLogger.Info("creating signer client", "url", config.BLSRemoteSignerUrl)
+		creds := insecure.NewCredentials()
+		if config.BLSSignerTLSCertFilePath != "" {
+			creds, err = credentials.NewClientTLSFromFile(config.BLSSignerTLSCertFilePath, "")
+			if err != nil {
+				return nil, err
+			}
+		}
+		conn, err := grpc.NewClient(
```
ian-shim, line 151: What happens if this connection gets broken while node is running?

node/node.go
```diff
@@ -114,6 +116,47 @@ func NewNode(reg *prometheus.Registry, config *Config, pubIPProvider pubip.Provi
 	// Create ChainState Client
 	cst := eth.NewChainState(tx, client)
 
+	var keyPair *core.KeyPair
+	var blsClient blssignerV1.SignerClient
```
ian-shim, line 120: I feel like this one-of pattern where one is expected to be nil and the other is not is very brittle. 
There is this pattern in `Config` where `PrivateBls` can be empty and here where one of either `keyPair` or `blsClient` is nil. 
Downstream, these attributes can't be assumed to be non-nil and there needs to be if/else nil check whenever they're accessed. 
Is there a blsClient abstraction that allows local signing with private bls key?

samlaf: Agreed! This is what I was trying to describe in https://github.com/Layr-Labs/eigenda/pull/745#discussion_r1838528665

node/node.go
samlaf: A grpc « connection » is really a virtual connection. It’s not 1-1 with the underlying tcp connection. If tcp conn drops the grpc runtime dials a new connection.

don’t think grpc « connection » can ever be disconnected. But worth checking if there are conditions that need to be checked/handled.

node/flags/flags.go
shrimalmadhur: right now it doesn't support auth - you can ask for any public key hex. but I think we will support it eventually.

node/grpc/server.go
shrimalmadhur: yea - good suggestion

node/node.go
shrimalmadhur: > Is there a blsClient abstraction that allows local signing with private bls key?
I am thinking of having this abstraction in eigensdk and then you just initialize signer and based on config it would sign. Once I do that - we can switch it out.

node/node.go
shrimalmadhur: > test mode: bls private key
> prod mode: bls sk file OR bls remote signer
> ?

Yes - I am not aware where test mode is use? maybe I think in inabox testing but yes that's the intention - in prod you can use either encrypted sk file or remote signer

shrimalmadhur: so what I can do is - have an explicit flag which switches between sk file and bls remote signer. `PrivateBls` is not really used in prod. maybe @ian-shim can confirm but I think thats true.

node/node.go
shrimalmadhur: nope it does not. nothing ever returns any sensitive information from the signer itself.

node/node.go
shrimalmadhur: maybe we should eventually remove out test node and use either sk file or remote signer as that clearly mimics prod usage

shrimalmadhur: the ugly thing about testMode and prod mode with sk is - both these cases populate `privateBls` and that's why checking testMode is not going to work here. since `privateBls` is populated in non testMode too. https://github.com/Layr-Labs/eigenda/pull/745/files#diff-923e406f4251573464f7748834b59f4747b3f6ec4eadda667614e2875a3b811aR166-R174

I think ideal would be to remove this testMode and eventually put the prod signing behind eigensdk as I said in above comment.

node/grpc/server.go
shrimalmadhur: 40777795d4f4d89cea862f8d8a7f5115e12af6dd

node/node.go
samlaf: Test mode I think is just used to allow passing a sk directly? Think it should be allowed to simplify tests.

shrimalmadhur: if that would remain - in current way - `privateBls` will be populated for both. I guess we could change eventually when we abstract out signing behind eigensdk

pschork:
Bump go-ethereum to latet to resolve unit tests

go-verkle was causing unit test failures with `type *banderwagon.Element has no field or method BytesUncompressed`

```
    go mod graph | grep go-verkle
    github.com/Layr-Labs/eigenda github.com/gballet/go-verkle@v0.1.1-0.20231031103413-a67434b50f46
    github.com/ethereum/go-ethereum@v1.14.0 github.com/gballet/go-verkle@v0.1.1-0.20231031103413-a67434b50f46
    github.com/gballet/go-verkle@v0.1.1-0.20231031103413-a67434b50f46 github.com/crate-crypto/go-ipa@v0.0.0-20231025140028-3c0104f4b233
    github.com/gballet/go-verkle@v0.1.1-0.20231031103413-a67434b50f46 golang.org/x/sync@v0.1.0
    github.com/gballet/go-verkle@v0.1.1-0.20231031103413-a67434b50f46 github.com/bits-and-blooms/bitset@v1.7.0
    github.com/gballet/go-verkle@v0.1.1-0.20231031103413-a67434b50f46 github.com/consensys/bavard@v0.1.13
    github.com/gballet/go-verkle@v0.1.1-0.20231031103413-a67434b50f46 github.com/consensys/gnark-crypto@v0.12.1
    github.com/gballet/go-verkle@v0.1.1-0.20231031103413-a67434b50f46 github.com/mmcloughlin/addchain@v0.4.0
    github.com/gballet/go-verkle@v0.1.1-0.20231031103413-a67434b50f46 golang.org/x/sys@v0.9.0github.com/gballet/go-verkle@v0.1.1-0.20231031103413-a67434b50f46 rsc.io/tmplfunc@v0.0.3
```

Resolve grpc.Dial deprecations due to cerbereus bumping grpc from v1.59.0 -> v1.64.1
 - Updates `grpc.Dial` -> `grpc.NewClient`
 - Removes `withBlock()` option from `grpcConnection.Dial` used by syntetic checks.

see https://github.com/grpc/grpc-go/blob/master/Documentation/anti-patterns.md for why `withBlock()` was deprecated.

node/node.go
ian-shim: Is this something that will be done & reflected in this PR? Or in a future PR?

ian-shim (commented):
one last comment

ian-shim (approved):

PR #749, opened 2024-09-05
Reader fragment

cody-littley:

PR 8 of 9 for the traffic generator project. This PR adds a worker that reads blobs.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests


tools/traffic/workers/mock_chain_client.go
```diff
@@ -0,0 +1,24 @@
+package workers
```
ian-shim, line 1: Do we need this new mock impl for chain client and retrieval client below? 
Looks like we already have mock implementations

api/clients/mock/retrieval_client.go
```diff
@@ -35,3 +35,21 @@ func (c *MockRetrievalClient) RetrieveBlob(
...
 }
+
+func (c *MockRetrievalClient) RetrieveBlobChunks(
+	ctx context.Context,
+	batchHeaderHash [32]byte,
+	blobIndex uint32,
+	referenceBlockNumber uint,
+	batchRoot [32]byte,
+	quorumID core.QuorumID) (*clients.BlobChunks, error) {
+
```
ian-shim, line 46: Do we want `args := c.Called()` and pass return values?
I guess this method isn't being used in any of test cases rn so wouldn't matter, but would be good to implement it instead of always returning nil

tools/traffic/workers/blob_reader.go
```diff
@@ -0,0 +1,223 @@
...
+	"github.com/Layr-Labs/eigenda/tools/traffic/metrics"
+	"github.com/Layr-Labs/eigenda/tools/traffic/table"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	gcommon "github.com/ethereum/go-ethereum/common"
+	"math/big"
+	"sync"
+	"time"
+)
+
+// BlobReader reads blobs from a disperser at a configured rate.
```
ian-shim, line 22: reads blobs from _the DA network_?

tools/traffic/workers/blob_reader.go
```diff
@@ -0,0 +1,223 @@
...
+	config *config.WorkerConfig
+
+	retriever   clients.RetrievalClient
+	chainClient eth.ChainClient
+
+	// blobsToRead blobs we are required to read a certain number of times.
+	blobsToRead *table.BlobStore
+
+	generatorMetrics           metrics.Metrics
+	fetchBatchHeaderMetric     metrics.LatencyMetric
```
ian-shim, line 43: could be cleaner to consolidate these metrics into a struct

tools/traffic/workers/blob_reader.go
```diff
@@ -0,0 +1,223 @@
...
+
+// Start begins a blob reader goroutine.
+func (reader *BlobReader) Start() {
+	reader.waitGroup.Add(1)
+	ticker := time.NewTicker(reader.config.ReadRequestInterval)
+	go func() {
+		defer reader.waitGroup.Done()
+		for {
+			select {
+			case <-(*reader.ctx).Done():
```
ian-shim, line 104: nit: log or return `reader.ctx.Err()`

tools/traffic/workers/blob_reader.go
```diff
@@ -0,0 +1,223 @@
...
+		readFailureMetric:          generatorMetrics.NewCountMetric("read_failure"),
+		operatorSuccessMetrics:     make(map[core.OperatorID]metrics.CountMetric),
+		operatorFailureMetrics:     make(map[core.OperatorID]metrics.CountMetric),
+		requiredReadPoolSizeMetric: generatorMetrics.NewGaugeMetric("required_read_pool_size"),
+		optionalReadPoolSizeMetric: generatorMetrics.NewGaugeMetric("optional_read_pool_size"),
+	}
+}
+
+// Start begins a blob reader goroutine.
+func (reader *BlobReader) Start() {
```
ian-shim, line 97: nit: usually, receivers are named with just one letter like `r`

tools/traffic/workers/blob_reader.go
```diff
@@ -0,0 +1,223 @@
...
+	cancel()
+	if err != nil {
+		reader.logger.Error("failed to get batch header", "err:", err)
+		reader.fetchBatchHeaderFailure.Increment()
+		return
+	}
+	reader.fetchBatchHeaderSuccess.Increment()
+
+	var batchHeaderHash [32]byte
+	copy(batchHeaderHash[:], metadata.BatchHeaderHash[:])
```
ian-shim, line 142: why do we need this copy?

tools/traffic/workers/blob_reader.go
```diff
@@ -0,0 +1,223 @@
...
+
+	ctxTimeout, cancel = context.WithTimeout(*reader.ctx, reader.config.RetrieveBlobChunksTimeout)
```
ian-shim, line 144: usually, clean ups are done like `defer cancel()` right after these methods are initialized

tools/traffic/workers/blob_reader.go
```diff
@@ -0,0 +1,223 @@
...
+		reader.logger.Error("failed to read chunks", "err:", err)
+		reader.readFailureMetric.Increment()
+		return
+	}
+	reader.readSuccessMetric.Increment()
+
+	assignments := chunks.Assignments
+
+	data, err := reader.retriever.CombineChunks(chunks)
+	if err != nil {
```
ian-shim, line 165: This is an interesting case and may warrant different error handling vs. the read failure above.
Read failures may happen for different reasons but mostly suggest some issues with network health. 
Blob recovery failure here most certainly suggests a bug. So we may set up noisier alert for this one

ian-shim (commented):
Looks good! Just few comments

tools/traffic/workers/mock_chain_client.go
cody-littley: I've removed the new mock implementations and used the existing ones.

api/clients/mock/retrieval_client.go
cody-littley: I had to fix this in order to get my tests working (after deleting the duplicate implementation of the mock retrieval client).

tools/traffic/workers/blob_reader.go
cody-littley: Good catch. The code is reading from the DA nodes, the comment was wrong. Fixed.

tools/traffic/workers/blob_reader.go
cody-littley: done

tools/traffic/workers/blob_reader.go
cody-littley: log statement added

tools/traffic/workers/blob_reader.go
cody-littley: renamed to `r`

tools/traffic/workers/blob_reader.go
cody-littley: Hmm... I honestly can't remember why I did it that way. Refactored so that it doesn't use a copy.

tools/traffic/workers/blob_reader.go
cody-littley: change made

tools/traffic/workers/blob_reader.go
cody-littley: This makes an error log and updates a metric. What did you have in mind for a noisier alert?

tools/traffic/workers/blob_reader.go
```diff
@@ -0,0 +1,231 @@
...
+		waitGroup:   waitGroup,
+		logger:      logger,
+		config:      config,
+		retriever:   retriever,
+		chainClient: chainClient,
+		blobsToRead: blobStore,
+		metrics: &blobReaderMetrics{
+			generatorMetrics:           generatorMetrics,
+			fetchBatchHeaderMetric:     generatorMetrics.NewLatencyMetric("fetch_batch_header"),
+			fetchBatchHeaderSuccess:    generatorMetrics.NewCountMetric("fetch_batch_header_success"),
```
jianoaix, line 86: How does it compare to having a metric with a "status" label for success/failure or valid/invalid (instead of one metric for each status)?

tools/traffic/workers/blob_reader.go
```diff
@@ -0,0 +1,231 @@
...
+	if metadata == nil {
+		// There are no blobs that we are required to read.
+		return
+	}
+
+	r.metrics.requiredReadPoolSizeMetric.Set(float64(r.blobsToRead.Size()))
+
+	ctxTimeout, cancel := context.WithTimeout(*r.ctx, r.config.FetchBatchHeaderTimeout)
+	defer cancel()
+	batchHeader, err := metrics.InvokeAndReportLatency(r.metrics.fetchBatchHeaderMetric,
```
jianoaix, line 136: I wouldn't consider a mix of core logic with monitoring code quite clean, and would prefer a separation.

tools/traffic/workers/blob_reader.go
```diff
@@ -0,0 +1,231 @@
...
+	chunks, err := metrics.InvokeAndReportLatency(r.metrics.readLatencyMetric, func() (*clients.BlobChunks, error) {
+		return r.retriever.RetrieveBlobChunks(
+			ctxTimeout,
+			metadata.BatchHeaderHash,
+			uint32(metadata.BlobIndex),
+			uint(batchHeader.ReferenceBlockNumber),
+			batchHeader.BlobHeadersRoot,
+			core.QuorumID(0))
+	})
+	cancel()
```
jianoaix, line 162: Why cancel the context here? Do you want to wait for timeout?

tools/traffic/workers/blob_reader.go
```diff
@@ -0,0 +1,231 @@
...
+			}
+		}
+	}
+}
+
+// reportChunk reports a successful chunk read.
+func (r *BlobReader) reportChunk(operatorId core.OperatorID) {
+	metric, exists := r.metrics.operatorSuccessMetrics[operatorId]
+	if !exists {
+		metric = r.metrics.generatorMetrics.NewCountMetric(fmt.Sprintf("operator_%x_returned_chunk", operatorId))
```
jianoaix, line 202: So there will be hundreds of metrics, not sure if that's a quite usable thing.

cody-littley: I agree that this is going to be noisy. Is it critical for this utility to determine which nodes are failing to return chunks, or would it be ok just to report the fraction of nodes that return chunks?

If knowledge of which nodes are failing to return chunks is important, do you have any recommendations for how to extract this sort of data without using metrics this way?

tools/traffic/workers/blob_reader.go
cody-littley: The way this code is currently set up, all of the latency metrics use the same base metric name but with different labels. The same is true for the count metrics and the gauge metrics. If anything, I feel like I'm using labels too much right now. 🙃 

This metrics pattern is something that extends to all three components in the traffic pattern (the writer, the status checker, and the reader). Should I work on this as a part of this PR, or split that work into a separate one?

tools/traffic/workers/blob_reader.go
cody-littley: There are a few patterns I could utilize for reporting latency. Which do you prefer?

# Pattern A

Wrap the base function in another function that calls the base function and records the amount of time it took to execute. (This is the pattern currently in the PR)

# Pattern B

Measure the time before and after the function call.

```
start := time.Now()
foo()
end := time.Now()
metrics.reportDurationOfFoo(start, end)
```

# Pattern C

Embed the measurement of the time to execute the method inside the method itself.

```
func foo() {
    start := time.now()
    // business logic
    end := time.now()
    metrics.reportDurationOfFoo(start, end)
```

# Pattern D

Is there a strategy you like that I didn't mention?

tools/traffic/workers/blob_reader.go
ian-shim: Hmm would keeping track of counts for only failed retrieval reduce the number of metrics here?

ian-shim (commented):
lgtm, @jianoaix can you chime in to Cody's responses?

tools/traffic/workers/blob_reader.go
cody-littley: sounds like a plan

tools/traffic/workers/blob_reader.go
cody-littley: I've removed the `invokeAndReportLatency()` metric in favor of pattern B

tools/traffic/workers/blob_reader.go
cody-littley: change made

tools/traffic/workers/blob_reader.go
ian-shim: I think this could be updated in this PR

ian-shim (approved):
lgtm

tools/traffic/workers/blob_reader.go
```diff
@@ -0,0 +1,238 @@
...
+		big.NewInt(int64(0)),
+		nil)
+	if err != nil {
+		r.logger.Error("failed to get batch header", "err:", err)
+		r.metrics.fetchBatchHeaderFailure.Increment()
+		return
+	} else {
+		end := time.Now()
+		duration := end.Sub(start)
+		r.metrics.fetchBatchHeaderMetric.ReportLatency(duration)
```
jianoaix, line 150: These 3 lines can be simplified as `r.metrics.fetchBatchHeaderMetric.ReportLatency(time.Since(start))`

tools/traffic/workers/blob_reader.go
```diff
@@ -0,0 +1,238 @@
...
+		ctxTimeout,
+		gcommon.HexToAddress(r.config.EigenDAServiceManager),
+		metadata.BatchHeaderHash[:],
+		big.NewInt(int64(0)),
+		nil)
+	if err != nil {
+		r.logger.Error("failed to get batch header", "err:", err)
+		r.metrics.fetchBatchHeaderFailure.Increment()
+		return
+	} else {
```
jianoaix, line 147: No need `else` after a `return`

tools/traffic/workers/blob_reader.go
```diff
@@ -0,0 +1,238 @@
...
+		metadata.BatchHeaderHash,
+		uint32(metadata.BlobIndex),
+		uint(batchHeader.ReferenceBlockNumber),
+		batchHeader.BlobHeadersRoot,
+		core.QuorumID(0))
+	if err != nil {
+		r.logger.Error("failed to read chunks", "err:", err)
+		r.metrics.readFailureMetric.Increment()
+		return
+	} else {
```
jianoaix, line 170: similar here

tools/traffic/workers/blob_reader.go
jianoaix: This part needs to be careful about cardinality. SG to follow up to cap the scope here. LMK if you need some inputs on refactoring this later.

tools/traffic/workers/blob_status_tracker.go
```diff
@@ -203,14 +203,16 @@ func (tracker *BlobStatusTracker) getBlobStatus(key *UnconfirmedKey) (*disperser
...
-		func() (*disperser.BlobStatusReply, error) {
-			return tracker.disperser.GetBlobStatus(ctxTimeout, key.Key)
-		})
+	start := time.Now()
+	status, err := tracker.disperser.GetBlobStatus(ctxTimeout, key.Key)
 
 	if err != nil {
 		tracker.getStatusErrorCountMetric.Increment()
 		return nil, err
+	} else {
```
jianoaix, line 212: Here and below: can be simplified as mentioned above

tools/traffic/workers/blob_reader.go
cody-littley: The primary reason why I'd like to address metrics as a follow up is because there are other metrics I added in prior PRs that also need to be considered. Makes sense to me to do all the metrics at once.

tools/traffic/workers/blob_reader.go
```diff
@@ -0,0 +1,238 @@
...
+	r.waitGroup.Add(1)
+	ticker := time.NewTicker(r.config.ReadRequestInterval)
+	go func() {
+		defer r.waitGroup.Done()
+		for {
+			select {
+			case <-(*r.ctx).Done():
+				err := (*r.ctx).Err()
+				if err != nil {
+					r.logger.Info("blob r context closed", "err:", err)
```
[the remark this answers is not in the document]
cody-littley: fixed

tools/traffic/workers/blob_reader.go
cody-littley: Makes sense. Will address this in the follow up PR.

tools/traffic/workers/blob_reader.go
cody-littley: change made

tools/traffic/workers/blob_reader.go
cody-littley: simplified

tools/traffic/workers/blob_reader.go
cody-littley: done

tools/traffic/workers/blob_status_tracker.go
cody-littley: done

PR #750, opened 2024-09-06
Keep encoded blob result until blob reaches terminal state

ian-shim:
Currently, the encoded blob result is removed once blob state transitions from `PROCESSING` to `DISPERSING`. 
However, a blob can be [fetched](https://github.com/Layr-Labs/eigenda/blob/master/disperser/batcher/encoding_streamer.go#L205-L208) for encoding and then be transitioned to `DISPERSING` state before the encoding request is complete because the mutex is released and allows the blob to be returned as part of a batch. 
This can cause an issue where the blob is successfully confirmed, but the remaining encoding result causes the blob to be dispersed again. The encoded blob state management needs to be enforced more tightly. 
This PR does not remove the encoded result when it's returned as part of the batch. Instead, it keeps it until the blob reaches a terminal state so that it can be properly [dedup'd](https://github.com/Layr-Labs/eigenda/blob/master/disperser/batcher/encoding_streamer.go#L241) from the concurrent encoding requests. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


disperser/batcher/batcher.go
```diff
@@ -312,10 +312,12 @@ func (b *Batcher) updateConfirmationInfo(
 
 		if status == disperser.Confirmed {
 			if _, updateConfirmationInfoErr = b.Queue.MarkBlobConfirmed(ctx, metadata, confirmationInfo); updateConfirmationInfoErr == nil {
+				b.EncodingStreamer.RemoveEncodedBlob(metadata)
```
bxue-l2, line 315: what is the chance of memory leak, when it falls into the last else statement part. CAn we just remove it regardless of the "if" condition

bxue-l2: in addition to that. the memory load on the batcher for 600sec batch interval would be 10MB/s*600 = 6GB. It looks manageable

ian-shim: Sg, removing regardless of the if condition seems benign

PR #751, opened 2024-09-07
Fix grpc size comments

jianoaix:
proto.Size() returns int, which is 64 bits on 64 bits machine.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


jianoaix:
Ping?

dmanc (approved):

PR #752, opened 2024-09-07
chore(inabox): multiplex inabox logs to stdout to help debug easier

samlaf:

Debugging why inabox was failing on my machine took me longer than it should have. Couldn't figure out why the log statements were not logging to stdout. Added a multiplexer to send logs both to the output file, but also to stdout. If a deploy command fails, I should be able to see the logs directly without needing to figure out which file the logs are piped too.

Also added a print statement to print the forge command being executed to help with debugging.

### DIFF
--- a/inabox/deploy/deploy.go
+++ b/inabox/deploy/deploy.go
@@ -3,6 +3,7 @@ package deploy
 import (
 	"encoding/json"
 	"fmt"
+	"io"
 	"log"
 	"math/big"
 	"os"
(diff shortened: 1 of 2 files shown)


jianoaix (approved):
Have you checked if the stdout will be overwhelmed by logs?

samlaf:
> Have you checked if the stdout will be overwhelmed by logs?

What does overwhelmed by logs mean? It's just "deploying xyz..." logs.
Even if it was huge logs I think its fine. Think of when you `make` a huge c binary and there's just infinite streams of logs. It doesn't really matter. What you want is the error printout if it ever crashes.

You can merge whenver you feel like it's fine. I'm not authorized to merge somehow.

PR #754, opened 2024-09-08
chore: refactor to single multistage dockerfile with bake file

samlaf:

All of the Dockerfiles were doing very similar things. Refactored to make use of buildkit multistage build (see [this article](https://depot.dev/blog/buildkit-in-depth) for explanation). Thought there would be greater speedup but it went from ~2min20sec to ~1min30sec. Greatest advantage is cleanliness and maintainability. For eg with a single Dockerfile we can see the commonalities between the different targets and refactor as needed.

Also added a bake file with different targets. This replaces the docker compose files. Bake files are a new feature of docker which allows for grouping targets together. They also support extra useful features like inheritance, so easy to make release targets that add extra platforms for eg.

This refactor also makes it easier to inject into all images the same ldflags that we currently only inject into the node release build, but not into the nodeplugin or any other image. Easier to see patterns like this.

## Testing

Have built the images manually successfully, but haven't done any testing with those images. Also haven't tested the pipeline changes that push to ghcr.


docker-bake.hcl
```diff
@@ -0,0 +1,111 @@
...
+}
+
+# RELEASE TARGETS
+
+target "_release" {
+    platforms = ["linux/amd64", "linux/arm64"]
+}
+
+target "node-release" {
+  inherits  = ["node", "_release"]
```
shrimalmadhur, line 106: the image built by release should be in format `opr-node` and `opr-nodeplugin` since that's what we do for public release. so you might want to override the tags in here.

docker-bake.hcl
```diff
@@ -0,0 +1,111 @@
...
+group "all" {
+  targets = ["node-group", "disperser-group"]
+}
+
+group "node-group" {
+  targets = ["node", "nodeplugin"]
+}
+
+group "disperser-group" {
+  targets = ["batcher", "disperser", "encoder", "retriever", "churner"]
```
jianoaix, line 34: Technically retriever and churner are not part of disperser

jianoaix (commented):
Look good

pschork (approved):

docker-bake.hcl
samlaf: ah good point will leave them as separate individual targets then.

.github/workflows/docker-publish.yaml
```diff
@@ -47,12 +47,12 @@ jobs:
...
-      # Build And Push Image
-      - name: Build Docker image
-        run: docker compose -f docker-compose-internal.yaml build
-      - name: Push Docker image
+      # Build And Push Images
+      - name: Build Docker images
+        run: docker buildx bake
+      - name: Push Docker images
         if: github.ref == 'refs/heads/master'
-        run: docker compose -f docker-compose-internal.yaml push
```
dmanc, line 55: What do you think of making it a separate workflow and then deprecating this one?

Dockerfile
```diff
@@ -0,0 +1,109 @@
+# syntax=docker/dockerfile:1
+
+# Declare build arguments
+# TODO: this is only used for node image right now, should we also use it for nodeplugin?
```
dmanc, line 4: Yeah we should use it for all

disperser/cmd/apiserver/Dockerfile
```diff
@@ -1,26 +0,0 @@
-FROM golang:1.21.1-alpine3.18 as builder
```
dmanc, line 1: Similar to previous comment, do we want to make the PR additive first and then deprecate this stuff later?

.github/workflows/docker-publish.yaml
samlaf: Feel it's not needed. If ever we have a problem with this we can git revert the PR.

docker-bake.hcl
samlaf: [a906a10](https://github.com/Layr-Labs/eigenda/pull/754/commits/a906a105d6229041cc1508d4477ff69c3e2ce8e1)

disperser/cmd/apiserver/Dockerfile
samlaf: How would you test this if we don't remove it? I feel let's just test it, and if for some reason the releases don't work or there's something super buggy with it, then we just git revert.

docker-bake.hcl
samlaf: [97b4eed](https://github.com/Layr-Labs/eigenda/pull/754/commits/97b4eedce69bfd32aef7b2209d2081a7d14ebc6b)

docker-bake.hcl
```diff
@@ -0,0 +1,111 @@
...
+target "nodeplugin" {
+  context    = "."
+  dockerfile = "./Dockerfile"
+  target     = "nodeplugin"
+  tags       = ["ghcr.io/layr-labs/eigenda/nodeplugin:${BUILD_TAG}"]
+}
+
+# RELEASE TARGETS
+
+target "_release" {
```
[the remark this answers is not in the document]
samlaf: seems like it's not private, since you can still print it with `docker buildx bake _release --print`
Just indicative that it's not meant to be a real target.

Dockerfile
samlaf: just checked, nodeplugin doesn't even define those variables. Let's do it in a separate PR if it's ever needed. Probably add it to all binaries in one go.

disperser/cmd/apiserver/Dockerfile
dmanc: We can still test it because there would be two separate pipelines. I'm fine with reverting if there's problems.

dmanc (approved):

PR #755, opened 2024-09-09
feat: dockerfile for deploying contracts

samlaf:

Need an image for deploying contracts as part of my [kurtosis inabox](https://github.com/Layr-Labs/eigenda-package) project. Generally think this is something we should have as well. Every time we deploy, we should create an image tagged with the release name/version number that contains the exact contracts that were deployed. This way anyone that wants to work locally can use that image to deploy to anvil or some local devnet and reproduce the contracts at that point.

Not set on the exact dockerfile; open to improvements. Also think we should have all of our ad-hoc docker images in the bake file in #754.

Published this image at https://github.com/orgs/Layr-Labs/packages/container/package/eigenda%2Fcontract-deployer. Would a kind admin soul please link that image with the eigenda repo for easier discovery. cc @anupsv 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/contracts/script/SetUpEigenDA.s.sol
+++ b/contracts/script/SetUpEigenDA.s.sol
@@ -25,7 +25,7 @@ import "forge-std/StdJson.sol";
 // forge script script/Deployer.s.sol:SetupEigenDA --rpc-url $RPC_URL  --private-key $PRIVATE_KEY --broadcast -vvvv
 contract SetupEigenDA is EigenDADeployer, EigenLayerUtils {
 
-    string deployConfigPath = "script/eigenda_deploy_config.json";
+    string deployConfigPath = "script/input/eigenda_deploy_config.json";
 
     // deploy all the EigenDA contracts. Relies on many EL contracts having already been deployed.
     function run() external {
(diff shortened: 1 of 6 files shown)


contracts/script/SetUpEigenDA.s.sol
```diff
@@ -25,7 +25,7 @@ import "forge-std/StdJson.sol";
 // forge script script/Deployer.s.sol:SetupEigenDA --rpc-url $RPC_URL  --private-key $PRIVATE_KEY --broadcast -vvvv
 contract SetupEigenDA is EigenDADeployer, EigenLayerUtils {
 
-    string deployConfigPath = "script/eigenda_deploy_config.json";
+    string deployConfigPath = "script/input/eigenda_deploy_config.json";
```
samlaf, line 28: had to change this path to make the kurtosis devnet work (otherwise uploading the config would overwrite the rest of the directory.

I think I changed everywhere else that depended on this but please double check. Also I feel like we might want to have this path as an arg to the run function instead of hardcoded?

0x0aa0 (approved):

PR #761, opened 2024-09-14
Fix docker-bake issues with devops builds

pschork:
On ECR there is no "repository" subpath defined

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -87,7 +87,7 @@ target "traffic-generator" {
   inherits = ["docker-metadata-action"]
   context    = "."
   dockerfile = "./trafficgenerator.Dockerfile"
-  target     = "traffic-generator"
+  target     = "generator"
 }
 
 target "dataapi" {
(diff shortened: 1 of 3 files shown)


docker-bake.hcl
```diff
@@ -1,10 +1,6 @@
 # VARIABLES
 variable "REGISTRY" {
-  default = "ghcr.io"
-}
-
-variable "REPOSITORY" {
-  default = "layr-labs/eigenda"
+  default = "ghcr.io/layr-labs/eigenda"
```
samlaf, line 3: Do you want to add the ecr repo as a comment for reference?

samlaf (approved):

docker-bake.hcl
pschork: There is ecr for each account dev/preprod/testnet/mainnet, but mainly I don't want to leak internal account ecr urls in oss repo

samlaf: fair point

PR #770, opened 2024-09-20
rm sig check

bxue-l2:
## Feature addition
 - allowing users without disperser to use EigenDAClient. For rollup, rollup nodes that do not assume the batcher role do not need the ability to disperse blobs. But the current EigenDAClient necessitate a hex private keys. This PR serves to remove such restrictions. Such that anyone can read from EigenDA without a privateKey

## Method
 - this PR adds a new struct that implement the interface [BlobRequestSigner](https://github.com/Layr-Labs/eigenda/blob/master/core/auth.go#L7C6-L7C23) called LocalNoopSigner, it returns error for all methods. For a user without setting the private key, but uses the signer to send the blob. It will encounter a problem. 

### Send without privateKey
Screenshot 2024-09-20 at 12 37 39 PM

### Send with privateKey (i.e it is working)
Screenshot 2024-09-20 at 1 36 23 PM

### retrieve without privatekey (working)
Screenshot 2024-09-20 at 2 43 32 PM

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/config.go
+++ b/api/clients/config.go
@@ -52,6 +52,10 @@ func (c *EigenDAClientConfig) CheckAndSetDefaults() error {
 	if c.ResponseTimeout == 0 {
 		c.ResponseTimeout = 30 * time.Second
 	}
+	if len(c.SignerPrivateKeyHex) > 0 && len(c.SignerPrivateKeyHex) != 64 {
+		return fmt.Errorf("a valid length SignerPrivateKeyHex needs to have 64 bytes")
+	}
+
 	if len(c.RPC) == 0 {
 		return fmt.Errorf("EigenDAClientConfig.RPC not set")
 	}
--- a/api/clients/config.go
+++ b/api/clients/config.go
@@ -52,9 +52,6 @@ func (c *EigenDAClientConfig) CheckAndSetDefaults() error {
 	if c.ResponseTimeout == 0 {
 		c.ResponseTimeout = 30 * time.Second
 	}
-	if len(c.SignerPrivateKeyHex) != 64 {
-		return fmt.Errorf("EigenDAClientConfig.SignerPrivateKeyHex should be 64 hex characters long, should not have 0x prefix")
-	}
 	if len(c.RPC) == 0 {
 		return fmt.Errorf("EigenDAClientConfig.RPC not set")
 	}
--- a/api/clients/disperser_client.go
+++ b/api/clients/disperser_client.go
@@ -107,10 +107,14 @@ func (c *disperserClient) DisperseBlob(ctx context.Context, data []byte, quorums
 }
 
 func (c *disperserClient) DisperseBlobAuthenticated(ctx context.Context, data []byte, quorums []uint8) (*disperser.BlobStatus, []byte, error) {
+	if c.signer == nil {
+		return nil, nil, fmt.Errorf("uninitialized signer for authenticated dispersal")
+	}
+
 	// first check if signer is valid
 	accountId, err := c.signer.GetAccountID()
 	if err != nil {
-		return nil, nil, err
+		return nil, nil, fmt.Errorf("please configure signer key if you want to use authenticated endpoint %w", err)
 	}
 
 	quorumNumbers := make([]uint32, len(quorums))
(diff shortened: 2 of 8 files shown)


api/clients/config.go
```diff
@@ -52,9 +52,6 @@ func (c *EigenDAClientConfig) CheckAndSetDefaults() error {
 	if c.ResponseTimeout == 0 {
 		c.ResponseTimeout = 30 * time.Second
 	}
-	if len(c.SignerPrivateKeyHex) != 64 {
```
ian-shim, line 55: Instead of removing this check, check length only when this value is not nil?

api/clients/disperser_client.go
```diff
@@ -107,6 +107,28 @@ func (c *disperserClient) DisperseBlob(ctx context.Context, data []byte, quorums
 }
 
 func (c *disperserClient) DisperseBlobAuthenticated(ctx context.Context, data []byte, quorums []uint8) (*disperser.BlobStatus, []byte, error) {
+	// first check if signer is valid
+	accountId, err := c.signer.GetAccountID()
```
ian-shim, line 111: check if `c.signer == nil`?

api/clients/eigenda_client.go
```diff
@@ -41,7 +42,13 @@ func NewEigenDAClient(log log.Logger, config EigenDAClientConfig) (*EigenDAClien
 		return nil, fmt.Errorf("failed to parse EigenDA RPC: %w", err)
 	}
 
-	signer := auth.NewLocalBlobRequestSigner(config.SignerPrivateKeyHex)
+	var signer core.BlobRequestSigner
+	if len(config.SignerPrivateKeyHex) == 64 {
+		signer = auth.NewLocalBlobRequestSigner(config.SignerPrivateKeyHex)
+	} else {
```
ian-shim, line 48: only do this when `config.SignerPrivateKeyHex` is nil, and return error when `config.SignerPrivateKeyHex` is anything other than nil or has length != 64

api/clients/disperser_client.go
```diff
@@ -107,6 +107,28 @@ func (c *disperserClient) DisperseBlob(ctx context.Context, data []byte, quorums
...
+	if err != nil {
+		return nil, nil, err
```
ian-shim, line 113: would it be even more clear if we add something like `please configure signer key if you want to use authenticated endpoint`?

ian-shim (commented):
looks good, left some comments

api/clients/config.go
bxue-l2: SignerPrivateKeyHex is a string defined in the config, it cannot be nil.

ian-shim: *empty string

api/clients/disperser_client.go
bxue-l2: after some reflections, we could have just assign the signer interface as nil if privateKeyHex is not provided. This make NoopSigner impl looks like a over-do. But I kind of like noopSigner because it complies to the interface, and leave less space for error.

api/clients/eigenda_client.go
bxue-l2: I can add the error case, but `config.SignerPrivateKeyHex` is a string, what I am missing here, there is no reason to check nil

ian-shim: I want to make sure we check the key length. If this key is provided (non-empty string), it should be length 64 and error out otherwise.

bxue-l2: yes, pushed that change

api/clients/config.go
bxue-l2: https://go.dev/play/p/6gLQTyOMg5o

empty string cannot be compared with nil

ian-shim: Oh sorry what I meant was we need to keep this check when the key is non-empty. 
i.e. `if len(c.SignerPrivateKeyHex) > 0 && len(c.SignerPrivateKeyHex) != 64` then return error

api/clients/disperser_client.go
bxue-l2: my main thing about it is because the signer is a interface, if it is a struct pointer, I will fall immediately after nil check

ian-shim: Yeah shouldn't it fail immediately if `c.signer` is nil?

api/clients/config.go
bxue-l2: got it, added

api/clients/disperser_client.go
bxue-l2: I did it, and the commit actually check it.

https://github.com/Layr-Labs/eigenda/pull/770/files#diff-80719721baf29a1f6b39d999668efd8ae1938a4d2ece27e1f9889edb486afb87R110

Not sure why github won't show it.

ian-shim (approved):

PR #771, opened 2024-09-20
Instrument the retrieval latency stages

jianoaix:
We need visibility to understand the retrieval latency, which sometimes goes above 1s (too high).

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/apiserver/server.go
+++ b/disperser/apiserver/server.go
@@ -4,13 +4,14 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"google.golang.org/grpc/status"
 	"math/rand"
 	"net"
 	"slices"
 	"strings"
 	"sync"
 	"time"
+
+	"google.golang.org/grpc/status"
 
 	"github.com/Layr-Labs/eigenda/api"
 	commonpb "github.com/Layr-Labs/eigenda/api/grpc/common"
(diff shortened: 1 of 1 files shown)


disperser/apiserver/server.go
```diff
@@ -702,7 +704,7 @@ func (s *DispersalServer) RetrieveBlob(ctx context.Context, req *pb.RetrieveBlob
 			return nil, api.NewResourceExhaustedError(errorString)
 		}
 	}
-
+	s.logger.Debug("checked retrieval blob rate limiting", "requesterID", fmt.Sprintf("%s:%s", origin, RetrievalBlobRateType.Plug()), "duration (ms)", time.Since(stageTimer).Milliseconds())
```
ian-shim, line 707: nit: could we name labels camelcase? or `time.Since(stageTimer).String()` should format it in a human readable way then we probably don't need to indicate the unit in the label

ian-shim (approved):

disperser/apiserver/server.go
jianoaix: The mixed cap is pretty compatible with golang. Sounds good to let `String()` to pretty print.

ian-shim: It's compatible in Golang, but doesn't work nicely when filtering by labels in Loki

jianoaix: feel free to start a conversation on this topic, and if it's indeed a good idea, make a migration cross the board. It's not that useful to start ad hoc style at random places.

PR #774, opened 2024-09-24
Table store

cody-littley:

This PR adds the ability to use tables on top of a store. A very nice to have feature for use cases that want to use multiple logical tables.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests


common/kvstore/batch.go
```diff
@@ -0,0 +1,20 @@
+package kvstore
+
+// Batch is a collection of operations that can be applied atomically to a TableStore.
+type Batch[T any] interface {
```
jianoaix, line 4: Not sure this needs template, Batch may just work with table key

common/kvstore/batch.go
```diff
@@ -0,0 +1,20 @@
...
+	// Delete removes the key from the batch.
+	Delete(key T)
+	// Apply atomically writes all the key / value pairs in the batch to the database.
+	Apply() error
+	// Size returns the number of operations in the batch.
+	Size() uint32
+}
+
+// BatchOperator is an interface for creating new batches.
+type BatchOperator[T any] interface {
```
jianoaix, line 17: For simplicity, it looks Batch creation can be a function of TableStore

common/kvstore/tablestore/table_store_wrapper.go
```diff
@@ -0,0 +1,353 @@
...
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"github.com/Layr-Labs/eigenda/common/kvstore"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	"math"
+	"sort"
+)
+
+// Table ID 0 is reserved for use internal use by the metadata table.
```
jianoaix, line 13: Comment doesn't match code

common/kvstore/tablestore/table_store_wrapper.go
```diff
@@ -0,0 +1,353 @@
...
+const metadataTableID uint32 = math.MaxUint32
+
+// Table ID 1 is reserved for use by the namespace table. This stores a mapping between IDs and table names.
```
jianoaix, line 16: Comment doesn't match code

common/kvstore/tablestore/table_store_wrapper.go
```diff
@@ -0,0 +1,353 @@
...
+	// Builds keys for the namespace table.
+	namespaceTable kvstore.Table
+}
+
+// Wrapper wraps the given Store to create a TableStore.
+//
+// WARNING: it is not safe to access the wrapped store directly while the TableStore is in use. The TableStore uses
+// special key formatting, and direct access to the wrapped store may violate the TableStore's invariants, resulting
+// in undefined behavior.
+func Wrapper(logger logging.Logger, base kvstore.Store) (kvstore.TableStore, error) {
```
jianoaix, line 68: I'd create a TableStore, instead of asking for an external Store and wrapping it. This way the Store can be internally created and no risk of breaking invariants by accessing it directly outside the TableStore.

common/kvstore/tablestore/table_store_wrapper.go
```diff
@@ -0,0 +1,353 @@
...
+
+// Get the prefix for the given table ID and prefix length.
+func getPrefix(tableID uint32) []byte {
+	bytes := make([]byte, 4)
+	binary.BigEndian.PutUint32(bytes, tableID)
+	return bytes
+}
+
+// GetTable gets the table with the given name. If the table does not exist, it is first created.
+func (t *tableStore) GetTable(name string) (kvstore.Table, error) {
```
jianoaix, line 185: I don't see this TableStore is thread-safe (which is required per interface)

cody-littley: In the interface, this method is specifically documented not to be thread safe (although the godoc in the implementation doesn't copy the warning). The code to make this thread simple to write, but I'm concerned about adding locks that could slow other operations. Do you think it's worth adding locks so that these methods are safe to call concurrently?

```
	// GetTable gets the table with the given name. If the table does not exist, it is first created.
	//
	// WARNING: this method is not thread safe with respect to any other methods in this interface or
	// any methods on any Table objects associated with this store.
	GetTable(name string) (Table, error)
```

common/kvstore/tablestore/table_store_wrapper.go
cody-littley: fixed

common/kvstore/tablestore/table_store_wrapper.go
cody-littley: fixed

common/kvstore/batch.go
cody-littley: `NewBatch` is also a function on the base `Store` object, but I agree this interface isn't really necessary. Removed.

common/kvstore/tablestore/table_store_wrapper.go
cody-littley: The purpose of requiring the caller to provide the `Store` implementation is to support multiple `Store` implementations. Currently this is used in unit tests, where I sometimes use a store implementation that is built on top of an in memory map. Let's discuss this.

cody-littley: Changed based on our slack discussion.

common/kvstore/tablestore/table_store_wrapper.go
jianoaix: I think it needs to, otherwise other methods like GetTableCount (which is supposed to be thread safe) won't be safe as they share the states.

common/kvstore/batch.go
jianoaix: The issue is this interface intends to be used for `TableStore` (like it comments out), but it's now got used to batch operations for `Store` as well.

jianoaix: One idea is to subclass this interface with two types, like (demo not real code):

```
type StoreBatch struct {
        // key type is []byte
	deletes    map[[]byte]struct{}
        ...
}

type TableStoreBatch struct {
        // key type is TableKey
	deletes    map[TableKey]struct{} 
        ...
}
```

It establishes common interface but users doesn't need to use template. Wdyt?

jianoaix: It's a good pattern to have a Base<T> and then subclass it with CaseA and CaseB without template. The interface helps constrain what they can be.

common/kvstore/tablestore/table_store_builder.go
```diff
@@ -0,0 +1,374 @@
...
+	return newTableStore(logger, base, tableMap), nil
+}
+
+// validateSchema loads/initiates the schema version in the metadata table.
+func validateSchema(metadataTable kvstore.Table) error {
+
+	schemaKey := []byte(metadataSchemaVersionKey)
+	onDiskSchemaBytes, err := metadataTable.Get(schemaKey)
+
+	if errors.Is(err, kvstore.ErrNotFound) {
```
jianoaix, line 107: The control below can be made more readable:
```
if err != nil {
    if !errors.Is(err, kvstore.ErrNotFound) {
        return ...
    }
    // handle not found case
    ...

   return nil
}

// Verify schema version
...
```

common/kvstore/tablestore/table_store_builder.go
```diff
@@ -0,0 +1,374 @@
...
+		if err != nil {
+			return fmt.Errorf("error setting schema version in metadata table: %w", err)
+		}
+
+		return nil
+	} else if err == nil {
+		// Verify schema version.
+		onDiskSchema := binary.BigEndian.Uint64(onDiskSchemaBytes)
+		if onDiskSchema != currentSchemaVersion {
+			// In the future if we change schema versions, we may need to write migration code here.
```
jianoaix, line 122: How do you migrate? The data will need to be re-processed.

common/kvstore/tablestore/table_view.go
```diff
@@ -0,0 +1,183 @@
...
+import (
+	"encoding/binary"
+	"github.com/Layr-Labs/eigenda/common/kvstore"
+	"github.com/syndtr/goleveldb/leveldb/iterator"
+	"github.com/syndtr/goleveldb/leveldb/util"
+)
+
+var _ kvstore.Table = &tableView{}
+
+// tableView allows table in a New to be accessed as if it were a Store.
```
jianoaix, line 12: The comment not readable/has typo ("in a New"?, "as it were a Store"?)

common/kvstore/tablestore/table_store_builder.go
```diff
@@ -0,0 +1,374 @@
...
+	"fmt"
+	"github.com/Layr-Labs/eigenda/common/kvstore"
+	"github.com/Layr-Labs/eigenda/common/kvstore/leveldb"
+	"github.com/Layr-Labs/eigenda/common/kvstore/mapstore"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	"math"
+	"sort"
+)
+
+// The table ID reserved for the metadata table.
```
jianoaix, line 15: Comment what's in the metadata table (like what you did for namespaceTable)

common/kvstore/tablestore/table_store_builder.go
```diff
@@ -0,0 +1,374 @@
...
+	LevelDB StoreType = iota
+	// MapStore is an in-memory store. This store does not preserve data across restarts.
+	MapStore
+)
+
+// Create creates a new TableStore instance of the given type. The store will be created at the given path.
+// This method will set up a table for each table name provided, and will drop all tables not in the list.
+// Dropping a table is irreversible and will delete all data in the table, so be very careful not to call
+// this method with table names omitted by mistake.
+func (t StoreType) Create(logger logging.Logger, path string, tables ...string) (kvstore.TableStore, error) {
```
jianoaix, line 46: It looks a more frequent operation will be `Load`, where it's given the path, and will load the existing base Store as well as existing tables.

common/kvstore/tablestore/table_store_wrapper.go
cody-littley: There are now no unsafe methods (due to schema pattern)

common/kvstore/tablestore/table_store_builder.go
cody-littley: refactored

common/kvstore/tablestore/table_store_builder.go
cody-littley: Most likely, a migration will require us to iterate all data in the DB and to change its on disk format. The schema version is included here as a "just in case" sort of thing. I REALLY hope we never have to do this, so we should do our best to design the schema right the first time.

common/kvstore/tablestore/table_view.go
cody-littley:
```
// tableView allows data in a table to be accessed as if it were a Store.
```

This is an artifact of my IDE attempting to automatically refactor things when I changed the name of the method.

common/kvstore/tablestore/table_store_builder.go
cody-littley:
```
// The table ID reserved for the metadata table. The metadata table is used for internal bookkeeping.
// The following data is currently stored in the metadata table:
// - Schema version (in case we ever need to do a schema migration)
// - Table deletion markers used to detect a crash during table deletion and to complete 
//   the deletion when the store is next started.
```

common/kvstore/tablestore/table_store_builder.go
cody-littley: I've renamed this to `Start()`, since that name is accurate for both a new setup and for when we load existing data.

common/kvstore/batch.go
cody-littley: done

common/kvstore/tablestore/table_store_builder.go
```diff
@@ -0,0 +1,381 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error loading namespace table: %w", err)
+	}
+
+	err = handleIncompleteDeletion(logger, base, metadataTable, namespaceTable)
+	if err != nil {
+		return nil, fmt.Errorf("error handling incomplete deletion: %w", err)
+	}
+
+	err = addAndRemoveTables(base, metadataTable, namespaceTable, tableIDMap, tables)
```
jianoaix, line 94: I'd prefer not deleting tables synchronously. After the database running for a while, the deletion can take quite a while. It'd be better to:
- Do lazy deletion: write the table to be deleted to the deletion table
- Run a thread in the background to perform the deletion - and cap the resource it can use

common/kvstore/tablestore/table_store_builder.go
```diff
@@ -0,0 +1,381 @@
...
+	"github.com/Layr-Labs/eigenda/common/kvstore/mapstore"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	"math"
+	"sort"
+)
+
+// The table ID reserved for the metadata table. The metadata table is used for internal bookkeeping.
+// The following data is currently stored in the metadata table:
+//   - Schema version (in case we ever need to do a schema migration)
+//   - Table deletion markers used to detect a crash during table deletion and to complete
```
jianoaix, line 18: If you adopt lazy deletion, then this can be a separate DeletionTable, which records each deleted table as an entry in this table. The background thread work on the DeletionTable and perform GC.

common/kvstore/tablestore/table_store_builder.go
jianoaix: Potential UX issues to check: 1) the users mostly may just want to load existing tables (adding/dropping tables are rare operations), and 2) the user may not know the existing tables, unless they can build this TableStore, which then requires knowing the existing tables (a loop here)

cody-littley: I added a method `Load()` which can be used to load the store without making any modifications to the schema.

common/kvstore/tablestore/table_store_builder.go
cody-littley: We discussed this, repeating it here for the sake of others. This PR is already fairly large, and so we've decided to implement async table deletion as a follow up PR.

common/kvstore/tablestore/table_store_builder.go
```diff
@@ -0,0 +1,402 @@
...
+			"incompatible schema version: code is at version %d, data on disk is at version %d",
+			currentSchemaVersion, onDiskSchema)
+	}
+
+	return nil
+}
+
+// This method adds and removes tables as needed to match the given list of tables. The table ID map is updated
+// to reflect the new state of the tables.
+func addAndRemoveTables(
```
jianoaix, line 164: Potentially more readability if we build this into:
- computeSchemaChanges: returns tablesToAdd and tablesToDrop
- addTables(tablesToAdd)
- dropTables(tablesToDrop)

common/kvstore/tablestore/table_store_builder.go
```diff
@@ -0,0 +1,402 @@
...
+	if err != nil {
+		return fmt.Errorf("error sanity checking namespace table: %w", err)
+	}
+
+	return nil
+}
+
+// Perform sanity checks on the namespace table.
+// This method makes potential logic in the namespace errors fail fast and visibly.
+func sanityCheckNamespaceTable(
```
jianoaix, line 250: Is this for testing/debugging the implementation is correct?

common/kvstore/tablestore/table_store_builder.go
```diff
@@ -0,0 +1,402 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error validating schema: %w", err)
+	}
+
+	tableIDMap, err := loadNamespaceTable(namespaceTable)
+	if err != nil {
+		return nil, fmt.Errorf("error loading namespace table: %w", err)
+	}
+
+	err = handleIncompleteDeletion(logger, base, metadataTable, namespaceTable)
```
jianoaix, line 96: Shouldn't this deletion happen before even loading the namespace table? Otherwise a name/ID that's still pending for deletion may get reused

common/kvstore/tablestore/table_store_builder.go
```diff
@@ -0,0 +1,402 @@
...
+// The table ID reserved for the metadata table. The metadata table is used for internal bookkeeping.
+// The following data is currently stored in the metadata table:
+//   - Schema version (in case we ever need to do a schema migration)
+//   - Table deletion markers used to detect a crash during table deletion and to complete
+//     the deletion when the store is next started.
+const metadataTableID uint32 = math.MaxUint32
+
+// The table ID reserved for the namespace table. Keys in the namespace table are table IDs (uint32)
+// and values are table names (string).
+const namespaceTableID uint32 = math.MaxUint32 - 1
```
jianoaix, line 24: It's helpful to call out in comment that table name and ID can be reused after a table is completely dropped from the TableStore.

common/kvstore/tablestore/table_store_builder.go
```diff
@@ -0,0 +1,402 @@
...
+	if len(tablesToAdd) == 0 {
+		// No tables to add, bail out early.
+		return nil
+	}
+
+	// Determine the table IDs for the new tables to be added.
+	// We want to fill gaps prior to assigning new IDs.
+	newTableIDs := make([]uint32, 0, len(tablesToAdd))
+	nextID := uint32(0)
+	for i := 0; i < len(tablesToAdd); i++ {
```
jianoaix, line 215: Slightly simpler/more readable way to express this code:
```
for len(newTableIDs) < len(tablesToAdd) {
	if _, alreadyUsed := tableIDMap[nextID]; !alreadyUsed {
		newTableIDs = append(newTableIDs, nextID)
	}
	nextID++
}
```

jianoaix (commented):
Almost there!

common/kvstore/tablestore/table_store_builder.go
```diff
@@ -0,0 +1,402 @@
...
+// to reflect the new state of the tables.
+func addAndRemoveTables(
+	base kvstore.Store,
+	metadataTable kvstore.Table,
+	namespaceTable kvstore.Table,
+	tableIDMap map[uint32]string,
+	currentTables []string) error {
+
+	// Determine which tables to keep and which to drop.
+	originalTablesSet := make(map[string]bool)
```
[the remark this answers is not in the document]
cody-littley: Done. One of the things that gets on my nerves with golang is the absence of a set implementation. Sure, you can mimic it with a map. It's just syntactically ugly.

common/kvstore/tablestore/table_store_builder.go
cody-littley: done

common/kvstore/tablestore/table_store_builder.go
cody-littley: This is sanity check code intended to make noise if a bug in the code results in inconsistencies between `tableIDMap`, `namespaceTable`, and the user provided schema. I intended this check to run even in production environments, since if the tables are corrupted we will have really big problems.

That being said, if you don't think it belongs I'm willing to remove it.

common/kvstore/tablestore/table_store_builder.go
cody-littley: You are correct that the partially deleted table may already be inside the namespace table. This edge case is handled by `handleIncompleteDeletion()`, which will perform the necessary deletion within the namespace table (last line of `handleIncompleteDeletion()` calls `dropTable()`, which deletes the entry from the namespace table).

common/kvstore/tablestore/table_store_builder.go
cody-littley:
```
// The table ID reserved for the namespace table. Keys in the namespace table are table IDs (uint32)
// and values are table names (string). Although no two tables are permitted to have share a name or a table ID,
// once a table is dropped, future tables may be instantiated with the same name and the table ID may be reused.
const namespaceTableID uint32 = math.MaxUint32 - 1
```

common/kvstore/tablestore/table_store_builder.go
cody-littley: Yes, that's a lot cleaner. Done.

common/kvstore/tablestore/table_store_builder.go
jianoaix: If there is a pending deletion, the `loadNamespaceTable` will include that deleted table in the returned `tableIDMap` (because that table may be still in the namespace table, haven't fully deleted), which is not desired

common/kvstore/tablestore/table_store_builder.go
jianoaix: Sounds fine to keep then

common/kvstore/tablestore/table_store_builder.go
cody-littley: I walked this code with a debugger. It turns out this was accidentally working, since the method `addAndRemoveTables()` deletes the spurious table ID. There is a unit test that is sensitive to spurious entries in the table ID map, and that was not failing. 

In order to make this more understandable and less easy to accidentally break, I've moved the loading of the table ID map until after `handleIncompleteDeletion()`.

common/kvstore/Batch.go
```diff
@@ -0,0 +1,14 @@
+package kvstore
+
+// Batch is a collection of key / value pairs that can be written atomically to a database.
```
jianoaix, line 3: nit: `can be written` -> `will be written` atomically?

Mention that the Batch is not thread safe

jianoaix (approved):

PR #779, opened 2024-09-30
disperser meterer for payments

hopeyen:

PoC inside disperser for handling payments

summary of changes 
- Create a meterer package inside the disperser that can be used to replace `ratelimiter` that validate the requests before storing them
- Add mocked interfaces for contracts
- Add dynamo DB for managing offchain state
- Using simple data types (first `u64` before worrying about `u128`)
- Unit tests
- Add flag for payments and take precedence over ratelimit when disperser serve requests
  - ineffective as blob request header must be updated first

unit tests

- reservations
	- test for invalid signatures
	- test for invalid bin index
	- test for non existent accounts
	- if blob size > bin, request is rejected 
	- blob size <= bin, given total throughput is still under reservation bin limit, requests is okay (need to add handling and tests for less than chargeable size)
	- first overflow is okay, check that bin+2 gets charged
	- following overflows for the same bin is rejected
	- --> to add periodic bin cleanup/removal
- on demand
	- test for invalid signatures
	- test for invalid bin index
	- test for non existent accounts
	- --> to add chargeable and ensure that the chargeable unit is recorded when blob is smaller 
	- if blobSize > cumulative payments, request is rejected
	- if blob size > global rate limit, request is rejected
	- for everything in between, if globally okay, things should work
	- reject requests exceeding the cumulative payment limit

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests


disperser/meterer/meterer.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	return nil
+}
+
+// ValidateBinIndex checks if the provided bin index is valid
+func (m *Meterer) ValidateBinIndex(blobHeader BlobHeader, reservation *ActiveReservation) bool {
+	currentBinIndex := GetCurrentBinIndex()
+	// Valid bin indexes are either the current bin or the previous bin
+	if (blobHeader.BinIndex != currentBinIndex && blobHeader.BinIndex != (currentBinIndex-1)) || (reservation.StartEpoch > blobHeader.BinIndex || blobHeader.BinIndex > reservation.EndEpoch) {
```
mooselumph, line 151: Does it make sense for Epochs to use the same units as Bins?

disperser/meterer/meterer.go
```diff
@@ -0,0 +1,273 @@
...
+		m.OffchainStore.UpdateReservationBin(ctx, blobHeader.AccountID, uint64(blobHeader.BinIndex+2), newUsage-binLimit)
+		return nil
+	}
+	return fmt.Errorf("Overflow usage exceeds bin limit")
+}
+
+// GetCurrentBinIndex returns the current bin index based on time
+func GetCurrentBinIndex() uint64 {
+	currentTime := time.Now().Unix()
+	binInterval := 375 // Interval that allows 3 32MB blobs at minimal throughput
```
mooselumph, line 189: Should this be configuraable? What are the units?

disperser/meterer/meterer.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	// check against the onchain state for overflow (TODO: update to real provider calls)
+	blockNumber := uint(0)
+	reservation, err := m.ChainState.MockedGetActiveReservationByAccount(ctx, blockNumber, blobHeader.AccountID)
+	if err != nil {
+		return fmt.Errorf("failed to get active reservation by account: %w", err)
+	}
+	if newUsage <= 2*binLimit && blobHeader.BinIndex+2 <= reservation.EndEpoch {
+		m.OffchainStore.UpdateReservationBin(ctx, blobHeader.AccountID, uint64(blobHeader.BinIndex+2), newUsage-binLimit)
```
mooselumph, line 180: Note: this is not a race condition because previous logic ensures this can only happen in a single thread.

disperser/meterer/offchain_store.go
```diff
@@ -0,0 +1,269 @@
...
+
+	return nil
+}
+
+// relevant on-demand payment records: previous cumulative payment, next cumulative payment, blob size of next payment
+func (s *OffchainStore) GetRelevantOnDemandRecords(ctx context.Context, accountID string, cumulativePayment uint64) (uint64, uint64, uint32, error) {
+	result, err := s.dynamoClient.QueryIndex(ctx, s.onDemandTableName, "AccountIDIndex", "AccountID = :account", commondynamodb.ExpresseionValues{
+		":account": &types.AttributeValueMemberS{
+			Value: accountID,
+		}})
```
mooselumph, line 228: I think we should be able to configure a sort key associated with the `CumulativePayment` field and directly retrieve the records with cumulative payment above and below the provided payment.

disperser/meterer/meterer.go
```diff
@@ -0,0 +1,273 @@
...
+	return nil
+}
+
+// ValidateBinIndex checks if the provided bin index is valid
+func (m *Meterer) ValidateGlobalBinIndex(blobHeader BlobHeader) (uint64, error) {
+	// Deterministic function: local clock -> index (1second intervals)
+	currentBinIndex := uint64(time.Now().Unix())
+
+	// Valid bin indexes are either the current bin or the previous bin (allow this second or prev sec)
+	if blobHeader.BinIndex != currentBinIndex && blobHeader.BinIndex != (currentBinIndex-1) {
```
mooselumph, line 253: For global rate limit, I think we can argue that we don't need to include the BinIndex within the header, but just use the receipt time directly.

disperser/meterer/meterer.go
```diff
@@ -0,0 +1,273 @@
...
+	if newUsage <= binLimit {
+		return nil
+	} else if newUsage-blobHeader.BlobSize >= binLimit {
+		// metered usage before updating the size already exceeded the limit
+		return fmt.Errorf("Bin has already been overflowed")
+	}
+
+	// check against the onchain state for overflow (TODO: update to real provider calls)
+	blockNumber := uint(0)
+	reservation, err := m.ChainState.MockedGetActiveReservationByAccount(ctx, blockNumber, blobHeader.AccountID)
```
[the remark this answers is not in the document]
hopeyen: updated!

disperser/meterer/meterer.go
hopeyen: I'm open to change this for easier usage. What about using start and end timestamps instead of epochs?

disperser/meterer/meterer.go
hopeyen: yes! changed this to use reservationWindow config and added an indication in the help string that this is in seconds

disperser/meterer/meterer.go
hopeyen: makes sense, in this case we don't need to validate the bin index, maybe not even receipt time, but the time when global rate gets checked (since global rate limit check happens after payment validations, we avoid failing on cases where payment validation ever gets to more than 2 seconds)

common/aws/dynamodb/client.go
```diff
@@ -156,6 +157,51 @@ func (c *Client) UpdateItem(ctx context.Context, tableName string, key Key, item
 	return resp.Attributes, err
 }
 
+func (c *Client) UpdateItemIncrement(ctx context.Context, tableName string, key Key, item Item) (Item, error) {
+	update := expression.UpdateBuilder{}
+	for itemKey, itemValue := range item {
+		if _, ok := key[itemKey]; ok {
+			// Cannot update the key
+			continue
+		}
```
mooselumph, line 166: I realize this is using the template above, but it's unclear to me why this is failing silently.

common/aws/dynamodb/client.go
```diff
@@ -156,6 +157,51 @@ func (c *Client) UpdateItem(ctx context.Context, tableName string, key Key, item
...
+		// fmt.Println("updating item", itemKey, itemValue)
+		// ADD numeric values
+		if n, ok := itemValue.(*types.AttributeValueMemberN); ok {
+			// update = update.Add(expression.Name(itemKey), expression.Value(n.Value))
+			// update = update.Add(expression.Name(itemKey), expression.Value(n.Value))
```
mooselumph, line 171: Remove?

disperser/meterer/offchain_store.go
```diff
@@ -0,0 +1,269 @@
...
+
+	binUsageValue, err := strconv.ParseUint(binUsageAttr.Value, 10, 32)
+	if err != nil {
+		return 0, err
+	}
+
+	return binUsageValue, nil
+}
+
+func (s *OffchainStore) FindReservationBin(ctx context.Context, accountID string, binIndex uint64) (*ReservationBin, error) {
```
mooselumph, line 134: Is this used?

disperser/meterer/offchain_store.go
```diff
@@ -0,0 +1,269 @@
...
+	err = attributevalue.UnmarshalMap(result, &reservation)
+	if err != nil {
+		return nil, err
+	}
+
+	return &reservation, nil
+}
+
+// Find all reservation bins for a given account
+func (s *OffchainStore) FindReservationBins(ctx context.Context, accountID string) ([]ReservationBin, error) {
```
mooselumph, line 159: Is this used?

mooselumph (commented):
Some more comments!

disperser/meterer/onchain_state.go
```diff
@@ -0,0 +1,87 @@
...
+)
+
+/* HEAVILY MOCKED */
+// PaymentAccounts (For reservations and on-demand payments)
+
+type TokenAmount uint64 // TODO: change to uint128
+
+// OperatorInfo contains information about an operator which is stored on the blockchain state,
+// corresponding to a particular quorum
+type ActiveReservation struct {
```
mooselumph, line 17: Note: neeed to add `QuorumNumbers` as discussed

disperser/meterer/types.go
```diff
@@ -0,0 +1,243 @@
...
+	AdversaryThreshold uint32
+	QuorumThreshold    uint32
+}
+
+// BlobHeader represents the header information for a blob
+type BlobHeader struct {
+	// Existing fields
+	Commitment       core.G1Point
+	DataLength       uint32
+	BlobQuorumParams []BlobQuorumParam
```
mooselumph, line 30: This can just be a list of QuorumNumbers, per the new header spec.

disperser/meterer/types.go
```diff
@@ -0,0 +1,243 @@
...
+
+	// New fields
+	Version   uint32
+	AccountID string
+	Nonce     uint32 // use nonce to prevent duplicate payments in the same reservation window
+	BinIndex  uint64
+
+	Signature []byte
+	BlobSize  uint32
```
mooselumph, line 39: This is redundant with DataLength. We should decide on canonical naming.

disperser/meterer/types.go
```diff
@@ -0,0 +1,243 @@
...
+// Protocol defines parameters: epoch length and rate-limit window interval
+type Reservation struct {
+	dataRate    uint32 // bandwith being reserved
+	startEpoch  uint32 // index of epoch where reservation begins
+	endEpoch    uint32 // index of epoch where reservation ends
+	quorumSplit []byte // each byte is a percentage at the corresponding quorum index
+}
+
+// Protocol defines parameters: FixedFeePerByte; fine to leave global rate-limit offchain atm
+type OnDemand struct {
```
mooselumph, line 62: Is this used?

disperser/meterer/types.go
```diff
@@ -0,0 +1,243 @@
...
+	// Existing fields
+	Commitment       core.G1Point
+	DataLength       uint32
+	BlobQuorumParams []BlobQuorumParam
+
+	// New fields
+	Version   uint32
+	AccountID string
+	Nonce     uint32 // use nonce to prevent duplicate payments in the same reservation window
+	BinIndex  uint64
```
mooselumph, line 36: It seems like `uint32` would be sufficient for this?

common/aws/dynamodb/client.go
hopeyen: Apparently dynamoDB doesn't let user update the primary key or sort key of an existing item, so if the current itemKey is part of the item's key, we simply move on to the next attribute in the item map. I think this is failing silently because we want to operate on other fields anyway.

some 3 alternative ways that still ensure consistency: 
1) a loop to check the keys (and return err) before the updating loop, 
2) do the same as now, but revert the changes later on if we cannot update the key
3) do the same as now, and return a list of keys that couldn't get updated

disperser/meterer/offchain_store.go
hopeyen: Nope atm, it was intended for an API, which i now realized I've yet to add

disperser/meterer/offchain_store.go
hopeyen: updated!

disperser/meterer/types.go
hopeyen: went with `DataLength`

disperser/meterer/types.go
hopeyen: updated to be more memory efficient; I'm assuming we don't need to worry about an index in the unit of seconds (previously global bin index is in seconds so 2^32 in unix let this go up to Jan 19, 2038, but with that being removed, uint32 is fine for reservation bin indices

I also updated start & end epochs to use `uint32`, as well as `PricePerChargeable` (2^64 gwei ~= 18M Eth; uint32 => ~4ETH), `MinChargeableSize` (in bytes, this is ~4GB/s), and `ReservationWindow` (136years but uint16 is only 18hours if we are counting in seconds; we could instead use hour as window unit and allow for 7years). 

Left `GlobalBytesPerSecond` with `uint64` for the aspiration of supporting rate-limit greater than 4GB/s network-wide ✨

PR #780, opened 2024-10-02
Style: improve api clients comments

samlaf:
Small improvements to comments/docs for api clients.
Had some fun in the plane while reading the code.
Added some TODOs in a few places which might be worth checking through, perhaps we can solve some of them right away @bxue-l2

### DIFF
--- a/api/clients/codecs/blob_codec.go
+++ b/api/clients/codecs/blob_codec.go
@@ -30,7 +30,9 @@ func GenericDecodeBlob(data []byte) ([]byte, error) {
 	if len(data) <= 32 {
 		return nil, fmt.Errorf("data is not of length greater than 32 bytes: %d", len(data))
 	}
-	// TODO: why [1]? What's in [0]?
+	// version byte is stored in [1], because [0] is always 0 to ensure the codecBlobHeader is a valid bn254 element
+	// see https://github.com/Layr-Labs/eigenda/blob/master/api/clients/codecs/default_blob_codec.go#L21
+	// TODO: we should prob be working over a struct with methods such as GetBlobEncodingVersion() to prevent index errors
 	version := BlobEncodingVersion(data[1])
 	codec, err := BlobEncodingVersionToCodec(version)
 	if err != nil {
--- a/api/clients/codecs/blob_codec.go
+++ b/api/clients/codecs/blob_codec.go
@@ -6,9 +6,9 @@ import (
 
 type BlobEncodingVersion byte
 
-// All blob encodings are IFFT'd before being dispersed
 const (
-	// This minimal blob encoding includes a version byte, a length uint32, and 31 byte field element mapping.
+	// This minimal blob encoding contains a 32 byte header = [0x00, version byte, uint32 len of data, 0x00, 0x00,...]
+	// followed by the encoded data [0x00, 31 bytes of data, 0x00, 31 bytes of data,...]
 	DefaultBlobEncoding BlobEncodingVersion = 0x0
 )
 
@@ -30,6 +30,7 @@ func GenericDecodeBlob(data []byte) ([]byte, error) {
 	if len(data) <= 32 {
 		return nil, fmt.Errorf("data is not of length greater than 32 bytes: %d", len(data))
 	}
+	// TODO: why [1]? What's in [0]?
 	version := BlobEncodingVersion(data[1])
 	codec, err := BlobEncodingVersionToCodec(version)
 	if err != nil {
@@ -38,7 +39,7 @@ func GenericDecodeBlob(data []byte) ([]byte, error) {
 
 	data, err = codec.DecodeBlob(data)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("unable to decode blob: %w", err)
 	}
 
 	return data, nil
--- a/api/clients/config.go
+++ b/api/clients/config.go
@@ -27,6 +27,7 @@ type EigenDAClientConfig struct {
 	WaitForFinalization bool
 
 	// The quorum IDs to write blobs to using this client. Should not include default quorums 0 or 1.
+	// TODO: should we change this to core.QuorumID instead? https://github.com/Layr-Labs/eigenda/blob/style--improve-api-clients-comments/core/data.go#L18
 	CustomQuorumIDs []uint
 
 	// Signer private key in hex encoded format. This key is currently purely used for authn/authz on the disperser.
--- a/api/clients/config.go
+++ b/api/clients/config.go
@@ -8,22 +8,32 @@ import (
 )
 
 type EigenDAClientConfig struct {
-	// RPC is the HTTP provider URL for the Data Availability node.
+	// RPC is the HTTP provider URL for the EigenDA Disperser
 	RPC string
 
-	// The total amount of time that the client will spend waiting for EigenDA to confirm a blob
+	// Timeout used when making dispersals to the EigenDA Disperser
+	// TODO: we should change this param as its name is quite confusing
+	ResponseTimeout time.Duration
+
+	// The total amount of time that the client will spend waiting for EigenDA
+	// to confirm a blob after it has been dispersed
+	// Note that reasonable values for this field will depend on the value of WaitForFinalization.
 	StatusQueryTimeout time.Duration
 
 	// The amount of time to wait between status queries of a newly dispersed blob
 	StatusQueryRetryInterval time.Duration
 
-	// The total amount of time that the client will waiting for a response from the EigenDA disperser
-	ResponseTimeout time.Duration
+	// If true, will wait for the blob to finalize, if false, will wait only for the blob to confirm.
+	WaitForFinalization bool
 
 	// The quorum IDs to write blobs to using this client. Should not include default quorums 0 or 1.
 	CustomQuorumIDs []uint
 
-	// Signer private key in hex encoded format. This key should not be associated with an Ethereum address holding any funds.
+	// Signer private key in hex encoded format. This key is currently purely used for authn/authz on the disperser.
+	// For security, it should not be associated with an Ethereum address holding any funds.
+	// This might change once we introduce payments.
+	// OPTIONAL: this value is optional, and if set to "", will result in a read-only eigenDA client,
+	// that can retrieve blobs but cannot disperse blobs.
 	SignerPrivateKeyHex string
 
 	// Whether to disable TLS for an insecure connection when connecting to a local EigenDA disperser instance.
@@ -37,9 +47,6 @@ type EigenDAClientConfig struct {
 	// the commitment. With this mode disabled, you will need to supply the entire blob to perform a verification
 	// that any part of the data matches the KZG commitment.
 	DisablePointVerificationMode bool
-
-	// If true, will wait for the blob to finalize, if false, will wait only for the blob to confirm.
-	WaitForFinalization bool
 }
 
 func (c *EigenDAClientConfig) CheckAndSetDefaults() error {
--- a/api/clients/disperser_client.go
+++ b/api/clients/disperser_client.go
@@ -20,6 +20,9 @@ import (
 type Config struct {
 	Hostname          string
 	Port              string
+	// BlobDispersal Timeouts for both authenticated and unauthenticated dispersals
+	// GetBlobStatus and RetrieveBlob timeouts are hardcoded to 60seconds
+	// TODO: do we want to add config timeouts for those separate requests?
 	Timeout           time.Duration
 	UseSecureGrpcFlag bool
 }
--- a/api/clients/eigenda_client.go
+++ b/api/clients/eigenda_client.go
@@ -155,7 +155,7 @@ func (m EigenDAClient) putBlob(ctx context.Context, rawData []byte, resultChan c
 		customQuorumNumbers[i] = uint8(e)
 	}
 	// disperse blob
-	// TODO: can we consider passing a requestID directly to requests, to get idempotency and tracing?
+	// TODO: would be nice to add a trace-id key to the context, to be able to follow requests from batcher->proxy->eigenda
 	blobStatus, requestID, err := m.Client.DisperseBlobAuthenticated(ctx, data, customQuorumNumbers)
 	if err != nil {
 		errChan <- fmt.Errorf("error initializing DisperseBlobAuthenticated() client: %w", err)
--- a/api/clients/eigenda_client.go
+++ b/api/clients/eigenda_client.go
@@ -150,7 +150,6 @@ func (m EigenDAClient) putBlob(ctx context.Context, rawData []byte, resultChan c
 		return
 	}
 
-	// TODO: should we just use uint8s directly in the config?
 	customQuorumNumbers := make([]uint8, len(m.Config.CustomQuorumIDs))
 	for i, e := range m.Config.CustomQuorumIDs {
 		customQuorumNumbers[i] = uint8(e)
--- a/api/clients/eigenda_client.go
+++ b/api/clients/eigenda_client.go
@@ -16,13 +16,19 @@ import (
 	"github.com/ethereum/go-ethereum/log"
 )
 
+// IEigenDAClient is a wrapper around the DisperserClient interface which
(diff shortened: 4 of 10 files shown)


api/clients/codecs/blob_codec.go
```diff
@@ -30,6 +30,7 @@ func GenericDecodeBlob(data []byte) ([]byte, error) {
 	if len(data) <= 32 {
 		return nil, fmt.Errorf("data is not of length greater than 32 bytes: %d", len(data))
 	}
+	// TODO: why [1]? What's in [0]?
```
samlaf, line 33: @bxue-l2 you prob know this

api/clients/eigenda_client.go
```diff
@@ -16,13 +16,19 @@ import (
...
 	GetBlob(ctx context.Context, batchHeaderHash []byte, blobIndex uint32) ([]byte, error)
 	PutBlob(ctx context.Context, txData []byte) (*grpcdisperser.BlobInfo, error)
 	GetCodec() codecs.BlobCodec
 }
 
+// EigenDAClient is a wrapper around the DisperserClient which
+// encodes blobs before dispersing them, and decodes them after retrieving them.
 type EigenDAClient struct {
+	// TODO: all of these should be private, to prevent users from using them directly,
+	// which breaks encapsulation and makes it hard for us to do refactors or changes
```
samlaf, line 31: this I personally think is a refactor worth doing... but it might break some client code... :\

api/clients/eigenda_client.go
```diff
@@ -84,27 +93,30 @@ func (m EigenDAClient) GetCodec() codecs.BlobCodec {
...
 	data, err := m.Client.RetrieveBlob(ctx, batchHeaderHash, blobIndex)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("could not retrieve blob: %w", err)
 	}
 
 	if len(data) == 0 {
+		// TODO: explain when/why/how this can happen
 		return nil, fmt.Errorf("blob has length zero")
 	}
```
samlaf, line 106: @bxue-l2 do you know?

api/proto/disperser/disperser.proto
```diff
@@ -78,12 +78,12 @@ message AuthenticationData {
...
 message DisperseBlobRequest {
 	// The data to be dispersed.
-	// The size of data must be <= 2MiB. Every 32 bytes of data chunk is interpreted as an integer in big endian format
+	// The size of data must be <= 16MiB. Every 32 bytes of data chunk is interpreted as an integer in big endian format
 	// where the lower address has more significant bits. The integer must stay in the valid range to be interpreted
 	// as a field element on the bn254 curve. The valid range is 
 	// 0 <= x < 21888242871839275222246405745257275088548364400416034343698204186575808495617
-	// containing slightly less than 254 bits and more than 253 bits. If any one of the 32 bytes chunk is outside the range, 
-	// the whole request is deemed as invalid, and rejected. 
+	// which is a 254 bit number (meaning the first 2 bits of each chunk must always be 00).
```
jianoaix, line 85: This isn't accurate though, `21888242871839275222246405745257275088548364400416034343698204186575808495617` is smaller than `2^254-1` (i.e. the original comment is accurate that it doesn't have 254 bits).

Also shall we not use `chunk` here since it's easy to confuse with the encoding chunks (which will become more user visible in the future)?

api/clients/eigenda_client.go
```diff
@@ -137,11 +150,13 @@ func (m EigenDAClient) putBlob(ctx context.Context, rawData []byte, resultChan c
 		return
 	}
 
+	// TODO: should we just use uint8s directly in the config?
```
jianoaix, line 153: Or change it to `core.QuorumID`?

api/clients/config.go
```diff
@@ -8,22 +8,32 @@ import (
 )
 
 type EigenDAClientConfig struct {
-	// RPC is the HTTP provider URL for the Data Availability node.
+	// RPC is the HTTP provider URL for the EigenDA Disperser
 	RPC string
 
-	// The total amount of time that the client will spend waiting for EigenDA to confirm a blob
+	// Timeout used when making dispersals to the EigenDA Disperser
```
jianoaix, line 14: This reads like it's waiting for the gRPC call to finish DisperseBlob(), but the old comment says it's waiting for confirmation. Is this intended change?

bxue-l2: I think it is indeed just the timeout for dispersal call, https://github.com/Layr-Labs/eigenda/blob/master/api/clients/disperser_client.go#L148

api/clients/config.go
```diff
@@ -8,22 +8,32 @@ import (
...
+	// TODO: we should change this param as its name is quite confusing
+	ResponseTimeout time.Duration
```
bxue-l2, line 16: agree, how about DispersalTimeout?

api/clients/disperser_client.go
```diff
@@ -20,6 +20,9 @@ import (
 type Config struct {
 	Hostname          string
 	Port              string
+	// BlobDispersal Timeouts for both authenticated and unauthenticated dispersals
```
bxue-l2, line 23: v2 might have an architecture change on how authentication will be done. We can revisit at that time. For now, the latency is dominated by data size, so timeout should be identical I think

api/clients/disperser_client.go
```diff
@@ -20,6 +20,9 @@ import (
...
+	// GetBlobStatus and RetrieveBlob timeouts are hardcoded to 60seconds
+	// TODO: do we want to add config timeouts for those separate requests?
```
bxue-l2, line 25: configurable is better, more flexible to explore fallback option

api/clients/eigenda_client.go
bxue-l2: agree this should be private. From my experience, it is difficult to understand data semantics in each stage or encoding and decoding.

api/clients/eigenda_client.go
```diff
@@ -84,27 +93,30 @@ func (m EigenDAClient) GetCodec() codecs.BlobCodec {
 // data, which is necessary for generating KZG proofs for data's correctness.
 // The function handles potential errors during blob retrieval, data length
 // checks, and decoding processes.
+// TODO: should we use a pointer receiver instead, to prevent unnecessary copying of the EigenDAClient struct?
```
bxue-l2, line 96: what do you mean by this?

api/clients/eigenda_client.go
```diff
@@ -137,11 +150,13 @@ func (m EigenDAClient) putBlob(ctx context.Context, rawData []byte, resultChan c
 		return
 	}
 
+	// TODO: should we just use uint8s directly in the config?
 	customQuorumNumbers := make([]uint8, len(m.Config.CustomQuorumIDs))
 	for i, e := range m.Config.CustomQuorumIDs {
 		customQuorumNumbers[i] = uint8(e)
 	}
 	// disperse blob
+	// TODO: can we consider passing a requestID directly to requests, to get idempotency and tracing?
```
bxue-l2, line 159: requestID is currently computed by disperser, that uses timestamp taken at the server side

bxue-l2 (dismissed):

api/clients/codecs/blob_codec.go
samlaf: Thanks, added explanation comment: [bb1eb0e](https://github.com/Layr-Labs/eigenda/pull/780/commits/bb1eb0e4c7d0d9df17ae8fd8eca2900e9f190ef5)

api/clients/eigenda_client.go
samlaf: Don't understand your comment, why are you mentioning data semantics here?
And are you saying I should change the fields to private right now? Are we not scared of breaking user code that might be using these?

api/clients/eigenda_client.go
samlaf: rip, @jianoaix @ian-shim any ideas here?

api/proto/disperser/disperser.proto
samlaf: Took me a while to figure this one out haha. Doesn't matter that it's < 2^254-1, what matters is that it's > 2^253-1, and so actually needs 254 bits.

The number is 254 bits according to https://hackmd.io/@jpw/bn254 (which also explains why its called bn254 and not bn253). Here's also a python script showing that it's a 254 bit number:
```
number = 21888242871839275222246405745257275088548364400416034343698204186575808495617
binary_representation = bin(number)

# Count the number of bits (excluding the '0b' prefix)
bit_count = len(binary_representation) - 2

print(f"The number is: {number}")
print(f"Binary representation: {binary_representation}")
print(f"Number of bits: {bit_count}")
```

samlaf: What's a better word that you would recommend instead of `chunk`? @bxue-l2 maybe you have an idea since you initially wrote this comment I believe?

api/clients/eigenda_client.go
samlaf: moved the TODO comment to the config and mentioned `core.QuorumID`: [839ffbb](https://github.com/Layr-Labs/eigenda/pull/780/commits/839ffbb7d800e6f90ad6bbc4fe061eca6cd3e759)
Do we have a policy regarding breaking changes like this? Are we fine with just pulling the trigger right now?

api/clients/eigenda_client.go
```diff
@@ -84,27 +93,30 @@ func (m EigenDAClient) GetCodec() codecs.BlobCodec {
...
 	decodedData, err := m.Codec.DecodeBlob(data)
 	if err != nil {
-		return nil, fmt.Errorf("error getting blob: %w", err)
+		return nil, fmt.Errorf("error decoding blob: %w", err)
 	}
 
 	return decodedData, nil
 }
 
 // PutBlob encodes and writes a blob to EigenDA, waiting for it to be finalized
```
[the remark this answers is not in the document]
samlaf: [4c8a2d7](https://github.com/Layr-Labs/eigenda/pull/780/commits/4c8a2d7b4ba6ef5ca0c8f0bc650e2ccaf0d2dfc5)

api/clients/config.go
samlaf: The old comment was not accurate... not sure if/when it was changed.

api/proto/disperser/disperser.proto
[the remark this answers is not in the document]
samlaf: [6b78c52](https://github.com/Layr-Labs/eigenda/pull/780/commits/6b78c526afb2f451149007cc80a4859b9269a504)

samlaf: EDIT: based on bowen's comment below, I think I understand what you guys are saying now. Still think my comment is more accurate and the original one weird/wrong (a number cannot have more than 253 bits and less than 254 bits, it has a fixed number of bits, and could happen to have 254 bits).
Updated the comment order to reflect you guys' confusion, hopefully this helps: [6b78c52](https://github.com/Layr-Labs/eigenda/pull/780/commits/6b78c526afb2f451149007cc80a4859b9269a504)

api/clients/config.go
samlaf: I'm fine with doing that, but as I mentioned in the other comment https://github.com/Layr-Labs/eigenda/pull/780#discussion_r1806364950, do we have a policy for these breaking changes? Or are we fine with just doing it now?

api/clients/disperser_client.go
samlaf: sgtm. will resolve this as there's already a ton of comments on this PR. But let me know if you wanted me to add a TODO comment in there or something.

api/clients/disperser_client.go
samlaf: agree but I'm getting tired of this PR haha, let's do it later, sorry, running out of steam.

api/clients/eigenda_client.go
samlaf: `func (m *EigenDAClient) PutBlob` instead of `func (m EigenDAClient) PutBlob`

api/proto/disperser/disperser.proto
```diff
@@ -78,12 +78,12 @@ message AuthenticationData {
 
 message DisperseBlobRequest {
```
jianoaix, line 79: The other copy: https://github.com/Layr-Labs/eigenda/blob/master/api/proto/disperser/v2/disperser_v2.proto#L25

api/clients/eigenda_client.go
jianoaix: Empty blobs should have been rejected from entering the system: https://github.com/Layr-Labs/eigenda/blob/master/disperser/apiserver/server.go#L930

api/proto/disperser/disperser.proto
pschork: > (meaning the first 2 bits of each chunk must always be 00)

Pedantically, this is not true. Disperser will happily encode & disperse a blob that does not prefix chunks with 00 as long as all 32 byte chunks are in range. The 00 prefix is just the easiest way to achieve lossless chunk encoding.

samlaf: That's not pedantic that's important info! So @bxue-l2 was the original comment wrong then? Disperser doesn't check range?

samlaf: @pschork maybe wording is getting everyone confused? By "chunk" in here we are talking about 32byte field elements. If those are not prefixed by 00, then they are surely a number > the max allowed FE, and so therefore wouldn't be in range, right?

pschork: When encoding, if your chunk size is **32 bytes** and the current chunk is within range, you don't need to pad it, but if given chunk exceeds the range then you need to truncate the chunk and pad it with zero to fit within range. The problem is when do this, you won't be able to decode the encoding because you wont know which chunks were or were not padded. To solve this you set the chunk size to **31 bytes** and reserve the *1 byte* as a prefix with `\x00` so that, when decoding, you can identify the start of a chunk. That is why each chunk (typically) starts with 00.

pschork: https://github.com/Layr-Labs/eigenda/blob/master/encoding/utils/codec/codec.go

jianoaix: please avoid using chunk here, it'll quickly get overloaded and confusing.

I think we don't need to mention it's prefixed with 00 (it's not precise anyway), we just need to mention each 32 bytes should fall in that value range so it's a valid field element.

api/clients/eigenda_client.go
bxue-l2: just from the developers' view, it tool me a while to understand the intermediate phases. If someone else would use it, I doubted it would be correct

api/clients/eigenda_client.go
bxue-l2: make sense

api/clients/eigenda_client.go
samlaf: Thanks: updated [5bb9efc](https://github.com/Layr-Labs/eigenda/pull/780/commits/5bb9efc3b69dfde838eb1a5bcbed41b34f191aa1)

api/proto/disperser/disperser.proto
samlaf: [795868e](https://github.com/Layr-Labs/eigenda/pull/780/commits/795868e27c83f5701419f5e42c72237b64e4de26)

api/proto/disperser/disperser.proto
samlaf: [9d46d92](https://github.com/Layr-Labs/eigenda/pull/780/commits/9d46d9283af20fbf6c2c5dbe48f1cdcc6c6e23e2)

api/clients/config.go
samlaf: Note: we decided to do all the breaking changes in a future PR.

api/clients/eigenda_client.go
samlaf: Note: we decided to do all the breaking changes in a future PR.

api/clients/eigenda_client.go
samlaf: Note: we decided to do all the breaking changes in a future PR.

bxue-l2 (approved):
lg

PR #781, opened 2024-10-02
V2 Blob Verification

0x0aa0:
- V2
- consolidates structs into common interface
- moves V1 thresholds to ThresholdRegistry while maintaining DASM interface
- renames RollupUtils to BlobVerificationUtils which implements V2 and batched util for V1
- introduces chain abstract immutable BlobVerifier contract to interact with BlobVerificationUtils
- adds relay registry


contracts/src/libraries/EigenDABlobVerificationUtils.sol
```diff
@@ -24,6 +25,29 @@ library EigenDABlobVerificationUtils {
...
         bytes quorumIndices;
     }
+
+    struct SignedCertificate {
+        BlobCertificate blobCertificate;
+        Attestation nonSignerStakesAndSignature;
+    }
+
+    struct BlobCertificate {
+        bytes blobKey;
```
ian-shim, line 35: We've removed blobKey from cert in the core struct because it can be derived from blob header. Unless we need `blobKey` in the verifier logic, we can probably remove it here too

contracts/src/interfaces/IEigenDABlobVerifier.sol
```diff
@@ -29,5 +29,12 @@ interface IEigenDABlobVerifier is IEigenDAThresholdRegistry {
...
-    function verifyBlobV2() external view;
-}
\ No newline at end of file
+    function verifyBlobV2(
+        EigenDABlobVerificationUtils.SignedCertificate calldata signedCertificate
+    ) external view;
+
+    function verifyBlobV2(
+        EigenDABlobVerificationUtils.SignedCertificate calldata signedCertificate,
+        bytes calldata additionalQuorumNumbersRequired
```
ian-shim, line 38: blob header (contained in blob cert) contains list of all quorums. Can we infer quorum numbers from the blob header? Then we can just have one interface above

contracts/src/core/EigenDAThresholdRegistry.sol
```diff
@@ -0,0 +1,66 @@
+// SPDX-License-Identifier: MIT
```
ian-shim, line 1: Would the `blob version -> {coding rate, reconstruction threshold, confirmation threshold, num chunks}` map be defined here?
cc @mooselumph

contracts/src/libraries/EigenDABlobVerificationUtils.sol
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    struct SignedCertificate {
+        BlobCertificate blobCertificate;
+        Attestation nonSignerStakesAndSignature;
+    }
+
+    struct BlobCertificate {
+        bytes blobKey;
+        IEigenDAServiceManager.BlobHeader blobHeader;
```
ian-shim, line 36: Kind of janky, but the v2 blob header looks different from the existing one

contracts/src/core/EigenDABlobVerifier.sol
```diff
@@ -0,0 +1,296 @@
...
+        operatorStateRetriever = _operatorStateRetriever;
+        registryCoordinator = _registryCoordinator;
+    }
+
+    /**
+     * @notice Verifies a the blob is valid for the required quorums
+     * @param blobHeader The blob header to verify
+     * @param blobVerificationProof The blob verification proof to verify the blob against
+     */
+    function verifyBlobV1(
```
ian-shim, line 42: do existing v1 integrations need to update and call this new method or can they continue using the old verify method?

contracts/src/core/EigenDARelayRegistry.sol
```diff
@@ -0,0 +1,43 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+import {IEigenDARelayRegistry} from "../interfaces/IEigenDARelayRegistry.sol";
+import {OwnableUpgradeable} from "@openzeppelin-upgrades/contracts/access/OwnableUpgradeable.sol";
+
+contract EigenDARelayRegistry is IEigenDARelayRegistry, OwnableUpgradeable {
+
+    mapping(uint32 => string) public relayIdToURL;
+    mapping(address => uint32) public relayAddressToId;
```
ian-shim, line 10: Curious what's a relay address?

contracts/src/core/EigenDAThresholdRegistry.sol
```diff
@@ -0,0 +1,90 @@
...
+        uint16[] memory _versions,
+        VersionedBlobParams[] memory _versionedBlobParams,
+        SecurityThresholds memory _defaultSecurityThresholdsV2
+    ) external initializer {
+        _transferOwnership(_initialOwner);
+
+        quorumAdversaryThresholdPercentages = _quorumAdversaryThresholdPercentages;
+        quorumConfirmationThresholdPercentages = _quorumConfirmationThresholdPercentages;
+        quorumNumbersRequired = _quorumNumbersRequired;
+        defaultSecurityThresholdsV2 = _defaultSecurityThresholdsV2;
```
ian-shim, line 30: is there no update function for `defaultSecurityThresholdsV2`?

contracts/src/interfaces/IEigenDAStructs.sol
```diff
@@ -0,0 +1,128 @@
...
+
+struct BlobCommitment {
+    BN254.G1Point commitment;
+    BN254.G2Point lengthCommitment;
+    BN254.G2Point lengthProof;
+    uint32 dataLength;
+}
+
+struct SignedBatch {
+	BatchHeaderV2 batchHeader;
```
ian-shim, line 87: nit: indent

ian-shim (commented):
few questions

contracts/src/core/EigenDAThresholdRegistryStorage.sol
```diff
@@ -0,0 +1,20 @@
...
+import {IEigenDAThresholdRegistry} from "../interfaces/IEigenDAThresholdRegistry.sol";
+import "../interfaces/IEigenDAStructs.sol";
+
+abstract contract EigenDAThresholdRegistryStorage is IEigenDAThresholdRegistry {
+
+    bytes public quorumAdversaryThresholdPercentages;
+
+    bytes public quorumConfirmationThresholdPercentages;
+
+    bytes public quorumNumbersRequired;
```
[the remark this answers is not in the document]
0x0aa0: ya this is still needed for V1 verification

contracts/src/core/EigenDAThresholdRegistryStorage.sol
```diff
@@ -0,0 +1,20 @@
...
+
+    mapping(uint16 => VersionedBlobParams) public versionedBlobParams;
+
+    SecurityThresholds public defaultSecurityThresholdsV2;
```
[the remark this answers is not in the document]
0x0aa0: yes you can supply custom security thresholds in V2 verification

contracts/src/core/EigenDABlobVerifier.sol
0x0aa0: old library will continue to work

contracts/src/core/EigenDARelayRegistry.sol
0x0aa0: onchain address of a relay, the thinking here is that in the future we may want this mapping to permission updates from relays

contracts/src/core/EigenDAThresholdRegistry.sol
0x0aa0: good call

contracts/src/core/EigenDABlobVerifier.sol
```diff
@@ -0,0 +1,302 @@
...
+    /**
+     * @notice Verifies a blob for the base required quorums and the default security thresholds
+     * @param batchHeader The batch header of the blob
+     * @param blobVerificationProof The blob verification proof for the blob
+     * @param nonSignerStakesAndSignature The nonSignerStakesAndSignature for the blob
+     */
+    function verifyBlobV2(
+        BatchHeaderV2 calldata batchHeader,
+        BlobVerificationProofV2 calldata blobVerificationProof,
+        NonSignerStakesAndSignature calldata nonSignerStakesAndSignature
```
samlaf, line 126: we talked about this during the call, but would be nice to have a simple API `verifyBlobFromSignedBatchV2(SignedBatch calldata signedBatch)` that would call `_getNonSignerStakesAndSignature`, when doing eth_calls

contracts/src/core/EigenDABlobVerifier.sol
```diff
@@ -0,0 +1,302 @@
...
+     */
+    function getNonSignerStakesAndSignature(
+        SignedBatch calldata signedBatch,
+        BlobHeaderV2 calldata blobHeader
+    ) external view returns (NonSignerStakesAndSignature memory) {
+        return EigenDABlobVerificationUtils._getNonSignerStakesAndSignature(
+            operatorStateRetriever, 
+            registryCoordinator, 
+            signedBatch.attestation, 
+            blobHeader.quorumNumbers
```
samlaf, line 230: this is apparently a bug, and we should use the quorumNumbers from the signedBatch. cc @mooselumph for details (I didnt understand why)

contracts/compile.sh
```diff
@@ -21,7 +21,7 @@ function create_binding {
 forge clean
 forge build
 
-contracts="PaymentVault SocketRegistry AVSDirectory DelegationManager BitmapUtils OperatorStateRetriever RegistryCoordinator BLSApkRegistry IndexRegistry StakeRegistry BN254 EigenDAServiceManager IEigenDAServiceManager MockRollup EjectionManager IEigenDARelayRegistry"
+contracts="PaymentVault SocketRegistry AVSDirectory DelegationManager BitmapUtils OperatorStateRetriever RegistryCoordinator BLSApkRegistry IndexRegistry StakeRegistry BN254 EigenDAServiceManager IEigenDAServiceManager MockRollup EjectionManager EigenDABlobVerifier EigenDAThresholdRegistry IEigenDARelayRegistry"
```
ian-shim, line 24: I think it should be `IEigenDARelayRegistry` instead of `IEigenDARelayRegistry`

contracts/src/core/EigenDAThresholdRegistry.sol
```diff
@@ -0,0 +1,97 @@
...
+
+    function updateQuorumConfirmationThresholdPercentages(bytes memory _quorumConfirmationThresholdPercentages) external onlyOwner {
+        quorumConfirmationThresholdPercentages = _quorumConfirmationThresholdPercentages;
+    }
+
+    function updateQuorumNumbersRequired(bytes memory _quorumNumbersRequired) external onlyOwner {
+        quorumNumbersRequired = _quorumNumbersRequired;
+    }
+
+    function updateVersionedBlobParams(uint16 version, VersionedBlobParams memory _versionedBlobParams) external onlyOwner {
```
ian-shim, line 50: What are your thoughts on making these values immutable and append only? 
Instead of allowing caller to set values for arbitrary versions, can we make it append only and enforce you can only add a new version in increments of 1?

contracts/src/core/EigenDARelayRegistry.sol
```diff
@@ -0,0 +1,44 @@
...
+    function addRelayURL(address relay, string memory relayURL) external onlyOwner returns (uint32) {
+        relayKeyToURL[nextRelayKey] = relayURL;
+        relayAddressToKey[relay] = nextRelayKey;
+        relayKeyToAddress[nextRelayKey] = relay;
+
+        emit RelayAdded(relay, nextRelayKey, relayURL);
+        return nextRelayKey++;
+    }
+
+    function getRelayURL(uint32 key) external view returns (string memory) {
```
ian-shim, line 33: should it check if `key < nextRelayKey`?

contracts/src/core/EigenDAThresholdRegistry.sol
```diff
@@ -0,0 +1,105 @@
...
+    ///////////////////////// V2 ///////////////////////////////
+
+    /// @notice Gets the default security thresholds for V2
+    function getDefaultSecurityThresholdsV2() external view returns (SecurityThresholds memory) {
+        return defaultSecurityThresholdsV2;
+    }
+
+    /// @notice Returns the blob params for a given blob version
+    function getBlobParams(uint16 version) external view returns (VersionedBlobParams memory) {
+        return versionedBlobParams[version];
```
ian-shim, line 103: same here, should we check `version < nextBlobVersion`?

ian-shim:
One thing to keep consistent is whether to use which uint types to represent blob versions/relay keys. Offchain, we use `uint8` for blob versions and `uint16` for relay keys, but here we're using `uint16` and `uint32` respectively. Which types do we want? cc @mooselumph @cody-littley

contracts/src/core/EigenDAThresholdRegistry.sol
0x0aa0: dont think it is totally necessary to require on the getter

contracts/script/EigenDADeployer.s.sol
```diff
@@ -226,6 +242,42 @@ contract EigenDADeployer is DeployOpenEigenLayer {
             )
         );
 
+        eigenDAThresholdRegistryImplementation = new EigenDAThresholdRegistry();
+
+        VersionedBlobParams[] memory versionedBlobParams = new VersionedBlobParams[](0);
```
ian-shim, line 247: Does it make sense to hardcode the params for version 0 here if we're going to use the same values across all environments?

contracts/src/libraries/EigenDAHasher.sol
```diff
@@ -117,6 +117,17 @@ library EigenDAHasher {
      * @param blobHeader the V2 blob header to hash
      */
     function hashBlobHeaderV2(BlobHeaderV2 memory blobHeader) internal pure returns(bytes32) {
```
ian-shim, line 119: Do we ever have to hash BlobCertificate from chain? 
Wondering how this affects hasher of BlobCertificate if it exists

contracts/src/core/EigenDARelayRegistry.sol
```diff
@@ -0,0 +1,44 @@
...
+ * @title Registry for EigenDA relay keys
+ * @author Layr Labs, Inc.
+ */
+contract EigenDARelayRegistry is OwnableUpgradeable, EigenDARelayRegistryStorage, IEigenDARelayRegistry {
+
+    constructor() {
+        _disableInitializers();
+    }
+
+    function initialize(
```
gpsanant, line 18: addd natspec to interfaces and use @inheritdoc

contracts/src/core/EigenDARelayRegistry.sol
```diff
@@ -0,0 +1,44 @@
...
+        address _initialOwner
+    ) external initializer {
+        _transferOwnership(_initialOwner);
+    }
+
+    function addRelayURL(address relay, string memory relayURL) external onlyOwner returns (uint32) {
+        relayKeyToURL[nextRelayKey] = relayURL;
```
gpsanant, line 25: if we end up storing more per relay, lets make this

```solidity
mapping(uint32 => Relay) public relays;
```

contracts/src/core/EigenDARelayRegistry.sol
```diff
@@ -0,0 +1,44 @@
...
+    function getRelayURL(uint32 key) external view returns (string memory) {
+        return relayKeyToURL[key];
+    }
+
+    function getRelayKey(address relay) external view returns (uint32) {
+        return relayAddressToKey[relay];
+    }
+
+    function getRelayAddress(uint32 key) external view returns (address) {
+        return relayKeyToAddress[key];
```
gpsanant, line 42: these should all be public maps?

contracts/src/core/EigenDAServiceManager.sol
```diff
@@ -152,24 +152,19 @@ contract EigenDAServiceManager is EigenDAServiceManagerStorage, ServiceManagerBa
...
     }
 
     /// @notice Returns the bytes array of quorumAdversaryThresholdPercentages
     function quorumConfirmationThresholdPercentages() external view returns (bytes memory) {
-        return hex"373737";
+        return eigenDAThresholdRegistry.quorumConfirmationThresholdPercentages();
     }
 
     /// @notice Returns the bytes array of quorumsNumbersRequired
     function quorumNumbersRequired() external view returns (bytes memory) {
```
gpsanant, line 166: is this for backwards compatability or?

contracts/src/core/EigenDAThresholdRegistry.sol
```diff
@@ -0,0 +1,105 @@
...
+        return nextBlobVersion++;
+    }
+
+    ///////////////////////// V1 ///////////////////////////////
+
+    /// @notice Gets the adversary threshold percentage for a quorum
+    function getQuorumAdversaryThresholdPercentage(
+        uint8 quorumNumber
+    ) public view virtual returns (uint8 adversaryThresholdPercentage) {
+        if(quorumAdversaryThresholdPercentages.length > quorumNumber){
```
gpsanant, line 72: we should revert if no?

contracts/src/core/EigenDABlobVerifier.sol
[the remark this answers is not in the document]
epociask: we implicitly use this term across our documentation and integrations - would be nice to have an explicit definition

ian-shim (commented):
One comment outside this PR but relevant, shouldn't the [reference block number](https://github.com/Layr-Labs/eigenda/blob/master/contracts/src/interfaces/IEigenDAStructs.sol#L68) be removed from blob certificate?

ian-shim (approved):

contracts/src/interfaces/IEigenDAStructs.sol
```diff
@@ -65,7 +70,6 @@ struct BlobVerificationProofV2 {
 
 struct BlobCertificate {
     BlobHeaderV2 blobHeader;
-    uint32 referenceBlockNumber;
```
ian-shim, line 68: I think it should be removed from `Attestation` as well

PR #782, opened 2024-10-02
Fix docker bake tagging

pschork:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -82,10 +82,6 @@ integration-tests-dataapi:
 	make dataapi-build
 	go test -v ./disperser/dataapi
 
-docker-internal-release:
-	BUILD_TAG=${BRANCH} SEMVER=${SEMVER} GITDATE=${GITDATE} GIT_SHA=${GITSHA} GIT_SHORT_SHA=${GITCOMMIT} \
-	docker buildx bake internal-release ${PUSH_FLAG}
-
 docker-release-build:
 	BUILD_TAG=${SEMVER} SEMVER=${SEMVER} GITDATE=${GITDATE} GIT_SHA=${GITSHA} GIT_SHORT_SHA=${GITCOMMIT} \
 	docker buildx bake node-group-release ${PUSH_FLAG}
--- a/Makefile
+++ b/Makefile
@@ -78,6 +78,10 @@ integration-tests-dataapi:
 	make dataapi-build
 	go test -v ./disperser/dataapi
 
+docker-internal-build:
+	SEMVER=${SEMVER} GITCOMMIT=${GITCOMMIT} GITDATE=${GITDATE} \
+	docker buildx bake ${PUSH_FLAG}
+
 docker-release-build:
 	BUILD_TAG=${SEMVER} SEMVER=${SEMVER} GITCOMMIT=${GITCOMMIT} GITDATE=${GITDATE} \
 	docker buildx bake node-group-release ${PUSH_FLAG}
--- a/Makefile
+++ b/Makefile
@@ -3,11 +3,15 @@
 ifeq ($(wildcard .git/*),)
 $(warning semver disabled - building from release zip)
 GITCOMMIT := ""
+GITSHA := ""
 GITDATE := ""
+BRANCH := ""
 SEMVER := $(shell basename $(CURDIR))
 else
 GITCOMMIT := $(shell git rev-parse --short HEAD)
 GITDATE := $(shell git log -1 --format=%cd --date=unix)
+GITSHA := $(shell git rev-parse HEAD)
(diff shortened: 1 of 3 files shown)


docker-bake.hcl
```diff
@@ -15,8 +15,12 @@ variable "SEMVER" {
...
 }
 
-variable "GITCOMMIT" {
-  default = "dev"
+variable "GIT_SHA" {
+  default = "$GIT_SHA"
+}
+
+variable "GIT_SHORT_SHA" {
+  default = "$GIT_SHORT_SHA"
```
pschork, line 23: These defaults are designed to fail (via invalid tag format) if someone tries to bake a release target without exporting `GIT_SHA` or `GIT_SHORT_SHA` env variables
```
docker buildx bake internal-release
[+] Building 0.1s (1/1) FINISHED                                                                                                                                                                                docker:desktop-linux
 => [internal] load local bake definitions                                                                                                                                                                                      0.0s
 => => reading docker-bake.hcl 5.14kB / 5.14kB                                                                                                                                                                                  0.0s
ERROR: invalid tag "ghcr.io/eigenda-traffic-generator:$GIT_SHA": invalid reference format
```
without a default, the build will still fail, but error message is harder to grok what the issue is.
```
docker buildx bake internal-release
[+] Building 0.1s (1/1) FINISHED                                                                                                                                                                                docker:desktop-linux
 => [internal] load local bake definitions                                                                                                                                                                                      0.0s
 => => reading docker-bake.hcl 5.14kB / 5.14kB                                                                                                                                                                                  0.0s
ERROR: invalid tag "ghcr.io/eigenda-traffic-generator:": invalid reference format
```

Makefile
```diff
@@ -3,11 +3,15 @@
...
 GITCOMMIT := ""
+GITSHA := ""
 GITDATE := ""
+BRANCH := ""
 SEMVER := $(shell basename $(CURDIR))
 else
 GITCOMMIT := $(shell git rev-parse --short HEAD)
 GITDATE := $(shell git log -1 --format=%cd --date=unix)
+GITSHA := $(shell git rev-parse HEAD)
+BRANCH := $(shell git rev-parse --abbrev-ref HEAD | sed 's/[^[:alnum:]\.\_\-]/-/g')
```
pschork, line 14: Note: this sanitizes the branch name to conform with docker tagging syntax rules.

docker-bake.hcl
```diff
@@ -30,99 +34,166 @@ group "default" {
...
-# DOCKER METADATA TARGET
-# See https://github.com/docker/metadata-action?tab=readme-ov-file#bake-definition
+# Github CI builds
+group "ci-release" {
+  targets = ["node-group", "batcher", "disperser", "encoder", "retriever", "churner", "dataapi"]
+}
+
+# Internal devops builds
+group "internal-release" {
+  targets = ["node-internal", "batcher-release", "disperser-release", "encoder-release", "retriever-release", "churner-release", "dataapi-release", "traffic-generator-release"]
```
dmanc, line 56: Why is it node-internal and not node-release?

dmanc (approved):
lgtm :shipit:

docker-bake.hcl
samlaf: I'd prob add this as a comment above the variables themselves for documentation. Otherwise I can see myself reading this and being super confused wondering where/how this variable is defined and used.

docker-bake.hcl
pschork: node-release was already used for public node releases and is incompatible with internal release targets.

pschork: I'll rename `node-internal-release`

samlaf: also confused. is it because node-release is already something separate which builds with the arm target?

Should we call all the other internal-release targets xxx-internal instead of release then? Especially if those things are only internal, why call them release?

pschork: > also confused. is it because node-release is already something separate which builds with the arm target?

it's because the existing `node-release` is incompatible with out internal release tagging.

```
target "node-release" {
  inherits = ["node", "_release"]
  tags     = ["${REGISTRY}/${REPO}/opr-node:${BUILD_TAG}"]
}
```

```
target "node-internal-release" {
  inherits = ["node"]
  tags       = ["${REGISTRY}/eigenda-node:${BUILD_TAG}",
                "${REGISTRY}/eigenda-node:${GIT_SHA}",
                "${REGISTRY}/eigenda-node:sha-${GIT_SHORT_SHA}",
               ]
}
```

> Should we call all the other internal-release targets xxx-internal instead of release then? Especially if those things are only internal, why call them release?

Definitely agree that "release" is way overloaded. I'll rename

PR #789, opened 2024-10-07
[payments] meterer structs and helpers

hopeyen:

adding structures used by payment meterer, with utility functions and unit tests

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests


core/meterer/meterer.go
```diff
@@ -0,0 +1,64 @@
...
+
+type TimeoutConfig struct {
+	ChainReadTimeout    time.Duration
+	ChainWriteTimeout   time.Duration
+	ChainStateTimeout   time.Duration
+	TxnBroadcastTimeout time.Duration
+}
+
+// network parameters (this should be published on-chain and read through contracts)
+type Config struct {
```
mooselumph, line 17: It would be useful to have comments describing what these parameters mean.

hopeyen: sounds good! added some comments

core/meterer/offchain_store.go
```diff
@@ -0,0 +1,293 @@
...
+}
+
+func (s *OffchainStore) UpdateReservationBin(ctx context.Context, accountID string, binIndex uint64, size uint64) (uint64, error) {
+	key := map[string]types.AttributeValue{
+		"AccountID": &types.AttributeValueMemberS{Value: accountID},
+		"BinIndex":  &types.AttributeValueMemberN{Value: strconv.FormatUint(binIndex, 10)},
+	}
+
+	update := map[string]types.AttributeValue{
+		"BinUsage": &types.AttributeValueMemberN{Value: strconv.FormatUint(uint64(size), 10)},
```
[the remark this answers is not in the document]
hopeyen: thanks! updated UpdateItemIncrement function to `IncrementBy` that only takes in a key name and a value uint64

core/meterer/offchain_store.go
```diff
@@ -0,0 +1,293 @@
...
+	}
+
+	fmt.Println("increment the item in a table", s.reservationTableName)
```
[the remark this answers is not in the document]
hopeyen: removed!

core/meterer/offchain_store.go
```diff
@@ -0,0 +1,293 @@
...
+	err = attributevalue.UnmarshalMap(result, &reservation)
+	if err != nil {
+		return nil, err
+	}
+
+	return &reservation, nil
+}
+
+// Find all reservation bins for a given account
+func (s *OffchainStore) FindReservationBins(ctx context.Context, accountID string) ([]ReservationBin, error) {
```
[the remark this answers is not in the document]
hopeyen: nope, this was for later exposing API to the user, but removed for now

core/meterer/types.go
```diff
@@ -0,0 +1,192 @@
...
+	"github.com/ethereum/go-ethereum/signer/core/apitypes"
+)
+
+/* SUBJECT TO BIG MODIFICATIONS */
+
+// BlobHeader represents the header information for a blob
+type BlobHeader struct {
+	// Existing fields
+	Commitment    core.G1Point
+	DataLength    uint32
```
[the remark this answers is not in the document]
hopeyen: I added a comment for this, though should I update the name to be something like `NumSymbols`?

core/meterer/onchain_state.go
```diff
@@ -0,0 +1,110 @@
...
+/* HEAVILY MOCKED */
+
+var (
+	DummyReservationBytesLimit    = uint64(1024)
+	DummyPaymentLimit             = TokenAmount(512)
+	DummyMinimumChargeableSize    = uint32(128)
+	DummyMinimumChargeablePayment = uint32(128)
+
+	DummyReservation     = ActiveReservation{DataRate: DummyReservationBytesLimit, StartTimestamp: 0, EndTimestamp: math.MaxUint32, QuorumSplit: []byte{50, 50}, QuorumNumbers: []uint8{0, 1}}
+	DummyOnDemandPayment = OnDemandPayment{CumulativePayment: DummyPaymentLimit}
```
ian-shim, line 21: instead of mocking directly in the implementation, the implementation should use `tx` to load these onchain parameters, and the `OnchainPaymentState` used in tests should load a mocked `tx` that returns the mock values

core/meterer/offchain_store.go
```diff
@@ -0,0 +1,239 @@
...
+
+	dynamoClient, err := commondynamodb.NewClient(cfg, logger)
+	if err != nil {
+		return nil, err
+	}
+	if reservationTableName == "" || onDemandTableName == "" || globalBinTableName == "" {
+		return nil, fmt.Errorf("table names cannot be empty")
+	}
+
+	err = CreateReservationTable(cfg, reservationTableName)
```
ian-shim, line 42: Tables should be created outside the application code (devops). We can check if the tables with corresponding names exist in the constructor and error out if they don't exist

core/meterer/util.go
```diff
@@ -0,0 +1,164 @@
...
+				KeyType:       types.KeyTypeHash,
+			},
+			{
+				AttributeName: aws.String("BinIndex"),
+				KeyType:       types.KeyTypeRange,
+			},
+		},
+		GlobalSecondaryIndexes: []types.GlobalSecondaryIndex{
+			{
+				IndexName: aws.String("AccountIDIndex"),
```
ian-shim, line 45: Do we need this index? Looks like we already key by `AccountID`

core/meterer/util.go
```diff
@@ -0,0 +1,164 @@
...
+				KeyType:       types.KeyTypeHash,
+			},
+			{
+				AttributeName: aws.String("CumulativePayments"),
+				KeyType:       types.KeyTypeRange,
+			},
+		},
+		GlobalSecondaryIndexes: []types.GlobalSecondaryIndex{
+			{
+				IndexName: aws.String("AccountIDIndex"),
```
ian-shim, line 137: same here

core/meterer/util.go
```diff
@@ -0,0 +1,164 @@
...
+
+	commonaws "github.com/Layr-Labs/eigenda/common/aws"
+	test_utils "github.com/Layr-Labs/eigenda/common/aws/dynamodb/utils"
+	"github.com/Layr-Labs/eigenda/core"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb/types"
+)
+
+func DummyCommitment() core.G1Point {
```
ian-shim, line 15: This doesn't seem used anywhere

core/meterer/util.go
```diff
@@ -0,0 +1,164 @@
...
+	commitment := core.NewG1Point(big.NewInt(123), big.NewInt(456))
+	return *commitment
+}
+
+func CreateReservationTable(clientConfig commonaws.ClientConfig, tableName string) error {
```
ian-shim, line 20: I feel like we can collapse these tables into a single table design. Wdyt?

core/meterer/meterer.go
```diff
@@ -0,0 +1,66 @@
...
+// To check if the payment is valid, the meterer will:
+//  1. check if the signature is valid
+//     (against the CumulativePayments and BinIndex fields ;
+//     maybe need something else to secure against using this appraoch for reservations when rev request comes in same bin interval; say that nonce is signed over as well)
+//  2. For reservations, check offchain bin state as demonstrated in pseudocode, also check onchain state before rejecting (since onchain data is pulled)
+//  3. For on-demand, check against payments and the global rates, similar to the reservation case
+//
+// If the payment is valid, the meterer will add the blob header to its state and return a success response to the disperser API server.
+// if any of the checks fail, the meterer will return a failure response to the disperser API server.
+var OnDemandQuorumNumbers = []uint8{0, 1}
```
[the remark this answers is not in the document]
hopeyen: Do you think they should be published on-chain?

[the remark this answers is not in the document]
ian-shim: We already have them onchain depending on what you need. 
QuorumCount [here](https://etherscan.io/address/0x0baac79acd45a023e19345c352d8a7a83c4e5656#readProxyContract#F27) and required quorums [here](https://etherscan.io/address/0x870679e138bcdf293b7ff14dd44b70fc97e12fc0#readProxyContract#F20)

core/meterer/meterer.go
```diff
@@ -0,0 +1,66 @@
...
+)
+
+type TimeoutConfig struct {
+	ChainReadTimeout    time.Duration
+	ChainWriteTimeout   time.Duration
+	ChainStateTimeout   time.Duration
+	TxnBroadcastTimeout time.Duration
+}
+
+// network parameters that should be published on-chain. We currently configure these params through disperser env vars.
```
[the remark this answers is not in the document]
hopeyen: updated, thanks for the tip!

core/meterer/meterer.go
```diff
@@ -0,0 +1,66 @@
...
+type Config struct {
```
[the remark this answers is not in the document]
hopeyen: I deleted this field and then added a timeout configuration to `Config`. (I realized that we are not writing or boardcasting for the initial scope, so only reading the state from chain is needed)

core/meterer/onchain_state.go
hopeyen: moved to use mocked on-chain state, lmk if this is what you had in mind 🤔 I wanted to have some kind of cache of on-chain state in-memory and `tx` updates it periodically, so the focus was just grabbing cached states

core/meterer/offchain_store.go
hopeyen: makes sense, updated to use check and error instead of making db changes

core/meterer/util.go
hopeyen: During testing, I found that this IndexName was needed to enable `QueryIndex`. I referenced the pattern like [here] with `BlobStatus` key and `StatusIndex` index (https://github.com/Layr-Labs/eigenda/blob/35e0b6b11182f1ab0fbbc7453f515b78703d9b26/disperser/common/blobstore/blob_metadata_store.go#L462).

core/meterer/util.go
hopeyen: this is used in later PRs but I've removed it since we are removing commitment from the Payment header

core/meterer/util.go
hopeyen: 🤔 something like [this](https://gist.github.com/hopeyen/7fa8f7c38b9d9a9b78beadf4c818c57e)?
My main consideration is the read/write performance based on the evenness of key distributions, and think it is clearer if different payment methods live in different tables, but I'm interested in learning the advantages of combining these into a single table

core/auth/payment_metadata.go
```diff
@@ -0,0 +1,169 @@
...
+			VerifyingContract: verifyingContract.Hex(),
+		},
+		types: apitypes.Types{
+			"EIP712Domain": []apitypes.Type{
+				{Name: "name", Type: "string"},
+				{Name: "version", Type: "string"},
+				{Name: "chainId", Type: "uint256"},
+				{Name: "verifyingContract", Type: "address"},
+			},
+			"PaymentMetadata": []apitypes.Type{
```
ian-shim, line 39: Does this depend on the structure of `BlobHeader` in v2? (i.e. would signature over `BlobHeader` be sufficient without one over `PaymentMetadata`?)
Do we need this struct signed in both v1 & v2?

core/data.go
```diff
@@ -470,3 +470,36 @@ func (cb Bundles) FromEncodedBundles(eb EncodedBundles) (Bundles, error) {
 	}
 	return c, nil
 }
+
+// PaymentMetadata represents the header information for a blob
+type PaymentMetadata struct {
```
ian-shim, line 475: Should this be independent of what BlobHeader looks like?

core/meterer/meterer.go
```diff
@@ -0,0 +1,57 @@
...
+	"time"
+
+	"github.com/Layr-Labs/eigensdk-go/logging"
+)
+
+// Config contains network parameters that should be published on-chain. We currently configure these params through disperser env vars.
+type Config struct {
+	// for rate limiting 2^64 ~= 18 exabytes per second; 2^32 ~= 4GB/s
+	// for payments      2^64 ~= 18M Eth;                2^32 ~= 4ETH
+	GlobalBytesPerSecond uint64 // Global rate limit in bytes per second for on-demand payments
```
ian-shim, line 13: nit: it's nicer to put comment for field description above the definition, starting with the name of the field

core/meterer/meterer.go
```diff
@@ -0,0 +1,57 @@
...
+	MinChargeableSize    uint32 // Minimum size of a chargeable unit in bytes, used as a floor for on-demand payments
+	PricePerChargeable   uint32 // Price per chargeable unit in gwei, used for on-demand payments
+	ReservationWindow    uint32 // Duration of all reservations in seconds, used to calculate bin indices
+
+	ChainReadTimeout time.Duration // Timeout for reading payment state from chain
+}
+
+// disperser API server will receive requests from clients. these requests will be with a blobHeader with payments information (CumulativePayments, BinIndex, and Signature)
```
ian-shim, line 21: nit: `Meterer handles payment accounting across different accounts. Disperser API server receives requests from clients and each request contains a blob header....`

core/meterer/meterer.go
```diff
@@ -0,0 +1,57 @@
...
+//  3. For on-demand, check against payments and the global rates, similar to the reservation case
+//
+// If the payment is valid, the meterer will add the blob header to its state and return a success response to the disperser API server.
+// if any of the checks fail, the meterer will return a failure response to the disperser API server.
+
+type Meterer struct {
+	Config
+
+	ChainState    OnchainPayment
+	OffchainStore OffchainStore
```
ian-shim, line 37: nit: brief description of what gets stored on which store and how they're used differently?

core/meterer/util.go
ian-shim: Yeah if you need to query by something other than the table's default keys, then building indexes on those attributes may help. However, looks like we can query by the table's default keys, in which case we can query from the table [directly](https://github.com/Layr-Labs/eigenda/blob/35e0b6b11182f1ab0fbbc7453f515b78703d9b26/common/aws/dynamodb/client.go#L195)

core/auth/payment_metadata.go
hopeyen: I think payment metadata will depend on `BlobHeader` v2 in which `PaymentMetadata` includes `BlobHeader`'s hash, a field like `BlobKey`. 
This struct should always be signed, but I guess payments is currently disjointed from blob information. We could discuss whether it is okay to let payment be "blob-agnostic" in v1?

common/aws/dynamodb/client.go
```diff
@@ -2,8 +2,10 @@ package dynamodb
```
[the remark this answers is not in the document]
hopeyen: ahh just realized that this branch was configured to merge into the dynamo-update branch. updated the merge base so the changes should be hidden now

core/data.go
hopeyen: hmm we could add a field like blob key or blob pointer? this would mean we need to do blob lookup. With these current fields, we don't need the blob or blob header to do metering

core/auth/payment_metadata_test.go
```diff
@@ -0,0 +1,173 @@
...
+		BinIndex:          0,
+		CumulativePayment: 1000,
+		DataLength:        1024,
+		QuorumNumbers:     []uint8{1},
+	}
+
+	t.Run("SignPaymentMetadata", func(t *testing.T) {
+		signature, err := signer.SignPaymentMetadata(header, privateKey)
+		require.NoError(t, err)
+		assert.NotEmpty(t, signature)
```
mooselumph, line 33: One thing you can do is use a contract script like this to generate the same signature on chain and compare. 
https://github.com/Layr-Labs/eigenda/blob/fab84e38d27c61d6c0cb66c77d25f6a9cfc0dc99/contracts/script/GenerateUnitTestHashes.s.sol#L13

core/auth/payment_metadata.go
```diff
@@ -0,0 +1,168 @@
...
+/* SUBJECT TO MODIFICATIONS */
+
+// EIP712Signer handles EIP-712 domain specific signing operations over typed and structured data
+type EIP712Signer struct {
+	domain apitypes.TypedDataDomain
+	types  apitypes.Types
+}
+
+// NewEIP712Signer creates a new EIP712Signer instance
+func NewEIP712Signer(chainID *big.Int, verifyingContract common.Address) *EIP712Signer {
```
mooselumph, line 24: Does the `verifyingContract` ever have a function?

core/auth/payment_metadata.go
mooselumph: In order to support both v1 and v2, what if we were to add a `BlobKey` field here which is not populated in v1?

core/auth/payment_metadata.go
```diff
@@ -0,0 +1,168 @@
...
+				{Name: "name", Type: "string"},
+				{Name: "version", Type: "string"},
+				{Name: "chainId", Type: "uint256"},
+				{Name: "verifyingContract", Type: "address"},
+			},
+			"PaymentMetadata": []apitypes.Type{
+				{Name: "accountID", Type: "string"},
+				{Name: "binIndex", Type: "uint32"},
+				{Name: "cumulativePayment", Type: "uint64"},
+				{Name: "dataLength", Type: "uint32"},
```
mooselumph, line 43: I'm not sure the dataLength would be needed in v2 if we also have the blob key.

core/auth/payment_metadata.go
```diff
@@ -0,0 +1,168 @@
...
+
+	signature, err := s.signTypedData(typedData, privateKey)
+	if err != nil {
+		return nil, fmt.Errorf("error signing payment metadata (header): %v", err)
+	}
+
+	return signature, nil
+}
+
+func convertUint8SliceToMap(params []uint8) []string {
```
mooselumph, line 73: `convertUint8SliceToStringSlice`?

core/auth/payment_metadata.go
hopeyen: Could I assume meterer will receive validated blob header which contains payment metadata and the quorum numbers and length? If so I could remove `validateSignature` from the meterer process and make a new blobHeader struct?

core/auth/payment_metadata.go
hopeyen: Nope, verifyingContract would only affect the output signature and the recovery

common/aws/dynamodb/client.go
```diff
@@ -408,3 +409,17 @@ func (c *Client) readItems(ctx context.Context, tableName string, keys []Key) ([
 
 	return items, nil
 }
+
+// TableCheck checks if a table exists and can be described
+func (c *Client) TableCheck(ctx context.Context, name string) error {
```
ian-shim, line 414: nit: `TableExists`?

core/data.go
```diff
@@ -470,3 +471,50 @@ func (cb Bundles) FromEncodedBundles(eb EncodedBundles) (Bundles, error) {
...
+	data = append(data, binIndexBytes...)
+
+	paymentBytes := make([]byte, 8)
+	binary.BigEndian.PutUint64(paymentBytes, pm.CumulativePayment)
+	data = append(data, paymentBytes...)
+
+	return crypto.Keccak256(data)
+}
+
+type TokenAmount uint64 // TODO: change to uint128
```
ian-shim, line 505: should it be big.Int?

core/meterer/meterer.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// Meterer handles payment accounting across different accounts. Disperser API server receives requests from clients and each request contains a blob header
+// with payments information (CumulativePayments, BinIndex, and Signature). Disperser will pass the blob header to the meterer, which will check if the
+// payments information is valid.
+type Meterer struct {
+	Config
+
+	// ChainState reads on-chain payment state periodically and cache it in memory
+	ChainState OnchainPayment
```
ian-shim, line 31: `OnchainPayment`? `ChainState` sounds like the `ChainState` struct in chainio 😭

core/meterer/onchain_state.go
```diff
@@ -0,0 +1,93 @@
...
+	"github.com/Layr-Labs/eigenda/core"
+	"github.com/Layr-Labs/eigenda/core/eth"
+)
+
+// PaymentAccounts (For reservations and on-demand payments)
+
+// OnchainPaymentState is an interface for getting information about the current chain state for payments.
+type OnchainPayment interface {
+	GetCurrentBlockNumber(ctx context.Context) (uint32, error)
+	CurrentOnchainPaymentState(ctx context.Context, tx *eth.Transactor) (OnchainPaymentState, error)
```
ian-shim, line 16: should we use reader/writer from `chainio` to replace transactor?

core/meterer/util.go
ian-shim: Added a suggestion in the design doc. We can discuss it there

core/meterer/meterer.go
hopeyen: Can we perhaps move it later?🫠 similar to replacing transactor with reader, I would prefer creating a separate PR after both of these are merged or at least `chainio` gets merged??

core/data.go
```diff
@@ -470,3 +472,48 @@ func (cb Bundles) FromEncodedBundles(eb EncodedBundles) (Bundles, error) {
...
+	DataRate       uint64 // Bandwidth per reservation bin
+	StartTimestamp uint64 // Unix timestamp that's valid for basically eternity
+	EndTimestamp   uint64
+
+	QuorumNumbers []uint8
+	QuorumSplit   []byte // ordered mapping of quorum number to payment split; on-chain validation should ensure split <= 100
+}
+
+type OnDemandPayment struct {
+	CumulativePayment big.Int // Total amount deposited by the user
```
mooselumph, line 518: I think it's common practice to use a pointer for big.Int, i.e. `CumulativePayment *big.Int`, as most methods consume the pointer.

hopeyen: sg, updated

mooselumph (approved):

ian-shim (approved):

PR #790, opened 2024-10-08
[payments] meterer core logic

hopeyen:

Core logic of payments metering

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests


core/meterer/meterer_test.go
```diff
@@ -0,0 +1,455 @@
...
+	unregisteredUser, err := crypto.GenerateKey()
+	if err != nil {
+		t.Fatalf("Failed to generate key: %v", err)
+	}
+	header, err := meterer.ConstructPaymentMetadata(signer, 1, 1, 1000, quorumNumbers, unregisteredUser)
+	assert.NoError(t, err)
+	err = mt.MeterRequest(ctx, *header)
+	assert.Error(t, err, "failed to get on-demand payment by account: payment not found")
+
+	// test invalid quorom ID
```
mooselumph, line 278: Why is the `BinIndex` set to non-zero here and in the next one?

core/meterer/meterer_test.go
```diff
@@ -0,0 +1,455 @@
...
+	header, err = meterer.ConstructPaymentMetadata(signer, binIndex, 1, 1, quorumNumbers, privateKey2)
+	assert.NoError(t, err)
+	err = mt.MeterRequest(ctx, *header)
+	assert.Error(t, err, "insufficient cumulative payment increment")
+
+	// test cannot insert cumulative payment in out of order
+	header, err = meterer.ConstructPaymentMetadata(signer, binIndex, 0, 50, quorumNumbers, privateKey2)
+	assert.NoError(t, err)
+	err = mt.MeterRequest(ctx, *header)
+	assert.Error(t, err, "cannot insert cumulative payment in out of order")
```
mooselumph, line 323: I don't completely understand how this is different from the previous check...

core/meterer/meterer.go
```diff
@@ -62,3 +68,223 @@ func NewMeterer(
...
+
+	return nil
+}
+
+// ServeReservationRequest handles the rate limiting logic for incoming requests
+func (m *Meterer) ServeReservationRequest(ctx context.Context, blobHeader BlobHeader, reservation *ActiveReservation) error {
+	if err := m.ValidateQuorum(blobHeader, reservation.QuorumNumbers); err != nil {
+		return fmt.Errorf("invalid quorum for reservation: %w", err)
+	}
+	if !m.ValidateBinIndex(blobHeader, reservation) {
```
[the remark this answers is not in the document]
hopeyen: I think this idea is similar to having disperser set the bin index of a message. The main trade-off was client cannot easily predict the `ReceiptTime` and the behavior of the message in the local accounting. I think this is a good idea with tolerance, something like `MaxDelayReceiveTime`?

core/meterer/meterer_test.go
```diff
@@ -153,3 +174,321 @@ func teardown() {
...
+	assert.Equal(t, strconv.Itoa(int(5)), item["BinUsage"].(*types.AttributeValueMemberN).Value)
+
+	// second over flow
+	header, err = auth.ConstructPaymentMetadata(&signer, binIndex, 0, 1, quoromNumbers, privateKey2)
+	assert.NoError(t, err)
+	err = mt.MeterRequest(ctx, *header)
+	assert.ErrorContains(t, err, "bin has already been filled")
+
+	// overwhelming bin overflow for empty bins (assuming all previous requests happened within 1 reservation window)
+	header, err = auth.ConstructPaymentMetadata(&signer, binIndex-1, 0, 1000, quoromNumbers, privateKey2)
```
mooselumph, line 267: Shoud we also have a passing test for `binIndex-1`?

core/meterer/meterer_test.go
```diff
@@ -96,12 +105,15 @@ func setup(_ *testing.M) {
...
 
 	logger = logging.NewNoopLogger()
+	signer = auth.NewEIP712Signer(big.NewInt(17000), gethcommon.HexToAddress("0x1234000000000000000000000000000000000000"))
 	config := meterer.Config{
-		PricePerChargeable:   1,
-		MinChargeableSize:    1,
-		GlobalBytesPerSecond: 1000,
-		ReservationWindow:    60,
-		ChainReadTimeout:     3 * time.Second,
+		PricePerSymbol:         1,
```
mooselumph, line 110: Would it be worth using some unique prime values here for different parameters (e.g. 2,3,5,7) to give better confidence that the calcualtions are being done correctly?

mooselumph (commented):
looks good! A few more comments

core/meterer/meterer_test.go
hopeyen: yep! added it and ready for review

core/data.go
```diff
@@ -490,22 +490,24 @@ type PaymentMetadata struct {
 	BinIndex uint32
 	// TODO: we are thinking the contract can use uint128 for cumulative payment,
 	// but the definition on v2 uses uint64. Double check with team.
-	CumulativePayment uint64
+	CumulativePayment big.Int
```
mooselumph, line 493: I think it's common practice to use a pointer for big.Int, i.e. CumulativePayment *big.Int, as most methods consume the pointer.

hopeyen: sg, updated

mooselumph (approved):

core/meterer/meterer.go
```diff
@@ -41,13 +44,248 @@ func NewMeterer(
...
 		Config: config,
 
 		ChainState:    paymentChainState,
 		OffchainStore: offchainStore,
 
 		logger: logger.With("component", "Meterer"),
-	}, nil
+	}
+	// goroutine to periodically refresh the on-chain state
+	go func() {
```
ian-shim, line 57: Can we have `Start()` method instead of starting a new goroutine in constructor?

core/meterer/onchain_state.go
```diff
@@ -12,82 +12,111 @@ import (
...
 	if err != nil {
-		return nil, nil, err
+		return err
 	}
 
-	activeReservations, err := tx.GetActiveReservations(ctx, blockNumber)
-	if err != nil {
-		return nil, nil, err
+	accountIDs := make([]string, 0, len(pcs.ActiveReservations))
+	for accountID := range pcs.ActiveReservations {
```
ian-shim, line 58: where does `pcs.ActiveReservations` first get populated?

core/meterer/meterer.go
```diff
@@ -41,13 +44,248 @@ func NewMeterer(
...
+		ticker := time.NewTicker(1 * time.Hour)
+		defer ticker.Stop()
+
+		for {
+			select {
+			case <-ticker.C:
+				if err := paymentChainState.RefreshOnchainPaymentState(ctx, nil); err != nil {
```
ian-shim, line 64: is this thread safe?

core/meterer/onchain_state.go
hopeyen: `ActiveReservations` gets populated on the request basis. When a request comes in, we would check if its account id exists in this map, and query the contract if not. We can't populate it at initialization because we don't want to store a list of account IDs on the contract and there's no function to take the keys from the contract storage mapping

core/meterer/meterer.go
hopeyen: I think this is okay since when `RefreshOnChainPaymentState` swaps out the chain state's reservation and on-demand maps atomically, and individual queries (like this [one](https://github.com/Layr-Labs/eigenda/blob/67c0e1d0186b392087793f76a76a466cd4e7ad93/core/meterer/onchain_state.go#L17) takes the struct and not pointer. 
Behaviorally, if a refresh happened in the middle of a request in which the meterer is using an outdated on-chain state, we would need to explain it takes some time for the on-chain adjustment to be reflected. Please let me know if you have concerns about this, ideally I would want to avoid locks if they are not necessary

ian-shim: I don't think `RefreshOnChainPaymentState` swaps out the maps atomically. It reads the onchain state and then updates its internal state, right?

core/meterer/onchain_state.go
```diff
@@ -12,82 +12,111 @@ import (
...
+		accountIDs = append(accountIDs, accountID)
 	}
 
-	onDemandPayments, err := tx.GetOnDemandPayments(ctx, blockNumber)
+	activeReservations, err := tx.GetActiveReservations(ctx, blockNumber, accountIDs)
 	if err != nil {
-		return nil, nil, err
+		return err
 	}
+	pcs.ActiveReservations = activeReservations
```
ian-shim, line 66: I think we should lock this method unless there is a strong reason not to, as i don't think it will cause any performance issues but there is a chance `ActiveReservations` and `OnDemandPayments` are accessed in parallel?

ian-shim (approved):
just one comment around locking

core/meterer/onchain_state.go
```diff
@@ -12,82 +13,121 @@ import (
...
 	}
-	return blockNumber, nil
+	pcs.OnDemandPayments = onDemandPayments
+	pcs.OnDemandLocks.Unlock()
+
+	return nil
 }
 
 func (pcs *OnchainPaymentState) GetActiveReservations(ctx context.Context, blockNumber uint) (map[string]core.ActiveReservation, error) {
 	return pcs.ActiveReservations, nil
```
ian-shim, line 90: I think it needs to get RLock here

core/meterer/onchain_state.go
```diff
@@ -12,82 +13,121 @@ import (
...
+	}
+
+	pcs.ReservationsLock.Lock()
+	pcs.ActiveReservations[accountID] = res
+	pcs.ReservationsLock.Unlock()
+	return res, nil
 }
 
 func (pcs *OnchainPaymentState) GetOnDemandPayments(ctx context.Context, blockNumber uint) (map[string]core.OnDemandPayment, error) {
 	return pcs.OnDemandPayments, nil
```
ian-shim, line 111: same here

PR #791, opened 2024-10-08
[payments] placeholder paid blob dispersal api

hopeyen:

Added a separate dispersal API and the corresponding placeholder client functions. Content of the server and client will each be filled in through individual PRs

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proto/disperser/disperser.proto
```diff
@@ -89,13 +96,44 @@ message DisperseBlobRequest {
...
+	repeated uint32 custom_quorum_numbers = 2;
+	
+	// The account ID of the client. This should be a hex-encoded string of the ECSDA public key
+	// corresponding to the key used by the client to sign the BlobAuthHeader.
+	string account_id = 3;
+
+	// The bin index of the blob.
+	uint32 bin_index = 6;
+	// The cumulative payment of the blob.
+	uint64 cumulative_payment = 7;
```
ian-shim, line 124: should we use [common.PaymentHeader](https://github.com/Layr-Labs/eigenda/blob/master/api/proto/common/common.proto#L32)?

api/proto/disperser/disperser.proto
```diff
@@ -11,6 +11,12 @@ service Disperser {
 	// processing status of the blob.
 	rpc DisperseBlob(DisperseBlobRequest) returns (DisperseBlobReply) {}
 
+	// This API require valid payments to accept blob to disperse from clients.
+	// This executes the dispersal async, i.e. it returns once the request
+	// is accepted. The client could use GetBlobStatus() API to poll the the
+	// processing status of the blob.
+	rpc DispersePaidBlob(DispersePaidBlobRequest) returns (DisperseBlobReply) {}
```
ian-shim, line 18: should we implement a canned method in disperser server which returns `Unimplemented` error? ex. https://github.com/Layr-Labs/eigenda/blob/master/disperser/apiserver/server_v2.go#L39-L41

hopeyen: yep, I will add

api/proto/disperser/disperser.proto
hopeyen: oh! didn't realize this was merged. Going to do a more general update to reflect the changes

api/proto/disperser/disperser.proto
```diff
@@ -42,6 +48,7 @@ message AuthenticatedRequest {
 	oneof payload {
 		DisperseBlobRequest disperse_request = 1;
 		AuthenticationData authentication_data = 2;
+		DispersePaidBlobRequest paid_disperse_request = 3;
```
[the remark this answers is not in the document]
hopeyen: yes, `DisperseBlobAuthenticated` doesn't use it, I was thinking to make an `DispersePaidBlobAuthenticated` version but somehow it is here 🤔 gonna remove it. I think we want payment metadata to be attached to the blob header which has signature over. going to take a look on whether that's available to use

api/proto/disperser/disperser.proto
hopeyen: Second thought is we don't need to add this in server v2, because `DisperseBlobRequest` contains `common.v2.BlobHeader` which contains `common.PaymentHeader`. So I think `DisperseBlobRequest` API should be able to handle payments and no additional API is needed

ian-shim: yeah i meant to add it in v1 server

api/proto/disperser/disperser.proto
```diff
@@ -22,6 +28,11 @@ service Disperser {
 	// 4. The Disperser verifies the signature and returns a DisperseBlobReply message.
 	rpc DisperseBlobAuthenticated(stream AuthenticatedRequest) returns (stream AuthenticatedReply);
 
+	// DispersePaidBlobAuthenticated is similar to DispersePaidBlob, except that it requires the
+	// client to authenticate itself via the AuthenticationData message. The protoco is similar to 
+	// DisperseBlobAuthenticated.
+	rpc DispersePaidBlobAuthenticated(stream AuthenticatedPaidRequest) returns (stream AuthenticatedReply);
```
ian-shim, line 34: Not sure why we'd need another endpoint. `DispersePaidBlob` already has authenticated built in and has a simpler interface. I feel like having 4 different dispersal endpoints would be very confusing

core/meterer/onchain_state.go
```diff
@@ -14,9 +14,7 @@ import (
 // OnchainPaymentState is an interface for getting information about the current chain state for payments.
 type OnchainPayment interface {
 	RefreshOnchainPaymentState(ctx context.Context, tx *eth.Transactor) error
-	GetActiveReservations(ctx context.Context) (map[string]core.ActiveReservation, error)
```
ian-shim, line 17: Do we no longer need these getters? How is the internal state being accessed?

api/proto/disperser/disperser.proto
hopeyen: I don't think `DispersePaidBlob` has authentication built in since we removed signature for PaymentHeader. Perhaps we don't want a `DispersePaidBlob` but we do want `DispersePaidBlobAuthenticated`?

core/meterer/onchain_state.go
hopeyen: We never used these getters, so I decided to remove them. 

The internal state gets accessed in other OnchainPayment methods for either the [length](https://github.com/Layr-Labs/eigenda/blob/48ac72718d68beab5af6d0a6a078872b0e822b61/core/meterer/onchain_state.go#L59) of the map (with locks) or account ID specific inner [struct](https://github.com/Layr-Labs/eigenda/blob/48ac72718d68beab5af6d0a6a078872b0e822b61/core/meterer/onchain_state.go#L89) (no lock). 

I could add a lock around accessing the internal state?

api/proto/disperser/disperser.proto
hopeyen: added to v1

api/proto/disperser/disperser.proto
ian-shim: Hmm what if we add `signature` as a top level field in `DispersePaidBlobRequest`, which would sign over the payment header? 
@mooselumph any thoughts?

core/meterer/onchain_state.go
ian-shim: Got it. These reads look fine to me!

api/proto/disperser/disperser.proto
hopeyen: I would agree that `DispersePaidBlobAuthenticated` and `DispersePaidBlobRequest` don't need to exist at the same time and payment should be authenticated regardless of what the method name is called

ian-shim (approved):

mooselumph (approved):

PR #792, opened 2024-10-08
[payments] disperser server metering

hopeyen:

Disperser server integrated with payments metering

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Integration tests


core/data.go
```diff
@@ -494,23 +496,27 @@ type PaymentMetadata struct {
...
 
-	// Append CumulativePayment
 	paymentBytes := pm.CumulativePayment.Bytes()
 	data = append(data, paymentBytes...)
 
-	return crypto.Keccak256(data)
+	return crypto.Keccak256Hash(data)
+}
+
+// Hash returns the Keccak256 hash of the PaymentMetadata
```
ian-shim, line 513: `ConvertPaymentHeader` converts...

core/meterer/meterer.go
```diff
@@ -138,9 +130,10 @@ func (m *Meterer) ValidateQuorum(headerQuorums []uint8, allowedQuorums []uint8)
 // ValidateBinIndex checks if the provided bin index is valid
 func (m *Meterer) ValidateBinIndex(header core.PaymentMetadata, reservation *core.ActiveReservation) bool {
 	now := uint64(time.Now().Unix())
-	currentBinIndex := GetBinIndex(now, m.ReservationWindow)
+	reservationWindow := m.ChainState.GetReservationWindow()
```
ian-shim, line 133: does it make onchain query every time? should we cache these?

disperser/apiserver/server.go
```diff
@@ -303,6 +304,107 @@ func (s *DispersalServer) disperseBlob(ctx context.Context, blob *core.Blob, aut
...
+		// Note the DispersePaidBlob already updated metrics for this error.
+		s.logger.Info("failed to disperse blob", "err", err)
+	} else {
+		s.metrics.HandleSuccessfulRpcRequest("DispersePaidBlob")
+	}
+	return reply, err
+}
+
+// dispersePaidBlob checks for payment metering, otherwise does the same thing as disperseBlob
+func (s *DispersalServer) dispersePaidBlob(ctx context.Context, blob *core.Blob, quorumNumbers []uint32, binIndex uint32, cumulativePayment *big.Int, signature []byte, authenticatedAddress string, apiMethodName string) (*pb.DisperseBlobReply, error) {
```
ian-shim, line 334: looks like its mostly copied from `disperseBlob` except for the `s.meterer != nil` block. Why don't we integrate it into the existing method?

disperser/apiserver/server.go
```diff
@@ -303,6 +304,107 @@ func (s *DispersalServer) disperseBlob(ctx context.Context, blob *core.Blob, aut
...
+			s.metrics.HandleFailedRequest(codes.InvalidArgument.String(), fmt.Sprintf("%d", param.QuorumID), blobSize, apiMethodName)
+		}
+		s.metrics.HandleInvalidArgRpcRequest(apiMethodName)
+		return nil, api.NewInvalidArgError(err.Error())
+	}
+
+	s.logger.Debug("received a new paid blob dispersal request", "authenticatedAddress", authenticatedAddress, "origin", origin, "blobSizeBytes", blobSize, "securityParams", strings.Join(securityParamsStrings, ", "))
+
+	// payments before ratelimits
+	if s.meterer != nil {
```
ian-shim, line 360: How does it interact with rate limiter? Would there ever be a case where request is limited by both meterer and rate limiter? i.e. is it possible that there's been sufficient amount paid, but is above the rate limit?

disperser/apiserver/server.go
```diff
@@ -1020,3 +1122,96 @@ func (s *DispersalServer) validateRequestAndGetBlob(ctx context.Context, req *pb
...
+	}
+
+	if len(req.GetCustomQuorumNumbers()) > int(quorumConfig.QuorumCount) {
+		return nil, errors.New("number of custom_quorum_numbers must not exceed number of quorums")
+	}
+
+	seenQuorums := make(map[uint8]struct{})
+
+	// TODO: validate payment signature against payment metadata
+	if !auth.VerifyPaymentSignature(req.GetPaymentHeader(), req.GetPaymentSignature()) {
```
ian-shim, line 1161: Same here. The method looks exactly same as `validateRequestAndGetBlob` except this check. Could we extend the existing method?

core/meterer/meterer.go
hopeyen: I don't expect this to make onchain query every time. `ChainState` tracks a `PaymentVaultParams` struct that caches all the contract constants so this should query directly from the cache ([function](https://github.com/Layr-Labs/eigenda/blob/03b04753e5264a0cafb96bff676ec7651b36fb2c/core/meterer/onchain_state.go#L222)) and I expect onchain reads only when 1) [initializing](https://github.com/Layr-Labs/eigenda/blob/03b04753e5264a0cafb96bff676ec7651b36fb2c/core/meterer/onchain_state.go#L49) `ChainState struct` and 2) [refreshed](https://github.com/Layr-Labs/eigenda/blob/03b04753e5264a0cafb96bff676ec7651b36fb2c/core/meterer/onchain_state.go#L136) periodically along with registered payments

disperser/apiserver/server.go
hopeyen: I didn't dare mess around `disperseBlob` 😬 and wanted to keep the two flows cleanly separated from each other. Additionally, `dispersePaidBlob` has these extra params ` quorumNumbers []uint32, binIndex uint32, cumulativePayment *big.Int, signature []byte`.

To make the code cleaner, I could add a `PaymentMetadata` param to `disperseBlob` and pass in an empty one when calling from `DisperseBlob` and construct a real one in `DispersePaidBlob`. What do you think?

disperser/apiserver/server.go
hopeyen: There won't ever be a case where request is limited by both meterer and rate limiter. If meterer is configured, rate limiter should not be used. 
Do we want a different type of behavior? Like, using meterer if there's `PaymentMetadata`, rate limiter if not?

core/data.go
hopeyen: thanks 😅 updated

core/auth/payment_signer.go
```diff
@@ -0,0 +1,79 @@
...
+	}
+
+	return &PaymentSigner{
+		PrivateKey: privateKey,
+	}
+}
+
+func (s *PaymentSigner) SignBlobPayment(header *commonpb.PaymentHeader) ([]byte, error) {
+	// Set the account id to the hex encoded public key of the signer
+	header.AccountId = hex.EncodeToString(crypto.FromECDSAPub(&s.PrivateKey.PublicKey))
```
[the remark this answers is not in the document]
hopeyen: It was intended to be the public key, but I've updated it to use the Eth account address

disperser/apiserver/server.go
mooselumph: > To make the code cleaner, I could add a PaymentMetadata param to disperseBlob and pass in an empty one when calling from DisperseBlob and construct a real one in DispersePaidBlob. What do you think?

I'd be in favor of something like this to reduce code duplication

disperser/apiserver/server.go
mooselumph: What about something simple like this?
- If the payment metadata is included, we use the meterer. If the meterer is unavailable, we fail. 
- If the payment metadata is not included, we use the rate limiter. If rate limiter is unavailable, pass (current behavior)

The DisperseBlobPaid method will always pass the payment metadata.

disperser/cmd/apiserver/main.go
```diff
@@ -101,6 +103,37 @@ func RunDisperserServer(ctx *cli.Context) error {
...
+		}
+
+		paymentChainState, err := mt.NewOnchainPaymentState(context.Background(), transactor)
+		if err != nil {
+			return fmt.Errorf("failed to create onchain payment state: %w", err)
+		}
+
+		offchainStore, err := mt.NewOffchainStore(
+			config.AwsClientConfig,
+			"reservations",
```
mooselumph, line 119: Are these switched to constants in a future PR by chance?

disperser/apiserver/server_test.go
```diff
@@ -647,6 +649,34 @@ func newTestServer(transactor core.Transactor) *apiserver.DispersalServer {
 	if err != nil {
 		panic("failed to create bucket store")
 	}
+	meterConfig := meterer.Config{
+		PricePerSymbol:         1,
+		MinNumSymbols:          1,
+		GlobalSymbolsPerSecond: 1000,
+		ReservationWindow:      60,
```
mooselumph: outdated?

core/meterer/meterer.go
mooselumph: Is there some kind of retry on failure mechanism?

hopeyen: nada, although I don't think a regular read from cached struct will have any failure. There's periodic polling against the on-chain contract to refresh the state, so if one poll fails, the next poll will be the next try

disperser/apiserver/server.go
hopeyen: updated! removed `dispersePaidBlob` and added `PaymentMetadata` to `disperseBlob`

disperser/apiserver/server.go
hopeyen: great, now `disperseBlob` will check for payment metadata to use metering, and then ratelimiter if it exists, otherwise continues as usual

disperser/cmd/apiserver/main.go
hopeyen: nope, I updated to use config flag (optional with defaults)

disperser/apiserver/server_test.go
hopeyen: yep, rebased just now

core/meterer/meterer.go
ian-shim: lol thought `ChainState` was this [ChainState](https://github.com/Layr-Labs/eigenda/blob/master/core/eth/state.go#L11). Should we just call this `OnchainPaymentState` or sth?

core/meterer/meterer.go
```diff
@@ -12,14 +12,6 @@ import (
 
 // Config contains network parameters that should be published on-chain. We currently configure these params through disperser env vars.
 type Config struct {
```
ian-shim, line 14: It looks like it needs to wait for 1 hour to load the onchain payment state (in `Start` L59). 
Should it refresh it immediately on start?

hopeyen: sounds good, I added a separate refresh before the ticker loop

core/meterer/meterer.go
hopeyen: yep we could do that! updated so it is more clear:)

core/meterer/onchain_state.go
```diff
@@ -101,15 +157,27 @@ func (pcs *OnchainPaymentState) GetActiveReservationByAccount(ctx context.Contex
...
 
+// GetActiveReservationByAccountOnChain returns on-chain reservation for the given account ID
+func (pcs *OnchainPaymentState) GetActiveReservationByAccountOnChain(ctx context.Context, accountID string) (core.ActiveReservation, error) {
+	blockNumber, err := pcs.tx.GetCurrentBlockNumber(ctx)
+	if err != nil {
+		return core.ActiveReservation{}, err
+	}
+	res, err := pcs.tx.GetActiveReservationByAccount(ctx, blockNumber, accountID)
+	if err != nil {
+		return core.ActiveReservation{}, errors.New("reservation account not found on-chain")
```
ian-shim, line 168: nit: wrap the actual error when returning the error, i.e. `fmt.Errorf("reservation account not found on-chain: %w", err)`

core/meterer/meterer.go
```diff
@@ -60,10 +52,15 @@ func (m *Meterer) Start(ctx context.Context) {
 		ticker := time.NewTicker(1 * time.Hour)
 		defer ticker.Stop()
 
+		// initial tick immediately upto Start
```
ian-shim, line 55: maybe we can put this outside the goroutine and immediately fail & return error if this initial sync fails?

disperser/apiserver/server.go
```diff
@@ -282,7 +283,13 @@ func (s *DispersalServer) disperseBlob(ctx context.Context, blob *core.Blob, aut
 
 	s.logger.Debug("received a new blob dispersal request", "authenticatedAddress", authenticatedAddress, "origin", origin, "blobSizeBytes", blobSize, "securityParams", strings.Join(securityParamsStrings, ", "))
 
-	if s.ratelimiter != nil {
+	// If paymentHeader is not empty, we use the meterer, otherwise we use the ratelimiter if the ratelimiter is available
+	if *paymentHeader != (core.PaymentMetadata{}) {
```
ian-shim, line 287: Need to check if paymentHeader is nil first

contracts/bindings/AVSDirectory/binding.go
```diff
@@ -39,7 +39,7 @@ type ISignatureUtilsSignatureWithSaltAndExpiry struct {
...
...ff80821115611bb757600080fd5b818501915085601f830112611bcb57600080fd5b813581811115611bda57600080fd5b866020828501011115611bec57600080fd5b60209290920196919550909350505050565b600060208284031215611c1057600080fd5b81516118c881611896565b6020808252602a908201527f6d73672e73656e646572206973206e6f74207065726d697373696f6e6564206160408201526939903ab73830bab9b2b960b11b606082015260800190565b600060208284031215611c7757600080fd5b815180151581146118c857600080fd5b60208082526028908201527f6d73672e73656e646572206973206e6f74207065726d697373696f6e6564206160408201526739903830bab9b2b960c11b606082015260800190565b60208152816020820152818360408301376000818301604090810191909152601f909201601f19160101919050565b82815260006020604081840152835180604085015260005b81811015611d3257858101830151858201606001528201611d16565b81811115611d44576000606083870101525b50601f01601f191692909201606001949350505050565b600060208284031215611d6d57600080fd5b81516001600160e01b0319811681146118c857600080fd5b60008219821115611da657634e487b7160e01b600052601160045260246000fd5b50019056fe4156534469726563746f72792e72656769737465724f70657261746f72546f41a2646970667358221220721a456eaad610451a09e43118388166ab29494af7846e6ac7b3c70777b8643864736f6c634300080c0033",
```
hopeyen, line 42: I'm not quite sure what happened here. The bindings are automatically generated when I ran lint, but the value generated is different 🤔

disperser/cmd/apiserver/main.go
ian-shim: should these be read from config?

disperser/cmd/apiserver/main.go
```diff
@@ -101,6 +103,40 @@ func RunDisperserServer(ctx *cli.Context) error {
...
+		)
+		if err != nil {
+			return fmt.Errorf("failed to create offchain store: %w", err)
+		}
+		// add some default sensible configs
+		meterer = mt.NewMeterer(
+			mtConfig,
+			&paymentChainState,
+			offchainStore,
+			logging.NewNoopLogger(),
```
ian-shim, line 135: why not reuse/inherit existing logger?

disperser/apiserver/server.go
ian-shim: sg, lets add a TODO here

ian-shim (commented):
are the bindings update intentional? I see no contract changes

core/meterer/onchain_state.go
hopeyen: thanks, updated

core/meterer/meterer.go
hopeyen: makes sense, I moved it out:)

disperser/apiserver/server.go
hopeyen: interesting 🤔 the header passed in should either be empty by `DisperseBlob` or non-empty by `DispersePaidBlob`, so thought there won't be a nil case. Added anyway for safety

ian-shim (approved):

hopeyen:
> are the bindings update intentional? I see no contract changes

this was not intentional, I'm also confused by what happened (this was from running lint, but the value of bin changes without code changes). I will revert the bin values back before merging

core/meterer/meterer.go
```diff
@@ -57,13 +49,13 @@ func NewMeterer(
 // Start starts to periodically refreshing the on-chain state
 func (m *Meterer) Start(ctx context.Context) {
 	go func() {
-		ticker := time.NewTicker(1 * time.Hour)
+		ticker := time.NewTicker(1 * time.Second)
```
mooselumph, line 52: Should this be configurable?

disperser/apiserver/server.go
```diff
@@ -209,7 +214,7 @@ func (s *DispersalServer) DisperseBlobAuthenticated(stream pb.Disperser_Disperse
 	}
 
 	// Disperse the blob
-	reply, err := s.disperseBlob(ctx, blob, authenticatedAddress, "DisperseBlobAuthenticated")
+	reply, err := s.disperseBlob(ctx, blob, authenticatedAddress, "DisperseBlobAuthenticated", &core.PaymentMetadata{})
```
mooselumph, line 217: Just pass in `nil` here?

disperser/apiserver/server.go
```diff
@@ -282,7 +283,13 @@ func (s *DispersalServer) disperseBlob(ctx context.Context, blob *core.Blob, aut
 
 	s.logger.Debug("received a new blob dispersal request", "authenticatedAddress", authenticatedAddress, "origin", origin, "blobSizeBytes", blobSize, "securityParams", strings.Join(securityParamsStrings, ", "))
 
-	if s.ratelimiter != nil {
+	// If paymentHeader is not empty, we use the meterer, otherwise we use the ratelimiter if the ratelimiter is available
+	if paymentHeader != nil && *paymentHeader != (core.PaymentMetadata{}) {
```
mooselumph, line 287: Probably just check for nil and pass in nil for the non-paid endpoint.

mooselumph (commented):
Looks good!

disperser/apiserver/server.go
hopeyen: sounds good, updated!

disperser/apiserver/server.go
hopeyen: updated!

core/meterer/meterer.go
hopeyen: yep, added to the Config and Flags

disperser/apiserver/server.go
```diff
@@ -282,7 +283,13 @@ func (s *DispersalServer) disperseBlob(ctx context.Context, blob *core.Blob, aut
 
 	s.logger.Debug("received a new blob dispersal request", "authenticatedAddress", authenticatedAddress, "origin", origin, "blobSizeBytes", blobSize, "securityParams", strings.Join(securityParamsStrings, ", "))
 
-	if s.ratelimiter != nil {
+	// If paymentHeader is not empty, we use the meterer, otherwise we use the ratelimiter if the ratelimiter is available
+	if paymentHeader != nil {
+		err := s.meterer.MeterRequest(ctx, *blob, *paymentHeader)
```
mooselumph, line 288: The blob length is passed in in the next PR

disperser/apiserver/server.go
```diff
@@ -311,6 +318,38 @@ func (s *DispersalServer) disperseBlob(ctx context.Context, blob *core.Blob, aut
 	}, nil
 }
 
+func (s *DispersalServer) DispersePaidBlob(ctx context.Context, req *pb.DispersePaidBlobRequest) (*pb.DisperseBlobReply, error) {
```
mooselumph, line 321: Should we put this code behind a feature flag? Right now, if we were to release from this code, we'd be running code here which is still incomplete. (The previous PR hits the unimplemented fallback)

disperser/apiserver/server.go
hopeyen: yep, all the changes regarding to our v1/v2 payment discussions were made in the next PR along with the integration test that initiated that discussion. I could repeat the changes here again, thinking it might make intuitive sense in the review process

disperser/apiserver/server.go
hopeyen: :thinking: like if `paymentFeatureEnabled` do this, otherwise hit unimplemented fallback? the main incomplete part is the contract, could we wait til code complete to make a release?

disperser/apiserver/server.go
mooselumph: This was mostly just a note for myself. I think it's fine as is!

disperser/apiserver/server.go
mooselumph: Could a feature flag be useful for rollout plan? I'd imagine we might want to test this on preprod and testnet before mainnet? cc @ian-shim

hopeyen: sounds good to me:) I added a check with if `meterer` is nil then the function fallback. This is because meterer is only set when `EnablePaymentMeterer` is true, and wanted to avoid adding additional config to DisperserServer

core/meterer/meterer.go
```diff
@@ -47,23 +42,23 @@ func NewMeterer(
...
 		logger: logger.With("component", "Meterer"),
 	}
 }
 
 // Start starts to periodically refreshing the on-chain state
-func (m *Meterer) Start(ctx context.Context) {
+func (m *Meterer) Start(ctx context.Context, updateInterval time.Duration) {
 	go func() {
-		ticker := time.NewTicker(1 * time.Hour)
+		ticker := time.NewTicker(updateInterval)
```
ian-shim, line 55: nit: would `updateInterval` be ever different from `m.UpdateInterval`? Should we just use `m.UpdateInterval` instead of passing it in as a param?

ian-shim (approved):

core/meterer/meterer.go
hopeyen: thanks for pointing it out! updated

mooselumph (approved):

PR #793, opened 2024-10-08
[payments] disperser client local metering

hopeyen:

Disperser client integrated with payments metering with inabox integration tests

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Integration tests


api/clients/accountant.go
```diff
@@ -0,0 +1,154 @@
...
+	close(a.stopRotation)
+}
+
+// accountant calculates and records payment information
+func (a *Accountant) BlobPaymentInfo(ctx context.Context, dataLength uint64) (uint32, *big.Int, error) {
+	//TODO: do we need to lock the binUsages here in case the blob rotation happens in the middle of the function?
+	// binUsage := a.binUsages[0] + dataLength
+	a.usageLock.Lock()
+	defer a.usageLock.Unlock()
+	a.binUsages[0] += dataLength
```
mooselumph, line 90: The way that the accountant is doing bin accounting looks to me like it might be incorrect; in particular, it might not line up with the server, since the transitions are happening at different points on the client than on the server. 

I have another idea for how we could possibly do this. Will follow up!

disperser/apiserver/server.go
```diff
@@ -285,7 +285,7 @@ func (s *DispersalServer) disperseBlob(ctx context.Context, blob *core.Blob, aut
 
 	// If paymentHeader is not empty, we use the meterer, otherwise we use the ratelimiter if the ratelimiter is available
 	if paymentHeader != nil && *paymentHeader != (core.PaymentMetadata{}) {
-		err := s.meterer.MeterRequest(ctx, *blob, *paymentHeader)
+		err := s.meterer.MeterRequest(ctx, *paymentHeader, uint(blobSize), blob.GetQuorumNumbers())
```
mooselumph, line 288: I think this needs to be converted from byte length to symbol length (https://github.com/Layr-Labs/eigenda/blob/db3da32518159617d301128d660362dc3b05b2a5/encoding/utils.go#L10)

disperser/apiserver/payment_test.go
```diff
@@ -0,0 +1,150 @@
...
+		},
+	}
+	ctx := peer.NewContext(context.Background(), p)
+
+	transactor.On("GetRequiredQuorumNumbers", tmock.Anything).Return([]uint8{0, 1}, nil).Twice()
+
+	pk := "0x0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdeb"
+	signer := auth.NewPaymentSigner(pk)
+
+	dataLength := len(data)
```
mooselumph, line 59: Similarly to the `server.go` comment, I believe we should be using the number of symbols, here, correct?

mooselumph (commented):
Looks really good! I'll continue reviewing after the discussed changes to the accountant!

api/clients/accountant.go
hopeyen: updated to use the wrapped bin array

disperser/apiserver/server.go
hopeyen: thank for the link, I've updated to use the conversion

api/clients/accountant.go
```diff
@@ -0,0 +1,158 @@
...
+		binRecords:        []BinRecord{{Index: 0, Usage: 0}, {Index: 1, Usage: 0}, {Index: 2, Usage: 0}},
+		cumulativePayment: big.NewInt(0),
+		stopRotation:      make(chan struct{}),
+		paymentSigner:     paymentSigner,
+	}
+	// TODO: add a routine to refresh the on-chain state occasionally?
+	return &a
+}
+
+// accountant calculates and records payment information
```
ian-shim, line 61: nit: `BlobPaymentInfo calculates...`

api/clients/accountant.go
```diff
@@ -0,0 +1,158 @@
...
+	reservationWindow uint32
+	pricePerSymbol    uint32
+	minNumSymbols     uint32
+
+	// local accounting
+	// contains 3 bins; circular wrapping of indices
+	binRecords        []BinRecord
+	usageLock         sync.Mutex
+	cumulativePayment *big.Int
+	stopRotation      chan struct{}
```
ian-shim, line 32: How is this used?

api/clients/accountant.go
ian-shim: can we also document what's being returned?

api/clients/accountant.go
```diff
@@ -0,0 +1,158 @@
...
+
+	paymentSigner core.PaymentSigner
+}
+
+type BinRecord struct {
+	Index uint32
+	Usage uint64
+}
+
+func NewAccountant(reservation core.ActiveReservation, onDemand core.OnDemandPayment, reservationWindow uint32, pricePerSymbol uint32, minNumSymbols uint32, paymentSigner core.PaymentSigner) *Accountant {
```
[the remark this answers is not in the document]
hopeyen: I think a lot of what's going on here will be replaced by https://github.com/Layr-Labs/eigenda/pull/835 . binRecords and cumulativePayment should change more frequently compared to the global parameters, and no one else needs access to the structs. would you help me understand what we will get by using pointers here?

api/clients/accountant.go
```diff
@@ -0,0 +1,158 @@
...
+	"math/big"
+	"sync"
+	"time"
+
+	commonpb "github.com/Layr-Labs/eigenda/api/grpc/common"
+	"github.com/Layr-Labs/eigenda/core"
+	"github.com/Layr-Labs/eigenda/core/meterer"
+)
+
+type IAccountant interface {
```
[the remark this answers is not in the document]
hopeyen: is https://github.com/Layr-Labs/eigenda/pull/847/files a fix to this? I think it looks good

api/clients/eigenda_client.go
```diff
@@ -298,6 +304,125 @@ func (m *EigenDAClient) putBlob(ctx context.Context, rawData []byte, resultChan
 	}
 }
 
+// PaidPutBlob behaves like PutBlob but with authenticated payment.
```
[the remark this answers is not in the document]
hopeyen: yep, did a similar refactoring that we did on the server side; if there's a setting on "PaymentSignerPrivateKeyHex", then we call out to `DispersePaidBlob`, otherwise `DisperseBlobAuthenticated`. lmk if this is okay?

api/clients/disperser_client.go
```diff
@@ -146,9 +127,50 @@ func (c *disperserClient) DisperseBlob(ctx context.Context, data []byte, quorums
...
+		quorumNumbers[i] = uint32(q)
+	}
+
+	// check every 32 bytes of data are within the valid range for a bn254 field element
+	_, err = rs.ToFrArray(data)
+	if err != nil {
+		return nil, nil, fmt.Errorf("encountered an error to convert a 32-bytes into a valid field element, please use the correct format where every 32bytes(big-endian) is less than 21888242871839275222246405745257275088548364400416034343698204186575808495617 %w", err)
+	}
+
+	header, signature, err := c.accountant.AccountBlob(ctx, uint64(len(data)), quorums)
```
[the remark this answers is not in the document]
hopeyen: the inner level `SignBlobPayment(protoPaymentHeader)` should error out if the header is nil so I don't expect this to happen, but added it for additional safety

api/clients/disperser_client.go
```diff
@@ -146,9 +127,50 @@ func (c *disperserClient) DisperseBlob(ctx context.Context, data []byte, quorums
 	return blobStatus, reply.GetRequestId(), nil
 }
 
-// TODO: implemented in subsequent PR
+// DispersePaidBlob disperses a blob with a payment header and signature. Similar to DisperseBlob but with signed payment header.
 func (c *disperserClient) DispersePaidBlob(ctx context.Context, data []byte, quorums []uint8) (*disperser.BlobStatus, []byte, error) {
-	return nil, nil, api.NewErrorInternal("not implemented")
```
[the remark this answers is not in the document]
hopeyen: sounds good, added it

api/clients/disperser_client.go
```diff
@@ -63,37 +63,18 @@ type disperserClient struct {
...
-	conn   *grpc.ClientConn
-	client disperser_rpc.DisperserClient
+	conn       *grpc.ClientConn
+	client     disperser_rpc.DisperserClient
+	accountant *Accountant
 }
 
 var _ DisperserClient = &disperserClient{}
 
-// DisperserClient maintains a single underlying grpc connection to the disperser server,
```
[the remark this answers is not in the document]
hopeyen: Sorry for taking it out, I'm not sure what happened, putting it back in

api/clients/accountant.go
hopeyen: outdated concept, removed:)

api/clients/accountant.go
hopeyen: updated the comments

mooselumph (commented):
Nice!

api/clients/accountant.go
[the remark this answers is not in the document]
ian-shim: Passing by values will make copies of the structs. Unless you're modifying the input structs, it's more efficient to pass by pointer

api/clients/accountant.go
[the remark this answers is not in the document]
ian-shim: Nah that one does something similar to `PaymentSigner`, not to `Accountant`

api/clients/accountant.go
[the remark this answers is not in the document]
hopeyen: New accountant would modify the input structs in later functions. Since there's no other use for the input params after being passed in, I thought it would be good for the accountant to take ownership of the params and imagined a garbage collector. Now I realize I'm thinking in rust 😅 gonna update this to be more efficient

disperser/apiserver/server_test.go
```diff
@@ -683,6 +685,22 @@ func newTestServer(transactor core.Writer, testName string) *apiserver.Dispersal
 		panic("failed to make initial query to the on-chain state")
 	}
 
+	mockState.On("GetPricePerSymbol").Return(uint32(encoding.BYTES_PER_SYMBOL), nil)
```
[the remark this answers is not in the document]
hopeyen: ha uhhh this is because the previous tests was made on the basis of charging per byte instead of per symbol, and I was lazy to redo the calculation after the blob size conversion, so I just scaled the price 😅

api/clients/accountant.go
```diff
@@ -0,0 +1,169 @@
...
+func (a *Accountant) SymbolsCharged(numSymbols uint) uint32 {
+	if numSymbols <= uint(a.minNumSymbols) {
+		return a.minNumSymbols
+	}
+	// Round up to the nearest multiple of MinNumSymbols
+	return uint32(core.RoundUpDivide(uint(numSymbols), uint(a.minNumSymbols))) * a.minNumSymbols
+}
+
+func (a *Accountant) GetRelativeBinRecord(index uint32) *BinRecord {
+	relativeIndex := index % 3
```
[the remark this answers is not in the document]
hopeyen: sure 🤔 maybe a client wants more insights into the past and use a bigger number? I added a config to EigenDA proxy client and the accountant will take the maximum between the setting and the minimum number of bins (hardcoded to 3)

api/clients/accountant.go
```diff
@@ -0,0 +1,169 @@
...
+
+	accountID := a.paymentSigner.GetAccountID()
+	pm := &core.PaymentMetadata{
+		AccountID:         accountID,
+		BinIndex:          binIndex,
+		CumulativePayment: cumulativePayment,
+	}
+	protoPaymentHeader := pm.ConvertToProtoPaymentHeader()
+
+	signature, err := a.paymentSigner.SignBlobPayment(protoPaymentHeader)
```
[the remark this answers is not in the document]
hopeyen: uhhhhh I think it was because `DispersePaidBlobRequest` gave me the proto struct. This is a good question, I thought about it a bit and added a conversion function (proto->golang) so we are now using the local struct

api/clients/accountant.go
```diff
@@ -0,0 +1,169 @@
...
+		return currentBinIndex, big.NewInt(0), nil
+	}
+
+	// reservation not available, attempt on-demand
+	//todo: rollback later if disperser respond with some type of rejection?
+	relativeBinRecord.Usage -= numSymbols
+	incrementRequired := big.NewInt(int64(a.PaymentCharged(uint(numSymbols))))
+	a.cumulativePayment.Add(a.cumulativePayment, incrementRequired)
+	if a.cumulativePayment.Cmp(a.onDemand.CumulativePayment) <= 0 {
+		if err := QuorumCheck(quorumNumbers, []uint8{0, 1}); err != nil {
```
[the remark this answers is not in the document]
hopeyen: I didn't intend on reading this from disperser endpoint. Everything else made sense to, but the quorum numbers were not included in the next PR because I was expecting them to stay as 0 and 1. alternatively, I could mark them as a hard coded constant and/or add to accountant struct.

api/clients/accountant.go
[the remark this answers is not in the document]
hopeyen: made similar changes to `accountant`

api/clients/accountant.go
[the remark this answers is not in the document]
mooselumph: I think either of these is fine. In the event that this set increased, it seems like it wouldn't be a breaking change for clients still using {0,1}. Maybe just a hard-coded constant for legibility would be good!

mooselumph (approved):

api/clients/accountant.go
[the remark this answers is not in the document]
ian-shim: Can we remove the `I` prefix now?

api/clients/disperser_client.go
```diff
@@ -89,11 +91,12 @@ var _ DisperserClient = &disperserClient{}
 //
 //	// Subsequent calls will use the existing connection
 //	status2, requestId2, err := client.DisperseBlob(ctx, otherData, otherQuorums)
-func NewDisperserClient(config *Config, signer core.BlobRequestSigner) *disperserClient {
+func NewDisperserClient(config *Config, signer core.BlobRequestSigner, accountant *accountant) DisperserClient {
```
ian-shim, line 94: nit: take the interface `Account` as the input?

core/auth/payment_signer.go
```diff
@@ -31,9 +30,10 @@ func NewPaymentSigner(privateKeyHex string) *PaymentSigner {
 }
 
 // SignBlobPayment signs the payment header and returns the signature
-func (s *PaymentSigner) SignBlobPayment(header *commonpb.PaymentHeader) ([]byte, error) {
-	header.AccountId = s.GetAccountID()
-	pm := core.ConvertPaymentHeader(header)
+// func (s *PaymentSigner) SignBlobPayment(header *commonpb.PaymentHeader) ([]byte, error) {
```
ian-shim, line 33: do we need the commented code?

api/clients/disperser_client.go
```diff
@@ -10,6 +10,7 @@ import (
```
ian-shim, line 10: This client is only compatible with v1 disperser. Do we need this change if we don't support payments in v1?

disperser/apiserver/server.go
```diff
@@ -260,7 +260,7 @@ func (s *DispersalServer) DisperseBlob(ctx context.Context, req *pb.DisperseBlob
 // to track the error again.
```
ian-shim, line 260: similarly, not necessarily in this PR, but we'd need some clean up in v1 server to remove payments

api/clients/disperser_client.go
hopeyen: I think my impression for what we want to do next is finishing the code but will not enable the feature flag?

core/auth/payment_signer.go
hopeyen: nope, removed

api/clients/disperser_client.go
ian-shim: Not sure if we need this complexity if we're not going to turn it on v1. 
Especially it changes API for any external package using this client. When we remove `accountant` from this client, they would have to update how they use this client

ian-shim (commented):
Can you address the conflicts?

api/clients/disperser_client.go
hopeyen: I don't think we would need to remove accountant from v1 client, and they would need to update for v2 anyways, and we could simply set accountant to always be nil? any way, I removed it for now and can implement for v2 later on.
I also removed the API from disperser server. if it makes the review process cleaner, I could make a separate PR that focus on removing the payment APIs

api/clients/accountant.go
```diff
@@ -0,0 +1,180 @@
...
+	if a.cumulativePayment.Cmp(a.onDemand.CumulativePayment) <= 0 {
+		if err := QuorumCheck(quorumNumbers, requiredQuorums); err != nil {
+			return 0, big.NewInt(0), err
+		}
+		return 0, a.cumulativePayment, nil
+	}
+	return 0, big.NewInt(0), fmt.Errorf("neither reservation nor on-demand payment is available")
+}
+
+// accountant provides and records payment information
```
ian-shim, line 118: nit: `AccountBlob provides...`

api/clients/disperser_client.go
```diff
@@ -78,6 +78,8 @@ type disperserClient struct {
 	//       instead of a real network connection for eg.
 	conn   *grpc.ClientConn
 	client disperser_rpc.DisperserClient
+
+	accountant Accountant
```
ian-shim, line 82: should this be removed?

ian-shim (approved):

api/clients/disperser_client.go
hopeyen: I thought it would be okay to leave as is, but removing for clarity

PR #797, opened 2024-10-09
[v2] Node v2 gRPC Entrypoint

ian-shim:
Adding entrypoint for node v2 grpc server. Both v1 and v2 endpoints are configured to the same ports. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/node/grpc/run.go
+++ b/node/grpc/run.go
@@ -1,0 +1,81 @@
+package grpc
+
+import (
+	"errors"
+	"fmt"
+	"net"
+
+	pb "github.com/Layr-Labs/eigenda/api/grpc/node"
+	pbv2 "github.com/Layr-Labs/eigenda/api/grpc/node/v2"
+	"github.com/Layr-Labs/eigenda/common/healthcheck"
+	"github.com/Layr-Labs/eigenda/node"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	"google.golang.org/grpc"
+	"google.golang.org/grpc/reflection"
+)
+
+const localhost = "0.0.0.0"
+
+func RunServers(serverV1 *Server, serverV2 *ServerV2, config *node.Config, logger logging.Logger) error {
+	if serverV1 == nil {
+		return errors.New("node V1 server is not configured")
+	}
+	if serverV2 == nil {
+		return errors.New("node V2 server is not configured")
+	}
+
+	go func() {
+		for {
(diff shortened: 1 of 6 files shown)


node/grpc/run.go
```diff
@@ -0,0 +1,81 @@
...
+
+	go func() {
+		for {
+			addr := fmt.Sprintf("%s:%s", localhost, config.InternalDispersalPort)
+			listener, err := net.Listen("tcp", addr)
+			if err != nil {
+				logger.Fatalf("Could not start tcp listener: %v", err)
+			}
+
+			opt := grpc.MaxRecvMsgSize(60 * 1024 * 1024 * 1024) // 60 GiB
```
cody-littley, line 35: Is this safe? Could an attacker force us to load 60gb into memory in order to force us to OOM?

node/grpc/run.go
```diff
@@ -0,0 +1,81 @@
...
+
+	go func() {
+		for {
+			addr := fmt.Sprintf("%s:%s", localhost, config.InternalRetrievalPort)
+			listener, err := net.Listen("tcp", addr)
+			if err != nil {
+				logger.Fatalf("Could not start tcp listener: %v", err)
+			}
+
+			opt := grpc.MaxRecvMsgSize(1024 * 1024 * 300) // 300 MiB
```
cody-littley, line 62: In the medium to long term, we might want to think about strategies for breaking up super large messages into smaller ones. If an attacker starts sending us lots of garbage data over multiple connections, 300mb in memory per connection is possibly non-trivial.

node/grpc/run.go
ian-shim: We don't have such protection on DA nodes right now (so reducing the max message size wouldn't make it any safer), and that's why they need to set up firewall and accept traffic from whitelisted origins for now.

node/grpc/run.go
ian-shim: Yes, we'd need more comprehensive defensive strategy to lift the need for firewall

cody-littley (approved):

node/grpc/run.go
jianoaix, line 35: 60GiB doesn't have effect as it's capped at 4GiB by gRPC itself

jianoaix (approved):

PR #800, opened 2024-10-10
payments vault

0x0aa0:
adds payment vault contract


contracts/src/payments/PaymentVault.sol
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * @title Entrypoint for making reservations and on demand payments for EigenDA.
+ * @author Layr Labs, Inc.
+**/
+contract PaymentVault is PaymentVaultStorage, OwnableUpgradeable {
+ 
+    constructor(
+        uint256 _reservationBinInterval,
+        uint256 _reservationBinStartTimestamp,
```
hopeyen, line 16: hey arch! Would you please remove this timestamp? I think this won't be needed as we chatted offline:)?

contracts/src/payments/PaymentVaultStorage.sol
```diff
@@ -0,0 +1,40 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+import {IPaymentVault} from "../interfaces/IPaymentVault.sol";
+
+abstract contract PaymentVaultStorage is IPaymentVault {
+    
+    /// @notice reservation bin duration 
+    uint256 public immutable reservationBinInterval;
```
hopeyen, line 9: could we also make this one mutable by onlyOwner?

mooselumph (approved):

PR #803, opened 2024-10-11
Protobufs for the relay

cody-littley:

This PR adds protobufs for the relay. WIP.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proto/relay/relay.proto
```diff
@@ -7,23 +7,95 @@ option go_package = "github.com/Layr-Labs/eigenda/api/grpc/relay";
...
+message Chunks {
+  repeated common.ChunkData chunks = 1;
 }
 
-// Reply to GetChunkRequest
-message GetChunkReply {
-	// The chunk requested.
-	common.ChunkData chunk = 1;
+// The error that occurred while trying to fetch a blob or chunks.
+enum RelayError {
```
ian-shim, line 94: Should we make the `RelayError` class a combination of an enum field + a string field for error message? Having error message will be helpful for 1) providing detailed error message, 2) representing more granular errors within error code

cody-littley: Makes sense. Done.

ian-shim (approved):

PR #806, opened 2024-10-14
Core/v2

mooselumph:

Adds a `core/v2` library which implements the core logic of v2 system.

Note: PR size will be smaller when #805 is merged. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


ian-shim:
Do you mind commenting on which files are copy paste of existing logic vs. which have logic updates? Maybe we can cover copy pastes in this PR and logic updates in separate PRs along with tests?

mooselumph:
> Do you mind commenting on which files are copy paste of existing logic vs. which have logic updates? Maybe we can cover copy pastes in this PR and logic updates in separate PRs along with tests?

Most of the copy-past is coming from #805 . I could rebase this PR to master to remove all of those delta. The tests will fail for now, but we can fix it later. Would that help? 

For this PR itself, the main copy-pastes are `core/v2/aggreation*.go`.  I would treat everything else in `core/v2` as new.

core/v2/core_test.go
```diff
@@ -0,0 +1,268 @@
...
+			BlobCommitments: commitments,
+		},
+		ReferenceBlockNumber: refBlockNumber,
+	}
+
+	return header, data
+
+}
+
+// prepareBatch takes in multiple blob, encodes them, generates the associated assignments, and the batch header.
```
hopeyen, line 116: `prepareBlobs`?

core/v2/core_test.go
```diff
@@ -0,0 +1,268 @@
...
+func TestValidationSucceeds(t *testing.T) {
+
+	// operatorCounts := []uint{1, 2, 4, 10, 30}
+
+	// numBlob := 3 // must be greater than 0
+	// blobLengths := []int{1, 32, 128}
+
+	operatorCounts := []uint{4}
+
+	numBlob := 1 // must be greater than 0
```
hopeyen, line 235: could we add more tests such as different number of blobs and blobs that would be considered invalid by the verifier?

hopeyen (commented):
I only reviewed some files by now, will look at the remaining in another pass

core/v2/types.go
```diff
@@ -0,0 +1,96 @@
...
+
+var (
+	ParametersMap = map[uint8]BlobVersionParameters{
+		0: {CodingRate: 8, ReconstructionThreshold: 0.22, NumChunks: 8192},
+	}
+)
+
+type QuorumID = uint8
+
+type OperatorID = [32]byte
```
ian-shim, line 17: Were we going to use those in `chainio`?

core/v2/types.go
```diff
@@ -0,0 +1,96 @@
...
+	}
+
+	return encoding.EncodingParams{
+		NumChunks:   uint64(params.NumChunks),
+		ChunkLength: uint64(length),
+	}, nil
+
+}
+
+type PaymentHeader struct {
```
ian-shim, line 66: we should probably just use the one in `core`?

core/v2/types.go
```diff
@@ -0,0 +1,96 @@
...
+}
+
+// BlobHeader contains all metadata related to a blob including commitments and parameters for encoding
+type BlobHeader struct {
+	Version uint8
+
+	encoding.BlobCommitments
+
+	// QuorumInfos contains the quorum specific parameters for the blob
+	QuorumNumbers []uint8
```
ian-shim, line 41: we should probably use something like `BlobQuorumInfo`

core/v2/types.go
```diff
@@ -0,0 +1,96 @@
...
+
+	// PaymentHeader contains the payment information for the blob
+	PaymentHeader
+
+	// AuthenticationData is the signature of the blob header by the account ID
+	AuthenticationData []byte `json:"authentication_data"`
```
ian-shim, line 47: authentication data is not part of blob header right?

ian-shim (commented):
@bxue-l2 can you take a look at the encoder changes?

core/v2/types.go
```diff
@@ -0,0 +1,96 @@
...
+
+import (
+	"math"
+
+	"github.com/Layr-Labs/eigenda/encoding"
+)
+
+var (
+	ParametersMap = map[uint8]BlobVersionParameters{
+		0: {CodingRate: 8, ReconstructionThreshold: 0.22, NumChunks: 8192},
```
ian-shim, line 11: How is `ReconstructionThreshold` determined? Does it not depend on onchain parameters?

core/v2/types.go
mooselumph: Why do you say this? The v2 `BlobHeader` only needs `QuorumNumbers`

core/v2/types.go
mooselumph: I think it makes sense to store it in the internal data structure, even it if isn't contained within the hash (`BlobKey`)?

core/v2/types.go
mooselumph: We can do some reconciliation once one or the other of the PRs is merged.

core/v2/types.go
ian-shim: Don't we need to know the security parameters in those quorums?

core/v2/types.go
ian-shim: Hmm does it get used after the initial request validation?

mooselumph: Will be passed onto DA nodes eventually.

core/v2/types.go
mooselumph: This is derived from the blob version.

ian-shim: I guess it's related to [this question](https://github.com/Layr-Labs/eigenda/pull/806#discussion_r1804103114). Should we get rid of the onchain parameters then?

core/v2/types.go
ian-shim: Do DA nodes validate this signature again? Then we need to think about how the interface changes there. 
Even when that's the case, I think putting this in a separate data structure like `BlobMetadata` would be cleaner. wdyt?

jianoaix:
Full copy of code generally worries me. The old code usually lives longer than people would think, and that can be an issue for maintenance.

core/v2/aggregation.go
```diff
@@ -0,0 +1,366 @@
...
+	// PercentSigned is percentage of the total stake for the quorum that signed for a particular batch.
+	PercentSigned uint8
+}
+
+// QuorumAttestation contains the results of aggregating signatures from a set of operators by quorums
+// It also returns map of all signers across all quorums
+type QuorumAttestation struct {
+	// QuorumAggPubKeys contains the aggregated public keys for all of the operators each quorum,
+	// including those that did not sign
+	QuorumAggPubKey map[QuorumID]*bn254.G1Point
```
hopeyen, line 53: plural versus singular: `QuorumAggPubKeys`?

I'm also thinking if there's better ways to arrange these structs (`QuorumResult`, `QuorumAttestation`, and `SignatureAggregation`). It probably doesn't make material differences though

ian-shim:
> Full copy of code generally worries me. The old code usually lives longer than people would think, and that can be an issue for maintenance.

I tend to agree with this. How about we create new library under `core/v2` only for those with updated logic?
@mooselumph I know you preferred not to have any dependencies on `core` in v2 components, but I think using `core` libraries with no logic changes is fine, can be helpful for maintenance, and migration remains straight forward. Any thoughts?

core/v2/validator.go
```diff
@@ -0,0 +1,199 @@
...
+		verifier:   v,
+		chainState: cst,
+		operatorID: operatorID,
+	}
+}
+
+func (v *ShardValidator) validateBlobQuorum(quorum QuorumID, blob *BlobShard, operatorState *chainio.OperatorState) ([]*encoding.Frame, *Assignment, error) {
+
+	// Check if the operator is a member of the quorum
+	if _, ok := operatorState.Operators[quorum]; !ok {
```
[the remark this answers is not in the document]
mooselumph: It looks like both of these checks are done in the next function call. I'll add a comment and remove this check.

encoding/kzg/prover/prover_cpu.go
```diff
@@ -21,8 +21,13 @@ import (
 
 func (g *Prover) newProver(params encoding.EncodingParams) (*ParametrizedProver, error) {
 
-	// Check that the parameters are valid with respect to the SRS.
-	if params.ChunkLength*params.NumChunks >= g.SRSOrder {
+	// Check that the parameters are valid with respect to the SRS. The precomputed terms of the amortized KZG
+	// prover use up to order params.ChunkLen*params.NumChunks-1 for the SRS, so we must have
+	// params.ChunkLen*params.NumChunks-1 <= g.SRSOrder. The condition below could technically
```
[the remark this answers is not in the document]
mooselumph: The constraint is that `params.ChunkLen*params.NumChunks-1 <= g.SRSOrder`, so the error case (below) is when the inequality is flipped (`>`)!

encoding/kzg/prover/parametrized_prover.go
```diff
@@ -67,25 +67,58 @@ func (g *ParametrizedProver) Encode(inputFr []fr.Element) (*bn254.G1Affine, *bn2
 
 	encodeStart := time.Now()
 
-	rsChan := make(chan RsEncodeResult, 1)
-	lengthCommitmentChan := make(chan LengthCommitmentResult, 1)
-	lengthProofChan := make(chan LengthProofResult, 1)
-	commitmentChan := make(chan CommitmentResult, 1)
-	proofChan := make(chan ProofsResult, 1)
+	type commitments struct {
```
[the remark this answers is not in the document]
mooselumph: I think it might be suitable since it's only being used here. I moved it out for consistency!

core/v2/assignment.go
```diff
@@ -0,0 +1,124 @@
...
+		index += a.chunks
+	}
+
+	return assignments, nil
+
+}
+
+func GetAssignment(state *chainio.OperatorState, blobVersion byte, quorum QuorumID, id core.OperatorID) (Assignment, error) {
+
+	assignments, err := GetAssignments(state, blobVersion, quorum)
```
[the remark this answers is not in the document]
mooselumph: I think this is basically a convenience function meant to be called by an operator who only needs the assignemnt for itself.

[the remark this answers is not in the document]
mooselumph: Maybe we could just remove it.

PR #807, opened 2024-10-15
Add feature that can disable traffic generator metrics.

cody-littley:

We may not want all metrics to be enabled for the traffic generator.

## Checks

- Testing Strategy
   - [x] hand tests 
   - [ ] Unit tests
   - [ ] Integration tests


tools/traffic/metrics/metrics.go
```diff
@@ -31,22 +32,39 @@ type metrics struct {
...
 	reg := prometheus.NewRegistry()
 	reg.MustRegister(collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}))
 	reg.MustRegister(collectors.NewGoCollector())
 
+	if metricsBlacklist == nil {
+		metricsBlacklist = []string{}
+	}
+	if metricsFuzzyBlacklist == nil {
+		metricsFuzzyBlacklist = []string{}
+	}
```
ian-shim, line 57: nit: I don't think these are needed

tools/traffic/config/config.go
```diff
@@ -106,6 +106,9 @@ func NewConfig(ctx *cli.Context) (*Config, error) {
 			EigenDAServiceManager: retrieverConfig.EigenDAServiceManagerAddr,
 			SignerPrivateKey:      ctx.String(SignerPrivateKeyFlag.Name),
 			CustomQuorums:         customQuorumsUint8,
+
+			MetricsBlacklist:      ctx.StringSlice(MetricsBlacklistFlag.Name),
```
ian-shim, line 110: why do we need both flags? 
could we just use `MetricsFuzzyBlacklist`? not sure what use cases there are where we want exact matching vs. fuzzy matching (fuzzy matching can cover all use cases we need for exact matching)

tools/traffic/metrics/count_metric.go
```diff
@@ -14,14 +14,19 @@ type CountMetric interface {
 type countMetric struct {
 	metrics     *metrics
 	description string
+	// If true, the metric is disabled and should behave as a no-op.
```
ian-shim, line 17: nit: `disabled specifies whether the metrics should behave as a no-op` or something like that

ian-shim (approved):
lgtm, few comments

tools/traffic/metrics/count_metric.go
cody-littley: wording updated

tools/traffic/config/config.go
cody-littley: Fuzzy matching is needed to handle metric labels that contain DA node IDs. Exact matching is needed to disable metrics where one metric is a substring of another. For example, there are two metrics labeled `read` and `read_success`. Without exact matching, we couldn't disable the `read` metric without also disabling the `read_success` metric.

The only straight forward way to avoid having two lists would be to interpret each entry as a regex. In your opinion, would you prefer to have a single list of regex blacklist arguments, or the current pattern?

ian-shim: I see. That makes sense. I think the current pattern is simpler in that case

PR #810, opened 2024-10-16
Adds operator ejections dataapi endpoint

pschork:
- Default look back is 1 day.
- Max lookback is 90 days
- Optional `operator_id` can be specified to limit ejection query to specific operator.

```
curl -X 'GET' \
  'https://dataapi-preprod-holesky.eigenda.xyz/api/v1/operators-info/operator-ejections?days=90' \
  -H 'accept: application/json'
{
  "ejections": [
    {
      "operator_id": "0x45fc52054b0b4b95f9188959eac542abf08537b62575fb0c89a8e42a77a93497",
      "quorum": 1,
      "block_number": 1956080,
      "block_timestamp": "2024-07-19T01:48:48Z",
      "transaction_hash": "0x3d0e493e09e960a8d373fdb904ab7cc6ece5d7f7385602a791e14aa44b2fed2e"
    },
    {
      "operator_id": "0x45fc52054b0b4b95f9188959eac542abf08537b62575fb0c89a8e42a77a93497",
      "quorum": 2,
      "block_number": 1956080,
      "block_timestamp": "2024-07-19T01:48:48Z",
      "transaction_hash": "0x3d0e493e09e960a8d373fdb904ab7cc6ece5d7f7385602a791e14aa44b2fed2e"
    },
    {
      "operator_id": "0xf512db9a07eefc22336c0de9f0dd8cb71400d0718e9ed00cb4af2643dc64325e",
      "quorum": 1,
      "block_number": 2126464,
      "block_timestamp": "2024-08-13T18:47:12Z",
      "transaction_hash": "0x23ba5240a21cb7e56a8deb971833d188ff5e30562bed5fb70e44b2a624809367"
    },
    {
      "operator_id": "0x45fc52054b0b4b95f9188959eac542abf08537b62575fb0c89a8e42a77a93497",
      "quorum": 1,
      "block_number": 2290044,
      "block_timestamp": "2024-09-07T18:47:12Z",
      "transaction_hash": "0x72044cb9bd5a05c5de466d698ffeff475d2e2e2df104e73b67bfac72bcc7814a"
    },
    {
      "operator_id": "0x45fc52054b0b4b95f9188959eac542abf08537b62575fb0c89a8e42a77a93497",
      "quorum": 2,
      "block_number": 2290044,
      "block_timestamp": "2024-09-07T18:47:12Z",
      "transaction_hash": "0x72044cb9bd5a05c5de466d698ffeff475d2e2e2df104e73b67bfac72bcc7814a"
    },
    {
      "operator_id": "0x45fc52054b0b4b95f9188959eac542abf08537b62575fb0c89a8e42a77a93497",
      "quorum": 1,
      "block_number": 2329091,
      "block_timestamp": "2024-09-13T18:47:24Z",
      "transaction_hash": "0x987fcdd225d72f2f22f9ae4eb32eaeb2f52dc301e585e7c1c503382c8c0e07cd"
    },
    {
      "operator_id": "0x45fc52054b0b4b95f9188959eac542abf08537b62575fb0c89a8e42a77a93497",
      "quorum": 2,
      "block_number": 2329091,
      "block_timestamp": "2024-09-13T18:47:24Z",
      "transaction_hash": "0x987fcdd225d72f2f22f9ae4eb32eaeb2f52dc301e585e7c1c503382c8c0e07cd"
    }
  ]
}
```

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/dataapi/server.go
+++ b/disperser/dataapi/server.go
@@ -878,7 +878,9 @@ func (s *server) FetchRegisteredOperators(c *gin.Context) {
 //	@Tags		OperatorsInfo
 //	@Produce	json
 //	@Param		days		query		int		false	"Lookback in days [default: 1]"
-//	@Param		operator_id	query		string	false	"Operator ID filter [default: all operators]"
+//	@Param		operator_id	query		string	false	"Operator ID filter"
+//	@Param		first		query		int		false	"Return first N ejections [default: 1000]"
+//	@Param		skip		query		int		false	"Skip first N ejections [default: 0]"
 //	@Success	200			{object}	QueriedOperatorEjectionsResponse
 //	@Failure	400			{object}	ErrorResponse	"error: Bad request"
 //	@Failure	404			{object}	ErrorResponse	"error: Not found"
@@ -891,9 +893,8 @@ func (s *server) FetchOperatorEjections(c *gin.Context) {
 	defer timer.ObserveDuration()
 
 	operatorId := c.DefaultQuery("operator_id", "") // If not specified, defaults to all operators
-	days := c.DefaultQuery("days", "1")             // If not specified, defaults to 1
 
-	// Convert days to integer
+	days := c.DefaultQuery("days", "1") // If not specified, defaults to 1
 	daysInt, err := strconv.Atoi(days)
 	if err != nil {
 		c.JSON(http.StatusBadRequest, gin.H{"error": "Invalid 'days' parameter"})
@@ -905,7 +906,31 @@ func (s *server) FetchOperatorEjections(c *gin.Context) {
 		return
 	}
 
-	operatorEjections, err := s.getOperatorEjections(c.Request.Context(), int32(daysInt), operatorId)
+	first := c.DefaultQuery("first", "1000") // If not specified, defaults to 1000
+	firstInt, err := strconv.Atoi(first)
+	if err != nil {
+		c.JSON(http.StatusBadRequest, gin.H{"error": "Invalid 'first' parameter"})
+		return
+	}
+
+	if firstInt < 1 || firstInt > 10000 {
+		c.JSON(http.StatusBadRequest, gin.H{"error": "Invalid 'first' parameter. Value must be between 1..10000"})
+		return
+	}
+
+	skip := c.DefaultQuery("skip", "0") // If not specified, defaults to 0
+	skipInt, err := strconv.Atoi(skip)
+	if err != nil {
+		c.JSON(http.StatusBadRequest, gin.H{"error": "Invalid 'skip' parameter"})
+		return
+	}
+
+	if skipInt < 0 || skipInt > 1000000000 {
+		c.JSON(http.StatusBadRequest, gin.H{"error": "Invalid 'skip' parameter. Value must be between 0..1000000000"})
+		return
+	}
+
+	operatorEjections, err := s.getOperatorEjections(c.Request.Context(), int32(daysInt), operatorId, uint(firstInt), uint(skipInt))
 	if err != nil {
 		s.logger.Error("Failed to fetch ejected operators", "error", err)
 		s.metrics.IncrementFailedRequestNum("FetchOperatorEjections")
--- a/disperser/dataapi/server.go
+++ b/disperser/dataapi/server.go
@@ -900,8 +900,8 @@ func (s *server) FetchOperatorEjections(c *gin.Context) {
 		return
 	}
 
-	if daysInt > 30 {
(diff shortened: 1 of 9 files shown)


disperser/dataapi/server.go
```diff
@@ -859,6 +872,54 @@ func (s *server) FetchRegisteredOperators(c *gin.Context) {
 	})
 }
 
+// FetchOperatorEjections godoc
+//
+//	@Summary	Fetch list of operator ejections over last N days.
+//	@Tags		OperatorsInfo
+//	@Produce	json
+//	@Param		days		query		int		false	"Lookback in days [default: 1]"
+//	@Param		operator_id	query		string	false	"Operator ID filter"
```
jianoaix, line 881: Also add a "[default: all operators]"?

disperser/dataapi/subgraph_client.go
```diff
@@ -84,6 +85,14 @@ type (
 	IndexedQueriedOperatorInfo struct {
 		Operators map[core.OperatorID]*QueriedOperatorInfo
 	}
+
+	//IndexedQueriedOperatorEjections struct {
```
jianoaix, line 89: remove if not used

disperser/dataapi/subgraph/api.go
```diff
@@ -209,6 +211,37 @@ func (a *api) QueryOperatorInfoByOperatorIdAtBlockNumber(ctx context.Context, op
...
+}
+
+func (a *api) QueryOperatorEjectionsGteBlockTimestamp(ctx context.Context, blockTimestamp uint64) ([]*OperatorEjection, error) {
+	var (
+		query     queryOperatorEjectedsGteBlockTimestamp
+		variables = map[string]any{
+			"blockTimestamp_gte": graphql.Int(blockTimestamp),
+		}
+	)
+	err := a.operatorStateGql.Query(context.Background(), &query, variables)
```
jianoaix, line 237: It may have a limit of how many items to return, if there are more ejections than that, it'll cut off (hence not returning e.g. 90 days worth of ejections)

jianoaix (commented):
It will be useful to also return whether the ejection hit rate limiting which is logged as well cc @0x0aa0

bxue-l2 (approved):

disperser/dataapi/server.go
```diff
@@ -859,6 +872,54 @@ func (s *server) FetchRegisteredOperators(c *gin.Context) {
 	})
 }
 
+// FetchOperatorEjections godoc
+//
+//	@Summary	Fetch list of operator ejections over last N days.
+//	@Tags		OperatorsInfo
+//	@Produce	json
+//	@Param		days		query		int		false	"Lookback in days [default: 1]"
```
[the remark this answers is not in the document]
jianoaix: It can be any number of times per day. And the query can be interested in a number of hours.

disperser/dataapi/server.go
```diff
@@ -859,6 +872,74 @@ func (s *server) FetchRegisteredOperators(c *gin.Context) {
...
+//	@Param		operator_id	query		string	false	"Operator ID filter [default: all operators]"
+//	@Param		first		query		int		false	"Return first N ejections [default: 1000]"
```
jianoaix, line 882: I'm not sure we want to expose these two at API, they can be internal controls

pschork: The reason subgraph has these is to force pagination on the caller to limit response from getting too big. I think we should pass pagination onto our callers as well. We shouldn't do pagination for them and thus return a huge response payload.
I considered setting internal limits and not exposing these, but it seems like we should allow people to find every ejection that has ever occurred - in which case we should just provide them the tools to do it.

disperser/dataapi/server.go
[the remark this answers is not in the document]
pschork: Whats the use case where you only want ejections in the last hour, but don't want ejections from the previous 24hr, and you are unable to filter results by block timestamp?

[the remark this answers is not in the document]
jianoaix: It's all possible, but just adding friction

disperser/dataapi/server.go
jianoaix: This deviates from the existing API norms and creates inconsistent API experience.
Adding more parameters also complicates things, even this implementation of pagination is not quite correct (a second call will is not strictly relative to the first one). I would prefer a simpler semantics with a caveat that it'll provide at most MIN{N, pastNumHoursWindow} items.

PR #811, opened 2024-10-16
TTL + table store

cody-littley:

This PR unifies the concepts of a "table store" and a "ttl store". Table stores now support TTLs, and there is no stand alone ttl store.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests


common/kvstore/tablestore/table_view.go
```diff
@@ -5,14 +5,17 @@ import (
 	"github.com/Layr-Labs/eigenda/common/kvstore"
 	"github.com/syndtr/goleveldb/leveldb/iterator"
 	"github.com/syndtr/goleveldb/leveldb/util"
+	"time"
 )
 
 var _ kvstore.Table = &tableView{}
 
 // tableView allows data in a table to be accessed as if it were a Store.
```
jianoaix, line 13: Table view tries to appear like a store, by wrapping two stores...these are getting too complicated and won't be simple to use.

common/kvstore/ttl_store.go
```diff
@@ -2,6 +2,12 @@ package kvstore
 
 import "time"
 
+// TTLStoreBatch is a collection of key / value pairs that will be written atomically to a database with
+// time-to-live (TTL) or expiration times. Although it is thread safe to modify different batches in
+// parallel or to modify a batch while the store is being modified, it is not thread safe to concurrently
+// modify the same batch.
+type TTLStoreBatch TTLBatch[[]byte]
+
```
jianoaix, line 10: since you are trying to merge ttl into tablestore, can these be deleted?

common/kvstore/tablestore/table_view.go
cody-littley: This isn't actually wrapping two stores, this class has access to two different interfaces on the same store. This constructor isn't publicly exported, so it's impossible for an end user to instantiate this class with two different stores.

That being said, in order to simplify this method, I've modified the function signature to make things easier to read. 
The reason why this class needed access to the parent `TableStore` class was for the methods `Shutdown()`, `Destroy()`, and `NewBatch()`. Instead of passing in the whole `TableStore`, we can just pass in lambdas for the functionality needed by the table view.

```
func newTableView(
	base kvstore.Store,
	name string,
	prefix uint32,
	shutdown func() error,
	destroy func() error,
	batchBuilder func() kvstore.TableStoreBatch) kvstore.Table {
```

common/kvstore/batch.go
```diff
@@ -0,0 +1,32 @@
...
+	Apply() error
+	// Size returns the number of operations in the batch.
+	Size() uint32
+}
+
+// TTLBatch is a collection of key / value pairs that will be written atomically to a database with
+// time-to-live (TTL) or expiration times. Although it is thread safe to modify different batches in
+// parallel or to modify a batch while the store is being modified, it is not thread safe to concurrently
+// modify the same batch.
+type TTLBatch[K any] interface {
```
jianoaix, line 24: It looks this is not needed?
- if it's a `Table`, there is `TableBatch`, operating on the raw keys
- if it's a `TableStore`, there is `TableStoreBatch`, operating on the namespaced keys

And the ttl related interfaces are always there.

common/kvstore/tablestore/table_view.go
jianoaix: The specific issue here is that `tableView` doesn't appear to be Store, it's trying to be a `Table` (e.g. the TTL related interfaces do not exist for Store)

common/kvstore/tablestore/table_view.go
```diff
@@ -5,30 +5,43 @@ import (
...
 type tableView struct {
-	// base is the underlying store.
+	// base is the underlying base store.
 	base kvstore.Store
 	// name is the name of the table.
 	name string
 	// prefix is the prefix for all keys in the table.
 	prefix uint32
+	// shutdown is a function that shuts down the table store.
```
jianoaix, line 21: Why the following 3 interfaces make sense to be here? They are all operating as the TableStore, not Table

common/kvstore/tablestore/table_store_builder.go
```diff
@@ -375,22 +371,46 @@ func loadNamespaceTable(namespaceTable kvstore.Table) (map[uint32]string, error)
 	return tableIDMap, nil
 }
 
-// CreateTable creates a new table with the given name. If a table with the given name already exists,
-// this method becomes a no-op. Returns ErrTableLimitExceeded if the maximum number of tables has been reached.
+// CreateTable creates a new table with the given name.
```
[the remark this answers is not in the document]
cody-littley: fixed

common/kvstore/tablestore/table_view.go
cody-littley: The interface `Table` currently extends the interface `Store`. A user can interact with a table as if it were a regular `Store` object.

```
type Table interface {
	Store

	Name() string
	TableKey(key []byte) TableKey
	PutWithTTL(key []byte, value []byte, ttl time.Duration) error
	PutWithExpiration(key []byte, value []byte, expiryTime time.Time) error
	NewTTLBatch() TTLStoreBatch
}
```

common/kvstore/tablestore/table_view.go
cody-littley: The core problem is that `Table` implements `Store`, which means it needs to implement methods like `Shutdown()`. If you'd like to change this, I can think of the following options:

- make methods like `Shutdown()` into no-ops
- make methods like `Shutdown()` return an error
- split the `Store` interface into several interfaces:
    - the interface called `Store` will just have the methods `Get()`, `Put()`, `Delete()`, etc.
    - There will be an interface called `DBStore` (or similar) that implements `Store` and adds methods such as `Shutdown()`
    - Things such as `LevelDBStore` will implement `DBStore`, while `Table` will only implement `Store`. `Table` will no longer have odd methods such as `Shutdown()`.
 - Replace `Table`s with key builders (I don't think you like this option, mentioning it for completeness 😜)

common/kvstore/batch.go
cody-littley: The reason why we need a batch with TTLs and a batch without TTLs is because `Store` supports batches but has no concept of TTLs

In the previous PR, you requested that I have a `Batch[K any]` interface that is inherited by the subtypes `StoreBatch[[]byte]` and `TableStoreBatch[TableKey]`. The generic interface `TTLBatch[K any] is a continuation of that pattern.

I agree that these interfaces are currently kind of clunky. But I'm not sure how to make them better while still respecting some of your prior requests. We should discuss strategies for improving this.

cody-littley: The batch APIs have now been simplified so that they don't have generics. Are there additional simplifications you'd like to see?

cody-littley: Note that we still need two batch interfaces if we want to support batches that can set different TTLs for operations within the same batch. The core reason is that the basic `Store` API does not have a concept of a TTL, meaning we need a batch interface that doesn't have TTLs.

common/kvstore/tablestore/table_view.go
jianoaix: I'm open to either option if it makes sense. I think the "weirdness" in APIs may indicate we could structure them better (like the case here). Do you prefer using key builder? Option 3 seems complicated to me.

common/kvstore/key.go
```diff
@@ -0,0 +1,49 @@
+package kvstore
+
+import "errors"
+
+// ErrInvalidKey is returned when a key cannot be interpreted as the requested type.
+var ErrInvalidKey = errors.New("invalid key")
+
+// Key represents a key in a TableStore. Each key is scoped to a specific table.
+type Key interface {
+	// AsString interprets the key as a string and returns it.
```
jianoaix, line 10: These are a lot of types, can we start with what we know we need?

jianoaix (commented):
lgtm mostly

common/kvstore/key.go
cody-littley: I've removed all of the methods except for the ones that deal with bytes. The other methods were not strictly necessary, just syntactic sugar.

jianoaix (approved):

PR #816, opened 2024-10-17
Update v2 disperser protos

ian-shim:
Updating blob header & cert schema 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proto/common/common.proto
```diff
@@ -29,13 +29,20 @@ message BlobCommitment {
...
 // BlobCertificate is what gets attested by the network
 message BlobCertificate {
-	uint32 version = 1;
-	bytes blob_key = 2;
-	BlobCommitment blob_commitment = 3;
-	repeated uint32 quorum_numbers = 4;
+	bytes blob_key = 1;
+	BlobHeader blob_header = 2;
 	uint32 reference_block_number = 5;
+	repeated string relays = 4;
```
cody-littley, line 45: We need to decide what form the relay identifier will take. If relays are registered on chain then we can use a short identifier, e.g. a `uint32` or similar. Passing the relay URLs here will be convenient but could bloat the size of the certificate.

api/proto/common/common.proto
```diff
@@ -29,13 +29,20 @@ message BlobCommitment {
...
+	bytes payment_header_hash = 4;
+}
+
 // BlobCertificate is what gets attested by the network
 message BlobCertificate {
-	uint32 version = 1;
-	bytes blob_key = 2;
-	BlobCommitment blob_commitment = 3;
-	repeated uint32 quorum_numbers = 4;
+	bytes blob_key = 1;
```
cody-littley, line 42: Is the blob key going to be equivalent to the blob hash? If "yes", perhaps if we should just call this field `blob_hash`.

ian-shim: `blob_key` is used throughout the system as a `hash(blob_header)`. What's the blob hash?

cody-littley: Makes sense, I think I was mistaken about the blob hash thing.

api/proto/common/common.proto
ian-shim: updated

api/proto/common/v2/common.proto
```diff
@@ -0,0 +1,24 @@
...
+import "common/common.proto";
+option go_package = "github.com/Layr-Labs/eigenda/api/grpc/common/v2";
+
+
+message BlobHeader {
+	// Blob version
+	uint32 version = 1;
+	repeated uint32 quorum_numbers = 2;
+	common.BlobCommitment commitment = 3;
+	bytes payment_header_hash = 4;
```
mooselumph, line 12: Should we just include the payment header now so as to minimize api changes in the future?

api/proto/common/v2/common.proto
```diff
@@ -0,0 +1,24 @@
...
+
+	// signature over keccak hash of the blob_header that can be verified by blob_header.account_id
+	bytes signature = 5;
+}
+
+// BlobCertificate is what gets attested by the network
+message BlobCertificate {
+	bytes blob_key = 1;
```
mooselumph, line 20: Is the blob key needed when the blob header is here?

mooselumph (approved):

PR #817, opened 2024-10-17
Add traffic generator 2 docker file.

cody-littley:
This PR adds a new docker file for the new traffic generator. Since I do not want to delete the old traffic generator yet, this should not yet replace the existing traffic generator.


dmanc (approved):
lgtm

PR #820, opened 2024-10-19
ci: add eigenda-client testnet-integration test to integration-tests.yml workflow

samlaf:

Not sure if this was not included on purpose (possible since this test might take a while to run? depending on batching interval on testnet), but added in case.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -28,7 +28,7 @@ jobs:
 
       - uses: actions/setup-go@v3
         with:
-          go-version: '1.21' # The Go version to download (if necessary) and use.
+          go-version: "1.21" # The Go version to download (if necessary) and use.
       - run: go version
 
       - name: Checkout EigenDA
@@ -78,6 +78,9 @@ jobs:
 
       - name: Node Plugin
         run: go test -v ./node/plugin/tests
+
+      - name: eigenda-client
+        run: go test -v ./api/clients --testnet-integration
 
       - name: Graph Indexer
         run: make build && go test -v ./core/thegraph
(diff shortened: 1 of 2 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -79,6 +79,9 @@ jobs:
       - name: Node Plugin
         run: go test -v ./node/plugin/tests
 
+      - name: eigenda-client
```
jianoaix, line 82: How long does it take to run? If too slow, we may only run it on master branch i.e. when PR is merged.

jianoaix (commented):
lgtm

.github/workflows/integration-tests.yml
bxue-l2: @samlaf

samlaf: hmm depends on batcher basically. batching interval is 10 mins on holesky testnet right now?
I think we should remove waiting for finalization.. that's just dumb to test I feel, and adds a 12 minute interval. And its testing the disperser itself more so than the client.

samlaf: set waitForFinalization=false: [eca995e](https://github.com/Layr-Labs/eigenda/pull/820/commits/eca995e4700fb173a95deceec46ce14ed084e55e)
Ran locally and it took <2mins. Let's see here. Think it depends on when in batching interval we send the request, but should hopefully be < 10 mins otherwise we'll need to increase default 10min golang test limit just for this test.

ian-shim (approved):

PR #821, opened 2024-10-19
feat!: eigenda client confirmation depth

samlaf:
Closes EBE-107.

Starting to upstream some of the eigenda-proxy features to eigenda-client.
This PR is needed for https://github.com/Layr-Labs/eigenda-proxy/issues/187

Did not add a unit test, as I'm not a big fan of mock-based tests, which is the only pattern we have currently to test the eigenda-client. Added an e2e test on holesky instead, but this does not seem to be checked by any of our ci workflows... PR incoming (https://github.com/Layr-Labs/eigenda/pull/820 ).

BREAKING CHANGES:
- eth-rpc-url and eigenda-service-manager address are now mandatory configs to instantiate the eigenda-client

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/config_test.go
+++ b/api/clients/config_test.go
@@ -106,52 +106,6 @@ func TestEigenDAClientConfig_CheckAndSetDefaults(t *testing.T) {
 		}
 	})
 
-	t.Run("WaitForFinalization and WaitForConfirmationDepth interaction", func(t *testing.T) {
-		testCases := []struct {
-			name                   string
-			waitForFinalization    bool
-			confirmationDepth      uint64
-			shouldWarnFinalization bool
-			shouldWarnDepthTooHigh bool
-		}{
-			{
-				name:                   "WaitForFinalization true ignores depth",
-				waitForFinalization:    true,
-				confirmationDepth:      10,
-				shouldWarnFinalization: true,
-				shouldWarnDepthTooHigh: false,
-			},
-			{
-				name:                   "High confirmation depth without finalization",
-				waitForFinalization:    false,
-				confirmationDepth:      25,
-				shouldWarnFinalization: false,
-				shouldWarnDepthTooHigh: true,
-			},
-			{
-				name:                   "Normal confirmation depth",
-				waitForFinalization:    false,
-				confirmationDepth:      12,
-				shouldWarnFinalization: false,
-				shouldWarnDepthTooHigh: false,
-			},
-		}
-
-		for _, tc := range testCases {
-			t.Run(tc.name, func(t *testing.T) {
-				config := newValidConfig()
-				config.WaitForFinalization = tc.waitForFinalization
-				config.WaitForConfirmationDepth = tc.confirmationDepth
-				err := config.CheckAndSetDefaults()
-				require.NoError(t, err)
-
-				// Note: In a real implementation, you might want to use a custom logger
-				// to capture and verify the warning messages. This test structure
-				// demonstrates what we would verify if we had that capability.
-			})
-		}
-	})
-
 	t.Run("Custom timeouts", func(t *testing.T) {
 		config := newValidConfig()
 		customRetryInterval := 10 * time.Second
--- a/api/clients/config_test.go
+++ b/api/clients/config_test.go
@@ -1,0 +1,186 @@
+package clients
+
+import (
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+	"testing"
+	"time"
+)
+
+// Claude generated tests... don't blame the copy paster.
+func TestEigenDAClientConfig_CheckAndSetDefaults(t *testing.T) {
+	// Helper function to create a valid base config
+	newValidConfig := func() *EigenDAClientConfig {
+		return &EigenDAClientConfig{
+			RPC:            "http://localhost:8080",
+			EthRpcUrl:      "http://localhost:8545",
+			SvcManagerAddr: "0x1234567890123456789012345678901234567890",
+		}
+	}
+
+	t.Run("Valid minimal configuration", func(t *testing.T) {
+		config := newValidConfig()
+		err := config.CheckAndSetDefaults()
+		require.NoError(t, err)
+
+		// Check default values are set
+		assert.Equal(t, 5*time.Second, config.StatusQueryRetryInterval)
+		assert.Equal(t, 25*time.Minute, config.StatusQueryTimeout)
+		assert.Equal(t, 30*time.Second, config.ResponseTimeout)
+	})
+
+	t.Run("Missing required fields", func(t *testing.T) {
+		testCases := []struct {
+			name        string
+			modifyConf  func(*EigenDAClientConfig)
+			expectedErr string
+		}{
+			{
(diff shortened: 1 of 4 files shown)


api/clients/config_test.go
```diff
@@ -0,0 +1,186 @@
...
+			},
+		}
+
+		for _, tc := range testCases {
+			t.Run(tc.name, func(t *testing.T) {
+				config := newValidConfig()
+				config.WaitForFinalization = tc.waitForFinalization
+				config.WaitForConfirmationDepth = tc.confirmationDepth
+				err := config.CheckAndSetDefaults()
+				require.NoError(t, err)
```
bxue-l2, line 146: since warning isn't error, so the test won't fail. Then it relies on human to catch it?

api/clients/eigenda_client.go
```diff
@@ -265,3 +301,33 @@ func (m *EigenDAClient) putBlob(ctx context.Context, rawData []byte, resultChan
...
+// getConfDeepBlockNumber returns the block number that is `depth` blocks behind the current block number.
+func (m EigenDAClient) getConfDeepBlockNumber(ctx context.Context, depth uint64) (*big.Int, error) {
+	curBlockNumber, err := m.ethClient.BlockNumber(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get latest block number: %w", err)
+	}
+	// If curBlock < depth, this will return the genesis block number (0),
+	// which would cause to accept as confirmed a block that isn't depth deep.
+	// TODO: there's prob a better way to deal with this, like returning a special error
+	return new(big.Int).SetUint64(max(curBlockNumber-depth, 0)), nil
```
bxue-l2, line 314: nix, since depth is uint64, logically, it can underflow

api/clients/config_test.go
samlaf: Good point... this whole test file was Claude generated. I removed this specific test as its just a warning like you said. [5439149](https://github.com/Layr-Labs/eigenda/pull/821/commits/5439149e745a9f5420a418da1e79043f9329c73e)

api/clients/eigenda_client.go
samlaf: good find: [32a6095](https://github.com/Layr-Labs/eigenda/pull/821/commits/32a60954433a474ce12fc82f7be77d67949dee13)

api/clients/eigenda_client_e2e_test.go
```diff
@@ -18,29 +19,71 @@ func init() {
...
-		// because we already test for this correct behavior in the unit tests using a mock disperser
-		// which is much faster.
-		WaitForFinalization: false,
+
+	t.Run("PutBlobWaitForConfirmationDepth0AndGetBlob", func(t *testing.T) {
+		t.Parallel()
+		logger := log.NewLogger(log.NewTerminalHandler(os.Stdout, true))
+		client, err := NewEigenDAClient(logger, EigenDAClientConfig{
+			RPC:                      "disperser-holesky.eigenda.xyz:443",
+			StatusQueryTimeout:       25 * time.Minute,
```
bxue-l2, line 35: should we comment with an equation on how to compute this depth.
Say batching interval is 3 mins, then attestation timeout is 80% of that. And writing to chain timeout is 72sec. So overall below 6 mins.

api/clients/eigenda_client_e2e_test.go
```diff
@@ -18,29 +19,71 @@ func init() {
...
+		batchHeaderHash := blobInfo.BlobVerificationProof.BatchMetadata.BatchHeaderHash
+		blobIndex := blobInfo.BlobVerificationProof.BlobIndex
+		blob, err := client.GetBlob(context.Background(), batchHeaderHash, blobIndex)
+		assert.NoError(t, err)
+		assert.Equal(t, data, string(blob))
+
+		// assert confirmation depth
+		blockNumCur, err := client.ethClient.BlockNumber(context.Background())
+		assert.NoError(t, err)
+		blockNumAtDepth := new(big.Int).SetUint64(blockNumCur - confDepth)
```
bxue-l2, line 83: how do you check at least confDepth is waited?

bxue-l2 (approved):

api/clients/eigenda_client_e2e_test.go
samlaf: wdym by attestation timeout? Is that the thing for 503? Haven't added that yet actually! So good reminder to add it, added comment in the 503 PR.

Why is your writing to chain timeout 72 sec here? Just random? It's only 3 confirmations, so should be 36 secs just waiting for ethereum, so you doubled that in case?

api/clients/eigenda_client_e2e_test.go
samlaf: Extended the comment: [b29d79a](https://github.com/Layr-Labs/eigenda/pull/821/commits/b29d79a7c017c832f4ff9eef59c38c9168dafa97)
But basically we check that the batch was onchain at `blockNumCur - confDepth`, meaning the batch was confirmed at least confDepth blocks ago.

api/clients/eigenda_client_e2e_test.go
samlaf: Added comment in [7f7fdaf](https://github.com/Layr-Labs/eigenda/pull/821/commits/7f7fdaf885be3046c006002a1edb073248518827)
Not sure if it's what you had in mind..

api/clients/eigenda_client.go
```diff
@@ -70,6 +79,17 @@ func NewEigenDAClient(log log.Logger, config EigenDAClientConfig) (*EigenDAClien
 		return nil, err
 	}
 
+	var ethClient *ethclient.Client
+	var edasmCaller *edasm.ContractEigenDAServiceManagerCaller
+	ethClient, err = ethclient.Dial(config.EthRpcUrl)
+	if err != nil {
+		return nil, fmt.Errorf("failed to dial ETH RPC node: %w", err)
+	}
+	edasmCaller, err = edasm.NewContractEigenDAServiceManagerCaller(common.HexToAddress(config.SvcManagerAddr), ethClient)
```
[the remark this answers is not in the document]
samlaf: No, but I still made the eth-rpc and service-manager address mandatory, because I realized we should probably be checking the confirmed status of the batch onchain when the disperser tells us (defensive programming).

Realized that it didn't make sense to require them only when confirmationDepth > 0, because why also not verify that the batch has landed onchain when confirmationDepth = 0 (meaning its included only in the currentBlock)? See https://github.com/Layr-Labs/eigenda/blob/7f7fdaf885be3046c006002a1edb073248518827/api/clients/eigenda_client.go#L270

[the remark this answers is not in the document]
samlaf: @epociask is that OK with you or do you see a problem with this?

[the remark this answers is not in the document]
samlaf: Talked offline. We are fine with this. Added a breaking change comment to the PR description.

epociask (approved):
LGTM

PR #823, opened 2024-10-21
feat: add unavailable 503 errors in disperser for retry

samlaf:

See https://linear.app/eigenlabs/issue/EBE-101/implement-503-error-code-logic-in-eigenda-proxy for background.
Basically trying to implement retry/fallback logic in eigenda-proxy. There are 2 main contributions from this PR:

### 1. refactor error codes to be more easily parsable by the human eye

My new programming bible is https://github.com/tigerbeetle/tigerbeetle/blob/main/docs/TIGER_STYLE.md
They suggest using most-significant-words-first style for variables and functions:
> Add units or qualifiers to variable names, and put the units or qualifiers last, sorted by descending significance, so that the variable starts with the most significant word, and ends with the least significant word. For example, latency_ms_max rather than max_latency_ms. This will then line up nicely when latency_ms_min is added, as well as group all variables that relate to latency.

 so I implemented this in [058e1a8](https://github.com/Layr-Labs/eigenda/pull/823/commits/058e1a88bc4ad592d031127776ba702d09716b16)

This way the errors show up in my IDE as
image
which I find way easier to read than
image

### 2. Add a 503 error and refactor DisperseBlob() function to return 503s instead of 500s when it applies

This is in commit [89a278d](https://github.com/Layr-Labs/eigenda/pull/823/commits/89a278d2c01799d03e5130eb19f4b3eae744a84e)

I've given this a try and will be curious for feedback, but honestly it feels like an impossible ask, because most libraries are not written with this kind of detail in mind, and simply return formatted string errors, so its impossible for me to know whether the errors are 400s or 500s for eg, and even less whether they are transient or permanent errors (typically 400s are in this category for eg).

aws sdk stands apart in this category, in that it uses their own smithy error library which encodes all of this information. I tried using that to parse errors from writes to S3, but even that is hard to use. Their ErrorCode method returns a string instead of an enum, and I havent even been able to find a list of their strings anywhere... so don't know what to compare them to (do they use "503" or "Unavailable" for eg?)

Overall I'm thinking of just forgetting about this 500/503 distinction and in eigenda-client using a simpler retry policy that always retries for 500s.... unless anyone thinks this 503 stuff is worth pushing through.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/errors.go
+++ b/api/errors.go
@@ -1,39 +1,49 @@
 package api
 
 import (
+	"time"
+
+	"google.golang.org/genproto/googleapis/rpc/errdetails"
 	"google.golang.org/grpc/codes"
 	"google.golang.org/grpc/status"
+	"google.golang.org/protobuf/types/known/durationpb"
 )
 
 // The canonical errors from the EigenDA gRPC API endpoints.
 //
 // Notes:
-// - Start the error space small (but sufficient), and expand when there is an important
-//   failure case to separate out.
-// - Avoid simply wrapping system-internal errors without checking if they are appropriate
-//   in user-facing errors defined here. Consider map and convert system-internal errors
-//   before return to users from APIs.
+// - We start with a small (but sufficient) subset of google's error code convention, 
+//   and expand when there is an important failure case to separate out. See:
+//   https://cloud.google.com/apis/design/errors#handling_errors
+// - Make sure that internally propagated errors are eventually wrapped in one of the 
+//   user-facing errors defined here, since grpc otherwise returns an UNKNOWN error code,
+//   which is harder to debug and understand for users.
 
-func NewGRPCError(code codes.Code, msg string) error {
+func newErrorGRPC(code codes.Code, msg string) error {
 	return status.Error(code, msg)
 }
 
 // HTTP Mapping: 400 Bad Request
-func NewInvalidArgError(msg string) error {
-	return NewGRPCError(codes.InvalidArgument, msg)
+func NewErrorInvalidArg(msg string) error {
+	return newErrorGRPC(codes.InvalidArgument, msg)
 }
 
 // HTTP Mapping: 404 Not Found
-func NewNotFoundError(msg string) error {
-	return NewGRPCError(codes.NotFound, msg)
+func NewErrorNotFound(msg string) error {
+	return newErrorGRPC(codes.NotFound, msg)
 }
 
 // HTTP Mapping: 429 Too Many Requests
-func NewResourceExhaustedError(msg string) error {
-	return NewGRPCError(codes.ResourceExhausted, msg)
+func NewErrorResourceExhausted(msg string) error {
+	return newErrorGRPC(codes.ResourceExhausted, msg)
 }
 
 // HTTP Mapping: 500 Internal Server Error
-func NewInternalError(msg string) error {
-	return NewGRPCError(codes.Internal, msg)
+func NewErrorInternal(msg string) error {
+	return newErrorGRPC(codes.Internal, msg)
+}
+
+// HTTP Mapping: 501 Not Implemented
+func NewErrorUnimplemented() error {
+	return newErrorGRPC(codes.Unimplemented, "not implemented")
 }
--- a/api/errors.go
+++ b/api/errors.go
@@ -47,3 +47,29 @@ func NewErrorInternal(msg string) error {
 func NewErrorUnimplemented() error {
 	return newErrorGRPC(codes.Unimplemented, "not implemented")
 }
+
+// HTTP Mapping: 503 Service Unavailable
+// Unavailable is used instead of 500 to indicate to the client that it can retry the operation.
+// See the documentation for the FAILED_PRECONDITION error code in https://grpc.io/docs/guides/status-codes/
+// which compares FAILED_PRECONDITION, UNAVAILABLE, and ABORTED.
+func NewErrorUnavailable(msg string) error {
+	return newErrorGRPC(codes.Unavailable, msg)
+}
+
+// HTTP Mapping: 503 Service Unavailable
+// NewErrorUnavailableWithRetry is like NewUnavailableError, but allows the caller to specify a retry delay.
+func NewErrorUnavailableWithRetry(msg string, delay time.Duration) error {
+	st := status.New(codes.Unavailable, msg)
+
+	retry := &errdetails.RetryInfo{
+		RetryDelay: durationpb.New(delay),
+	}
+
+	statusWithDetails, err := st.WithDetails(retry)
+	if err != nil {
+		// If adding details failed, just return the status without details
+		return st.Err()
+	}
+
+	return statusWithDetails.Err()
+}
--- a/api/proto/disperser/disperser.proto
+++ b/api/proto/disperser/disperser.proto
@@ -5,10 +5,16 @@ option go_package = "github.com/Layr-Labs/eigenda/api/grpc/disperser";
 
 // Disperser defines the public APIs for dispersing blobs.
 service Disperser {
-	// This API accepts blob to disperse from clients.
+	// DisperseBlob accepts a single blob to be dispersed.
 	// This executes the dispersal async, i.e. it returns once the request
-	// is accepted. The client could use GetBlobStatus() API to poll the the
+	// is accepted. The client should use GetBlobStatus() API to poll the
 	// processing status of the blob.
+	// 
+	// If DisperseBlob returns the following error codes:
+	// INVALID_ARGUMENT (400): request is invalid for a reason specified in the error msg.
+	// RESOURCE_EXHAUSTED (429): request is rate limited for the quorum specified in the error msg.
+	//                           user should retry after the specified duration.
+	// INTERNAL (500): serious error, user should NOT retry.
 	rpc DisperseBlob(DisperseBlobRequest) returns (DisperseBlobReply) {}
 
 
--- a/api/proto/disperser/disperser.proto
+++ b/api/proto/disperser/disperser.proto
@@ -15,6 +15,7 @@ service Disperser {
 	// RESOURCE_EXHAUSTED (429): request is rate limited for the quorum specified in the error msg.
 	//                           user should retry after the specified duration.
 	// INTERNAL (500): serious error, user should NOT retry.
+	// UNAVAILABLE (503): intermittent failure, user can retry after 1s & exponential backoff.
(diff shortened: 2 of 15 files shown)


disperser/common/blobstore/blob_metadata_store.go
```diff
@@ -526,10 +533,14 @@ func GenerateTableSchema(metadataTableName string, readCapacityUnits int64, writ
 	}
 }
 
+// MarshalBlobMetadata marshals the blob metadata into a DynamoDB item, or else returns a grpc internal error
 func MarshalBlobMetadata(metadata *disperser.BlobMetadata) (commondynamodb.Item, error) {
+	// We use MarshalMap a few times here. Errors from MarshalMap are most likely
+	// a bug in our code (badly set configurations), so not worth client retrying,
+	// hence we send a 500s instead of 503s
```
pschork, line 540: Not sure this comment section is necessary and it seems kinda passive. An error here definitely indicates a bug in our code. The idea of returning 503s here is the wrong mental model for 503s. So adding `instead of 503s` seems like an odd thing to bring up.

pschork (approved):

disperser/common/blobstore/blob_metadata_store.go
samlaf: My idea was that if there’s a bug in our code, then sending a 503 and asking the client to resubmit won’t help; they will just hit the same bug again with the next request. 

This is why this separation between 500s and 503s feels so hard to me. Because im never sure if the error is not actually triggered by some one in a million scenario which wouldn’t happen if they retry for whatever reason. Maybe it is the case (then should send 503), and maybe it isn’t (then should send 500 to prevent them from spamming us with more requests)

api/errors.go
```diff
@@ -47,3 +47,29 @@ func NewErrorInternal(msg string) error {
...
+// HTTP Mapping: 503 Service Unavailable
+// Unavailable is used instead of 500 to indicate to the client that it can retry the operation.
+// See the documentation for the FAILED_PRECONDITION error code in https://grpc.io/docs/guides/status-codes/
+// which compares FAILED_PRECONDITION, UNAVAILABLE, and ABORTED.
+func NewErrorUnavailable(msg string) error {
+	return newErrorGRPC(codes.Unavailable, msg)
+}
+
+// HTTP Mapping: 503 Service Unavailable
+// NewErrorUnavailableWithRetry is like NewUnavailableError, but allows the caller to specify a retry delay.
```
bxue-l2, line 60: since it is not used, should we remove it?

api/proto/disperser/disperser.proto
```diff
@@ -15,6 +15,7 @@ service Disperser {
 	// RESOURCE_EXHAUSTED (429): request is rate limited for the quorum specified in the error msg.
 	//                           user should retry after the specified duration.
 	// INTERNAL (500): serious error, user should NOT retry.
+	// UNAVAILABLE (503): intermittent failure, user can retry after 1s & exponential backoff.
```
bxue-l2, line 18: I remember in the proxy, 503 corresponds to "fallback to ETH", meaning it is less of a transient error. 500 corresponds to some other error. It is bit counter-intuitive

disperser/common/blobstore/blob_metadata_store.go
```diff
@@ -59,11 +61,16 @@ func (s *BlobMetadataStore) QueueNewBlobMetadata(ctx context.Context, blobMetada
 	if s.shadowTableName != "" && s.shadowTableName != s.tableName {
 		err = s.dynamoDBClient.PutItem(ctx, s.shadowTableName, item)
 		if err != nil {
+			// TODO: should we return an error here? If not, then we should document why we are ignoring the error
```
bxue-l2, line 64: @ian-shim

ian-shim: shadow tables aren't used. I will open a PR to clean them up

disperser/common/blobstore/blob_metadata_store.go
```diff
@@ -59,11 +61,16 @@ func (s *BlobMetadataStore) QueueNewBlobMetadata(ctx context.Context, blobMetada
...
 			s.logger.Error("failed to put item into shadow table %s : %v", s.shadowTableName, err)
 		}
 	}
 
-	return s.dynamoDBClient.PutItem(ctx, s.tableName, item)
+	err = s.dynamoDBClient.PutItem(ctx, s.tableName, item)
+	if err != nil {
+		return api.NewErrorUnavailable(fmt.Sprintf("failed to put item into table %s : %v", s.tableName, err))
```
ian-shim, line 71: I don't think we should be returning api errors from libraries. 
Maybe we can create new errors at the library level and match those in apiserver to identify which api error should wrap it?

disperser/common/blobstore/shared_storage.go
```diff
@@ -78,22 +80,31 @@ func NewSharedStorage(bucketName string, s3Client s3.Client, blobMetadataStore *
 func (s *SharedBlobStore) StoreBlob(ctx context.Context, blob *core.Blob, requestedAt uint64) (disperser.BlobKey, error) {
```
ian-shim, line 80: same with errors in this file

disperser/common/blobstore/blob_metadata_store.go
ian-shim: https://github.com/Layr-Labs/eigenda/pull/830

disperser/common/blobstore/blob_metadata_store.go
samlaf: Agree. I started working on a generic API error in https://github.com/Layr-Labs/eigenda/pull/828/files#diff-b14f8e128e2a8997e2b4f5f10af13deeba5cc983f7aad6c540227c0908fada4a

It's under api/errors.go right now, since I'm only using it as part of the clients, but perhaps I can move it higher (I'd be in favor of having it in errors.go under the root dir and using such an error across the codebase personally.

WDYT?

disperser/common/blobstore/shared_storage.go
samlaf: See my comment in the other thread: https://github.com/Layr-Labs/eigenda/pull/823#discussion_r1812887627

api/errors.go
samlaf: think it doesn't hurt to keep around, especially if we *might* use it in the future, since it's non-trivial code to write, so its good to keep the example around I feel.

api/proto/disperser/disperser.proto
samlaf: Ya I agree......... it's hard here because what we're trying to do with the fallback is not really typical Http behavior, so doesn't fit very well into any grpc/http codes. Here I was following the grpc codes https://grpc.io/docs/guides/status-codes/
image
which mention that 500 is serious error. Guess it doesn't mean that we shouldn't retry though.... But also 503 here means "you can retry with exponential backoff", which is kind of annoying because it goes against what we're trying to do with the 503 failover in batcher->proxy communication. Perhaps we need to be more precise and failover on 503 that has a `RETRY-AFTER` header of >= 1 hour? We're getting into complex, annoying, and bug-prone territory for implementers though...

disperser/common/blobstore/blob_metadata_store.go
samlaf: Thanks @ian-shim ! Removed the TODO here since this piece of code will be removed once we merge your PR. [d5a75f1](https://github.com/Layr-Labs/eigenda/pull/823/commits/d5a75f1af353d73b6684dad71d603e14e491ed08)

disperser/common/blobstore/blob_metadata_store.go
samlaf: Ahhhh I see what you are saying now! You're saying that there's no point in even comparing to 503 because 503 is "clearly" wrong. I guess that's true.. I was stuck in this 503 vs 500 model while coding this PR, hence the comment. Will remove.

ian-shim:
Sounds good. Yeah totally fine with using the generic ErrorAPI defined there to surface API errors from API server, but doesn't make a lot of sense for internal libraries to return API errors since they don't implement http service.

samlaf:
> Sounds good. Yeah totally fine with using the generic ErrorAPI defined there to surface API errors from API server, but doesn't make a lot of sense for internal libraries to return API errors since they don't implement http service.

This I disagree! I actually wish every single golang library out there implemented ErrorAPI instead of error, because every library implements an API, it's not only about networking APIs!

And it's very annoying that most libraries out there don't return whether the error is a client or server/system error. Because whenever a webserver needs to make use of those libraries it needs to do some ugly regexp pattern matching against errors to resurface whether its a 400 or 500 error.

Even crypto libraries should implement ErrorAPI imo, as for eg if some input is larger than modulus, then you can resurface a Client error to the caller, which is easier to debug.

ian-shim:
What a hot take 😭 
Not every library is built to be consumed by an http service and http status codes are defined specifically to describe the status of the http response. imo, error codes in libraries should be more generic to accommodate usage in any service, not just http services. I think the right way to address your annoyance is well defined internal error codes, not reusing loosely analogous error codes where only subset of them are relevant anyways

samlaf:
> What a hot take 😭 Not every library is built to be consumed by an http service and http status codes are defined specifically to describe the status of the http response. imo, error codes in libraries should be more generic to accommodate usage in any service, not just http services. I think the right way to address your annoyance is well defined internal error codes, not reusing loosely analogous error codes where only subset of them are relevant anyways

Ah ok agree that this is a good point for not using http error codes but defining our own. Definitely not married to those. The main thing I like is the ErrorFault() which is just a ternary (client,server,unknown). That is enough to be able to at least classify 400 and 500 errors. Potentially I could remove error codes from ErrorAPI then?

Otherwise we can dive very deep into different error codes, with each library defining its own error codes and then we need complicated mappings between them and the ultimate networking error codes. grpc has a translation between their codes and http codes for eg: https://cloud.google.com/apis/design/errors#handling_errors for eg.

Btw I took the idea for ErrorAPI from the aws sdk, which is built using their smithy IDL. All of their sdk methods return an APIError (see https://aws.github.io/aws-sdk-go-v2/docs/handling-errors/). But agree that aws sdk always talks over network (I think?), so it makes mroe sense in this case than a library which talks only to fs, or only to CPU for eg. Will have to think about this more but I still think at least enforcing the 400/500 distinction should be mandatory lol.

PR #824, opened 2024-10-21
Core/v2

mooselumph:

v2 of core library

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/utils.go
+++ b/core/utils.go
@@ -1,4 +1,4 @@
-package corev2
+package core
 
 import (
 	"math"
--- a/core/utils.go
+++ b/core/utils.go
@@ -13,7 +13,6 @@ func RoundUpDivideBig(a, b *big.Int) *big.Int {
 	num := new(big.Int).Sub(new(big.Int).Add(a, b), one) // a + b - 1
 	res := new(big.Int).Div(num, b)                      // (a + b - 1) / b
 	return res
-
 }
 
 func RoundUpDivide[T constraints.Integer](a, b T) T {
--- a/core/v2/assignment.go
+++ b/core/v2/assignment.go
@@ -1,4 +1,4 @@
-package corev2
+package v2
 
 import (
 	"fmt"
@@ -24,8 +24,8 @@ func GetAssignments(state *core.OperatorState, blobVersion byte, quorum uint8) (
 		return nil, fmt.Errorf("too many operators for blob version %d", blobVersion)
 	}
 
-	n := big.NewInt(int64(len(ops)))
-	m := big.NewInt(int64(params.NumChunks))
+	numOperators := big.NewInt(int64(len(ops)))
+	numChunks := big.NewInt(int64(params.NumChunks))
 
 	type assignment struct {
 		id     core.OperatorID
@@ -37,10 +37,10 @@ func GetAssignments(state *core.OperatorState, blobVersion byte, quorum uint8) (
 	chunkAssignments := make([]assignment, 0, len(ops))
 	for ID, r := range state.Operators[quorum] {
 
-		num := new(big.Int).Mul(r.Stake, new(big.Int).Sub(m, n))
+		num := new(big.Int).Mul(r.Stake, new(big.Int).Sub(numChunks, numOperators))
 		denom := state.Totals[quorum].Stake
 
-		chunks := RoundUpDivideBig(num, denom)
+		chunks := core.RoundUpDivideBig(num, denom)
 
 		chunkAssignments = append(chunkAssignments, assignment{id: ID, index: uint32(r.Index), chunks: uint32(chunks.Uint64()), stake: r.Stake})
 	}
@@ -70,12 +70,10 @@ func GetAssignments(state *core.OperatorState, blobVersion byte, quorum uint8) (
 			a.chunks++
 		}
 
-		assignment := Assignment{
+		assignments[a.id] = Assignment{
 			StartIndex: index,
 			NumChunks:  a.chunks,
 		}
-
-		assignments[a.id] = assignment
 		index += a.chunks
 	}
 
--- a/core/v2/assignment.go
+++ b/core/v2/assignment.go
@@ -1,0 +1,123 @@
+package corev2
+
+import (
+	"fmt"
+	"math/big"
+	"sort"
+
+	"github.com/Layr-Labs/eigenda/core"
+)
+
+func GetAssignments(state *core.OperatorState, blobVersion byte, quorum uint8) (map[core.OperatorID]Assignment, error) {
+
+	params, ok := ParametersMap[blobVersion]
+	if !ok {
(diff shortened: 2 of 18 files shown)


core/v2/core_test.go
```diff
@@ -0,0 +1,267 @@
...
+func TestValidationSucceeds(t *testing.T) {
+
+	// operatorCounts := []uint{1, 2, 4, 10, 30}
+
+	// numBlob := 3 // must be greater than 0
+	// blobLengths := []int{1, 32, 128}
+
+	operatorCounts := []uint{4}
+
+	numBlob := 1 // must be greater than 0
```
mooselumph, line 234: comment from @hopeyen: could we add more tests such as different number of blobs and blobs that would be considered invalid by the verifier?

core/v2/types.go
```diff
@@ -0,0 +1,99 @@
...
+	RelayKeys []uint16
+}
+
+type BlobVersionParameters struct {
+	CodingRate              uint32
+	ReconstructionThreshold float64
+	NumChunks               uint32
+}
+
+func (p BlobVersionParameters) MaxNumOperators() uint32 {
```
ian-shim, line 88: This is theoretical max number of operators based on the config, which is different from the max number of operators imposed on contracts. Do we need this method or just use what's on the contracts? 
It's problematic if a `BlobVersionParameters` is set in a way that allows max # operators smaller than what's configured onchain, so it looks like we already have to make sure the set of parameters allow the max # operators configured onchain when they're added

core/v2/assignment.go
```diff
@@ -0,0 +1,123 @@
...
+	if !ok {
+		return nil, fmt.Errorf("no operators found for quorum %d", quorum)
+	}
+
+	if len(ops) > int(params.MaxNumOperators()) {
+		return nil, fmt.Errorf("too many operators for blob version %d", blobVersion)
+	}
+
+	n := big.NewInt(int64(len(ops)))
+	m := big.NewInt(int64(params.NumChunks))
```
ian-shim, line 28: nit: `numOperators` and `numChunks` instead of `n, m`?

core/v2/assignment.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+
+	assignments := make(map[core.OperatorID]Assignment, len(chunkAssignments))
+	index := uint32(0)
+	for i, a := range chunkAssignments {
+		if i < delta {
+			a.chunks++
+		}
+
+		assignment := Assignment{
```
ian-shim, line 73: nit: `assgn := Assignment{...`
variable name is same as type name in L30

core/v2/types.go
```diff
@@ -0,0 +1,99 @@
...
+	}
+
+	return encoding.EncodingParams{
+		NumChunks:   uint64(params.NumChunks),
+		ChunkLength: uint64(length),
+	}, nil
+
+}
+
+type PaymentHeader struct {
```
ian-shim, line 62: We can reuse `core.PaymentMetadata`?

core/v2/assignment.go
```diff
@@ -0,0 +1,123 @@
+package corev2
```
ian-shim, line 1: should we just name the package `v2`?

core/v2/utils.go
```diff
@@ -0,0 +1,26 @@
+package corev2
```
ian-shim, line 1: should we replace existing usages?

ian-shim (approved):

core/v2/types.go
mooselumph: Good callout. Since all of this will be on chain, what about having some on-chain checks to ensure consistency when any of these parameters are updated?

ian-shim: That's a good idea. Can you add a quick comment over `BlobVersionParameters` that this is supposed to go onchain?

cc @0x0aa0

core/utils.go
```diff
@@ -0,0 +1,26 @@
...
+	"golang.org/x/exp/constraints"
+)
+
+func RoundUpDivideBig(a, b *big.Int) *big.Int {
+
+	one := new(big.Int).SetUint64(1)
+	num := new(big.Int).Sub(new(big.Int).Add(a, b), one) // a + b - 1
+	res := new(big.Int).Div(num, b)                      // (a + b - 1) / b
+	return res
+
```
dmanc, line 16: nit: remove extra new lines

encoding/kzg/prover/prover_cpu.go
```diff
@@ -21,8 +21,13 @@ import (
...
 func (g *Prover) newProver(params encoding.EncodingParams) (*ParametrizedProver, error) {
 
-	// Check that the parameters are valid with respect to the SRS.
-	if params.ChunkLength*params.NumChunks >= g.SRSOrder {
+	// Check that the parameters are valid with respect to the SRS. The precomputed terms of the amortized KZG
+	// prover use up to order params.ChunkLen*params.NumChunks-1 for the SRS, so we must have
+	// params.ChunkLen*params.NumChunks-1 <= g.SRSOrder. The condition below could technically
+	// be relaxed to params.ChunkLen*params.NumChunks > g.SRSOrder+1, but because all of the paramters are
+	// powers of 2, the stricter condition is equivalent.
+
```
dmanc, line 29: nit: remove new line

encoding/kzg/prover/parametrized_prover.go
```diff
@@ -67,30 +73,56 @@ func (g *ParametrizedProver) Encode(inputFr []fr.Element) (*bn254.G1Affine, *bn2
 
 	encodeStart := time.Now()
 
-	rsChan := make(chan RsEncodeResult, 1)
-	lengthCommitmentChan := make(chan LengthCommitmentResult, 1)
-	lengthProofChan := make(chan LengthProofResult, 1)
-	commitmentChan := make(chan CommitmentResult, 1)
-	proofChan := make(chan ProofsResult, 1)
+	commitmentsChan := make(chan commitmentsResult, 1)
```
dmanc, line 76: Just making sure, the changes here are just refactors to allow us to expose GetFrames and GetCommitments. None of the functionality changes in any major way?

dmanc (approved):
lgtm, left some comments

encoding/kzg/prover/prover_cpu.go
```diff
@@ -21,8 +21,13 @@ import (
...
+	if params.ChunkLength*params.NumChunks > g.SRSOrder {
```
dmanc, line 30: We have a function called ValidateEncodingParams, let's just use that.

core/v2/types.go
mooselumph: Done

encoding/kzg/prover/parametrized_prover.go
mooselumph: That's right!

PR #826, opened 2024-10-22
refactor: make disperser client reuse same grpc connection

samlaf:

grpc "connections" are actually virtual connections: the docs/specs calls them channels.
They are not a physical underlying tcp/http connection but actually manage underlying connections. Hence, they are not meant to be redialed on every request. They should be created once, and reused for multiple requests, which is what this PR does.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/disperser_client.go
+++ b/api/clients/disperser_client.go
@@ -28,6 +28,7 @@ type Config struct {
 	UseSecureGrpcFlag bool
 }
 
+// Deprecated: Use &Config{...} directly instead
 func NewConfig(hostname, port string, timeout time.Duration, useSecureGrpcFlag bool) *Config {
 	return &Config{
 		Hostname:          hostname,
@@ -53,8 +54,15 @@ type disperserClient struct {
 	// whenever a method is called, using initOnce to make sure initialization happens only once
 	// and is thread-safe
 	initOnce sync.Once
-	conn     *grpc.ClientConn
-	client   disperser_rpc.DisperserClient
+	// We use a single grpc connection, which allows a max number of concurrent open streams (from DisperseBlobAuthenticated).
+	// This should be fine in most cases, as each such request should take <1sec per 1MB blob.
+	// The MaxConcurrentStreams parameter is set by the server. If not set, then it defaults to the stdlib's
+	// http2 default of 100-1000: https://github.com/golang/net/blob/4783315416d92ff3d4664762748bd21776b42b98/http2/transport.go#L55
+	// This means a conservative estimate of 100-1000MB/sec, which should be amply sufficient.
+	// If we ever need to increase this, we could either consider asking the disperser to increase its limit,
+	// or to use a pool of connections here.
+	conn   *grpc.ClientConn
+	client disperser_rpc.DisperserClient
 }
 
 var _ DisperserClient = &disperserClient{}
@@ -299,4 +307,3 @@ func getGrpcDialOptions(useSecureGrpcFlag bool) []grpc.DialOption {
 	}
 	return options
 }
-
--- a/api/clients/disperser_client.go
+++ b/api/clients/disperser_client.go
@@ -5,6 +5,7 @@ import (
 	"crypto/tls"
 	"errors"
 	"fmt"
+	"sync"
 	"time"
 
 	disperser_rpc "github.com/Layr-Labs/eigenda/api/grpc/disperser"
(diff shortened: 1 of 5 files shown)


api/clients/disperser_client.go
```diff
@@ -37,47 +38,73 @@ func NewConfig(hostname, port string, timeout time.Duration, useSecureGrpcFlag b
...
+	initOnce sync.Once
+	conn     *grpc.ClientConn
+	client   disperser_rpc.DisperserClient
 }
 
 var _ DisperserClient = &disperserClient{}
 
-func NewDisperserClient(config *Config, signer core.BlobRequestSigner) DisperserClient {
+// DisperserClient maintains a single underlying grpc connection to the disperser server,
+// through which it sends requests to disperse blobs, get blob status, and retrieve blobs.
```
bxue-l2, line 63: probably should take care of the max concurrent requests, see https://learn.microsoft.com/en-us/aspnet/core/grpc/performance?view=aspnetcore-8.0#connection-concurrency. 

If it reaches the max, the new call get blocked. It can happen if OP uses concurrent blobs spinning up 100 conn

samlaf: Ah good catch. We can bump that limit or make it a configuration actually.
But I really don't think it's that big of a worry. OP would have to literally make 100 requests at the same time, which I think is unlikely? How long does a typical DisperseBlobAuthenticated take? < 5secs hopefully?
Also what happens if that limit is reached? it returns a 500 or something? Then batcher would just resubmit the blob a little bit later once the other connections are done processing.

api/clients/disperser_client.go
```diff
@@ -37,47 +38,73 @@ func NewConfig(hostname, port string, timeout time.Duration, useSecureGrpcFlag b
...
 }
 
+// See the NewDisperserClient constructor's documentation for details and usage examples.
 type disperserClient struct {
 	config *Config
 	signer core.BlobRequestSigner
+	// conn and client are not initialized in the constructor, but are initialized lazily
+	// whenever a method is called, using initOnce to make sure initialization happens only once
+	// and is thread-safe
+	initOnce sync.Once
```
ian-shim, line 55: 💯

api/clients/disperser_client.go
```diff
@@ -37,47 +38,73 @@ func NewConfig(hostname, port string, timeout time.Duration, useSecureGrpcFlag b
...
+//
+//	// The connection will be established on the first call
+//	status, requestId, err := client.DisperseBlob(ctx, someData, someQuorums)
+//	if err != nil {
+//	    // Handle error
+//	}
+//
+//	// Subsequent calls will use the existing connection
+//	status2, requestId2, err := client.DisperseBlob(ctx, otherData, otherQuorums)
+func NewDisperserClient(config *Config, signer core.BlobRequestSigner) *disperserClient {
```
ian-shim, line 82: Why did we switch the return type to struct pointer vs. interface?

api/clients/disperser_client.go
bxue-l2: DisperseBlobAuthenticated takes about 300ms for 1MB blob

ian-shim (commented):
Do we have tests for these clients?

api/clients/disperser_client.go
samlaf: https://en.m.wikipedia.org/wiki/Robustness_principle instantiated in golang is "take interfaces as input, and return structs" (see https://bryanftan.medium.com/accept-interfaces-return-structs-in-go-d4cab29a301b for eg).

This is not a breaking change because the pointer implements the interface.

api/clients/disperser_client.go
samlaf: Update comment: [ef1c7a5](https://github.com/Layr-Labs/eigenda/pull/826/commits/ef1c7a591109da15e9759020783b0cbb3821242d)

samlaf:
> Do we have tests for these clients?

No we don't! :O
Didn't notice this.. we only have unit tests for the eigenda_client, which use a mock disperser_client.
We do have an eigenda_client e2e test, that I'm trying to add to CI in #820 

What do you recommend is best way to test the disperser client? Writing a fake grpc server is prob the easiest way to go right?

api/clients/disperser_client.go
ian-shim: This is a good read. Thanks!

ian-shim (approved):

PR #827, opened 2024-10-22
[payment] transactor ping payment vault contract

hopeyen:

filling in the transactor functions to read the payment vault contract on-chain

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


core/chainio.go
```diff
@@ -107,16 +107,16 @@ type Reader interface {
...
+	// GetActiveReservations returns active reservations (end timestamp > current timestamp)
+	GetActiveReservationByAccount(ctx context.Context, accountID string) (ActiveReservation, error)
 
 	// GetOnDemandPayments returns all on-demand payments
-	GetOnDemandPayments(ctx context.Context, blockNumber uint32, accountIDs []string) (map[string]OnDemandPayment, error)
+	GetOnDemandPayments(ctx context.Context, accountIDs []string) (map[string]OnDemandPayment, error)
 
-	// GetOnDemandPaymentByAccount returns on-demand payment of an account
-	GetOnDemandPaymentByAccount(ctx context.Context, blockNumber uint32, accountID string) (OnDemandPayment, error)
+	// GetOnDemandPayments returns all on-demand payments
```
ian-shim, line 118: nit: this change needs to be undone

core/chainio.go
```diff
@@ -107,16 +107,16 @@ type Reader interface {
 	GetAllVersionedBlobParams(ctx context.Context) (map[uint8]*BlobVersionParameters, error)
 
 	// GetActiveReservations returns active reservations (end timestamp > current timestamp)
-	GetActiveReservations(ctx context.Context, blockNumber uint32, accountIDs []string) (map[string]ActiveReservation, error)
+	GetActiveReservations(ctx context.Context, accountIDs []string) (map[string]ActiveReservation, error)
 
-	// GetActiveReservationByAccount returns active reservation by account ID
-	GetActiveReservationByAccount(ctx context.Context, blockNumber uint32, accountID string) (ActiveReservation, error)
+	// GetActiveReservations returns active reservations (end timestamp > current timestamp)
```
ian-shim, line 112: nit: this change needs to be undone

core/data.go
```diff
@@ -9,6 +9,7 @@ import (
 
 	commonpb "github.com/Layr-Labs/eigenda/api/grpc/common"
 	"github.com/Layr-Labs/eigenda/common"
+	paymentvault "github.com/Layr-Labs/eigenda/contracts/bindings/PaymentVault"
```
ian-shim, line 12: can we not have contract dependency on core data structs?

core/eth/reader.go
```diff
@@ -37,6 +38,7 @@ type ContractBindings struct {
 	EigenDAServiceManager *eigendasrvmg.ContractEigenDAServiceManager
 	EjectionManager       *ejectionmg.ContractEjectionManager
 	AVSDirectory          *avsdir.ContractAVSDirectory
+	PaymentVault          *paymentvault.ContractPaymentVault
```
ian-shim, line 41: should this be initialized in `updateContractBindings`?

core/eth/reader.go
```diff
@@ -616,42 +618,119 @@ func (t *Reader) GetAllVersionedBlobParams(ctx context.Context) (map[uint8]*core
...
-	return map[string]core.ActiveReservation{}, nil
+func (t *Reader) GetActiveReservations(ctx context.Context, accountIDs []string) (map[string]core.ActiveReservation, error) {
+	// map accountIDs to addresses
+	accountAddresses := make([]gethcommon.Address, len(accountIDs))
+	for i, accountID := range accountIDs {
+		accountAddresses[i] = gethcommon.HexToAddress(accountID)
+	}
+
+	reservations_map := make(map[string]core.ActiveReservation)
+	reservations, err := t.bindings.PaymentVault.GetReservations(&bind.CallOpts{
```
ian-shim, line 629: would this crash if `PaymentVault` is not initialized

core/eth/reader.go
```diff
@@ -616,42 +618,119 @@ func (t *Reader) GetAllVersionedBlobParams(ctx context.Context) (map[uint8]*core
...
-	// contract is not implemented yet
-	return map[string]core.ActiveReservation{}, nil
+func (t *Reader) GetActiveReservations(ctx context.Context, accountIDs []string) (map[string]core.ActiveReservation, error) {
+	// map accountIDs to addresses
+	accountAddresses := make([]gethcommon.Address, len(accountIDs))
+	for i, accountID := range accountIDs {
+		accountAddresses[i] = gethcommon.HexToAddress(accountID)
+	}
+
+	reservations_map := make(map[string]core.ActiveReservation)
```
ian-shim, line 628: `reservationsMap`

core/eth/reader.go
```diff
@@ -616,42 +618,119 @@ func (t *Reader) GetAllVersionedBlobParams(ctx context.Context) (map[uint8]*core
...
+	reservations, err := t.bindings.PaymentVault.GetReservations(&bind.CallOpts{
+		Context: ctx,
+	}, accountAddresses)
+	if err != nil {
+		return nil, err
+	}
+
+	// since reservations are returned in the same order as the accountIDs, we can directly map them
+	for i, reservation := range reservations {
+		reservations_map[accountIDs[i]] = reservation
```
ian-shim, line 638: Is it guaranteed that `PaymentVault.GetReservations` returns reservations in the same order?
validate that `len(accountIDs) = len(reservations)` before this loop?

core/eth/reader.go
```diff
@@ -616,42 +618,119 @@ func (t *Reader) GetAllVersionedBlobParams(ctx context.Context) (map[uint8]*core
...
 }
 
-func (t *Reader) GetActiveReservations(ctx context.Context, blockNumber uint32, accountIDs []string) (map[string]core.ActiveReservation, error) {
-	// contract is not implemented yet
-	return map[string]core.ActiveReservation{}, nil
+func (t *Reader) GetActiveReservations(ctx context.Context, accountIDs []string) (map[string]core.ActiveReservation, error) {
+	// map accountIDs to addresses
+	accountAddresses := make([]gethcommon.Address, len(accountIDs))
+	for i, accountID := range accountIDs {
+		accountAddresses[i] = gethcommon.HexToAddress(accountID)
```
ian-shim, line 625: Should `accountID` always be a hex string representing an address? 
Should we type the input as `[]gethcommon.Address` then?

core/chainio.go
```diff
@@ -107,16 +107,16 @@ type Reader interface {
 	GetAllVersionedBlobParams(ctx context.Context) (map[uint8]*BlobVersionParameters, error)
 
 	// GetActiveReservations returns active reservations (end timestamp > current timestamp)
-	GetActiveReservations(ctx context.Context, blockNumber uint32, accountIDs []string) (map[string]ActiveReservation, error)
+	GetActiveReservations(ctx context.Context, accountIDs []string) (map[string]ActiveReservation, error)
```
ian-shim, line 110: nit: return pointers to the object (same with return types in other methods below)?

core/eth/reader.go
```diff
@@ -616,42 +618,119 @@ func (t *Reader) GetAllVersionedBlobParams(ctx context.Context) (map[uint8]*core
...
+		reservations_map[accountIDs[i]] = reservation
+	}
+	return reservations_map, nil
 }
 
-func (t *Reader) GetActiveReservationByAccount(ctx context.Context, blockNumber uint32, accountID string) (core.ActiveReservation, error) {
-	// contract is not implemented yet
-	return core.ActiveReservation{}, nil
+func (t *Reader) GetActiveReservationByAccount(ctx context.Context, accountID string) (core.ActiveReservation, error) {
+	reservation, err := t.bindings.PaymentVault.GetReservation(&bind.CallOpts{
```
ian-shim, line 644: If account is not part of the mapping, contract will return zero value for the reservation instead of an error. We should probably validate that and return err here instead.

Same with other methods in this file that accesses mapping onchain

core/meterer/onchain_state.go
```diff
@@ -107,6 +109,33 @@ func (pcs *OnchainPaymentState) RefreshOnchainPaymentState(ctx context.Context,
 	}
 	// These parameters should be rarely updated, but we refresh them anyway
 	pcs.PaymentVaultParams = paymentVaultParams
+
```
ian-shim, line 112: not necessarily this PR, but FYI. I ended up using atomic.Pointer to keep these "read-mostly" values

core/meterer/onchain_state.go
```diff
@@ -107,6 +109,33 @@ func (pcs *OnchainPaymentState) RefreshOnchainPaymentState(ctx context.Context,
 	}
 	// These parameters should be rarely updated, but we refresh them anyway
 	pcs.PaymentVaultParams = paymentVaultParams
```
ian-shim, line 111: does this need locking?

core/eth/reader.go
hopeyen: it looks like the ordering and length is guaranteed by [this loop](https://github.com/Layr-Labs/eigenda/pull/800/files#diff-2f1b4eb845063422ced2032eace4bbe25fc506e57398f8ed8d1f0fa00a971b4eR131). If there's no match in the mapping, as you commented, would be a zero value. I will update it to set a nil or empty reservation when there's a zero value, do you think that's sufficient?

core/meterer/onchain_state.go
hopeyen: hmmm the case I would be concerned about is this gets updated in the middle of an on-demand payment, where it passes the old price per symbol but fails the new global rate. But with payment rollback, price per symbol isn't used. So far, I don't think we need locking and this seems atomic to me, but I'm happy to update if there's something I'm not seeing

core/meterer/onchain_state.go
hopeyen: cool, updated to use a pointer for this field

core/eth/reader.go
hopeyen: Yep! made an update

core/eth/reader.go
```diff
@@ -616,42 +619,130 @@ func (t *Reader) GetAllVersionedBlobParams(ctx context.Context) (map[uint8]*core
...
-	return core.ActiveReservation{}, nil
+func (t *Reader) GetActiveReservationByAccount(ctx context.Context, accountID gethcommon.Address) (*core.ActiveReservation, error) {
+	reservation, err := t.bindings.PaymentVault.GetReservation(&bind.CallOpts{
+		Context: ctx,
+	}, accountID)
+	if err != nil {
+		return nil, err
+	}
+	if isZeroValuedReservation(reservation) {
+		return nil, errors.New("reservation is zero-valued")
```
ian-shim, line 654: nit: `"reservation does not exist for given account"`

core/chainio.go
```diff
@@ -107,16 +107,16 @@ type Reader interface {
 	GetAllVersionedBlobParams(ctx context.Context) (map[uint8]*BlobVersionParameters, error)
 
 	// GetActiveReservations returns active reservations (end timestamp > current timestamp)
-	GetActiveReservations(ctx context.Context, blockNumber uint32, accountIDs []string) (map[string]ActiveReservation, error)
+	GetActiveReservations(ctx context.Context, accountIDs []gethcommon.Address) (*map[gethcommon.Address]ActiveReservation, error)
```
ian-shim, line 110: map is already a pointer, you probably want `map[gethcommon.Address]*ActiveReservation` here and below

core/meterer/onchain_state.go
```diff
@@ -28,17 +30,18 @@ var _ OnchainPayment = (*OnchainPaymentState)(nil)
 type OnchainPaymentState struct {
 	tx *eth.Reader
 
-	ActiveReservations map[string]core.ActiveReservation
-	OnDemandPayments   map[string]core.OnDemandPayment
+	ActiveReservations *map[gethcommon.Address]core.ActiveReservation
```
ian-shim, line 33: you probably want `map[gethcommon.Address]*core.ActiveReservation`

core/meterer/onchain_state.go
ian-shim: I don't think atomicity is guaranteed for pointer assignment. 
Can we use `atomic.Pointer`?

core/meterer/onchain_state.go
hopeyen: yes, thank you

core/meterer/onchain_state.go
hopeyen: I attempted this, lmk if that looks like what you are thinking

core/meterer/onchain_state.go
```diff
@@ -106,63 +111,84 @@ func (pcs *OnchainPaymentState) RefreshOnchainPaymentState(ctx context.Context,
 		return err
 	}
 	// These parameters should be rarely updated, but we refresh them anyway
-	pcs.PaymentVaultParams = paymentVaultParams
+	atomic.StorePointer((*unsafe.Pointer)(unsafe.Pointer(&pcs.PaymentVaultParams)), unsafe.Pointer(paymentVaultParams))
```
ian-shim, line 114: Do we need to use unsafe pointer? Can we just type `PaymentVaultParams` as `atomic.Pointer[PaymentVaultParams]` in L41?

core/eth/reader.go
ian-shim: Cool, are you planning to leave this PR open till the contract code is ready and deployed?
Maybe we should just add nil checks and proceed for now? We can update/remove the nil checks when contracts are ready

core/meterer/onchain_state.go
hopeyen: aah cool! updated

core/eth/reader.go
hopeyen: I added nil checks before the contract calls so we can go ahead to merge! I think the payment DASM link is almost ready, can work on that as soon as that one gets merged

core/eth/reader.go
```diff
@@ -656,44 +659,159 @@ func (t *Reader) GetAllVersionedBlobParams(ctx context.Context) (map[uint8]*core
...
+	for i, reservation := range reservations {
+		reservationsMap[accountIDs[i]] = &reservation
+	}
+
+	// filter out all zero-valued reservations
+	for accountID, reservation := range reservationsMap {
+		if isZeroValuedReservation(*reservation) {
+			delete(reservationsMap, accountID)
+		}
+	}
```
ian-shim, line 684: We can move this check inside the for loop above (L675) and only set the value if reservation isn't zero valued

core/eth/utils.go
```diff
@@ -126,3 +128,9 @@ func bitmapToBytesArray(bitmap *big.Int) []byte {
 	}
 	return bytesArray
 }
+
+func isZeroValuedReservation(reservation paymentvault.IPaymentVaultReservation) bool {
+	zeroReservation := paymentvault.IPaymentVaultReservation{}
+
+	return reflect.DeepEqual(reservation, zeroReservation)
```
ian-shim, line 135: `maps.Equal` for performance

core/eth/reader.go
```diff
@@ -656,44 +659,159 @@ func (t *Reader) GetAllVersionedBlobParams(ctx context.Context) (map[uint8]*core
...
+	if t.bindings.PaymentVault == nil {
+		return nil, errors.New("payment vault not deployed")
+	}
+	onDemandPayment, err := t.bindings.PaymentVault.GetOnDemandAmount(&bind.CallOpts{
+		Context: ctx,
+	}, accountID)
+	if err != nil {
+		return nil, err
+	}
+	if onDemandPayment == big.NewInt(0) {
```
ian-shim, line 744: Has this been tested? This will result in false even if `onDemandPayment` has value of 0 because you're comparing bigint pointers. 
Use something like `onDemandPayment.Cmp(big.NewInt(0)) == 0` instead

core/meterer/onchain_state.go
```diff
@@ -106,63 +113,84 @@ func (pcs *OnchainPaymentState) RefreshOnchainPaymentState(ctx context.Context,
...
+	pcs.OnDemandLocks.Unlock()
+
 	return nil
 }
 
 // GetActiveReservationByAccount returns a pointer to the active reservation for the given account ID; no writes will be made to the reservation
-func (pcs *OnchainPaymentState) GetActiveReservationByAccount(ctx context.Context, accountID string) (core.ActiveReservation, error) {
-	if reservation, ok := pcs.ActiveReservations[accountID]; ok {
+func (pcs *OnchainPaymentState) GetActiveReservationByAccount(ctx context.Context, accountID gethcommon.Address) (*core.ActiveReservation, error) {
+	if reservation, ok := (pcs.ActiveReservations)[accountID]; ok {
```
ian-shim, line 149: Don't we need rlock here? I think it can crash if a map is read at the same time it's being written

core/data.go
ian-shim: can we do something here?

core/eth/utils.go
hopeyen: I had trouble with this because `IPaymentVaultReservation` is not a map. is it more performant to put the struct into a map and use `maps.Equal`?

core/eth/reader.go
hopeyen: I don't have test coverage over this. Do we have anywhere that puts tests on contract calls?
(updated to use your suggestion)

core/meterer/onchain_state.go
hopeyen: oof, I added RLock and defer RUnlock

core/data.go
hopeyen: what do you think about moving `type ActiveReservation = paymentvault.IPaymentVaultReservation` to core/eth?

core/eth/reader.go
```diff
@@ -654,44 +657,151 @@ func (t *Reader) GetAllVersionedBlobParams(ctx context.Context) (map[uint8]*core
...
+		Context: ctx,
+	}, accountIDs)
+	if err != nil {
+		return nil, err
+	}
+
+	// since reservations are returned in the same order as the accountIDs, we can directly map them
+	for i, reservation := range reservations {
+		if isZeroValuedReservation(reservation) {
+			delete(reservationsMap, accountIDs[i])
```
ian-shim, line 675: If a reservation is zero, this delete would be a no-op and it would be added in L677, no? 
Don't we want `if !isZeroValuedReservation(reservation) { reservationsMap[accountIDs[i]] = &reservation }` ?

core/eth/utils.go
ian-shim: Ah it's a struct. We could compare each field individually (or i'm fine with using reflect as well).

Do you know if eth rpc will return nil vs. empty byte slice for `[]byte` type fields? 
If they return empty byte slice, this comparison will evaluate to false

core/data.go
ian-shim: How does that remove the contract dependency?

hopeyen: core/eth already has contract dependency, so we are won't introducing it in data.go, we just need to avoid circular dependencies in this case

core/meterer/onchain_state.go
ian-shim: If you `defer pcs.ReservationsLock.RUnlock()` it will hold the rlock until the end of the execution. Will it cause conflict with write lock in L160?

core/eth/utils.go
```diff
@@ -126,3 +128,27 @@ func bitmapToBytesArray(bitmap *big.Int) []byte {
...
+		reservation.EndTimestamp == 0 &&
+		len(reservation.QuorumNumbers) == 0 &&
+		len(reservation.QuorumSplits) == 0
+}
+
+// ConvertToActiveReservation converts a upstream binding data structure to local definition.
+// Returns an error if the input reservation is zero-valued.
+func ConvertToActiveReservation(reservation paymentvault.IPaymentVaultReservation) (*core.ActiveReservation, error) {
+	if isZeroValuedReservation(reservation) {
+		return nil, fmt.Errorf("reservation does not exist for given account")
```
ian-shim, line 144: This error message is not accurate given this is a generic util method

core/eth/reader.go
```diff
@@ -654,44 +657,152 @@ func (t *Reader) GetAllVersionedBlobParams(ctx context.Context) (map[uint8]*core
...
+		Context: ctx,
+	}, accountIDs)
+	if err != nil {
+		return nil, err
+	}
+
+	// since payments are returned in the same order as the accountIDs, we can directly map them
+	for i, payment := range payments {
+		if payment.Cmp(big.NewInt(0)) == 0 {
+			t.logger.Warn("failed to get on demand payment for account", "account", accountIDs[i], "err", err)
```
ian-shim, line 714: I don't think `err` is relevant here?

ian-shim (approved):
couple of nits, but lgtm!

core/eth/reader.go
hopeyen: you are right! removed

core/eth/utils.go
hopeyen: updated to say that `reservation is not a valid active reservation`

PR #828, opened 2024-10-22
feat: eigenda client returns 503 errors (for failover purpose)

samlaf:

Related to batcher failover project. See https://linear.app/eigenlabs/issue/EBE-60/design-fallback-mechanism-for-both-op-and-arb and https://github.com/ethereum-optimism/specs/issues/434 for more details.

Basically we want eigenda-proxy to return 503 to batcher to signify "eigenda is down, failover and start submitting blobs to ethereum instead".

This PR makes the eigenda-client return 503s. eigenda-proxy will then use these, retry 3 times, and if it is still returning 503s, then return a 503 to the batcher.

The 2 sources of 503 errors that we have identified are:
1. disperseBlob call failed (disperser is down) - TODO: this is not implemented yet b/c disperser client still doesn't return errors with error codes. I can either update this PR or make a followup one to do that.
2. blob stuck in PROCESSING or DISPERSING status and call times out before reaching CONFIRMED state (landing onchain) - this is implemented

### TODO
- [x] implement 503s in disperserClient: [3a7b16e](https://github.com/Layr-Labs/eigenda/pull/828/commits/3a7b16edbbeb456af671a137bac594438bb86478)
- [x] Add documentation for how to set the different TIMEOUTS (see this [discussion](https://github.com/Layr-Labs/eigenda/pull/821#discussion_r1816913522)): [efd6d28](https://github.com/Layr-Labs/eigenda/pull/828/commits/efd6d28fe35dff3ed4f9ca43d4b0ae6229b01a93)

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/errors.go
```diff
@@ -37,3 +40,105 @@ func NewResourceExhaustedError(msg string) error {
...
+// ErrorAPI is the most generic API and protocol agnostic error interface that we eventually
+// want every api clients returned errors to implement.
+//
+// This way, consumers of the api clients can tell what kind of error they are dealing with,
+// most broadly whether it is a client or server fault (see the ErrorFault type).
+type ErrorAPI interface {
+	error
+
+	// ErrorCode returns the error code for the API exception.
+	ErrorCode() ErrorCode
```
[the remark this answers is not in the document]
samlaf: Good idea: [b4b3441](https://github.com/Layr-Labs/eigenda/pull/828/commits/b4b34411c210de18b23bc9841a0b7186ce0856e3)

I guess my main concern was whether we eventually switch to grpc error codes instead. In that case we might want ErrorCode to be a string for eg instead of an int. But in the meantime agree using stdlib is best.

api/clients/eigenda_client.go
```diff
@@ -178,18 +199,34 @@ func (m *EigenDAClient) putBlob(ctx context.Context, rawData []byte, resultChan
...
+			// Assuming that the timeout is correctly set (long enough to both land onchain + finalize),
+			// 1. means that there is a problem with EigenDA, so we return 503 to let the batcher failover to ethda
+			// 2. means that there is a problem with Ethereum, so we return 500.
+			//    batcher would most likely resubmit another blob, which is not ideal but there isn't much to be done...
+			//    eigenDA v2 will have idempotency so one can just resubmit the same blob safely.
+			if latestBlobStatus == grpcdisperser.BlobStatus_PROCESSING || latestBlobStatus == grpcdisperser.BlobStatus_DISPERSING {
+				errChan <- api.NewErrorAPIGeneric(http.StatusServiceUnavailable,
+					fmt.Errorf("eigenda might be down. timed out waiting for blob to land onchain (request id=%s): %w", base64RequestID, ctx.Err()))
+			}
+			// but not else (otherwise it might be a problem with ethereum, so fallbacking to ethda wouldnt help)
```
bxue-l2, line 220: should we be more explicit, with else if (grpcdisperser.Confirmed)

and default to return something else. Maybe useful for debugging?

api/clients/eigenda_client.go
```diff
@@ -200,10 +237,12 @@ func (m *EigenDAClient) putBlob(ctx context.Context, rawData []byte, resultChan
 					alreadyWaitingForDispersal = true
 				}
 			case grpcdisperser.BlobStatus_FAILED:
-				errChan <- fmt.Errorf("EigenDA blob dispersal failed in processing, requestID=%s: %w", base64RequestID, err)
+				// TODO: when exactly does this happen? I think only happens if ethereum reorged and the blob was lost
```
bxue-l2, line 240: 1. blob has expired, a client retrieve after 14 days. Sounds like 400 errors. (fine to ignore, since it would timeout already) 
2. internal logic error while requesting encoding (shouldn't happen), but it should return 503
3. wait for blob finalization from confirmation and blob retry has exceeded its limit. Thinking a bit more, it is possible that a chain re-org triggered a Failed status. See code (https://github.com/Layr-Labs/eigenda/blob/master/disperser/batcher/finalizer.go#L179-L189). So we should be returning 500, not 503. 

In any cases, EigenDA returned error into the failed status. So it makes sense that we return 500, i.e. internalServerError.

api/clients/eigenda_client.go
```diff
@@ -200,10 +237,12 @@ func (m *EigenDAClient) putBlob(ctx context.Context, rawData []byte, resultChan
...
+				errChan <- api.NewErrorAPIGeneric(http.StatusInternalServerError, fmt.Errorf("blob dispersal (requestID=%s) reached failed status. please resubmit the blob.", base64RequestID))
 				return
 			case grpcdisperser.BlobStatus_INSUFFICIENT_SIGNATURES:
-				errChan <- fmt.Errorf("EigenDA blob dispersal failed in processing with insufficient signatures, requestID=%s: %w", base64RequestID, err)
+				// this might be a temporary condition where some eigenda nodes were temporarily offline, so we should retry
+				errChan <- api.NewErrorAPIGeneric(http.StatusInternalServerError, fmt.Errorf("blob dispersal (requestID=%s) failed with insufficient signatures. please resubmit the blob.", base64RequestID))
```
bxue-l2, line 245: I think it should be 503. Effectively, it means the service is down for this batch.

disperser/disperser.go
```diff
@@ -17,16 +17,43 @@ import (
...
+	// - blob has expired
+	// - internal logic error while requesting encoder API 
+	// - blob retry has exceeded its limit while waiting for blob finalization after confirmation
 	Failed
+	// FINALIZED means that the block containing the blob's confirmation transaction has been finalized on Ethereum
 	Finalized
+	// INSUFFICIENT_SIGNATURES means that the confirmation threshold for the blob was not met
+	// for at least one quorum.
 	InsufficientSignatures
+	// DISPERSING means that the blob is currently being dispersed to DA Nodes and being confirmed onchain
```
bxue-l2, line 56: have not yet confirmed onchain

disperser/disperser.go
```diff
@@ -17,16 +17,43 @@ import (
...
+	// signatures, which is a separate state). This status is somewhat of a catch-all category,
+	// containg (but not necessarily exclusively as errors can be added in the future):
+	// - blob has expired
+	// - internal logic error while requesting encoder API 
+	// - blob retry has exceeded its limit while waiting for blob finalization after confirmation
 	Failed
+	// FINALIZED means that the block containing the blob's confirmation transaction has been finalized on Ethereum
 	Finalized
+	// INSUFFICIENT_SIGNATURES means that the confirmation threshold for the blob was not met
+	// for at least one quorum.
```
bxue-l2, line 54: I think once a blob received an insufficient signature, it won't be retried.

samlaf: Yes that is mentioned above:
```
// Terminal states are states that will not be updated to a different state:
// - FAILED
// - FINALIZED
// - INSUFFICIENT_SIGNATURES
```

disperser/disperser.go
samlaf: I literally just copied these from disperser.proto. But agree, updated both: [69f0b50](https://github.com/Layr-Labs/eigenda/pull/828/commits/69f0b50ea521607dfdc96caf170e07936d1835e6)

api/clients/eigenda_client.go
samlaf: Updated comments with your info: [585d68e](https://github.com/Layr-Labs/eigenda/pull/828/commits/585d68eb588fceebe90644810733d3373d83f371)

api/clients/eigenda_client.go
samlaf: good idea

samlaf: [5f8c3fb](https://github.com/Layr-Labs/eigenda/pull/828/commits/5f8c3fba836405ec937521367d36e5577bbf5d88)

api/clients/eigenda_client.go
```diff
@@ -178,18 +199,34 @@ func (m *EigenDAClient) putBlob(ctx context.Context, rawData []byte, resultChan
...
+			// We can only land here if blob is still in
+			// 1. processing or dispersing status: waiting to land onchain
+			// 2. or confirmed status: landed onchain, waiting for finalization
+			// because all other statuses return immediately below.
+			//
+			// Assuming that the timeout is correctly set (long enough to both land onchain + finalize),
+			// 1. means that there is a problem with EigenDA, so we return 503 to let the batcher failover to ethda
+			// 2. means that there is a problem with Ethereum, so we return 500.
+			//    batcher would most likely resubmit another blob, which is not ideal but there isn't much to be done...
+			//    eigenDA v2 will have idempotency so one can just resubmit the same blob safely.
```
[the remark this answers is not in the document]
samlaf: Not sure I'm following your comment. v2 will have idempotency. You can resubmit the same blob, and disperser will just give you the blobStatus blob right away. Same thing as sending a PUT request with idempotency key. If its already there, it doesn't get written twice. That's the whole point of idempotency.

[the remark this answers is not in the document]
bxue-l2: Ok, it is one way to achieve idempotency within some time window. It is interesting that, we can achieve it by making it impossible to disperser to accept blobs for that time window. 

However, once that time window passed. User can disperse the same blob twice.

api/clients/eigenda_client.go
```diff
@@ -219,9 +219,18 @@ func (m *EigenDAClient) putBlob(ctx context.Context, rawData []byte, resultChan
 			if latestBlobStatus == grpcdisperser.BlobStatus_PROCESSING || latestBlobStatus == grpcdisperser.BlobStatus_DISPERSING {
 				errChan <- api.NewErrorAPIGeneric(http.StatusServiceUnavailable,
 					fmt.Errorf("eigenda might be down. timed out waiting for blob to land onchain (request id=%s): %w", base64RequestID, ctx.Err()))
+			} else if latestBlobStatus == grpcdisperser.BlobStatus_CONFIRMED {
+				// Timeout'ing in confirmed state means one of two things:
+				// 1. (if timeout was long enough to finalize in normal conditions): problem with ethereum, so we return 504
```
bxue-l2, line 224: 503

api/clients/eigenda_client.go
```diff
@@ -156,13 +169,21 @@ func (m *EigenDAClient) putBlob(ctx context.Context, rawData []byte, resultChan
...
+			Code:  0,
+			Fault: api.ErrorFaultUnknown,
+		}
 		return
 	}
 
 	// process response
 	if *blobStatus == disperser.Failed {
-		errChan <- fmt.Errorf("unable to disperse blob to eigenda (reply status %d): %w", blobStatus, err)
+		// Don't think this state should be reachable. DisperseBlobAuthenticated should return an error instead of a failed status.
```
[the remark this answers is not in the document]
samlaf: Are you sure about this? Why don't we just remove the blobStatus from DisperseBlobReply then?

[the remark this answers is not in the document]
bxue-l2: We don't failed inside disperser api server. 
https://github.com/Layr-Labs/eigenda/blob/master/disperser/apiserver/server.go#L101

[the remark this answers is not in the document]
samlaf: I see. Removed this check then. Also updated proto and disperser-client api to mention this: [2185cd3](https://github.com/Layr-Labs/eigenda/pull/828/commits/2185cd3223e7e1b3cc05c2daae7e3b867fdfd5dd)

Think we should just remove the BlobStatus from the Reply then... it's useless.

api/proto/disperser/disperser.proto
```diff
@@ -159,8 +159,8 @@ enum BlobStatus {
 	// INSUFFICIENT_SIGNATURES means that the confirmation threshold for the blob was not met
 	// for at least one quorum.
 	INSUFFICIENT_SIGNATURES = 5;
-
-	// DISPERSING means that the blob is currently being dispersed to DA Nodes and being confirmed onchain
+	// DISPERSING means that the blob is currently being dispersed to DA Nodes, 
+	// and the batcher is gathering their signatures.
```
bxue-l2, line 164: Between DISPERSING and its next state (InsufficientSig, Confirmed, or failed), there are two phases
1. Dispersing to DA nodes and collecting signature
3. submitting the transaction on chain and waiting for tx receipt

bxue-l2: my bad, not very accurate in the first comment

api/clients/eigenda_client.go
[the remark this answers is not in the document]
samlaf: Ah I see. That is true, depending on whether we keep the metadata table (history of requests) forever or not.
I had another comment somewhere in this PR I think mentioning this regarding FAILED status. did you answer it?

api/proto/disperser/disperser.proto
samlaf: [1588266](https://github.com/Layr-Labs/eigenda/pull/828/commits/15882662fd15d415bf0aab2e2610b48e49400f92)

api/errors.go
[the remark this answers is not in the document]
samlaf: Actually changed back in [8eeb3e0](https://github.com/Layr-Labs/eigenda/pull/828/commits/8eeb3e0968b90128578dfd494b469c7b56afdb81) to grpc error codes only. I think this whole tanged with ErrorAPI was a mistake... mea culpa.

api/clients/eigenda_client.go
```diff
@@ -156,13 +169,21 @@ func (m *EigenDAClient) putBlob(ctx context.Context, rawData []byte, resultChan
 	// TODO: would be nice to add a trace-id key to the context, to be able to follow requests from batcher->proxy->eigenda
 	blobStatus, requestID, err := m.Client.DisperseBlobAuthenticated(ctx, data, customQuorumNumbers)
 	if err != nil {
-		errChan <- fmt.Errorf("error initializing DisperseBlobAuthenticated() client: %w", err)
+		errChan <- &api.ErrorAPIGeneric{
+			Err: fmt.Errorf("error submitting authenticated blob to disperser: %w", err),
+			// We set to unknown fault b/c disperser client returns a mix of 400s and 500s currently.
+			// TODO: update disperser client to also return ErrorAPIGeneric errors
+			Code:  0,
+			Fault: api.ErrorFaultUnknown,
```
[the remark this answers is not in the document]
samlaf: Changed back to using the grpc codes directly in [8eeb3e0](https://github.com/Layr-Labs/eigenda/pull/828/commits/8eeb3e0968b90128578dfd494b469c7b56afdb81), so we return the disperser-client's grpc error codes directly. Proxy will have to be updated to read those.

bxue-l2 (approved):

api/clients/eigenda_client.go
```diff
@@ -229,18 +252,41 @@ func (m *EigenDAClient) putBlob(ctx context.Context, rawData []byte, resultChan
...
 		case <-ctx.Done():
-			errChan <- fmt.Errorf("timed out waiting for EigenDA blob to confirm blob with request id=%s: %w", base64RequestID, ctx.Err())
+			// We can only land here if blob is still in
+			// 1. processing or dispersing status: waiting to land onchain
+			// 2. or confirmed status: landed onchain, waiting for finalization
+			// because all other statuses return immediately below.
+			//
+			// Assuming that the timeout is correctly set (long enough to both land onchain + finalize),
+			// 1. means that there is a problem with EigenDA, so we return an ErrorFailover to let the batcher failover to ethda
+			// 2. means that there is a problem with Ethereum, so we return 500.
```
bxue-l2, line 266: 500, or 504. Probably more accurate to point out the grpc error

api/clients/eigenda_client.go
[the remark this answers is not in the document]
bxue-l2: I think how it is going to be implemented with bins or leaky bucket alike, like how current rate limitor works. The bin window is about 10-30min? (not super sure). But in the long term, it is going to be an onchain parameter I think

api/clients/eigenda_client.go
```diff
@@ -251,10 +297,21 @@ func (m *EigenDAClient) putBlob(ctx context.Context, rawData []byte, resultChan
 					alreadyWaitingForDispersal = true
 				}
 			case grpcdisperser.BlobStatus_FAILED:
-				errChan <- fmt.Errorf("EigenDA blob dispersal failed in processing, requestID=%s: %w", base64RequestID, err)
+				// This can happen for a few reasons:
+				// 1. blob has expired, a client retrieve after 14 days. Sounds like 400 errors, but not sure this can happen during dispersal...
+				// 2. internal logic error while requesting encoding (shouldn't happen), but should probably return api.ErrorFailover
+				// 3. wait for blob finalization from confirmation and blob retry has exceeded its limit.
+				//    Probably from a chain re-org. See https://github.com/Layr-Labs/eigenda/blob/master/disperser/batcher/finalizer.go#L179-L189.
+				//    So we should be returning 500 to force a blob resubmission (not eigenda's fault but until
```
bxue-l2, line 305: In v2, disperser likely will not return a failed state, because there is no onchain transaction. Then the source to enter into a failed state most is mostly because of insufficient signature. In that case, the disperser will keep retry, and it would not return a failed state. Failed state probably just is going to be some catastrophic event on disperser.

api/clients/eigenda_client.go
```diff
@@ -205,15 +232,11 @@ func (m *EigenDAClient) putBlob(ctx context.Context, rawData []byte, resultChan
 	}
 	// disperse blob
 	// TODO: would be nice to add a trace-id key to the context, to be able to follow requests from batcher->proxy->eigenda
-	blobStatus, requestID, err := m.Client.DisperseBlobAuthenticated(ctx, data, customQuorumNumbers)
+	_, requestID, err := m.Client.DisperseBlobAuthenticated(ctx, data, customQuorumNumbers)
```
bxue-l2, line 235: since err is wrapped, the proxy client should handle this error

api/clients/eigenda_client.go
```diff
@@ -229,18 +252,41 @@ func (m *EigenDAClient) putBlob(ctx context.Context, rawData []byte, resultChan
...
+			} else if latestBlobStatus == grpcdisperser.BlobStatus_CONFIRMED {
+				// Timeout'ing in confirmed state means one of two things:
+				// 1. (if timeout was long enough to finalize in normal conditions): problem with ethereum, so we return 504 (DeadlineExceeded)
+				// 2. TODO: (if timeout was not long enough to finalize in normal conditions): eigenda-client is badly configured, should be a 400 (INVALID_ARGUMENT)
+				errChan <- api.NewErrorDeadlineExceeded(
+					fmt.Sprintf("timed out waiting for blob that landed onchain to finalize (request id=%s). "+
+						"Either timeout not long enough, or ethereum might be experiencing difficulties: %v. ", base64RequestID, ctx.Err()))
+			} else {
+				// this should not be reachable...
+				errChan <- api.NewErrorInternal(fmt.Sprintf("timed out in a state that shouldn't be possible (request id=%s): %s", base64RequestID, ctx.Err()))
```
bxue-l2, line 280: Since it is not unexpected, should we let the client to EthDA, since EigenDA logic is so wrong at this point.

api/clients/eigenda_client.go
```diff
@@ -291,7 +348,7 @@ func (m *EigenDAClient) putBlob(ctx context.Context, rawData []byte, resultChan
 				return
 			default:
 				// this should never happen. If it does, the blob is in a heisenberg state... it could either eventually get confirmed or fail
-				errChan <- fmt.Errorf("unknown reply status %d. ask for assistance from EigenDA team, using requestID %s", statusRes.Status, base64RequestID)
+				errChan <- api.NewErrorInternal(fmt.Sprintf("unknown reply status %d. ask for assistance from EigenDA team, using requestID %s", statusRes.Status, base64RequestID))
```
bxue-l2, line 351: Same here, if EigenDA logics is so wrong at this point, should we just let the client to use EthDA?

api/clients/eigenda_client.go
samlaf: oh ya good catch: [2912230](https://github.com/Layr-Labs/eigenda/pull/828/commits/29122300ece440a9269756fdb9c41b07545f77f4)

api/clients/eigenda_client.go
samlaf: Think we shouldn't add comments about v2 until its actually implemented though? Do you want to just add this note to the notion doc we started yesterday?

api/clients/eigenda_client.go
samlaf: Not sure what you are saying here. Are you saying the wrapping here will break the eigenda-proxy logic? It won't, this error will be handled fine, the wrapping doesn't change anything, because grpc's `status.FromError()` function will unwrap until it find a grpc error.

api/clients/eigenda_client.go
samlaf: Yes good suggestion: [c9ea9c7](https://github.com/Layr-Labs/eigenda/pull/828/commits/c9ea9c75243e772a5abc8f157742e681053a1e1d)

api/clients/eigenda_client.go
samlaf: This one I'm not sure... actually do you know how we can be in this state? Not sure that the heisenberg state means. It sounds like its not a problem with eigenda but possibly just some weird logic with this particular blob, so just resubmitting the blob should be the correct behavior?

api/clients/eigenda_client.go
bxue-l2: just a place holder, so that comment from the other PR can reference

api/clients/eigenda_client.go
bxue-l2: yeah, internalError should be good. There is something wrong, but nothing shows that it is catastrophic

api/clients/eigenda_client.go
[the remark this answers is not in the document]
samlaf: Oh? I think I misunderstood your previous comment then.
That seems wrong then? Idempotency shouldn't have anything to do with rate limits and small window size... it should have to do with whether we already received the blob. So we should just check if that blobHeader is already in the dynamodb metadata table (or wherever we store those things in v2), and if yes, then return the blobStatus for that blob instead of redispersing it. No?

PR #829, opened 2024-10-23
Replace `math/rand` with `crypto/rand` for generating challenge

ian-shim:
Using `crypto/rand` to generate challenge for disperser endpoint

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/apiserver/server.go
+++ b/disperser/apiserver/server.go
@@ -2,9 +2,10 @@ package apiserver
 
 import (
 	"context"
+	"crypto/rand"
+	"encoding/binary"
 	"errors"
 	"fmt"
-	"math/rand"
 	"net"
 	"slices"
 	"strings"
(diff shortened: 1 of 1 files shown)


disperser/apiserver/server.go
```diff
@@ -146,7 +147,14 @@ func (s *DispersalServer) DisperseBlobAuthenticated(stream pb.Disperser_Disperse
...
 	// Send back challenge to client
-	challenge := rand.Uint32()
+	challengeBytes := make([]byte, 32)
+	_, err = rand.Read(challengeBytes)
+	if err != nil {
+		s.metrics.HandleInvalidArgRpcRequest("DisperseBlobAuthenticated")
+		s.metrics.HandleInvalidArgRequest("DisperseBlobAuthenticated")
+		return api.NewInvalidArgError(fmt.Sprintf("failed to generate challenge: %v", err))
+	}
+	challenge := binary.LittleEndian.Uint32(challengeBytes)
```
anupsv, line 157: any preference to little or big endian

ian-shim: I don't think that matters much here as we're converting one random byte array into uint32

anupsv (approved):

PR #830, opened 2024-10-23
Remove shadow write from disperser

ian-shim:
Shadow table was added to allow testing two different batchers (existing batcher + minibatcher) with traffic sent to one shared disperser. 
We ended up removing minibatcher and having separate disperser in v2, so shadow writes won't be needed anymore. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


samlaf (commented):
LGTM but I'm not code owner so will let someone else approve.

pschork (approved):

PR #832, opened 2024-10-23
Refactor Transactor into Reader/Writer

mooselumph:

Breaks transactor into reader/writer

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


core/tx.go
```diff
@@ -29,7 +29,7 @@ type OperatorSetParam struct {
 
 type OperatorStakes map[QuorumID]map[OperatorIndex]OperatorStake
 
-type Transactor interface {
+type TransactorOld interface {
```
hopeyen, line 32: not deleting this yet? I don't think it is used anywhere

hopeyen (approved):
:shipit:

core/mock/writer.go
```diff
@@ -12,29 +12,29 @@ import (
 	"github.com/stretchr/testify/mock"
```
ian-shim, line 12: should this be split into MockWriter and MockReader?

core/tx.go
ian-shim: +1

core/eth/state.go
```diff
@@ -10,10 +10,10 @@ import (
 
 type ChainState struct {
 	Client common.EthClient
-	Tx     core.Transactor
+	Tx     core.Reader
 }
 
-func NewChainState(tx core.Transactor, client common.EthClient) *ChainState {
+func NewChainState(tx core.Reader, client common.EthClient) *ChainState {
```
ian-shim, line 16: will this be part of `Reader` down the road?

core/mock/writer.go
mooselumph: Since the `Writer` implements `Reader`, for mocking purposes it seems fine to combine? What do you think?

core/eth/state.go
mooselumph: I think so. Will be a follow up.

core/mock/writer.go
ian-shim: How about having MockReader embedded into MockWriter just like how Writer embeds Reader?

ian-shim (approved):

PR #833, opened 2024-10-24
S3 relay interface

cody-littley:

This PR adds utilities for uploading and downloading files from S3 in a way that supports breaking files into smaller ones.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests


go.mod
```diff
@@ -6,6 +6,7 @@ toolchain go1.21.1
 
 require (
 	github.com/Layr-Labs/eigensdk-go v0.1.7-0.20240507215523-7e4891d5099a
+	github.com/aws/aws-sdk-go v1.55.5
```
dmanc, line 9: Is it possible to do it with just v2 library?

relay/dataplane/config.go
```diff
@@ -0,0 +1,46 @@
...
+	Bucket string
+	// If true then the bucket will be created if it does not already exist. If false and the bucket does not exist
+	// then the S3Client will return an error when it is created. Default is false.
+	AutoCreateBucket bool
+	// The number of characters of the key to use as the prefix. A value of "3" for the key "ABCDEFG" would result in
+	// the prefix "ABC". Default is 3.
+	PrefixChars int
+	// This framework utilizes a pool of workers to help upload/download files. This value specifies the number of
+	// workers to use. Default is 32.
+	Parallelism int
```
dmanc, line 24: Should we set the default as number of cpus in machine? Like by using gomaxprocs

dmanc (commented):
Looks great, some initial comments

relay/dataplane/s3_client.go
```diff
@@ -0,0 +1,241 @@
...
+	// ctx is the context for the client.
+	ctx context.Context
+	// cancel is called to cancel the context.
+	cancel context.CancelFunc
+	// the AWS S3 handle to use to talk to S3.
+	svc *s3.S3
+	// tasks are placed into this channel to be processed by workers.
+	tasks chan func()
+	// this wait group is completed when all workers have finished.
+	wg *sync.WaitGroup
```
ian-shim, line 28: I think what you're doing here can be simplified & abstracted out by using `workerpool` (ex. https://github.com/Layr-Labs/eigenda/blob/master/api/clients/retrieval_client.go#L181)

ian-shim (commented):
Looks good! Few comments

relay/dataplane/s3_client.go
cody-littley: This is a neat little library, good to know about. Done.

relay/dataplane/config.go
cody-littley: I've turned this into two parameters: `ParallelismFactor` and `ParallelismConstant`. The total number of workers is set equal to the formula `ParallelismFactor * numCores + ParallelismConstant`.

We will want to have a lot more workers than cores, since most of the time the workers are blocked on IO tasks. This allows us to set a sane default that uses a good number of workers as the number of cores grows.

common/aws/cli.go
```diff
@@ -3,20 +3,46 @@ package aws
...
+	FragmentParallelismFactorFlagName   = "aws.fragment-parallelism-factor"
+	FragmentParallelismConstantFlagName = "aws.fragment-parallelism-constant"
+	FragmentReadTimeoutFlagName         = "aws.fragment-read-timeout"
+	FragmentWriteTimeoutFlagName        = "aws.fragment-write-timeout"
 )
 
 type ClientConfig struct {
-	Region          string
-	AccessKey       string
+	// The region to use when interacting with S3. Default is "us-east-2".
```
ian-shim, line 22: nit: start comment with the name of the field..

common/aws/cli.go
```diff
@@ -3,20 +3,46 @@ package aws
...
+	// A value of "3" for the key "ABCDEFG" will result in the prefix "ABC". Default is 3.
+	FragmentPrefixChars int
+	// This framework utilizes a pool of workers to help upload/download files. A non-zero value for this parameter
+	// adds a number of workers equal to the number of cores times this value. Default is 8. In general, the number
+	// of workers here can be a lot larger than the number of cores because the workers will be blocked on I/O most
+	// of the time.
+	FragmentParallelismFactor int
+	// This framework utilizes a pool of workers to help upload/download files. A non-zero value for this parameter
+	// adds a constant number of workers. Default is 0.
+	FragmentParallelismConstant int
```
ian-shim, line 41: Why not just use a single field to explicitly define the number of workers?

common/aws/s3/client.go
```diff
@@ -148,3 +175,162 @@ func (s *client) ListObjects(ctx context.Context, bucket string, prefix string)
...
+		return err
+	}
+	resultChannel := make(chan error, len(fragments))
+
+	ctx, cancel := context.WithTimeout(ctx, s.cfg.FragmentWriteTimeout)
+	defer cancel()
+
+	for _, fragment := range fragments {
+		fragmentCapture := fragment
+		s.pool.Submit(func() {
```
ian-shim, line 208: This pool is shared across all ongoing uploads and downloads. Which means when there are many uploads/downloads in flight, it can build backpressure. Is that problematic?

ian-shim (approved):
lgtm! 
Please resolve the test failures before merging

common/aws/cli.go
cody-littley: fixed

common/aws/cli.go
cody-littley: This is based on @dmanc's request to scale the number of workers based on the number of cores on the machine. The pattern here allows for the user to specify either a fixed number of threads or a number that varies with the number of cores.

If you think this is over complicated I'm willing to go back to having a constant number of worker threads.

common/aws/s3/client.go
cody-littley: My previous iteration had a config parameter that allowed the client to be have a configurable sized work queue (i.e. when I was using my own implementation of `workerpool`. Unfortunately, the `workerpool` library does not allow us to override the channel used to send data to the workers.

From the implementation:
```
workerQueue: make(chan func())
```

By default, channels have a buffer size of 0. This means that if the number of read/write tasks exceed the number of available workers, the caller will block until all tasks are accepted by a worker. This will provide back pressure if more work is scheduled than there are workers to handle that work.

Are you ok with the way this is configured by default? If not, we will probably need our own implementation of `workerpool`.

PR #838, opened 2024-10-25
Normalize allowlist lookups to use lowercase non-checksummed authenticated address

pschork:

This mitigates a recent incident where the allowlist contained a non-checksummed address for LayerN, but LayerN requests contained a checksummed address resulting in failed rateConfig lookup.

This change normalizes the addresses to lowercase when loading the allowlist rateConfig, and always lowers the authenticatedAccount when doing rate bucket lookup.

This approach allows the allowList to be specified as checksummed or non-checksummed address, as well as the dispersalRequest account to be specified as checksummed or non-checksummed address.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/apiserver/config.go
+++ b/disperser/apiserver/config.go
@@ -8,6 +8,7 @@ import (
 	"log"
 	"os"
 	"strconv"
+	"strings"
 	"time"
 
 	"github.com/Layr-Labs/eigenda/common"
@@ -164,9 +165,11 @@ func ReadAllowlistFromFile(f string) (Allowlist, error) {
 	}
 
 	for _, entry := range allowlistEntries {
-		rateInfoByQuorum, ok := allowlist[entry.Account]
+		// normalize to lowercase (non-checksummed) address
+		account := strings.ToLower(entry.Account)
+		rateInfoByQuorum, ok := allowlist[account]
 		if !ok {
-			allowlist[entry.Account] = map[core.QuorumID]PerUserRateInfo{
+			allowlist[account] = map[core.QuorumID]PerUserRateInfo{
 				core.QuorumID(entry.QuorumID): {
 					Name:       entry.Name,
 					Throughput: common.RateParam(entry.ByteRate),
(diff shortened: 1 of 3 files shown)


disperser/apiserver/server.go
```diff
@@ -326,6 +326,15 @@ func (s *DispersalServer) getAccountRate(origin, authenticatedAddress string, qu
 	// Check if the address is in the allowlist
 	if len(authenticatedAddress) > 0 {
 		quorumRates, ok := s.rateConfig.Allowlist[authenticatedAddress]
```
ian-shim, line 328: how about we always make the keys in this map lowercase and compare with lowercased addresses? 
https://github.com/Layr-Labs/eigenda/blob/master/disperser/apiserver/config.go#L167-L170

pschork: Thx for pointing this out. That's a way more systemic solution and avoids us having to normalize the allowlist configs.

disperser/apiserver/config.go
```diff
@@ -164,9 +165,11 @@ func ReadAllowlistFromFile(f string) (Allowlist, error) {
 	}
 
 	for _, entry := range allowlistEntries {
-		rateInfoByQuorum, ok := allowlist[entry.Account]
+		// normalize to lowercase (non-checksummed) address
```
ian-shim, line 168: nit: can we say this is either address or IP address? (not necessarily address)

disperser/apiserver/server.go
```diff
@@ -325,6 +325,9 @@ func (s *DispersalServer) getAccountRate(origin, authenticatedAddress string, qu
 
 	// Check if the address is in the allowlist
 	if len(authenticatedAddress) > 0 {
+		// normalize to lowercase (non-checksummed) address
```
ian-shim, line 328: same here

ian-shim (approved):
Thanks for fixing this!

bxue-l2 (approved):

PR #839, opened 2024-10-26
Fix pointer issue

mooselumph:

Addresses a pointer issue in the below portion of the rate limiter implementation:

```
// Calculate updated bucket levels
	allowed := true
	for i, size := range d.globalRateParams.BucketSizes {

		// Determine bucket deduction
		deduction := time.Microsecond * time.Duration(1e6*float32(params.BlobSize)/float32(params.Rate)/d.globalRateParams.Multipliers[i])

		prevLevel := bucketParams.BucketLevels[i]

		// Update the bucket level
		bucketParams.BucketLevels[i] = getBucketLevel(bucketParams.BucketLevels[i], size, interval, deduction)
		allowed = allowed && bucketParams.BucketLevels[i] > 0

	}
```

This code returns the `bucketParams` struct, which is then push back to the rate limiter's KV store when the call is allowed. 

The problem here is that `bucketParams.BucketLevels` is an array slice and the KV store is an in-memory LRU cache. Even though `bucketParams` is a copy of the object stored in the KV store, both objects reference the same slice. The code here will directly update the slice, so that when `allowed` is false, the bucket level is inadvertently updated. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/common/ratelimit/limiter.go
+++ b/common/ratelimit/limiter.go
@@ -129,7 +129,7 @@ func (d *rateLimiter) checkAllowed(ctx context.Context, params common.RequestPar
 				"requester_id":   params.RequesterID,
 				"requester_name": params.RequesterName,
 				"bucket_index":   strconv.Itoa(i),
-			}).Set(float64(bucketParams.BucketLevels[i]))
+			}).Set(float64(bucketLevels[i]))
 		}
 	}
 
--- a/common/ratelimit/limiter.go
+++ b/common/ratelimit/limiter.go
@@ -100,11 +100,13 @@ func (d *rateLimiter) checkAllowed(ctx context.Context, params common.RequestPar
 		}
 	}
 
+	bucketLevels := make([]time.Duration, len(d.globalRateParams.BucketSizes))
+
 	// Check whether the request is allowed based on the rate
 
 	// Get interval since last request
 	interval := time.Since(bucketParams.LastRequestTime)
-	bucketParams.LastRequestTime = time.Now().UTC()
+	lastRequestTime := time.Now().UTC()
 
 	// Calculate updated bucket levels
 	allowed := true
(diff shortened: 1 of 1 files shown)


pschork (approved):
RIP mutation roulette. Nice fix. LGTM

common/ratelimit/limiter.go
```diff
@@ -113,13 +115,11 @@ func (d *rateLimiter) checkAllowed(ctx context.Context, params common.RequestPar
...
 		// Update the bucket level
-		bucketParams.BucketLevels[i] = getBucketLevel(bucketParams.BucketLevels[i], size, interval, deduction)
-		allowed = allowed && bucketParams.BucketLevels[i] > 0
+		bucketLevels[i] = getBucketLevel(bucketParams.BucketLevels[i], size, interval, deduction)
+		allowed = allowed && bucketLevels[i] > 0
 
-		d.logger.Debug("Bucket level updated", "key", params.RequesterID, "name", params.RequesterName, "prevLevel", prevLevel, "level", bucketParams.BucketLevels[i], "size", size, "interval", interval, "deduction", deduction, "allowed", allowed)
+		d.logger.Debug("Bucket level updated", "key", params.RequesterID, "name", params.RequesterName, "prevLevel", bucketParams.BucketLevels[i], "level", bucketLevels[i], "size", size, "interval", interval, "deduction", deduction, "allowed", allowed)
 
 		// Update metrics only if the requester name is provided. We're making
```
dmanc, line 124: Should we update the metric to use  `bucketLevels[i]`?

dmanc (commented):
I didn't fully understand how this change will fix the problem. A better description in the PR with an example would be helpful for me.

common/ratelimit/limiter.go
mooselumph: Updated

mooselumph:
> I didn't fully understand how this change will fix the problem. A better description in the PR with an example would be helpful for me.

I updated the PR description with a better explanation!

common/ratelimit/limiter.go
```diff
@@ -129,10 +129,15 @@ func (d *rateLimiter) checkAllowed(ctx context.Context, params common.RequestPar
 				"requester_id":   params.RequesterID,
 				"requester_name": params.RequesterName,
 				"bucket_index":   strconv.Itoa(i),
-			}).Set(float64(bucketParams.BucketLevels[i]))
+			}).Set(float64(bucketLevels[i]))
```
bxue-l2, line 132: not understanding it. Wouldn't the bucketLevel get updated anyway in both cases?

ian-shim (approved):

PR #841, opened 2024-10-28
[v2] s3 Blob Store

ian-shim:
Thin wrapper around s3 for blob storage

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


disperser/common/v2/blobstore/s3_blob_store.go
```diff
@@ -0,0 +1,42 @@
...
+
+// GetBlob retrieves a blob from the blob store
+func (b *BlobStore) GetBlob(ctx context.Context, blobKey string) ([]byte, error) {
+	data, err := b.s3Client.DownloadObject(ctx, b.bucketName, blobKey)
+	if err != nil {
+		b.logger.Errorf("failed to download blob from bucket %s: %v", b.bucketName, err)
+		return nil, err
+	}
+	return data, nil
+}
```
cody-littley, line 42: When we eventually merge the [fragmented uploading utility](https://github.com/Layr-Labs/eigenda/pull/833), we will want to revisit this code. A prerequisite to using fragmented uploads will be to store extra metadata for each blob (i.e. fragment size + blob size needed to download without round trips).

No need to wait to merge this PR. The fact that blobs are fragmeneted in S3 should be invisible to things utilizing this API. As long as we don't deploy this code to production environments, a switchover to the fragmented strategy won't be that tricky.

cody-littley (approved):

disperser/common/v2/blobstore/s3_blob_store.go
ian-shim: Not sure if we want the fragmented uploading in the blob store. This is not the relay store that stores chunks, but the one that stores whole blobs. I think this simple wrapper can serve the blob store in the long run too?

cody-littley: I'm ok waiting to have the discussion for how we want to store blobs. It's mostly a question as to whether we want to try and improve latency for really big blobs. Since we don't really support super big blobs currently, it might be something we can avoid dealing with for the MVP.

PR #844, opened 2024-10-29
[v2] `BlobMetadataStore`: Blob Certificate Operations

ian-shim:
## Why are these changes needed?
- updates conditional puts in dynamodb client
- refactors put operations in v2 metadata store with conditional
- adds operations to put/get blob certs 

## Checks

- [ ] I've made sure the lint is passing in this PR.
- [ ] I've made sure the tests are passing. Note that there might be a few flaky tests, in that case, please comment that they are not relevant.
- [ ] I've checked the new test coverage and the coverage percentage didn't drop.
- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests
   - [ ] This PR is not tested :(

PR #846, opened 2024-10-29
[v2] Encoding manager

ian-shim:
- Adds EncodingManager component
- Adds conditional updates for blob status
- Adds interface for encoding client v2

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


ian-shim:
To be merged on top of https://github.com/Layr-Labs/eigenda/pull/844

disperser/controller/encoding_manager.go
```diff
@@ -0,0 +1,200 @@
...
+
+type EncodingManagerConfig struct {
+	PullInterval time.Duration
+
+	EncodingRequestTimeout time.Duration
+	StoreTimeout           time.Duration
+	// NumEncodingRetries defines how many times the encoding will be retried
+	NumEncodingRetries int
+	// NumRelayAssignment defines how many relays will be assigned to a blob
+	NumRelayAssignment uint16
```
pschork, line 32: nit: seems like this should be plural

disperser/controller/encoding_manager.go
```diff
@@ -0,0 +1,200 @@
...
+}
+
+func GetRelayKeys(numAssignment uint16, availableRelays []corev2.RelayKey) ([]corev2.RelayKey, error) {
+	if int(numAssignment) > len(availableRelays) {
+		return nil, fmt.Errorf("numAssignment (%d) cannot be greater than numRelays (%d)", numAssignment, len(availableRelays))
+	}
+	relayKeys := availableRelays
+	// shuffle relay keys
+	for i := len(relayKeys) - 1; i > 0; i-- {
+		j := rand.Intn(i + 1)
```
pschork, line 195: should this be cryptographically secure randomness? If so `crypto/rand` should be used.

pschork (approved):

disperser/controller/encoding_manager.go
ian-shim: I don't think so. It just has to be random enough. Performance is probably more important than security

disperser/controller/encoding_manager.go
```diff
@@ -0,0 +1,200 @@
...
+			e.logger.Error("failed to get blob key", "err", err, "requestedAt", blob.RequestedAt, "paymentMetadata", blob.BlobHeader.PaymentMetadata)
+			continue
+		}
+		e.lastUpdatedAt = blob.UpdatedAt
+
+		// Encode the blobs
+		e.pool.Submit(func() {
+			for i := 0; i < e.NumEncodingRetries+1; i++ {
+				encodingCtx, cancel := context.WithTimeout(ctx, e.EncodingRequestTimeout)
+				fragmentInfo, err := e.encodeBlob(encodingCtx, blobKey, blob)
```
dmanc, line 129: Is this for adding metadata for how to retrieve from the S3 bucket?

ian-shim: Yeah current thinking is that encoder returns the size of the chunks and the number of fragments (files stored), and controller stores them in metadata. Wdyt?

dmanc: Yeah I see, it makes sense to me now since we are writing the chunks through the S3 client with the new fragmenting logic.

disperser/controller/encoding_manager.go
```diff
@@ -0,0 +1,200 @@
...
+		blobKey, err := blob.BlobHeader.BlobKey()
+		if err != nil {
+			e.logger.Error("failed to get blob key", "err", err, "requestedAt", blob.RequestedAt, "paymentMetadata", blob.BlobHeader.PaymentMetadata)
+			continue
+		}
+		e.lastUpdatedAt = blob.UpdatedAt
+
+		// Encode the blobs
+		e.pool.Submit(func() {
+			for i := 0; i < e.NumEncodingRetries+1; i++ {
```
mooselumph, line 127: I don't really love how this all uses one retry loop. If we are just failing to update dynamo on the last step, do we really need to reencode the blob?

disperser/controller/encoding_manager.go
```diff
@@ -0,0 +1,200 @@
...
+				storeCtx, cancel := context.WithTimeout(ctx, e.StoreTimeout)
+				err = e.blobMetadataStore.PutBlobCertificate(storeCtx, cert)
+				cancel()
+				if err != nil && !errors.Is(err, dispcommon.ErrAlreadyExists) {
+					e.logger.Error("failed to put blob certificate", "err", err)
+					continue
+				}
+
+				storeCtx, cancel = context.WithTimeout(ctx, e.StoreTimeout)
+				err = e.blobMetadataStore.MarkBlobEncoded(storeCtx, blobKey, fragmentInfo)
```
mooselumph, line 156: Why are we updating the blob record with the `fragmentInfo` instead of just pushing it next to the `BlobCert`?

disperser/controller/encoding_manager.go
ian-shim: The plan was to keep encoder request idempotent (i.e. encoder checks if the blob has been encoded and returns immediately in that case). Then the extra encoding request is trivial

disperser/controller/encoding_manager.go
ian-shim: It would be nicer to put it in BlobCert. I'm not sure if it belongs there though. Is it something that should go onchain?

mooselumph: Couldn't we just store a wrapper at this stage which contains both the `BlobCert` and this additional metadata? I agree it should not be in the actual canonical `BlobCert` structure.

disperser/controller/encoding_manager.go
mooselumph: Ah, ok 👍

disperser/controller/encoding_manager.go
ian-shim: My worry is that now there are two different data structures representing `BlobCert` (one used in disperser & relay, and another one in node). I'd rather keep one consistent struct across. Is updating the blob metadata that bad?

disperser/controller/encoding_manager.go
dmanc: In order to keep the encoder request idempotent that means we do something like ListObject on s3 and make sure stuff exists there?

cc: @cody-littley we were discussing this part in the morning.

disperser/controller/encoding_manager.go
ian-shim: Actually, ignore that. I ended up marshaling/unmarshaling them separately and putting them in BlobCertificate: https://github.com/Layr-Labs/eigenda/pull/846/commits/e76e3a62acddefcf9538aa731142d0fd8385f508

disperser/controller/encoding_manager.go
mooselumph: Hmm, that could be pretty slow. This seems a bit unideal. @ian-shim , do you have thoughts?

ian-shim: We don't have to list whole bucket. Since we already know the object names, we can check if those [objects exist](https://docs.aws.amazon.com/cli/latest/reference/s3api/head-object.html) right?

mooselumph (approved):

PR #848, opened 2024-10-30
Chunk store

cody-littley:

This PR adds a framework that will be used by encoders to upload data.

As requested by @dmanc, I split apart the logic for uploading proofs and uploading coefficients into separate methods. 

Since this functionality is needed to unblock @dmanc, I'm pushing it in a partially completed form. Namely, the framework does not currently break large files into smaller ones when pushing to S3. I plan on adding that as a follow up task.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


disperser/common/chunkstore/chunk_metadata_store.go
cody-littley, line 1: A note to the reviewers:

This file is intentionally left as a stub. I'm pushing it in an incomplete form so that other tasks that are dependent on the chunk store can proceed while I work on the chunk metadata store in parallel.

disperser/common/chunkstore/chunk_writer.go
```diff
@@ -0,0 +1,103 @@
+package chunkstore
```
ian-shim, line 1: I feel like chunkstore belongs outside `disperser`. Should it be under `relay`?

encoding/rs/frame.go
```diff
@@ -33,3 +37,61 @@ func Decode(b []byte) (Frame, error) {
 	}
 	return f, nil
 }
+
+// EncodeFrames serializes a slice of frames into a byte slice.
+func EncodeFrames(frames []*Frame) ([]byte, error) {
```
ian-shim, line 42: Why don't we use existing serialization methods like [this](https://github.com/Layr-Labs/eigenda/blob/master/core/data.go#L96)?

disperser/common/chunkstore/chunk_reader.go
```diff
@@ -0,0 +1,111 @@
...
+		proofs[i] = &proof
+	}
+
+	return proofs, nil
+}
+
+func (r *chunkReader) GetChunkCoefficients(
+	ctx context.Context,
+	blobKey disperser.BlobKey,
+	metadata *ChunkCoefficientMetadata) ([]*rs.Frame, error) {
```
ian-shim, line 92: Does chunk reader need `chunkMetadataStore` if this method takes metadata as input?

disperser/common/chunkstore/chunk_metadata_store.go
```diff
@@ -0,0 +1,42 @@
...
+
+import (
+	commondynamodb "github.com/Layr-Labs/eigenda/common/aws/dynamodb"
+	"github.com/Layr-Labs/eigenda/disperser"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	"time"
+)
+
+// ChunkMetadataStore is an interface for storing and retrieving metadata about chunks.
+type ChunkMetadataStore interface {
```
ian-shim, line 11: Not sure if we need another store abstraction for writing/reading chunk metadata. 
Since chunk metadata lives inside blob metadata, I think write/read should happen via [blob metadata store](https://github.com/Layr-Labs/eigenda/pull/846/files#diff-e7315f00f40f5ec590e1c7e8b6a46fb07d214931d8ea9d776fd57266dd6e7e39R74).

encoding/rs/frame.go
dmanc: From split encoder perspective we want to serialize the proofs and the coefficients separately. From node perspective we could make sure it receives the chunks in the expected serialized format.

disperser/common/chunkstore/chunk_writer.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/consensys/gnark-crypto/ecc/bn254"
+)
+
+// ChunkCoefficientMetadata contains metadata about how chunk coefficients are stored.
+// Required for reading chunk coefficients using ChunkReader.GetChunkCoefficients().
+type ChunkCoefficientMetadata struct {
+	// The total size of file containing all chunk coefficients for the blob.
+	DataSize int
+	// The maximum fragment size used to store the chunk coefficients.
+	FragmentSize int
```
anupsv, line 20: Could we be more specific here and say its uint64 or what's appropriate than generic int type ?

disperser/common/chunkstore/chunk_writer.go
```diff
@@ -0,0 +1,103 @@
...
+func (c *chunkWriter) PutChunkProofs(ctx context.Context, blobKey disperser.BlobKey, proofs []*encoding.Proof) error {
+	s3Key := blobKey.String()
+
+	bytes := make([]byte, 0, bn254.SizeOfG1AffineCompressed*len(proofs))
+	for _, proof := range proofs {
+		proofBytes := proof.Bytes()
+		bytes = append(bytes, proofBytes[:]...)
+	}
+
+	err := c.s3Client.UploadObject(ctx, c.bucketName, s3Key, bytes)
```
anupsv, line 67: What is our s3 object versioning policy ? would we need to care for objects that already exist before uploading ?

disperser/common/chunkstore/chunk_writer.go
cody-littley: Moved it to the `relay`.

I was actually a little confused about where this should live. The relay reads from it, but the encoder will be writing to it. But I guess `relay` is as good as anywhere (unless we decide to make it a top level directory, which I don't think it deserves to be).

encoding/rs/frame.go
cody-littley: @dmanc requested that the chunk store had the capability of uploading the `encoding.Proof` objects and `rs.Frame` objects separately. The code you point to is capable of serializing both at the same time, but does provide a way to serialize/deserialize them separately.

disperser/common/chunkstore/chunk_reader.go
cody-littley: Very good point, it won't need this to be passed in. I didn't notice because it's not actually used in this PR's iteration of the feature. Removed.

disperser/common/chunkstore/chunk_metadata_store.go
cody-littley: This was something separate since we had originally discussed not putting the extra metadata into the regular blob metadata store. Now that this data has merged into the other blob matadata, I agree that it doesn't make sense to have a separate chunk metadata store. Removed.

disperser/common/chunkstore/chunk_writer.go
cody-littley: Will do. Let's go with `uint32`.

As a side note, one of the quirks of golang that drives me up a wall is how they strongly encourage everybody to use the `int` type everywhere. For example, why does `len(x)` return a signed value? If I were in charge of the language design, I'd never have supported the types `int` and `uint` in the first place. /rant

disperser/common/chunkstore/chunk_writer.go
cody-littley: Discussed offline, repeating conclusion of that talk here for others.

Since each key is unique and has a deterministic value, writing a value to a key more than once is harmless (i.e. the data is overwritten with the exact same data).

encoding/rs/frame.go
ian-shim: I see. How about [this one](https://github.com/Layr-Labs/eigenda/blob/master/core/data.go#L376)?

ian-shim (commented):
lgtm! one comment re: serialization. Can you also take a look at the lint failure?

encoding/rs/frame.go
```diff
@@ -33,3 +37,106 @@ func Decode(b []byte) (Frame, error) {
 	}
 	return f, nil
 }
+
+// GnarkEncodeFrames serializes a slice of frames into a byte slice.
+func GnarkEncodeFrames(frames []*Frame) ([]byte, error) {
+
+	// Serialization format:
```
dmanc, line 44: Should we move this to above the function so it shows up in the go docs?

encoding/rs/frame.go
```diff
@@ -33,3 +37,106 @@ func Decode(b []byte) (Frame, error) {
...
+	}
+
+	frameCount := binary.BigEndian.Uint32(serializedFrame)
+	index := uint32(4)
+
+	if len(serializedFrame) < int(index+frameCount*encoding.BYTES_PER_SYMBOL) {
+		return nil, 0, fmt.Errorf("invalid frame size: %d", len(serializedFrame))
+	}
+
+	coeffs := make([]encoding.Symbol, frameCount)
```
dmanc, line 131: nit: I feel like encoding.Symbol is not really used anywhere. Maybe it's worth deprecating it and just using fr.Element.

```
// Symbol is a symbol in the field used for polynomial commitments
type Symbol = fr.Element
```

dmanc (approved):
Looks good, left a couple small comments

encoding/rs/frame.go
cody-littley: Done.

```
// GnarkEncodeFrames serializes a slice of frames into a byte slice.
//
// Serialization format:
// [number of frames: 4 byte uint32]
// [size of frame 1: 4 byte uint32][frame 1]
// [size of frame 2: 4 byte uint32][frame 2]
// ...
// [size of frame n: 4 byte uint32][frame n]
//
// Where relevant, big endian encoding is used.
func GnarkEncodeFrames(frames []*Frame) ([]byte, error) {
```

encoding/rs/frame.go
cody-littley: done

ian-shim (approved):

relay/chunkstore/chunk_reader.go
```diff
@@ -0,0 +1,110 @@
...
+		metadataStore: metadataStore,
+		client:        s3Client,
+		bucket:        bucketName,
+		shards:        shards,
+	}
+}
+
+func (r *chunkReader) GetChunkProofs(
+	ctx context.Context,
+	blobKey disperser.BlobKey) ([]*encoding.Proof, error) {
```
dmanc, line 58: Seems like this blobkey references the v1 blob key. In StoreBlob for V2 we use blobKey.Hex() = string https://github.com/Layr-Labs/eigenda/blob/c63dd61bf633c82ba6b30e7f71c5a17e5c452b7c/disperser/common/v2/blobstore/s3_blob_store.go#L27. 

V2 blob key:
```
type BlobKey [32]byte

func (b BlobKey) Hex() string {
	return hex.EncodeToString(b[:])
}
```

dmanc: Also when we fetch for proofs vs coefficients don't we need a different S3 key to differentiate it?

relay/chunkstore/chunk_writer.go
```diff
@@ -0,0 +1,103 @@
...
+	"github.com/Layr-Labs/eigenda/disperser"
+	"github.com/Layr-Labs/eigenda/encoding"
+	"github.com/Layr-Labs/eigenda/encoding/rs"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	"github.com/consensys/gnark-crypto/ecc/bn254"
+)
+
+// ChunkCoefficientMetadata contains metadata about how chunk coefficients are stored.
+// Required for reading chunk coefficients using ChunkReader.GetChunkCoefficients().
+type ChunkCoefficientMetadata struct {
```
dmanc, line 16: Are these the same? [encoding.FragmentInfo](https://github.com/Layr-Labs/eigenda/blob/c63dd61bf633c82ba6b30e7f71c5a17e5c452b7c/encoding/data.go#L91)

```
type FragmentInfo struct {
	TotalChunkSizeBytes uint32
	NumFragments        uint32
}
```

relay/chunkstore/chunk_reader.go
cody-littley: I've been assuming we'd use different buckets. Started a slack conversation to discuss. Will circle back on this once we decide how we want to handle buckets and name spacing.

cody-littley: I've switched over to using `v2.BlobKey` as recommended by @ian-shim.

PR #849, opened 2024-10-30
fix(disperser-client): RetrieveBlob grpc max size regression bug

samlaf:
A previous commit 6894828 (https://github.com/Layr-Labs/eigenda/pull/826) caused regression by forgetting to set grpc max msg size in retrieveBlob call

Caught by integration test from eigenda-proxy: https://github.com/Layr-Labs/eigenda-proxy/actions/runs/11584563970/job/32251942842?pr=196

Added unit-test here to catch any future regression on this feature. Also made that connection size a parameter of the disperser-client (keeps the default 100MiB that was previously set).

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/disperser_client.go
+++ b/api/clients/disperser_client.go
@@ -102,20 +102,36 @@ var _ DisperserClient = &disperserClient{}
 //
 //	// Subsequent calls will use the existing connection
 //	status2, requestId2, err := client.DisperseBlob(ctx, otherData, otherQuorums)
-func NewDisperserClient(config *Config, signer core.BlobRequestSigner) *disperserClient {
-	if config == nil {
-		config = &Config{}
-	}
-	if config.MaxRetrieveBlobSizeBytes == 0 {
-		// Set to 100MiB for forward compatibility.
-		// Check official documentation for current max blob size on mainnet.
-		config.MaxRetrieveBlobSizeBytes = 100 * 1024 * 1024
+func NewDisperserClient(config *Config, signer core.BlobRequestSigner) (*disperserClient, error) {
+	if err := checkConfigAndSetDefaults(config); err != nil {
+		return nil, fmt.Errorf("invalid config: %w", err)
 	}
 	return &disperserClient{
 		config: config,
 		signer: signer,
 		// conn and client are initialized lazily
+	}, nil
+}
+
+func checkConfigAndSetDefaults(c *Config) error {
+	if c == nil {
+		return fmt.Errorf("config is nil")
 	}
+	if c.Hostname == "" {
+		return fmt.Errorf("config.Hostname is empty")
+	}
+	if c.Port == "" {
+		return fmt.Errorf("config.Port is empty")
+	}
+	if c.Timeout == 0 {
+		return fmt.Errorf("config.Timeout is 0")
+	}
+	if c.MaxRetrieveBlobSizeBytes == 0 {
+		// Set to 100MiB for forward compatibility.
+		// Check official documentation for current max blob size on mainnet.
+		c.MaxRetrieveBlobSizeBytes = 100 * 1024 * 1024
+	}
+	return nil
 }
 
 // Close closes the grpc connection to the disperser server.
--- a/api/clients/disperser_client.go
+++ b/api/clients/disperser_client.go
@@ -32,7 +32,8 @@ type Config struct {
 	// but all other methods use the default 4MiB max message size, whereas RetrieveBlob
 	// potentially needs a larger size.
 	//
-	// If not set, the default value is 100MiB, to be backward compatible.
+	// If not set, default value is 100MiB for forward compatibility.
+	// Check official documentation for current max blob size on mainnet.
 	MaxRetrieveBlobSizeBytes int
 }
 
@@ -293,8 +294,8 @@ func (c *DisperserClient) RetrieveBlob(ctx context.Context, batchHeaderHash []by
 	ctxTimeout, cancel := context.WithTimeout(ctx, time.Second*60)
 	defer cancel()
 	if c.config.MaxRetrieveBlobSizeBytes == 0 {
-		// max blob size on mainnet is currently 16MiB but we set this to 100MiB for backward compatibility
-		// to what it originally was before this config was added
+		// Set to 100MiB for forward compatibility.
+		// Check official documentation for current max blob size on mainnet.
 		c.config.MaxRetrieveBlobSizeBytes = 100 * 1024 * 1024
 	}
 	reply, err := c.client.RetrieveBlob(ctxTimeout,
@@ -302,7 +303,7 @@ func (c *DisperserClient) RetrieveBlob(ctx context.Context, batchHeaderHash []by
 			BatchHeaderHash: batchHeaderHash,
 			BlobIndex:       blobIndex,
 		},
-		grpc.MaxCallRecvMsgSize(c.config.MaxRetrieveBlobSizeBytes))
+		grpc.MaxCallRecvMsgSize(c.config.MaxRetrieveBlobSizeBytes)) // for client
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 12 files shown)


api/clients/disperser_client.go
```diff
@@ -26,6 +26,14 @@ type Config struct {
 	// TODO: do we want to add config timeouts for those separate requests?
 	Timeout           time.Duration
 	UseSecureGrpcFlag bool
+	// MaxRetrieveBlobSizeBytes is the maximum size of a blob that can be retrieved by using
+	// the RetrieveBlob method. This is used to set the max message size for the grpc client.
+	// DisperserClient uses a single underlying grpc channel shared for all methods,
+	// but all other methods use the default 4MiB max message size, whereas RetrieveBlob
```
bxue-l2, line 32: I think dispersal and retrieval for a blob have the same blob length.

It is just the receiving ends are different. On our dispersal endpoint, we also set a very large buffer size. Then similarly, if you retrieve, you should set a large buffer. The buffer size on the sender side does not matter

api/clients/disperser_client.go
```diff
@@ -38,7 +46,7 @@ func NewConfig(hostname, port string, timeout time.Duration, useSecureGrpcFlag b
 	}
 }
 
-type DisperserClient interface {
+type IDisperserClient interface {
```
bxue-l2, line 49: I don't think we use this convention, this looks solidity

 what do you think @ian-shim

api/clients/disperser_client.go
```diff
@@ -273,18 +284,25 @@ func (c *disperserClient) GetBlobStatus(ctx context.Context, requestID []byte) (
...
 		return nil, api.NewErrorInternal(err.Error())
 	}
 
 	ctxTimeout, cancel := context.WithTimeout(ctx, time.Second*60)
 	defer cancel()
-	reply, err := c.client.RetrieveBlob(ctxTimeout, &disperser_rpc.RetrieveBlobRequest{
-		BatchHeaderHash: batchHeaderHash,
-		BlobIndex:       blobIndex,
-	})
+	if c.config.MaxRetrieveBlobSizeBytes == 0 {
```
bxue-l2, line 295: should we have a flag for this? this looks adhoc

api/clients/disperser_client.go
```diff
@@ -273,18 +284,25 @@ func (c *disperserClient) GetBlobStatus(ctx context.Context, requestID []byte) (
...
+		// max blob size on mainnet is currently 16MiB but we set this to 100MiB for backward compatibility
```
pschork, line 296: @samlaf are you planning to update this comment every time the disperser's `MAX_BLOB_SIZE` config is changed? I would rephrase this to `set to 100MiB for forward compatibility. check official documentation for current max blob size on mainnet`

api/clients/disperser_client.go
```diff
@@ -26,6 +26,14 @@ type Config struct {
 	// TODO: do we want to add config timeouts for those separate requests?
 	Timeout           time.Duration
 	UseSecureGrpcFlag bool
+	// MaxRetrieveBlobSizeBytes is the maximum size of a blob that can be retrieved by using
+	// the RetrieveBlob method. This is used to set the max message size for the grpc client.
+	// DisperserClient uses a single underlying grpc channel shared for all methods,
+	// but all other methods use the default 4MiB max message size, whereas RetrieveBlob
+	// potentially needs a larger size.
+	//
+	// If not set, the default value is 100MiB, to be backward compatible.
```
pschork, line 35: forward compatible

samlaf: It's actually backward compatible. The code before my previous PR (that caused the regression) had a default 100MiB.

api/clients/disperser_client.go
samlaf: I know, I don't like it either, but we already use this for eigenda_client, so I just followed the convention there.
Was forced to make the `disperserClient` struct public for the tests, so had to find another name for the interface. More golangy might be `DisperserClienter`. Honestly I think defining the interface here is code smell and should eventually all be refactored. Interfaces should be defined case-by-case at the call sites.

api/clients/disperser_client.go
samlaf: It does, both sides set their respective length that they want.
image

Disperser needs large buffer to receive large blobs for dispersal, and client needs large buffer size to retrieve blobs.

api/clients/disperser_client.go
samlaf: Ah just saw your previous comment, I see what you mean now! Good point

samlaf: [a07336c](https://github.com/Layr-Labs/eigenda/pull/849/commits/a07336cf2fce85b7d1cf886a3d07f3306ac459b8)

api/clients/disperser_client.go
samlaf: Good call: [a07336c](https://github.com/Layr-Labs/eigenda/pull/849/commits/a07336cf2fce85b7d1cf886a3d07f3306ac459b8)

api/clients/disperser_client.go
bxue-l2: commented in the wrong place, make c.config.MaxRetrieveBlobSizeBytes set in the constructor with the flag, so no if statement here

api/clients/disperser_client.go
bxue-l2: Ok, saw it. we used IEigenDAClient.

Looks fine to me

api/clients/disperser_client.go
bxue-l2: I mean MaxRetrieveBlobSizeBytes is always 0. since there is no flag to set it

api/clients/disperser_client.go
ian-shim: > Was forced to make the disperserClient struct public for the tests

Don't see any place that uses it. We couldn't use the constructor in the test?

api/clients/disperser_client.go
samlaf: It's not. constructor takes a config, which could have set MaxRetrieveBlobSizeBytes
```
func NewDisperserClient(config *Config, signer core.BlobRequestSigner) *DisperserClient {
```

And as I said above, I prefer not to set in constructor, because one can still build a DisperserClient without using the constructor. This is often done in tests for eg, like
```
client := &DisperserClient{
   Conn: ...
   Foo: ...
   
```

samlaf: Ahhhh scratch that... fields are private, so constructor actually has to be used. Ok, you're right, will update

samlaf: [04499a5](https://github.com/Layr-Labs/eigenda/pull/849/commits/04499a54c72071151d1bee67b1ba8e68d98328c6)

bxue-l2 (approved):

bxue-l2:
lgtm, but address Ian's comment

api/clients/disperser_client.go
samlaf: Ah, you're right. Originally was trying to put the tests in the same package so I can build the struct manually injecting the connection to test something. But then realized I can just add the config parameter.

samlaf: So hmmm... could revert to the old names. But somehow I feel like our clients should follow the same convention.

samlaf: Reverted: [dc98581](https://github.com/Layr-Labs/eigenda/pull/849/commits/dc985811c0bdc2d47e4e046c67b4756294258505)
Also added TODO to change eigenda-client to use same naming convention as disperser-client: [57d716f](https://github.com/Layr-Labs/eigenda/pull/849/commits/57d716fec20db630599a969ce1b4b7ecbe61242c)

api/clients/disperser_client.go
```diff
@@ -91,6 +103,14 @@ var _ DisperserClient = &disperserClient{}
 //	// Subsequent calls will use the existing connection
 //	status2, requestId2, err := client.DisperseBlob(ctx, otherData, otherQuorums)
 func NewDisperserClient(config *Config, signer core.BlobRequestSigner) *disperserClient {
+	if config == nil {
+		config = &Config{}
```
ian-shim, line 107: should it return error if config is nil?

ian-shim (approved):
lgtm, left one more comment!

api/clients/disperser_client.go
samlaf: yep good call: [f2100c8](https://github.com/Layr-Labs/eigenda/pull/849/commits/f2100c8f0055c27ac0678ae54e8f8c3eccc377cd)

PR #851, opened 2024-10-31
Use serialized bytes for commitments proto

ian-shim:
Instead of specifying each field of G1/G2 commitment, use serialized bytes to represent them like it does for encoding today. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proto/common/common.proto
```diff
@@ -9,24 +9,13 @@ message G1Commitment {
 	bytes y = 2;
 }
 
-message G2Commitment {
```
ian-shim, line 12: shouldn't be used other than in `BlobCommitment` in this file, which isn't in production use

api/proto/common/common.proto
```diff
@@ -9,24 +9,13 @@ message G1Commitment {
...
-	bytes x_a1 = 2;
-	// The A0 element of the Y coordinate of G2 point.
-	bytes y_a0 = 3;
-	// The A1 element of the Y coordinate of G2 point.
-	bytes y_a1 = 4;
-}
-
 // BlobCommitment represents commitment of a specific blob, containing its
 // KZG commitment, degree proof, the actual degree, and data length in number of symbols.
 message BlobCommitment {
```
ian-shim, line 14: only used in v2 protos, not in production use

disperser/api/proto/encoder/encoder.proto
```diff
@@ -7,7 +7,8 @@ service Encoder {
   rpc EncodeBlob(EncodeBlobRequest) returns (EncodeBlobReply) {}
 }
 
-// BlomCommitments contains the blob's commitment, degree proof, and the actual degree
+// BlobCommitments contains the blob's commitment, degree proof, and the actual degree
+// DEPRECATED: use common.BlobCommitment instead
```
ian-shim, line 11: Copying this existing scheme in `common` instead of moving because this is currently being used by encoders

api/proto/common/common.proto
```diff
@@ -9,24 +9,13 @@ message G1Commitment {
...
-	G1Commitment commitment = 1;
-	G2Commitment length_commitment = 2;
-	G2Commitment length_proof = 3;
-	uint32 data_length = 4;
+  bytes commitment = 1;
```
dmanc, line 15: nit: can we make sure the spacing is consistent in the file

disperser/api/proto/encoder/encoder.proto
dmanc: Once we are fully on v2 we can remove this?

dmanc (approved):

disperser/api/proto/encoder/encoder.proto
ian-shim: Yes!

PR #850, opened 2024-10-30
Revert "S3 relay interface"

cody-littley:
Reverts Layr-Labs/eigenda#833

Turns out this caused integration tests to fail. I will fix before attempting to re-merge this code.

PR #853, opened 2024-10-31
Revert "Revert "S3 relay interface""

cody-littley:
Reverts Layr-Labs/eigenda#850

I merged [this PR](https://github.com/Layr-Labs/eigenda/pull/833) which caused a CI breakage. I reverted the change to unblock others. This PR re-enables the changes with the required fixes in place.

There are two lines that deviate from the original PR:

- removal of "debug.PrintStack() // TODO do not merge" from file `inabox/deploy/localstack.go`. This was not the cause of the failure, but an intentional oversight that created some spam in stout.
- fixed one of the new flags in `common/cli.go` that was misconfiguration. This was the root cause of the failing integration-tests panel.


common/aws/s3/s3.go
```diff
@@ -2,10 +2,47 @@ package s3
...
 import "context"
 
+// Client encapsulates the functionality of an S3 client.
 type Client interface {
-	CreateBucket(ctx context.Context, bucket string) error
+
+	// DownloadObject downloads an object from S3.
 	DownloadObject(ctx context.Context, bucket string, key string) ([]byte, error)
+
+	// UploadObject uploads an object to S3.
```
samlaf, line 11: Thank you for adding these comments everywhere!

common/aws/s3/s3.go
```diff
@@ -2,10 +2,47 @@ package s3
...
+	// CreateBucket creates a bucket in S3.
+	CreateBucket(ctx context.Context, bucket string) error
+
+	// FragmentedUploadObject uploads a file to S3. The fragmentSize parameter specifies the maximum size of each
+	// file uploaded to S3. If the file is larger than fragmentSize then it will be broken into
+	// smaller parts and uploaded in parallel. The file will be reassembled on download.
+	//
+	// Note: if a file is uploaded with this method, only the FragmentedDownloadObject method should be used to
+	// download the file. It is not advised to use DeleteObject on files uploaded with this method (if such
+	// functionality is required, a new method to do so should be added to this interface).
```
samlaf, line 30: Interesting, how do we garbage collect this stuff after 2 weeks then? Or is the idea that we don't?

common/aws/s3/client.go
```diff
@@ -159,3 +175,162 @@ func (s *client) ListObjects(ctx context.Context, bucket string, prefix string)
...
+	resultChannel := make(chan error, len(fragments))
+
+	ctx, cancel := context.WithTimeout(ctx, s.cfg.FragmentWriteTimeout)
+	defer cancel()
+
+	for _, fragment := range fragments {
+		fragmentCapture := fragment
+		s.pool.Submit(func() {
+			s.fragmentedWriteTask(ctx, resultChannel, fragmentCapture, bucket)
+		})
```
samlaf, line 210: Any benefits of having our own worker pool implementation instead of using https://pkg.go.dev/golang.org/x/sync/errgroup?

common/aws/s3/fragment.go
```diff
@@ -0,0 +1,128 @@
+package s3
+
+import (
+	"fmt"
+	"sort"
+	"strings"
+)
+
+// GetFragmentCount returns the number of fragments that a file of the given size will be broken into.
+func GetFragmentCount(fileSize int, fragmentSize int) int {
```
samlaf, line 10: I feel like if these Fragment related functions are going to be in the s3 package, then they should all be private (lowerCase)? Can be tested by moving the `fragment_test.go` function inside the s3 package.

common/aws/s3/fragment.go
```diff
@@ -0,0 +1,128 @@
...
+}
+
+// GetFragmentKey returns the key for the fragment at the given index.
+//
+// Fragment keys take the form of "prefix/body-index[f]". The prefix is the first prefixLength characters
+// of the file key. The body is the file key. The index is the index of the fragment. The character "f" is appended
+// to the key of the last fragment in the series.
+//
+// Example: fileKey="abc123", prefixLength=2, fragmentCount=3
+// The keys will be "ab/abc123-0", "ab/abc123-1", "ab/abc123-2f"
```
samlaf, line 27: why is the prefix needed?

common/aws/test/fragment_test.go
```diff
@@ -0,0 +1,330 @@
...
+
+		if i == segmentCount-1 {
+			assert.True(t, strings.HasSuffix(fragmentKey, "f"))
+		} else {
+			assert.False(t, strings.HasSuffix(fragmentKey, "f"))
+		}
+	}
+}
+
+func TestExampleInGodoc(t *testing.T) {
```
samlaf, line 141: Is the goal for this function to appear in godoc? I think for that to happen the function name has to start with `Example` and not `Test`: https://go.dev/blog/examples

common/aws/test/client_test.go
```diff
@@ -0,0 +1,176 @@
...
+type clientBuilder struct {
+	// This method is called at the beginning of the test.
+	start func() error
+	// This method is called to build a new client.
+	build func() (s3.Client, error)
+	// This method is called at the end of the test when all operations are done.
+	finish func() error
+}
+
+var clientBuilders = []*clientBuilder{
```
samlaf, line 38: This is nice! The idea is to test every test against a mock (unit) and localstack (integration)?

common/aws/test/client_test.go
```diff
@@ -0,0 +1,176 @@
...
+	"testing"
+)
+
+var (
+	dockertestPool     *dockertest.Pool
+	dockertestResource *dockertest.Resource
+)
+
+const (
+	localstackPort = "4570"
```
samlaf, line 24: consider using 0 to let the os assign a random port for each instantiation of localstack. With this you can then run all tests in parallel by using `t.parallel()` at the beginning of your tests: https://pkg.go.dev/testing#T.Parallel

common/aws/s3/s3.go
```diff
@@ -2,10 +2,47 @@ package s3
 
 import "context"
 
+// Client encapsulates the functionality of an S3 client.
 type Client interface {
-	CreateBucket(ctx context.Context, bucket string) error
```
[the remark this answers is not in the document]
cody-littley: Not a mistake. Ian and I both added an identical `CreateBucket()` function with identical API in parallel PRs, and the resulting merge changed the location of the method in the file, but does not remove it..

common/aws/s3/s3.go
cody-littley: Documenting every function is one of my pet peeves. :)

common/aws/s3/s3.go
cody-littley: The current use case for this method will use a bucket that enforces a TTL for all files inside of it, so we won't have to do any sort of manual deletion. Straight forward to add methods for deleting fragmented files, but my instinct is to avoid adding code that I don't have an obvious current use for.

common/aws/s3/fragment.go
cody-littley: Done. Easy enough to make them public again later if we discover a need.

common/aws/s3/fragment.go
cody-littley: S3 throttles bandwidth based on prefix. That means reading and writing keys with two different prefixes has an aggregate upper bandwidth limit twice as high as reading and writing keys with a single prefix.

https://docs.aws.amazon.com/AmazonS3/latest/userguide/optimizing-performance.html

> For example, your application can achieve at least 3,500 PUT/COPY/POST/DELETE or 5,500 GET/HEAD requests per second per partitioned Amazon S3 prefix. There are no limits to the number of prefixes in a bucket. You can increase your read or write performance by using parallelization. For example, if you create 10 prefixes in an Amazon S3 bucket to parallelize reads, you could scale your read performance to 55,000 read requests per second. Similarly, you can scale write operations by writing to multiple prefixes.

The goal here is to spread the data into different prefixes so that we have a much higher upper limit on bandwidth.

common/aws/test/fragment_test.go
cody-littley: I give an example in the godoc of the function `getFragmentKey()`. Updated the godoc to mention that:

```
// TestExampleInGodoc tests the example provided in the documentation for getFragmentKey().
//
// Example: fileKey="abc123", prefixLength=2, fragmentCount=3
// The keys will be "ab/abc123-0", "ab/abc123-1", "ab/abc123-2f"
```

common/aws/test/client_test.go
cody-littley: Yup! It's kind of like a meta-test to sanity check the unit tests. Since the mock implementation is super simple and easy to reason about, it helps me to distinguish between bugs in the unit test definition and bugs in the implementation of the code.

common/aws/test/client_test.go
cody-littley: This is the only port number I can get to work. When this test runs in the github CI job, there is already a local stack instance with this port. In that environment, my test harness doesn't start a new instance, it just uses the existing one. Although it was easy to get the test running locally with a port of my choosing, I was unable to get that working in the CI pipeline.

common/aws/cli.go
```diff
@@ -3,20 +3,48 @@ package aws
...
+	FragmentParallelismFactor int
+	// FragmentParallelismConstant helps determine the size of the pool of workers to help upload/download files.
+	// A non-zero value for this parameter adds a constant number of workers. Default is 0.
+	FragmentParallelismConstant int
+	// FragmentReadTimeout is used to bound the maximum time to wait for a single fragmented read.
+	// Default is 30 seconds.
+	FragmentReadTimeout time.Duration
+	// FragmentWriteTimeout is used to bound the maximum time to wait for a single fragmented write.
+	// Default is 30 seconds.
+	FragmentWriteTimeout time.Duration
```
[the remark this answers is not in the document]
cody-littley: Your suggestion makes sense. The core difficulty is that `aws.ClientConfig` is currently used to build an S3 client by `disperser/cmd/apiserver`, `disperser/cmd/batcher`, and `disperser/cmd/dataapi`. Even harder is the fact that these use cases use the `aws.ClientConfig` object to instantiate both an s3 client and a dynamoDB client, meaning that if I nest `aws.ClientConfig` inside S3 config then it gets duplicated.

I'm sure it's possible to untangle the mess I describe above, but my concern is scope creep for this PR. I've got some ideas for how we might simplify the way we manage our project's configuration, would you be interested in discussing this offline?

common/aws/s3/s3.go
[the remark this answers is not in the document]
samlaf: Oh yeah I totally missed that it was being readded.

common/aws/cli.go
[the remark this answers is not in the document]
samlaf: Yes, interested in discussing, no need for scope creep I agree, since these are only used by us internally so its fine to have breaking changes.

That being said, I don't see why having the duplicate ClientConfig is bad. In fact, one could want an s3 client and dynamoDB client with different aws configs (say the dbs live in different regions for eg).

common/aws/s3/s3.go
samlaf: Agree with not adding a method that we don't currently need! Maybe document the TTL behavior here though?

common/aws/s3/client.go
samlaf: errgroup has a limit feature now! https://pkg.go.dev/golang.org/x/sync/errgroup#Group.SetLimit

common/aws/test/client_test.go
samlaf: I see.... that's kinda weird. So our github workflow starts a global localstack instance before running the tests? Feel like the localstack should be controlled by the tests and started by them, but maybe there's something I'm missing..

common/aws/s3/fragment.go
samlaf: This is cool, didn't know about it. But I'm still a bit puzzled by how you are making use of this though. Given that the prefix length is a global config that is always the same, then all fragments of a file will have the same prefix right? Compare
```
"ab/abc123-0", "ab/abc123-1", "ab/abc123-2f"
```
to
```
"abc123-0", "abc123-1", "abc123-2f"
```

All fragments in both cases have the same prefix, so I don't quite understand how using an extra prefix is helping here. Am I missing something?

ian-shim: +1 
The prefix helps with creating virtual hierarchy within the bucket, but all files with the same prefix is already in the same bucket, right?

ian-shim (approved):
lgtm
Please address Sam's comments before merging

common/aws/s3/s3.go
cody-littley: Added the following:

```
	// Note: if this operation fails partway through, some file fragments may have made it to S3 and others may not.
	// In order to prevent long term accumulation of fragments, it is suggested to use this method in conjunction with
	// a bucket configured to have a TTL.
```

common/aws/s3/fragment.go
cody-littley: All files end up in the same bucket. The prefix has nothing to do with which bucket each file is written to. Our use of prefixes is entirely a tactic to improve our bandwidth with S3, not as a mechanism used to organize our data for our own purposes.

common/aws/s3/client.go
cody-littley: Neat. Switched over to using errgroup.

common/aws/s3/fragment.go
samlaf: I see. Ok so the idea is that all files are in the same bucket, which means without prefixes all reads/writes to this bucket would have a total bandwidth limit. Since we want a per-file bandwidth limit, you add these prefixes.

Why don't you just use "-" as the prefix delimiter then, instead of adding this prefix? Still feels to me like the real useful prefix is the file name, not this deterministically shortened version of it (which I still think isn't useful), no?

On a side note, how exactly are buckets organized? When do we create a new bucket, if ever?

PR #858, opened 2024-11-01
fix(api): failover error bugs

samlaf:

2 big bugs in my ErrorFailover implementation:
1. I misunderstood how errors.Is works, so could only compare if the underlying wrapped error was exactly equal, which is not useful (and made the logic in disperser/eigenda-clients incorrect)
2. err.Error() would panic when ErrorFailover{} was created with a nil wrapped error

Added tests to make sure the above are fixed, forever.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/errors.go
+++ b/api/errors.go
@@ -77,6 +77,13 @@ func NewErrorDeadlineExceeded(msg string) error {
 //  1. Failed to put the blob in the disperser's queue (disperser is down)
 //  2. Timed out before getting confirmed onchain (batcher is down)
 //  3. Insufficient signatures (eigenda network is down)
+//
+// One can check if an error is an ErrorFailover by using errors.Is:
+//
+//	failoverErr := NewErrorFailover(someOtherErr)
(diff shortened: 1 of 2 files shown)


bxue-l2 (approved):

api/errors.go
```diff
@@ -90,9 +97,28 @@ func NewErrorFailover(err error) *ErrorFailover {
...
+//
+//	baseErr := fmt.Errorf("some error")
+//	failoverErr := NewErrorFailover(baseErr)
+//	wrappedFailoverErr := fmt.Errorf("some extra context: %w", failoverErr)
+//
+//	if !errors.Is(wrappedFailoverErr, &ErrorFailover{}) {
+//	  // do something...
+//	}
+func (e *ErrorFailover) Is(target error) bool {
+	_, ok := target.(*ErrorFailover)
```
bxue-l2, line 122: As is probably better

bxue-l2: https://pkg.go.dev/errors

PR #859, opened 2024-11-01
style: remove unneeded lint checkbox in pull_request_template.md

samlaf:

Not sure why we have this checkbox since we have a CI action that lints. We should instead just make it impossible to merge a PR where the lint workflow is failing (I can't change these settings because not admin on this repo).

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -4,7 +4,6 @@
 
 ## Checks
 
-- [ ] I've made sure the lint is passing in this PR.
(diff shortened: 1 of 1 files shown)


dmanc:
We could also shift this further left by using pre-commit hooks 
- https://github.com/golangci/golangci-lint/blob/master/.pre-commit-hooks.yaml
- https://pre-commit.com/

jianoaix:
It's there maybe due to that we were in a unhealthy state long time ago. But sounds fine to make lint non-optional and auto enforced (hence remove it from check box).

jianoaix (approved):

PR #860, opened 2024-11-04
Cache operator state

ian-shim:
Integrating operator state cache directly into `ChainState`

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/indexer/state.go
+++ b/core/indexer/state.go
@@ -2,16 +2,20 @@ package indexer
 
 import (
 	"context"
+	"encoding/binary"
 	"errors"
 
 	"github.com/Layr-Labs/eigenda/core"
 	"github.com/Layr-Labs/eigenda/indexer"
+	lru "github.com/hashicorp/golang-lru/v2"
 )
 
 type IndexedChainState struct {
 	core.ChainState
 
 	Indexer indexer.Indexer
+
+	operatorStateCache *lru.Cache[string, *core.IndexedOperatorState]
 }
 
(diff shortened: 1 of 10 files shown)


core/indexer/state.go
```diff
@@ -138,3 +162,10 @@ func (ics *IndexedChainState) getObjects(blockNumber uint) (*OperatorPubKeys, Op
 	return pubkeys, sockets, nil
 
 }
+
+func computeCacheKey(blockNumber uint, quorumIDs []uint8) string {
+	bytes := make([]byte, 8+len(quorumIDs))
```
anupsv, line 167: would be good to comment as to why 8+value

core/indexer/state.go
```diff
@@ -2,28 +2,42 @@ package indexer
...
 func NewIndexedChainState(
 	chainState core.ChainState,
 	indexer indexer.Indexer,
+	cacheSize int,
 ) (*IndexedChainState, error) {
-
+	operatorStateCache := (*lru.Cache[string, *core.IndexedOperatorState])(nil)
+	var err error
+	if cacheSize > 0 {
+		operatorStateCache, err = lru.New[string, *core.IndexedOperatorState](cacheSize)
```
anupsv, line 31: 2 questions:
1. Should we set a max cache size based on available memory etc?
2. Is there a default expiry of the cache ?

ian-shim:
realized this won't be needed in the short term with batching. will close this one for now

PR #861, opened 2024-11-04
Add cache wrapper that handles parallel access.

cody-littley:

This PR adds a wrapper around `lru.Cache`. The purpose of this wrapper is to facilitate efficient lookup if two RPCs request the same data from the relay concurrently. Ideally we only want to fetch the data once in order to serve both requests.

Example scenario without cache wrapper:
- at `t=0`, client A requests data X. X is not present in the cache, so the relay requests it from S3.
- at `t=50`, client B requests data X. X is not present in the cache, so the relay requests it from S3.
- at `t=200`, the request initiated by client A is completed. X is now present in the cache.
- at `t=250`, the request initiated by client B is completed.

Example scenario with cache wrapper:
- at `t=0`, client A requests data X. X is not present in the cache, so the relay requests it from S3.
- at `t=50`, client B requests data X. X is not present in the cache, but since the lookup is already in progress, B's request is paused until A's request is completed.
- at `t=200`, the request initiated by client A is completed. X is now present in the cache.
- at `t=201`, the request initiated by client B is wakes up and uses the same data that was used to serve the request from client A.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/relay/cached_accessor.go
+++ b/relay/cached_accessor.go
@@ -34,14 +34,12 @@ var _ CachedAccessor[string, string] = &cachedAccessor[string, string]{}
 
 // cachedAccessor is an implementation of CachedAccessor.
 type cachedAccessor[K comparable, V any] struct {
-	// used to synchronize access to the cache and lookupsInProgress
-	lock sync.Mutex
 
 	// lookupsInProgress has an entry for each key that is currently being looked up via the accessor. The value
 	// is written into the channel when it is eventually fetched. If a key is requested more than once while a
 	// lookup in progress, the second (and following) requests will wait for the result of the first lookup
 	// to be written into the channel.
-	lookupsInProgress map[K]*accessResult[V]
+	lookupsInProgress *sync.Map
 
 	// cache is the LRU cache used to store values fetched by the accessor.
 	cache *lru.Cache[K, *V]
@@ -59,8 +57,7 @@ func NewCachedAccessor[K comparable, V any](cacheSize int, accessor Accessor[K, 
 	}
 
 	return &cachedAccessor[K, V]{
-		lock:              sync.Mutex{},
-		lookupsInProgress: make(map[K]*accessResult[V]),
+		lookupsInProgress: &sync.Map{},
 		cache:             cache,
 		accessor:          accessor,
 	}, nil
@@ -82,21 +79,16 @@ func (c *cachedAccessor[K, V]) Get(key K) (*V, error) {
 	}
 
 	// if that fails, check if a lookup is already in progress. If not, start a new one.
-	c.lock.Lock()
-	result, lookupInProgress := c.lookupsInProgress[key]
-	if !lookupInProgress {
-		// No lookup in progress, start a new one.
-		result = c.newAccessResult()
-		c.lookupsInProgress[key] = result
-	}
-	c.lock.Unlock()
+	result := c.newAccessResult()
+	actual, alreadyLoading := c.lookupsInProgress.LoadOrStore(key, result)
+	result = actual.(*accessResult[V]) // sync.Map was written prior to generics in golang ;(
 
-	if lookupInProgress {
-		// Wait for the value to be fetched by the other goroutine.
+	if alreadyLoading {
+		// The result is being fetched on another goroutine. Wait for it to finish.
 		result.wg.Wait()
 		return result.value, result.err
 	} else {
-		// We are responsible for fetching the value.
+		// We are the first goroutine to request this key.
 		value, err := c.accessor(key)
 
 		// Update the cache if the fetch was successful.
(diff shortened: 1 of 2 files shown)


relay/cached_accessor.go
```diff
@@ -0,0 +1,118 @@
+package relay
+
+import (
+	lru "github.com/hashicorp/golang-lru/v2"
+	"sync"
+)
+
+// CachedAccessor is an interface for accessing a resource that is cached. It assumes that cache misses
+// are expensive, and prevents multiple concurrent cache misses for the same key.
+type CachedAccessor[K comparable, V any] interface {
```
ian-shim, line 10: Does it make sense to not generalize this interface for now as the only use case is retrieving blobs from the store? I'm generally against generalizing something like this unless there are 3+ immediate use cases.
I think the interface will be a lot simpler (and don't need to pass in function as `Accessor`) if we just made this `CachedBlobStore` or something that has the s3 store and accessor logic implemented directly. Wdyt?

relay/cached_accessor.go
```diff
@@ -0,0 +1,118 @@
...
+// cachedAccessor is an implementation of CachedAccessor.
+type cachedAccessor[K comparable, V any] struct {
+	// used to synchronize access to the cache and lookupsInProgress
+	lock sync.Mutex
+
+	// lookupsInProgress has an entry for each key that is currently being looked up via the accessor. The value
+	// is written into the channel when it is eventually fetched. If a key is requested more than once while a
+	// lookup in progress, the second (and following) requests will wait for the result of the first lookup
+	// to be written into the channel.
+	lookupsInProgress map[K]*accessResult[V]
```
ian-shim, line 44: maybe we can use `sync.Map` and ditch the mutex?

relay/cached_accessor.go
cody-littley: I agree with your general sentiment on this, making something generic adds complexity to its implementation. But in this case, there are actually three types of data that I'd like to use this pattern for:

- fetching blob metadata
- fetching unencoded blobs
- fetching encoded chunks

I expect that preventing parallel fetches for each of these data types in the advent of a cache miss will be beneficial.

relay/cached_accessor.go
cody-littley: Done.

relay/cached_accessor.go
```diff
@@ -0,0 +1,109 @@
...
+	}
+
+	return &cachedAccessor[K, V]{
+		lookupsInProgress: &sync.Map{},
+		cache:             cache,
+		accessor:          accessor,
+	}, nil
+}
+
+func (c *cachedAccessor[K, V]) newAccessResult() *accessResult[V] {
```
ian-shim, line 66: should this be a static method?

relay/cached_accessor.go
```diff
@@ -0,0 +1,109 @@
...
+// whose size can be specified by memory footprint in bytes.
+
+// cachedAccessor is an implementation of CachedAccessor.
+type cachedAccessor[K comparable, V any] struct {
+
+	// lookupsInProgress has an entry for each key that is currently being looked up via the accessor. The value
+	// is written into the channel when it is eventually fetched. If a key is requested more than once while a
+	// lookup in progress, the second (and following) requests will wait for the result of the first lookup
+	// to be written into the channel.
+	lookupsInProgress *sync.Map
```
ian-shim, line 42: nit: you don't need to initialize `lookupsInProgress` if it's typed as `sync.Map`

relay/cached_accessor_test.go
```diff
@@ -0,0 +1,248 @@
...
+			value, err := ca.Get(0)
+			require.NoError(t, err)
+			require.Equal(t, baseData[0], *value)
+		}()
+	}
+
+	// Wait for the goroutines to start. We want to give the goroutines a chance to do naughty things if they want.
+	// Eliminating this sleep will not cause the test to fail, but it may cause the test not to exercise the
+	// desired race condition.
+	time.Sleep(100 * time.Millisecond)
```
ian-shim, line 166: Sleep is generally not reliable in tests. i.e. I don't think it's guaranteed that it always sleep the right amount in test and it can result in flaky tests. 
Maybe we can use something like a buffered channel to wait for all goroutines to trigger?

relay/cached_accessor.go
ian-shim: I see. Makes sense

ian-shim (approved):
💯

relay/cached_accessor.go
cody-littley: done

relay/cached_accessor.go
cody-littley: neat, fixed

cody-littley: I ended up having to back out this change in favor of using a mutex. The core issue is that reading the `lookupsInProgress` map needs to be atomic with respect to reading from the cache. Was able to provoke a race condition in a unit test that caused an unnecessary cache miss.

relay/cached_accessor_test.go
cody-littley: In this scenario, I'm not relying on the sleep to make the test pass. The purpose of the sleep statement is to allow background threads a chance to do bad things if they want to do bad things.

In order to show that test stability does not depend on the sleep, I've set this up so that it runs twice: once with the sleep and once without it. Are you ok with this approach? If you'd still prefer to avoid having a sleep, we should chat. One option might just be to remove the sleep entirely.

PR #863, opened 2024-11-05
[SocketRegistry] get operator socket

hopeyen:

Check on-chain socket registry state with indexed operator info

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/eth/reader.go
+++ b/core/eth/reader.go
@@ -678,7 +678,14 @@ func (t *Reader) GetReservationWindow(ctx context.Context) (uint32, error) {
 }
 
 func (t *Reader) GetOperatorSocket(ctx context.Context, operatorId core.OperatorID) (string, error) {
-	return t.bindings.SocketRegistry.GetOperatorSocket(&bind.CallOpts{
+	socket, err := t.bindings.SocketRegistry.GetOperatorSocket(&bind.CallOpts{
 		Context: ctx,
 	}, [32]byte(operatorId))
+	if err != nil {
+		return "", err
+	}
+	if socket == "" {
+		return "", errors.New("operator socket string is empty, check operator with id: " + operatorId.Hex())
+	}
+	return socket, nil
 }
--- a/core/eth/reader.go
+++ b/core/eth/reader.go
@@ -15,6 +15,7 @@ import (
 	indexreg "github.com/Layr-Labs/eigenda/contracts/bindings/IIndexRegistry"
 	opstateretriever "github.com/Layr-Labs/eigenda/contracts/bindings/OperatorStateRetriever"
 	regcoordinator "github.com/Layr-Labs/eigenda/contracts/bindings/RegistryCoordinator"
+	socketreg "github.com/Layr-Labs/eigenda/contracts/bindings/SocketRegistry"
 	stakereg "github.com/Layr-Labs/eigenda/contracts/bindings/StakeRegistry"
 	"github.com/Layr-Labs/eigenda/core"
 	"github.com/Layr-Labs/eigensdk-go/logging"
@@ -37,6 +38,7 @@ type ContractBindings struct {
 	EigenDAServiceManager *eigendasrvmg.ContractEigenDAServiceManager
 	EjectionManager       *ejectionmg.ContractEjectionManager
 	AVSDirectory          *avsdir.ContractAVSDirectory
+	SocketRegistry        *socketreg.ContractSocketRegistry
 }
 
 type Reader struct {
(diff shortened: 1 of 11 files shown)


node/node.go
```diff
@@ -223,6 +222,15 @@ func (n *Node) Start(ctx context.Context) error {
 			return fmt.Errorf("failed to register the operator: %w", err)
 		}
 	} else {
+		registeredSocket, err := n.Transactor.GetOperatorSocket(ctx, n.Config.ID)
+		// Error out if registration on-chain is a requirement
```
hopeyen, line 226: I was thinking that we can replace indexed operator info after testing here?

mooselumph (approved):
lgtm

disperser/dataapi/queried_operators_handlers.go
```diff
@@ -295,6 +295,19 @@ func (s *server) scanOperatorsHostInfo(ctx context.Context) (*SemverReportRespon
 	if err != nil {
 		return nil, fmt.Errorf("failed to fetch indexed operator info - %s", err)
 	}
+
+	// check operator socket registration against the indexed state
+	for operatorID, operatorInfo := range operators {
+		socket, err := s.chainState.GetOperatorSocket(context.Background(), currentBlock, operatorID)
```
ian-shim, line 301: will this make extra N RPC calls?

core/eth/reader.go
```diff
@@ -619,3 +640,9 @@ func (t *Reader) GetReservationWindow(ctx context.Context) (uint32, error) {
 	// contract is not implemented yet
 	return 0, nil
 }
+
+func (t *Reader) GetOperatorSocket(ctx context.Context, operatorId core.OperatorID) (string, error) {
```
ian-shim, line 644: whenever we call this, should we check if the string is empty?

ian-shim (commented):
Looks like we're adding these checks to ensure there's no discrepancy.
Are these check some temporary measure or going to stay here forever?

node/node.go
```diff
@@ -223,6 +222,15 @@ func (n *Node) Start(ctx context.Context) error {
...
+		if err != nil {
+			n.Logger.Warn("failed to get operator socket: %w", err)
```
pschork, line 228: `Warnf`

node/node.go
```diff
@@ -223,6 +222,15 @@ func (n *Node) Start(ctx context.Context) error {
...
+		}
+		if registeredSocket != socket {
+			n.Logger.Warn("registered socket %s does not match expected socket %s", registeredSocket, socket)
```
pschork, line 231: `Warnf`

disperser/dataapi/queried_operators_handlers.go
hopeyen: Yes, there's no multicast call available from the contract unfortunately

core/eth/reader.go
hopeyen: 🤔 I don't think the check is necessary, we will return whatever the contract returns with an empty string. Although this is a local optimization we can make, if we expect the input operator id is ever empty?

hopeyen:
> Are these check some temporary measure or going to stay here forever?

I think the checks are temporary and should be removed once we are confident we don't need the indexed operator state for the socket

core/eth/reader.go
ian-shim: hmm should we just place that check in this method? i.e. return error if result is empty

disperser/dataapi/queried_operators_handlers.go
ian-shim: @pschork how often is `scanOperatorsHostInfo` called? I believe we have some cronjob?

core/eth/reader.go
hopeyen: If the result is empty, then the check against the expected operator socket would fail. Is there a reason we want to handle it here?

hopeyen: updated to check for empty string here

pschork (approved):

disperser/dataapi/queried_operators_handlers.go
pschork: It's called every 12hrs, but the dataAPI is public, so anyone can call it on demand

hopeyen: 🤔 should we add authentication to this method so not everyone can call it?

ian-shim (approved):

PR #865, opened 2024-11-05
Use keys that do not collide when data is stored in a single bucket.

cody-littley:

This PR makes it so that the data stored in S3 does not have name collisions between blobs, chunks, and proofs. Prior to this change, I had been thinking that we'd use different buckets. After this merges, all v2 data will be able to be stored in a single bucket.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


dmanc (approved):
:shipit:

PR #869, opened 2024-11-07
Implement core relay functionality

cody-littley:

This PR adds core functionality for the relay server. The following features have intentionally been omitted, and will be merged as follow-up PRs:

- main function entrypoint
- rate limiting
- request timeouts
- benchmarks
- authentication for GetChunks() requests
- metrics

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proto/relay/relay.proto
```diff
@@ -9,43 +9,23 @@ option go_package = "github.com/Layr-Labs/eigenda/api/grpc/relay";
 
 // Relay is a service that provides access to public relay functionality.
 service Relay {
```
ian-shim, line 11: This change looks good to me. Do you want to have a PR with this change and merge it while working on the relay? 
This could unblock node endpoint implementation.

ian-shim: Should we update `GetChunksRequest` to allow ranges as well?

relay/metadata_server.go
```diff
@@ -0,0 +1,153 @@
+package relay
```
ian-shim, line 1: Why do we need relay to serve blob metadata? 
I think the only service it needs are blobs & chunks delivery

cody-littley: This is for internal relay use. When an external user requests data, the relay needs to first fetch the metadata. It uses the metadata to decide if the blob is in the correct relay shard, as well as to figure out the proper file sizes for downloading fragmented files.

ian-shim: Ah I should have taken a closer look. This is a library for fetching the metadata, not a server 😭 
Can we rename it to something other than "server"

api/proto/relay/relay.proto
```diff
@@ -9,43 +9,23 @@ option go_package = "github.com/Layr-Labs/eigenda/api/grpc/relay";
...
-    bytes blob = 1;
-    // An error (if any) that occurred while trying to fetch the blob.
-    RelayError error = 2;
-  }
+message GetBlobReply {
+  // The blob requested.
+  bytes blob = 1;
 }
 
 // Request chunks from blobs stored by this relay.
```
mooselumph, line 31: I think there's an argument to be made that we only need to be able to address `(list of keys)\times(list of indices)` instead of `list of (keys,list of indices)`, since DA nodes will always be requesting the same set of indices for the blobs in a given batch. 

This will decrease the size of the request that the DA nodes are sending. I think this would have been a larger benefit for light node sampling since light nodes must send a list of indices. But maybe still worth considering.

.github/workflows/golangci-lint.yml
```diff
@@ -24,4 +24,4 @@ jobs:
         uses: golangci/golangci-lint-action@v3
         with:
           version: v1.60
-          args: --timeout 3m --verbose
+          args: --timeout 3m --verbose --out-format=colored-line-number
```
cody-littley, line 27: Note to the reviewers:

As previously configured, if lint failed during the CI job, it would not tell you which line needed to be corrected. Normally this isn't a problem, since I just run lint locally and fix whatever it tells me to. However, I encountered a rule that was being enforced by the CI job but not my local lint tool. This extra argument causes the CI job to tell you which line needs to be corrected.

relay/chunk_server_test.go
```diff
@@ -0,0 +1,158 @@
...
+		fMap, err := server.GetFrames(context.Background(), &mMap)
+		require.NoError(t, err)
+
+		require.Equal(t, 1, len(*fMap))
+		readFrames := (*fMap)[key]
+		require.NotNil(t, readFrames)
+
+		// TODO: when I inspect this data using a debugger, the proofs are all made up of 0s... something
+		//  is wrong with the way the data is generated in the test.
+		require.Equal(t, frames, readFrames)
```
cody-littley, line 69: @ian-shim prior to merge, I'd like to talk to you about how to properly generate data here.

relay/metadata_server.go
cody-littley: I've renamed these helper functions to `blobManager`, `chunkManager`, and `metadataManager`.

api/proto/relay/relay.proto
cody-littley: Change made.

api/proto/relay/relay.proto
ian-shim: Thanks! Can you rebase this PR onto master?

api/proto/relay/relay.proto
ian-shim: > since DA nodes will always be requesting the same set of indices for the blobs in a given batch

I don't think this is necessarily true. For any given blob, a node may have to download two different sets of index ranges because the assigned index range is different for each quorum?

mooselumph: Good point. I suppose the the node could make one request per quorum / batch?

relay/chunk_manager.go
```diff
@@ -0,0 +1,165 @@
...
+
+	keys := make([]*blobKeyWithMetadata, 0, len(*mMap))
+	for k, v := range *mMap {
+		keys = append(keys, &blobKeyWithMetadata{blobKey: k, metadata: *v})
+	}
+
+	fMap := make(frameMap, len(keys))
+	hadError := atomic.Bool{}
+
+	wg := sync.WaitGroup{}
```
ian-shim, line 85: same comment about the wg and errgroup

relay/chunk_manager.go
```diff
@@ -0,0 +1,165 @@
...
+			proofs, proofsErr = s.chunkReader.GetChunkProofs(s.ctx, key.blobKey)
+			return nil
+		})
+
+	fragmentInfo := &encoding.FragmentInfo{
+		TotalChunkSizeBytes: key.metadata.totalChunkSizeBytes,
+		FragmentSizeBytes:   key.metadata.fragmentSizeBytes,
+	}
+
+	coefficients, err := s.chunkReader.GetChunkCoefficients(s.ctx, key.blobKey, fragmentInfo)
```
ian-shim, line 132: should this request also execute from the pool?

relay/config.go
```diff
@@ -0,0 +1,41 @@
...
+
+	// ChunkCacheSize is the size of the chunk cache. Default is 32.
+	ChunkCacheSize int
+
+	// ChunkWorkPoolSize is the size of the chunk work pool. Default is 32.
+	ChunkWorkPoolSize int
+}
+
+// DefaultConfig returns the default configuration for the relay Server.
+func DefaultConfig() *Config {
```
ian-shim, line 32: Is this used anywhere?

relay/metadata_manager.go
```diff
@@ -0,0 +1,154 @@
...
+}
+
+// metadataMap is a map of blob keys to metadata.
+type metadataMap map[v2.BlobKey]*blobMetadata
+
+// GetMetadataForBlobs retrieves metadata about multiple blobs in parallel.
+func (m *metadataManager) GetMetadataForBlobs(keys []v2.BlobKey) (*metadataMap, error) {
+
+	mapLock := sync.Mutex{}
+	mMap := make(metadataMap)
```
ian-shim, line 89: nit: does `sync.Map` work here?

relay/metadata_manager.go
```diff
@@ -0,0 +1,154 @@
...
+			mapLock.Lock()
+			mMap[boundKey] = metadata
+			mapLock.Unlock()
+
+			return nil
+		})
+	}
+	wg.Wait()
+
+	if hadError.Load() {
```
ian-shim, line 118: should it early terminate if any of the requests failed instead of waiting all requests to finish?

relay/blob_manager.go
```diff
@@ -0,0 +1,96 @@
...
+	server.blobCache = cache
+
+	return server, nil
+}
+
+// GetBlob retrieves a blob from the blob store.
+func (s *blobManager) GetBlob(blobKey v2.BlobKey) ([]byte, error) {
+
+	// Even though we don't need extra parallelism here, we still use the work pool to ensure that we don't
+	// permit too many concurrent requests to the blob store.
```
dmanc, line 61: What happens if there are too many concurrent requests to S3? I'm wondering if it could be worth using this blob manager on the encoder (although it doesn't need the caching part).

See: https://github.com/Layr-Labs/eigenda/blob/94ab42f939b12d7e4ea3957aa488a460ef5d76d0/disperser/encoder/server_v2.go#L128

relay/config.go
```diff
@@ -0,0 +1,41 @@
...
+import core "github.com/Layr-Labs/eigenda/core/v2"
+
+// Config is the configuration for the relay Server.
+type Config struct {
+
+	// Shards contains the IDs of the relays that this server is willing to serve data for. If empty, the server will
+	// serve data for any shard it can.
+	Shards []core.RelayKey
+
+	// MetadataCacheSize is the size of the metadata cache. Default is 1024 * 1024.
```
dmanc, line 12: Can we add some units to the config descriptions?

dmanc (commented):
initial review, will give it another go tomorrow

relay/blob_manager.go
cody-littley: Nothing catastrophic if there is too much S3 traffic, other than a potential slowdown. After a certain point, adding more goroutines attempting to pull data will result in a net reduction in capacity. Think of this sort of code as a defensive programming in case the node is overloaded with work.

api/proto/relay/relay.proto
cody-littley: This branch is currently up to date with master.

relay/config.go
cody-littley: In this PR, it's just used in unit tests. I've got a follow up PR in the works that adds a main function that utilizes this method.

relay/chunk_manager.go
cody-littley: `GetChunkCoefficients()` fans out into the pool of workers used by the S3 client. It does this since we break the files into tiny fragments for parallel read/write. So even if we have a very large number of parallel requests, the total number of active S3/dynamoDB connections has a firm upper limit controlled by us.

relay/blob_manager.go
```diff
@@ -0,0 +1,96 @@
...
+
+// newBlobManager creates a new blobManager.
+func newBlobManager(
+	ctx context.Context,
+	logger logging.Logger,
+	blobStore *blobstore.BlobStore,
+	blobCacheSize int,
+	workPoolSize int) (*blobManager, error) {
+
+	pool, _ := errgroup.WithContext(ctx)
```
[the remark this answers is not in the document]
cody-littley: Changed construction to `pool := &errgroup.Group{}`

relay/blob_manager.go
```diff
@@ -0,0 +1,96 @@
...
+	return server, nil
+}
+
+// GetBlob retrieves a blob from the blob store.
+func (s *blobManager) GetBlob(blobKey v2.BlobKey) ([]byte, error) {
+
+	// Even though we don't need extra parallelism here, we still use the work pool to ensure that we don't
+	// permit too many concurrent requests to the blob store.
+
+	wg := sync.WaitGroup{}
```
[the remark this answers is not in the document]
cody-littley: We could do this with just a bounded channel, the errgroup just encapsulates that logic for us. This is the implementation of `errgroup.Group.Go()`:

```
func (g *Group) Go(f func() error) {
	if g.sem != nil {
		g.sem <- token{}
	}

	g.wg.Add(1)
	go func() {
		defer g.done()

		if err := f(); err != nil {
			g.errOnce.Do(func() {
				g.err = err
				if g.cancel != nil {
					g.cancel(g.err)
				}
			})
		}
	}()
}
```

relay/chunk_manager.go
cody-littley: Constructor changed. Let me know if you'd like to move away from the errgroup abstraction, and I'll change both locations.

relay/metadata_manager.go
cody-littley: Yes, although it would require me to copy the contents of the map into a regular map before returning it (return type is `type metadataMap map[v2.BlobKey]*blobMetadata`). I'm willing to make this change, but wanted to double check with you in case you think it's cleaner just to use a mutex.

cody-littley: `sync.Map` is non-ideal, since the return type is a normal map and I'd have to copy the data. Figured out a better way to do this without `sync.Map` or a mutex by using a channel.

relay/config.go
cody-littley: Description added.

```
// MetadataCacheSize is the size of the metadata cache. Current cache implementation is unaware of data sizes, and 
	// so this is a total count, not a size in bytes. Default is 1024 * 1024.
	MetadataCacheSize int
```

relay/blob_manager.go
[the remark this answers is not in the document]
ian-shim: If we don't need to block individual requests, errgroup makes sense. But using a bounded channel seems simpler than using both errgroup and a new instance of waitgroup for every request?

relay/chunk_manager.go
ian-shim: Ah makes sense!

relay/blob_manager.go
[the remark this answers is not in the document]
mooselumph: @ian-shim's example seems a lot easier to understand.

api/proto/common/common.proto
```diff
@@ -23,12 +23,3 @@ message PaymentHeader {
   uint32 bin_index = 2;
   bytes cumulative_payment = 3;
 }
-
```
ian-shim, line 26: 🙏

relay/server.go
```diff
@@ -0,0 +1,196 @@
...
+
+// GetBlob retrieves a blob stored by the relay.
+func (s *Server) GetBlob(ctx context.Context, request *pb.GetBlobRequest) (*pb.GetBlobReply, error) {
+
+	// Future work	:
+	//  - global throttle
+	//  - per-connection throttle
+	//  - timeouts
+
+	keys := []v2.BlobKey{v2.BlobKey(request.BlobKey)}
```
ian-shim, line 86: using `corev2.BytesToBlobKey` will also validate the key

relay/server.go
```diff
@@ -0,0 +1,196 @@
...
+	reply := &pb.GetBlobReply{
+		Blob: data,
+	}
+
+	return reply, nil
+}
+
+// GetChunks retrieves chunks from blobs stored by the relay.
+func (s *Server) GetChunks(ctx context.Context, request *pb.GetChunksRequest) (*pb.GetChunksReply, error) {
+
```
ian-shim, line 112: check if `len(request.GetChunkRequests) > 0`?

relay/blob_manager.go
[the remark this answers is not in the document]
ian-shim: hmm don't see it yet. Has it been pushed?

cody-littley:
> hmm don't see it yet. Has it been pushed?

@ian-shim Check again. Forgot to push. 🙃

relay/server.go
cody-littley: added

relay/server.go
cody-littley: Neat, I've added this to all places where an end user can supply a blob key

ian-shim (commented):
Few more comments. LGTM otherwise

ian-shim (approved):
lgtm, please address the remaining comments

relay/blob_manager.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/Layr-Labs/eigensdk-go/logging"
+)
+
+// blobManager encapsulates logic for fetching blobs. Utilized by the relay Server.
+// This struct adds caching and threading on top of blobstore.BlobStore.
+type blobManager struct {
+	ctx    context.Context
+	logger logging.Logger
+
+	// blobStore can be used to read blobs from S3.
```
jianoaix, line 18: "can be used to" -> "is used to"

relay/blob_manager.go
```diff
@@ -0,0 +1,78 @@
...
+	"context"
+	"fmt"
+	"github.com/Layr-Labs/eigenda/core/v2"
+	"github.com/Layr-Labs/eigenda/disperser/common/v2/blobstore"
+	"github.com/Layr-Labs/eigenda/relay/cache"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+)
+
+// blobManager encapsulates logic for fetching blobs. Utilized by the relay Server.
+// This struct adds caching and threading on top of blobstore.BlobStore.
```
jianoaix, line 13: It doesn't seem there is any threading below. It's just request limiting.

relay/blob_manager.go
```diff
@@ -0,0 +1,78 @@
...
+
+// fetchBlob retrieves a single blob from the blob store.
+func (s *blobManager) fetchBlob(blobKey v2.BlobKey) (*[]byte, error) {
+	data, err := s.blobStore.GetBlob(s.ctx, blobKey)
+	if err != nil {
+		s.logger.Error("Failed to fetch blob: %v", err)
+		return nil, err
+	}
+
+	return &data, nil
```
jianoaix, line 77: Is this pointer needed? The slice is already a pointer

relay/blob_manager.go
```diff
@@ -0,0 +1,78 @@
...
+	concurrencyLimiter chan struct{}
+}
+
+// newBlobManager creates a new blobManager.
+func newBlobManager(
+	ctx context.Context,
+	logger logging.Logger,
+	blobStore *blobstore.BlobStore,
+	blobCacheSize int,
+	workPoolSize int) (*blobManager, error) {
```
jianoaix, line 34: It doesn't seem a worker pool, but more like max concurrent blob fetches

api/proto/relay/relay.proto
```diff
@@ -70,7 +70,8 @@ message ChunkRequest {
 }
 
 // The reply to a GetChunks request.
-// Requests are fulfilled in all-or-nothing fashion. If any of the requested chunks are not found, the entire request will fail.
+// Requests are fulfilled in all-or-nothing fashion. If any of the requested chunks are not found, the entire request
```
jianoaix, line 73: This probably better to be placed next to the request or the relevant field in the request message.

relay/chunk_manager.go
```diff
@@ -0,0 +1,174 @@
...
+	frames, err := assembleFrames(coefficients, proofs)
+	if err != nil {
+		return nil, err
+	}
+
+	return &frames, nil
+}
+
+// assembleFrames assembles a slice of frames from its composite proofs and coefficients.
+func assembleFrames(frames []*rs.Frame, proof []*encoding.Proof) ([]*encoding.Frame, error) {
```
jianoaix, line 160: nit: proof -> proofs

relay/chunk_manager.go
```diff
@@ -0,0 +1,174 @@
...
+	"github.com/Layr-Labs/eigenda/relay/chunkstore"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	"sync"
+)
+
+type chunkManager struct {
+	ctx    context.Context
+	logger logging.Logger
+
+	// metadataCache is an LRU cache of blob metadata. Blobs that do not belong to one of the relay shards
```
jianoaix, line 20: What is "relay shard"?

relay/config.go
```diff
@@ -0,0 +1,44 @@
+package relay
+
+import core "github.com/Layr-Labs/eigenda/core/v2"
+
+// Config is the configuration for the relay Server.
+type Config struct {
+
+	// Shards contains the IDs of the relays that this server is willing to serve data for. If empty, the server will
+	// serve data for any shard it can.
+	Shards []core.RelayKey
```
jianoaix, line 10: "contains the IDs of the relays", then shouldn't it be called `relays`? Unclear what is shard here

relay/config.go
```diff
@@ -0,0 +1,44 @@
...
+
+	// MetadataCacheSize is the size of the metadata cache. Current cache implementation is unaware of data sizes, and
```
jianoaix, line 12: This and below: can just say it's the max num of entries/items in the the cache (if there is limit for bytes, it can be named properly as XXSizeBytes)

relay/metadata_manager.go
```diff
@@ -0,0 +1,169 @@
...
+	server.metadataCache = metadataCache
+
+	return server, nil
+}
+
+// metadataMap is a map of blob keys to metadata.
+type metadataMap map[v2.BlobKey]*blobMetadata
+
+// GetMetadataForBlobs retrieves metadata about multiple blobs in parallel.
+func (m *metadataManager) GetMetadataForBlobs(keys []v2.BlobKey) (*metadataMap, error) {
```
jianoaix, line 80: It's probably simpler to use workerpool than handwriting the concurrency limiting for this function

relay/metadata_manager.go
```diff
@@ -0,0 +1,169 @@
...
+	// the size of the blob in bytes
+	blobSizeBytes uint32
+	// the size of the file containing the encoded chunks
+	totalChunkSizeBytes uint32
+	// the fragment size used for uploading the encoded chunks
+	fragmentSizeBytes uint32
+}
+
+// metadataManager encapsulates logic for fetching metadata for blobs. Utilized by the relay Server.
+type metadataManager struct {
```
jianoaix, line 25: Would it be more accurate to name these XXManagers as XXReader or XXFetcher/XXAccessor? It appears no management work needed

relay/chunk_manager.go
```diff
@@ -0,0 +1,174 @@
...
+	server.frameCache = c
+
+	return server, nil
+}
+
+// frameMap is a map of blob keys to frames.
+type frameMap map[v2.BlobKey][]*encoding.Frame
+
+// GetFrames retrieves the frames for a blob.
+func (s *chunkManager) GetFrames(ctx context.Context, mMap *metadataMap) (*frameMap, error) {
```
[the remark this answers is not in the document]
cody-littley: All map pointers converted to non-pointer type

relay/chunk_manager.go
```diff
@@ -0,0 +1,174 @@
...
+
```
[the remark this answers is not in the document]
cody-littley: added check

relay/chunk_manager.go
```diff
@@ -0,0 +1,174 @@
...
+			return nil, fmt.Errorf("error fetching frames for blob %v: %w", result.key, result.err)
+		}
+		fMap[result.key] = result.data
+	}
+
+	return &fMap, nil
+}
+
+// fetchFrames retrieves the frames for a single blob.
+func (s *chunkManager) fetchFrames(key blobKeyWithMetadata) (*[]*encoding.Frame, error) {
```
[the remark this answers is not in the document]
cody-littley: fixed

relay/blob_manager.go
cody-littley: Comment fixed. More accurately, this is not rate limiting, it's instead putting an upper limit on the maximum concurrency for network IO operations. True rate limiting is in a branch that is not yet open for review (will be ready later today, probably).

relay/blob_manager.go
cody-littley: fixed

api/proto/relay/relay.proto
cody-littley: moved

relay/chunk_manager.go
cody-littley: fixed

relay/chunk_manager.go
cody-littley: I've updated this comment to omit the term "relay shard". For interpreting this, you can just interpret `relay shard` to mean `relay ID`.

```
	// metadataCache is an LRU cache of blob metadata. Each relay is authorized to serve data assigned to one or more
	// relay IDs. Blobs that do not belong to one of the relay IDs assigned to this server will not be in the cache.
```

relay/config.go
cody-littley: I've been using the term shard and relay ID interchangeably... will update the naming here to use the term "relay ID"

relay/config.go
cody-littley: updated

relay/metadata_manager.go
cody-littley: Getting conflicting recommendations here. I originally had a worker pool, but changed it at the request of @ian-shim and @mooselumph. I will start a conversation on slack to discuss.

relay/metadata_manager.go
cody-littley: Starting an offline conversation with you on naming, will update naming once we arrive at a mutually acceptable naming schema.

jianoaix (approved):

relay/blob_provider.go
```diff
@@ -0,0 +1,78 @@
...
+func (s *blobProvider) GetBlob(blobKey v2.BlobKey) ([]byte, error) {
+
+	s.concurrencyLimiter <- struct{}{}
+	data, err := s.blobCache.Get(blobKey)
+	<-s.concurrencyLimiter
+
+	if err != nil {
+		// It should not be possible for external users to force an error here since we won't
+		// even call this method if the blob key is invalid (so it's ok to have a noisy log here).
+		s.logger.Error("Failed to fetch blob: %v", err)
```
dmanc, line 62: Should it be errorf?, same comment for fetchBlob

relay/chunk_provider.go
```diff
@@ -0,0 +1,178 @@
...
+	// Channel for results.
+	completionChannel := make(chan *framesResult, len(keys))
+
+	for _, key := range keys {
+
+		boundKey := key
+		go func() {
+			frames, err := s.frameCache.Get(*boundKey)
+			if err != nil {
+				s.logger.Error("Failed to get frames for blob %v: %v", boundKey.blobKey, err)
```
dmanc, line 95: Same comment

dmanc (approved):
left some minor comments but lgtm

relay/blob_provider.go
cody-littley: Good catch, fixed

relay/chunk_provider.go
cody-littley: fixed

PR #870, opened 2024-11-07
[v2] Refactor object serialization

ian-shim:
- Moves all serialization code out of `v2/types.go` and into its own `v2/serialization.go`
- Adds a serialization method for `BlobCertificate`

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


core/v2/types.go
```diff
@@ -413,9 +218,11 @@ type Attestation struct {
 }
 
 type BlobVerificationInfo struct {
-	BlobCertificate *BlobCertificate
-	BlobIndex       uint32
-	InclusionProof  []byte
+	*BatchHeader
```
mooselumph, line 221: We had this in the design doc: 

```
message SignedBatch {
	BatchHeader batch_header = 1;
  Attestation non_signer_stakes_and_signature = 2;
}

message BlobVerificationInfo {
  BlobCertificate blob_certificate = 1;
  uint32 blob_index = 2;
  bytes inclusion_proof = 3;
}
```

Is there an update here?

core/v2/serialization.go
```diff
@@ -0,0 +1,265 @@
...
+func (c *BlobCertificate) Hash() ([32]byte, error) {
+	if c.BlobHeader == nil {
+		return [32]byte{}, fmt.Errorf("blob header is nil")
+	}
+
+	blobCertType, err := abi.NewType("tuple", "", []abi.ArgumentMarshaling{
+		{
+			Name:       "blobHeader",
+			Type:       "tuple",
+			Components: blobHeaderArgMarshaling(),
```
mooselumph, line 179: Wouldn't it make sense to just use the `BlobKey` here?

mooselumph (approved):
Just a few questions

core/v2/types.go
ian-shim: Yeah protobuf interface isn't changed. Just internal structs. 
In fact, I was thinking about removing `BlobHeader` from `BlobVerificationInfo` as the cert can be looked up by the blobKey. What do you think?

core/v2/serialization.go
ian-shim: As long as we stay consistent in contracts, we can do that. Then the hash will be applied to a struct like this instead of actual blob cert: 
```
struct {
  bytes32 blobHeaderHash;
  uint16[] relayKeys;
}
```

core/v2/types.go
mooselumph: >In fact, I was thinking about removing BlobHeader from BlobVerificationInfo as the cert can be looked up by the blobKey. What do you think?
Removing from the internal struct or the interface? It looks like you already removed it from the struct? That seems fine to me, but haven't yet thought through exactly how this is going to be used...

core/v2/serialization.go
mooselumph: It's probably cheaper to do it the way you have it here. Let's leave it for now. I can't think of any situations where this would matter.

core/v2/types.go
ian-shim: Oops. Right I already did lol 
From controllers perspective, this will be used to look up attestation given some blob.

PR #871, opened 2024-11-07
[v2] Dispatcher Implementation

ian-shim:
Dispatcher implementation which takes encoded blobs, forms a batch, distributes the data across DA nodes, then gathers, validates, and aggregates the attestation. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/controller/dispatcher.go
+++ b/disperser/controller/dispatcher.go
@@ -1,0 +1,412 @@
+package controller
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"math"
+	"time"
+
+	"github.com/Layr-Labs/eigenda/api/clients"
+	"github.com/Layr-Labs/eigenda/common"
+	"github.com/Layr-Labs/eigenda/core"
+	corev2 "github.com/Layr-Labs/eigenda/core/v2"
+	v2 "github.com/Layr-Labs/eigenda/disperser/common/v2"
+	"github.com/Layr-Labs/eigenda/disperser/common/v2/blobstore"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	gethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/wealdtech/go-merkletree/v2"
+	"github.com/wealdtech/go-merkletree/v2/keccak256"
+)
+
+var errNoBlobsToDispatch = errors.New("no blobs to dispatch")
+
+type DispatcherConfig struct {
+	PullInterval time.Duration
+
+	FinalizationBlockDelay uint64
+	NodeRequestTimeout     time.Duration
+	NumRequestRetries      int
+}
+
+type Dispatcher struct {
+	DispatcherConfig
+
+	blobMetadataStore *blobstore.BlobMetadataStore
+	pool              common.WorkerPool
+	chainState        core.IndexedChainState
+	aggregator        core.SignatureAggregator
+	nodeClientManager NodeClientManager
+	logger            logging.Logger
+
+	lastUpdatedAt uint64
+}
+
+type batchData struct {
+	Batch           *corev2.Batch
+	BatchHeaderHash [32]byte
+	BlobKeys        []corev2.BlobKey
+	OperatorState   *core.IndexedOperatorState
+}
+
+func NewDispatcher(
+	config DispatcherConfig,
+	blobMetadataStore *blobstore.BlobMetadataStore,
+	pool common.WorkerPool,
+	chainState core.IndexedChainState,
+	aggregator core.SignatureAggregator,
+	nodeClientManager NodeClientManager,
+	logger logging.Logger,
+) (*Dispatcher, error) {
+	return &Dispatcher{
+		DispatcherConfig: config,
+
+		blobMetadataStore: blobMetadataStore,
+		pool:              pool,
+		chainState:        chainState,
+		aggregator:        aggregator,
+		nodeClientManager: nodeClientManager,
+		logger:            logger.With("component", "Dispatcher"),
+
+		lastUpdatedAt: 0,
+	}, nil
+}
+
+func (d *Dispatcher) Start(ctx context.Context) error {
+	go func() {
+		ticker := time.NewTicker(d.PullInterval)
+		defer ticker.Stop()
+		for {
+			select {
+			case <-ctx.Done():
+				return
+			case <-ticker.C:
+				sigChan, batchData, err := d.HandleBatch(ctx)
+				if err != nil {
+					if errors.Is(err, errNoBlobsToDispatch) {
+						d.logger.Warn("no blobs to dispatch")
+					} else {
+						d.logger.Error("failed to process a batch", "err", err)
+					}
+					continue
+				}
+
+				go func() {
+					err := d.HandleSignatures(ctx, batchData, sigChan)
+					if err != nil {
+						d.logger.Error("failed to handle signatures", "err", err)
+					}
+				}()
+			}
+		}
+	}()
+
+	return nil
+
+}
+
+func (d *Dispatcher) HandleBatch(ctx context.Context) (chan core.SigningMessage, *batchData, error) {
+	currentBlockNumber, err := d.chainState.GetCurrentBlockNumber()
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to get current block number: %w", err)
+	}
+	referenceBlockNumber := uint64(currentBlockNumber) - d.FinalizationBlockDelay
+
+	// Get a batch of blobs to dispatch
+	// This also writes a batch header and blob verification info for each blob in metadata store
+	batchData, err := d.NewBatch(ctx, referenceBlockNumber)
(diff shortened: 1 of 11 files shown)


core/v2/serialization.go
```diff
@@ -0,0 +1,265 @@
...
+			},
+			LengthCommitment: abiG2Commit{
+				X: [2]*big.Int{
+					b.BlobCommitments.LengthCommitment.X.A0.BigInt(new(big.Int)),
+					b.BlobCommitments.LengthCommitment.X.A1.BigInt(new(big.Int)),
+				},
+				Y: [2]*big.Int{
+					b.BlobCommitments.LengthCommitment.Y.A0.BigInt(new(big.Int)),
+					b.BlobCommitments.LengthCommitment.Y.A1.BigInt(new(big.Int)),
+				},
```
anupsv, line 121: Should this in thr form of A1, A0 ?

disperser/controller/dispatcher.go
```diff
@@ -0,0 +1,432 @@
...
+			}
+		})
+	}
+
+	return sigChan, batchData, nil
+}
+
+// HandleSignatures receives signatures from operators, validates, and aggregates them
+func (d *Dispatcher) HandleSignatures(ctx context.Context, batchData *batchData, sigChan chan core.SigningMessage) error {
+	quorumAttestation, err := d.aggregator.ReceiveSignatures(ctx, batchData.OperatorState, batchData.BatchHeaderHash, sigChan)
```
anupsv, line 186: Is there any risk of receiving signatures from the same operator twice or more times ? Even if we do, is it a problem ?

disperser/controller/dispatcher.go
```diff
@@ -0,0 +1,432 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to receive and validate signatures: %w", err)
+	}
+	numPassed, passedQuorums := numBlobsAttestedByQuorum(quorumAttestation.QuorumResults, batchData.Batch.BlobCertificates)
```
anupsv, line 190: Is deduplication of certs already handled upstream ?

core/v2/serialization.go
ian-shim: Each extension in G2Point is represented as `uint256[2]` [onchain](https://github.com/Layr-Labs/eigenlayer-middleware/blob/mainnet/src/libraries/BN254.sol#L44)

disperser/controller/dispatcher.go
ian-shim: No. Each request should [result](https://github.com/Layr-Labs/eigenda/pull/871/files/469b084bd08f16502f6037b5260eff7103a9747e#diff-96cf41ed19eafb59625ea2cc3208dd5cc930d68d6f64a278dfd2783a620b98a9R135) in one (or no) response back to this channel

disperser/controller/dispatcher.go
ian-shim: There is no deduplication. New batch should be constructed by blobs that haven't been processed yet. `GetBlobMetadataByStatus` takes `lastUpdatedAt` which gets updated as we pull new blobs and acts as a cursor. It only takes blobs that are updated after this value. 

`numBlobsAttestedByQuorum` is added to 1) early terminate if no blobs received sufficient amount of signatures, or 2) remove unsuccessful quorums (to help reduce gas costs for bridging). However, this optimization seems pretty minor, and I think I'll remove this check to simplify.

disperser/controller/dispatcher.go
```diff
@@ -0,0 +1,432 @@
...
+	certs, _, err := d.blobMetadataStore.GetBlobCertificates(ctx, keys)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get blob certificates: %w", err)
+	}
+
+	if len(certs) != len(keys) {
+		return nil, fmt.Errorf("blob certificates not found for all blob keys")
+	}
+
+	certsMap := make(map[corev2.BlobKey]*corev2.BlobCertificate, len(certs))
```
mooselumph, line 270: Instead of reordering the certs, would it be simpler to just rederive the `keys` array from the `certs` array?

mooselumph: I suppose the current approach forces a fault if there is somehow a redundant cert.

disperser/controller/dispatcher.go
```diff
@@ -0,0 +1,432 @@
...
+		BatchRoot:            [32]byte{},
+		ReferenceBlockNumber: referenceBlockNumber,
+	}
+
+	tree, err := BuildMerkleTree(certs)
+	if err != nil {
+		return nil, fmt.Errorf("failed to build merkle tree: %w", err)
+	}
+
+	copy(batchHeader.BatchRoot[:], tree.Root())
```
mooselumph, line 299: Is it simpler to create the tree first, then populate the `batchHeader` with the root when you create it?

disperser/controller/dispatcher.go
```diff
@@ -0,0 +1,432 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to hash batch header: %w", err)
+	}
+
+	err = d.blobMetadataStore.PutBatchHeader(ctx, batchHeader)
+	if err != nil {
+		return nil, fmt.Errorf("failed to put batch header: %w", err)
+	}
+
+	for i, cert := range certs {
```
mooselumph, line 311: Is there any way to batch Puting the verification info here to keep the latency cost in check?

disperser/controller/dispatcher.go
ian-shim: Yes, that would be simpler but also true that it enforces cert is present for every key. We'd also need to reorder metadatas (ordering is not a requirement though, it's just for sanity)

disperser/controller/dispatcher.go
ian-shim: Unfortunately, batch write operations don't support conditional puts 😭

disperser/controller/dispatcher.go
ian-shim: I think that's just matter of different order of execution vs. necessarily simpler?

disperser/controller/dispatcher.go
mooselumph: Hmm, I'm a little uncertain about whether the conditional check is needed in this case. If we somehow overwrote, it's guaranteed to be the same blob + batch.

ian-shim: Done!

ian-shim:
Now depends on https://github.com/Layr-Labs/eigenda/pull/878

core/v2/types.go
```diff
@@ -239,6 +239,7 @@ type BlobVerificationInfo struct {
 type BlobVersionParameters struct {
 	CodingRate              uint32
 	ReconstructionThreshold float64
+	ConfirmationThreshold   uint8
```
mooselumph, line 242: Remove this?

ian-shim: done

disperser/controller/node_client_manager.go
```diff
@@ -0,0 +1,60 @@
...
+		return nil, fmt.Errorf("failed to create LRU cache: %w", err)
+	}
+
+	return &nodeClientManager{
+		nodeClients: nodeClients,
+		logger:      logger,
+	}, nil
+}
+
+func (m *nodeClientManager) GetClient(operatorID core.OperatorID, socket string) (clients.NodeClientV2, error) {
```
mooselumph, line 40: What happens if there is a socket update? Since this is only using the operatorID as the key, if there is a new socket it won't get reflected. Wouldn't it make more sense to use the socket as the key?

disperser/controller/dispatcher.go
```diff
@@ -0,0 +1,408 @@
...
+
+var errNoBlobsToDispatch = errors.New("no blobs to dispatch")
+
+type DispatcherConfig struct {
+	PullInterval time.Duration
+
+	FinalizationBlockDelay uint64
+	NodeRequestTimeout     time.Duration
+	NumRequestRetries      int
+	NumConnectionsToNodes  int
```
mooselumph, line 30: Is this parameter used?

disperser/controller/node_client_manager.go
```diff
@@ -0,0 +1,62 @@
...
+	}, nil
+}
+
+func (m *nodeClientManager) GetClient(operatorSocket core.OperatorSocket) (clients.NodeClientV2, error) {
+	host, dispersalPort, _, err := core.ParseOperatorSocket(operatorSocket.String())
+	if err != nil {
+		return nil, fmt.Errorf("failed to parse operator socket: %w", err)
+	}
+
+	client, ok := m.nodeClients.Get(dispersalPort)
```
mooselumph, line 47: Why is it just using the `dispersalPort`? Couldn't this be non-unique?

ian-shim: Ah this should just take generic socket

mooselumph (approved):

PR #879, opened 2024-11-09
[v2] Make `ChunkWriter` requests idempotent

ian-shim:
Makes `ChunkWriter` operations idempotent by querying (without downloading) the existing items. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/common/aws/s3/fragment.go
+++ b/common/aws/s3/fragment.go
@@ -3,6 +3,7 @@ package s3
 import (
 	"fmt"
 	"sort"
+	"strconv"
 	"strings"
 )
 
@@ -80,6 +81,23 @@ func GetFragmentKeys(fileKey string, fragmentCount int) ([]string, error) {
 		keys[i] = fragmentKey
 	}
 	return keys, nil
+}
+
+// AllFragmentsExist returns true if all keys are fragment keys.
+// It checks if all fragment keys starting with 0th fragment and ending with nth fragment (marked by "f" postfix) exist.
+// Warning: this function sorts the input slice in place and therefore mutates the ordering of the input slice.
+func SortAndCheckAllFragmentsExist(keys []string) bool {
+	sort.Strings(keys)
+	for i, key := range keys {
+		if !strings.HasSuffix(key, "-"+strconv.Itoa(i)) {
+			if strings.HasSuffix(key, "-"+strconv.Itoa(i)+"f") {
+				return i == len(keys)-1
+			}
+			return false
+		}
+	}
+
+	return false
 }
 
 // recombineFragments recombines fragments into a single file.
--- a/common/aws/s3/fragment.go
+++ b/common/aws/s3/fragment.go
@@ -45,8 +45,8 @@ type Fragment struct {
 	Index       int
 }
 
-// breakIntoFragments breaks a file into fragments of the given size.
-func breakIntoFragments(fileKey string, data []byte, fragmentSize int) ([]*Fragment, error) {
+// BreakIntoFragments breaks a file into fragments of the given size.
+func BreakIntoFragments(fileKey string, data []byte, fragmentSize int) ([]*Fragment, error) {
 	fragmentCount := getFragmentCount(len(data), fragmentSize)
 	fragments := make([]*Fragment, fragmentCount)
 	for i := 0; i < fragmentCount; i++ {
@@ -69,8 +69,8 @@ func breakIntoFragments(fileKey string, data []byte, fragmentSize int) ([]*Fragm
 	return fragments, nil
 }
 
-// getFragmentKeys returns the keys for all fragments of a file.
-func getFragmentKeys(fileKey string, fragmentCount int) ([]string, error) {
+// GetFragmentKeys returns the keys for all fragments of a file.
+func GetFragmentKeys(fileKey string, fragmentCount int) ([]string, error) {
 	keys := make([]string, fragmentCount)
 	for i := 0; i < fragmentCount; i++ {
 		fragmentKey, err := getFragmentKey(fileKey, fragmentCount, i)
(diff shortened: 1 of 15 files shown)


relay/chunkstore/chunk_writer.go
```diff
@@ -48,6 +48,15 @@ func NewChunkWriter(
 }
 
 func (c *chunkWriter) PutChunkProofs(ctx context.Context, blobKey corev2.BlobKey, proofs []*encoding.Proof) error {
```
ian-shim, line 50: Main changes in this file

common/aws/s3/fragment.go
```diff
@@ -45,8 +45,8 @@ type Fragment struct {
 	Index       int
 }
 
-// breakIntoFragments breaks a file into fragments of the given size.
-func breakIntoFragments(fileKey string, data []byte, fragmentSize int) ([]*Fragment, error) {
+// BreakIntoFragments breaks a file into fragments of the given size.
+func BreakIntoFragments(fileKey string, data []byte, fragmentSize int) ([]*Fragment, error) {
```
anupsv, line 49: could we use `uint32` or `uint64` instead of `int` ?

anupsv (approved):

relay/chunkstore/chunk_writer.go
```diff
@@ -85,3 +104,26 @@ func (c *chunkWriter) PutChunkCoefficients(
 		FragmentSizeBytes:   uint32(c.fragmentSize),
 	}, nil
 }
+
+func (c *chunkWriter) proofExists(ctx context.Context, blobKey corev2.BlobKey) bool {
```
dmanc, line 108: Can we make `proofExists` and `coefficientsExists` public so that the encoder could use them to also be idempotent?

relay/chunkstore/chunk_writer.go
```diff
@@ -85,3 +104,26 @@ func (c *chunkWriter) PutChunkCoefficients(
...
+	size, err := c.s3Client.HeadObject(ctx, c.bucketName, s3.ScopedProofKey(blobKey))
+	if err == nil && size != nil && *size > 0 {
+		return true
+	}
+
+	return false
+}
+
+func (c *chunkWriter) coefficientsExists(ctx context.Context, blobKey corev2.BlobKey, dataSize int64) bool {
+	objs, err := c.s3Client.ListObjects(ctx, c.bucketName, s3.ScopedChunkKey(blobKey))
```
mooselumph, line 118: What do you think about adding a log for the latency here so that we can double check the performance?

relay/chunkstore/chunk_writer.go
mooselumph: If the check only comes at the writing stage, then failures in the encoding manager metadata update stage will result in the blob being reencoded...

ian-shim: Ah you're right. Will export these methods

relay/chunkstore/chunk_writer.go
ian-shim: I was planning on adding metrics for all v2 components as a follow up (none of them export any metrics rn). I'll add this as part of those. Does that work?

mooselumph: Yep!

mooselumph (approved):

cody-littley (commented):
I'm a little confused why this is necessary. Uploading the same chunks multiple times has no bad effects other than wasting a bit of bandwidth. If the types of crashes that lead to duplicate uploads are relatively rare, then we should not have to pay this cost very often. Is it worth adding additional latency to each request during the happy case in order to save time in the rare (but still functionally correct) situation?

mooselumph:
> I'm a little confused why this is necessary. Idempotent usually just means that doing the same request twice doesn't have a negative impact. I'd argue that the code is already idempotent in that sense... uploading multiple times has no bad effects other than wasting a bit of bandwidth. If the types of crashes that lead to duplicate uploads are relatively rare, then we should not have to pay this cost very often. Is it worth adding additional latency to each request during the happy case in order to save time in the rare (but still functionally correct) situation?

I think we are worried about wasting bandwidth + encoding resources. The check here doesn't save us from re-encoding. 

Supposing we put the check before doing the encoding, would it make sense to remove the check from the writing phase in order to avoiding adding uneccessary latency? @ian-shim

ian-shim:
@cody-littley @mooselumph That's a good point. I've moved these checks to encoding server from chunk writer.

common/aws/s3/fragment_test.go
```diff
@@ -282,10 +284,50 @@ func TestMissingFinalFragment(t *testing.T) {
...
+	}
+	require.False(t, SortAndCheckAllFragmentsExist(keys))
+
+	keys = []string{ // extra fragment after final fragment
+		"abc-2f",
+		"abc-3",
+		"abc-1",
+		"abc-0",
+	}
+	require.False(t, SortAndCheckAllFragmentsExist(keys))
```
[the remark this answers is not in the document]
cody-littley: You are correct, this should be considered an invalid set of keys.

common/aws/s3/fragment.go
```diff
@@ -82,6 +83,23 @@ func getFragmentKeys(fileKey string, fragmentCount int) ([]string, error) {
...
+		if !strings.HasSuffix(key, "-"+strconv.Itoa(i)) {
+			if strings.HasSuffix(key, "-"+strconv.Itoa(i)+"f") {
+				return i == len(keys)-1
+			}
+			return false
+		}
+	}
+
+	return false
+}
```
cody-littley, line 101: Do you intend for this to fail if there is a missing key in the middle? E.g. `abc-0`, `abc-1`, `--missing--`, `abc-3f`. The function `recombineFragments()` in `fragment.go` performs this sort of check, albeit on the `s3.Fragment` type, not on strings.

cody-littley (approved):

dmanc (approved):

PR #875, opened 2024-11-08
Relay Client

ian-shim:
## Why are these changes needed?
Scaffold implementation of client that talks to a set of relay services. 
A follow up PR will handle gRPC calls when the interface has been finalized and merged 

## Checks

- [ ] I've made sure the lint is passing in this PR.
- [ ] I've made sure the tests are passing. Note that there might be a few flaky tests, in that case, please comment that they are not relevant.
- [ ] I've checked the new test coverage and the coverage percentage didn't drop.
- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests
   - [ ] This PR is not tested :(

PR #880, opened 2024-11-09
[v2] Node `DownloadBatch` method

ian-shim:
Implements a utility method for downloading a batch from relays. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/serialization.go
+++ b/core/serialization.go
@@ -412,8 +412,8 @@ func BatchHeaderFromProtobuf(in *pb.BatchHeader) (*BatchHeader, error) {
 	}, nil
 }
 
-// BlobHeaderFromProto constructs a core.BlobHeader from a proto of pb.BlobHeader.
-func BlobHeaderFromProto(h *pb.BlobHeader) (*BlobHeader, error) {
+// BlobHeaderFromProtobuf constructs a core.BlobHeader from a proto of pb.BlobHeader.
+func BlobHeaderFromProtobuf(h *pb.BlobHeader) (*BlobHeader, error) {
 	if h == nil {
 		return nil, fmt.Errorf("GetBlobHeaderFromProto: blob header is nil")
 
--- a/core/v2/types.go
+++ b/core/v2/types.go
@@ -211,6 +211,77 @@ type Batch struct {
 	BlobCertificates []*BlobCertificate
 }
 
+func (b *Batch) ToProtobuf() (*commonpb.Batch, error) {
+	if b.BatchHeader == nil {
+		return nil, errors.New("batch header is nil")
+	}
+
+	if b.BatchHeader.BatchRoot == [32]byte{} {
+		return nil, errors.New("batch root is empty")
+	}
+
+	if b.BatchHeader.ReferenceBlockNumber == 0 {
+		return nil, errors.New("reference block number is 0")
+	}
+
+	blobCerts := make([]*commonpb.BlobCertificate, len(b.BlobCertificates))
+	for i, cert := range b.BlobCertificates {
+		blobCert, err := cert.ToProtobuf()
+		if err != nil {
+			return nil, fmt.Errorf("failed to convert blob certificate to protobuf: %v", err)
+		}
+		blobCerts[i] = blobCert
+	}
+
+	return &commonpb.Batch{
+		Header: &commonpb.BatchHeader{
+			BatchRoot:            b.BatchHeader.BatchRoot[:],
+			ReferenceBlockNumber: b.BatchHeader.ReferenceBlockNumber,
+		},
+		BlobCertificates: blobCerts,
+	}, nil
+}
+
+func BatchFromProtobuf(proto *commonpb.Batch) (*Batch, error) {
+	if len(proto.GetBlobCertificates()) == 0 {
+		return nil, errors.New("missing blob certificates in batch")
+	}
+
+	if proto.GetHeader() == nil {
+		return nil, errors.New("missing header in batch")
+	}
+
+	if len(proto.GetHeader().GetBatchRoot()) != 32 {
+		return nil, errors.New("batch root must be 32 bytes")
+	}
+
+	batchHeader := &BatchHeader{
+		BatchRoot:            [32]byte(proto.GetHeader().GetBatchRoot()),
+		ReferenceBlockNumber: proto.GetHeader().GetReferenceBlockNumber(),
+	}
+
+	blobCerts := make([]*BlobCertificate, len(proto.GetBlobCertificates()))
+	for i, cert := range proto.GetBlobCertificates() {
+		blobHeader, err := NewBlobHeader(cert.GetBlobHeader())
+		if err != nil {
+			return nil, fmt.Errorf("failed to create blob header: %v", err)
+		}
+
+		blobCerts[i] = &BlobCertificate{
+			BlobHeader: blobHeader,
+			RelayKeys:  make([]RelayKey, len(cert.GetRelays())),
+		}
+		for j, r := range cert.GetRelays() {
+			blobCerts[i].RelayKeys[j] = RelayKey(r)
+		}
+	}
+
+	return &Batch{
+		BatchHeader:      batchHeader,
+		BlobCertificates: blobCerts,
+	}, nil
+}
+
 type Attestation struct {
 	*BatchHeader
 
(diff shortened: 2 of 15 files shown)


core/v2/validator.go
```diff
@@ -16,8 +16,8 @@ var (
 )
 
 type BlobShard struct {
-	BlobCertificate
-	Chunks map[core.QuorumID][]*encoding.Frame
+	*BlobCertificate
+	Chunks map[core.QuorumID]core.Bundle
```
ian-shim, line 20: @mooselumph can you comment if you like this change?

ian-shim:
This PR depends on https://github.com/Layr-Labs/eigenda/pull/875

core/serialization.go
```diff
@@ -393,6 +397,94 @@ func (h *BlobHeader) Deserialize(data []byte) (*BlobHeader, error) {
 	return h, err
 }
 
+// GetBatchHeader constructs a core.BatchHeader from a proto of pb.StoreChunksRequest.
+// Note the StoreChunksRequest is validated as soon as it enters the node gRPC
+// interface, see grpc.Server.validateStoreChunkRequest.
+func BatchHeaderFromProtobuf(in *pb.BatchHeader) (*BatchHeader, error) {
```
jianoaix: nit: `BatchHeaderFromProto` to be consistent (or `BlobHeaderFromProtobuf` below if full spell is preferred)

core/v2/types.go
```diff
@@ -211,6 +211,77 @@ type Batch struct {
 	BlobCertificates []*BlobCertificate
 }
 
+func (b *Batch) ToProtobuf() (*commonpb.Batch, error) {
+	if b.BatchHeader == nil {
+		return nil, fmt.Errorf("batch header is nil")
```
jianoaix, line 216: nit: here and below, `errors.New` since there is not formatting param

core/v2/validator.go
```diff
@@ -16,8 +16,8 @@ var (
 )
 
 type BlobShard struct {
-	BlobCertificate
-	Chunks map[core.QuorumID][]*encoding.Frame
+	*BlobCertificate
+	Bundles map[core.QuorumID]core.Bundle
```
jianoaix, line 20: `Bundles core.Bundles`?

node/node_v2.go
```diff
@@ -0,0 +1,126 @@
...
+	metadata []*requestMetadata
+	bundles  [][]byte
+}
+
+type RawBundles struct {
+	BlobCertificate *corev2.BlobCertificate
+	Bundles         map[core.QuorumID][]byte
+}
+
+func (n *Node) DownloadBatch(ctx context.Context, batch *corev2.Batch) ([]*corev2.BlobShard, []*RawBundles, error) {
```
jianoaix, line 35: It looks odd to supply the batch as input parameter, while the goal is to produce (download) a batch (unless the parameter is also used to bring out the output which isn't the case here)

node/node_v2.go
```diff
@@ -0,0 +1,126 @@
...
+			}
+			bundleChan <- response{
+				metadata: req.metadata,
+				bundles:  bundles,
+			}
+		})
+	}
+	pool.StopWait()
+
+	for i := 0; i < len(requests); i++ {
```
jianoaix, line 113: The pool may send less than `len(requests)` responses to the channel as it exits on error (L103), so this may hang/stuck the server.

node/node_v2.go
```diff
@@ -0,0 +1,126 @@
...
+	"github.com/gammazero/workerpool"
+)
+
+type requestMetadata struct {
+	blobShardIndex int
+	quorum         core.QuorumID
+}
+type relayRequest struct {
+	chunkRequests []*clients.ChunkRequestByRange
+	metadata      []*requestMetadata
```
jianoaix, line 23: This looks not metadata, but an integral part of the request itself (without quorum ID, the request is incomplete). That is, a well defined request should have blobkey, chunk index range and quorum ID all together.

node/grpc/server_v2.go
```diff
@@ -18,23 +18,26 @@ type ServerV2 struct {
...
 	return &ServerV2{
-		config:      config,
-		logger:      logger,
-		node:        node,
-		ratelimiter: ratelimiter,
-		mu:          &sync.Mutex{},
+		config:                config,
+		node:                  node,
+		ratelimiter:           ratelimiter,
+		assignmentCoordinator: &core.StdAssignmentCoordinator{},
```
mooselumph, line 39: Why is the v1 assignment coordinator needed here?

node/node_v2.go
mooselumph, line 35: It seems making robust retry logic for this method could become pretty complex. Are you thinking of doing this as a follow-on/post-MVP?

node/node_v2.go
```diff
@@ -0,0 +1,126 @@
...
+		}
+
+		if len(cert.RelayKeys) == 0 {
+			return nil, nil, fmt.Errorf("no relay keys in the certificate")
+		}
+		blobShards[i] = &corev2.BlobShard{
+			BlobCertificate: cert,
+			Bundles:         make(map[core.QuorumID]core.Bundle),
+		}
+		rawBundles[i] = &RawBundles{
```
mooselumph, line 61: It seems like it could be cleaner to do something like 

```
type BlobShardAndRawBundles struct{
    core.BlobShard
    Bundles: map[core.QuorumID][]byte
}
```

and just return one []BlobShardAndRawBundles array?

mooselumph: Or maybe the []core.BlobShard array is needed by another interface.

node/node_v2.go
```diff
@@ -0,0 +1,126 @@
...
+				requests[relayKey] = req
+			}
+			// Chunks from one blob are requested to the same relay
+			req.chunkRequests = append(req.chunkRequests, &clients.ChunkRequestByRange{
+				BlobKey: blobKey,
+				Start:   assgn.StartIndex,
+				End:     assgn.StartIndex + assgn.NumChunks,
+			})
+			req.metadata = append(req.metadata, &requestMetadata{
+				blobShardIndex: i,
```
mooselumph, line 88: Is there any potential bug with this being bound to the iterator variable?

node/node_v2.go
```diff
@@ -0,0 +1,126 @@
...
+		blobKey, err := cert.BlobHeader.BlobKey()
+		if err != nil {
+			return nil, nil, fmt.Errorf("failed to get blob key: %v", err)
+		}
+
+		if len(cert.RelayKeys) == 0 {
+			return nil, nil, fmt.Errorf("no relay keys in the certificate")
+		}
+		blobShards[i] = &corev2.BlobShard{
+			BlobCertificate: cert,
```
mooselumph, line 58: Any potential bug with this being bound to iterator variable?

node/grpc/server_v2.go
ian-shim: removed

node/node_v2.go
ian-shim: This is metadata for keeping track of the requests, i.e. the actual request to relay doesn't require quorum IDs, just blob key and indices. This metadata is used on the client side (node) to put the responses in the right place.

node/node_v2.go
ian-shim: Agree with the confusion. I'll rename this method to `DownloadBundles`. Is that more clear?

ian-shim: I was thinking about not retrying at all (at least for the MVP) and returning error immediately if download fails. What do you think?

node/node_v2.go
ian-shim: Not sure what you mean by this being bound to iterator variable?

node/node_v2.go
ian-shim: It's useful to keep the return type `[]*corev2.BlobShard` and `[]*RawBundles` (separate slices for each type of bundles) because you want the raw bundles for storage and blob shards for validation. If combined, we'd end up reconstructing these slices downstream

node/node_v2.go
ian-shim: Thanks for catching this. Fixed!

core/v2/validator.go
mooselumph: Looks fine to me!

mooselumph (approved):

PR #890, opened 2024-11-13
Fix Subgraph README.md typo

gwenall:

There is a typo int a shell example were there are 4 hyphens instead of 2 on the graph init parameters.
This commit is fixing this typo

## Checks
This is a Readme typo. 
The command has been tested locally.


jianoaix (approved):

jianoaix:
> This looks reasonable to me. My guess is they wanted `-- --from-contract ...`, but not sure if the `--` separation is needed for graph init.

`from-contract` is a normal argument from graph, should not need the `--` separation

PR #892, opened 2024-11-13
Relay main

cody-littley:

This PR adds a main function and GRPC server for the relay.

Not yet ready for review, needs https://github.com/Layr-Labs/eigenda/pull/869 to merge first.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


go.mod
cody-littley, line 1: These changes were the result of installing viper via `go get github.com/spf13/viper`

https://github.com/spf13/viper

relay/config.go
```diff
@@ -1,14 +1,61 @@
...
+const (
+	// relayEnvPrefix is the prefix for all environment variables used by the relay.
+	relayEnvPrefix = "RELAY"
+)
 
 // Config is the configuration for the relay Server.
+//
+// Environment variables are mapped into this struct by taking the name of the field in this struct,
+// converting to upper case, and prepending "RELAY_". For example, "BlobCacheSize" can be set using the
+// environment variable "RELAY_BLOBCACHESIZE".
```
jianoaix, line 22: Is it a constraint of viper? This is slightly different format than what we currently use, which is underscore connected, e.g. `MAX_BLOB_SIZE`.

relay/config.go
```diff
@@ -1,14 +1,61 @@
...
+	AWS aws.ClientConfig
+
 	// RelayIDs contains the IDs of the relays that this server is willing to serve data for. If empty, the server will
 	// serve data for any shard it can.
 	RelayIDs []core.RelayKey
 
+	// GRPCPort is the port that the relay server listens on. Default is 50051. // TODO what is a good port?
+	GRPCPort int
+
+	// BucketName is the name of the S3 bucket that stores blobs. Default is "relay".
```
jianoaix, line 49: Are chunks and blobs in different buckets? If so, might be better to make the names more specific

jianoaix (approved):

ian-shim (approved):

relay/config.go
cody-littley: No, they are in the same bucket. We discussed this as a group when you were on PTO, and came to the conclusion that it was simpler to use a single bucket. We use [this code](https://github.com/Layr-Labs/eigenda/blob/master/common/aws/s3/scoped_keys.go) to ensure we don't have collisions between different data types.

relay/config.go
cody-littley: There is a way, but lets leave that to a follow up PR (as discussed offline).

PR #893, opened 2024-11-14
Node v2 `StoreChunks` and `GetChunks` endpoints

ian-shim:
Implementation of v2 `StoreChunks` and `GetChunks` endpoints

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/node/node_v2.go
+++ b/node/node_v2.go
@@ -33,14 +33,9 @@ type RawBundles struct {
 	Bundles         map[core.QuorumID][]byte
 }
 
-func (n *Node) DownloadBundles(ctx context.Context, batch *corev2.Batch) ([]*corev2.BlobShard, []*RawBundles, error) {
+func (n *Node) DownloadBundles(ctx context.Context, batch *corev2.Batch, operatorState *core.OperatorState) ([]*corev2.BlobShard, []*RawBundles, error) {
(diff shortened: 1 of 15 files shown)


node/node_v2.go
```diff
@@ -134,3 +131,26 @@ func (n *Node) DownloadBundles(ctx context.Context, batch *corev2.Batch) ([]*cor
 
 	return blobShards, rawBundles, nil
 }
+
+func (n *Node) ValidateBatchV2(
```
mooselumph, line 135: Should this just go in the core/validator object?

mooselumph (approved):
Looks good. Idempotency as follow-up?

node/node_v2.go
ian-shim: I moved batch header validation as another method in validator object, but kept them separated

ian-shim:
> Looks good. Idempotency as follow-up?

For now, I added [a check](https://github.com/Layr-Labs/eigenda/pull/893/files#diff-e0709c067d80ae092e5eb445765c32358b1161585f60e22f065f7e01e9fa90a3R59-R61) if the batch header has been stored and abort the entire store operation if it has been stored. 
We can add more sophisticated logic later but this would be on par with the v1 logic.

PR #895, opened 2024-11-14
GPU accelerated encoder

dmanc:

This PR adds support for multiple backends for the encoder.  Initially we only have the `gnark` and `icicle` backends, see https://dev.ingonyama.com/icicle/overview for more information on the icicle backend. 

A backend may include GPU acceleration as an option, see https://dev.ingonyama.com/icicle/install_cuda_backend for icicle's GPU accelerated backend.

In order to use the icicle backend it must be compiled with the `icicle` build tag:

```
go build -tags=icicle main.go
```

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


disperser/cmd/encoder/config.go
```diff
@@ -28,7 +28,7 @@ type Config struct {
 	EncoderConfig    kzg.KzgConfig
 	LoggerConfig     common.LoggerConfig
 	ServerConfig     *encoder.ServerConfig
-	MetricsConfig    encoder.MetrisConfig
+	MetricsConfig    encoder.MetricsConfig
```
samlaf, line 31: is there any reason why ServerConfig is a pointer but MetricsConfig is not? Is ServerConfig being shared with another service? If so, add a comment. Otherwise, I'd prob make all of them use same pattern, just for consistency.

samlaf (commented):
Yikes lost all my comments while reviewing on github.dev.... friggin github. Some graphql internal error or something. So I'll just submit this first set of reviews in case. Don't really feel like reviewing again lol... another reason why small PRs are gud, big PRs bad.

samlaf (commented):
Just spent 2 hours reviewing and I barely even got to the important stuff. Suggest no more PRs like this, for the sake of our industry, and of your reviewers' mental health. :D

encoding/icicle/device_setup.go
```diff
@@ -0,0 +1,123 @@
...
+)
+
+// IcicleDevice wraps the core device setup and configurations
+type IcicleDevice struct {
+	Device         icicle_runtime.Device
+	NttCfg         core.NTTConfig[[icicle_bn254.SCALAR_LIMBS]uint32]
+	MsmCfg         core.MSMConfig
+	FlatFFTPointsT []icicle_bn254.Affine
+	SRSG1Icicle    []icicle_bn254.Affine
+}
```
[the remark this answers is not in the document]
dmanc: No there's only going to be a single configuration loaded at startup time. Right now we don't support multiple GPUs for an encoder.

encoding/icicle/device_setup.go
```diff
@@ -0,0 +1,123 @@
...
+
+// IcicleDeviceConfig holds configuration options for device setup
+type IcicleDeviceConfig struct {
+	EnableGPU bool
+	NTTSize   uint8
+	// MSM setup parameters (optional)
```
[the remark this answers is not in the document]
dmanc: Reed-solomon encoder only uses NTT so doesn't need the MSM configuration

encoding/icicle/msm_setup.go
```diff
@@ -0,0 +1,40 @@
...
+package icicle
+
+import (
+	"github.com/consensys/gnark-crypto/ecc/bn254"
+	"github.com/ingonyama-zk/icicle/v3/wrappers/golang/core"
+	icicle_bn254 "github.com/ingonyama-zk/icicle/v3/wrappers/golang/curves/bn254"
+	"github.com/ingonyama-zk/icicle/v3/wrappers/golang/runtime"
+)
+
+func SetupMsm(rowsG1 [][]bn254.G1Affine, srsG1 []bn254.G1Affine) ([]icicle_bn254.Affine, []icicle_bn254.Affine, core.MSMConfig, core.MSMConfig, runtime.EIcicleError) {
```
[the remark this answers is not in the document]
dmanc: G2 MSM is not used right now but if we wanted to accelerate the length proof commitment then we would need it. Going to remove it from the configuration for now.

encoding/icicle/ntt_setup.go
```diff
@@ -0,0 +1,47 @@
...
+
+	"github.com/consensys/gnark-crypto/ecc/bn254/fr/fft"
+	"github.com/ingonyama-zk/icicle/v3/wrappers/golang/core"
+	"github.com/ingonyama-zk/icicle/v3/wrappers/golang/curves/bn254"
+	"github.com/ingonyama-zk/icicle/v3/wrappers/golang/curves/bn254/ntt"
+	"github.com/ingonyama-zk/icicle/v3/wrappers/golang/runtime"
+)
+
+// batchSize is number of batches
+func SetupNTT(maxScale uint8) (core.NTTConfig[[bn254.SCALAR_LIMBS]uint32], runtime.EIcicleError) {
```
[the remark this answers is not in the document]
dmanc: Right now we don't support batching on the encoding side but pretty much the idea is we could do multiple blobs in a batch on the GPU. Think @bxue-l2 measured it's only 2x better

encoding/bench/Makefile
```diff
@@ -1,9 +1,15 @@
...
 
-benchmark_cpu:
+build_icicle:
+	go build -tags=icicle -gcflags="all=-N -l" -ldflags="-s=false -w=false" -o bin/main_icicle main.go
+
+benchmark_default:
 	go run main.go -cpuprofile cpu.prof -memprofile mem.prof
 
+benchmark_icicle:
+	go run -tags=icicle main.go -cpuprofile cpu.prof -memprofile mem.prof
```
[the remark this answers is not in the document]
dmanc: I just am not sure how to use it. Also the benchmarking code here was used to compare the default CPU implementation vs icicle GPU implementation in terms of speed-up so needed the raw timings.

docker-bake.hcl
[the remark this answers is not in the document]
dmanc: I'm not sure what tool can lint it I just gave it to Claude.

disperser/cmd/encoder/flags/flags.go
```diff
@@ -67,6 +68,19 @@ var (
 		Required: false,
 		EnvVar:   common.PrefixEnvVar(envVarPrefix, "ENABLE_GNARK_CHUNK_ENCODING"),
 	}
+	EnableGPUFlag = cli.BoolFlag{
```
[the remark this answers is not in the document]
dmanc: Oh that's pretty good actually. Maybe we should create a style guide of our own that we enforce somehow.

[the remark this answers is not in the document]
dmanc: I'll make the change for the new stuff I'm adding but keeping the rest the same for the purpose of backwards compatability.

core/test/core_test.go
```diff
@@ -53,12 +53,20 @@ func makeTestComponents() (encoding.Prover, encoding.Verifier, error) {
...
 	if err != nil {
 		return nil, nil, err
 	}
 
-	v, err := verifier.NewVerifier(config, true)
+	vopts := []verifier.VerifierOption{
+		verifier.WithKZGConfig(config),
+		verifier.WithLoadG2Points(true),
+	}
+	v, err := verifier.NewVerifier(vopts...)
```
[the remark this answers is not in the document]
dmanc: Not relevant to this PR so gonna just leave it in a ticket.

encoding/icicle/device_setup.go
```diff
@@ -0,0 +1,123 @@
...
+
+	// Setup NTT and optionally MSM on device
+	icicle_runtime.RunOnDevice(&device, func(args ...any) {
+		defer wg.Done()
+
+		// Setup NTT
+		nttCfg, icicleErr = SetupNTT(config.NTTSize)
+		if icicleErr != icicle_runtime.Success {
+			setupErr = fmt.Errorf("could not setup NTT")
+			return
```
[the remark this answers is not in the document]
dmanc: cannot return because I don't think that RunOnDevice func will let me. Looks like this:

```
func RunOnDevice(device *Device, funcToRun func(args ...any), args ...any) {
    go func(deviceToRunOn *Device) {
        defer runtime.UnlockOSThread()
        runtime.LockOSThread()
        originalDevice, _ := GetActiveDevice()
        SetDevice(deviceToRunOn)
        funcToRun(args...)
        SetDevice(originalDevice)
    }(device)
}
```

encoding/rs/serialization.go
```diff
@@ -0,0 +1,43 @@
+package rs
+
+import (
+	"errors"
+
+	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
+)
+
+const BYTES_PER_SYMBOL = 32 // fr.Element serializes to 32 bytes
```
jianoaix, line 9: We can reuse already defined constant: https://github.com/Layr-Labs/eigenda/blob/82802dbcf1cf929781c6249ab476f4f03769a532/encoding/constants.go#L9

encoding/rs/noicicle.go
```diff
@@ -0,0 +1,15 @@
...
+import (
+	"fmt"
+
+	"github.com/Layr-Labs/eigenda/encoding"
+	"github.com/Layr-Labs/eigenda/encoding/fft"
+)
+
+func CreateIcicleBackendEncoder(p *Encoder, params encoding.EncodingParams, fs *fft.FFTSettings) (*ParametrizedEncoder, error) {
+	// Not supported
+	return nil, fmt.Errorf("icicle backend called without icicle build tag")
```
jianoaix, line 14: May use `errors.New("...")` since this has no formatting needs

encoding/rs/serialization.go
```diff
@@ -0,0 +1,43 @@
...
+
+// Serialize serializes Frame to a compact byte format
+// Each fr.Element is serialized consecutively
+func Serialize(f *Frame) ([]byte, error) {
+	// Pre-allocate buffer for all coefficients
+	coded := make([]byte, 0, BYTES_PER_SYMBOL*len(f.Coeffs))
+
+	// Append each coefficient's bytes
+	for _, coeff := range f.Coeffs {
```
jianoaix, line 18: What about the Frame.Proof? Without that, it's a lossy serialization.

jianoaix (commented):
Can we have a diff-neutral PR just for refactoring? It seems there are code here which doesn't need deep review as they are just existing code moved around?

encoding/rs/serialization.go
dmanc: Sorry meant to delete this file, it's not relevant

encoding/rs/serialization.go
dmanc: same comment, file not needed.

dmanc:
> Can we have a diff-neutral PR just for refactoring? It seems there are code here which doesn't need deep review as they are just existing code moved around?

Yes, if we're happy with this new configuration method then I'll make another PR with that and rebase this one

disperser/cmd/encoder/main.go
```diff
@@ -63,9 +65,31 @@ func RunEncoderServer(ctx *cli.Context) error {
...
 		// We no longer compute the commitments in the encoder, so we don't need to load the G2 points
-		prover, err := prover.NewProver(&config.EncoderConfig, false)
+		popts := []prover.ProverOption{
+			prover.WithKZGConfig(&config.EncoderConfig),
+			prover.WithLoadG2Points(false),
+			prover.WithBackend(backendType),
+			prover.WithGPU(config.ServerConfig.GPUEnable),
+			prover.WithRSEncoder(rsEncoder),
+		}
+		prover, err := prover.NewProver(popts...)
```
dmanc, line 92: Trying to follow options pattern with this change https://golang.cafe/blog/golang-functional-options-pattern.html

encoding/kzg/prover/noicicle.go
```diff
@@ -0,0 +1,16 @@
...
+	"fmt"
+
+	"github.com/Layr-Labs/eigenda/encoding"
+	"github.com/Layr-Labs/eigenda/encoding/fft"
+	"github.com/Layr-Labs/eigenda/encoding/kzg"
+)
+
+func CreateIcicleBackendProver(p *Prover, params encoding.EncodingParams, fs *fft.FFTSettings, ks *kzg.KZGSettings) (*ParametrizedProver, error) {
+	// Not supported
+	return nil, fmt.Errorf("icicle backend called without icicle build tag")
```
dmanc, line 15: The point of noicicle.go file is that the icicle backend should only be functional when the icicle build tag is used. This is so we don't need to include the icicle backend files in the traditional encoder backend, otherwise it may break the existing build.

encoding/kzg/prover/proof_backend.go
```diff
@@ -0,0 +1,18 @@
...
+	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
+)
+
+// Proof device represents a backend capable of computing KZG multiproofs.
+type KzgMultiProofsBackend interface {
+	ComputeMultiFrameProof(blobFr []fr.Element, numChunks, chunkLen, numWorker uint64) ([]bn254.G1Affine, error)
+}
+
+// CommitmentDevice represents a backend capable of computing various KZG commitments.
+type KzgCommitmentsBackend interface {
```
dmanc, line 14: Moved commitments to its own backend in case we would like to accelerate them in the future.

encoding/kzg/prover/prover.go
```diff
@@ -331,3 +457,99 @@ func toUint64Array(chunkIndices []encoding.ChunkNumber) []uint64 {
 	}
 	return res
 }
+
+func (p *Prover) newProver(params encoding.EncodingParams) (*ParametrizedProver, error) {
```
dmanc, line 461: this file should be reviewed carefully

encoding/kzg/verifier/verifier.go
```diff
@@ -18,10 +19,16 @@ import (
 	"github.com/consensys/gnark-crypto/ecc"
 	"github.com/consensys/gnark-crypto/ecc/bn254"
 	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
+	_ "go.uber.org/automaxprocs"
 )
 
+// VerifierOption defines a function that configures a Verifier
+type VerifierOption func(*Verifier) error
```
dmanc, line 26: This can go in it's own PR, it's applying the same options pattern refactor to verifier so not super relevant to GPU encoder

encoding/rs/parametrized_encoder.go
```diff
@@ -0,0 +1,125 @@
...
+import (
+	"fmt"
+
+	"github.com/Layr-Labs/eigenda/encoding"
+	"github.com/Layr-Labs/eigenda/encoding/fft"
+	rb "github.com/Layr-Labs/eigenda/encoding/utils/reverseBits"
+	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
+)
+
+type ParametrizedEncoder struct {
```
dmanc, line 12: Refactor to make the encoder parametrized based on FFT settings (similar to Prover). Don't think it's too valuable since max time is ~600ms for loading largest fft setting. Am thinking of reverting

encoding/icicle/utils.go
```diff
@@ -0,0 +1,103 @@
+//go:build icicle
+
+package icicle
```
dmanc, line 3: This folder is icicle backend related setup

encoding/kzg/prover/icicle/ecntt.go
```diff
@@ -0,0 +1,46 @@
+//go:build icicle
+
+package icicle
```
dmanc, line 3: This folder is icicle related computations for MultiProofs

encoding/rs/icicle/extend_poly.go
```diff
@@ -0,0 +1,40 @@
...
+
+	"github.com/Layr-Labs/eigenda/encoding/icicle"
+	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
+	"github.com/ingonyama-zk/icicle/v3/wrappers/golang/core"
+	icicle_bn254 "github.com/ingonyama-zk/icicle/v3/wrappers/golang/curves/bn254"
+	"github.com/ingonyama-zk/icicle/v3/wrappers/golang/curves/bn254/ntt"
+	"github.com/ingonyama-zk/icicle/v3/wrappers/golang/runtime"
+)
+
+type RsIcicleComputeDevice struct {
```
dmanc, line 16: This is icicle related computations for RS Encode

dmanc:
One thing I'm not too sure how to deal with is testing Icicle/ GPU encoder correctness. The only strategy I can think of is to compare output of default backend and icicle backend.

Also if we wanted it to be part of CI, then we need to acquire GPU based github action runners.

encoding/rs/icicle.go
```diff
@@ -0,0 +1,35 @@
+//go:build icicle
+
+package rs
+
+import (
+	"github.com/Layr-Labs/eigenda/encoding"
+	"github.com/Layr-Labs/eigenda/encoding/fft"
+	"github.com/Layr-Labs/eigenda/encoding/icicle"
+	rs_icicle "github.com/Layr-Labs/eigenda/encoding/rs/icicle"
```
jianoaix, line 9: It usually doesn't use snake case for alias

encoding/rs/icicle.go
```diff
@@ -0,0 +1,35 @@
...
+)
+
+const (
+	defaultNTTSize = 25 // Used for NTT setup in Icicle backend
+)
+
+func CreateIcicleBackendEncoder(e *Encoder, params encoding.EncodingParams, fs *fft.FFTSettings) (*ParametrizedEncoder, error) {
+	icicleDevice, err := icicle.NewIcicleDevice(icicle.IcicleDeviceConfig{
+		GPUEnable: e.Config.GPUEnable,
```
jianoaix, line 18: It looks Icicle will support also CPU now?

encoding/rs/icicle/extend_poly.go
```diff
@@ -0,0 +1,40 @@
...
+}
+
+// Encoding Reed Solomon using FFT
+func (g *RsIcicleComputeDevice) ExtendPolyEval(coeffs []fr.Element) ([]fr.Element, error) {
+	wg := sync.WaitGroup{}
+	wg.Add(1)
+
+	var evals []fr.Element
+	g.NttCfg.BatchSize = int32(1)
+	runtime.RunOnDevice(&g.Device, func(args ...any) {
```
jianoaix, line 28: If the GPU is enabled, then this func will be executed on GPU by the runtime?

encoding/kzg/verifier/verifier.go
jianoaix: It looks more verbose and is deviating from the convention, what're the main problems it solves here compared to passing in a config struct?

encoding/rs/encoder.go
```diff
@@ -44,21 +107,42 @@ type RsComputeDevice interface {
...
+	if e.Config.GPUEnable {
+		return nil, fmt.Errorf("GPU is not supported in default backend")
+	}
 
-	return &Encoder{
-		EncodingParams: params,
-		Fs:             fs,
-		verbose:        verbose,
-		NumRSWorker:    runtime.GOMAXPROCS(0),
+	return &ParametrizedEncoder{
```
jianoaix, line 138: So when you say "backend", it looks a software/library that computes RS encoding? I.e. it decoupled device (CPU/GPU) and backend (the software that executes compute on device)

encoding/rs/extend_poly.go
```diff
@@ -1,19 +1,16 @@
...
+package rs
 
 import (
-	"github.com/Layr-Labs/eigenda/encoding"
 	"github.com/Layr-Labs/eigenda/encoding/fft"
 	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
 )
 
-type RsCpuComputeDevice struct {
+type RsDefaultComputeDevice struct {
```
jianoaix, line 8: Can "default" be made specific? Like RsGnarkComputeDevice (suppose it's based on that)

encoding/kzg/verifier/verifier.go
dmanc: One argument I would give for this configuration pattern is that it provides a simpler constructor that is extensible and backwards compatible. 

For example, with a config struct we would have to modify every client that calls the constructor when we add a new configuration value vs. this pattern you add a `WithX` function in the library that sets the field and does whatever validation it needs to. Then clients can optionally update to use that new added configuration.

encoding/rs/icicle/extend_poly.go
dmanc: yes because the device variable will be set to the GPU device here https://github.com/Layr-Labs/eigenda/blob/dfe247d701aea8685f265d34b2f553bfec8863ff/encoding/icicle/device_setup.go#L88

encoding/rs/encoder.go
dmanc: Right now I consider backend as library that implements the cryptographic primitive like NTT, MSM. We have our own implementation of NTT borrowed from protolambda and MSM from gnark. 

It could be plausible that there exists libraries that implement the higher level primitives like multiproofs, commitment, RS encode and we can try to make our code more of a frontend that calls those. But for now we focus on the core primitives that give the speedup factor.

encoding/rs/extend_poly.go
dmanc: i think if we replaced protolambdas FFT implementation that we are borrowing with the gnark FFT then it would make sense to be called gnark ebcause then the FFT and MSM are both from the gnark library.

encoding/rs/icicle/extend_poly.go
jianoaix: For performance, it looks the main data movement to/from GPU will be the coefficients and the resulting evaluations?

encoding/rs/extend_poly.go
jianoaix: "default" is not quite informative for what this does. Also if icicle becomes the primary backend, this as the default doesn't look well.

The argument here seems that it's not purely Gnark but a mixed use of libraries. If we plan to move away from protoblambda's FFT, it may be fine to preliminarily call it Gnark? But open to other good way to name it.

encoding/kzg/verifier/verifier.go
jianoaix: >with a config struct we would have to modify every client that calls the constructor when we add a new configuration value

Not sure about this, if that new field is optional, the caller doesn't need to do anything. If it's required, callers will have to make changes in either case.

My concerns are the boilplate and the scalabilty of keeping adding WithXX, we may have a long list of fields (like in some big structs).
Also it looks it still need to operate on config struct, if the validation runs across multiple fields (eg. invariants defined over multiple fields).

encoding/rs/extend_poly.go
dmanc: I'm happy with calling it Gnark backend

encoding/kzg/verifier/verifier.go
dmanc: > My concerns are the boilplate and the scalabilty of keeping adding WithXX, we may have a long list of fields (like in some big structs).

These seem tractable problem though, here is an example in grpc library:
https://github.com/grpc/grpc-go/blob/66385b28b3fe71a4895f00d581ede0a344743a3f/dialoptions.go#L248

encoding/rs/icicle.go
dmanc: Yes but it's not as performant as the gnark backend, am trying to look into why

api/clients/retrieval_client_test.go
```diff
@@ -37,12 +37,20 @@ func makeTestComponents() (encoding.Prover, encoding.Verifier, error) {
 		NumWorker:       uint64(runtime.GOMAXPROCS(0)),
 	}
 
-	p, err := prover.NewProver(config, true)
+	opts := []prover.ProverOption{
+		prover.WithKZGConfig(config),
+		prover.WithLoadG2Points(true),
+	}
+	p, err := prover.NewProver(opts...)
```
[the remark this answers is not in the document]
dmanc: done

encoding/bench/Makefile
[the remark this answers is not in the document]
samlaf: Feel like go benchmark should be able to give you that, but also not super familiar, and don't want to add more stuff to this PR, so resolving.

encoding/icicle/device_setup.go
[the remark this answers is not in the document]
samlaf: ah my bad I completely missed the function signature.
I see so the code passed to RunOnDevice is run on the gpu...? Did you test that your shared pointer for setupErr is working then? This would mean the gpu is able to go write in the cpu's RAM. Does this only work on mac where cpu/gpu share memory? Or is this something that always works?

encoding/kzg/verifier/verifier.go
samlaf: Think I agree with both of you. I've grown fond of the options pattern and like it best, but I still think consistency across a codebase is the most important thing, so kind of agree with jian about this.

encoding/icicle/device_setup.go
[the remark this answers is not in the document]
dmanc: No, RunOnDevice locks the OS thread so it makes all of that code run on the same OS thread. It's because with go routines you can call `SetDevice` to set the GPU device and that can run on thread 1 but the go routine can end up switching to thread 2 in the middle which means that it will use the default device (CPU).

encoding/kzg/verifier/verifier.go
dmanc: No problem, I'll revert to the previous method

disperser/cmd/encoder/main.go
dmanc: reverted

encoding/icicle/device_setup.go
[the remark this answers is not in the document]
dmanc: added more info in the comment above

encoding/kzg/prover/prover.go
```diff
@@ -335,3 +349,98 @@ func toUint64Array(chunkIndices []encoding.ChunkNumber) []uint64 {
...
+}
+
+// Helper methods for setup
+func (p *Prover) SetupFFTPoints(params encoding.EncodingParams) ([][]bn254.G1Affine, [][]bn254.G1Affine, error) {
+	subTable, err := NewSRSTable(p.KzgConfig.CacheDir, p.Srs.G1, p.KzgConfig.NumWorker)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to create SRS table: %w", err)
+	}
+
+	fftPoints, err := subTable.GetSubTables(params.NumChunks, params.ChunkLength)
```
bxue-l2, line 432: I don't think we ever need fftPoints, chance for optimization

encoding/rs/parametrized_encoder.go
```diff
@@ -0,0 +1,125 @@
...
+
+	numWorker := uint64(g.Config.NumWorker)
+	if numWorker > g.NumChunks {
+		numWorker = g.NumChunks
+	}
+
+	jobChan := make(chan JobRequest, numWorker)
+	results := make(chan error, numWorker)
+
+	for w := uint64(0); w < numWorker; w++ {
```
bxue-l2, line 61: try workerpool?

encoding/kzg/prover/prover.go
dmanc: Will include this in followup optimizations

encoding/kzg/prover/commitments.go
```diff
@@ -0,0 +1,49 @@
...
+	"github.com/consensys/gnark-crypto/ecc/bn254"
+	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
+)
+
+type KzgCommitmentsGnarkBackend struct {
+	KzgConfig  *kzg.KzgConfig
+	Srs        *kzg.SRS
+	G2Trailing []bn254.G2Affine
+}
+
```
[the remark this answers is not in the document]
bxue-l2: should have been 
e(g1_shifted, ComputeLengthCommitment) = e(g1, ComputeLengthProof)
e(ComputeCommitment, g2) = e(g1, ComputeLengthCommitment)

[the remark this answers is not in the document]
dmanc: I decided to do documentation update in a separate PR

encoding/kzg/prover/icicle.go
```diff
@@ -0,0 +1,66 @@
...
+
+	"github.com/Layr-Labs/eigenda/encoding"
+	"github.com/Layr-Labs/eigenda/encoding/fft"
+	"github.com/Layr-Labs/eigenda/encoding/icicle"
+	"github.com/Layr-Labs/eigenda/encoding/kzg"
+	icicleprover "github.com/Layr-Labs/eigenda/encoding/kzg/prover/icicle"
+)
+
+const (
+	MAX_NTT_SIZE = 25
```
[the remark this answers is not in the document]
dmanc: added comment

encoding/rs/parametrized_encoder.go
dmanc: Will include in followup. Wrote ticket

disperser/cmd/apiserver/main.go
```diff
@@ -162,7 +162,8 @@ func RunDisperserServer(ctx *cli.Context) error {
 	bucketName := config.BlobstoreConfig.BucketName
 	logger.Info("Blob store", "bucket", bucketName)
 	if config.DisperserVersion == V2 {
-		prover, err := prover.NewProver(&config.EncodingConfig, true)
+		config.EncodingConfig.LoadG2Points = true
+		prover, err := prover.NewProver(&config.EncodingConfig, nil)
```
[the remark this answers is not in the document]
dmanc: I think it's acceptable

encoding/kzg/prover/icicle.go
```diff
@@ -0,0 +1,70 @@
...
+		SRSIcicle:      icicleDevice.SRSG1Icicle,
+		SFs:            sfs,
+		Srs:            p.Srs,
+		NttCfg:         icicleDevice.NttCfg,
+		MsmCfg:         icicleDevice.MsmCfg,
+		KzgConfig:      p.KzgConfig,
+		Device:         icicleDevice.Device,
+	}
+
+	// Set up default commitments backend
```
jianoaix, line 55: Re "default": don't we only have gnark backend for commitment anyway?

encoding/kzg/prover/icicle/ecntt.go
```diff
@@ -0,0 +1,46 @@
...
+}
+
+func (c *KzgMultiProofIcicleBackend) ECNttOnDevice(batchPoints core.DeviceSlice, isInverse bool, totalSize int) (core.DeviceSlice, error) {
+	var p iciclebn254.Projective
+	var out core.DeviceSlice
+
+	output, err := out.Malloc(p.Size(), totalSize)
+
+	if err != runtime.Success {
+		return out, fmt.Errorf("%v", "Allocating bytes on device for Projective results failed")
```
jianoaix, line 30: here and below: use errors.New() should be sufficient if there is no formatting.

Also start message with lower case

encoding/kzg/prover/icicle/msm.go
```diff
@@ -0,0 +1,34 @@
...
+func (c *KzgMultiProofIcicleBackend) MsmBatchOnDevice(rowsFrIcicleCopy core.DeviceSlice, rowsG1Icicle []iciclebn254.Affine, totalSize int) (core.DeviceSlice, error) {
+	rowsG1IcicleCopy := core.HostSliceFromElements[iciclebn254.Affine](rowsG1Icicle)
+
+	var p iciclebn254.Projective
+	var out core.DeviceSlice
+
+	_, err := out.Malloc(p.Size(), totalSize)
+
+	if err != runtime.Success {
+		return out, fmt.Errorf("%v", "Allocating bytes on device for Projective results failed")
```
jianoaix, line 25: ditto for error message

encoding/kzg/prover/multiframe_proof.go
```diff
@@ -13,66 +14,18 @@ import (
 	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
 )
 
-type KzgCpuProofDevice struct {
+type KzgMultiProofGnarkBackend struct {
```
jianoaix, line 17: Shall we create a `gnark` folder similar to `icicle` to organize these implementations with gnark?

encoding/kzg/prover/icicle/multiframe_proof.go
```diff
@@ -0,0 +1,212 @@
...
+	runtime.RunOnDevice(&p.Device, func(args ...any) {
+		defer wg.Done()
+
+		// Copy the flatten coeff to device
+		var flattenStoreCopyToDevice core.DeviceSlice
+		flattenCoeffStoreCopy.CopyToDevice(&flattenStoreCopyToDevice, true)
+
+		sumVec, err := p.MsmBatchOnDevice(flattenStoreCopyToDevice, p.FlatFFTPointsT, int(numPoly)*int(dimE)*2)
+		if err != nil {
+			icicleErr = fmt.Errorf("msm error: %w", err)
```
jianoaix, line 104: here and below: why not return on such an error? The proof computing is already failed at these points

encoding/kzg/prover/noicicle.go
jianoaix, line 1: How is this used?

encoding/kzg/prover/prover.go
```diff
@@ -336,3 +350,98 @@ func toUint64Array(chunkIndices []encoding.ChunkNumber) []uint64 {
...
+		return p.createIcicleBackendProver(params, fs, ks)
+	default:
+		return nil, fmt.Errorf("unsupported backend type: %v", p.Config.BackendType)
+	}
+
+}
+
+func (p *Prover) createGnarkBackendProver(params encoding.EncodingParams, fs *fft.FFTSettings, ks *kzg.KZGSettings) (*ParametrizedProver, error) {
+	if p.Config.GPUEnable {
+		return nil, fmt.Errorf("GPU is not supported in default backend")
```
jianoaix, line 385: nit: "default" -> "gnark" (some occurrences below as well)

encoding/kzg/verifier/batch_commit_equivalence_test.go
```diff
@@ -14,17 +14,21 @@ import (
 )
 
 func TestBatchEquivalence(t *testing.T) {
```
jianoaix, line 16: It'll be nice to have a icicle-gnark equivalence test for proofs, but may need gpu setup for unit test... Maybe we can start with CPU (for both), in a follow up

encoding/kzg/pointsIO.go
```diff
@@ -121,12 +119,6 @@ func ReadG1Points(filepath string, n uint64, numWorker uint64) ([]bn254.G1Affine
 		return nil, err
 	}
 
-	// measure reading time
-	t := time.Now()
-	elapsed := t.Sub(startTimer)
-	log.Printf("    Reading G1 points (%v bytes) takes %v\n", (n * G1PointBytes), elapsed)
```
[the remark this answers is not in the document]
dmanc: Will revert, was doing some refactors on logging but can do that separately.

disperser/cmd/encoder/icicle.Dockerfile
```diff
@@ -0,0 +1,62 @@
...
+    cp -r ./icicle/include/icicle/ /usr/local/include/ && \
+    tar xzf /tmp/icicle-cuda.tar.gz -C /opt && \
+    rm /tmp/icicle.tar.gz /tmp/icicle-cuda.tar.gz
+
+# Build the server with icicle backend
+WORKDIR /app/disperser
+RUN go build -tags=icicle -o ./bin/server ./cmd/encoder
+
+# Start a new stage for the base image
+FROM nvidia/cuda:12.2.2-base-ubuntu22.04
```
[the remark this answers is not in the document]
dmanc: Didn't look into but my guess right now would be the cuda base [image](https://hub.docker.com/r/nvidia/cuda) which contains the cuda runtime.

encoding/kzg/prover/noicicle.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/895#discussion_r1844364488

Essentially just makes it so we don't have to compile icicle libraries unless we explicitly add the build tag.

encoding/rs/icicle/extend_poly.go
```diff
@@ -0,0 +1,65 @@
...
+type RsIcicleBackend struct {
+	NttCfg  core.NTTConfig[[iciclebn254.SCALAR_LIMBS]uint32]
+	Device  icicleRuntime.Device
+	GpuLock sync.Mutex
+}
+
+// Encoding Reed Solomon using FFT
+func (g *RsIcicleBackend) ExtendPolyEval(coeffs []fr.Element) ([]fr.Element, error) {
+	// Lock the GPU for operations
+	g.GpuLock.Lock()
```
jianoaix, line 26: If it's not using GPU (which is possible for icicle), then wouldn't this lock hurt performance?

jianoaix (commented):
Mostly lg. There were a couple comments before would be nice to address

encoding/kzg/prover/proof_backend.go
```diff
@@ -0,0 +1,19 @@
+package prover
+
+import (
+	"github.com/consensys/gnark-crypto/ecc/bn254"
+	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
+)
+
+// Proof device represents a backend capable of computing KZG multiproofs.
+type KzgMultiProofsBackend interface {
+	ComputeMultiFrameProof(blobFr []fr.Element, numChunks, chunkLen, numWorker uint64) ([]bn254.G1Affine, error)
```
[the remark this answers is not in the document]
dmanc: blobFr are the coefficients of the polynomial. We actually have some documentation about it here https://github.com/Layr-Labs/eigenda/blob/master/docs/spec/attestation/amortized-proving.md#amortized-multireveal-proof-generation-with-the-fft.  

These two resources were good for understanding the implementation details (although I still don't fully grasp it yet):
https://alinush.github.io/2021/06/17/Feist-Khovratovich-technique-for-computing-KZG-proofs-fast.html
https://alinush.github.io/2020/03/19/multiplying-a-vector-by-a-toeplitz-matrix.html

encoding/rs/icicle/extend_poly.go
dmanc: The perf is already quite poor on the CPU version of icicle. I would prefer to tackle that separately if that's okay.

jianoaix (approved):
LGTM, well done!

PR #899, opened 2024-11-15
[v2] Disperser server `GetBlobCommitment` endpoint

ian-shim:
Loads a prover with G2 points and computes commitment on `GetBlobCommitment` request

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/apiserver/server_v2.go
+++ b/disperser/apiserver/server_v2.go
@@ -8,6 +8,7 @@ import (
 	"time"
 
 	"github.com/Layr-Labs/eigenda/api"
+	pbcommon "github.com/Layr-Labs/eigenda/api/grpc/common"
 	pb "github.com/Layr-Labs/eigenda/api/grpc/disperser/v2"
 	"github.com/Layr-Labs/eigenda/common"
 	healthcheck "github.com/Layr-Labs/eigenda/common/healthcheck"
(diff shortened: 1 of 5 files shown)


disperser/cmd/apiserver/main.go
```diff
@@ -161,6 +162,10 @@ func RunDisperserServer(ctx *cli.Context) error {
 	bucketName := config.BlobstoreConfig.BucketName
 	logger.Info("Blob store", "bucket", bucketName)
 	if config.DisperserVersion == V2 {
+		prover, err := prover.NewProver(&config.EncodingConfig, true)
```
dmanc, line 165: The prover setup takes around ~20 seconds with 16 cpus since we need to load the G2 points. Also is the disperser still a singleton in V2?

disperser/apiserver/server_v2.go
```diff
@@ -149,7 +154,36 @@ func (s *DispersalServerV2) GetBlobStatus(ctx context.Context, req *pb.BlobStatu
 }
 
 func (s *DispersalServerV2) GetBlobCommitment(ctx context.Context, req *pb.BlobCommitmentRequest) (*pb.BlobCommitmentReply, error) {
-	return &pb.BlobCommitmentReply{}, api.NewErrorUnimplemented()
+	if s.prover == nil {
+		return nil, api.NewErrorUnimplemented()
+	}
+	if len(req.GetData()) == 0 {
```
dmanc, line 160: Should we have bounds check on the largest supported blob size?

ian-shim: good call

disperser/cmd/apiserver/main.go
ian-shim: I don't think it has to be a singleton anymore now that payments aren't handled locally. @mooselumph is that correct?

cody-littley (approved):

PR #900, opened 2024-11-15
Disperser Client v2

ian-shim:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/disperser_client_v2.go
+++ b/api/clients/disperser_client_v2.go
@@ -1,0 +1,239 @@
+package clients
+
+import (
+	"context"
+	"fmt"
+	"math/big"
+	"sync"
+
+	"github.com/Layr-Labs/eigenda/api"
+	disperser_rpc "github.com/Layr-Labs/eigenda/api/grpc/disperser/v2"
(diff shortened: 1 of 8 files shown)


api/clients/disperser_client_v2.go
```diff
@@ -0,0 +1,234 @@
...
+		return nil, api.NewErrorInternal(err.Error())
+	}
+
+	request := &disperser_rpc.BlobStatusRequest{
+		BlobKey: blobKey[:],
+	}
+	return c.client.GetBlobStatus(ctx, request)
+}
+
+func (c *disperserClientV2) GetBlobCommitment(ctx context.Context, data []byte) (*disperser_rpc.BlobCommitmentReply, error) {
```
cody-littley, line 203: Since this will be more on the public facing side of things, wondering if it's worth it to write godocs.

cody-littley (approved):

api/clients/disperser_client_v2.go
ian-shim: Good call. Done.

PR #904, opened 2024-11-15
[relay] Replace viper to common cli parsing framework

ian-shim:
We will probably want to use viper for handling configs, but for now, it's easier to compose flags with existing framework. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


jianoaix:
Does using Viper for now cause issue (hence the replacement here)?

ian-shim:
> Does using Viper for now cause issue (hence the replacement here)?

Yeah forgot to update makefile

relay/cmd/config.go
```diff
@@ -0,0 +1,103 @@
...
+// 	viper.SetEnvKeyReplacer(strings.NewReplacer(`.`, `_`))
+// 	viper.AutomaticEnv()
+
+// 	err := viper.Unmarshal(config)
+// 	if err != nil {
+// 		return nil, fmt.Errorf("fatal error unmarshaling configuration: %s \n", err)
+// 	}
+
+// 	return config, nil
+// }
```
cody-littley, line 103: this can be removed

relay/cmd/config.go
```diff
@@ -0,0 +1,103 @@
...
+func NewConfig(ctx *cli.Context) (Config, error) {
+	loggerConfig, err := common.ReadLoggerCLIConfig(ctx, flags.FlagPrefix)
+	if err != nil {
+		return Config{}, err
+	}
+	awsClientConfig := aws.ReadClientConfig(ctx, flags.FlagPrefix)
+	relayIDs := ctx.IntSlice(flags.RelayIDsFlag.Name)
+	if len(relayIDs) == 0 {
+		return Config{}, fmt.Errorf("no relay IDs specified")
+	}
```
cody-littley, line 57: Having zero relay IDs is interpreted as the node permitting service for all relay IDs. Do you think it's worth allowing this feature?

relay/cmd/main.go
```diff
@@ -54,4 +80,6 @@ func main() {
 	if err != nil {
 		panic(fmt.Sprintf("fatal error starting relay server: %s", err))
 	}
+
+	return nil
```
cody-littley, line 84: If this function now returns an error, should we convert the `panic()` operations in this function into errors?

cody-littley (approved):
For the sake of moving this effort forward I'm willing to accept this change without further discussion. I still think we are paying a high ongoing cost for the way we currently manage our configuration, though.

relay/cmd/config.go
ian-shim: I feel like there is a chance relayIDs are left empty unintentionally then it will serve for all IDs inadvertently. If the relay is to serve for all IDs, the IDs should be explicitly configured. Wdyt?

relay/cmd/config.go
ian-shim: Left it for the time we convert it back to using viper, but agree it should be removed for now

jianoaix:
Sounds fine to keep using existing config handling, we have a good number of moving pieces now

PR #906, opened 2024-11-18
Relay rate limits

cody-littley:

Adds server side rate limiting to the relay.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


relay/config.go
```diff
@@ -76,23 +77,32 @@ type Config struct {
 	// ChunkMaxConcurrency is the size of the work pool for fetching chunks. Default is 32. Note that this does not
 	// impact concurrency utilized by the s3 client to upload/download fragmented files.
 	ChunkMaxConcurrency int
+
+	// MaxKeysPerGetChunksRequest is the maximum number of keys that can be requested in a single GetChunks request.
+	// Default is 1024. // TODO should this be the max batch size? What is that?
+	MaxKeysPerGetChunksRequest int
```
cody-littley, line 83: @ian-shim what is the maximum number of keys we should permit in a single request?

relay/relay_test_utils.go
```diff
@@ -177,7 +177,7 @@ func buildChunkStore(t *testing.T, logger logging.Logger) (chunkstore.ChunkReade
 
 func randomBlob(t *testing.T) (*v2.BlobHeader, []byte) {
 
-	data := tu.RandomBytes(128)
+	data := tu.RandomBytes(225) // TODO talk to Ian about this
```
cody-littley, line 180: @ian-shim a blob size of 128 bytes results in 5 commitment length of 5, which the method to determine the length of a chunk says is illegal because it's not a power of 2.

relay/limiter/config.go
```diff
@@ -0,0 +1,88 @@
...
+	// within a short time window. Default is 2MiB.
+	GetChunkBytesBurstinessClient int
+
+	// MaxConcurrentGetChunkOpsClient is the maximum number of concurrent GetChunk operations that are permitted.
+	// Default is 1.
+	MaxConcurrentGetChunkOpsClient int
+}
+
+// DefaultConfig returns a default rate limit configuration.
+func DefaultConfig() *Config {
```
ian-shim, line 68: same with comment above. For now, can we create the flags and set the default values there?

relay/metadata_provider.go
```diff
@@ -153,8 +155,15 @@ func (m *metadataProvider) fetchMetadata(key v2.BlobKey) (*blobMetadata, error)
 		}
 	}
 
+	blobSize := uint32(cert.BlobHeader.BlobCommitments.Length)
```
ian-shim, line 158: `cert.BlobHeader.BlobCommitments.Length` gives you the blob length in symbols. 
I think what we want is here `cert.BlobHeader.BlobSize` (size of raw blob data in bytes) instead

relay/server.go
```diff
@@ -136,14 +158,58 @@ func (s *Server) GetChunks(ctx context.Context, request *pb.GetChunksRequest) (*
...
 
 	if len(request.ChunkRequests) <= 0 {
 		return nil, fmt.Errorf("no chunk requests provided")
 	}
+	if len(request.ChunkRequests) > s.config.MaxKeysPerGetChunksRequest {
+		return nil, fmt.Errorf(
+			"too many chunk requests provided, max is %d", s.config.MaxKeysPerGetChunksRequest)
+	}
+
+	// Future work: client IDs will be fixed when authentication is implemented
```
ian-shim, line 171: nit: `// TODO(cody-littley): ...`

relay/server.go
```diff
@@ -136,14 +158,58 @@ func (s *Server) GetChunks(ctx context.Context, request *pb.GetChunksRequest) (*
...
+	clientID := fmt.Sprintf("%d", request.RequesterId)
```
ian-shim, line 172: do we allow unauthenticated traffic or all requests need to be signed (for MVP)?

relay/server.go
```diff
@@ -226,14 +288,35 @@ func (s *Server) GetChunks(ctx context.Context, request *pb.GetChunksRequest) (*
...
+// computeChunkRequestRequiredBandwidth computes the bandwidth required to fulfill a GetChunks request.
+func computeChunkRequestRequiredBandwidth(request *pb.GetChunksRequest, mMap metadataMap) int {
+	requiredBandwidth := 0
+	for _, req := range request.ChunkRequests {
+		var metadata *blobMetadata
+		var requestedChunks int
+
+		if req.GetByIndex() != nil {
+			key := v2.BlobKey(req.GetByIndex().GetBlobKey())
+			metadata = mMap[key]
```
ian-shim, line 303: check if the key exists in the map?

relay/server.go
```diff
@@ -226,14 +288,35 @@ func (s *Server) GetChunks(ctx context.Context, request *pb.GetChunksRequest) (*
...
+			requestedChunks = len(req.GetByIndex().ChunkIndices)
+		} else {
+			key := v2.BlobKey(req.GetByRange().GetBlobKey())
+			metadata = mMap[key]
+			requestedChunks = int(req.GetByRange().EndIndex - req.GetByRange().StartIndex)
+		}
+
+		requiredBandwidth += requestedChunks * int(metadata.chunkSizeBytes)
```
ian-shim, line 311: check if `metadata` is nil?

relay/metadata_provider.go
```diff
@@ -153,8 +155,15 @@ func (m *metadataProvider) fetchMetadata(key v2.BlobKey) (*blobMetadata, error)
...
+	chunkSize, err := v2.GetChunkLength(cert.BlobHeader.BlobVersion, blobSize)
```
ian-shim, line 159: Similarly, I think this returns the chunk length in number of symbols (not bytes). 
You would have to multiply this by `encoding.BYTES_PER_SYMBOL`

relay/limiter/blob_rate_limiter.go
```diff
@@ -0,0 +1,93 @@
...
+		return nil
+	}
+
+	countInFlight := l.operationsInFlight.Add(1)
+	if countInFlight > int64(l.config.MaxConcurrentGetBlobOps) {
+		l.operationsInFlight.Add(-1)
+		return fmt.Errorf("global concurrent request limit exceeded for getBlob operations, try again later")
+	}
+
+	allowed := l.opLimiter.AllowN(now, 1)
```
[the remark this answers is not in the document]
cody-littley: The only difference is that the time is supplied from the outside context. This allows me to write unit tests that aren't wall clock dependent.

relay/config.go
```diff
@@ -76,23 +77,32 @@ type Config struct {
...
+
+	// RateLimits contains configuration for rate limiting.
+	RateLimits limiter.Config
 }
 
 // DefaultConfig returns the default configuration for the relay Server.
 func DefaultConfig() *Config {
```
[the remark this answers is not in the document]
cody-littley: moved

relay/server.go
```diff
@@ -86,23 +97,29 @@ func NewServer(
...
 // GetBlob retrieves a blob stored by the relay.
 func (s *Server) GetBlob(ctx context.Context, request *pb.GetBlobRequest) (*pb.GetBlobReply, error) {
 
-	// Future work	:
-	//  - global throttle
-	//  - per-connection throttle
+	// Future work:
 	//  - timeouts
 
+	err := s.blobRateLimiter.BeginGetBlobOperation(time.Now())
```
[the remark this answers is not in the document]
cody-littley: Yes, this was missing. Passed unit tests because the default concurrency limit is higher than the maximum number of operations executed by any single unit test on the server. Fixed.

relay/limiter/config.go
cody-littley: moved

relay/limiter/blob_rate_limiter.go
```diff
@@ -0,0 +1,93 @@
...
+	if l == nil {
+		// If the rate limiter is nil, do not enforce rate limits.
+		return nil
+	}
+
+	countInFlight := l.operationsInFlight.Add(1)
+	if countInFlight > int64(l.config.MaxConcurrentGetBlobOps) {
+		l.operationsInFlight.Add(-1)
+		return fmt.Errorf("global concurrent request limit exceeded for getBlob operations, try again later")
+	}
```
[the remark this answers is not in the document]
cody-littley: The situation where this logic may fail is if the request rate is so high that there are always  `MaxConcurrentGetBlobOps` or more goroutines that happen to be inside this one block of code, which could in theory prevent any actual operations from being served.

Although such a scenario is unlikely, probably better to err on the side of safety. This now uses a mutex to guarantee atomicity.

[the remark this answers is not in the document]
cody-littley: Playing around with this a little, the code gets really ugy when I try and to low resolution locking. Methods in this class (as well as `ChunkRateLimiter`) are just doing basic arithmetic... let's just use a single mutext to protect the entire class for the sake of simplicity.

relay/server.go
cody-littley: done

relay/server.go
cody-littley: My plan was to only permit authenticated traffic for `GetChunks()`. Is there any use case for allowing unauthenticated traffic? I can add it if you think it would be helpful.

relay/server.go
cody-littley: Should not be possible, since process will fail before this point if there is missing metadata. But it's a useful safety check, so I don't mind adding it.

relay/server.go
cody-littley: added

relay/metadata_provider.go
cody-littley: done, thanks

relay/metadata_provider.go
cody-littley: There is currently no such field. Is that something that is going to be added?

relay/limiter/chunk_rate_limiter.go
```diff
@@ -0,0 +1,151 @@
...
+			rate.Limit(l.config.MaxGetChunkOpsPerSecondClient),
+			l.config.GetChunkOpsBurstinessClient)
+
+		l.perClientBandwidthLimiter[requesterID] = rate.NewLimiter(
+			rate.Limit(l.config.MaxGetChunkBytesPerSecondClient),
+			l.config.GetChunkBytesBurstinessClient)
+	}
+
+	if l.globalOperationsInFlight >= l.config.MaxConcurrentGetChunkOps {
+		return fmt.Errorf("global concurrent request limit exceeded for GetChunks operations, try again later")
```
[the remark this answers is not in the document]
cody-littley: added to this error message and all similar error messages

relay/limiter/chunk_rate_limiter.go
```diff
@@ -0,0 +1,151 @@
...
+	}
+
+	// no lock needed here, as the bandwidth limiters themselves are thread-safe
+
+	allowed := l.globalBandwidthLimiter.AllowN(now, bytes)
+	if !allowed {
+		return fmt.Errorf("global rate limit exceeded for GetChunk bandwidth, try again later")
+	}
+
+	allowed = l.perClientBandwidthLimiter[requesterID].AllowN(now, bytes)
```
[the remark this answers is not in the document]
cody-littley: added sanity check

PR #908, opened 2024-11-18
[v2] node/relay bug fixes

ian-shim:
A number of bug fixes in relay and v2 node.
Also, it adds a feature flag for enabling v2 features on the node. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/relay_client.go
+++ b/api/clients/relay_client.go
@@ -45,11 +45,15 @@ type RelayClient interface {
 type relayClient struct {
 	config *RelayClientConfig
 
-	initOnce map[corev2.RelayKey]*sync.Once
-	conns    map[corev2.RelayKey]*grpc.ClientConn
-	logger   logging.Logger
+	// initOnce is used to ensure that the connection to each relay is initialized only once.
+	// It maps relay key to a sync.Once instance: `map[corev2.RelayKey]*sync.Once`
+	initOnce *sync.Map
+	// conns maps relay key to the gRPC connection: `map[corev2.RelayKey]*grpc.ClientConn`
+	conns  sync.Map
+	logger logging.Logger
 
-	grpcClients map[corev2.RelayKey]relaygrpc.RelayClient
+	// grpcClients maps relay key to the gRPC client: `map[corev2.RelayKey]relaygrpc.RelayClient`
+	grpcClients sync.Map
 }
 
 var _ RelayClient = (*relayClient)(nil)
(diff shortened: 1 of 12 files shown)


api/clients/relay_client.go
```diff
@@ -45,49 +45,44 @@ type RelayClient interface {
...
-	grpcClients map[corev2.RelayKey]relaygrpc.RelayClient
+	// initOnce is used to ensure that the connection to each relay is initialized only once.
+	// It maps relay key to a sync.Once instance: `map[corev2.RelayKey]*sync.Once`
+	initOnce *sync.Map
+	// conns maps relay key to the gRPC connection: `map[corev2.RelayKey]*grpc.ClientConn`
+	conns  sync.Map
+	logger logging.Logger
+
+	// grpcClients maps relay key to the gRPC client: `map[corev2.RelayKey]relaygrpc.RelayClient`
+	grpcClients sync.Map
```
ian-shim, line 56: All maps need to be thread safe

relay/metadata_provider.go
```diff
@@ -119,8 +127,7 @@ func (m *metadataProvider) GetMetadataForBlobs(keys []v2.BlobKey) (metadataMap,
 		}()
 	}
 
-	mMap := make(metadataMap)
-	for len(mMap) < len(keys) {
```
ian-shim, line 123: This hangs if there are duplicates in `keys`

jianoaix: Add a unit test for this

node/grpc/server_v2_test.go
```diff
@@ -84,6 +84,23 @@ func TestV2NodeInfoRequest(t *testing.T) {
 	assert.True(t, err == nil)
 }
 
+func TestV2ServerWithoutV2(t *testing.T) {
+	config := makeConfig(t)
+	config.EnableV2 = false
+	c := newTestComponents(t, config)
+	_, err := c.server.StoreChunks(context.Background(), &pbv2.StoreChunksRequest{})
+	require.Error(t, err)
```
jianoaix, line 92: These block of code may be benefited from a small helper function (parse error and assert the error status), as they repeat many times in this file

api/clients/relay_client.go
```diff
@@ -45,49 +45,44 @@ type RelayClient interface {
...
 }
 
 var _ RelayClient = (*relayClient)(nil)
 
 // NewRelayClient creates a new RelayClient that connects to the relays specified in the config.
 // It keeps a connection to each relay and reuses it for subsequent requests, and the connection is lazily instantiated.
```
jianoaix, line 62: Why lazy init? It's not that expensive to create a connection once per Node creation

node/grpc/server_test.go
```diff
@@ -83,6 +83,7 @@ func makeConfig(t *testing.T) *node.Config {
 		DbPath:                    t.TempDir(),
 		ID:                        opID,
 		NumBatchValidators:        runtime.GOMAXPROCS(0),
+		EnableV2:                  true,
```
jianoaix, line 86: This removes the test coverage for the existing v1, we need to keep testing it

api/clients/relay_client.go
ian-shim: Just preventing a case that creates a large number of connections. This client is not necessarily only used by the node as it's meant to be used as a library

node/grpc/server_test.go
ian-shim: Sounds good. Defaulted it to false and overwrote it to true in `server_v2_test`

node/grpc/server_v2_test.go
ian-shim: Done

relay/metadata_provider.go
ian-shim: Done

cody-littley (approved):

PR #911, opened 2024-11-19
Relay authentication

cody-littley:

Add authentication for GetChunks() requests. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proto/relay/relay.proto
```diff
@@ -34,7 +34,7 @@ message GetChunksRequest {
 
   // If this is an authenticated request, this should hold the ID of the requester. If this
   // is an unauthenticated request, this field should be empty.
-  uint64 requester_id = 2;
+  bytes requester_id = 2;
```
jianoaix, line 37: From the code below, this is interpreted as operator ID. It'll be helpful to document it here so the users know how to  set it.

relay/auth/authenticator.go
```diff
@@ -0,0 +1,159 @@
...
+		authenticationTimeoutDuration: authenticationTimeoutDuration,
+	}
+}
+
+func (a *requestAuthenticator) AuthenticateGetChunksRequest(
+	address string,
+	request *pb.GetChunksRequest,
+	now time.Time) error {
+
+	if a == nil {
```
jianoaix, line 68: How does this compare to check the authenticator from outside?

relay/auth/authenticator.go
```diff
@@ -0,0 +1,159 @@
...
+	}
+	operators, err := a.ics.GetIndexedOperators(context.Background(), blockNumber)
+	if err != nil {
+		return fmt.Errorf("failed to get operators: %w", err)
+	}
+
+	operatorID := core.OperatorID(request.RequesterId)
+	operator, ok := operators[operatorID]
+	if !ok {
+		return errors.New("operator not found")
```
jianoaix, line 90: It'll be helpful to include the block number in error message

relay/auth/authenticator.go
```diff
@@ -0,0 +1,159 @@
...
+	if a.isAuthenticationStillValid(now, address) {
+		// We've recently authenticated this client. Do not authenticate again for a while.
+		return nil
+	}
+
+	blockNumber, err := a.ics.GetCurrentBlockNumber()
+	if err != nil {
+		return fmt.Errorf("failed to get current block number: %w", err)
+	}
+	operators, err := a.ics.GetIndexedOperators(context.Background(), blockNumber)
```
jianoaix, line 82: There are 2 RPC calls here, it'd be helpful to cache the operator state

relay/auth/authenticator.go
```diff
@@ -0,0 +1,159 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to get operators: %w", err)
+	}
+
+	operatorID := core.OperatorID(request.RequesterId)
+	operator, ok := operators[operatorID]
+	if !ok {
+		return errors.New("operator not found")
+	}
+	key := operator.PubkeyG2
```
jianoaix, line 92: I'd cache `operatorID->G2`, this is a mapping that'll not change (hence highly valuable to cache)

api/proto/relay/relay.proto
```diff
@@ -34,7 +34,7 @@ message GetChunksRequest {
...
 
   // If this is an authenticated request, this field will hold a signature by the requester
   // on the chunks being requested.
```
jianoaix, line 40: Document how this signature should be computed

cody-littley:
```
  // If this is an authenticated request, this field will hold a BLS signature by the requester
  // on the hash of this request. Relays may choose to reject unauthenticated requests.
  //
  // The following describes the schema for computing the hash of this request
  // This algorithm is implemented in golang using relay.auth.HashGetChunksRequest().
  //
  // All integers are encoded as unsigned 4 byte big endian values.
  //
  // Perform a keccak256 hash on the following data in the following order:
  // 1. the operator id
  // 2. for each chunk request:
  //    a. if the chunk request is a request by index:
  //       i. the blob key
  //       ii. the start index
  //       iii. the end index
  //    b. if the chunk request is a request by range:
  //       i. the blob key
  //       ii. each requested chunk index, in order
  bytes operator_signature = 3;
```

relay/auth/authenticator.go
cody-littley: caching added

relay/auth/authenticator.go
cody-littley: Done. This will break if we ever allow operators to change keys, but that's a bridge we can cross in the future.

api/proto/relay/relay.proto
cody-littley: I've renamed this field to `operator_id`. Is that sufficient, or do you think I should add additional documentation?

relay/auth/request_signing.go
```diff
@@ -0,0 +1,50 @@
...
+			binary.BigEndian.PutUint32(endBytes, getByRange.EndIndex)
+			hasher.Write(endBytes)
+		}
+	}
+
+	return hasher.Sum(nil)
+}
+
+// SignGetChunksRequest signs the given GetChunksRequest with the given private key.
+func SignGetChunksRequest(keys *core.KeyPair, request *pb.GetChunksRequest) {
```
ian-shim, line 46: In general, I think methods should not modify input parameters in place unless it's strictly necessary. 
Can this method return the signature and caller place the return value in the request?

relay/cmd/main.go
```diff
@@ -83,3 +96,42 @@ func RunRelay(ctx *cli.Context) error {
...
+		logger.Error("Cannot create chain.Client", "err", err)
+		return nil, err
+	}
+
+	tx, err := coreeth.NewWriter(logger, client, config.BLSOperatorStateRetrieverAddr, config.EigenDAServiceManagerAddr)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create eth writer: %w", err)
+	}
+
+	idx, err := coreindexer.CreateNewIndexer(
```
ian-shim, line 117: Is this using a built-in indexer? Should we use a graph indexer instead?

relay/server.go
```diff
@@ -200,9 +219,22 @@ func (s *Server) GetChunks(ctx context.Context, request *pb.GetChunksRequest) (*
 			"too many chunk requests provided, max is %d", s.config.MaxKeysPerGetChunksRequest)
 	}
 
-	// Future work: client IDs will be fixed when authentication is implemented
-	clientID := fmt.Sprintf("%d", request.RequesterId)
+	client, ok := peer.FromContext(ctx)
+	if !ok {
+		return nil, errors.New("could not get peer information")
+	}
+	clientAddress := client.Addr.String()
```
ian-shim, line 226: should we place this block inside the conditional (L228)?

ian-shim (commented):
lgtm! few comments

relay/auth/authenticator.go
```diff
@@ -0,0 +1,174 @@
...
+	// authenticationTimeoutDuration is the duration for which an auth is valid.
+	// If this is zero, then auth saving is disabled, and each request will be authenticated independently.
+	authenticationTimeoutDuration time.Duration
+
+	// savedAuthLock is used for thread safe atomic modification of the authenticatedClients map and the
+	// authenticationTimeouts queue.
+	savedAuthLock sync.Mutex
+
+	// keyCache is used to cache the public keys of operators. Operator keys are assumed to never change.
+	keyCache sync.Map
```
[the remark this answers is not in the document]
jianoaix: We'd LRU this (with a capacity greater than max operators we allow)

relay/auth/authenticator.go
```diff
@@ -0,0 +1,174 @@
...
+		ics:                           ics,
+		authenticatedClients:          make(map[string]struct{}),
+		authenticationTimeouts:        make([]*authenticationTimeout, 0),
+		authenticationTimeoutDuration: authenticationTimeoutDuration,
+		keyCache:                      sync.Map{},
+	}
+}
+
+func (a *requestAuthenticator) AuthenticateGetChunksRequest(
+	address string,
```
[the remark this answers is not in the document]
cody-littley: renamed to `origin`

relay/auth/authenticator.go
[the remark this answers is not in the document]
cody-littley: Probably yes, given practical limitations on the number of DA nodes we will have. But in the interest in safety, I've converted this to a cache with a maximum size.

relay/auth/authenticator.go
```diff
@@ -0,0 +1,174 @@
...
+
+	a.savedAuthLock.Lock()
+	defer a.savedAuthLock.Unlock()
+
+	a.removeOldAuthentications(now)
+	_, ok := a.authenticatedClients[address]
+	return ok
+}
+
+// removeOldAuthentications removes any authentications that have expired.
```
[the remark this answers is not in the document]
cody-littley: comment added

relay/auth/request_signing.go
```diff
@@ -0,0 +1,50 @@
...
+
+import (
+	"encoding/binary"
+	pb "github.com/Layr-Labs/eigenda/api/grpc/relay"
+	"github.com/Layr-Labs/eigenda/core"
+	"golang.org/x/crypto/sha3"
+)
+
+// HashGetChunksRequest hashes the given GetChunksRequest.
+func HashGetChunksRequest(request *pb.GetChunksRequest) []byte {
```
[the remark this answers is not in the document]
cody-littley: Hmm... you are correct. Not sure if it's a security concern, since I'm not sure an attacker could really exploit this to do anything interesting. But probably best to be safe rather than sorry. I've updated this so that these two types of requests can't collide so trivially.

relay/auth/request_signing.go
cody-littley: change made

relay/server.go
cody-littley: Good idea. Done.

relay/auth/authenticator.go
```diff
@@ -0,0 +1,174 @@
...
+		return nil, fmt.Errorf("failed to get operators: %w", err)
+	}
+
+	operator, ok := operators[operatorID]
+	if !ok {
+		return nil, errors.New("operator not found")
+	}
+	key := operator.PubkeyG2
+
+	a.keyCache.Store(operatorID, key)
```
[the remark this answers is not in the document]
cody-littley: Good idea. Done.

```
func (a *requestAuthenticator) PreloadCache() error {
	blockNumber, err := a.ics.GetCurrentBlockNumber()
	if err != nil {
		return fmt.Errorf("failed to get current block number: %w", err)
	}
	operators, err := a.ics.GetIndexedOperators(context.Background(), blockNumber)
	if err != nil {
		return fmt.Errorf("failed to get operators: %w", err)
	}

	for operatorID, operator := range operators {
		a.keyCache.Add(operatorID, operator.PubkeyG2)
	}

	return nil
}
```

relay/auth/authenticator.go
```diff
@@ -47,29 +51,52 @@ type requestAuthenticator struct {
...
 		ics:                           ics,
 		authenticatedClients:          make(map[string]struct{}),
 		authenticationTimeouts:        make([]*authenticationTimeout, 0),
 		authenticationTimeoutDuration: authenticationTimeoutDuration,
-		keyCache:                      sync.Map{},
+		keyCache:                      keyCache,
+	}, nil
+}
+
+func (a *requestAuthenticator) PreloadCache() error {
```
jianoaix, line 77: Make it private and run in the the NewRequestAuthenticator?

jianoaix (approved):

relay/cmd/main.go
cody-littley: Converted to use a thegraph indexer

relay/cmd/flags/flags.go
```diff
@@ -221,6 +264,10 @@ var optionalFlags = []cli.Flag{
 	MaxGetChunkBytesPerSecondClientFlag,
 	GetChunkBytesBurstinessClientFlag,
 	MaxConcurrentGetChunkOpsClientFlag,
+	IndexerPullIntervalFlag,
```
ian-shim, line 267: We can remove this, right?

PR #912, opened 2024-11-19
Use power of 2 for blob length

mooselumph:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/encoding/kzg/prover/parametrized_prover.go
+++ b/encoding/kzg/prover/parametrized_prover.go
@@ -32,19 +32,19 @@ type rsEncodeResult struct {
 }
 
 type lengthCommitmentResult struct {
-	LengthCommitment bn254.G2Affine
+	LengthCommitment *bn254.G2Affine
 	Duration         time.Duration
 	Err              error
 }
 
 type lengthProofResult struct {
-	LengthProof bn254.G2Affine
+	LengthProof *bn254.G2Affine
 	Duration    time.Duration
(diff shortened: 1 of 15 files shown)


dmanc (approved):

ian-shim (approved):

encoding/kzg/prover/parametrized_prover.go
```diff
@@ -128,7 +128,7 @@ func (g *ParametrizedProver) GetCommitments(inputFr []fr.Element) (*bn254.G1Affi
 		start := time.Now()
 		commit, err := g.Computer.ComputeCommitment(inputFr)
 		commitmentChan <- commitmentResult{
-			Commitment: *commit,
```
mooselumph, line 131: This causes a nil-pointer dereference error when there is an error.

encoding/kzg/prover/parametrized_prover.go
```diff
@@ -32,19 +32,19 @@ type rsEncodeResult struct {
 }
 
 type lengthCommitmentResult struct {
-	LengthCommitment bn254.G2Affine
+	LengthCommitment *bn254.G2Affine
```
mooselumph, line 35: Is there any issue with using pointers here?

ian-shim:
@cody-littley This is the change that will make `Length` power of 2

PR #914, opened 2024-11-20
[v2] Paginate metadata records in controller

ian-shim:
Instead of querying all available metadata to process, limit the number of metadata to process per iteration and keep the cursor for the next iteration. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


disperser/common/v2/blobstore/dynamo_metadata_store.go
```diff
@@ -50,6 +50,11 @@ var (
 	ErrInvalidStateTransition = errors.New("invalid state transition")
 )
 
+type StatusIndexCursor struct {
+	PK        string
+	UpdatedAt uint64
+}
```
cody-littley, line 56: Low priority ask, could you add a short godoc for `PK`, it's not obvious at first glance what it means.

cody-littley (approved):

disperser/common/v2/blobstore/dynamo_metadata_store.go
ian-shim: Actually, you made me realize `PK` is probably not the right field to export from the cursor. Will change it to `BlobKey`

disperser/common/v2/blobstore/dynamo_metadata_store.go
```diff
@@ -174,6 +193,107 @@ func (s *BlobMetadataStore) GetBlobMetadataByStatus(ctx context.Context, status
 	return metadata, nil
 }
 
+// GetBlobMetadataByStatusPaginated returns all the metadata with the given status that were updated after the given cursor.
+// It also returns a new cursor (last evaluated key) to be used for the next page
+// even when there are no more results or there are no results at all.
+// This cursor can be used to get new set of results when they become available.
+// Therefore, it's possible to get an empty result from a request with exclusive start key returned from previous response.
+func (s *BlobMetadataStore) GetBlobMetadataByStatusPaginated(
```
jianoaix, line 201: How about actually build a cursor that streams the data?
The `StatusIndexCursor` is really a position or offset; and an actual Cursor will take a cursor position/offset, and keep iterating forward (and internalizing the bookkeeping of moving position).
A Cursor (or Iterator) will provide `Next()`, `Item()`, `Position()` etc. so the client (encoder manager and dispatcher) can stream through the data.
It'll be a better abstraction.

ian-shim: I agree that's a better abstraction for querying individual blobs one by one. 
However, there isn't a use case for that right now as the controller will always request a batch of the blobs by status. 
We can probably make your cursor abstraction work with batches, but I'm not sure if the usability will significantly improve vs. effort to introduce this new abstraction

jianoaix: My cursor abstraction can take in status which serves as a filter of blobs to iterate over

ian-shim: My main point wasn't that it can't take in status. It was more about the need to support iteration one by one vs. batches

disperser/common/v2/blobstore/dynamo_metadata_store.go
```diff
@@ -174,6 +193,107 @@ func (s *BlobMetadataStore) GetBlobMetadataByStatus(ctx context.Context, status
...
+	ctx context.Context,
+	status v2.BlobStatus,
+	exclusiveStartKey *StatusIndexCursor,
+	limit int32,
+) ([]*v2.BlobMetadata, *StatusIndexCursor, error) {
+	var cursor map[string]types.AttributeValue
+	if exclusiveStartKey != nil {
+		pk := blobKeyPrefix
+		if exclusiveStartKey.BlobKey != nil && len(exclusiveStartKey.BlobKey) == 32 {
```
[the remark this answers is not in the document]
ian-shim: Empty blob key is allowed though (no cursor on blob key)

disperser/common/v2/blobstore/dynamo_metadata_store.go
```diff
@@ -174,6 +193,107 @@ func (s *BlobMetadataStore) GetBlobMetadataByStatus(ctx context.Context, status
...
+		}
+		cursor = map[string]types.AttributeValue{
+			"PK": &types.AttributeValueMemberS{
+				Value: pk,
+			},
+			"SK": &types.AttributeValueMemberS{
+				Value: blobMetadataSK,
+			},
+			"UpdatedAt": &types.AttributeValueMemberN{
+				Value: strconv.FormatUint(exclusiveStartKey.UpdatedAt, 10),
```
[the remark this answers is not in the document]
ian-shim: If `len(blobKey) != 32`, exclusiveStartKey isn't ignored. PK defaults to `BlobKey#`, which means it should get all PKs

disperser/common/v2/blobstore/dynamo_metadata_store.go
```diff
@@ -174,6 +193,107 @@ func (s *BlobMetadataStore) GetBlobMetadataByStatus(ctx context.Context, status
...
+		// return the same cursor
+		return nil, exclusiveStartKey, nil
+	}
+
+	metadata := make([]*v2.BlobMetadata, len(res.Items))
+	for i, item := range res.Items {
+		metadata[i], err = UnmarshalBlobMetadata(item)
+		if err != nil {
+			return nil, nil, err
+		}
```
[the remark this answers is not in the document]
ian-shim: Updated it to it logs and skips those instead

pschork (approved):

disperser/common/v2/blobstore/dynamo_metadata_store.go
jianoaix: No the underlying mechanism isn't one by one -- it's the same by a small batch and buffer the results. Consuming that buffer is one by one, which is no difference than what's written here with a for loop.

jianoaix (approved):

disperser/common/v2/blobstore/dynamo_metadata_store.go
ian-shim: punting this as discussed offline

anupsv (commented):
lg

PR #915, opened 2024-11-20
[v2] Disperser request validation

ian-shim:
More validations on dispersal request

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/apiserver/disperse_blob_v2.go
+++ b/disperser/apiserver/disperse_blob_v2.go
@@ -70,11 +70,12 @@ func (s *DispersalServerV2) StoreBlob(ctx context.Context, data []byte, blobHead
 func (s *DispersalServerV2) validateDispersalRequest(req *pb.DisperseBlobRequest) error {
 	data := req.GetData()
 	blobSize := len(data)
-	if uint64(blobSize) > s.maxNumSymbolsPerBlob*encoding.BYTES_PER_SYMBOL {
-		return api.NewErrorInvalidArg(fmt.Sprintf("blob size cannot exceed %v bytes", s.maxNumSymbolsPerBlob*encoding.BYTES_PER_SYMBOL))
-	}
 	if blobSize == 0 {
 		return api.NewErrorInvalidArg("blob size must be greater than 0")
+	}
+	blobLength := encoding.GetBlobLengthPowerOf2(uint(blobSize))
+	if blobLength > uint(s.maxNumSymbolsPerBlob) {
+		return api.NewErrorInvalidArg("blob size too big")
 	}
 
 	blobHeaderProto := req.GetBlobHeader()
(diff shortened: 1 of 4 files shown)


disperser/apiserver/disperse_blob_v2.go
```diff
@@ -70,33 +70,40 @@ func (s *DispersalServerV2) StoreBlob(ctx context.Context, data []byte, blobHead
...
+	}
+
 	if len(blobHeaderProto.GetQuorumNumbers()) > int(s.onchainState.QuorumCount) {
 		return api.NewErrorInvalidArg(fmt.Sprintf("too many quorum numbers specified: maximum is %d", s.onchainState.QuorumCount))
 	}
 
+	for _, quorum := range blobHeaderProto.GetQuorumNumbers() {
+		if quorum > corev2.MaxQuorumID || uint8(quorum) > s.onchainState.QuorumCount {
+			return api.NewErrorInvalidArg(fmt.Sprintf("invalid quorum number %d; maximum is %d", quorum, s.onchainState.QuorumCount))
+		}
```
cody-littley, line 97: The quorum ID in the protobuf is `uint32`, why do you cast it into a `uint8`?

Do quorum numbers start at `0` or `1`? If they start at `0`, should it instead be `>= s.onchainState.QuorumCount`?

disperser/cmd/apiserver/flags/flags.go
```diff
@@ -150,7 +150,7 @@ var (
 	MaxNumSymbolsPerBlob = cli.UintFlag{
 		Name:     common.PrefixFlag(FlagPrefix, "max-num-symbols-per-blob"),
 		Usage:    "max number of symbols per blob. This flag is only relevant in v2",
-		Value:    65_536,
+		Value:    524_288, // this should allow for 16MiB blobs
```
cody-littley, line 153: Nit, would it make sense to express this as a mathematical expression? I verified that your number made sense, but I personally find it easier to do when the parts are split out. Feel free to ignore suggestion if you don't like it.

```suggestion
		Value:    1024 * 1024 * 16 / encoding.BYTES_PER_SYMBOL, // this should allow for 16MiB blobs
```

cody-littley (approved):

disperser/apiserver/disperse_blob_v2.go
ian-shim: > The quorum ID in the protobuf is uint32, why do you cast it into a uint8?

It needs to be converted to compare to `s.onchainState.QuorumCount`, which is `uint8`

> Do quorum numbers start at 0 or 1? If they start at 0, should it instead be >= s.onchainState.QuorumCount?

Nice catch! Will update.

PR #918, opened 2024-11-20
Relay timeouts

cody-littley:

Adds proper timeouts to relay operations. Should only be reviewed once https://github.com/Layr-Labs/eigenda/pull/911 merges.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


relay/cache/cached_accessor.go
```diff
@@ -114,11 +116,42 @@ func (c *cachedAccessor[K, V]) Get(key K) (V, error) {
...
+		return c.fetchResult(ctx, key, result)
+	}
+}
 
+// waitForResult waits for the result of a lookup that was initiated by another requester and returns it
+// when it becomes is available. This method will return quickly if the provided context is cancelled.
+// Doing so does not disrupt the other requesters that are also waiting for this result.
+func (c *cachedAccessor[K, V]) waitForResult(ctx context.Context, result *accessResult[V]) (V, error) {
+	wgChan := make(chan struct{}, 1)
+	go func() {
```
ian-shim, line 131: I think this goroutine will leak if context times out first.
If we replaced waitgroup with semaphore, I think you can do something like `sem.Acquire(ctx, N)`

relay/chunk_provider.go
```diff
@@ -11,6 +11,7 @@ import (
 	"github.com/Layr-Labs/eigenda/relay/chunkstore"
 	"github.com/Layr-Labs/eigensdk-go/logging"
 	"sync"
+	"time"
 )
 
 type chunkProvider struct {
```
ian-shim, line 17: same here, can we remove the ctx from struct?

relay/auth/authenticator_test.go
```diff
@@ -69,6 +69,7 @@ func TestValidRequest(t *testing.T) {
 	now := time.Now()
 
 	err = authenticator.AuthenticateGetChunksRequest(
+		context.Background(),
```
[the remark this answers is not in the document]
cody-littley: done

relay/blob_provider.go
```diff
@@ -20,6 +21,9 @@ type blobProvider struct {
 
 	// blobCache is an LRU cache of blobs.
 	blobCache cache.CachedAccessor[v2.BlobKey, []byte]
+
```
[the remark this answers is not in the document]
cody-littley: I think this actually needs to diverge from the normal patterns a bit.

The complicating factor is that we want to prevent two or more GRPCs that request the same blob from all doing an S3 fetch. This is the core purpose of the `CachedAccessor` class -- it forces duplicate requests to wait if there is already a fetch in progress. If we used the context from one of the callers, but then that connection broke and the context got cancelled, then other callers waiting on that data would instead get an error. So instead, the internal fetch happens using the server's context which intentionally ignores if the original caller cancels its context.

relay/cache/cached_accessor.go
cody-littley: I don't think this leaks. The wait group is completed inside `fetchResult()` on line 175 when the requested data has either been fetched or has returned an error while being fetched. That goroutine isn't impeded if the client's context gets cancelled.

That being said, it's better to have code that is obviously correct at first glance, rather than code that you have to stare at a while to realize that it's correct. I've switched this over to using a semaphore.

relay/cache/cached_accessor.go
```diff
@@ -1,24 +1,28 @@
...
 	"sync"
 )
 
 // CachedAccessor is an interface for accessing a resource that is cached. It assumes that cache misses
 // are expensive, and prevents multiple concurrent cache misses for the same key.
 type CachedAccessor[K comparable, V any] interface {
 	// Get returns the value for the given key. If the value is not in the cache, it will be fetched using the Accessor.
-	Get(key K) (V, error)
+	// If the context is cancelled, the function may abort early. If multiple goroutines request the same key,
+	// cancellation of one request will not affect the others.
```
jianoaix, line 15: nit: the cancellation of the first request can affect others (all of them get error)

relay/cache/cached_accessor.go
```diff
@@ -86,14 +89,13 @@ func NewCachedAccessor[K comparable, V any](
 
 func newAccessResult[V any]() *accessResult[V] {
 	result := &accessResult[V]{
-		wg: sync.WaitGroup{},
+		sem: semaphore.NewWeighted(1),
 	}
-	result.wg.Add(1)
+	result.sem.TryAcquire(1)
```
jianoaix, line 94: This always succeeds, should it just Acquire()?

jianoaix (approved):

relay/cache/cached_accessor.go
ian-shim: Right, what i meant is that this goroutine will survive after the parent method returns and meaninglessly block on wg until the wg completes

relay/cache/cached_accessor.go
cody-littley: This will always succeed either way, since this happens right after the semaphore is created and since no other goroutine has access to the semaphore yet. The reason why I use `TryAcquire()` is that it doesn't return an error. If I use `Acquire()` then I have to add error handling, even though it's impossible for an error to be returned.

With this in mind, are you ok with `TryAcquire(1)` here?

relay/cache/cached_accessor.go
cody-littley: Ah, right. The point I was making was that the wait group wouldn't block forever. But I agree that the goroutine would survive for longer than necessary. Regardless, this has now been refactored into a simpler solution using a semaphore.

relay/blob_provider.go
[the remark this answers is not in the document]
cody-littley: > The problem with using a global ctx is that this whole component is dead when that context gets killed

Yes. The context passed into the constructor of the provider should survive as long as the relay is running. Let's discuss this in our 1-1 today.

relay/cache/cached_accessor.go
cody-littley: The way I implemented this class, cancellation of the first request should not cause there to be an error for the following requests. Do you see a bug in the implementation?

jianoaix: Ok, I think it's about how to read "not affect", the failure of fetch is not the failure of first request, then this comment is correct. Anyway, it's fine to keep this.

relay/cache/cached_accessor.go
jianoaix: Does `_ = result.sem.Acquire()` work (with comment that this should always succeed)?

TryAquire() returns a boolean, so it's the same in the sense that the return value is ignored.

So both will work, the difference I think is that TryAcquire reads like it doesn't care about whether it has grabbed the resource.

ian-shim (approved):

relay/cache/cached_accessor.go
cody-littley: I've switched this to `_ = result.sem.Acquire(context.Background(), 1)`

PR #891, opened 2024-11-13
Integrate the grpc native instrumentation

jianoaix:
## Why are these changes needed?
Gain visibility at gRPC level.

## Checks

- [ ] I've made sure the tests are passing. Note that there might be a few flaky tests, in that case, please comment that they are not relevant.
- [ ] I've checked the new test coverage and the coverage percentage didn't drop.
- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests
   - [ ] This PR is not tested :(

PR #919, opened 2024-11-20
Add additional queue metrics

dmanc:

image

Added max queue capacity so we can do computations to get utilization (useful if we want to scale using percentage of the request pool). Also adds overall request pool utilization for convenience.

This PR also adds GRPC metrics similar to https://github.com/Layr-Labs/eigenda/pull/891

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/encoder/metrics.go
+++ b/disperser/encoder/metrics.go
@@ -27,6 +27,8 @@ type Metrics struct {
 	BlobSizeTotal         *prometheus.CounterVec
 	Latency               *prometheus.SummaryVec
 	BlobQueue             *prometheus.GaugeVec
+	QueueCapacity         prometheus.Gauge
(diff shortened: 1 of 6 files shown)


disperser/encoder/metrics.go
```diff
@@ -71,6 +73,20 @@ func NewMetrics(httpPort string, logger logging.Logger) *Metrics {
 			},
 			[]string{"size_bucket"},
 		),
+		QueueCapacity: promauto.With(reg).NewGauge(
```
jianoaix, line 76: This is configured, so doesn't look it should be exported via metric again.

Utilization is a division of queue size (already exported) by capacity.

jianoaix (approved):

PR #922, opened 2024-11-21
[v2] Read blob version params from chain

ian-shim:
Remove hardcoded blob version params map, and read the values from chain. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


ian-shim:
Depends on https://github.com/Layr-Labs/eigenda/pull/921

common/read_only_map.go
```diff
@@ -0,0 +1,26 @@
+package common
+
+type ReadOnlyMap[K comparable, V any] struct {
```
jianoaix, line 3: ah I'd wish golang has `const` which will make it self-explain and enforced

node/node.go
```diff
@@ -79,6 +80,10 @@ type Node struct {
 
 	mu            sync.Mutex
 	CurrentSocket string
+
+	// BlobVersionParams is a map of blob version parameters loaded from the chain.
+	// It is used to determine blob parameters based on the version number.
+	BlobVersionParams *corev2.BlobVersionParameterMap
```
jianoaix, line 86: This needs mutex protection for concurrent read/write

node/node_v2.go
```diff
@@ -61,7 +65,11 @@ func (n *Node) DownloadBundles(ctx context.Context, batch *corev2.Batch, operato
 		relayIndex := rand.Intn(len(cert.RelayKeys))
 		relayKey := cert.RelayKeys[relayIndex]
 		for _, quorum := range cert.BlobHeader.QuorumNumbers {
-			assgn, err := corev2.GetAssignment(operatorState, batch.BlobCertificates[0].BlobHeader.BlobVersion, quorum, n.Config.ID)
+			blobParams, ok := n.BlobVersionParams.Get(cert.BlobHeader.BlobVersion)
```
jianoaix, line 68: It looks when there is a new version upgrade, it may have an outage since different dispersers/nodes up to the different versions (and potentially cannot reach quorum threshold)?

node/node.go
ian-shim: The underlying map is read only and this pointer to the read only map gets swapped out atomically occasionally. Do we need synchronization here?

node/node_v2.go
ian-shim: Yes. The blob version needs to be registered onchain well before new blobs start using the new version.

node/node.go
ian-shim: I think you're right. Converted these into `atomic.Pointer`

node/node_v2.go
ian-shim: 1) Add new blob version onchain (multisig transaction)
2) Wait for `OnchainStateRefreshInterval`. If operator nodes, disperser, and relays use the default value (1h), all parts of system will be aware of the new version within this time
3) Blob with new version can be dispersed

node/flags/flags.go
```diff
@@ -224,6 +224,13 @@ var (
 		Required: false,
 		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "ENABLE_V2"),
 	}
+	OnchainStateRefreshIntervalFlag = cli.DurationFlag{
```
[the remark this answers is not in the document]
ian-shim: I was going to have this control onchain state that changes very infrequently, for now such as relay addresses. 
Maybe we can keep this name, and if we need to add other state that needs to be refreshed at different rate, we can add separate controls for those?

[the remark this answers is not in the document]
ian-shim: This reminds me that blob store duration isn't getting refreshed at all. It only gets read on bootup. Probably fine as they don't really change, but should we refresh those as part of this loop?

anupsv (approved):
few small comments

core/v2/assignment.go
```diff
@@ -58,9 +56,9 @@ func GetAssignments(state *core.OperatorState, blobVersion BlobVersion, quorum u
 		mp += int(a.chunks)
 	}
 
-	delta := int(params.NumChunks) - mp
+	delta := int(blobParams.NumChunks) - mp
```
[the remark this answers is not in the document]
ian-shim: I think it should be int. Otherwise, `delta` can underflow if `mp > numChunks`

node/flags/flags.go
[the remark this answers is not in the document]
jianoaix: They also change very infrequently, maybe at similar level as blob params. SG to include them.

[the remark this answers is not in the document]
ian-shim: Will include them in a follow up PR

PR #924, opened 2024-11-21
Size aware cache

cody-littley:

Makes relay caches aware of the size of their content (necessary for sane caching with if we want to support both large blobs and small blobs).

This should not be reviewed until https://github.com/Layr-Labs/eigenda/pull/918 merges.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


relay/cache/fifo-cache.go
```diff
@@ -0,0 +1,84 @@
...
+import (
+	"errors"
+	"github.com/Layr-Labs/eigenda/common/queue"
+)
+
+var _ Cache[string, string] = &FIFOCache[string, string]{}
+
+// FIFOCache is a cache that evicts the least recently added item when the cache is full. Useful for situations
+// where time of addition is a better predictor of future access than time of most recent access.
+type FIFOCache[K comparable, V any] struct {
```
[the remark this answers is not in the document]
cody-littley: No, it's just that FIFO is a lot easier to implement. I think it's premature to declare either to be faster at the moment, since we don't know for certain what the access pattern is going to be.

relay/cache/cache.go
```diff
@@ -0,0 +1,30 @@
+package cache
+
+// WeightCalculator is a function that calculates the weight of a key-value pair in a Cache.
+// By default, the weight of a key-value pair is 1. Cache capacity is always specified in terms of
+// the weight of the key-value pairs it can hold, rather than the number of key-value pairs.
+//
+// Unless otherwise noted, Cache implementations are not required to be thread safe.
+type WeightCalculator[K comparable, V any] func(key K, value V) uint64
```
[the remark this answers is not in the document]
cody-littley: This generalization is needed to support the current use case. There isn't a trivial and cheap way we can determine the size of an arbitrary data structure, so we need the ability to compute the size for different data types. For things like blobs that are just an array of bytes, it's easy. For things like the chunks that are in a complex data structure, we need an abstraction like this.

common/queue/queue.go
```diff
@@ -0,0 +1,18 @@
+package queue
+
+// Queue is an interface for a generic queue. It's absurd there isn't an equivalent in the standard golang libraries.
```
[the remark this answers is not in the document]
cody-littley: I've switched over to this library.

relay/cache/cache.go
```diff
@@ -0,0 +1,30 @@
...
+
+	// Put adds a key-value pair to the cache. After this operation, values may be dropped if the total weight
+	// exceeds the configured maximum weight. Will ignore the new value if it exceeds the maximum weight
+	// of the cache in and of itself.
+	Put(key K, value V)
+
+	// WithWeightCalculator sets the weight calculator for the cache. May only be called
+	// when the cache is empty. The weight calculator should be an idempotent function that
+	// always returns the same output given the same input.
+	WithWeightCalculator(weightCalculator WeightCalculator[K, V]) error
```
jianoaix, line 23: It'd be better to drop this method and pass it in from the constructor since it's required to have this calculator

relay/server.go
```diff
@@ -78,14 +78,14 @@ type Config struct {
 	// goroutines.
 	MetadataMaxConcurrency int
 
-	// BlobCacheSize is the maximum number of items in the blob cache.
-	BlobCacheSize int
+	// BlobCacheSize is the maximum size of the blob cache, in bytes.
+	BlobCacheSize uint64
```
jianoaix, line 82: May call it `BlobCacheSizeBytes` and then it's self-explaining

relay/cache/cache.go
```diff
@@ -0,0 +1,30 @@
+package cache
+
+// WeightCalculator is a function that calculates the weight of a key-value pair in a Cache.
+// By default, the weight of a key-value pair is 1. Cache capacity is always specified in terms of
+// the weight of the key-value pairs it can hold, rather than the number of key-value pairs.
+//
+// Unless otherwise noted, Cache implementations are not required to be thread safe.
```
jianoaix, line 7: why? It's quite often the use cases will need thread safety.

Also please move this line down to the Cache

relay/cache/cache.go
cody-littley: done

relay/server.go
cody-littley: done

relay/cache/cache.go
cody-littley: With the cache accessor, we are forced to use a mutex in order to perform atomic operations on the cache and the map of in-progress operations. Since that is currently our only use case, adding thread safety to the cache would be redundant and would provide no immediate benefit.

I agree that having a thread safe cache may be useful if we want to reuse it as a general purpose cache. I see two options:

- we make the cache inherently thread safe, even though not all use cases may need thread safety.
- we create a wrapper object that implements `Cache` that adds thread safety to another cache implementation, e.g. `func NewThreadSafeCache(cache Cache) Cache`. We could either write this now, or implement it later when we need it.

Of these options, which would you prefer?

jianoaix (approved):

relay/cache/cache.go
cody-littley: Currently there are no intended use cases outside of the cache accessor. Let's wait to add locking until we have a use case that justifies it.

PR #928, opened 2024-11-22
[v2] disperser client payments api

hopeyen:

properly construct payment header in the dispersal request

For easy initialization of DisperserClient, pass in an empty accountant `&accountant{}` and then call `client.PopulateAccountant(ctx)` to populate the accountant by getting payment information both on-chain and off-chain for the local signer.

For more customized initialization, the user can explicitly construct their accountant and pass in as the argument

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/accountant.go
+++ b/api/clients/accountant.go
@@ -159,9 +159,40 @@ func (a *accountant) GetRelativeBinRecord(index uint32) *BinRecord {
 	return &a.binRecords[relativeIndex]
 }
 
-func (a *accountant) SetPaymentState(paymentState *disperser_rpc.GetPaymentStateReply) {
-	a.minNumSymbols = uint32(paymentState.PaymentGlobalParams.MinNumSymbols)
+func (a *accountant) SetPaymentState(paymentState *disperser_rpc.GetPaymentStateReply) error {
+	if paymentState == nil {
+		return fmt.Errorf("payment state cannot be nil")
+	}
 
+	if paymentState.PaymentGlobalParams == nil {
+		return fmt.Errorf("payment global params cannot be nil")
+	}
+
+	if paymentState.OnchainCumulativePayment == nil {
+		return fmt.Errorf("onchain cumulative payment cannot be nil")
+	}
+
+	if paymentState.CumulativePayment == nil {
+		return fmt.Errorf("cumulative payment cannot be nil")
+	}
+
+	if paymentState.Reservation == nil {
+		return fmt.Errorf("reservation cannot be nil")
+	}
+
+	if paymentState.Reservation.QuorumNumbers == nil {
+		return fmt.Errorf("reservation quorum numbers cannot be nil")
+	}
+
+	if paymentState.Reservation.QuorumSplit == nil {
+		return fmt.Errorf("reservation quorum split cannot be nil")
+	}
+
+	if paymentState.BinRecords == nil {
+		return fmt.Errorf("bin records cannot be nil")
+	}
+
+	a.minNumSymbols = uint32(paymentState.PaymentGlobalParams.MinNumSymbols)
 	a.onDemand.CumulativePayment = new(big.Int).SetBytes(paymentState.OnchainCumulativePayment)
 	a.cumulativePayment = new(big.Int).SetBytes(paymentState.CumulativePayment)
 	a.pricePerSymbol = uint32(paymentState.PaymentGlobalParams.PricePerSymbol)
@@ -170,16 +201,19 @@ func (a *accountant) SetPaymentState(paymentState *disperser_rpc.GetPaymentState
 	a.reservation.StartTimestamp = uint64(paymentState.Reservation.StartTimestamp)
 	a.reservation.EndTimestamp = uint64(paymentState.Reservation.EndTimestamp)
 	a.reservationWindow = uint32(paymentState.PaymentGlobalParams.ReservationWindow)
+
 	quorumNumbers := make([]uint8, len(paymentState.Reservation.QuorumNumbers))
 	for i, quorum := range paymentState.Reservation.QuorumNumbers {
 		quorumNumbers[i] = uint8(quorum)
 	}
 	a.reservation.QuorumNumbers = quorumNumbers
+
 	quorumSplit := make([]uint8, len(paymentState.Reservation.QuorumSplit))
 	for i, quorum := range paymentState.Reservation.QuorumSplit {
 		quorumSplit[i] = uint8(quorum)
 	}
 	a.reservation.QuorumSplit = quorumSplit
+
 	binRecords := make([]BinRecord, len(paymentState.BinRecords))
 	for i, record := range paymentState.BinRecords {
 		binRecords[i] = BinRecord{
@@ -188,6 +222,8 @@ func (a *accountant) SetPaymentState(paymentState *disperser_rpc.GetPaymentState
 		}
 	}
 	a.binRecords = binRecords
+
+	return nil
 }
 
 // QuorumCheck eagerly returns error if the check finds a quorum number not an element of the allowed quorum numbers
--- a/api/clients/accountant.go
+++ b/api/clients/accountant.go
@@ -15,13 +15,7 @@ import (
 
 var requiredQuorums = []uint8{0, 1}
 
-type Accountant interface {
-	AccountBlob(ctx context.Context, numSymbols uint64, quorums []uint8) (*core.PaymentMetadata, error)
-}
-
-var _ Accountant = &accountant{}
-
-type accountant struct {
+type Accountant struct {
 	// on-chain states
 	accountID         string
 	reservation       *core.ActiveReservation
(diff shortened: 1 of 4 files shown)


hopeyen:
Waiting for #902 to merge before rebasing

api/clients/disperser_client_v2.go
```diff
@@ -60,7 +60,7 @@ var _ DisperserClientV2 = &disperserClientV2{}
 //
 //	// Subsequent calls will use the existing connection
 //	status2, blobKey2, err := client.DisperseBlob(ctx, data, blobHeader)
-func NewDisperserClientV2(config *DisperserClientV2Config, signer corev2.BlobRequestSigner, prover encoding.Prover) (*disperserClientV2, error) {
+func NewDisperserClientV2(config *DisperserClientV2Config, signer corev2.BlobRequestSigner, prover encoding.Prover, accountant Accountant) (*disperserClientV2, error) {
```
ian-shim, line 63: It's quite difficult to construct a new Accountant. Does it make sense to provide a helper method that creates an accountant from the `GetPaymentState` result? 

Let's also add a way to call `GetPaymentState` in this client

ian-shim (approved):

api/clients/disperser_client_v2.go
hopeyen: added a helper function `PopulateAccountant` that should run after creating `NewDisperserClientV2` by constructing accountant from `GetPaymentState`

What do you think about allowing the user of DisperserClientV2 to provide their own accountant, or simply passing in `&accountant{}` to `NewDisperserClientV2` and then explicitly calling `PopulateAccountant` to fill in the accountant state? ([5004cf9](https://github.com/Layr-Labs/eigenda/pull/928/commits/5004cf9ad401f641afce86088fd2633d1b947f87))

hopeyen: The reason is `GetPaymentState` is a function of `disperserClientV2`. If we write a standalone `PopulateAccountant`, then we would need it to be able to `initOnceGrpcConnection`, pass in a signer for creating the request, and a `disperser_rpc.DisperserClient` that makes the actual request. I figured it is cleaner to just use disperserClientV2

api/clients/disperser_client_v2.go
```diff
@@ -109,15 +121,11 @@ func (c *disperserClientV2) DisperseBlob(
 		return nil, [32]byte{}, api.NewErrorInternal("uninitialized signer for authenticated dispersal")
 	}
```
ian-shim, line 123: check if `c.accountant` is nil? 
we can instruct how to initialize accountant in the error message

api/clients/disperser_client_v2.go
```diff
@@ -75,13 +75,25 @@ func NewDisperserClientV2(config *DisperserClientV2Config, signer corev2.BlobReq
...
+		config:     config,
+		signer:     signer,
+		prover:     prover,
+		accountant: accountant,
 		// conn and client are initialized lazily
 	}, nil
 }
 
+// PopulateAccountant populates the accountant with the payment state from the disperser.
+// This function is required to be called before using the accountant. Perhaps rename to Start()?
```
ian-shim, line 87: This function isn't required if a valid accountant is passed in the constructor, right?
If you want to require this method, we can remove accountant as an argument from the constructor.

api/clients/accountant.go
```diff
@@ -160,6 +159,37 @@ func (a *accountant) GetRelativeBinRecord(index uint32) *BinRecord {
 	return &a.binRecords[relativeIndex]
 }
 
+func (a *accountant) SetPaymentState(paymentState *disperser_rpc.GetPaymentStateReply) {
+	a.minNumSymbols = uint32(paymentState.PaymentGlobalParams.MinNumSymbols)
```
ian-shim, line 163: these will panic if any of the attributes are nil?

hopeyen: yes.... added a check for all of them!
It would've been great if protobuf let me set fields to be required, but proto3 removed "required" keyword:(

api/clients/disperser_client_v2.go
hopeyen: Oh yeah, will update the comment

api/clients/accountant.go
ian-shim: If you use getters (i.e. `GetXXX`), you could chain them!

ian-shim (approved):
lgtm, lets address the lint issue

api/clients/accountant.go
```diff
@@ -160,6 +159,59 @@ func (a *accountant) GetRelativeBinRecord(index uint32) *BinRecord {
 	return &a.binRecords[relativeIndex]
 }
 
+func (a *accountant) SetPaymentState(paymentState *disperser_rpc.GetPaymentStateReply) error {
```
samlaf, line 162: This feels too coupled to me. A Public setter that is not part of the interface and it set by some response from another client. As discussed in other threads, could we not combine accountant and disperser_client more smoothly? Does accountant really need to be an independent struct?

api/clients/accountant.go
```diff
@@ -8,15 +8,15 @@ import (
...
 	"github.com/Layr-Labs/eigenda/core"
 	"github.com/Layr-Labs/eigenda/core/meterer"
 )
 
 var requiredQuorums = []uint8{0, 1}
 
 type Accountant interface {
-	AccountBlob(ctx context.Context, numSymbols uint64, quorums []uint8) (*commonpb.PaymentHeader, error)
+	AccountBlob(ctx context.Context, numSymbols uint64, quorums []uint8) (*core.PaymentMetadata, error)
 }
```
samlaf, line 20: Why is this interface defined if it's not being used in the disperser_client?

api/clients/disperser_client_v2.go
```diff
@@ -108,16 +123,15 @@ func (c *disperserClientV2) DisperseBlob(
 	if c.signer == nil {
 		return nil, [32]byte{}, api.NewErrorInternal("uninitialized signer for authenticated dispersal")
 	}
+	if c.accountant == nil {
+		return nil, [32]byte{}, api.NewErrorInternal("uninitialized accountant for paid dispersal; make sure to call PopulateAccountant after creating the client")
```
samlaf, line 127: how about we make the call to PopulateAccountant as part of the initOnce function? This way its hidden from the user.
Is there ever a need to recall PopulateAccountant? For eg if disperser went down, or whatever other weird case? To reset the accountant basically? If so, should it be called `ResetAccountantFromServer` or something like that?

api/clients/accountant.go
hopeyen: We wanted to keep accounting flow separate from the client, but arguably we can combine them into a bigger struct. What do you think @ian-shim ?

api/clients/disperser_client_v2.go
hopeyen: In this case, maybe we also move signer check to the initOnce?

There's some cases a user might recall populate accountant, I don't think it hurts to be able to reset. Can update the function name for sure

api/clients/accountant.go
hopeyen: `AccountBlob` is used in `DisperseBlob`, but I guess we wouldn't need the interface to access this function... I thought this is good practice, but happy to remove if that's not the case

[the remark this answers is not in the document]
hopeyen: Makes sense. Can update the names from `Accountant-> IAccountant` for interface and `accountant-> Accountant`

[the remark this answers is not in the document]
samlaf: Think `Accounter` would be more golangy? But in either case, Unless this interface is already used in many places, best practice is to just delete it. Interfaces should typically be defined at the call site, not where the struct is defined. That's because different call sites need different subsets of methods of a struct typically, so best to define the minimal interface needed (makes it easier to write a fake for that minimal interface).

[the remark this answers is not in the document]
hopeyen: I removed the interface definition and renamed the struct to Accounter

api/clients/disperser_client_v2.go
hopeyen: I think I'll leave this as a future refactoring

api/clients/accountant.go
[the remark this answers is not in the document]
samlaf: Wah sorry! The struct should still be called Accountant. I was saying IF we were to have an interface here (which I don't think we should), then it could be called `Accounter` instead of `IAccountant`. It's typical of golang to use verbs for interface, like Writer and Reader.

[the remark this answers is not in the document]
samlaf: LGTM after [6643ca9](https://github.com/Layr-Labs/eigenda/pull/928/commits/6643ca9280c1c4d8cd7c59cde05a939e9813a5eb); Ill resolve this

samlaf (approved):
I'll give a stamp on this in case its blocking progress on other work, but I would have liked to get @ian-shim's take on https://github.com/Layr-Labs/eigenda/pull/928#discussion_r1872241197. Would really like to find a better interface/decoupling between these. But we can do so later (hopefully...)

api/clients/accountant.go
samlaf: Makes sense! I would also not try to future-proof this design too much (to a case where there are multiple dispersers and you need some super complex accountant that manages multiple dispersers). We should try to make the simple case simple, true to golang's philosophy, and people who need complex integrations can take on the complexity hit around our simple API.

Another possibility is to have 2 constructors: one where we inject the accountant, and the other that initializes the accountant by calling the disperser?

hopeyen: Hmm all one needs to do is to call the current constructor and then call `PopulateAccountant`, so adding a separate constructor doesn't seem that beneficial to me. I won't mind adding it, probably in the follow-up [PR](https://github.com/Layr-Labs/eigenda/pull/965) that focus on integrations adaptions (maybe something like `NewDisperserClientV2AutoPay`?

PR #930, opened 2024-11-22
Optimize the Disperser perf

jianoaix:
The Disperser in particular the retrieval latency was 60+ seconds, when there was high traffic and/or large blobs.

There was due to inefficient memory management when dealing with blob buffer from S3, making it spend a lot of CPU/time on memory allocation and GC.

**Tested:**
From local retrieval query for a large blob,

Before:
```
time grpcurl -max-msg-sz 15000000 -proto ./api/proto/disperser/disperser.proto -import-path ./api/proto -d '{"batch_header_hash": "G9ntJR15dlOlljWQnx2avoT39XqLD5STAo91B8OMBqw=", "blob_index": 613 }' disperser-holesky.eigenda.xyz:443 disperser.Disperser/RetrieveBlob > /dev/null 2>&1
real	0m7.335s
user	0m0.357s
sys	0m0.157s
```

After:
```
time grpcurl -max-msg-sz 15000000 -proto ./api/proto/disperser/disperser.proto -import-path ./api/proto -d '{"batch_header_hash": "G9ntJR15dlOlljWQnx2avoT39XqLD5STAo91B8OMBqw=", "blob_index": 613 }' disperser-holesky.eigenda.xyz:443 disperser.Disperser/RetrieveBlob > /dev/null 2>&1

real	0m0.847s
user	0m0.382s
sys	0m0.152s
```

From pprof:

Before:
mem

After:

Screenshot 2024-11-22 at 2 45 13 PM

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/common/aws/s3/client.go
+++ b/common/aws/s3/client.go
@@ -4,6 +4,7 @@ import (
 	"bytes"
 	"context"
 	"errors"
+	"fmt"
 	"runtime"
 	"sync"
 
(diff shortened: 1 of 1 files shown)


dmanc (approved):
:shipit:

common/aws/s3/client.go
```diff
@@ -105,15 +106,32 @@ func NewClient(ctx context.Context, cfg commonaws.ClientConfig, logger logging.L
 	return ref, err
 }
 
+func PeekObjectSize(ctx context.Context, s3Client *s3.Client, bucket, key string) (int64, error) {
```
ian-shim, line 109: We already have this method in this file `HeadObject`. Could we have used that one?

PR #931, opened 2024-11-22
[v2][node] Construct relay client from relay address from chain

ian-shim:
Node creates relay client with relay address from chain

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/common/read_only_map.go
+++ b/common/read_only_map.go
@@ -1,10 +1,14 @@
 package common
 
-type ReadOnlyMap[K comparable, V any] struct {
+import (
+	"maps"
+)
+
+type ReadOnlyMap[K comparable, V comparable] struct {
 	data map[K]V
 }
 
-func NewReadOnlyMap[K comparable, V any](data map[K]V) *ReadOnlyMap[K, V] {
+func NewReadOnlyMap[K comparable, V comparable](data map[K]V) *ReadOnlyMap[K, V] {
 	return &ReadOnlyMap[K, V]{data: data}
 }
 
@@ -23,4 +27,8 @@ func (m *ReadOnlyMap[K, V]) Keys() []K {
 
 func (m *ReadOnlyMap[K, V]) Len() int {
 	return len(m.data)
+}
+
+func (m *ReadOnlyMap[K, V]) Equal(data map[K]V) bool {
+	return maps.Equal(m.data, data)
 }
--- a/core/v2/types.go
+++ b/core/v2/types.go
@@ -161,7 +161,7 @@ func (b *BlobHeader) GetEncodingParams(blobParams *core.BlobVersionParameters) (
 	}, nil
 }
 
-type RelayKey uint16
+type RelayKey = uint16
 
 type BlobCertificate struct {
 	BlobHeader *BlobHeader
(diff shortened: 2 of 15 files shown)


common/read_only_map.go
```diff
@@ -24,3 +26,7 @@ func (m *ReadOnlyMap[K, V]) Keys() []K {
 func (m *ReadOnlyMap[K, V]) Len() int {
 	return len(m.data)
 }
+
+func (m *ReadOnlyMap[K, V]) Equal(data map[K]V) bool {
+	return reflect.DeepEqual(m.data, data)
```
jianoaix, line 31: Reflection may have perf issue, so may be not a good choice for non testing use.

core/v2/types.go
```diff
@@ -161,7 +161,7 @@ func (b *BlobHeader) GetEncodingParams(blobParams *core.BlobVersionParameters) (
 	}, nil
 }
 
-type RelayKey uint16
+type RelayKey = uint16
```
jianoaix, line 164: What's the consideration of changing to alias? The current one provides type safety which is often desired

node/node.go
```diff
@@ -392,13 +410,47 @@ func (n *Node) RefreshOnchainState(ctx context.Context) error {
...
+				n.Logger.Error("error fetching relay URLs", "err", err)
+				continue
+			}
+
+			if reflect.DeepEqual(existingURLs, relayURLs) {
+				n.Logger.Info("No change in relay URLs")
+				continue
+			}
+
+			relayClient, err := clients.NewRelayClient(&clients.RelayClientConfig{
```
jianoaix, line 444: Relay url may be updated much more frequently than protocol params, should it be in a different interval/thread?

core/v2/types.go
ian-shim: It's mostly to interoperate with chainio library which still lives in core and can't import v2 due to import cycles. 
So chainio library uses `uint16` instead of `RelayKey` type, and I updated this to type alias to convert the output from chainio without an explicit conversion method

common/read_only_map.go
ian-shim: Good point

node/node.go
ian-shim: Yeah, I agree in terms of relative frequency, but I still think the frequency for new relays being added and getting used would be in order of hours if not days.

node/node_v2.go
```diff
@@ -34,7 +34,8 @@ type RawBundles struct {
 }
 
 func (n *Node) DownloadBundles(ctx context.Context, batch *corev2.Batch, operatorState *core.OperatorState) ([]*corev2.BlobShard, []*RawBundles, error) {
-	if n.RelayClient == nil {
+	relayClient, ok := n.RelayClient.Load().(clients.RelayClient)
```
jianoaix, line 37: will this panic if Load() returns nil?

jianoaix (approved):

node/node_v2.go
ian-shim: No. In that case, `relayClient` will be `nil` and `ok` will be false

PR #932, opened 2024-11-25
Do not fail the object fetch when object size is unknown

jianoaix:
If it fails to peek the object size, we just use a default size and continue to download the blob (instead of failing it).

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/common/aws/s3/client.go
+++ b/common/aws/s3/client.go
@@ -18,6 +18,10 @@ import (
 	"golang.org/x/sync/errgroup"
 )
 
+const (
+	defaultBlobBufferSizeByte = 128 * 1024
+)
+
 var (
 	once              sync.Once
 	ref               *client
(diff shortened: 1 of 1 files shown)


common/aws/s3/client.go
```diff
@@ -106,24 +105,13 @@ func NewClient(ctx context.Context, cfg commonaws.ClientConfig, logger logging.L
...
-		return 0, fmt.Errorf("failed to head object: %w", err)
-	}
-	return *result.ContentLength, nil
-}
-
 func (s *client) DownloadObject(ctx context.Context, bucket string, key string) ([]byte, error) {
-	size, err := PeekObjectSize(ctx, s.s3Client, bucket, key)
-	if err != nil {
-		return nil, err
+	objectSize := 128 * 1024
```
dmanc, line 109: How was this number chosen? Also should make it a const

dmanc (approved):

common/aws/s3/client.go
jianoaix: Just give it a moderate number so it won't waste a lot, nor need to realloc the memory too many times.

PR #938, opened 2024-11-27
Relay metrics

cody-littley:

Adds metrics to the relay server.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


common/metrics/metrics.go
```diff
@@ -140,3 +155,14 @@ type LatencyMetric interface {
 	// when creating the metric, an error will be returned.
 	ReportLatency(latency time.Duration, label ...any)
 }
+
+// RunningAverageMetric tracks the average of a series of values over a given time window.
+type RunningAverageMetric interface {
```
jianoaix, line 160: You probably don't need this. Running average is a dashboarding problem not metric.

cody-littley: Removed `RunningAverageMetric` and replaced with gauges.

relay/cache/cache_accessor.go
```diff
@@ -47,6 +48,9 @@ type cacheAccessor[K comparable, V any] struct {
 	// cache is the underlying cache that this wrapper manages.
 	cache Cache[K, V]
 
+	// weightCalculator is the function used to calculate the weight of a key-value pair.
+	calculator WeightCalculator[K, V]
```
ian-shim, line 52: What is this calculator or the weight used for? 
I feel like abstractions like this make the code bit harder to read

relay/cache/cache_accessor.go
```diff
@@ -55,17 +59,29 @@ type cacheAccessor[K comparable, V any] struct {
 
 	// accessor is the function used to fetch values that are not in the cache.
 	accessor Accessor[K, V]
+
+	// insertionTimes is a map of keys to the time they were inserted into the cache. Used to calculate the average
+	// lifespan of items in the cache.
+	insertionTimes map[K]time.Time
```
ian-shim, line 65: Is it going to keep the insertion times for every single key?

relay/cache/cache_accessor.go
```diff
@@ -74,11 +90,30 @@ func NewCacheAccessor[K comparable, V any](
...
+		evictionHandler = func(key K, _ V) {
+			if insertionTime, ok := insertionTimes[key]; ok {
+				lifespan := time.Since(insertionTime).Milliseconds()
+				metrics.lifespan.Set(float64(lifespan))
+				delete(insertionTimes, key)
+			}
+		}
+	}
+
+	cache := NewFIFOCache(maxWeight, calculator, evictionHandler)
```
ian-shim, line 107: why do we now create this cache from the constructor vs. taking it as a parameter?

relay/limiter/chunk_rate_limiter.go
```diff
@@ -36,12 +37,17 @@ type ChunkRateLimiter struct {
...
+
 	// this lock is used to provide thread safety
 	lock sync.Mutex
 }
 
 // NewChunkRateLimiter creates a new ChunkRateLimiter.
-func NewChunkRateLimiter(config *Config) *ChunkRateLimiter {
+func NewChunkRateLimiter(
+	config *Config,
+	limiterCounter metrics.CountMetric) *ChunkRateLimiter {
```
ian-shim, line 50: nit: it's kind of strange that it's only taking in the count metric. The metrics a component is tracking probably shouldn't be defined by constructor arguments

relay/server.go
```diff
@@ -124,8 +130,8 @@ func NewServer(
 	blobStore *blobstore.BlobStore,
 	chunkReader chunkstore.ChunkReader,
 	chainReader core.Reader,
-	ics core.IndexedChainState,
-) (*Server, error) {
```
ian-shim, line 128: nit: not sure if there's any explicit recommendations on this, but I find this format more readable 😅

relay/server.go
```diff
@@ -311,6 +346,10 @@ func (s *Server) GetChunks(ctx context.Context, request *pb.GetChunksRequest) (*
 		return nil, fmt.Errorf("error gathering chunk data: %w", err)
 	}
 
+	finishedFetchingData := time.Now()
+	s.metrics.GetChunksDataLatency.ReportLatency(finishedFetchingData.Sub(finishedFetchingMetadata))
```
ian-shim, line 350: nit: `time.Since(finishedFetchingMetadata)`

ian-shim (commented):
lgtm, just few comments

relay/cache/cache_accessor.go
cody-littley: It's used to calculate the size of each individual item in the cache. Things like blobs are straight forward, you just take the length of the byte array. Chunks are much less straight forward, since they are stored in a more complex data structure.

I agree that these types of abstractions add complexity. I'm just not sure how to make the cache sufficiently generic as to work with both blobs and chunks without it.

relay/cache/cache_accessor.go
cody-littley: Just the keys currently in the cache. When the cache evicts something, it uses the time to determine how long the item was in the cache, and then it removes it from this map.

relay/cache/cache_accessor.go
cody-littley: This is necessary due to the new `evictionHandler`, which is used to determine how long elements are in the cache before they are evicted. A few options if you'd like to back to the prior pattern:

- remove the metric that reports on the time each element remains in the cache (I think this will be a useful metric, would like to figure out how to keep it)
- add a method `Cache.SetEvictionHandler(handler)` that allows an eviction callback to be registered after the cache is constructed
- use a builder pattern, i.e. where there is a `CacheBuilder` class that you can configure by calling optional setters that eventually builds the cache

relay/server.go
cody-littley: reverted format change

relay/server.go
cody-littley: change made

relay/limiter/chunk_rate_limiter.go
cody-littley: I've refactored this. Instead of taking an individual metric, it now accepts an instance of the `RelayMetrics` wrapper that contains all relay metrics.

jianoaix (approved):

ian-shim (approved):

PR #939, opened 2024-11-27
Add latency metric for dispersal/retrieval with blob size breakdown

jianoaix:
To enable the understanding of dispersal/retrieval latency v.s. blob sizes.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/metrics.go
+++ b/disperser/metrics.go
@@ -24,6 +24,7 @@ type Metrics struct {
 	NumBlobRequests *prometheus.CounterVec
 	NumRpcRequests  *prometheus.CounterVec
 	BlobSize        *prometheus.GaugeVec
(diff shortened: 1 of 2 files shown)


disperser/metrics.go
```diff
@@ -78,6 +79,15 @@ func NewMetrics(reg *prometheus.Registry, httpPort string, logger logging.Logger
 			},
 			[]string{"method"},
 		),
+		BlobLatency: promauto.With(reg).NewGaugeVec(
```
jianoaix, line 82: Using gauge instead of summary, to reduce the number of time series -- it should be sufficient to just have gauge to get insights for this latency breakdown

dmanc (approved):

PR #941, opened 2024-12-02
Add timeout to download chunks from node

ian-shim:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/node/flags/flags.go
+++ b/node/flags/flags.go
@@ -231,6 +231,13 @@ var (
 		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "ONCHAIN_STATE_REFRESH_INTERVAL"),
 		Value:    1 * time.Hour,
 	}
+	ChunkDownloadTimeoutFlag = cli.DurationFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "chunk-download-timeout"),
+		Usage:    "The timeout for downloading chunks from the relay (default: 30s)",
+		Required: false,
+		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "CHUNK_DOWNLOAD_TIMEOUT"),
+		Value:    20 * time.Second,
+	}
 
 	// Test only, DO NOT USE the following flags in production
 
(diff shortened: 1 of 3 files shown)


node/flags/flags.go
```diff
@@ -231,6 +231,13 @@ var (
 		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "ONCHAIN_STATE_REFRESH_INTERVAL"),
 		Value:    1 * time.Hour,
 	}
+	ChunkDownloadTimeoutFlag = cli.DurationFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "chunk-download-timeout"),
+		Usage:    "The timeout for downloading chunks from the relay (default: 30s)",
+		Required: false,
+		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "CHUNK_DOWNLOAD_TIMEOUT"),
+		Value:    30 * time.Second,
+	}
```
cody-littley, line 240: The default timeout on the relay side of the fence is 20 seconds (`relay/cmd/flags/flags.go: GetChunksTimeoutFlag`). I don't really care what value we use, but we should probably assign the same value for both settings.

node/node_v2.go
```diff
@@ -103,21 +103,24 @@ func (n *Node) DownloadBundles(ctx context.Context, batch *corev2.Batch, operato
...
+			ctxTimeout, cancel := context.WithTimeout(ctx, n.Config.ChunkDownloadTimeout)
+			bundles, err := relayClient.GetChunksByRange(ctxTimeout, relayKey, req.chunkRequests)
 			if err != nil {
 				n.Logger.Errorf("failed to get chunks from relays: %v", err)
 				bundleChan <- response{
 					metadata: nil,
 					bundles:  nil,
 					err:      err,
 				}
+				cancel()
```
cody-littley, line 115: Could we just do a `defer cancel()` at the time we create the context?

cody-littley (approved):

jianoaix (approved):

PR #942, opened 2024-12-02
Payments storage packing

0x0aa0:
- introduces new storage layout
- rename bin -> period
- link PaymentVault on DASM


contracts/src/interfaces/IPaymentVault.sol
```diff
@@ -14,21 +14,21 @@ interface IPaymentVault {
...
-        uint256 previousMinNumSymbols, 
-        uint256 newMinNumSymbols, 
-        uint256 previousPricePerSymbol, 
-        uint256 newPricePerSymbol, 
-        uint256 previousPriceUpdateCooldown, 
-        uint256 newPriceUpdateCooldown
+        uint128 previousMinNumSymbols, 
+        uint128 newMinNumSymbols, 
+        uint128 previousPricePerSymbol, 
+        uint128 newPricePerSymbol,
```
ian-shim, line 29: How much gas savings are we talking here? I assume we're going to price this in mwei, but I wonder if we should just keep this in wei if the cost savings are trivial. 
(If we do need to use mwei, can we document that?)

contracts/src/payments/PaymentVault.sol
```diff
@@ -134,13 +134,13 @@ contract PaymentVault is PaymentVaultStorage, OwnableUpgradeable {
...
         return onDemandPayments[_account];
     }    
 
     /// @notice Fetches the current total on demand balances for a set of accounts
-    function getOnDemandAmounts(address[] memory _accounts) external view returns (uint256[] memory _payments) {
-        _payments = new uint256[](_accounts.length);
+    function getOnDemandAmounts(address[] memory _accounts) external view returns (uint128[] memory _payments) {
+        _payments = new uint128[](_accounts.length);
         for(uint256 i; i < _accounts.length; ++i){
             _payments[i] = onDemandPayments[_accounts[i]];
```
ian-shim, line 145: should we validate if `_accounts[i]` exists in `onDemandPayments`?

hopeyen (commented):
DASM link looks good. 
For storage packing, I think we can revert back to using uint256 for `onDemandPayments` and in the future we will need to add the collected amount to the same slot (like `_onDemandPaymentsCollected_ || _onDemandPaymentsDeposited_`. The current deposit function will only take a uint128 to update the lower slot, and a new collect function only callable by operators who supply a payment verification over the claimed collection. 

Sorry that this is going back and forth a little bit

hopeyen (approved):
lgtm!

PR #946, opened 2024-12-02
Automatically generate protobuf documentation

cody-littley:

Automatically generate protobuf documentation when compiling protobufs.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/docs/eigenda-protos.md
```diff
@@ -0,0 +1,1661 @@
+# Protocol Documentation
+<a name="top"></a>
+
+## Table of Contents
+
+- [churner/churner.proto](#churner_churner-proto)
```
jianoaix, line 6: It's unclear if it's best to put all docs together. 
There 3 categories of APIs for different roles: operators, users (consumers), and admin (-ish, like churner/ejector). It may be clearer for people to visit a page with APIs relevant to them.

api/grpc/lightnode/lightnode.pb.go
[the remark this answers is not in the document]
cody-littley: The light node protobufs are currently incomplete, and I think it's likely we will revisit the design before continuing again. I can restore this file if you think its worthwhile.

api/docs/eigenda-protos.md
cody-littley: The script now generates both the individual documents for each proto file as well as a unified document in case somebody prefers that format. If you don't think it's helpful, I can delete the unified file.

jianoaix (approved):

PR #948, opened 2024-12-03
Node metrics

cody-littley:

Adds metrics to the v2 DA node.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


node/config.go
```diff
@@ -207,7 +208,7 @@ func NewConfig(ctx *cli.Context) (*Config, error) {
 		EnableNodeApi:                  ctx.GlobalBool(flags.EnableNodeApiFlag.Name),
 		NodeApiPort:                    ctx.GlobalString(flags.NodeApiPortFlag.Name),
 		EnableMetrics:                  ctx.GlobalBool(flags.EnableMetricsFlag.Name),
-		MetricsPort:                    ctx.GlobalString(flags.MetricsPortFlag.Name),
+		MetricsPort:                    ctx.GlobalInt(flags.MetricsPortFlag.Name),
```
ian-shim, line 211: Is this backward compatible? i.e. if node doesn't update the env var from string to int, does it continue to work?

node/grpc/server_v2.go
```diff
@@ -101,6 +110,8 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
 			return
 		}
 
+		s.metrics.ReportStoreChunksDataSize(size)
```
ian-shim, line 113: what if the store operation gets reverted in L125?

node/grpc/server_v2.go
```diff
@@ -124,6 +135,10 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
 	}
 
 	sig := s.node.KeyPair.SignMessage(batchHeaderHash).Bytes()
+
+	timeElapsed := time.Since(start)
+	s.metrics.ReportStoreChunksLatency(timeElapsed)
```
ian-shim, line 140: nit: `s.metrics.ReportStoreChunksLatency(time.Since(start))`?

node/grpc/v2_metrics.go
```diff
@@ -0,0 +1,202 @@
...
+
+	if m.dbSizePollPeriod.Nanoseconds() == 0 {
+		return
+	}
+	go func() {
+		ticker := time.NewTicker(m.dbSizePollPeriod)
+
+		for m.isAlive.Load() {
+			var size int64
+			err := filepath.Walk(m.dbDir, func(_ string, info os.FileInfo, err error) error {
```
ian-shim, line 154: is this thread safe?

node/grpc/v2_metrics.go
```diff
@@ -0,0 +1,202 @@
...
+
+	dbSize           *prometheus.GaugeVec
+	dbSizePollPeriod time.Duration
+	dbDir            string
+	isAlive          *atomic.Bool
+}
+
+// NewV2Metrics creates a new V2Metrics instance. dbSizePollPeriod is the period at which the database size is polled.
+// If set to 0, the database size is not polled.
+func NewV2Metrics(
```
ian-shim, line 44: Is this compatible with v1 metrics? 
v1 metrics also registers mux and initializes its own server

node/grpc/server_v2.go
```diff
@@ -166,6 +183,15 @@ func (s *ServerV2) GetChunks(ctx context.Context, in *pb.GetChunksRequest) (*pb.
 		return nil, api.NewErrorInternal(fmt.Sprintf("failed to get chunks: %v", err))
 	}
 
+	var size uint64
+	for _, chunk := range chunks {
```
jianoaix, line 187: len(chunk) * len(chunks)?

node/config.go
cody-littley: Bash variables are untyped. From bash's perspective, everything is just a string (regardless of quotation mark usage). Changing the flag from a `GlobalString` to a `GlobalInt` just causes golang to parse the data into an int when it reads it.

https://tldp.org/LDP/abs/html/untyped.html

node/grpc/server_v2.go
cody-littley: As a general rule of thumb, should we report incremental metrics if the operation as a whole fails? Or should we only report metrics for an operation if it is successful? (in another PR, you suggested that I should report latencies even when there are failures).

I can make this only report if the request ends up being valid, but I want to be consistent with the way we handle scenarios like this.

node/grpc/server_v2.go
cody-littley: done

node/grpc/v2_metrics.go
cody-littley: It is almost certainly not thread safe (i.e. if level DB deletes a file or a directory mid-walk, then `filepath.Walk()` will return an error). My hope was that if the race condition was sufficiently rare, we could still extract meaningful metrics data.

Currently, this will log an error whenever this method is unable to fetch new data. Will an error log cause problems if it triggers every once in a while? If so, should this be downgraded to a a `logger.info()` call?

Unfortunately, levelDB doesn't expose API that tells you the size of the DB (that I know of). My reasoning was that this metric would be sufficiently valuable to justify a hacky collection method.

In theory, we could have the levelDB wrapper track the quantity of data, at the cost of some extra book keeping (every DB modification would need to update a special `size` key-value pair). This wouldn't tell us the size of the files on disk (which may vary depending on things like compaction and indexes), but would give us a very good idea of the approximate size if the DB. If I implemented such a thing, it would need to be in a stand alone PR though.

node/grpc/server_v2.go
cody-littley: done

```
	size := 0
	if len(chunks) > 0 {
		size = len(chunks[0]) * len(chunks)
	}
	s.metrics.ReportGetChunksDataSize(size)
```

node/grpc/v2_metrics.go
cody-littley: Ok, I've now made it so we use the same registry/server as v1. The result is a bit hacky since the v1 metrics code is in another repo.

node/grpc/server_v2.go
```diff
@@ -166,6 +184,15 @@ func (s *ServerV2) GetChunks(ctx context.Context, in *pb.GetChunksRequest) (*pb.
...
 	}
 
+	size := 0
+	if len(chunks) > 0 {
+		size = len(chunks[0]) * len(chunks)
+	}
+	s.metrics.ReportGetChunksDataSize(size)
+
+	elapsed := time.Since(start)
+	s.metrics.ReportGetChunksLatency(elapsed)
```
jianoaix, line 194: It seems "elapsed" can be inlined to Report call

node/grpc/v2_metrics.go
```diff
@@ -0,0 +1,175 @@
...
+	"os"
+	"path/filepath"
+	"sync/atomic"
+	"time"
+)
+
+const namespace = "eigenda_node"
+
+// V2Metrics encapsulates metrics for the v2 DA node.
+type V2Metrics struct {
```
jianoaix, line 19: MetricsV2? It's in "XxxV2" format for other components

node/grpc/v2_metrics.go
```diff
@@ -0,0 +1,175 @@
...
+}
+
+// GetGRPCServerOption returns the gRPC server option that enables automatic GRPC metrics collection.
+func (m *V2Metrics) GetGRPCServerOption() grpc.ServerOption {
+	return m.grpcServerOption
+}
+
+func (m *V2Metrics) ReportStoreChunksLatency(latency time.Duration) {
+	m.storeChunksLatency.WithLabelValues().Observe(
+		float64(latency.Nanoseconds()) / float64(time.Millisecond))
```
jianoaix, line 161: float64(latency.Milliseconds()) should work

node/store_v2.go
```diff
@@ -19,7 +19,9 @@ const (
 )
 
 type StoreV2 interface {
-	StoreBatch(batch *corev2.Batch, rawBundles []*RawBundles) ([]kvstore.Key, error)
+	// StoreBatch stores a batch and its raw bundles in the database. Returns the keys of the stored data
```
jianoaix, line 22: Any atomicity guarantees which should be commented?

Same for "DeleteKeys"

node/flags/flags.go
```diff
@@ -98,6 +98,13 @@ var (
 		Required: true,
 		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "DB_PATH"),
 	}
+	DBSizePollPeriodFlag = cli.DurationFlag{
```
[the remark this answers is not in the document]
cody-littley: added metrics prefix

node/grpc/server_v2.go
cody-littley: done

node/grpc/v2_metrics.go
cody-littley: renamed

node/grpc/v2_metrics.go
cody-littley: I'm intentionally not using that (@ian-shim made the same suggestion on another PR 😜). `latency.Milliseconds()` returns an `int`, meaning we lose all sub-millisecond fidelity in the measurement. For many metrics this isn't a hug deal (e.g. if you are measuring something that takes 100s of milliseconds), but for some the precision is nice to have.

Let me know if you'd like to discuss this further.

node/store_v2.go
cody-littley: added requested documentation

```
type StoreV2 interface {

	// StoreBatch stores a batch and its raw bundles in the database. Returns the keys of the stored data
	// and the size of the stored data, in bytes.
	//
	// All modifications to the database within this method are performed atomically.
	StoreBatch(batch *corev2.Batch, rawBundles []*RawBundles) ([]kvstore.Key, uint64, error)

	// DeleteKeys deletes the keys from local storage.
	//
	// All modifications to the database within this method are performed atomically.
	DeleteKeys(keys []kvstore.Key) error

	// GetChunks returns the chunks of a blob with the given blob key and quorum.
	GetChunks(blobKey corev2.BlobKey, quorum core.QuorumID) ([][]byte, error)
}
```

node/grpc/v2_metrics.go
jianoaix: I think it depends on what the expected number we will deal with here. If it's ultra low latency and sub-ms matters, `Microseconds()` may be an option.

node/grpc/metrics_v2.go
```diff
@@ -0,0 +1,175 @@
...
+	//registry.MustRegister(collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}))
+	//registry.MustRegister(collectors.NewGoCollector())
+
+	grpcMetrics := grpcprom.NewServerMetrics()
+	registry.MustRegister(grpcMetrics)
+	grpcServerOption := grpc.UnaryInterceptor(
+		grpcMetrics.UnaryServerInterceptor(),
+	)
+
+	storeChunksLatency := promauto.With(registry).NewSummaryVec(
```
jianoaix, line 54: We may not need these primitive level metrics as gRPC already provides (need to integrate / enable it)

node/grpc/v2_metrics.go
cody-littley: For the sake of consistency, I've been converting everything to `ms` for reporting latencies.

Although I can often make an educated guess as to the expected latency for an operation, absent experimental data it's only a guess. If I guess wrong, then we could end up in a situation where we guess wrong.

Would this be a topic you think worthwhile to schedule a short call to discuss?

node/grpc/v2_metrics.go
cody-littley: This metric is now removed.

node/grpc/server_v2.go
cody-littley: As discussed offline, this will be left the way it currently is.

node/grpc/v2_metrics.go
cody-littley: As discussed, all of these now use a utility method in common.

jianoaix (approved):

PR #951, opened 2024-12-03
fix: skip socket registry in contract bindings

hopeyen:

Don't error out when the socket registry is not deployed

! This should be reverted once the socket registry contract is deployed

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/eth/reader.go
+++ b/core/eth/reader.go
@@ -50,6 +50,7 @@ type Reader struct {
 }
 
 var _ core.Reader = (*Reader)(nil)
+var socketRegisterEnabled = false
 
 func NewReader(
 	logger logging.Logger,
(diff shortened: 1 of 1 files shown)


core/eth/reader.go
```diff
@@ -168,16 +168,18 @@ func (t *Reader) updateContractBindings(blsOperatorStateRetrieverAddr, eigenDASe
...
 	if err != nil {
-		t.logger.Error("Failed to fetch SocketRegistry address", "err", err)
-		return err
-	}
-
-	contractSocketRegistry, err := socketreg.NewContractSocketRegistry(socketRegistryAddr, t.ethClient)
-	if err != nil {
-		t.logger.Error("Failed to fetch SocketRegistry contract", "err", err)
-		return err
+		//TODO: Temporarily warn, but continue without the socket registry
```
jianoaix, line 174: Can you use a flag to control these?

hopeyen: I could! Although it looks like I would need to update NewReader to take in a flag, and that requires updating flags for a lot of components like the controller, retriever, dataapi, apiserver, disperser, ... Do you think it is still the best way to process?

jianoaix: It looks the easier way is to hold off the socket registry change until the contract is live. Is there anything blocked by not merging the change?

hopeyen: I agree that's the easier way. To my understanding this PR unblocks this week's release but the contract is going live at the end of the sprint. 
Alternatively, we revert the changes that adds the socket registry and go ahead with the release; but then we wanted to do decentralized retrieval this sprint, and devops is using semvar to check correctness of the socket registry (migration from thegraph indexer)

jianoaix: I was talking about not merging previous PR which added in this part of code.
It sounds nothing is blocked by this PR; all what you mentioned I'd say the blocker is the the deployment of the contract.

jianoaix: A more appropriate way seems still to be some CLI flags, maybe defined similar to the graph indexer with default to false, passed in as a config etc. It'd be a standard approach as we add other contracts in the future.

To unblock the release for now, what if you add a local flag (not a CLI flag), to control flow here and the below GetOperatorSocket? Then when you actually ready to enable it, there is only one place to change, instead of editing code around.

hopeyen: yep, that sounds great. I added a local variable `socketRegistryEnabled` and used that for control flow here and `GetOperatorSocket`

jianoaix (approved):

pschork (approved):

PR #953, opened 2024-12-03
v2 Retrieval Client

ian-shim:
Implements a client that queries chunks from DA nodes and reconstructs the blob

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/retrieval_client_v2.go
+++ b/api/clients/retrieval_client_v2.go
@@ -1,0 +1,205 @@
+package clients
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	grpcnode "github.com/Layr-Labs/eigenda/api/grpc/node/v2"
+	"github.com/Layr-Labs/eigenda/core"
+	corev2 "github.com/Layr-Labs/eigenda/core/v2"
+	"github.com/Layr-Labs/eigenda/encoding"
+	"github.com/Layr-Labs/eigensdk-go/logging"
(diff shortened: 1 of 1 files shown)


api/clients/retrieval_client_v2.go
```diff
@@ -0,0 +1,204 @@
...
+type retrievalClientV2 struct {
+	logger            logging.Logger
+	ethClient         core.Reader
+	indexedChainState core.IndexedChainState
+	verifier          encoding.Verifier
+	numConnections    int
+}
+
+// NewRetrievalClientV2 creates a new retrieval client.
+func NewRetrievalClientV2(
```
cody-littley, line 34: This fetches the data directly from the DA nodes. Should we first attempt to fetch the data from the relays? Or is this the job of another layer of the system?

cody-littley (approved):

api/clients/retrieval_client_v2.go
ian-shim: > This fetches the data directly from the DA nodes. Should we first attempt to fetch the data from the relays? Or is this the job of another layer of the system?

Yes, that is done using the relay client. This client is meant to be the client used to retrieve from the DA nodes. 
Whatever is integrating retrieval should choose to integrate using one or both of these clients.

ian-shim: Clarified this in the comment

PR #960, opened 2024-12-05
[v2] Offchain integrations with v2 contracts

ian-shim:
Bindings + client methods to get latest blob version and relay key

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


cody-littley (approved):
Not a blocker, more of a process question.

I'm always a bit suspicious of automatically generated code that we manually have to update. Would it be plausible (as a future change) to ensure that the bindings are automatically generated? We could perhaps use a CI pipeline strategy similar to how we currently enforce that the proto files get properly generated.

jianoaix (approved):

ian-shim:
> Not a blocker, more of a process question.
> 
> I'm always a bit suspicious of automatically generated code that we manually have to update. Would it be plausible (as a future change) to ensure that the bindings are automatically generated? We could perhaps use a CI pipeline strategy similar to how we currently enforce that the proto files get properly generated.

I agree. Right now, given that offchain codebase directly imports the bindings, it would be very risky to automatically generate the bindings following contract updates. So I think bindings should be generated/integrated manually at this point. 

imo, the right way to deal with this is to have a separate repo for contracts. This would allow contract development to make progress independently with separate release process, and offchain code can update dependency whenever it's ready. Then we can have bindings autogenerated in the contracts repo

jianoaix:
> > Not a blocker, more of a process question.
> > I'm always a bit suspicious of automatically generated code that we manually have to update. Would it be plausible (as a future change) to ensure that the bindings are automatically generated? We could perhaps use a CI pipeline strategy similar to how we currently enforce that the proto files get properly generated.
> 
> I agree. Right now, given that offchain codebase directly imports the bindings, it would be very risky to automatically generate the bindings following contract updates. So I think bindings should be generated/integrated manually at this point.
> 
Any examples of this risk?

> imo, the right way to deal with this is to have a separate repo for contracts. This would allow contract development to make progress independently with separate release process, and offchain code can update dependency whenever it's ready. Then we can have bindings autogenerated in the contracts repo

ian-shim:
> Any examples of this risk?

If we generate bindings automatically on contract changes without corresponding offchain updates, offchain code ends up importing the new bindings which may be no longer compatible

pschork (approved):

PR #965, opened 2024-12-05
payment struct and names ReservationPeriod and Salt

hopeyen:

Rename BinIndex to ReservationPeriod
Add Salt uint16 field to PaymentMetadata so that the same blob can be dispersed multiple times within a single reservation period (by assigning different blob key)

Rebase and merge after https://github.com/Layr-Labs/eigenda/pull/928

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


core/data.go
```diff
@@ -554,61 +562,57 @@ func (pm *PaymentMetadata) UnmarshalDynamoDBAttributeValue(av types.AttributeVal
...
 // ConvertToProtoPaymentHeader converts a PaymentMetadata to a protobuf payment header
 func ConvertToPaymentMetadata(ph *commonpb.PaymentHeader) *PaymentMetadata {
 	return &PaymentMetadata{
 		AccountID:         ph.AccountId,
-		BinIndex:          ph.BinIndex,
+		ReservationPeriod: ph.ReservationPeriod,
 		CumulativePayment: new(big.Int).SetBytes(ph.CumulativePayment),
+		Salt:              ph.Salt,
 	}
 }
```
ian-shim, line 611: There are two sets of methods to convert from/to protobuf (there are `ToProtobuf` & `ConvertPaymentHeader` above). Let's just keep 1

api/clients/accountant_test.go
```diff
@@ -271,7 +272,7 @@ func TestConcurrentBinRotationAndAccountBlob(t *testing.T) {
 			defer wg.Done()
 			// for j := 0; j < 5; j++ {
```
ian-shim, line 273: nit: remove these comments?

api/clients/accountant_test.go
```diff
@@ -271,7 +272,7 @@ func TestConcurrentBinRotationAndAccountBlob(t *testing.T) {
...
 			// fmt.Println("request ", i)
-			_, err := accountant.AccountBlob(ctx, 100, quorums)
+			_, err := accountant.AccountBlob(ctx, 100, quorums, salt)
 			assert.NoError(t, err)
 			time.Sleep(500 * time.Millisecond)
```
ian-shim, line 277: Do we need this sleep? Sleep in tests aren't reliable/guaranteed

api/clients/accountant_test.go
```diff
@@ -311,33 +312,33 @@ func TestAccountBlob_ReservationWithOneOverflow(t *testing.T) {
 	now := time.Now().Unix()
 
 	// Okay reservation
```
ian-shim, line 314: do we need equality assertions on salt values?

api/proto/disperser/disperser.proto
```diff
@@ -99,7 +99,7 @@ message DispersePaidBlobRequest {
 	// The quorums to which the blob to be sent
 	repeated uint32 quorum_numbers = 2;
 	
-	// Payment header contains AccountID, BinIndex, and CumulativePayment
+	// Payment header contains AccountID, ReservationPeriod, and CumulativePayment
```
ian-shim, line 102: nit: add salt?

api/proto/common/common.proto
```diff
@@ -20,6 +20,7 @@ message BlobCommitment {
 
 message PaymentHeader {
```
ian-shim, line 21: nit: can we document what these fields mean

core/data.go
```diff
@@ -554,61 +562,57 @@ func (pm *PaymentMetadata) UnmarshalDynamoDBAttributeValue(av types.AttributeVal
...
-	binIndex, err := strconv.ParseUint(m.Value["BinIndex"].(*types.AttributeValueMemberN).Value, 10, 32)
+	reservationPeriod, err := strconv.ParseUint(m.Value["ReservationPeriod"].(*types.AttributeValueMemberN).Value, 10, 32)
 	if err != nil {
-		return fmt.Errorf("failed to parse BinIndex: %w", err)
+		return fmt.Errorf("failed to parse ReservationPeriod: %w", err)
 	}
-	pm.BinIndex = uint32(binIndex)
+	pm.ReservationPeriod = uint32(reservationPeriod)
 	pm.CumulativePayment, _ = new(big.Int).SetString(m.Value["CumulativePayment"].(*types.AttributeValueMemberN).Value, 10)
 	return nil
```
ian-shim, line 571: do we need to add salt?

core/meterer/meterer.go
```diff
@@ -21,7 +21,7 @@ type Config struct {
 }
 
 // Meterer handles payment accounting across different accounts. Disperser API server receives requests from clients and each request contains a blob header
-// with payments information (CumulativePayments, BinIndex, and Signature). Disperser will pass the blob header to the meterer, which will check if the
+// with payments information (CumulativePayments, ReservationPeriod, and Signature). Disperser will pass the blob header to the meterer, which will check if the
```
ian-shim, line 24: add salt?

core/v2/types.go
```diff
@@ -111,7 +111,7 @@ func BlobHeaderFromProtobuf(proto *commonpb.BlobHeader) (*BlobHeader, error) {
 
 	paymentMetadata := core.PaymentMetadata{
```
ian-shim, line 112: salt?

actually, we can use the helper method in core/data here

api/clients/accountant_test.go
hopeyen: I wanted to simulate a consistent dispersal scenario across multiple reservation periods. Would this sit better with inabox tests?

core/data.go
hopeyen: Do we need to store salt? I was thinking if salt is different then the blobHeaderHash would be different and we don't need to explicitly store it anywhere?

api/clients/accountant_test.go
hopeyen: I added some assertions for coverage but didn't add it throughout, lmk if you think that will be useful

api/clients/accountant_test.go
ian-shim: If each call is made 500ms apart, are these calls actually concurrent? 
One way to remove sleep would be to take `now` as a parameter in `AccountBlob` instead of setting `now := time.Now().Unix()` in the method itself.

core/meterer/util.go
```diff
@@ -19,7 +19,7 @@ func CreateReservationTable(clientConfig commonaws.ClientConfig, tableName strin
 				AttributeType: types.ScalarAttributeTypeS,
 			},
 			{
-				AttributeName: aws.String("BinIndex"),
+				AttributeName: aws.String("ReservationPeriod"),
```
ian-shim, line 22: cc @pschork @dmanc for any infra changes

api/clients/accountant_test.go
hopeyen: You are right, I removed this sleep and the tests still passes:)

core/data.go
hopeyen: Okay cool, I added Salt as part of this unmarshal function. If we decide to store salt, I can add it in a subsequent PR since it doesn't seem necessary for now

ian-shim: sg lets add it in a follow up PR

ian-shim (approved):

PR #966, opened 2024-12-06
Support blob and batch fetch APIs

jianoaix:
Implementing dataapi v2 (https://github.com/Layr-Labs/eigenda/pull/955)

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/dataapi/docs/docs.go
+++ b/disperser/dataapi/docs/docs.go
@@ -15,6 +15,52 @@ const docTemplate = `{
     "host": "{{.Host}}",
     "basePath": "{{.BasePath}}",
     "paths": {
+        "/feed/batches/{batch_header_hash}": {
+            "get": {
+                "produces": [
+                    "application/json"
+                ],
+                "tags": [
+                    "Feed"
+                ],
+                "summary": "Fetch batch by the batch header hash",
+                "parameters": [
+                    {
+                        "type": "string",
+                        "description": "Batch header hash in hex string",
+                        "name": "batch_header_hash",
+                        "in": "path",
+                        "required": true
+                    }
+                ],
+                "responses": {
+                    "200": {
+                        "description": "OK",
+                        "schema": {
+                            "$ref": "#/definitions/dataapi.BlobResponse"
+                        }
+                    },
+                    "400": {
+                        "description": "error: Bad request",
+                        "schema": {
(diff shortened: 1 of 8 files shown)


disperser/dataapi/docs/docs.go
```diff
@@ -750,6 +796,27 @@ const docTemplate = `{
...
+                "account_id": {
+                    "description": "AccountID is the ETH account address for the payer",
+                    "type": "string"
+                },
+                "bin_index": {
+                    "description": "BinIndex represents the range of time at which the dispersal is made",
+                    "type": "integer"
+                },
+                "cumulative_payment": {
+                    "description": "TODO: we are thinking the contract can use uint128 for cumulative payment,\nbut the definition on v2 uses uint64. Double check with team.",
```
pschork, line 811: Do we really want this in the swagger doc?

jianoaix: It's the cost of reusing data types in "core", we otherwise have to duplicate it in "dataapi". I may do a manual fix on it to handle this as an exception.

disperser/dataapi/docs/swagger.json
```diff
@@ -136,7 +182,7 @@
                     "200": {
                         "description": "OK",
                         "schema": {
-                            "$ref": "#/definitions/dataapi.BlobMetadataResponse"
+                            "$ref": "#/definitions/dataapi.BlobResponse"
```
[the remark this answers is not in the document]
ian-shim: Hmm, is the base url not taken into account here? 
Anyway, that makes sense to me

ian-shim (approved):
lgtm

disperser/dataapi/docs/docs.go
jianoaix: I'm going to leave it for now until all APIs are implemented, otherwise each time we generate docs it will involve this manual edit.

disperser/dataapi/docs/swagger.json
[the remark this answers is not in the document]
jianoaix: Yea maybe adding "v1" / "v2" to the path will avoid this. But it seems more logical to have "/blob/..." or "/batch/..." than organizing around "/feed/...".

PR #972, opened 2024-12-09
Implement v2 client GET functionality

litt3:
## Why are these changes needed?

This PR creates a new `EigenDAClientV2`, and implements GET functionality for the new client. Additional client functionality will be implemented in upcoming PRs.

## Checks

- [X] I've made sure the tests are passing. Note that there might be a few flaky tests, in that case, please comment that they are not relevant.
- [X] I've checked the new test coverage and the coverage percentage didn't drop.
- Testing Strategy
   - [X] Unit tests
   - [ ] Integration tests
   - [ ] This PR is not tested :(

PR #976, opened 2024-12-10
Write TestRandom util

litt3:

- The existing random test utils in `common/testutils/test_utils.go` operate by seeding global rand. This isn't ideal, as it doesn't provide a source of randomness that can be passed into objects during testing.
- This PR creates a new utility class, which provides the same random utility methods from `test_utils.go`, but with a non-global instance of random.
- If this PR merges, I will open subsequent PRs to migrate usage of newly deprecated test methods.
- I created this new utility to provide functionality necessary to address [this comment](https://github.com/Layr-Labs/eigenda/pull/972#discussion_r1876903049) in an ongoing PR.
- I've discussed this approach with @cody-littley.

## Checks

- Testing Strategy
   - [X] Unit tests
   - [ ] Integration tests


common/testutils/test_random.go
```diff
@@ -0,0 +1,108 @@
+package testutils
```
samlaf, line 1: How about putting this in another package, like `testutils/random` instead?

common/testutils/test_random.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"fmt"
+	"math/rand"
+	"time"
+)
+
+// TestRandom provides all the functionality of math/rand.Rand, plus additional randomness functionality useful for testing
+// This struct wraps an instance of rand.Rand, and directly delegates all standard random methods to this internal instance
+type TestRandom struct {
+	internalRand *rand.Rand
```
samlaf, line 12:
```suggestion
	*rand.Rand
```

samlaf: Use struct embedding instead! https://gobyexample.com/struct-embedding
This way you automatically "inherit" all the methods of rand.Rand, and don't need to write all those boilerplate proxy methods.

common/testutils/test_random.go
```diff
@@ -0,0 +1,108 @@
...
+	_, err := r.internalRand.Read(bytes)
+	if err != nil {
+		panic(err)
+	}
+	return bytes
+}
+
+// RandomTime generates a random time.
+func (r *TestRandom) RandomTime() time.Time {
+	return time.Unix(int64(r.internalRand.Int31()), 0)
```
samlaf, line 50: Nit: any reason we use Int31 instead of Int63?
Also not sure how this function is used but might be good to have a `RandomTimeBetween(start, end time.Time)` version? We can add if/when needed though.

samlaf (commented):
Mostly LGTM, but have a few nits

common/testutils/test_random.go
litt3: Nice! Didn't know this was a thing

litt3: [09e6390](https://github.com/Layr-Labs/eigenda/pull/976/commits/09e63904bb2ac75517d8e9ff4f0c25ea31cda94d)

common/testutils/test_random.go
litt3: [09e6390](https://github.com/Layr-Labs/eigenda/pull/976/commits/09e63904bb2ac75517d8e9ff4f0c25ea31cda94d)

common/testutils/test_random.go
litt3: No reason, changed to int63

Agreed `RandomTimeBetween` would likely find usage, I'll leave for a future PR though

[09e6390](https://github.com/Layr-Labs/eigenda/pull/976/commits/09e63904bb2ac75517d8e9ff4f0c25ea31cda94d)

common/testutils/random/test_random.go
```diff
@@ -0,0 +1,60 @@
...
+	fmt.Printf("Random seed: %d\n", seed)
+	return &TestRandom{
+		rand.New(rand.NewSource(seed)),
+	}
+}
+
+// GetRand returns the underlying random instance
+func (r *TestRandom) GetRand() *rand.Rand {
+	return r.Rand
+}
```
samlaf, line 35: nit: getter is not needed with a public field

samlaf (approved):
LGTM!

cody-littley (approved):

common/testutils/random/test_random.go
litt3: [617c2e3](https://github.com/Layr-Labs/eigenda/pull/976/commits/617c2e3b06115f7848a6288df411fbe6b9f55d37)

PR #983, opened 2024-12-11
Add operator state cache to IndexedChainState

dmanc:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/indexer/state.go
+++ b/core/indexer/state.go
@@ -2,16 +2,21 @@ package indexer
 
 import (
 	"context"
+	"encoding/binary"
 	"errors"
 
 	"github.com/Layr-Labs/eigenda/core"
 	"github.com/Layr-Labs/eigenda/indexer"
+	lru "github.com/hashicorp/golang-lru/v2"
 )
+
+const operatorStateCacheSize = 32
 
 type IndexedChainState struct {
 	core.ChainState
 
-	Indexer indexer.Indexer
+	Indexer            indexer.Indexer
+	operatorStateCache *lru.Cache[string, *core.IndexedOperatorState]
 }
 
 var _ core.IndexedChainState = (*IndexedChainState)(nil)
@@ -20,10 +25,15 @@ func NewIndexedChainState(
 	chainState core.ChainState,
 	indexer indexer.Indexer,
 ) (*IndexedChainState, error) {
+	operatorStateCache, err := lru.New[string, *core.IndexedOperatorState](operatorStateCacheSize)
+	if err != nil {
+		return nil, err
+	}
 
 	return &IndexedChainState{
-		ChainState: chainState,
-		Indexer:    indexer,
+		ChainState:         chainState,
+		Indexer:            indexer,
+		operatorStateCache: operatorStateCache,
 	}, nil
 }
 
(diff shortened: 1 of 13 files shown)


core/indexer/state.go
```diff
@@ -2,16 +2,21 @@ package indexer
...
 	"context"
+	"encoding/binary"
 	"errors"
 
 	"github.com/Layr-Labs/eigenda/core"
 	"github.com/Layr-Labs/eigenda/indexer"
+	lru "github.com/hashicorp/golang-lru/v2"
 )
 
+const operatorStateCacheSize = 32
```
dmanc, line 13: Do we want to make it configurable?

ian-shim (commented):
looks good to me!
waiting for tests

core/thegraph/state.go
```diff
@@ -365,3 +392,13 @@ func convertIndexedOperatorInfoGqlToIndexedOperatorInfo(operator *IndexedOperato
 		Socket:   string(operator.SocketUpdates[0].Socket),
 	}, nil
 }
+
+// Computes a cache key for the operator state cache. The cache key is a
+// combination of the block number and the quorum IDs. Note: the order of the
```
dmanc, line 397: I don't know if this is a problem but noticed that the cache key is dependent on the order you input the quorum ids.

core/thegraph/config.go
```diff
@@ -39,15 +41,22 @@ func CLIFlags(envPrefix string) []cli.Flag {
 			Value:  5,
 			EnvVar: common.PrefixEnvVar(envPrefix, "GRAPH_MAX_RETRIES"),
 		},
+		cli.IntFlag{
+			Name:   OperatorStateCacheSize,
+			Usage:  "The size of the operator state cache in elements (0 to disable)",
+			Value:  0,
```
dmanc, line 47: Should default be 0 or 32?

core/thegraph/state.go
jianoaix: I think it's a unordered set, so you may eliminate the ordering effect by sorting the them

core/thegraph/config.go
jianoaix: What's the estimated size bytes for 32 entries and how many instances of this struct is expected to be created? If it's small enough it should be fine to enable by default

dmanc: Depends on the amount of operators but let's assume 200. The order of magnitude is ~1-10mb.

core/thegraph/state.go
dmanc: Sounds good I'll add sorting.

core/thegraph/state.go
```diff
@@ -124,6 +138,14 @@ func (ics *indexedChainState) Start(ctx context.Context) error {
 }
 
 func (ics *indexedChainState) GetIndexedOperatorState(ctx context.Context, blockNumber uint, quorums []core.QuorumID) (*core.IndexedOperatorState, error) {
+	// Check if the indexed operator state has been cached
+	cacheKey := computeCacheKey(blockNumber, quorums)
```
dmanc, line 142: If we are using cache then we need to assume that the blockNumber has been finalized already. I believe all users of this function would satisfy that assumption since they're passing in a reference block number.

core/thegraph/config.go
jianoaix: Actually, it seems this cache is not really useful:
- At different components, it different instances been created so they don't share the cache
- Within same components, like controller, it may not need to access same blockNumber many times, only when it creates batch (https://github.com/Layr-Labs/eigenda/blob/master/disperser/controller/dispatcher.go#L294), and then the blockNumber will advance

Where does the cache help?

disperser/cmd/batcher/main.go
```diff
@@ -219,7 +222,7 @@ func RunBatcher(ctx *cli.Context) error {
 		if err != nil {
 			return err
 		}
-		ics, err = coreindexer.NewIndexedChainState(cs, indexer)
+		ics, err = coreindexer.NewIndexedChainState(cs, indexer, 0)
```
gpsanant, line 225: y 0?

gpsanant: in other places too, ctrl f

PR #984, opened 2024-12-11
Disperser auth

cody-littley:

Authenticate `StoreChunks()` requests.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


node/auth/authenticator.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to preload cache: %w", err)
+	}
+
+	return authenticator, nil
+}
+
+func (a *requestAuthenticator) preloadCache(ctx context.Context, now time.Time) error {
+	// this will need to be updated for decentralized dispersers
+	_, err := a.getDisperserKey(ctx, now, 0)
```
ian-shim, line 94: `EigenLabsDisperserID`

core/eth/reader.go
```diff
@@ -916,3 +935,22 @@ func (t *Reader) GetRelayURLs(ctx context.Context) (map[uint32]string, error) {
...
+	address, err := registry.DisperserKeyToAddress(
+		&bind.CallOpts{
+			Context: ctx,
+		},
+		disperserID)
+
+	if err != nil {
+		return gethcommon.Address{}, fmt.Errorf("failed to get disperser address: %w", err)
+	}
+
```
ian-shim, line 954: I don't think the contract will return `err` if the `disperserID` doesn't exist. It will probably return zero address. 
We need to explicitly check this case

node/auth/authenticator.go
```diff
@@ -0,0 +1,172 @@
...
+	origin string,
+	request *grpc.StoreChunksRequest,
+	now time.Time) error {
+
+	if a.isAuthenticationStillValid(now, origin) {
+		// We've recently authenticated this client. Do not authenticate again for a while.
+		return nil
+	}
+
+	key, err := a.getDisperserKey(ctx, now, request.DisperserID)
```
ian-shim, line 113: This would result in an extra ETH call for any new `DisperserID` it sees. 
Since we're already hardcoding the only allowed disperser, does it make sense if we consider a request invalid if `DisperserID` is unknown (not EigenLabsDisperserID) in `isAuthenticationStillValid`?

ian-shim (commented):
lgtm, few comments

disperser/cmd/controller/main.go
```diff
@@ -148,7 +149,20 @@ func RunController(ctx *cli.Context) error {
 			return err
 		}
 	}
-	nodeClientManager, err := controller.NewNodeClientManager(config.NodeClientCacheSize, logger)
+
+	var requestSigner clients.RequestSigner
+	if !config.DisperserStoreChunksSigningDisabled {
```
anupsv, line 154: if the store chunks signing is disabled, we should probably throw a warning on the start up logs.

node/auth/authenticator.go
```diff
@@ -0,0 +1,172 @@
...
+	// reloaded from the chain state in case the key has been changed.
+	keyTimeoutDuration time.Duration
+
+	// authenticatedDispersers is a set of disperser addresses that have been recently authenticated, mapped
+	// to the time when that cached authentication will expire.
+	authenticatedDispersers *lru.Cache[string, time.Time]
+
+	// authenticationTimeoutDuration is the duration for which an auth is valid.
+	// If this is zero, then auth saving is disabled, and each request will be authenticated independently.
+	authenticationTimeoutDuration time.Duration
```
anupsv, line 54: Let's keep this value low for the beginning.

api/clients/v2/request_signer.go
```diff
@@ -0,0 +1,62 @@
...
+	"fmt"
+	grpc "github.com/Layr-Labs/eigenda/api/grpc/node/v2"
+	"github.com/Layr-Labs/eigenda/common"
+	"github.com/Layr-Labs/eigenda/node/auth"
+	"github.com/aws/aws-sdk-go-v2/aws"
+	"github.com/aws/aws-sdk-go-v2/service/kms"
+)
+
+// RequestSigner encapsulates the logic for signing GetChunks requests.
+type RequestSigner interface {
```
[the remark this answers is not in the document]
cody-littley: renamed to `DispersalRequestSigner`

api/clients/v2/node_client.go
```diff
@@ -60,16 +62,28 @@ func (c *nodeClient) StoreChunks(ctx context.Context, batch *corev2.Batch) (*cor
...
+	request := &nodegrpc.StoreChunksRequest{
 		Batch: &commonpb.Batch{
 			Header: &commonpb.BatchHeader{
 				BatchRoot:            batch.BatchHeader.BatchRoot[:],
 				ReferenceBlockNumber: batch.BatchHeader.ReferenceBlockNumber,
 			},
 			BlobCertificates: blobCerts,
 		},
-	})
+		DisperserID: 0, // this will need to be updated dispersers are decentralized
```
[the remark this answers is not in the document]
cody-littley:
```
// EigenLabsDisperserID is the ID of the disperser that is managed by Eigen Labs.
const EigenLabsDisperserID = 0
```

node/auth/authenticator.go
cody-littley:
```
	var defaultAddress gethcommon.Address
	if err != nil {
		return defaultAddress, fmt.Errorf("failed to get disperser address: %w", err)
	}
	if address == defaultAddress {
		return defaultAddress, fmt.Errorf("disperser with id %d not found", disperserID)
	}
```

disperser/cmd/controller/main.go
cody-littley: var requestSigner clients.DispersalRequestSigner
	if config.DisperserStoreChunksSigningDisabled {
		logger.Warn("StoreChunks() signing is disabled")
	} else {
			context.Background(),
			config.AwsClientConfig.Region,
			config.AwsClientConfig.EndpointURL,
			config.DisperserKMSKeyID)
		if err != nil {
			return fmt.Errorf("failed to create request signer: %v", err)
		}
	}

node/auth/authenticator.go
cody-littley: I've set this to be one minute by default. Is that sufficient, or should we lower it further?

api/clients/v2/node_client.go
```diff
@@ -24,21 +25,23 @@ type NodeClient interface {
 }
 
 type nodeClient struct {
-	config   *NodeClientConfig
-	initOnce sync.Once
-	conn     *grpc.ClientConn
+	config        *NodeClientConfig
+	initOnce      sync.Once
+	conn          *grpc.ClientConn
+	requestSigner DispersalRequestSigner
```
jianoaix, line 31: A slightly different structuring of it: shall we pass in an interface, not a concrete implementation?
Like we have RequestSigners interface (and passed around), but with a KMS based implementation. This leaves space for potential non-KMS (i.e. non AWS specific) options (in a decentralized scenario it may be needed).

api/clients/v2/request_signer.go
```diff
@@ -0,0 +1,62 @@
+package clients
```
ian-shim, line 1: nit: we can update the file name as well?

core/eth/reader.go
```diff
@@ -916,3 +935,26 @@ func (t *Reader) GetRelayURLs(ctx context.Context) (map[uint32]string, error) {
...
+	registry := t.bindings.DisperserRegistry
+	if registry == nil {
+		return gethcommon.Address{}, errors.New("disperser registry not deployed")
+	}
+
+	address, err := registry.DisperserKeyToAddress(
+		&bind.CallOpts{
+			Context: ctx,
+		},
+		disperserID)
```
jianoaix, line 949: Can we settle on either ID or Key? Using both is unnecessary and adds a little bit confusion.
(if this is meant to be an int, ID seems fit)

node/auth/authenticator.go
ian-shim: I think this reply is for another comment in `reader`.

node/auth/authenticator.go
```diff
@@ -0,0 +1,172 @@
...
+	expiration time.Time
+}
+
+var _ RequestAuthenticator = &requestAuthenticator{}
+
+type requestAuthenticator struct {
+	// chainReader is used to read the chain state.
+	chainReader core.Reader
+
+	// keyCache is used to cache the public keys of dispersers.
```
jianoaix, line 41: what public key has only 32 bits?

node/auth/authenticator.go
```diff
@@ -0,0 +1,172 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to get operator key: %w", err)
+	}
+
+	err = VerifyStoreChunksRequest(*key, request)
+	if err != nil {
+		return fmt.Errorf("failed to verify request: %w", err)
+	}
+
+	a.saveAuthenticationResult(now, origin)
```
jianoaix, line 123: nit: `cacheAuthenticationResult`, "save" feels like it's persisting the data which it's not

node/auth/authenticator.go
```diff
@@ -0,0 +1,172 @@
...
+	disperserID uint32) (*gethcommon.Address, error) {
+	key, ok := a.keyCache.Get(disperserID)
+	if ok {
+		expirationTime := key.expiration
+		if now.Before(expirationTime) {
+			return &key.key, nil
+		}
+	}
+
+	address, err := a.chainReader.GetDisperserAddress(ctx, disperserID)
```
jianoaix, line 140: I'd cache this ID to address mapping to save RPCs

node/flags/flags.go
```diff
@@ -238,6 +238,33 @@ var (
 		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "CHUNK_DOWNLOAD_TIMEOUT"),
 		Value:    20 * time.Second,
 	}
+	DisableDispersalAuthenticationFlag = cli.BoolFlag{
```
jianoaix, line 241: Why disabling is an option? shouldn't it be non optional?

node/auth/request_signing.go
jianoaix, line 1: This file might be better fit in `api/clients`

api/clients/v2/node_client.go
```diff
@@ -3,6 +3,7 @@ package clients
 import (
 	"context"
 	"fmt"
+	"github.com/Layr-Labs/eigenda/node/auth"
```
jianoaix, line 6: It doesn't seem to me this is a good dependency direction for api/ to depend on node/

node/auth/authenticator.go
```diff
@@ -0,0 +1,172 @@
...
+}
+
+// saveAuthenticationResult saves the result of an auth.
+func (a *requestAuthenticator) saveAuthenticationResult(now time.Time, origin string) {
+	if a.authenticationTimeoutDuration == 0 {
+		// Authentication saving is disabled.
+		return
+	}
+
+	a.authenticatedDispersers.Add(origin, now.Add(a.authenticationTimeoutDuration))
```
ian-shim, line 160: are we actually going to use this feature? 
nodes will be receiving batches very frequently from the centralized disperser, which means if an exploiter sends dispersal request with EigenDADisperserID, it will almost certainly bypass the authentication. 
Not sure if this optimization is meaningful

node/auth/authenticator.go
cody-littley: The `uint32` is the relay ID. These are serial numbers, not eth keys. Currently we have exactly one disperser, and its ID is hard coded to be `uint32(0)`.

node/auth/authenticator.go
cody-littley: This is cached (see `keyCache *lru.Cache[uint32, *keyWithTimeout]`). By design, we re-fetch the same key again after a timeout just in case the key was changed. But not for each RPC.

node/flags/flags.go
cody-littley: It's enabled by default. Necessary to get things working in inabox. Inabox integration is kind of complex, so I'd like to merge this code with it disabled in the e2e test, and to enable it in a follow up PR.

node/auth/request_signing.go
cody-littley: I've moved all GRPC hashing code to a new package `api/hashing`. There is now no longer a dependency (outside of test code) between the client and this file.

api/clients/v2/node_client.go
cody-littley: this dependency is no longer present

node/auth/authenticator.go
cody-littley: renamed

core/eth/reader.go
cody-littley: I'm happy with calling this an "ID". These are serial numbers. Eventually when we allow the community to register dispersers, we will allocate these in monotonic order via a smart contract.

Where do you see the terminology "key" being used in this context?

api/clients/v2/node_client.go
cody-littley: I'm a little confused about this comment. `DispersalRequestSigner` is already an interface. Is the interface ok in its current form?

```
// DispersalRequestSigner encapsulates the logic for signing GetChunks requests.
type DispersalRequestSigner interface {
	// SignStoreChunksRequest signs a StoreChunksRequest. Does not modify the request
	// (i.e. it does not insert the signature).
	SignStoreChunksRequest(ctx context.Context, request *grpc.StoreChunksRequest) ([]byte, error)
}
```

api/clients/v2/request_signer.go
cody-littley: Oops, forgot to rename the file. 😅 Fixed.

node/auth/authenticator.go
cody-littley: Oops, you're correct @ian-shim, wires got crossed here. 🙃 

The response I **intended** to type here  was that the code does cache this value inside this cache: `keyCache *lru.Cache[uint32, *keyWithTimeout]`.

core/eth/reader.go
cody-littley: Ah, I think I see what you are getting at. There is a method in the chain reader that references the disperser key:

```
getDisperserKey(
	ctx context.Context,
	now time.Time,
	disperserID uint32) (*gethcommon.Address, error)
```

The disperser key and the disperser ID are actually distinct things.

- Disperser ID: a unique identifier for the disperser, a 32 bit serial number
- Disperser (public) key: the public key used to verify `StoreChunks()` requests from the disperser, is an eth address

The reason why the disperser's public key is not a good identifier for the disperser is that we want the capability to re-key the disperser (e.g. if we lose the key, it gets compromised, or we just want to rotate it).

jianoaix: Ok, seems like the “operator address" v.s. "operator ID" case.

node/auth/authenticator.go
jianoaix: Can you fix the comment then?

core/eth/reader.go
jianoaix: Then why `DisperserKeyToAddress` is accepting a disperserID as param?

jianoaix (commented):
lgtm

core/eth/reader.go
cody-littley: There is a map from disperser IDs (serial numbers) to disperser public keys (eth addresses). This function looks up the address given the disperser's ID. Until we decentralize dispersers, there will always be exactly one disperser ID: `0`. 

The reason why we decided to add this complexity now is that we wanted the ability to change the public key in case we need to rotate the disperser's keys. The disperser ID part was super simple to have, and will mean that we don't need to change this contract in the future when we move to decentralized dispersers.

node/auth/authenticator.go
cody-littley: Updated comments to clarify:

```
// keyCache is used to cache the public keys of dispersers. The uint32 map keys are disperser IDs. Disperser
// IDs are serial numbers, with the original EigenDA disperser assigned ID 0. The map values contain
// the public key of the disperser and the time when the local cache of the key will expire.
keyCache *lru.Cache[uint32 /* disperser ID */, *keyWithTimeout]
```

api/clients/v2/dispersal_request_signer.go
```diff
@@ -0,0 +1,62 @@
+package clients
+
+import (
+	"context"
+	"crypto/ecdsa"
+	"fmt"
+	grpc "github.com/Layr-Labs/eigenda/api/grpc/node/v2"
+	"github.com/Layr-Labs/eigenda/api/hashing"
+	"github.com/Layr-Labs/eigenda/common"
+	"github.com/aws/aws-sdk-go-v2/aws"
```
ian-shim, line 10: nit: I'd prefer to have a thin kms wrapper client and use it to interface with kms instead of directly importing aws libraries. But feel free to ignore

common/kms.go
```diff
@@ -0,0 +1,167 @@
+package common
```
ian-shim, line 1: aws related utils are under `common/aws/`

node/auth/authenticator.go
ian-shim: The case I was worried about is an exploit with sending requests with  increasing disperser IDs

common/kms.go
cody-littley: Moved to `common/aws`

api/clients/v2/dispersal_request_signer.go
cody-littley: This class is already just a thin wrapper, although its use case specific and not general purpose. Let's discuss this and make any changes in a follow up PR.

node/auth/authenticator.go
cody-littley: Ah, I see what you mean. You don't need to send increasing disperser IDs, since nothing is cached when a non-existent disperser ID is queried.

The core purpose of this caching was to reduce the latency of this RPC. The latency of an invalid request is not really that important. Even if reading the chain state is not particularly expensive, I agree that it's generally good practice to avoid doing as much as  possible for invalid RPCs.

The solution in the immediate term is to simply reject any disperser ID other than `0`. I've made that change.

When we eventually have multiple dispersers, we will need to add a function to the smart contract that returns a list of all currently valid disperser IDs. We can cache this list and refresh it every X seconds, and always reject any request for a disperser that is unknown. Unfortunately this change will not be backwards comparable, meaning that node operators will have a required upgrade when we enable multiple dispersers. Not the end of the world, but not much we can do to avoid it now I think (short of updating contracts again).

jianoaix (approved):

PR #986, opened 2024-12-11
Separate out client v2 package

ian-shim:
Instead of suffixing `v2`, create a separate `v2` package under `clients`

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/clients/mock/v2/node_client.go
```diff
@@ -1,9 +1,9 @@
-package mock
+package v2
```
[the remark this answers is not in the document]
ian-shim: sg, done!

litt3 (approved):

api/clients/v2/disperser_client.go
```diff
@@ -1,4 +1,4 @@
-package clients
+package v2
```
samlaf, line 1: How about keeping the package name `clients` here?
Its typical for golang modules to use an "invisible" version suffix (this is actually part of how golang's algo does version management). See for eg this urfave example: https://github.com/urfave/cli/blob/main/examples/example-cli/example-cli.go (imported as cli/v3, but module is just used as `cli.`, not `v3.` which would be awkward).

I'm not sure if its typical to do the same thing for packages, but I feel like it would be cleaner?

api/clients/v2/disperser_client.go
```diff
@@ -16,21 +16,21 @@ import (
...
 
-type DisperserClientV2Config struct {
+type DisperserClientConfig struct {
 	Hostname          string
 	Port              string
 	UseSecureGrpcFlag bool
 }
 
-type DisperserClientV2 interface {
+type DisperserClient interface {
```
samlaf, line 25: If we keep the package name clients, then we could simplify these to `clients.DisperserConfig` and `clients.Disperser`.

api/clients/v2/disperser_client.go
ian-shim: To do this, don't you need to version the entire module (like [here](https://github.com/urfave/cli/blob/main/go.mod#L1) in your example)?

disperser/controller/dispatcher.go
```diff
@@ -4,11 +4,12 @@ import (
...
 	"errors"
 	"fmt"
-	"github.com/prometheus/client_golang/prometheus"
 	"math"
 	"time"
 
-	"github.com/Layr-Labs/eigenda/api/clients"
+	"github.com/prometheus/client_golang/prometheus"
+
+	clients "github.com/Layr-Labs/eigenda/api/clients/v2"
```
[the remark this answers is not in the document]
ian-shim: I don't think this would be a big refactor to do just within the `api` package. 
Yes, it would be more difficult to do this in other packages because there's a lot of sharing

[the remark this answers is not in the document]
samlaf: I think it would be very nice to do, but don't have enough context on everything in the api package, and whether this refactor would make sense with the rest of the code in this repo.

api/clients/v2/disperser_client.go
samlaf: No I'm just saying leave the directory structure the same (`api/clients/v2/.*.go`), but name the package `clients` (package can be named different than the directory).

samlaf (approved):
Approving so we can move on with our lives, although let's keep in mind a future move to versioning the entire api package (api/v2/...).
I'd also still vote for `clients.Disperser` instead of `clients.DisperserClient` but I'm fine if people like the more verbose version.

PR #990, opened 2024-12-12
reservation timestamp check

hopeyen:

Utilize the start and end timestamp to check if the reservation is active. Updated the name from `ActiveReservation` to `ReservedPayment` to make it less confusing and consistent with `OnDemandPayment`

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/data.go
+++ b/core/data.go
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"math/big"
 	"strconv"
-	"time"
 
 	commonpb "github.com/Layr-Labs/eigenda/api/grpc/common"
 	"github.com/Layr-Labs/eigenda/common"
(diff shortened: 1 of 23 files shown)


core/data.go
```diff
@@ -606,7 +607,7 @@ func ConvertToPaymentMetadata(ph *commonpb.PaymentHeader) *PaymentMetadata {
 
 // OperatorInfo contains information about an operator which is stored on the blockchain state,
 // corresponding to a particular quorum
-type ActiveReservation struct {
+type ReservedPayment struct {
 	SymbolsPerSecond uint64 // reserve number of symbols per second
 	//TODO: we are not using start and end timestamp, add check or remove
```
ian-shim, line 612: Is this TODO still relevant? Maybe we can update the documentation here for two timestamps.

Also nit: comments should be line above the fields starting with the name of the field

core/data.go
```diff
@@ -606,7 +607,7 @@ func ConvertToPaymentMetadata(ph *commonpb.PaymentHeader) *PaymentMetadata {
 
 // OperatorInfo contains information about an operator which is stored on the blockchain state,
```
ian-shim, line 608: nit: update the comment?

core/data.go
```diff
@@ -625,3 +626,8 @@ type BlobVersionParameters struct {
 	MaxNumOperators uint32
 	NumChunks       uint32
 }
+
+func (ar *ReservedPayment) IsActive() bool {
+	now := uint64(time.Now().Unix())
```
ian-shim, line 631: do we have unit tests for active/inactive reservations? 
would be useful to take this timestamp as input for testing

core/data.go
```diff
@@ -625,3 +626,8 @@ type BlobVersionParameters struct {
...
+	return ar.StartTimestamp <= now && ar.EndTimestamp >= now
```
ian-shim, line 632: are all `ReservedPayment`s guaranteed to have these timestamps populated?

ian-shim (commented):
lgtm, just few comments

core/data.go
hopeyen: I think yes, because before we make ReservedPayments, they go through the convert function [here](https://github.com/Layr-Labs/eigenda/blob/96215b135230ca5310c3ee7ffcf73491638d8dee/core/eth/utils.go#L132)

ian-shim: If reservation has zero timestamps and non-zero values for other fields, looks like it would still get converted successfully

hopeyen: yes, that is only for valid reservations, but this check should handle checking if the reservation is active or not

hopeyen: I don't think I understand what are your concern here 🤔 ?

ian-shim: My concern is whether it's possible to have a valid reservation that's missing `EndTimestamp` for example.

ian-shim: I guess the current logic would ignore those reservations if they're missing any timestamps

ian-shim (approved):

core/data.go
hopeyen: Yep!

PR #991, opened 2024-12-12
[v2] payments inabox 

hopeyen:

enabling inabox with payments

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


disperser/cmd/apiserver/main.go
```diff
@@ -107,10 +107,29 @@ func RunDisperserServer(ctx *cli.Context) error {
...
 
+		logger.Info("table names",
+			"res",
+			config.ReservationsTableName,
+			"ondemand",
+			config.OnDemandTableName,
+			"global",
+			config.GlobalRateTableName,
+			"config",
+			config.AwsClientConfig,
```
ian-shim, line 122: do we need this logged?

inabox/geth/genesis.json
```diff
@@ -18,126 +18,36 @@
     },
```
ian-shim, line 18: hmmm what caused these changes?

api/clients/v2/disperser_client.go
```diff
@@ -139,16 +138,11 @@ func (c *disperserClient) DisperseBlob(
 		return nil, [32]byte{}, api.NewErrorInternal("uninitialized signer for authenticated dispersal")
 	}
 
-	// TODO(hopeyen): uncomment this after the accountant is implemented
-	// if c.accountant == nil {
```
ian-shim, line 143: hmm shall we keep this check? 
what happens if `initOncePopulateAccountant` fails and client tries to call this method again?

contracts/script/SetUpEigenDA.s.sol
```diff
@@ -10,6 +10,8 @@ import {StakeRegistry} from "eigenlayer-middleware/StakeRegistry.sol";
 import {IIndexRegistry} from "eigenlayer-middleware/interfaces/IIndexRegistry.sol";
```
ian-shim, line 10: is this script only used for testing?
cc @0x0aa0

contracts/script/SetUpEigenDA.s.sol
```diff
@@ -111,9 +128,16 @@ contract SetupEigenDA is EigenDADeployer, EigenLayerUtils {
             operatorETHAmounts[i] = 5 ether;
         }
 
+        uint256 clientPrivateKey = 0x0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcded;
```
ian-shim, line 131: maybe we could use one of the accounts prefunded in anvil?

ian-shim (commented):
lgtm, couple comments

api/clients/v2/disperser_client.go
hopeyen: hmm I guess if the client tries this method again after the init failed, all subsequent calls will return the same error from the first attempt. I think it would be the same behavior as adding this check back?

contracts/script/SetUpEigenDA.s.sol
hopeyen: I realized that I could send deposits to the client as the owner and there's no need to access client private key here, just the target address:-)

contracts/script/SetUpEigenDA.s.sol
0x0aa0: yes!

ian-shim (approved):

PR #995, opened 2024-12-12
[v2] integrate with new payments contract

hopeyen:

Generated new contract bindings
Updated to corresponding names
Update contract bindings to include payment vault initialization 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


core/eth/reader.go
```diff
@@ -210,6 +210,18 @@ func (t *Reader) updateContractBindings(blsOperatorStateRetrieverAddr, eigenDASe
 		}
 	}
 
+	paymentVaultAddr, err := contractEigenDAServiceManager.PaymentVault(&bind.CallOpts{})
+	if err != nil {
```
ian-shim, line 214: let's not fail and account for the case contract isn't deployed like threshold registry above

core/eth/reader.go
```diff
@@ -691,9 +703,6 @@ func (t *Reader) GetAllVersionedBlobParams(ctx context.Context) (map[uint16]*cor
 }
 
 func (t *Reader) GetReservedPayments(ctx context.Context, accountIDs []gethcommon.Address) (map[gethcommon.Address]*core.ReservedPayment, error) {
-	if t.bindings.PaymentVault == nil {
```
ian-shim, line 694: let's keep them until contract is deployed on mainnet

core/eth/reader.go
hopeyen: okay, edited to just log

ian-shim (approved):

PR #997, opened 2024-12-13
Create codeQL-scanning.yaml

anupsv:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/.github/workflows/codeQL-scanning.yaml
+++ b/.github/workflows/codeQL-scanning.yaml
@@ -1,0 +1,49 @@
+name: "CodeQL scanning"
+
+on:
+  push:
+    branches:
+      - master
+      - 'release/*'
+  pull_request:
+    branches:
+      - main
+      - 'release/*'
+    paths:
+      - 'node/**'
+      - 'operators/**'
+      - 'retriever/**'
+      - 'disperser/**'
+      - 'core/**'
+      - 'common/**'
+      - 'api/**'
+      - '.github/codeql/**'
+      - '.github/workflows/codeql-analysis.yml'
+  schedule:
+    - cron: '0 9 * * 1'
+
+jobs:
+  CodeQL-Build:
+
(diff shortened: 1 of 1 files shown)


.github/workflows/codeQL-scanning.yaml
```diff
@@ -0,0 +1,49 @@
+name: "CodeQL scanning"
+
+on:
+  push:
+    branches:
+      - master
+      - 'release/*'
+  pull_request:
+    branches:
+      - main
```
jianoaix, line 10: There is no `main` branch

.github/workflows/codeQL-scanning.yaml
```diff
@@ -0,0 +1,49 @@
...
+      - 'operators/**'
+      - 'retriever/**'
+      - 'disperser/**'
+      - 'core/**'
+      - 'common/**'
+      - 'api/**'
+      - '.github/codeql/**'
+      - '.github/workflows/codeql-analysis.yml'
+  schedule:
+    - cron: '0 9 * * *'
```
jianoaix, line 23: It's triggered already on push, what this cron test adds?

.github/workflows/codeQL-scanning.yaml
anupsv: we should probably change to main. `master` seems deprecated as the default name.

.github/workflows/codeQL-scanning.yaml
anupsv: adds a cadence of running the codeql scanner everyday at 9 am. Helps keep a regular non PR triggered scan of the code. Idea is to use this for longer and slower codeql scans. Will be adding that soon-ish.

ian-shim (approved):

jianoaix (approved):

PR #1001, opened 2024-12-13
chore: change from pk to random bytes

shrimalmadhur:

- We don't need private key to have randomness for salt. 
- We are creating using `crypto/rand` library. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/node/churner_client.go
+++ b/node/churner_client.go
@@ -2,6 +2,7 @@ package node
 
 import (
 	"context"
+	"crypto/rand"
 	"crypto/tls"
 	"errors"
 	"time"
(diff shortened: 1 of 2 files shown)


node/churner_client.go
```diff
@@ -45,8 +46,12 @@ func (c *churnerClient) Churn(ctx context.Context, operatorAddress string, keyPa
...
 	}
 	// generate salt
-	privateKeyBytes := []byte(keyPair.PrivKey.String())
-	salt := crypto.Keccak256([]byte("churn"), []byte(time.Now().String()), quorumIDs[:], privateKeyBytes)
+	bytes := make([]byte, 32)
+	_, err := rand.Read(bytes)
+	if err != nil {
+		return nil, err
+	}
+	salt := crypto.Keccak256([]byte("churn"), []byte(time.Now().String()), quorumIDs[:], bytes)
```
bxue-l2, line 54: perhaps, we can set bytes = salts. Not sure why do keccak256().

But if leave as what it is, the code is still correct

bxue-l2 (approved):

node/churner_client.go
shrimalmadhur: @gpsanant might know why we do it?

PR #1005, opened 2024-12-16
fix: relay considered sensitive

hopeyen:

https://github.com/Layr-Labs/eigenda/security/code-scanning/18

Not sure why my commit was the source it identified 🤷 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/relay/cmd/config.go
+++ b/relay/cmd/config.go
@@ -106,3 +106,10 @@ func NewConfig(ctx *cli.Context) (Config, error) {
 	}
 	return config, nil
 }
+
+// SanitizedConfig returns a copy of the Config with sensitive information removed or obfuscated.
+func (c Config) SanitizedConfig() Config {
+	sanitized := c
+	sanitized.RelayConfig.AuthenticationKeyCacheSize = 0 // Obfuscate sensitive information
+	return sanitized
+}
--- a/relay/cmd/main.go
+++ b/relay/cmd/main.go
@@ -54,7 +54,7 @@ func RunRelay(ctx *cli.Context) error {
 	if err != nil {
 		return fmt.Errorf("failed to create logger: %w", err)
 	}
-	logger.Info(fmt.Sprintf("Relay configuration: %#v", config))
+	logger.Info(fmt.Sprintf("Relay configuration: %#v", config.SanitizedConfig()))
 
 	dynamoClient, err := dynamodb.NewClient(config.AWS, logger)
 	if err != nil {


relay/cmd/config.go
```diff
@@ -106,3 +106,10 @@ func NewConfig(ctx *cli.Context) (Config, error) {
 	}
 	return config, nil
 }
+
+// SanitizedConfig returns a copy of the Config with sensitive information removed or obfuscated.
+func (c Config) SanitizedConfig() Config {
+	sanitized := c
+	sanitized.RelayConfig.AuthenticationKeyCacheSize = 0 // Obfuscate sensitive information
```
samlaf, line 113: hmm this doesn't seem like a secret. It's just a cache size right? Maybe codeQL is just parsing "AuthenticationKey" and is not smart enough to understand the suffix?

relay/cmd/main.go
```diff
@@ -54,7 +54,7 @@ func RunRelay(ctx *cli.Context) error {
 	if err != nil {
 		return fmt.Errorf("failed to create logger: %w", err)
 	}
-	logger.Info(fmt.Sprintf("Relay configuration: %#v", config))
+	logger.Info(fmt.Sprintf("Relay configuration: %#v", config.SanitizedConfig()))
```
samlaf, line 57: let's make sure that none of the configs printed (including awsConfig, ethConfig, etc) don't leak any secrets, and if nothing is leaked maybe there's a comment we can add to shutup codeQL on this one if its a false positive

relay/cmd/config.go
hopeyen: there is aws `SecretAccessKey` but it isn't triggering anything for codeQL. None other config seems to be a secret. maybe we just add shutup?

samlaf: Let’s not log secretaccesskey and shut up the rest.
Still, for best practices going forward, I’m now a big fan of separating configs and secrets into two separate structs, and only logging configs on startup. This I feel is a safer way to not in the future add more secrets that would be logged and not picked up by the now shut up codeQL.

relay/cmd/main.go
ian-shim: I think this logging was meant for debugging. Maybe we can just log few specific fields instead of the whole struct?

hopeyen: which fields are useful for debugging? we got 
```
RelayIDs []v2.RelayKey, 
GRPCPort int
MaxGRPCMessageSize int
MetadataCacheSize int
MetadataMaxConcurrency int
BlobCacheBytes uint64
BlobMaxConcurrency int
ChunkCacheSize uint64
ChunkMaxConcurrency int
MaxKeysPerGetChunksRequest int
RateLimits limiter.Config
AuthenticationKeyCacheSize int <- codeQL doesn't like this one
AuthenticationTimeout time.Duration
AuthenticationDisabled bool
Timeouts TimeoutConfig
OnchainStateRefreshInterval time.Duration
MetricsPort int
```

PR #1013, opened 2024-12-17
fix: make the api path naming plural

jianoaix:
To be consistent with `/operators/...` (and plural seems more usual in such case)

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/dataapi/server_v2.go
+++ b/disperser/dataapi/server_v2.go
@@ -105,13 +105,13 @@ func (s *ServerV2) Start() error {
 	{
 		blob := v2.Group("/blob")
 		{
-			blob.GET("/blob/feed", s.FetchBlobFeedHandler)
-			blob.GET("/blob/:blob_key", s.FetchBlobHandler)
+			blob.GET("/blobs/feed", s.FetchBlobFeedHandler)
+			blob.GET("/blobs/:blob_key", s.FetchBlobHandler)
 		}
 		batch := v2.Group("/batch")
 		{
-			batch.GET("/batch/feed", s.FetchBatchFeedHandler)
-			batch.GET("/batch/:batch_header_hash", s.FetchBatchHandler)
+			batch.GET("/batches/feed", s.FetchBatchFeedHandler)
+			batch.GET("/batches/:batch_header_hash", s.FetchBatchHandler)
 		}
 		operators := v2.Group("/operators")
 		{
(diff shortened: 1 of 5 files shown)


disperser/dataapi/server_v2.go
```diff
@@ -105,13 +105,13 @@ func (s *ServerV2) Start() error {
...
+			blob.GET("/blobs/:blob_key", s.FetchBlobHandler)
 		}
 		batch := v2.Group("/batch")
 		{
-			batch.GET("/batch/feed", s.FetchBatchFeedHandler)
-			batch.GET("/batch/:batch_header_hash", s.FetchBatchHandler)
+			batch.GET("/batches/feed", s.FetchBatchFeedHandler)
+			batch.GET("/batches/:batch_header_hash", s.FetchBatchHandler)
 		}
 		operators := v2.Group("/operators")
```
jianoaix, line 116: Or we make it `/operator/...` if it's preferred

disperser/dataapi/server_v2_test.go
```diff
@@ -232,9 +232,9 @@ func TestFetchBatchHandlerV2(t *testing.T) {
 	err = blobMetadataStore.PutAttestation(context.Background(), attestation)
 	require.NoError(t, err)
 
-	r.GET("/v2/batch/:batch_header_hash", testDataApiServerV2.FetchBatchHandler)
+	r.GET("/v2/batches/:batch_header_hash", testDataApiServerV2.FetchBatchHandler)
```
dmanc, line 235: nit: Rename functions to match? `FetchBatchHandler` -> `FetchBatchesHandler`?

disperser/dataapi/server_v2.go
```diff
@@ -105,13 +105,13 @@ func (s *ServerV2) Start() error {
 	{
 		blob := v2.Group("/blob")
 		{
-			blob.GET("/blob/feed", s.FetchBlobFeedHandler)
-			blob.GET("/blob/:blob_key", s.FetchBlobHandler)
+			blob.GET("/blobs/feed", s.FetchBlobFeedHandler)
```
dmanc, line 108: nit: Same comment: `FetchBlob` -> `FetchBlobs`?

disperser/dataapi/server_v2.go
dmanc: Think plural is better since the response is for multiple operators right?

dmanc (approved):

disperser/dataapi/server_v2.go
jianoaix: FetchBlob is fetching exactly one blob. The edge case is "FetchBlobsFeed" or "FetchBlobFeed", which I am not sure which one, maybe "blob" is adjective here so "BlobFeed" sounds right?

disperser/dataapi/server_v2.go
jianoaix: It can be for a single operator as well. I think it's like "this is API for dealing with multi operators"  rather than saying it'll just be returning one or multiple

PR #1014, opened 2024-12-17
fix: properly set fields without onchain correspondance

hopeyen:

- do not let server return API errors when some users did not have payment set on-chain
- for the client's accoutant, use default values instead of failing when one of the payments info is missing

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/accountant.go
+++ b/api/clients/v2/accountant.go
@@ -39,6 +39,13 @@ type BinRecord struct {
 	Usage uint64
 }
 
+func DummyBinRecord() *BinRecord {
+	return &BinRecord{
+		Index: 0,
+		Usage: 0,
+	}
+}
+
 func NewAccountant(accountID string, reservation *core.ReservedPayment, onDemand *core.OnDemandPayment, reservationWindow uint32, pricePerSymbol uint32, minNumSymbols uint32, numBins uint32) *Accountant {
 	//TODO: client storage; currently every instance starts fresh but on-chain or a small store makes more sense
 	// Also client is currently responsible for supplying network params, we need to add RPC in order to be automatic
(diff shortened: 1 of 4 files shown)


api/clients/v2/accountant.go
```diff
@@ -159,51 +166,55 @@ func (a *Accountant) SetPaymentState(paymentState *disperser_rpc.GetPaymentState
...
-	}
-	a.reservation.QuorumSplits = quorumSplits
-
-	binRecords := make([]BinRecord, len(paymentState.BinRecords))
-	for i, record := range paymentState.BinRecords {
-		binRecords[i] = BinRecord{
-			Index: record.Index,
-			Usage: record.Usage,
+	}
+
```
ian-shim, line 170: can we document this behavior (which fields are optional and what the consequences are)?

api/clients/v2/accountant.go
```diff
@@ -159,51 +166,55 @@ func (a *Accountant) SetPaymentState(paymentState *disperser_rpc.GetPaymentState
...
+			quorumSplits[i] = uint8(quorum)
 		}
+		a.reservation.QuorumSplits = quorumSplits
 	}
-	a.binRecords = binRecords
 
+	binRecords := make([]BinRecord, len(paymentState.GetBinRecords()))
+	for i, record := range paymentState.GetBinRecords() {
+		if record == nil {
+			binRecords[i] = *DummyBinRecord()
```
ian-shim, line 209: nit: 
```
binRecords[i] = BinRecord{
	Index: 0,
	Usage: 0,
}
```
?

disperser/apiserver/server_v2.go
```diff
@@ -254,20 +254,22 @@ func (s *DispersalServerV2) GetPaymentState(ctx context.Context, req *pb.GetPaym
...
 	if err != nil {
-		return nil, api.NewErrorNotFound("failed to get largest cumulative payment")
+		s.logger.Debug("failed to get largest cumulative payment, use zero value", err)
 	}
 	// on-Chain account state
 	reservation, err := s.meterer.ChainPaymentState.GetReservedPaymentByAccount(ctx, accountID)
 	if err != nil {
-		return nil, api.NewErrorNotFound("failed to get active reservation")
+		s.logger.Debug("failed to get onchain reservation, use zero values", err)
+		reservation = core.DummyReservedPayment()
```
ian-shim, line 267: why don't we leave them as nil?

hopeyen: The way we [build reply](https://github.com/Layr-Labs/eigenda/blob/07670726c824e8228503a2799efc7183db73dab5/disperser/apiserver/server_v2.go#L291) struct assumes that there is a valid nonnil object. Would it make more sense to set the inner fields here and use those when we construct the reply?

ian-shim: Does the reply need to return zero valued object? Could we return nil instead?
Accountant handles nil reservation/ondemand payment [here](https://github.com/Layr-Labs/eigenda/pull/1014/files/07670726c824e8228503a2799efc7183db73dab5#diff-12b74ec71c12c2e17777a7dee108651d3a1cd184a32a2bb522a43c0f0baa204dR187)

hopeyen: We could return nil, it is just that we will add a separate step to convert cached struct to protobuf structs outside of the reply construction. 

This seems like a cleaner behavior, so I will update to do that

disperser/apiserver/server_v2.go
```diff
@@ -254,20 +255,44 @@ func (s *DispersalServerV2) GetPaymentState(ctx context.Context, req *pb.GetPaym
 	currentReservationPeriod := meterer.GetReservationPeriod(now, reservationWindow)
 	binRecords, err := s.meterer.OffchainStore.GetBinRecords(ctx, req.AccountId, currentReservationPeriod)
 	if err != nil {
-		return nil, api.NewErrorNotFound("failed to get active reservation")
+		s.logger.Debug("failed to get reservation records, use placeholders", err, accountID)
 	}
 	largestCumulativePayment, err := s.meterer.OffchainStore.GetLargestCumulativePayment(ctx, req.AccountId)
 	if err != nil {
-		return nil, api.NewErrorNotFound("failed to get largest cumulative payment")
+		s.logger.Debug("failed to get largest cumulative payment, use zero value", err)
```
ian-shim, line 262: nit: I think it would be helpful to log `accountID`s

disperser/apiserver/server_v2.go
```diff
@@ -254,20 +255,44 @@ func (s *DispersalServerV2) GetPaymentState(ctx context.Context, req *pb.GetPaym
 	currentReservationPeriod := meterer.GetReservationPeriod(now, reservationWindow)
 	binRecords, err := s.meterer.OffchainStore.GetBinRecords(ctx, req.AccountId, currentReservationPeriod)
 	if err != nil {
-		return nil, api.NewErrorNotFound("failed to get active reservation")
+		s.logger.Debug("failed to get reservation records, use placeholders", err, accountID)
```
ian-shim, line 258: use labels:
```
s.logger.Debug("failed to get reservation records, use placeholders", "err ", err, "accountID", accountID)
```

disperser/apiserver/server_v2.go
```diff
@@ -254,20 +255,44 @@ func (s *DispersalServerV2) GetPaymentState(ctx context.Context, req *pb.GetPaym
...
 	}
 	largestCumulativePayment, err := s.meterer.OffchainStore.GetLargestCumulativePayment(ctx, req.AccountId)
 	if err != nil {
-		return nil, api.NewErrorNotFound("failed to get largest cumulative payment")
+		s.logger.Debug("failed to get largest cumulative payment, use zero value", err)
 	}
 	// on-Chain account state
+	pbReservation := &pb.Reservation{}
```
ian-shim, line 265: if we declare it like `var pbReservation *pb.Reservation`, we don't have to set `pbReservation = nil` in L269

disperser/apiserver/server_v2.go
hopeyen: updated to include accountIDs and use labels

core/data.go
```diff
@@ -619,11 +619,27 @@ type ReservedPayment struct {
...
+		QuorumSplits:     []byte{},
+	}
+}
+
 type OnDemandPayment struct {
 	// Total amount deposited by the user
 	CumulativePayment *big.Int
 }
 
+func DummyOnDemandPayment() *OnDemandPayment {
```
ian-shim, line 637: Is this used anywhere?

api/clients/v2/accountant.go
```diff
@@ -154,56 +154,66 @@ func (a *Accountant) GetRelativeBinRecord(index uint32) *BinRecord {
...
+	}
+
+	if paymentState.GetCumulativePayment() == nil {
+		a.cumulativePayment = big.NewInt(0)
+	} else {
+		a.cumulativePayment = new(big.Int).SetBytes(paymentState.GetCumulativePayment())
+	}
+
+	if paymentState.GetReservation() == nil {
+		a.reservation = core.DummyReservedPayment()
```
ian-shim, line 187: nit: could we explicitly create empty object
```
a.reservation = &ReservedPayment{
	SymbolsPerSecond: 0,
	StartTimestamp:   0,
	EndTimestamp:     0,
	QuorumNumbers:    []uint8{},
	QuorumSplits:     []byte{},
}
```
and remove `DummyReservedPayment`?

ian-shim (approved):

core/data.go
hopeyen: removed

api/clients/v2/accountant.go
hopeyen: yes, updated

PR #1015, opened 2024-12-17
init once accountant

hopeyen:

Instead of relying on disperser client to call populate accountant before dispersing blobs, we add an init once function that automatically populate the accountant if it isn't constructed at the beginning 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/data.go
+++ b/core/data.go
@@ -619,25 +619,9 @@ type ReservedPayment struct {
 	QuorumSplits []byte
 }
 
-func DummyReservedPayment() *ReservedPayment {
-	return &ReservedPayment{
-		SymbolsPerSecond: 0,
-		StartTimestamp:   0,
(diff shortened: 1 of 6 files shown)


samlaf:
42 file changes... maybe target branch needs to be rebased as well or something?

hopeyen:
> 42 file changes... maybe target branch needs to be rebased as well or something?

Weird, this was supposed to be a single file change. taking a look now

core/data.go
```diff
@@ -619,11 +619,27 @@ type ReservedPayment struct {
 	QuorumSplits []byte
 }
```
ian-shim, line 621: we don't need this change right?

ian-shim (approved):

PR #1022, opened 2024-12-17
Sign get chunks

cody-littley:

Enable request signing for `GetChunks()` requests.

Companion PR to update the relay helm chart: https://github.com/Layr-Labs/eigenda-devops/pull/1256

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


inabox/tests/integration_v2_test.go
```diff
@@ -135,9 +135,18 @@ var _ = Describe("Inabox v2 Integration", func() {
 			}
 		}
 
+		// These values are only used by the relay client to get chunks, which we do not call from the relay
```
jianoaix, line 138: This looks just because of missing test coverage, then shall we add a call to test GetChunks?

api/clients/v2/relay_client.go
```diff
@@ -62,8 +69,8 @@ var _ RelayClient = (*relayClient)(nil)
 
 // NewRelayClient creates a new RelayClient that connects to the relays specified in the config.
 // It keeps a connection to each relay and reuses it for subsequent requests, and the connection is lazily instantiated.
-func NewRelayClient(config *RelayClientConfig, logger logging.Logger) (*relayClient, error) {
-	if config == nil || len(config.Sockets) <= 0 {
+func NewRelayClient(config *RelayClientConfig, logger logging.Logger) (RelayClient, error) {
+	if config == nil || len(config.Sockets) <= 0 || config.OperatorID == nil || config.MessageSigner == nil {
```
ian-shim, line 73: should we move this check `config.OperatorID == nil || config.MessageSigner == nil` to GetChunks methods? 
Other methods like `GetBlob` should work with nil operator ID and nil signer

relay/server_test.go
```diff
@@ -30,6 +30,7 @@ func defaultConfig() *Config {
 		ChunkCacheSize:             1024 * 1024,
 		ChunkMaxConcurrency:        32,
 		MaxKeysPerGetChunksRequest: 1024,
+		AuthenticationDisabled:     true,
```
ian-shim, line 33: if we set this to false, does integration test fail?

api/clients/v2/relay_client.go
cody-littley: I've moved these checks out of the constructor. They will cause the methods that get chunks to return an error if they are not properly configured.

relay/server_test.go
cody-littley: Disabling authentication doesn't cause signed requests to be rejected, it just causes signatures to be entirely ignored by the relay.

That being said, disabling authentication in the relay tests was a bit of a shortcut (the bad kind). I've modified the code in this test file to properly enable authentication.

inabox/tests/integration_v2_test.go
cody-littley: Talked to @ian-shim about this. He said the calls to `GetChunks()` are made from the operator nodes, not from this client instance. We are only using this client instance to call `GetBlobs()`.

relay/mock/ics.go
```diff
@@ -0,0 +1,68 @@
+package mock
```
ian-shim, line 1: why do we need a separate mock for relay? 
don't we have an existing mock for IndexedChainState?

ian-shim (approved):

api/clients/v2/relay_client.go
```diff
@@ -97,7 +104,25 @@ func (c *relayClient) GetBlob(ctx context.Context, relayKey corev2.RelayKey, blo
 	return res.GetBlob(), nil
 }
 
-func (c *relayClient) GetChunksByRange(ctx context.Context, relayKey corev2.RelayKey, requests []*ChunkRequestByRange) ([][]byte, error) {
+// signGetChunksRequest signs the GetChunksRequest with the operator's private key
+// and sets the signature in the request.
+func (c *relayClient) signGetChunksRequest(ctx context.Context, request *relaygrpc.GetChunksRequest) error {
+	hash := auth.HashGetChunksRequest(request)
```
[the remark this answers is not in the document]
jianoaix: Seems good area to focus for auditors

jianoaix (approved):

relay/mock/ics.go
cody-littley: The existing `ChainDataMock` struct has a bunch of internal logic that looks like it was written to support specific unit tests. I just need a mock implementation where I can use the pattern `when X method is called with Y arguments, return Z`.

PR #1023, opened 2024-12-17
fix: make password as non required field

### LINKED ISSUE #1021: [Bug]: Empty NODE_BLS_KEY_PASSWORD error
### What happened + What you expected to happen

Updated to the latest release on holesky, ` v0.8.5-rc.1` and encountered an issue starting the daemon.    We keep getting the following error:
```2024/12/17 14:48:26 application failed: BLS key file and password is required if BLS remote signer is disabled```

We are using an empty password for our local BLS key.   It appears this update checks that the key file name, and pass are NOT == "", which is leading to this. 

### Versions / Dependencies

v0.8.5-rc.1

### How to reproduce

Use local bls key, with an empty password.

### Issue Severity

High: It blocks me from completing my task.

### PR DESCRIPTION
**Pull request by shrimalmadhur**

Fixes #1021 

- Makes password as not required since folks do testing with no password encrypted keys.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/node/config.go
+++ b/node/config.go
@@ -165,8 +165,8 @@ func NewConfig(ctx *cli.Context) (*Config, error) {
 	if blsRemoteSignerEnabled && (ctx.GlobalString(flags.BLSRemoteSignerUrlFlag.Name) == "" || ctx.GlobalString(flags.BLSPublicKeyHexFlag.Name) == "") {
 		return nil, fmt.Errorf("BLS remote signer URL and Public Key Hex is required if BLS remote signer is enabled")
 	}
-	if !blsRemoteSignerEnabled && (ctx.GlobalString(flags.BlsKeyFileFlag.Name) == "" || ctx.GlobalString(flags.BlsKeyPasswordFlag.Name) == "") {
-		return nil, fmt.Errorf("BLS key file and password is required if BLS remote signer is disabled")
+	if !blsRemoteSignerEnabled && ctx.GlobalString(flags.BlsKeyFileFlag.Name) == "" {
+		return nil, fmt.Errorf("BLS key file is required if BLS remote signer is disabled")
 	}
 
 	// Decrypt BLS key

### CODE REVIEW

File: node/config.go [hunk 1]
```diff
@@ -165,8 +165,8 @@ func NewConfig(ctx *cli.Context) (*Config, error) {
 	if blsRemoteSignerEnabled && (ctx.GlobalString(flags.BLSRemoteSignerUrlFlag.Name) == "" || ctx.GlobalString(flags.BLSPublicKeyHexFlag.Name) == "") {
 		return nil, fmt.Errorf("BLS remote signer URL and Public Key Hex is required if BLS remote signer is enabled")
 	}
-	if !blsRemoteSignerEnabled && (ctx.GlobalString(flags.BlsKeyFileFlag.Name) == "" || ctx.GlobalString(flags.BlsKeyPasswordFlag.Name) == "") {
-		return nil, fmt.Errorf("BLS key file and password is required if BLS remote signer is disabled")
+	if !blsRemoteSignerEnabled && ctx.GlobalString(flags.BlsKeyFileFlag.Name) == "" {
```
dmanc, line 168: Maybe add a warn message?

dmanc (approved):

node/config.go
shrimalmadhur: warn message about not providing a password?

shrimalmadhur: it will anyway fail since the keys won't decrypt and server won't start. so I am not sure if warn is helpful for that.

ian-shim (commented):
Hmm I think we need an input from @anupsv 
I think an argument can be made that we can be more opinionated here and require password even if they just want to test it (or only allow no password if `testMode`?). So that there cannot be plain keys in production

shrimalmadhur:
> cannot be plain keys in production

hmm I think in testMode - we use plaintext pk. 
In `testnet` though - which is not testMode - they still use a file but what the issue says is they don't encrypt the file with password (which is kinda useless). But I am wondering if folks want to quickly test with no password on testnet. We can try to encore password on testnet too if that makes more sense.

shrimalmadhur:
Closing this. As Anup said we want to have similar behavior in testnet and mainnet.

PR #1024, opened 2024-12-17
Nil checks in disperser v2 server constructor

ian-shim:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/apiserver/server_v2.go
+++ b/disperser/apiserver/server_v2.go
@@ -74,7 +74,32 @@ func NewDispersalServerV2(
 	onchainStateRefreshInterval time.Duration,
 	_logger logging.Logger,
 	registry *prometheus.Registry,
-) *DispersalServerV2 {
+) (*DispersalServerV2, error) {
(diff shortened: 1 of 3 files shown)


disperser/apiserver/server_v2.go
```diff
@@ -232,6 +257,9 @@ func (s *DispersalServerV2) RefreshOnchainState(ctx context.Context) error {
 }
 
 func (s *DispersalServerV2) GetPaymentState(ctx context.Context, req *pb.GetPaymentStateRequest) (*pb.GetPaymentStateReply, error) {
+	if s.meterer == nil {
```
hopeyen, line 260: why do we check nil meterer here instead of in the `NewDispersalServerV2` constructor?

ian-shim: This is to get around with disabling payments. Currently, payment's disabled and meterer is not set. We still want the server to run until we reenable it

hopeyen (approved):

PR #1027, opened 2024-12-17
Update dependabot.yml

anupsv:

Updating dependabot to only notify security updates for version bumps.


jianoaix (commented):
This was quite noisy, is it ready to enable?

anupsv:
> This was quite noisy, is it ready to enable?

So what i've done is to use dependabot only for security updates and not version updates. We should do version updates as well as part of the code quality but we'll tackle that later. So for security updates for direct dependencies, it will create a PR along with the severity. I will help review these PR's as well.

jianoaix (approved):

PR #1034, opened 2024-12-19
symbols charged and payment metadata check

hopeyen:

Making sure that symbols charged are correctly tracked, and payment metadata is considered valid if one of the payment specification is missing

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/accountant.go
+++ b/api/clients/v2/accountant.go
@@ -67,14 +67,15 @@ func NewAccountant(accountID string, reservation *core.ReservedPayment, onDemand
 // then on-demand if the reservation is not available. The returned values are
 // reservation period for reservation payments and cumulative payment for on-demand payments,
 // and both fields are used to create the payment header and signature
-func (a *Accountant) BlobPaymentInfo(ctx context.Context, numSymbols uint64, quorumNumbers []uint8) (uint32, *big.Int, error) {
+func (a *Accountant) BlobPaymentInfo(ctx context.Context, numSymbols uint32, quorumNumbers []uint8) (uint32, *big.Int, error) {
 	now := time.Now().Unix()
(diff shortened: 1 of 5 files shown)


api/clients/v2/accountant.go
```diff
@@ -87,18 +88,19 @@ func (a *Accountant) BlobPaymentInfo(ctx context.Context, numSymbols uint64, quo
 
 	overflowBinRecord := a.GetRelativeBinRecord(currentReservationPeriod + 2)
 	// Allow one overflow when the overflow bin is empty, the current usage and new length are both less than the limit
-	if overflowBinRecord.Usage == 0 && relativeBinRecord.Usage-numSymbols < binLimit && numSymbols <= binLimit {
-		overflowBinRecord.Usage += relativeBinRecord.Usage - binLimit
+	overflowBinRecord.Usage += relativeBinRecord.Usage - binLimit
+	if overflowBinRecord.Usage == relativeBinRecord.Usage-binLimit && relativeBinRecord.Usage-symbolUsage < binLimit && symbolUsage <= binLimit {
```
ian-shim, line 92: not sure if i understand this change, it seems like this condition is same as before but we preemptively add the overflow usage before the check. 
why was this change necessary?

hopeyen: Yep, we added this so that we will always add to the bin if reservation is used, otherwise both current and overflow bin gets reverted when the accountant uses ondemand (rollback added for overflow at [line](https://github.com/Layr-Labs/eigenda/pull/1034/files/c75419c7617f9314434bb69f7920e9091823dc6b#diff-12b74ec71c12c2e17777a7dee108651d3a1cd184a32a2bb522a43c0f0baa204dR101))

ian-shim (approved):

PR #1033, opened 2024-12-19
fix: global symbols per period and interval

hopeyen:
## Why are these changes needed?

quick fixes related to global reservations

## Checks

- [ ] I've made sure the tests are passing. Note that there might be a few flaky tests, in that case, please comment that they are not relevant.
- [ ] I've checked the new test coverage and the coverage percentage didn't drop.
- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests
   - [ ] This PR is not tested :(

PR #1035, opened 2024-12-19
payment failure cases inabox

hopeyen:

explicitly running out of funds for both payments (will rebase after merging #991, #1033, and #1034 , don't need to review yet)

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/inabox/tests/integration_v2_payments_test.go
+++ b/inabox/tests/integration_v2_payments_test.go
@@ -1,0 +1,94 @@
+package integration_test
+
+import (
+	"context"
+	"crypto/rand"
+	"time"
+
+	"github.com/Layr-Labs/eigenda/api/clients/v2"
+	auth "github.com/Layr-Labs/eigenda/core/auth/v2"
+	dispv2 "github.com/Layr-Labs/eigenda/disperser/common/v2"
+	"github.com/Layr-Labs/eigenda/encoding/utils/codec"
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+)
+
+var _ = Describe("Inabox v2 Integration", func() {
+	It("reservation runs out", func() {
+		ctx, cancel := context.WithTimeout(context.Background(), 60*time.Second)
+		defer cancel()
+
+		privateKeyHex := "0x0123456789abcdef0123456789abcdef0123456789abcdef0123456789abcdee"
+		signer := auth.NewLocalBlobRequestSigner(privateKeyHex)
+
+		disp, err := clients.NewDisperserClient(&clients.DisperserClientConfig{
+			Hostname: "localhost",
+			Port:     "32005",
+		}, signer, nil, nil)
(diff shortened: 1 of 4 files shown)


inabox/tests/integration_v2_payments_test.go
```diff
@@ -0,0 +1,94 @@
+package integration_test
```
ian-shim, line 1: I wonder if these tests could be moved to unit tests? like in `accountant_test.go`? 
Also, should we also test payment failure cases from disperser server if there isn't already? (to cover the case that user isn't running accountatnt)

hopeyen: hmm I made these tests for the interaction of disperser client and the server, and shows when a blob should be dispersed and when it should fail with respect to payments. I don't think they could be moved to unit test.
Disperser client would build accountant if it wasn't provided with one, so alternatively we could add rogue accountant implementations that makes invalid payment headers

ian-shim: > hmm I made these tests for the interaction of disperser client and the server

Is the server logic relevant in these tests? It looks like they mostly test the accountant logic (server could have the payment metering disabled and allowed all requests and these tests would still pass)

> Disperser client would build accountant if it wasn't provided with one

Yes, but not all requests may originate from the disperser client shipped in this repo. Anyone is free to disperse blobs without running accountant, in which case, API server should enforce correct metering scheme. I was referring to payment failure cases from the API server payment metering, not from the accountant.

hopeyen: > Is the server logic relevant in these tests? It looks like they mostly test the accountant logic (server could have the payment metering disabled and allowed all requests and these tests would still pass)

Hmm I understood your concerns. I updated the tests to override the accountant with erroneous payment state so that accountant doesn't account correctly, and focus on testing failures from API server logic.  (if payment is disabled, the tests should not pass as the expected failure would instead be nil)

> not all requests may originate from the disperser client shipped in this repo. Anyone is free to disperse blobs without running accountant, in which case, API server should enforce correct metering scheme.

by this logic, should we have a set of tests that doesn't use DisperserClient? The existing tests all utilize `NewDisperserClient` to make API requests

ian-shim: Yes, I think tests for accountant and metering from API server should be independently done covering potentially identical/similar scenarios so that they're covered in both client & server side. If we do this, we could probably have them as unit tests in `accountant_test.go` and `apiserver/server_v2_test.go` instead of as part of the inabox test

hopeyen: There are existing unit tests for both meterer and accountants, covering similar logic on both sides. What's done here is the interaction between accountant and meterer. Do you think it is not necessary to test the interaction? I'm okay to close the PR if that's the case?

PR #1043, opened 2024-12-19
Fail v2 blobs for no attestations

ian-shim:
If a blob receives no attestation for any of its target quorums, update status as `INSUFFICIENT_SIGNATURES`

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


disperser/controller/dispatcher.go
```diff
@@ -489,12 +491,41 @@ func (d *Dispatcher) sendChunks(ctx context.Context, client clients.NodeClient,
...
+		for _, q := range cert.BlobHeader.QuorumNumbers {
+			if res, ok := attestation.QuorumResults[q]; !ok || res == 0 {
+				d.logger.Error("quorum result not found", "quorumID", q, "blobKey", blobKey.Hex())
+				failed = true
+				break
+			}
+		}
+
+		if failed {
+			err := d.blobMetadataStore.UpdateBlobStatus(ctx, blobKey, v2.INSUFFICIENT_SIGNATURES)
```
dmanc, line 517: So right now the insufficient signature state is for when a quorum has no attestations?

disperser/controller/encoding_manager_test.go
```diff
@@ -2,10 +2,11 @@ package controller_test
 
 import (
 	"context"
-	"github.com/prometheus/client_golang/prometheus"
 	"testing"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
+
```
dmanc, line 9: nit: remove newline

disperser/controller/dispatcher.go
ian-shim: Yeah
We can improve this by either considering the coding rate or the default confirmation threshold onchain, but this is a quick fix before further steps

PR #1042, opened 2024-12-19
Create util to verify KZG commitment by generation and comparison

litt3:
## Why are these changes needed?

- This PR is part of https://github.com/Layr-Labs/eigenda/issues/1000
- It includes a utility to verify kzg commitments by regenerating them, and comparing vs the claimed commitment
- The optimized version of verification using fiat shamir transformation will be included in a future PR

## Checks

- [X] I've made sure the tests are passing. Note that there might be a few flaky tests, in that case, please comment that they are not relevant.
- [ ] I've checked the new test coverage and the coverage percentage didn't drop.
- Testing Strategy
   - [X] Unit tests
   - [ ] Integration tests
   - [ ] This PR is not tested :(

PR #1051, opened 2024-12-20
Move commitment verification into verifier

litt3:
This PR moves the newly implemented commitment verification into the verifier, as suggested [here](https://github.com/Layr-Labs/eigenda/pull/1042#discussion_r1894059079)


encoding/kzg/verifier/verifier.go
```diff
@@ -329,3 +329,45 @@ func PairingsVerify(a1 *bn254.G1Affine, a2 *bn254.G2Affine, b1 *bn254.G1Affine,
 
 	return nil
 }
+
+// GenerateBlobCommitment computes a kzg-bn254 commitment of blob data using SRS
+func (v *Verifier) GenerateBlobCommitment(blob []byte) (*encoding.G1Commitment, error) {
```
samlaf, line 334: does this need to be public?

encoding/kzg/verifier/verifier.go
```diff
@@ -329,3 +329,45 @@ func PairingsVerify(a1 *bn254.G1Affine, a2 *bn254.G2Affine, b1 *bn254.G1Affine,
...
+	if err != nil {
+		return nil, fmt.Errorf("MultiExp: %w", err)
+	}
+
+	return &encoding.G1Commitment{X: commitment.X, Y: commitment.Y}, nil
+}
+
+// GenerateAndCompareBlobCommitment generates the kzg-bn254 commitment of the blob, and compares it with a claimed
+// commitment. An error is returned if there is a problem generating the commitment, or if the comparison fails.
+func (v *Verifier) GenerateAndCompareBlobCommitment(claimedCommitment *encoding.G1Commitment, blobBytes []byte) error {
```
samlaf, line 358: does this need to be public?

samlaf (approved):
LGTM! Let's just make sure to make the methods private IF they don't need to be public, to not clutter the verifier's API. Not sure how/where you plan to use them so fine with me if they need to be public.

encoding/kzg/verifier/verifier.go
litt3: Yes, this will be called by the v2 client

encoding/kzg/verifier/verifier.go
litt3: I don't have any particular external use cases in mind. My tendency would be to keep it public, since it's a general purpose method to compute commitment from bytes, with likelihood of being used elsewhere in the future. But I'd have no strong objection to making it private

samlaf: We can rework the API later but I find it a bit weird that this method is on a strict called Verifier. I might want to compute a kzg commitment without verifying anything.

litt3: I'll make it private for now, and we can reconsider moving somewhere else more suitable, as future use cases present themselves

litt3: I see your point. We could create a new type `Committer`, which would accept a kzg configuration, and just produce the commitments.

litt3:
Closing this PR, to move the methods to better places. See https://github.com/Layr-Labs/eigenda/pull/1051#discussion_r1894261800

PR #1052, opened 2024-12-20
Disperser auth innabox

cody-littley:

Enable `StoreChunks()` signing in inabox tests.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


inabox/deploy/deploy.go
```diff
@@ -155,12 +161,95 @@ func (env *Config) DeployExperiment() {
...
 }
 
+// GenerateDisperserKeypair generates a disperser keypair using AWS KMS.
+func (env *Config) GenerateDisperserKeypair() error {
+
+	// Generate a keypair in AWS KMS
+
+	keyManager := kms.New(kms.Options{
+		Region:       "us-east-1",
+		BaseEndpoint: aws.String("http://localhost:4570"), // TODO don't hard code this
```
ian-shim, line 183: can we do this in this PR?

ian-shim (commented):
lgtm!

ian-shim (approved):

inabox/deploy/deploy.go
cody-littley: Oops, was my intention but I forgot. Now fixed.

core/eth/writer.go
```diff
@@ -6,6 +6,9 @@ import (
 	"encoding/hex"
 	"encoding/json"
 	"fmt"
+	"github.com/Layr-Labs/eigenda/api"
+	contractEigenDADisperserRegistry "github.com/Layr-Labs/eigenda/contracts/bindings/EigenDADisperserRegistry"
```
jianoaix, line 10: use lowecase for alias

core/eth/writer.go
```diff
@@ -329,3 +332,39 @@ func (t *Writer) ConfirmBatch(ctx context.Context, batchHeader *core.BatchHeader
...
+	if err != nil {
+		return fmt.Errorf("failed to create transaction for setting disperser address: %w", err)
+	}
+
+	err = t.ethClient.SendTransaction(ctx, transaction)
+	if err != nil {
+		return fmt.Errorf("failed to set disperser address: %w", err)
+	}
+
+	return nil
```
jianoaix, line 369: Does this need to wait for txn to be mined?

core/eth/writer.go
cody-littley: renamed to `dreg`

core/eth/writer.go
cody-littley: I've added a retry loop in the function that calls this function. It will now wait for up to a minute if the disperser address is not available on chain.

```
        // Read the disperser's public key from on-chain storage to verify it was written correctly
	
	retryTimeout := time.Now().Add(1 * time.Minute)
	ticker := time.NewTicker(1 * time.Second)
	
	for time.Now().Before(retryTimeout) {
		address, err := writer.GetDisperserAddress(context.Background(), 0)
		if err != nil {
			logger.Errorf("could not get disperser address: %v", err)
		} else {
			if address != env.DisperserAddress {
				return fmt.Errorf("expected disperser address %s, got %s", env.DisperserAddress, address)
			}
			return nil
		}
		
		<- ticker.C
	}

	return fmt.Errorf("timed out waiting for disperser address to be set")
```

jianoaix: Would `EstimateGasPriceAndLimitAndSendTx` in the ethclient work cleaner?

cody-littley: This caused errors. I initially tried to use this method, but later in the test run I'd get failures with the following error message:

```
{
   "time":"2025-01-03T16:28:01.407636692Z",
   "level":"ERROR",
   "source":{
      "function":"github.com/Layr-Labs/eigenda/core/eth.(*Reader).GetOperatorStakesForQuorums",
      "file":"/home/runner/work/eigenda/eigenda/core/eth/reader.go",
      "line":437
   },
   "msg":"Failed to fetch operator state",
   "component":"Reader",
   "err":"execution reverted: IndexRegistry._operatorCountAtBlockNumber: quorum did not exist at given block number"
}
```

I spent a lot of time on this and was never able to determine how these two things were related. But as soon as I switched to using `SendTransaction`, the problem went away.

PR #1053, opened 2024-12-20
[v2] eth address as account id

hopeyen:

just for investigating account behaviors

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/auth/v2/authenticator.go
+++ b/core/auth/v2/authenticator.go
@@ -6,6 +6,7 @@ import (
 	"fmt"
 
 	core "github.com/Layr-Labs/eigenda/core/v2"
+	"github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/crypto"
 )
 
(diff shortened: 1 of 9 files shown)


core/auth/v2/authenticator.go
```diff
@@ -32,24 +30,16 @@ func (*authenticator) AuthenticateBlobRequest(header *core.BlobHeader) error {
...
 	sigPublicKeyECDSA, err := crypto.SigToPub(blobKey[:], sig)
 	if err != nil {
 		return fmt.Errorf("failed to recover public key from signature: %v", err)
 	}
 
-	if !bytes.Equal(pubKey.X.Bytes(), sigPublicKeyECDSA.X.Bytes()) || !bytes.Equal(pubKey.Y.Bytes(), sigPublicKeyECDSA.Y.Bytes()) {
+	accountId := header.PaymentMetadata.AccountID
+	pubKey := crypto.PubkeyToAddress(*sigPublicKeyECDSA).Hex()
+
+	if pubKey != accountId {
```
ian-shim, line 42: Here we're comparing strings, but this will fail if `accountID` isn't checksummed string. 
I think it's safer to convert `accountID` to common.Address (common.HexToAddress()) and compare bytearrays using `Address.Cmp()`

hopeyen: Updated to use address type.

We have auth_test that check for authenticating the requests, but not as specific as whether it is checksum or not.

ian-shim (approved):

PR #1054, opened 2024-12-20
Move SRS initialization logic into SRS constructor

litt3:
## Changes

- Add `G2Trailing` SRS data into `SRS` struct
    - Discussed with @bxue-l2, since these points are SRS, it makes sense to include them in the struct directly
    - This change enabled the SRS initialization simplifications
- Extract SRS initialization from `Prover` and `Verifier`
    - The code in these two locations was nearly identical
    - This abstraction allows me to avoid an additional duplication of this logic, while implementing a [further](https://github.com/Layr-Labs/eigenda/pull/1051#discussion_r1894274288) utility which requires SRS

### DIFF
--- a/encoding/kzg/prover/gnark/commitments.go
+++ b/encoding/kzg/prover/gnark/commitments.go
@@ -10,9 +10,8 @@ import (
 )
 
 type KzgCommitmentsGnarkBackend struct {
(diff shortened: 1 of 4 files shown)


litt3:
This is potentially good cleanup, but turned out to be unnecessary for now.

PR #1064, opened 2025-01-03
chore: fix quotation marks

xiaobei0715:
Incorrect use of quotation marks can lead to certain editors failing to accurately render text content, which may result in garbled output or other formatting issues that could confuse readers.


pschork:
Closing per policy to not accept minor documentation fixes.

PR #1066, opened 2025-01-06
Create a blob verification utility

litt3:

- This verification utility is needed for the v2 client

## Notes

- I made the decision to *not* add the verification contract logic into the existing `reader.go`, since doing so would cause a cascade of changes in unrelated files
    - IMO, I don't see the advantage of piling all of this stuff into a single class, when the only thing shared is the eth client

## Checks

- [X] I've made sure the tests are passing. Note that there might be a few flaky tests, in that case, please comment that they are not relevant.
- Testing Strategy: This verification utility cannot be meaningfully unit tested. E2E integration tests will be created in a future PR.


cody-littley (approved):

api/grpc/disperser/v2/conversion_utils.go
cody-littley, line 1: IMO, it's a little odd putting code into packages that otherwise just contain generated code. What if instead of this, we add constructors to the wrapper structs that accept a protobuf and produce the wrapper?

litt3: Agreed, this is better. Done.

cody-littley (approved):

api/clients/v2/verification/blob_verifier.go
```diff
@@ -0,0 +1,71 @@
...
+
+	disperser "github.com/Layr-Labs/eigenda/api/grpc/disperser/v2"
+	verifierBindings "github.com/Layr-Labs/eigenda/contracts/bindings/EigenDABlobVerifier"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	gethcommon "github.com/ethereum/go-ethereum/common"
+)
+
+// BlobVerifier is responsible for making eth calls to verify blobs that have been received by the client
+type BlobVerifier struct {
+	// go binding around the verifyBlobV2FromSignedBatch ethereum contract
```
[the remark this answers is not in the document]
litt3: [ca8d9d74](https://github.com/Layr-Labs/eigenda/pull/1066/commits/ca8d9d74a74b847b885a233185e44ab3661f49ba)

api/clients/v2/verification/blob_verifier.go
```diff
@@ -0,0 +1,71 @@
...
+	blobVerifierCaller *verifierBindings.ContractEigenDABlobVerifierCaller
+}
+
+// NewBlobVerifier constructs a BlobVerifier
+func NewBlobVerifier(
+	ethClient *common.EthClient,               // the eth client, which should already be set up
+	verifyBlobV2FromSignedBatchAddress string, // the hex address of the verifyBlobV2FromSignedBatch contract
```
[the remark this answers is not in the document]
litt3: [ca8d9d74](https://github.com/Layr-Labs/eigenda/pull/1066/commits/ca8d9d74a74b847b885a233185e44ab3661f49ba)

api/clients/v2/verification/blob_verifier.go
```diff
@@ -0,0 +1,71 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("bind to verifier contract at %s: %s", verifyBlobV2FromSignedBatchAddress, err)
+	}
+
+	return &BlobVerifier{
+		blobVerifierCaller: verifierCaller,
+	}, nil
+}
+
+// VerifyBlobV2FromSignedBatch makes a call to the verifyBlobV2FromSignedBatch contract
```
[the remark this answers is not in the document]
litt3: Fixed terminology [ca8d9d74](https://github.com/Layr-Labs/eigenda/pull/1066/commits/ca8d9d74a74b847b885a233185e44ab3661f49ba)

anupsv (approved):

api/clients/v2/verification/blob_verifier.go
```diff
@@ -0,0 +1,71 @@
...
+		return fmt.Errorf("convert signed batch: %s", err)
+	}
+
+	convertedBlobVerificationProof, err := blobVerificationProof.ToBinding()
+	if err != nil {
+		return fmt.Errorf("convert blob verification proof: %s", err)
+	}
+
+	err = v.blobVerifierCaller.VerifyBlobV2FromSignedBatch(
+		&bind.CallOpts{Context: ctx},
```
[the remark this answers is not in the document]
litt3: Added a comment clarifying that the caller is responsible for any timeout [c752e5b1](https://github.com/Layr-Labs/eigenda/pull/1066/commits/c752e5b1b186773cac3bd541d88e5f40d5a13a7e)

api/clients/v2/verification/blob_verifier.go
```diff
@@ -0,0 +1,74 @@
...
+
+	"github.com/Layr-Labs/eigenda/common"
+
+	disperser "github.com/Layr-Labs/eigenda/api/grpc/disperser/v2"
+	verifierBindings "github.com/Layr-Labs/eigenda/contracts/bindings/EigenDABlobVerifier"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	gethcommon "github.com/ethereum/go-ethereum/common"
+)
+
+// BlobVerifier is responsible for making eth calls against the BlobVerifier contract to ensure cryptographic and
```
samlaf, line 15: add link to BlobVerifier contract?

api/clients/v2/verification/blob_verifier.go
```diff
@@ -0,0 +1,74 @@
...
+	}
+
+	return &BlobVerifier{
+		blobVerifierCaller: verifierCaller,
+	}, nil
+}
+
+// VerifyBlobV2FromSignedBatch calls the verifyBlobV2FromSignedBatch view function on the EigenDABlobVerifier contract
+//
+// This method returns nil if the blob is successfully verified. Otherwise, it returns an error.
```
samlaf, line 43: I prefer when a method returns (bool, error) for these kinds of cases, because there are many error cases that are unrelated to the cert being verified or not (networking error being the most prominent one).

samlaf (commented):
@litt3 you must be a masochist asking for so many people to review your PR. As the saying goes: 10 line PR = 10 comments. 1000 line PR = LGTM. So here's my 2 comments :D

bxue-l2:
One thing that looks missing to me is the test.

api/clients/v2/verification/blob_verifier.go
litt3: @samlaf The problem is that the underlying function call, `VerifyBlobV2FromSignedBatch`, communicates a failed verification via an error. So, there is never a case where this method would return `false, nil`, nor `true, notNil`

A returned bool _could_ be used to convey useful information, if we defined it to mean "returns true if there are no errors preceding verification". But a bool being returned from a "verify" function already has strong baked in expectations of meaning "verified/not verified"

Or, we could make the return signature `(bool, error)`, even though it's a bit redundant. LMK what you guys think

api/clients/v2/verification/blob_verifier.go
litt3: Do you mean like this? I don't like that this sort of link is instantly out of date, but I don't know of a better way to link here

https://github.com/Layr-Labs/eigenda/blob/06e88b311ef27618b4b878a78cbbb24358104ab9/contracts/src/core/EigenDABlobVerifier.sol#L108

litt3:
> One thing that looks missing to me is the test. Though it looks difficult to test, unless we can mock the eth call, and prepare all the consistent data

@bxue-l2 I considered mocking the eth call, but it would be a lot of boilerplate to really only test the simple logic of whether errors are returned. This seems to fall into the "99% coverage" bucket.

I wouldn't have a problem with a or 99% coverage policy, if that's what we want to go for across the board. Thoughts?

contracts/bindings/EigenDABlobVerifier/conversion_utils.go
```diff
@@ -0,0 +1,227 @@
...
+
+	return &BN254G1Point{
+		X: g1Point.X.BigInt(new(big.Int)),
+		Y: g1Point.Y.BigInt(new(big.Int)),
+	}, nil
+}
+
+func bytesToBN254G2Point(bytes []byte) (*BN254G2Point, error) {
+	var g2Point bn254.G2Affine
+	_, err := g2Point.SetBytes(bytes)
```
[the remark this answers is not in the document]
litt3: Done [f31e4a6b](https://github.com/Layr-Labs/eigenda/pull/1066/commits/f43c7cebdc7b4f3b8e98790a70a46574bf68a5c8)

0x0aa0 (approved):

bxue-l2:
> > One thing that looks missing to me is the test. Though it looks difficult to test, unless we can mock the eth call, and prepare all the consistent data
> 
> @bxue-l2 I considered mocking the eth call, but it would be a lot of boilerplate to really only test the simple logic of whether errors are returned. This seems to fall into the "99% coverage" bucket.
> 
> I wouldn't have a problem with a 99% coverage policy, if that's what we want to go for across the board. Thoughts?

Got it, I am not trying to impose a high coverage policy, The code looks good to me.

contracts/bindings/EigenDABlobVerifier/conversion_utils.go
```diff
@@ -0,0 +1,229 @@
...
+
+	// SetBytes checks that the result is in the correct subgroup
+	_, err := g2Point.SetBytes(bytes)
+
+	if err != nil {
+		return nil, fmt.Errorf("deserialize g2 point: %s", err)
+	}
+
+	var x, y [2]*big.Int
+	x[0] = g2Point.X.A0.BigInt(new(big.Int))
```
ian-shim, line 205: Has this been tested? 
I'm pretty sure the [ordering](https://github.com/Layr-Labs/eigenlayer-middleware/blob/mainnet/src/libraries/BN254.sol#L43) should be reversed `X = {X.A1, X.A0}`

api/clients/v2/verification/blob_verifier.go
samlaf: Yes did mean like this. We can also just link to the contract itself, in the master branch, so that it doesn't get out of date instantly.

api/clients/v2/verification/blob_verifier.go
samlaf: Gotcha, that make sense. Was thinking the validation onchain was trivial, but I see it can already return many different types of errors, and it wouldn't make sense to compress those to a boolean result.

contracts/bindings/EigenDABlobVerifier/conversion_utils.go
litt3: > Has this been tested?

Not yet, integration tests are forthcoming

Good catch, thank you. [Fixed](https://github.com/Layr-Labs/eigenda/pull/1066/commits/66249bcaece421283828f8d8652538238d55fc39)

api/clients/v2/verification/blob_verifier.go
litt3: Added [8bba429a](https://github.com/Layr-Labs/eigenda/pull/1066/commits/8bba429aea90afe9366b87d94cb961c6eac459b4)

epociask (approved):
LGTM

samlaf (approved):

PR #1069, opened 2025-01-07
feat: add basic logging in disperser meterer

hopeyen:

There's no logging for meterer at all. Added 3 basic lines to log the general flow of meterer processing 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


core/meterer/meterer.go
```diff
@@ -98,6 +99,7 @@ func (m *Meterer) MeterRequest(ctx context.Context, header core.PaymentMetadata,
 
 // ServeReservationRequest handles the rate limiting logic for incoming requests
 func (m *Meterer) ServeReservationRequest(ctx context.Context, header core.PaymentMetadata, reservation *core.ReservedPayment, numSymbols uint, quorumNumbers []uint8) error {
+	m.logger.Info("Recording and validating reservation usage", "reservation", reservation)
```
dmanc, line 102: Should we add header to this log and below one too?

dmanc (approved):

core/meterer/meterer.go
hopeyen: yeah, added:)

PR #1073, opened 2025-01-07
fix: payment metadata invalid if both payment fields empty

hopeyen:

quick fix

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/apiserver/disperse_blob_v2.go
+++ b/disperser/apiserver/disperse_blob_v2.go
@@ -118,7 +118,7 @@ func (s *DispersalServerV2) validateDispersalRequest(ctx context.Context, req *p
 		return api.NewErrorInvalidArg("payment metadata is required")
 	}
 
-	if len(blobHeader.PaymentMetadata.AccountID) == 0 || blobHeader.PaymentMetadata.ReservationPeriod == 0 || blobHeader.PaymentMetadata.CumulativePayment == nil {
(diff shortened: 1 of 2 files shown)


ian-shim (approved):

disperser/apiserver/disperse_blob_v2.go
```diff
@@ -118,7 +118,7 @@ func (s *DispersalServerV2) validateDispersalRequest(ctx context.Context, req *p
 		return api.NewErrorInvalidArg("payment metadata is required")
 	}
 
-	if len(blobHeader.PaymentMetadata.AccountID) == 0 || blobHeader.PaymentMetadata.ReservationPeriod == 0 || blobHeader.PaymentMetadata.CumulativePayment == nil {
+	if len(blobHeader.PaymentMetadata.AccountID) == 0 || (blobHeader.PaymentMetadata.ReservationPeriod == 0 && blobHeader.PaymentMetadata.CumulativePayment.Cmp(big.NewInt(0)) == 0) {
```
ian-shim, line 121: does `blobHeader.PaymentMetadata.CumulativePayment.Cmp(big.NewInt(0))` still evaluate to 0 if `CumulativePayment` is nil?

hopeyen: No, `nil.Cmp(...)` would fail, but our conversion from proto to core data structure (`ConvertToPaymentMetadata`) would force big Int to be not nil

PR #1078, opened 2025-01-08
Add support for fetch all operators' responses for a batch

jianoaix:
Followup to https://github.com/Layr-Labs/eigenda/pull/1075

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/common/v2/blobstore/dynamo_metadata_store.go
+++ b/disperser/common/v2/blobstore/dynamo_metadata_store.go
@@ -480,6 +480,35 @@ func (s *BlobMetadataStore) GetDispersalResponse(ctx context.Context, batchHeade
 	return res, nil
 }
 
+func (s *BlobMetadataStore) GetDispersalResponses(ctx context.Context, batchHeaderHash [32]byte) ([]*corev2.DispersalResponse, error) {
+	items, err := s.dynamoDBClient.Query(ctx, s.tableName, "PK = :pk AND begins_with(SK, :prefix)", commondynamodb.ExpressionValues{
+		":pk": &types.AttributeValueMemberS{
+			Value: dispersalKeyPrefix + hex.EncodeToString(batchHeaderHash[:]),
+		},
+		":prefix": &types.AttributeValueMemberS{
+			Value: dispersalResponseSKPrefix,
+		},
+	})
+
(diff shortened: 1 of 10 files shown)


jianoaix:
ping

disperser/common/v2/blobstore/dynamo_metadata_store.go
```diff
@@ -480,6 +480,35 @@ func (s *BlobMetadataStore) GetDispersalResponse(ctx context.Context, batchHeade
 	return res, nil
 }
 
+func (s *BlobMetadataStore) GetDispersalResponses(ctx context.Context, batchHeaderHash [32]byte) ([]*corev2.DispersalResponse, error) {
+	items, err := s.dynamoDBClient.Query(ctx, s.tableName, "PK = :pk AND begins_with(SK, :prefix)", commondynamodb.ExpressionValues{
```
dmanc, line 484: Do you think we will need pagination? Or is the operator set small enough not to worry about it?

jianoaix: It should be small enough to worry about

dmanc (approved):

disperser/common/v2/blobstore/dynamo_metadata_store.go
jianoaix: One optimization we may do in the future is to denormalize and just store all operators' responses for a batch in a single message (and then this request will just need a lookup, no need for index)

dmanc: I wonder if that could hit size limit on DynamoDB for a record

dmanc: https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/ServiceQuotas.html, says around 400kKB

jianoaix: That's good angle. It should be sufficient I think: the response size is about 160 bytes (if no error message, which can be a variable length), which is about 2500 operators. We need to check it again if we take this approach (e.g. maybe we add more info to the response).

PR #1080, opened 2025-01-08
Add metrics to encoder v2

dmanc:

Brings metrics changes made from V1 encoder to V2 encoder. In order to support certain metrics from V1 we need to send the blob size as part of the request.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


disperser/encoder/server_v2.go
```diff
@@ -31,23 +35,38 @@ type EncoderServerV2 struct {
 	logger      logging.Logger
 	prover      encoding.Prover
 	metrics     *Metrics
+	grpcMetrics *grpcprom.ServerMetrics
 	close       func()
 
 	runningRequests chan struct{}
-	requestPool     chan struct{}
+	requestPool     chan blobRequest
```
jianoaix, line 42: nit: can we rename "pool" to "queue" (e.g. queueingRequests, and ServerConfig.RequestQueueSize) so it's consistent with the rest parts?

jianoaix (approved):

disperser/encoder/server_v2.go
dmanc: Done V2 code will use new request queue flag. Leaving the V1 server code as is though to prevent having to update the config values.

jianoaix (approved):

PR #1082, opened 2025-01-08
Added missing v2 protobuf documentation.

cody-littley:

Adds missing v2 protobuf documentation.


api/proto/common/common.proto
```diff
@@ -2,6 +2,10 @@ syntax = "proto3";
 package common;
 option go_package = "github.com/Layr-Labs/eigenda/api/grpc/common";
 
+// TODO: I generated the comments in this file using copilot. Reviewers, please verify accuracy,
+// and after you do I will remove this comment. This PR should not merge until this comment is removed.
```
cody-littley, line 6: Reviewers, please pay extra close attention to this file. I wasn't sure what to write so I went with copilot's suggestion, and it's quite possible that the comments are incorrect.

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -16,58 +16,73 @@ service Disperser {
...
+  // Although the length commitment is required to be a power of 2, the length of this byte array is not required to
+  // be a power of 2 as long as the length does not exceed the length commitment. The maximum length of the data is
+  // 16MiB. (In the future, the 16MiB limit may be increased, but this is not guaranteed to happen.)
+  //
+  // Every 32 bytes of data is interpreted as an integer in big endian format where the lower address has more
+  // significant bits. The integer must stay in the valid range to be interpreted as a field element on the bn254 curve.
+  // The valid range is 0 <= x < 21888242871839275222246405745257275088548364400416034343698204186575808495617.
   // If any one of the 32 bytes elements is outside the range, the whole request is deemed as invalid, and rejected.
   bytes data = 1;
+  // The header contains metadata about the blob.
```
jianoaix, line 42: Do users need to specify all quorums (including required ones), or just custom quorums?

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -101,21 +116,22 @@ message GetPaymentStateReply {
 // - FAILED
 // - INSUFFICIENT_SIGNATURES
 enum BlobStatus {
+  // UNKNOWN means that the status of the blob is unknown.
```
jianoaix, line 119: Will UNKNOWN be transitioned to a different state?

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -153,25 +169,39 @@ message Attestation {
...
 message Reservation {
+  // rate limit for the account
   uint64 symbols_per_second = 1;
+  // start timestamp of the reservation
   uint32 start_timestamp = 2;
+  // end timestamp of the reservation
   uint32 end_timestamp = 3;
+  // quorums allowed to make reserved dispersals
   repeated uint32 quorum_numbers = 4;
+  // quorum splits between allowed quorums
```
jianoaix, line 196: What is "split"?

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -153,25 +169,39 @@ message Attestation {
...
   repeated uint32 quorum_splits = 5;
 }
 
-// BinRecord is the usage record of an account in a bin. The API should return the active bin 
+// BinRecord is the usage record of an account in a bin. The API should return the active bin
```
jianoaix, line 200: What is "bin"?

api/proto/retriever/v2/retriever.proto
```diff
@@ -33,6 +34,7 @@ message BlobRequest {
 	uint32 quorum_id = 3;
```
jianoaix, line 34: Does this param make any difference? Shouldn't it return the same blob no matter which quorum is specified for that blob?

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -16,58 +16,73 @@ service Disperser {
...
+  // BlobVerificationInfo is the information needed to verify the inclusion of a blob in a batch.
   BlobVerificationInfo blob_verification_info = 3;
 }
 
-// Utility method used to generate the commitment of blob given its data.
-// This can be used to construct BlobHeader.commitment
+// A request to generate the commitment of a blob via BlobCommitmentRequest().
+// This can be used to construct a BlobHeader.commitment.
 message BlobCommitmentRequest {
+  // The blob data.
```
jianoaix, line 73: "The blob data to compute commitment for"?

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -16,58 +16,73 @@ service Disperser {
...
   bytes data = 1;
 }
 
+// A reply to a BlobCommitmentRequest.
```
jianoaix, line 77: nit: "A reply" (or "A request" above) sounds like an instance of the reply, but this is just describing the message

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -16,58 +16,73 @@ service Disperser {
...
 }
 
+// BlobStatusReply is the reply to a BlobStatusRequest.
 message BlobStatusReply {
   // The status of the blob.
   BlobStatus status = 1;
   // The signed batch
   SignedBatch signed_batch = 2;
+  // BlobVerificationInfo is the information needed to verify the inclusion of a blob in a batch.
   BlobVerificationInfo blob_verification_info = 3;
```
jianoaix, line 67: This probably be easier to understand if it's named as something like "BlobInclusionInfo"

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -16,58 +16,73 @@ service Disperser {
...
 message BlobStatusRequest {
+  // The unique identifier for the blob.
   bytes blob_key = 1;
 }
 
+// BlobStatusReply is the reply to a BlobStatusRequest.
 message BlobStatusReply {
   // The status of the blob.
   BlobStatus status = 1;
   // The signed batch
```
jianoaix, line 64: It's useful for users to know that the following two fields are conditionally set (only for some BlobStatus)

api/proto/common/common.proto
```diff
@@ -2,6 +2,10 @@ syntax = "proto3";
...
+
+// G1Commitment represents a commitment to a field element in the G1 group of the BLS12-381 curve.
```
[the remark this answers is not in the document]
cody-littley: Changed to the following:

```
// A KZG commitment to a G1 point.
```

api/proto/disperser/v2/disperser_v2.proto
cody-littley: change made

api/proto/disperser/v2/disperser_v2.proto
cody-littley: Changed to `// The input for a BlobCommitmentRequest().` and `// The result of a BlobCommitmentRequest.`

api/proto/disperser/v2/disperser_v2.proto
cody-littley: @ian-shim are you ok with this rename? If "yes", let me know and I'll do the refactor.

api/proto/disperser/v2/disperser_v2.proto
cody-littley: Based on my reading of the code, it looks like the last two fields are set only when the blob is `CERTIFIED`. @ian-shim is this correct?

```
  // The signed batch. Unset if the status is not CERTIFIED.
  SignedBatch signed_batch = 2;
  // BlobVerificationInfo is the information needed to verify the inclusion of a blob in a batch.
  // Unset if the status is not CERTIFIED.
  BlobVerificationInfo blob_verification_info = 3;
```

api/proto/common/common.proto
```diff
@@ -12,9 +16,13 @@ message G1Commitment {
 // BlobCommitment represents commitment of a specific blob, containing its
 // KZG commitment, degree proof, the actual degree, and data length in number of symbols.
 message BlobCommitment {
+  // A commitment to the blob data.
   bytes commitment = 1;
+  // A commitment to the degree of the polynomial used to generate the blob commitment.
   bytes length_commitment = 2;
+  // A proof that the degree of the polynomial used to generate the blob commitment is valid.
```
jianoaix, line 23: It seems not quite right. I think this is commitment to the shift polynomial (`x^(srsorder - length)`). cc @bxue-l2 to confirm.

bxue-l2: The comment tries to state the purpose for the proof, which is correct.

The comment for length_commitment is a bit confusing. I will suggest something like

// A commitment to the blob data on G2, used to work with length_proof
// such that the claimed length below is verifiable
length_commitment

// A proof that the degree of the polynomial used to generate the blob commitment is valid.
// It is computed such that the polynomial is lifted up to the highest monomial term (this line optional)
length_proof

// The length specifies the degree of the polynomial used to generate the blob commitment. The length 
// must equal to the degree + 1, and it must be a power of 2.
length = 4;

api/proto/disperser/v2/disperser_v2.proto
cody-littley: Talked to Ian. All quorums must be specified. Documentation updated as follows:

```
  // quorum_numbers is the list of quorum numbers that the blob is part of. 
  // All quorums must be specified (including required quorums).
  repeated uint32 quorum_numbers = 2;
```

api/proto/disperser/v2/disperser_v2.proto
cody-littley: `UNKNOWN` is a terminal status. Added to the list of terminal statuses, and also added this message:

```
  // UNKNOWN means that the status of the blob is unknown. 
  // This is a catch all and should not be encountered absent a bug.
  UNKNOWN = 0;
```

api/proto/disperser/v2/disperser_v2.proto
cody-littley:
```
// quorum splits describes how the payment is split among the quorums
repeated uint32 quorum_splits = 5;
```

api/proto/disperser/v2/disperser_v2.proto
cody-littley: Hope has since renamed this to `PeriodRecord`.

api/proto/retriever/v2/retriever.proto
cody-littley: Yes. This determines which quorum the retriever will go to to fetch the blob. If the blob retrieval is successful, then different quorums would return the same data. But if the quorum specified here doesn't possess the blob, the request will not be successful.

api/proto/disperser/v2/disperser_v2.proto
cody-littley: Talked to Ian to confirm.

api/proto/common/common.proto
cody-littley: Thanks @bxue-l2! I've incorporated these changes.

jianoaix: "A commitment to the blob data with G2 SRS", do you mean it's same as the kzg commitment above (on G1), just that it's computed on G2?

api/proto/common/common.proto
```diff
@@ -2,6 +2,10 @@ syntax = "proto3";
 package common;
 option go_package = "github.com/Layr-Labs/eigenda/api/grpc/common";
 
+// TODO: I generated the comments in this file using copilot. Reviewers, please verify accuracy,
```
jianoaix, line 5: Please remove

(btw you may just comment on the file to give reviewers notice instead of writing in the code)

jianoaix (approved):
LGTM

We may iterate further from @samlaf PR

api/proto/common/common.proto
```diff
@@ -2,6 +2,10 @@ syntax = "proto3";
...
+// and after you do I will remove this comment. This PR should not merge until this comment is removed.
+
+// A KZG commitment to a G1 point.
```
jianoaix, line 8: I think G1 point *is* a commitment, not that this is a commitment *to* G1

api/proto/node/v2/node_v2.proto
```diff
@@ -6,14 +6,18 @@ option go_package = "github.com/Layr-Labs/eigenda/api/grpc/node/v2";
 // The EigenDA Node implements two services, Dispersal and Retrieval, as defined below,
 // for better security and separation of concerns.
 
-// WARNING: the following RPCs are experimental and subject to change.
+// Dispersal is utilized to disperse chunk data. The disperser calls these RPCs to store chunks on individual DA nodes.
 service Dispersal {
+  // StoreChunks stores a batch of chunks on the Node.
   rpc StoreChunks(StoreChunksRequest) returns (StoreChunksReply) {}
+  // NodeInfo fetches metadata about the node.
   rpc NodeInfo(NodeInfoRequest) returns (NodeInfoReply) {}
```
jianoaix, line 14: Side note: maybe we should rename this to `GetNodeInfo` (to use a verb which is more consistent naming)

api/proto/common/common.proto
jianoaix: Ping here, if it's unclear even to us, it won't be clear to users of this API

api/proto/disperser/v2/disperser_v2.proto
jianoaix: Sg, it's much clear:)

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -16,58 +14,73 @@ service Disperser {
...
   bytes data = 1;
+  // The header contains metadata about the blob.
   common.v2.BlobHeader blob_header = 2;
 }
 
+// A reply to a DisperseBlob request.
 message DisperseBlobReply {
   // The status of the blob associated with the blob key.
   BlobStatus result = 1;
+  // The unique identifier for the blob. Unique even if blob data is the identical.
```
jianoaix, line 47: This is the first time the users see blob key. We may probably better document it eg:
- how long is this key? (32 bytes, length always fixed)
- how is it defined? (hash of BlobHeader)
etc

api/proto/common/common.proto
samlaf: Agree with Jian here, I think I would interpret it the same. Is length commitment really a commitment of the blob_data? Might be useful to link to our specs or docs or some place where this is explained using mathematical symbol, which is probably easier to describe than in words here.

api/proto/disperser/v2/disperser_v2.proto
samlaf: Agree. I'll suggest the same edit here I suggested for the blob_key in GetChunksRequest

samlaf:
```suggestion
  // The unique 32 bytes identifier for the blob.
  // The blob_key is the keccak hash of the rlp serialization of the BlobHeader, as computed here:
  // https://github.com/Layr-Labs/eigenda/blob/0f14d1c90b86d29c30ff7e92cbadf2762c47f402/core/v2/serialization.go#L30
  // The blob_key must thus be unique for every request, even if the same blob is being disperser.
  // Meaning the blob_header must be different for each request.
```

samlaf: What happens if we submit the exact same request to DisperseBlob? Does the Disperser return an error? Or does it just return the blob status of the blob identified by that request?

api/proto/common/v2/common_v2.proto
```diff
@@ -3,21 +3,26 @@ package common.v2;
...
+  // payment_header contains payment information for the blob
   common.PaymentHeader payment_header = 4;
-
   // signature over keccak hash of the blob_header that can be verified by blob_header.account_id
   bytes signature = 5;
 }
 
 // BlobCertificate is what gets attested by the network
 message BlobCertificate {
+  // blob_header is the header of the blob
```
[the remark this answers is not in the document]
jianoaix: Yea it's not that informative. We should try to actually convey useful things in documentation.

api/proto/disperser/v2/disperser_v2.proto
jianoaix: @cody-littley can we add what are required quorums v.s. custom quorums, and the current required quorums (0/ETH, 1/EIGEN)?

cody-littley: Updated comment as follows:
```
  // quorum_numbers is the list of quorum numbers that the blob is part of.
  // All quorums must be specified (including required quorums).
  // 
  // The following quorums are currently required:
  // - 0: ETH
  // - 1: EIGEN
  repeated uint32 quorum_numbers = 2;
```

api/proto/common/common.proto
cody-littley: removed

api/proto/common/common.proto
cody-littley: Changed to `// A KZG commitment`

api/proto/node/v2/node_v2.proto
cody-littley: Renamed to `GetNodeInfo()` (both Retrieval and Dispersal services).

I'm unsure why we duplicate this. @ian-shim can you comment?

api/proto/disperser/v2/disperser_v2.proto
cody-littley: > What happens if we submit the exact same request to DisperseBlob?

That's a good question. @jianoaix @jianoaix do you know the answer?

api/proto/common/v2/common_v2.proto
```diff
@@ -3,21 +3,26 @@ package common.v2;
 import "common/common.proto";
 option go_package = "github.com/Layr-Labs/eigenda/api/grpc/common/v2";
 
-
+// BlobHeader is the header of a blob
 message BlobHeader {
   // Blob version
   uint32 version = 1;
+  // quorum_numbers is the list of quorum numbers that the blob is part of.
+  // All quorums must be specified (including required quorums).
```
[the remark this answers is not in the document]
cody-littley: Suggested change made.

api/proto/common/v2/common_v2.proto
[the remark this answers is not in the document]
cody-littley: Changed to the following:

```
// blob_header contains metadata about the blob. The hash of this header is used to compute the blob key.
```

api/proto/common/v2/common_v2.proto
```diff
@@ -3,21 +3,26 @@ package common.v2;
...
+  // commitment is the KZG commitment of the blob
   common.BlobCommitment commitment = 3;
+  // payment_header contains payment information for the blob
   common.PaymentHeader payment_header = 4;
-
   // signature over keccak hash of the blob_header that can be verified by blob_header.account_id
   bytes signature = 5;
 }
 
 // BlobCertificate is what gets attested by the network
```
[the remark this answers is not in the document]
cody-littley: change adopted

api/proto/common/v2/common_v2.proto
```diff
@@ -3,21 +3,26 @@ package common.v2;
...
 message BlobCertificate {
+  // blob_header is the header of the blob
   BlobHeader blob_header = 1;
+  // relays is the list of relays that are in custody of the blob
```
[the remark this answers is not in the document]
cody-littley: Change adopted.

disperser/apiserver/disperse_blob_v2.go
```diff
@@ -109,6 +109,11 @@ func (s *DispersalServerV2) validateDispersalRequest(ctx context.Context, req *p
 		return api.NewErrorInvalidArg("blob header must contain commitments")
 	}
 
+	commitmentLength := blobHeaderProto.GetCommitment().Length
+	if commitmentLength != encoding.NextPowerOf2(commitmentLength) {
+		return api.NewErrorInvalidArg("invalid commitment length, must be a power of 2")
+	}
```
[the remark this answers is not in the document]
cody-littley: Updated as follows:


```
	if blobHeaderProto.GetCommitment() == nil {
		return api.NewErrorInvalidArg("blob header must contain a commitment")
	}
	commitmentLength := blobHeaderProto.GetCommitment().GetLength()
	if commitmentLength == 0 || commitmentLength != encoding.NextPowerOf2(commitmentLength) {
		return api.NewErrorInvalidArg("invalid commitment length, must be a power of 2")
	}
```

disperser/apiserver/server_v2_test.go
```diff
@@ -544,3 +544,51 @@ func newTestServerV2(t *testing.T) *testComponents {
 		Peer:              p,
 	}
 }
+
+func TestInvalidLength(t *testing.T) {
```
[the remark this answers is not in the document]
cody-littley: This is intentional. @mooselumph asked that I ensure that we catch this specific error scenario when a blob is initially dispersed. The concern was that we might sometimes accept a blob that the encoder would eventually reject, and this pattern can yield a bad user experience. Since I wrote new code I thought it would be wise to exercise it in a unit test.

api/proto/node/v2/node_v2.proto
```diff
@@ -59,32 +63,42 @@ message StoreChunksRequest {
...
 
+// StoreChunksReply is the message type used to respond to a StoreChunks() RPC.
 message StoreChunksReply {
+  // a custody signature of the received batch
   bytes signature = 1;
 }
 
+// The parameter for the GetChunks() RPC.
 message GetChunksRequest {
+  // The unique identifier for the blob the chunks are being requested for.
```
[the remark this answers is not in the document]
cody-littley: Change adopted.

api/proto/node/v2/node_v2.proto
```diff
@@ -59,32 +63,42 @@ message StoreChunksRequest {
   bytes signature = 3;
 }
 
+// StoreChunksReply is the message type used to respond to a StoreChunks() RPC.
 message StoreChunksReply {
+  // a custody signature of the received batch
```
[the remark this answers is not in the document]
cody-littley: Just the chunks. I've changed this comment to read `// a custody signature of the received chunks`

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -93,29 +106,32 @@ message GetPaymentStateReply {
...
 // - FAILED
 // - INSUFFICIENT_SIGNATURES
+// - UNKNOWN
 enum BlobStatus {
+  // UNKNOWN means that the status of the blob is unknown.
+  // This is a catch all and should not be encountered absent a bug.
   UNKNOWN = 0;
 
-  // QUEUED means that the blob has been queued by the disperser for processing
+  // QUEUED means that the blob has been queued by the disperser for processing.
```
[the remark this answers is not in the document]
cody-littley: change adopted

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -93,29 +106,32 @@ message GetPaymentStateReply {
...
   QUEUED = 1;
 
-  // ENCODED means that the blob has been encoded and is ready to be dispersed to DA Nodes
+  // ENCODED means that the blob has been encoded and is ready to be dispersed to DA Nodes.
   ENCODED = 2;
 
-  // CERTIFIED means the blob has been dispersed and attested by the DA nodes
+  // CERTIFIED means the blob has been dispersed and attested by the DA nodes.
```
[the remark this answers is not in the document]
cody-littley: Change adopted.

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -93,29 +106,32 @@ message GetPaymentStateReply {
...
   CERTIFIED = 3;
 
-  // FAILED means that the blob has failed permanently
+  // FAILED means that the blob has failed permanently.
```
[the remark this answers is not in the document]
cody-littley: @jianoaix @jianoaix Can you comment on this?

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -16,58 +14,73 @@ service Disperser {
...
 
+// The result of a BlobCommitmentRequest().
 message BlobCommitmentReply {
+  // The commitment of the blob.
   common.BlobCommitment blob_commitment = 1;
 }
 
 // GetPaymentStateRequest contains parameters to query the payment state of an account.
 message GetPaymentStateRequest {
+  // The ID of the account being queried.
```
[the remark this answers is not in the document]
hopeyen: AccountID is the eth wallet address of the user, and is to authenticate `GetPaymentStateRequest`. This is the same as AccountID included in payment metadata and the same AccountID generated by the blob request signer.
We could add something like timestamp as part of authentication but was considered not a priority by the team

api/proto/disperser/v2/disperser_v2.proto
ian-shim: I don't have a strong opinion on this. Summoning the naming council @samlaf @mooselumph @litt3

api/proto/disperser/v2/disperser_v2.proto
ian-shim: Yes, signed batch and blob verification info are set only under CERTIFIED status.

api/proto/node/v2/node_v2.proto
ian-shim: They can technically run those services separately on different machines

api/proto/disperser/v2/disperser_v2.proto
ian-shim: If the exact same request was made again after a successful initial request (not necessarily successful dispersal, but successful dispersal request), the second request will return `AlreadyExists` error

api/proto/disperser/v2/disperser_v2.proto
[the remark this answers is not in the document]
ian-shim: I believe they're still charged for the dispersal and yes they need another dispersal with different salt. Correct me if I'm wrong @hopeyen

api/proto/disperser/v2/disperser_v2.proto
samlaf: yeessssssss +10000 to this. @litt3 and I had this name change already proposed in austin's diagram.
image

api/grpc/common/common.pb.go
```diff
@@ -84,10 +85,18 @@ type BlobCommitment struct {
...
-	Commitment       []byte `protobuf:"bytes,1,opt,name=commitment,proto3" json:"commitment,omitempty"`
+	// A commitment to the blob data.
+	Commitment []byte `protobuf:"bytes,1,opt,name=commitment,proto3" json:"commitment,omitempty"`
+	// A commitment to the blob data with G2 SRS, used to work with length_proof
+	// such that the claimed length below is verifiable.
 	LengthCommitment []byte `protobuf:"bytes,2,opt,name=length_commitment,json=lengthCommitment,proto3" json:"length_commitment,omitempty"`
-	LengthProof      []byte `protobuf:"bytes,3,opt,name=length_proof,json=lengthProof,proto3" json:"length_proof,omitempty"`
-	Length           uint32 `protobuf:"varint,4,opt,name=length,proto3" json:"length,omitempty"`
+	// A proof that the degree of the polynomial used to generate the blob commitment is valid.
+	// It is computed such that the coefficient of the polynomial is committing with the G2 SRS
```
litt3, line 94: I'm personally not able to understand this explanation: not sure if I'm just lacking background knowledge, or if this explanation could be further fleshed out

api/grpc/disperser/v2/disperser_v2.pb.go
```diff
@@ -25,26 +25,31 @@ const (
...
 // The status of a blob can be queried by the client using the GetBlobStatus API.
-// Intermediate states are states that the blob can be in while being processed, and it can be updated to a differet state:
+// Intermediate states are states that the blob can be in while being processed, and it can be updated to a different state:
 // - QUEUED
 // - ENCODED
 // Terminal states are states that will not be updated to a different state:
 // - CERTIFIED
 // - FAILED
 // - INSUFFICIENT_SIGNATURES
+// - UNKNOWN
```
litt3, line 35: `UNKNOWN` is the first element in the enum, but the last element in this doc

api/grpc/disperser/v2/disperser_v2.pb.go
```diff
@@ -25,26 +25,31 @@ const (
...
+	// ENCODED means that the blob has been Reed-Solomon encoded into chunks and is ready to be dispersed to DA Nodes.
 	BlobStatus_ENCODED BlobStatus = 2
-	// CERTIFIED means the blob has been dispersed and attested by the DA nodes
+	// CERTIFIED means the blob has been dispersed and attested by the DA nodes.
 	BlobStatus_CERTIFIED BlobStatus = 3
-	// FAILED means that the blob has failed permanently
+	// FAILED means that the blob has failed permanently.
 	BlobStatus_FAILED BlobStatus = 4
-	// INSUFFICIENT_SIGNATURES means that the blob has failed to gather sufficient attestation
+	// INSUFFICIENT_SIGNATURES means that the blob has failed to gather sufficient attestation.
```
litt3, line 52: Is `INSUFFICIENT_SIGNATURES` also a mode of failure? I.e., is this also "permanent"?

api/grpc/disperser/v2/disperser_v2.pb.go
```diff
@@ -156,14 +166,21 @@ func (x *DisperseBlobRequest) GetBlobHeader() *v2.BlobHeader {
...
 
 	// The status of the blob associated with the blob key.
-	Result  BlobStatus `protobuf:"varint,1,opt,name=result,proto3,enum=disperser.v2.BlobStatus" json:"result,omitempty"`
-	BlobKey []byte     `protobuf:"bytes,2,opt,name=blob_key,json=blobKey,proto3" json:"blob_key,omitempty"`
+	Result BlobStatus `protobuf:"varint,1,opt,name=result,proto3,enum=disperser.v2.BlobStatus" json:"result,omitempty"`
+	// The unique 32 byte identifier for the blob.
+	//
+	// The blob_key is the keccak hash of the rlp serialization of the BlobHeader, as computed here:
+	// https://github.com/Layr-Labs/eigenda/blob/0f14d1c90b86d29c30ff7e92cbadf2762c47f402/core/v2/serialization.go#L30
+	// The blob_key must thus be unique for every request, even if the same blob is being disperser.
```
litt3, line 181: is being disperser -> is being dispersed

api/grpc/node/v2/node_v2_grpc.pb.go
```diff
@@ -20,15 +20,17 @@ const _ = grpc.SupportPackageIsVersion7
...
 )
 
 // DispersalClient is the client API for Dispersal service.
 //
 // For semantics around ctx use and closing/ending streaming RPCs, please refer to https://pkg.go.dev/google.golang.org/grpc/?tab=doc#ClientConn.NewStream.
 type DispersalClient interface {
+	// StoreChunks stores a batch of chunks on the Node.
 	StoreChunks(ctx context.Context, in *StoreChunksRequest, opts ...grpc.CallOption) (*StoreChunksReply, error)
-	NodeInfo(ctx context.Context, in *NodeInfoRequest, opts ...grpc.CallOption) (*NodeInfoReply, error)
+	// NodeInfo fetches metadata about the node.
```
litt3, line 32: NodeInfo -> GetNodeInfo

api/grpc/node/v2/node_v2_grpc.pb.go
```diff
@@ -61,8 +63,10 @@ func (c *dispersalClient) NodeInfo(ctx context.Context, in *NodeInfoRequest, opt
 // All implementations must embed UnimplementedDispersalServer
 // for forward compatibility
 type DispersalServer interface {
+	// StoreChunks stores a batch of chunks on the Node.
 	StoreChunks(context.Context, *StoreChunksRequest) (*StoreChunksReply, error)
-	NodeInfo(context.Context, *NodeInfoRequest) (*NodeInfoReply, error)
+	// NodeInfo fetches metadata about the node.
```
litt3, line 68: NodeInfo -> GetNodeInfo

api/proto/node/v2/node_v2.proto
```diff
@@ -6,17 +6,21 @@ option go_package = "github.com/Layr-Labs/eigenda/api/grpc/node/v2";
 // The EigenDA Node implements two services, Dispersal and Retrieval, as defined below,
 // for better security and separation of concerns.
 
-// WARNING: the following RPCs are experimental and subject to change.
+// Dispersal is utilized to disperse chunk data. The disperser calls these RPCs to store chunks on individual DA nodes.
 service Dispersal {
+  // StoreChunks stores a batch of chunks on the Node.
   rpc StoreChunks(StoreChunksRequest) returns (StoreChunksReply) {}
-  rpc NodeInfo(NodeInfoRequest) returns (NodeInfoReply) {}
+  // NodeInfo fetches metadata about the node.
```
litt3, line 13: NodeInfo -> GetNodeInfo

api/proto/disperser/v2/disperser_v2.proto
ian-shim: To be clear, v2 does not explicitly enforce the required quorums. It only enforces that price for using subset of the required quorums is the same as using all required quorums. So technically, users are free to choose whatever set of quorums as necessary (with custom verification method).

jianoaix: This sounds more complicated than before by giving more choice to users. I would prefer an enforcement and reject requests missing required quorums.

api/proto/disperser/v2/disperser_v2.proto
jianoaix: yea it's more accurate. I had to have a few turns in mind to comprehend BlobVerficationInfo.

Please note the renaming may affect more than "eigenda" and "eigenda-proxy" repos, e.g. https://github.com/Layr-Labs/hokulea/blob/master/crates/eigenda/src/certificate.rs#L50

api/proto/common/v2/common_v2.proto
```diff
@@ -3,21 +3,39 @@ package common.v2;
 import "common/common.proto";
 option go_package = "github.com/Layr-Labs/eigenda/api/grpc/common/v2";
 
-
+// BlobHeader contains the information needed to disperse a blob to the EigenDA network.
```
jianoaix, line 6: IMO this one-liner could be more about what does BlobHeader describe (definitive properties of a blob), rather than where it's used

api/proto/common/v2/common_v2.proto
```diff
@@ -3,21 +3,39 @@ package common.v2;
...
+// It can be thought of as an "eigenDA tx", in that it plays a purpose similar to an eth_tx to disperse a 4844 blob.
```
jianoaix, line 7: "It" means BlobHeader? I think Blob (which is different than BlobHeader) is analogous to a txn

api/proto/common/common.proto
bxue-l2: LengthCommitment is just the commitment in G2 on the same blbo

api/proto/disperser/v2/disperser_v2.proto
samlaf: That's fine, that'll teach @bxue-l2 for not using prost to generate his proto structs :P

api/proto/disperser/v2/disperser_v2.proto
samlaf: How does pricing work here? Are they charged per quorum? Where is the pricing defined?

api/proto/common/common.proto
samlaf: Hmm indeed bowen just pointed me to https://github.com/Layr-Labs/eigenda/blob/fcd50aeaa3231779506413516986b2be161b48eb/encoding/kzg/prover/gnark/commitments.go#L53 @jianoaix , which is called [here](https://github.com/Layr-Labs/eigenda/blob/fcd50aeaa3231779506413516986b2be161b48eb/encoding/kzg/prover/parametrized_prover.go#L142) using same input as that for the commitment.

api/proto/common/v2/common_v2.proto
samlaf: No the blob is the payload. the blobheader is the tx, specifying price, commitment, etc.

api/proto/common/v2/common_v2.proto
```diff
@@ -3,21 +3,39 @@ package common.v2;
...
+// Note that a call to DisperseBlob requires the blob and the blobHeader, which is similar to how dispersing a blob
+// to ethereum requires sending a tx who's data contains the hash of the kzg commit of the blob, which is
```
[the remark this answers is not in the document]
samlaf: Ah good point, agree about moving. As for suggestions, its much easier for the PR author if you make actual suggestions that they can just commit to @jianoaix. Something like I did above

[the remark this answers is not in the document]
samlaf:
```suggestion
// to ethereum requires sending a tx whose data contains the hash of the kzg commit of the blob, which is
```

api/proto/common/v2/common_v2.proto
```diff
@@ -3,21 +3,39 @@ package common.v2;
...
-
   // signature over keccak hash of the blob_header that can be verified by blob_header.account_id
   bytes signature = 5;
 }
 
-// BlobCertificate is what gets attested by the network
+// BlobCertificate is what gets attested by the network.
+// It gets constructed by the Disperser to which the DisperseBlob request was submitted.
 message BlobCertificate {
+  // blob_header contains metadata about the blob. The hash of this header is used to compute the blob key.
```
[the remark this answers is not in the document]
samlaf:
```suggestion
  // blob_header contains metadata about the blob.
```

api/proto/disperser/v2/disperser_v2.proto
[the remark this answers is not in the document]
samlaf: Can we add that as a comment @cody-littley ?

api/grpc/node/v2/node_v2_grpc.pb.go
samlaf:
```suggestion
	// GetNodeInfo fetches metadata about the node.
```

api/proto/common/v2/common_v2.proto
jianoaix: Now I don't know what is blob/blobheader/payload/txn... https://en.wikipedia.org/wiki/Semantic_satiation

api/proto/common/common.proto
samlaf: @jianoaix turns out its the proof that is a commitment of the shift poly you were mentioning: https://github.com/Layr-Labs/eigenda/blob/fcd50aeaa3231779506413516986b2be161b48eb/encoding/kzg/prover/gnark/commitments.go#L34

api/proto/disperser/v2/disperser_v2.proto
samlaf: Oh snap that's a good point always trips me that the errors are not represented in the proto file itself... easy to forget about them. This is a good reminder that we should be documenting the grpc errors that the different endpoints can return, and on which conditions.

For eg here @cody-littley can you add a comment saying that if the same blob_header is submitted, a grpc [AlreadyExists](https://github.com/grpc/grpc-go/blob/f35fb347b7a7b57b92d25ca2e012dd8359b2a305/codes/codes.go#L85) error is returned.

For reference the error is returned [here](https://github.com/Layr-Labs/eigenda/blob/fcd50aeaa3231779506413516986b2be161b48eb/disperser/apiserver/disperse_blob_v2.go#L69).

api/proto/disperser/v2/disperser_v2.proto
hopeyen: As for payments, reservation allows custom quorum numbers and doesn't enforce required quorums, but on-demand payments is only allowed for on-demand quorums. 

To Sam's question, there's no specific logic on pricing per quorum. For on-demand, pricing is currently defined across all required quorums in the contracts, which is a simple fixed price per symbol. For reservation (set by EigenDA governance only), we left pricing to the BD team and is not defined at all. 


(Pricing for reservations is in the making for permissionless reservations. Curious to learn what you think makes sense)

api/proto/disperser/v2/disperser_v2.proto
cody-littley: How long do we "remember" a previously submitted blob's hash? Probably not forever, right? What if we failed to disperse that blob due to some sort of error that puts the blob into a terminal state? Will that prevent us from submitting that same blob again in the future?

api/proto/disperser/v2/disperser_v2.proto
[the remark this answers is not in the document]
cody-littley: Updated comment.

> // The ID of the account being queried. This account ID is an eth wallet address of the user.
  string account_id = 1;

TODO commit link

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -93,29 +106,32 @@ message GetPaymentStateReply {
...
+  // UNKNOWN means that the status of the blob is unknown.
+  // This is a catch all and should not be encountered absent a bug.
   UNKNOWN = 0;
 
-  // QUEUED means that the blob has been queued by the disperser for processing
+  // QUEUED means that the blob has been queued by the disperser for processing.
   QUEUED = 1;
 
-  // ENCODED means that the blob has been encoded and is ready to be dispersed to DA Nodes
+  // ENCODED means that the blob has been encoded and is ready to be dispersed to DA Nodes.
```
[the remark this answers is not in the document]
cody-littley: change adopted

(I did this but forgot to comment, not sure which commit contains the changes)

api/proto/disperser/v2/disperser_v2.proto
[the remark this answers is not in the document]
cody-littley:
```
  // Note that attempting to disperse a blob with the same blob key as a previously dispersed blob may cause
  // the disperser to reject the blob (DisperseBlob() RPC will return an error).
  bytes blob_key = 2;
```

TODO commit link

api/grpc/common/v2/common_v2.pb.go
```diff
@@ -21,16 +21,29 @@ const (
...
 	unknownFields protoimpl.UnknownFields
 
 	// Blob version
-	Version       uint32                 `protobuf:"varint,1,opt,name=version,proto3" json:"version,omitempty"`
-	QuorumNumbers []uint32               `protobuf:"varint,2,rep,packed,name=quorum_numbers,json=quorumNumbers,proto3" json:"quorum_numbers,omitempty"`
-	Commitment    *common.BlobCommitment `protobuf:"bytes,3,opt,name=commitment,proto3" json:"commitment,omitempty"`
-	PaymentHeader *common.PaymentHeader  `protobuf:"bytes,4,opt,name=payment_header,json=paymentHeader,proto3" json:"payment_header,omitempty"`
+	Version uint32 `protobuf:"varint,1,opt,name=version,proto3" json:"version,omitempty"`
+	// quorum_numbers is the list of quorum numbers that the blob is part of.
+	// All quorums must be specified (including required quorums).
```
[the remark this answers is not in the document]
cody-littley: @ian-shim @jianoaix can you comment on this?

api/grpc/common/v2/common_v2.pb.go
```diff
@@ -21,16 +21,29 @@ const (
...
+	//
+	// The following quorums are currently required:
+	// - 0: ETH
+	// - 1: EIGEN
+	QuorumNumbers []uint32 `protobuf:"varint,2,rep,packed,name=quorum_numbers,json=quorumNumbers,proto3" json:"quorum_numbers,omitempty"`
+	// commitment is the KZG commitment of the blob
```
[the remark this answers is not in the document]
cody-littley: change made in source file

TODO commit link

api/grpc/common/v2/common_v2.pb.go
```diff
@@ -102,14 +115,21 @@ func (x *BlobHeader) GetSignature() []byte {
...
 
-// BlobCertificate is what gets attested by the network
+// BlobCertificate is what gets attested by the network.
+// It gets constructed by the Disperser to which the DisperseBlob request was submitted.
 type BlobCertificate struct {
 	state         protoimpl.MessageState
 	sizeCache     protoimpl.SizeCache
 	unknownFields protoimpl.UnknownFields
 
+	// blob_header contains metadata about the blob. The hash of this header is used to compute the blob key.
```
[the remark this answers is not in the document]
cody-littley: changed to `The hash of this header is the blob key`

TODO commit

api/grpc/common/common.pb.go
cody-littley: This is also beyond me, copied from text suggested by @bxue-l2

api/grpc/disperser/v2/disperser_v2.pb.go
cody-littley: Yes, it is also a failure. Just a more specific failure. Anything other than `CERTIFIED` in the terminal status list is functionally the same thing. The purpose of multiple statuses is to provide context as to why the blob failed.

I've added the following style of comment to the other terminal statuses.

```
// This status is functionally equivalent to FAILED, but is used to indicate that the failure is due to an 
  // an inability to gather sufficient signatures.
```

TODO commit

api/grpc/disperser/v2/disperser_v2.pb.go
cody-littley: yes, fixed

TODO commit

api/grpc/node/v2/node_v2_grpc.pb.go
cody-littley: fixed

TODO commit

api/grpc/node/v2/node_v2_grpc.pb.go
cody-littley: fixed

TODO commit

api/proto/node/v2/node_v2.proto
cody-littley: fixed

TODO commit

api/proto/node/v2/node_v2.proto
```diff
@@ -59,32 +63,44 @@ message StoreChunksRequest {
...
 
+// The response to the GetChunks() RPC.
 message GetChunksReply {
   // All chunks the Node is storing for the requested blob per RetrieveChunksRequest.
   repeated bytes chunks = 1;
 }
 
-// Node info request
-message NodeInfoRequest {
+// The parameter for the NodeInfo() RPC.
```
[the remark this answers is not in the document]
cody-littley: fixed

TODO commit

api/proto/node/v2/node_v2.proto
```diff
@@ -6,17 +6,21 @@ option go_package = "github.com/Layr-Labs/eigenda/api/grpc/node/v2";
 // The EigenDA Node implements two services, Dispersal and Retrieval, as defined below,
 // for better security and separation of concerns.
 
-// WARNING: the following RPCs are experimental and subject to change.
+// Dispersal is utilized to disperse chunk data. The disperser calls these RPCs to store chunks on individual DA nodes.
```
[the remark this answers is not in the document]
cody-littley: dropped

TODO commit

api/proto/common/v2/common_v2.proto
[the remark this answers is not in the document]
cody-littley: dropped

TODO commit

api/proto/common/v2/common_v2.proto
cody-littley: I switched the wording as follows. Note that this paragraph moved into `DisperseBlobRequest` (as requested by a prior comment).

```
  // This header can be thought of as an "eigenDA tx", in that it plays a purpose similar to an eth_tx to disperse a
  // 4844 blob. Note that a call to DisperseBlob requires the blob and the blobHeader, which is similar to how
  // dispersing a blob to ethereum requires sending a tx whose data contains the hash of the kzg commit of the blob,
  // which is dispersed separately.
```

api/proto/disperser/v2/disperser_v2.proto
cody-littley: TODO open ticket for rename

api/proto/node/v2/node_v2.proto
cody-littley: TODO create ticket, do we want to merge these services?

api/proto/common/v2/common_v2.proto
```diff
@@ -3,21 +3,35 @@ package common.v2;
 import "common/common.proto";
 option go_package = "github.com/Layr-Labs/eigenda/api/grpc/common/v2";
 
-
+// BlobHeader contains the information needed to disperse a blob to the EigenDA network.
 message BlobHeader {
   // Blob version
   uint32 version = 1;
```
samlaf, line 9: We need to describe what this version is, and that it's linked to https://github.com/Layr-Labs/eigenda/blob/c4567f90e835678fae4749f184857dea10ff330c/contracts/src/interfaces/IEigenDAStructs.sol#L49
@0x0aa0 can you add some description explaining how this version is used wrt the onchain thresholds and params?

api/proto/common/common.proto
```diff
@@ -12,12 +13,21 @@ message G1Commitment {
...
   bytes length_proof = 3;
+  // The length specifies the degree of the polynomial used to generate the blob commitment. The length
+  // must equal to the degree + 1, and it must be a power of 2.
   uint32 length = 4;
 }
 
+// PaymentHeader contains payment information for a blob.
 message PaymentHeader {
   // The account ID of the disperser client. This should be a hex-encoded string of the ECSDA public key
   // corresponding to the key used by the client to sign the BlobHeader.
```
samlaf, line 33:
```suggestion
  // The account ID of the disperser client. This account ID is an eth wallet address of the user,
  // corresponding to the key used by the client to sign the BlobHeader.
```

samlaf: confirmed with @hopeyen that this is outdated and needs to be changed.

litt3 (approved):

api/proto/common/v2/common_v2.proto
cody-littley: Changed to the following:

```
// BlobHeader contains the information describing a blob and the way it is to be dispersed.
```

TODO commit

api/proto/common/v2/common_v2.proto
```diff
@@ -3,21 +3,39 @@ package common.v2;
...
   common.BlobCommitment commitment = 3;
+  // payment_header contains payment information for the blob
   common.PaymentHeader payment_header = 4;
-
   // signature over keccak hash of the blob_header that can be verified by blob_header.account_id
   bytes signature = 5;
 }
 
-// BlobCertificate is what gets attested by the network
+// BlobCertificate is what gets attested by the network.
```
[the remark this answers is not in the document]
cody-littley: Updated to read as follows:

```
// BlobCertificate contains a full description of a blob and how it is dispersed. Part of the certificate
// is provided by the blob submitter (i.e. the blob header), and part is provided by the disperser (i.e. the relays).
// Validator nodes eventually sign the blob certificate once they are in custody of the required chunks
// (note that the signature is indirect; validators sign the hash of a Batch, which contains the blob certificate).
message BlobCertificate {
```

TODO commit

api/proto/common/v2/common_v2.proto
```diff
@@ -3,21 +3,37 @@ package common.v2;
...
   // signature over keccak hash of the blob_header that can be verified by blob_header.account_id
   bytes signature = 5;
 }
 
-// BlobCertificate is what gets attested by the network
+// BlobCertificate contains a full description of a blob and how it is dispersed. Part of the certificate
+// is provided by the blob submitter (i.e. the blob header), and part is provided by the disperser (i.e. the relays).
+// Validator nodes eventually sign the blob certificate once they are in custody of the required chunks
+// (note that the signature is indirect; validators sign the hash of a Batch, which contains the blob certificate).
 message BlobCertificate {
```
jianoaix, line 29: Since we are re-evaluating the naming, I think we may consider this name as well. It may be something like `TBSBlobCertificate`, TBS stands for "To Be Signed".

It's a practice seen in https://www.rfc-editor.org/rfc/rfc5280#section-4.1.1
Where a actual certificate is like:
```
Certificate ::= SEQUENCE {
    tbsCertificate          TBSCertificate,
    signatureAlgorithm      AlgorithmIdentifier,
    signature               BIT STRING
  }
```

api/proto/common/v2/common_v2.proto
0x0aa0: Blob versions are pushed onchain by EigenDA governance in an append only fashion and store the maximum number of operators, number of chunks, and coding rate for a blob. On blob verification, these values are checked against supplied or default security thresholds to validate the security assumptions of the blobs availability.

cody-littley: Thanks! I've added this paragraph to the field's documentation.

api/proto/common/v2/common_v2.proto
cody-littley: I've created a ticket to revisit this. I'd prefer not to rename any additional things as a part of this PR. https://linear.app/eigenlabs/issue/EGDA-808/rename-blobcertificate

api/proto/node/v2/node_v2.proto
samlaf: @cody-littley think that's the wrong linear ticket

cody-littley: Couldn't find the original ticket (not sure what happened). Created a new one: https://linear.app/eigenlabs/issue/EGDA-809/getnodeinfo-rpc-appears-twice

PR #1101, opened 2025-01-13
Better printing during inabox failure mode

cody-littley:

Improve behavior if inabox fails due to `wget` not being installed.

### DIFF
--- a/inabox/deploy/deploy.go
+++ b/inabox/deploy/deploy.go
@@ -329,7 +329,7 @@ func (env *Config) RegisterBlobVersionAndRelays(ethClient common.EthClient) map[
 // TODO: Supply the test path to the runner utility
 func (env *Config) StartBinaries() {
 	changeDirectory(filepath.Join(env.rootPath, "inabox"))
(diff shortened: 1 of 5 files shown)


inabox/deploy/deploy.go
```diff
@@ -339,23 +339,23 @@ func (env *Config) StartBinaries() {
...
+	err := execCmd("./bin.sh", []string{"stop"}, []string{}, true)
 	if err != nil {
 		log.Panicf("Failed to stop binaries. Err: %s", err)
 	}
 }
 
 func (env *Config) StartAnvil() {
 	changeDirectory(filepath.Join(env.rootPath, "inabox"))
-	err := execCmd("./bin.sh", []string{"start-anvil"}, []string{})
+	err := execCmd("./bin.sh", []string{"start-anvil"}, []string{}, false) // printing output causes hang
```
jianoaix, line 350: because too many prints?

jianoaix (approved):

inabox/deploy/deploy.go
cody-littley: I am unsure what the root cause is. It may be something like that.

PR #1104, opened 2025-01-13
docs: add documentation for disperser v2 grpc api and related functions/structs

samlaf:

The v2 interfaces are missing documentation, which makes it hard for intergration to interface with core. Adding some ad-hoc documentation here for thing that I needed to lookup today, but would be useful to do a uniform pass on the API to make sure its documentation is complete at some point.

Just learned about https://github.com/Layr-Labs/eigenda/pull/1082, so therre might be some redundancy here. Probably should merge that one first (it's documents every more broadly) and then I'll rebase mine on top (I have more precise documentation for much fewer things).

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proto/common/common.proto
```diff
@@ -10,11 +13,20 @@ message G1Commitment {
...
+  // Concatenation of the x and y coordinates of `common.G1Commitment`.
   bytes commitment = 1;
+  // Serialization of the G2Commitment to the blob length.
   bytes length_commitment = 2;
+  // Serialization of the G2Affine element representing the proof of the blob length.
   bytes length_proof = 3;
+  // The length of the blob in symbols (field elements).
+  // TODO: is this length always a power of 2? Are there any other characteristics that we should list? etc.
   uint32 length = 4;
 }
```
samlaf, line 31: Does anyone know for this one?

encoding/rs/utils.go
```diff
@@ -10,6 +10,15 @@ import (
...
 
+// ToFrArray deserializes a byte array into a list of bn254 field elements.,
+// where each 32-byte chunk needs to be a big-endian serialized bn254 field element.
+// The last chunk is allowed to not have 32-bytes, and will be padded with zeroes
+// on the right (so make sure that the last partial chunk represents a valid field element
+// when padded with zeroes on the right and interpreted as big-endian).
+//
+// TODO: we should probably just force the data to be a multiple of 32 bytes.
+// This would make the API and code simpler to read, and also allow the code
+// to be auto-vectorized by the compiler (it probably isn't right now given the if inside the for loop).
```
samlaf, line 21: Can we do this? Really think it would greatly help devex.

api/proto/common/common.proto
samlaf: Thanks, rebased against master which has the latest changes from cody's PR, where this TODO is solved.

samlaf:
@jianoaix @litt3 this is ready for review

api/proto/common/common.proto
```diff
@@ -11,19 +13,24 @@ message G1Commitment {
...
+  // Concatenation of the x and y coordinates of `common.G1Commitment`.
   bytes commitment = 1;
   // A commitment to the blob data with G2 SRS, used to work with length_proof
-  // such that the claimed length below is verifiable.
+	// such that the claimed length below is verifiable.
   bytes length_commitment = 2;
   // A proof that the degree of the polynomial used to generate the blob commitment is valid.
-  // It is computed such that the coefficient of the polynomial is committing with the G2 SRS
-  // at the end of the highest order.
+	// It is computed such that the coefficient of the polynomial is committing with the G2 SRS
```
jianoaix, line 28: Can we just say like it's `a commit to x^(SRSOrder-n) * P(x), where P(x) is polynomial of degree n representing the blob`?  This is not only clear, but concise.

api/proto/common/common.proto
```diff
@@ -11,19 +13,24 @@ message G1Commitment {
...
+	// at the end of the highest order.
   bytes length_proof = 3;
-  // The length specifies the degree of the polynomial used to generate the blob commitment. The length
-  // must equal to the degree + 1, and it must be a power of 2.
+  // The length of the blob in symbols (field elements), which must be a power of 2.
```
jianoaix, line 31: This is the one that I still have question about. I think it's just a upper bound of length/degree, rather than the length/degree of the poly.

api/proto/common/common.proto
```diff
@@ -2,7 +2,9 @@ syntax = "proto3";
 package common;
 option go_package = "github.com/Layr-Labs/eigenda/api/grpc/common";
 
-// A KZG commitment
+// G1Commitment represents the serialized coordinates of a G1 KZG commitment.
+// We use gnark-crypto so adopt its serialization, which is big-endian. See:
+// https://github.com/Consensys/gnark-crypto/blob/779e884dabb38b92e677f4891286637a3d2e5734/ecc/bn254/fp/element.go#L862
```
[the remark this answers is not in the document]
samlaf: not that I know

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -16,25 +16,41 @@ service Disperser {
...
 // Requests and Replies
 
 // A request to disperse a blob.
 message DisperseBlobRequest {
-  // The data to be dispersed.
+  // The encoded data to be dispersed to the EigenDA network.
   //
-  // The size of this byte array may be any size as long as it does not exceed the maximum length of 16MiB.
-  // (In the future, the 16MiB limit may be increased, but this is not guaranteed to happen.)
+  // Validation rules:
```
litt3, line 43: Size must also be >0

encoding/rs/utils.go
```diff
@@ -35,7 +44,9 @@ func ToFrArray(data []byte) ([]fr.Element, error) {
 	return eles, nil
 }
 
-// ToByteArray converts a list of Fr to a byte array
+// ToByteArray serializes a slice of fields elements to a slice of bytes.
+// The byte array is created by serializing each Fr element in big-endian format.
+// It is the reverse operation of ToFrArray.
```
litt3, line 49: Technically not quite the reverse, since `ToFrArray` adds padding, yet this function doesn't remove any padding

Doesn't necessarily need to be changed, though

api/proto/common/common.proto
```diff
@@ -2,7 +2,9 @@ syntax = "proto3";
...
 message G1Commitment {
   // The X coordinate of the KZG commitment. This is the raw byte representation of the field element.
   bytes x = 1;
```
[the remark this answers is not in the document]
samlaf: [2efb189](https://github.com/Layr-Labs/eigenda/pull/1104/commits/2efb189ae7a3e7597b8b606eec143c670a9c270d)

Specified number of bytes, but not what happens if x doesn't contain that number of bytes. Do we validate and enforce that x/y contain EXACLY 32 bytes? We could use https://github.com/bufbuild/protoc-gen-validate to enforce extra validation like this

api/proto/common/common.proto
samlaf: Not sure for v1, but for v2 I think its exactly the length now? cc @litt3

api/proto/common/common.proto
```diff
@@ -11,19 +13,24 @@ message G1Commitment {
 }
 
 // BlobCommitment represents commitment of a specific blob, containing its
-// KZG commitment, degree proof, the actual degree, and data length in number of symbols.
+// KZG commitment, degree proof, the actual degree, and data length in number of symbols (field elements).
+// It deserializes into https://github.com/Layr-Labs/eigenda/blob/ce89dab18d2f8f55004002e17dd3a18529277845/encoding/data.go#L27
+//
+// See https://github.com/Layr-Labs/eigenda/blob/master/docs/spec/attestation/encoding.md#validation-via-kzg
```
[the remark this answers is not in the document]
samlaf: [b94dbc7](https://github.com/Layr-Labs/eigenda/pull/1104/commits/b94dbc75744121bc2f1558a884f5a50929dbca5c)

api/proto/common/common.proto
```diff
@@ -11,19 +13,24 @@ message G1Commitment {
...
+// to understand how this commitment is used to validate the blob.
 message BlobCommitment {
-  // A commitment to the blob data.
+  // Concatenation of the x and y coordinates of `common.G1Commitment`.
   bytes commitment = 1;
   // A commitment to the blob data with G2 SRS, used to work with length_proof
-  // such that the claimed length below is verifiable.
+	// such that the claimed length below is verifiable.
```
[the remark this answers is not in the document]
samlaf: [0b6cd20](https://github.com/Layr-Labs/eigenda/pull/1104/commits/0b6cd20f8c85e9379af4d30a6993cfc3c2989476)

api/proto/common/common.proto
samlaf: [dd108ef](https://github.com/Layr-Labs/eigenda/pull/1104/commits/dd108efcb8080a5df28cb159d906535a9740e098)

encoding/rs/utils.go
samlaf: [d3e3b88](https://github.com/Layr-Labs/eigenda/pull/1104/commits/d3e3b882609bdba557227ab5e7ee45721f7ed442)

samlaf:
@jianoaix @litt3 sorry took super long to get back to this. fixed the conflicts and addressed all comments

api/proto/common/common.proto
litt3: I suppose you could call it an "upper bound", in the sense that you have the option choose to a power-of-2 length arbitrarily higher than the necessary length, without failing verification.

But I agree with @samlaf, we've been referring to this length as the source-of-truth, *actual length* of the blob. If the commitment creator chose a larger value than necessary, that means they essentially just appended a bunch of 0s to the blob polynomial, and therefore literally increased the length of the blob (even though they didn't add any additional non-0 information)

samlaf: But the appended 0s even though not transmitted from clients to disperser, will actually be encoded as chunks and transmitted to operators, which is why we charge for them right? So in this sense the blob_length is LITERALLY the blob_length, as seen from the network's perspective, right?

litt3: Yes, that matches my understanding.

encoding/rs/utils.go
```diff
@@ -14,7 +14,8 @@ import (
 // ToFrArray accept a byte array as an input, and converts it to an array of field elements
 //
 // TODO (litt3): it would be nice to rename this to "DeserializeFieldElements", as the counterpart to "SerializeFieldElements",
-//  but doing so would be a very large diff. I'm leaving this comment as a potential future cleanup.
+//
+//	but doing so would be a very large diff. I'm leaving this comment as a potential future cleanup.
```
litt3, line 18: I really hate this formatting that gets automatically imposed! Must be a golang thing, since goland does the same thing :( No need to change, manually fixing this reformatting over and over is more trouble than it's worth

litt3 (approved):

api/proto/common/common.proto
```diff
@@ -2,27 +2,35 @@ syntax = "proto3";
...
-  // A commitment to the blob data.
+  // Concatenation of the x and y coordinates of `common.G1Commitment`.
   bytes commitment = 1;
   // A commitment to the blob data with G2 SRS, used to work with length_proof
   // such that the claimed length below is verifiable.
   bytes length_commitment = 2;
   // A proof that the degree of the polynomial used to generate the blob commitment is valid.
-  // It is computed such that the coefficient of the polynomial is committing with the G2 SRS
-  // at the end of the highest order.
+  // It is computed as x^(SRSOrder-n) * P(x), where P(x) is polynomial of degree n representing the blob.
```
jianoaix, line 30: nit: not this x^(SRSOrder-n) * P(x) polynomial, but a kzg commitment to it

api/proto/common/common.proto
jianoaix: That makes sense to me.  I think the source of confusion might be the word "upper bound"; if we just say it's any power of 2 that's larger than raw/original blob, and this power of 2 is the *actual* blob viewed by network, that's clear to me.
Anyway, at least there is no use of "upper bound" in documentation here.

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -16,9 +16,21 @@ service Disperser {
   rpc GetBlobStatus(BlobStatusRequest) returns (BlobStatusReply) {}
 
   // GetBlobCommitment is a utility method that calculates commitment for a blob payload.
+  // It is provided to help clients who are trying to construct a DisperseBlobRequest.blob_header
+  // and don't have the ability to calculate the commitment themselves (expensive operation which requires SRS points).
+  //
+  // For an example usage, see how our disperser_client makes a call to this endpoint when it doesn't have a local prover:
```
jianoaix, line 22: This will be much better if we have API documentation or guide on how to use these APIs (with example), but that's outside the scope here

encoding/rs/utils.go
jianoaix: really auto imposed here? It's odd it adds a line here. Could "TODO(litt3)" help? (no space after TODO)

jianoaix (approved):

encoding/rs/utils.go
samlaf: this happens when you leave a space at the beginning of a line. I think ` but` has a space in front of it, so it did this. Ill fix it

api/proto/common/common.proto
samlaf: oh yeah thanks! [8920a23](https://github.com/Layr-Labs/eigenda/pull/1104/commits/8920a23a18c5a0ad9c95ce5a75d2d5238d93c504)

encoding/rs/utils.go
samlaf: [6feadaf](https://github.com/Layr-Labs/eigenda/pull/1104/commits/6feadaf01d5792ae1aac27568a7cefee3677767c)

PR #1107, opened 2025-01-14
Add utility function to call verifyBlobV2

litt3:

- Recent discussions have yielded consensus that the eigenDA cert which gets published to rollup inboxes should contain the `NonSignerStakesAndSignature`, rather than the `SignedBatch`
- This PR introduces a verification utility to call `VerifyBlobV2`, which accepts the `NonSignerStakesAndSignature`
- It's possible that the `VerifyBlobV2FromSignedBatch` verification utility will be deleted at some point, but I think it makes sense to keep for now, until the v2 client implementation has solidified.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/clients/v2/verification/blob_verifier.go
```diff
@@ -74,3 +75,40 @@ func (v *BlobVerifier) VerifyBlobV2FromSignedBatch(
 
 	return nil
 }
+
+// VerifyBlobV2 calls the VerifyBlobV2 view function on the EigenDABlobVerifier contract
+//
+// This method returns nil if the blob is successfully verified. Otherwise, it returns an error.
+//
+// It is the responsibility of the caller to configure a timeout on the ctx, if a timeout is required.
```
cody-littley, line 83: This sentence possibly not necessary, as this behavior is more or less the status-quo of a golang function that accepts a context as an argument but does not have a timeout parameter.

cody-littley (approved):

api/clients/v2/verification/blob_verifier.go
litt3: Removed

Issue #1100, opened 2025-01-13
Verify that the blob key returned in `DisperseBlob` matches what was sent


litt3:
- To avoid an additional trust assumption on the disperser, we should verify that the blob key returned from `DisperseBlob` matches what was originally sent https://github.com/Layr-Labs/eigenda/blob/6059c6a068298d11c41e50f5bcd208d0da44906a/api/clients/v2/disperser_client.go#L121
- With this check missing, the client wouldn't know if the disperser disperses something other than what the client created

PR #1109, opened 2025-01-14
Check blob key from disperser against actual key

litt3:

- Closes https://github.com/Layr-Labs/eigenda/issues/1100

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Integration tests


samlaf (approved):
Changes LGTM, but the DisperseBlob function is getting big and unweildy. Might be good to refactor at some point, create some functions to abstract some of the stuff out. Would need to think through best way to do this.

Your code here could clearly fit in a function though, for eg.

litt3:
> Changes LGTM, but the DisperseBlob function is getting big and unweildy. Might be good to refactor at some point, create some functions to abstract some of the stuff out. Would need to think through best way to do this.
> 
> Your code here could clearly fit in a function though, for eg.

@samlaf Good point. Split the blob key check into a separate function [617cbd5b](https://github.com/Layr-Labs/eigenda/pull/1109/commits/617cbd5b70c12a1e593bac6b63e430975fa16a92)

api/clients/v2/disperser_client.go
```diff
@@ -213,9 +213,46 @@ func (c *disperserClient) DisperseBlob(
...
+//
+// A successful verification guarantees that the disperser didn't make any modifications to the BlobHeader that it
+// received from this client.
+//
+// This function returns nil if the verification succeeds, and otherwise returns an error describing the failure
+func verifyReceivedBlobKey(
+	// the blob header which was constructed locally and sent to the disperser
+	blobHeader *corev2.BlobHeader,
+	// the reply received back from the disperser
+	disperserReply *disperser_rpc.DisperseBlobReply) error {
```
samlaf, line 234:
```suggestion
	disperserReply *disperser_rpc.DisperseBlobReply,
) error {
```

samlaf: we should probably have a formatter to make the code standardized everywhere. I personally prefer this formatting, but I'm fine with whatever, just suggesting.

samlaf (approved):
LGTM, thanks!

api/clients/v2/disperser_client.go
litt3: [78c58a1d](https://github.com/Layr-Labs/eigenda/pull/1109/commits/78c58a1dfa1a3798fef730620a4eeabb8bbe16bf)

I don't have a strong preference. Strongly agree, this is the sort of thing that ought to be managed by a standard formatter.

api/clients/v2/disperser_client.go
```diff
@@ -213,9 +213,47 @@ func (c *disperserClient) DisperseBlob(
...
+// A successful verification guarantees that the disperser didn't make any modifications to the BlobHeader that it
+// received from this client.
+//
+// This function returns nil if the verification succeeds, and otherwise returns an error describing the failure
+func verifyReceivedBlobKey(
+	// the blob header which was constructed locally and sent to the disperser
+	blobHeader *corev2.BlobHeader,
+	// the reply received back from the disperser
+	disperserReply *disperser_rpc.DisperseBlobReply,
+) error {
```
cody-littley, line 235: nit, would a simple unit test be possible? i.e. check both a valid and invalid header

cody-littley (approved):

api/clients/v2/disperser_client.go
litt3: Done

api/clients/v2/disperser_client.go
```diff
@@ -213,9 +213,47 @@ func (c *disperserClient) DisperseBlob(
...
+//
+// A successful verification guarantees that the disperser didn't make any modifications to the BlobHeader that it
+// received from this client.
+//
+// This function returns nil if the verification succeeds, and otherwise returns an error describing the failure
+func verifyReceivedBlobKey(
+	// the blob header which was constructed locally and sent to the disperser
+	blobHeader *corev2.BlobHeader,
+	// the reply received back from the disperser
+	disperserReply *disperser_rpc.DisperseBlobReply,
```
ian-shim, line 234: why don't we just take the received blob key instead of full reply?

api/clients/v2/disperser_client.go
```diff
@@ -213,9 +213,47 @@ func (c *disperserClient) DisperseBlob(
 		return nil, [32]byte{}, err
 	}
 
+	if verifyReceivedBlobKey(blobHeader, reply) != nil {
+		return nil, [32]byte{}, fmt.Errorf("verify received blob key: %w", err)
```
ian-shim, line 217: nit: "failed to verify that the reply contains locally computed blob key" or something like that..?

ian-shim (approved):

api/clients/v2/disperser_client.go
litt3: The idea was that we don't immediately have either blob key (computed or received) in the correct form: they need to be constructed and error checked. Moving the logic to the helper method was intended to keep `DisperseBlob` as clean as possible

I don't feel super strongly about this, I can remove the helper method if you would like

api/clients/v2/disperser_client.go
litt3: The helper method does contains more specific details about what specifically failed, e.g. `"blob key returned by disperser (%v) doesn't match blob which was dispersed (%v)"`

(Also, this message is attempting to follow [this standard](https://github.com/uber-go/guide/blob/master/style.md#error-wrapping) that @samlaf has been rallying support behind. It prescribes omitting the "failed to..." prefix)

Happy to still make an update here if you'd like, just pointing out the rationale of the current message

cody-littley (approved):

PR #1120, opened 2025-01-16
Create payload->blob type system

litt3:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/clients/codecs/blob.go
```diff
@@ -0,0 +1,79 @@
...
+)
+
+// Blob is data that is dispersed on eigenDA.
+//
+// A Blob is represented under the hood by a coeff polynomial
+type Blob struct {
+	coeffPolynomial *coeffPoly
+	// blobLength must be a power of 2, and should match the blobLength claimed in the BlobCommitment
+	// This is the blob length IN SYMBOLS, not in bytes
+	blobLength uint32
```
samlaf, line 14: consider renaming to `blobLengthSymbols`

samlaf: Although is symbol standard here? I feel like our whitepaper is the only place I've ever seen it. Should we just use FE for field element everywhere? BlobLengthFE?

samlaf, line 14: unclear how coeffPolynomial and blobLength are related. Does `len(coeffPolynomial) == blobLength`? Guessing we allow last zeros to be truncated. We should explain that

samlaf: also can't we just enforce that blobLength = nextPowerOf2(len(coeffPolynomial)) and then we don't need blobLength here?

api/clients/codecs/blob.go
```diff
@@ -0,0 +1,79 @@
...
+//
+// blobLength is the length of the blob IN SYMBOLS
+func BlobFromBytes(bytes []byte, blobLength uint32) (*Blob, error) {
+	poly, err := coeffPolyFromBytes(bytes)
+	if err != nil {
+		return nil, fmt.Errorf("polynomial from bytes: %w", err)
+	}
+
+	return BlobFromPolynomial(poly, blobLength)
+}
```
samlaf, line 27: see comment above, do we not want to enforce some relationship between len(bytes) and blobLength?

api/clients/codecs/blob.go
```diff
@@ -0,0 +1,79 @@
...
+	return &Blob{
+		coeffPolynomial: coeffPolynomial,
+		blobLength:      blobLength,
+	}, nil
+}
+
+// GetBytes gets the raw bytes of the Blob
+func (b *Blob) GetBytes() []byte {
+	return b.coeffPolynomial.getBytes()
+}
```
samlaf, line 42: have a preference for `Bytes()` without the `get` prefix

samlaf: or `toBytes()` to be consistent with `ToPayload`?

api/clients/codecs/blob.go
```diff
@@ -0,0 +1,79 @@
...
+
+// ToPayload converts the Blob into a Payload
+//
+// The payloadStartingForm indicates how payloads are constructed by the dispersing client. Based on the starting form
+// of the payload, we can determine what operations must be done to the blob in order to reconstruct the original payload
+func (b *Blob) ToPayload(payloadStartingForm PolynomialForm) (*Payload, error) {
```
samlaf, line 48: I think I prefer if get rid of the `starting` terminology. It introduces time which I personally find confusing. For me its really just "how does a rollup interpret its payloads as polynomial, ALWAYS", aka its more of an invariant or a way of thinking, than a "starting point".

Wdyt of just `payloadForm` or `payloadPolyForm`?

api/clients/codecs/blob.go
```diff
@@ -0,0 +1,79 @@
...
+	var encodedPayload *encodedPayload
+	var err error
+	switch payloadStartingForm {
+	case PolynomialFormCoeff:
+		// the payload started off in coefficient form, so no conversion needs to be done
```
samlaf, line 53: again to me its not about "starting". its more about the invariant that we want to convert a blob to a payload. This requires knowing 2 piece of info:
1. how does rollup interpret its payload
2. how does eigenda interpret its blobs (always coeff)

samlaf:
```suggestion
		// the payload is interpreted as coefficients of the polynomial, so no conversion needs to be done given that eigenda also interprets blobs as coefficients.
```

api/clients/codecs/blob.go
```diff
@@ -0,0 +1,79 @@
...
+		encodedPayload, err = b.coeffPolynomial.toEncodedPayload(b.blobLength)
+		if err != nil {
+			return nil, fmt.Errorf("coeff poly to encoded payload: %w", err)
+		}
+	case PolynomialFormEval:
+		// the payload started off in evaluation form, so we first need to convert the blob's coeff poly into an eval poly
```
samlaf, line 59:
```suggestion
		// the payload is interpreted as evaluations of a polynomial, so we need to convert the blob given that eigenda interprets blobs as coefficients.
```

encoding/fft/fft.go
```diff
@@ -88,3 +90,9 @@ func NewFFTSettings(maxScale uint8) *FFTSettings {
 		ReverseRootsOfUnity:  rootzReverse,
 	}
 }
+
+// FFTSettingsFromBlobLength accepts a blob length in symbols, and returns a new instance of FFT settings
+func FFTSettingsFromBlobLength(blobLength uint32) *FFTSettings {
```
samlaf, line 95: would really like blobLength to have a suffix, I feel it'll come back to byte (pun unintended at first lol) us otherwise

api/clients/codecs/blob_codec.go
```diff
@@ -4,22 +4,22 @@ import (
 	"fmt"
 )
 
-type BlobEncodingVersion byte
+type PayloadEncodingVersion byte
```
samlaf, line 7:
```suggestion
type PayloadEncodingVersion uint8
```

samlaf: byte and uint8 are aliases afaik. uint8 feels more appropriate here imo

api/clients/codecs/blob_codec.go
```diff
@@ -4,22 +4,22 @@ import (
...
 
 const (
-	// This minimal blob encoding contains a 32 byte header = [0x00, version byte, uint32 len of data, 0x00, 0x00,...]
+	// PayloadEncodingVersion0 entails a 32 byte header = [0x00, version byte, big-endian uint32 len of payload, 0x00, 0x00,...]
```
samlaf, line 10: can you add comment as to why we add the first 0x00 byte (so that it gets parsed as a bn254 FE). Took me a second to remember this

api/clients/codecs/default_blob_codec.go
```diff
@@ -22,7 +22,7 @@ func (v DefaultBlobCodec) EncodeBlob(rawData []byte) ([]byte, error) {
 	codecBlobHeader := make([]byte, 32)
 	// first byte is always 0 to ensure the codecBlobHeader is a valid bn254 element
 	// encode version byte
-	codecBlobHeader[1] = byte(DefaultBlobEncoding)
+	codecBlobHeader[1] = byte(PayloadEncodingVersion0)
```
samlaf, line 25: doesn't this function `EncodedBlob` need to be changed? it returns []byte, shouldnt it return a Blob in your type system? Or whatever else this is (prob just an encodedPayload?)

api/clients/codecs/blob_codec.go
```diff
@@ -4,22 +4,22 @@ import (
...
 	DecodeBlob(encodedData []byte) ([]byte, error)
 	EncodeBlob(rawData []byte) ([]byte, error)
 }
 
-func BlobEncodingVersionToCodec(version BlobEncodingVersion) (BlobCodec, error) {
+func BlobEncodingVersionToCodec(version PayloadEncodingVersion) (BlobCodec, error) {
 	switch version {
-	case DefaultBlobEncoding:
+	case PayloadEncodingVersion0:
 		return DefaultBlobCodec{}, nil
```
samlaf, line 23: this looks a bit weird now. Why is payload version 0 mapping to a defaultBlobCodec? Should we use BlobCodecVersion0 instead? Or do we want separate versioning for the payloadEncoding and BlobCodec? Or are these things the same?

api/clients/codecs/blob_test.go
```diff
@@ -0,0 +1,39 @@
...
+		originalData := testRandom.Bytes(testRandom.Intn(1024) + 1)
+		testBlobConversionForForm(t, originalData, PolynomialFormEval)
+		testBlobConversionForForm(t, originalData, PolynomialFormCoeff)
+	}
+}
+
+func testBlobConversionForForm(t *testing.T, payloadBytes []byte, form PolynomialForm) {
+	payload := NewPayload(payloadBytes)
+
+	blob, err := payload.ToBlob(form)
```
samlaf, line 26: The natural interpretation for this reads wrong :(
It reads "payload is getting to blob in this form" instead of "payload, interpreted in this form, gets converted to a blob".
Maybe there's a different way to structure the method such that the interpretation makes more sense?

samlaf: Like perhaps defining on Blob a method `.fromPayloadInForm()`? Then we could use like
`new(Blob).fromPayloadInForm(payload, form)`. It's def more ugly but at least it reads more easily I feel?

api/clients/codecs/blob_test.go
```diff
@@ -0,0 +1,39 @@
...
+	testRandom := random.NewTestRandom(t)
+
+	iterations := 1000
+
+	for i := 0; i < iterations; i++ {
+		originalData := testRandom.Bytes(testRandom.Intn(1024) + 1)
+		testBlobConversionForForm(t, originalData, PolynomialFormEval)
+		testBlobConversionForForm(t, originalData, PolynomialFormCoeff)
+	}
+}
```
samlaf, line 21: fuzzing would be better here as it does a smarter search than this random iid generation of test examples.

samlaf: also we should be testing the edge cases like empty bytes

api/clients/codecs/blob_codec.go
samlaf: Also should Payload related stuff be in its own codec file? (not sure, just asking)

api/clients/codecs/encoded_payload.go
```diff
@@ -0,0 +1,158 @@
...
+	}
+
+	return fieldElements, nil
+}
+
+// encodedPayloadFromElements accepts an array of field elements, and converts them into an encoded payload
+//
+// blobLength is the length of the blob IN SYMBOLS, as claimed by the blob commitment. This is needed to make sure
+// that the claimed length in the encoded payload header is valid relative to the total blob length
+func encodedPayloadFromElements(fieldElements []fr.Element, blobLength uint32) (*encodedPayload, error) {
```
samlaf, line 92: this comment makes me wonder whether we shouldn't instead define this method on the blob itself? Why do we need this method here? Seems like its at the wrong place given that it requires a blobLength?

samlaf (commented):
First pass. Still have more to review but tired.

api/clients/codecs/blob.go
litt3: I originally used `blobLength` to try to keep consistency with the struct in the commitment, but I agree, it's not ideal that I have to put in comments everywhere that it's in symbols

I don't have a preference between `BlobLengthFE` or `BlobLengthSymbols`. Usage of `Symbols` is pretty pervasive in our codebase, but also I don't think anyone would be confused by the `FE` terminology. Do you have a preference?

api/clients/codecs/coeff_poly.go
```diff
@@ -0,0 +1,63 @@
...
+	"github.com/Layr-Labs/eigenda/encoding/rs"
+	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
+)
+
+// coeffPoly is a polynomial in coefficient form.
+//
+// The underlying bytes represent 32 byte field elements, and each field element represents a coefficient
+type coeffPoly struct {
+	fieldElements []fr.Element
+}
```
cody-littley, line 16: This duplicates an existing type (currently in `encoding/rs/frame_coeffs.go` in `master, named something different if you aren't up to date with the latest `master`).

```
type FrameCoeffs []fr.Element
```

Also, there is a bunch of serialization logic in that file as well for going back and forth between serialized and deserialized form.

Is the plan to circle back and replace these old types in a follow up PR?

cody-littley: I see `[]fr.Element` used directly as a type in many places in this PR. Is there a reason why you don't use the fancy type you define here?

cody-littley: Is there a reason why you put this into a struct instead of doing `type coefPoly []fr.Element`?

api/clients/codecs/coeff_poly.go
```diff
@@ -0,0 +1,63 @@
...
+		return nil, fmt.Errorf("perform FFT: %w", err)
+	}
+
+	return evalPolyFromElements(fftedElements), nil
+}
+
+// GetBytes returns the bytes that underlie the polynomial
+func (p *coeffPoly) getBytes() []byte {
+	return rs.FieldElementsToBytes(p.fieldElements)
+}
```
cody-littley, line 55: My personal preference for methods like this where we convert to/from `[]byte` is to use a `serialize`/`deserialize` naming schema. Open to discussion if you don't agree.

api/clients/codecs/eval_poly.go
```diff
@@ -0,0 +1,46 @@
...
+	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
+)
+
+// evalPoly is a polynomial in evaluation form.
+//
+// The underlying bytes represent 32 byte field elements, and the field elements represent the polynomial evaluation
+// at roots of unity.
+type evalPoly struct {
+	fieldElements []fr.Element
+}
```
cody-littley, line 16: Particular reason why you use a struct instead of `type fieldElements []fr.Element`?

encoding/test_utils.go
```diff
@@ -0,0 +1,18 @@
...
+
+// GeneratePowersOfTwo creates a slice of integers, containing powers of 2 (starting with element == 1), with
+// powersToGenerate number of elements
+func GeneratePowersOfTwo[T constraints.Integer](powersToGenerate T) []T {
+	powers := make([]T, powersToGenerate)
+	for i := T(0); i < powersToGenerate; i++ {
+		powers[i] = T(math.Pow(2, float64(i)))
+	}
+
+	return powers
```
cody-littley, line 17: you can get more efficiently get powers of two by doing the following:
```suggestion
	powers := make([]T, powersToGenerate)
	for i := T(0); i < powersToGenerate; i++ {
		powers[i] = 1 << i
	}

	return powers
```

api/clients/codecs/blob.go
litt3: There is a corner case here:

- Imagine a user disperses a very small blob, only 64 bytes, and the last 40 bytes are trailing 0s
- When a different user fetches the blob from a relay, we've established that the relay _could_ remove trailing 0s
- If we were to say that `blobLength = nextPowerOf2(len(coeffPolynomial))`, then the user fetching and reconstructing this blob would determine that the blob length is 1 symbol, when it's actually 2

api/clients/codecs/blob.go
litt3: I personally tend to understand a nuanced difference between `To` and `Get`

-`to` (or in other contexts, `compute`) seems to indicate that the data is changing form, or being processed in some way
- `get` implies that the method is simply fetching something that already exists, byte for byte

So, I chose `ToPayload` because the method is transforming the data, to produce a payload. And it's `GetBytes`, because we are returning the literal bytes, as they already exist.

If this nuance isn't something that's generally understood, though, I'm ok changing the current names. What would your preferred naming be for these two methods? `Bytes()` and `ToPayload()`?

api/clients/codecs/blob.go
litt3: I'd be fine with either of these names, I agree `payloadStartingForm` isn't very good. To choose it, I first wrote the doc above, and then tried to come up with a parameter name to fit the description.

Let's say we changed the parameter name to `payloadForm`: how would you rephrase the method doc?

litt3: I think `payloadForm` is good. I will change it

encoding/fft/fft.go
litt3: Agreed, will change pending resolution of discussion [above](https://github.com/Layr-Labs/eigenda/pull/1120#discussion_r1954877635)

api/clients/codecs/blob_codec.go
litt3: [78010448](https://github.com/Layr-Labs/eigenda/pull/1120/commits/7801044891e39e9c9ed73b1404af20164c26a228)

api/clients/codecs/default_blob_codec.go
litt3: `default_blob_codec` will not be used in v2, all encoding is contained in the new structs

api/clients/codecs/blob_codec.go
litt3: I renamed the constant to what I felt made sense for the new code, and but the name doesn't work so well in the old `blob_codec` code. If I *don't* rename the constant, then the name doesn't work so well in the new code

> Or are these things the same?

Yes, they are essentially the same thing.

api/clients/codecs/blob_test.go
litt3: What do you think of this idea: when constructing a payload, declare in the constructor what form it is. The sequence would then read:

```
payload := NewPayload(payloadBytes, form)
blob, err := payload.ToBlob()
```

I personally *really* dislike like the pattern of creating a new empty object, just to call a method that creates _another_ instance of the object

api/clients/codecs/blob_test.go
litt3: I'm on board with fuzz testing, but when do we actually run our fuzz tests? Do they run in CI?

litt3: Took a stab at fuzz testing in [c70fad8d](https://github.com/Layr-Labs/eigenda/pull/1120/commits/c70fad8d2cb418706ae34cc9950887d9e1d826d7) LMK what you think

api/clients/codecs/blob.go
litt3: [e22238b1](https://github.com/Layr-Labs/eigenda/pull/1120/commits/e22238b117a6e22727a3f6a1d021fba6a38b9607)

api/clients/codecs/blob.go
litt3: see my [response](https://github.com/Layr-Labs/eigenda/pull/1120#discussion_r1954938990) above

api/clients/codecs/blob.go
litt3: Wording fixed in[e22238b1](https://github.com/Layr-Labs/eigenda/pull/1120/commits/e22238b117a6e22727a3f6a1d021fba6a38b9607)

api/clients/codecs/blob.go
litt3: Wording reworked in [e22238b1](https://github.com/Layr-Labs/eigenda/pull/1120/commits/e22238b117a6e22727a3f6a1d021fba6a38b9607)

api/clients/codecs/encoded_payload.go
litt3: In [617faa7b63](https://github.com/Layr-Labs/eigenda/pull/1120/commits/617faa7b638785ebbcfc71d02ebdc5e6d0795cf6#diff-1556913a0af60ea5bf48b4b5a78d03e1be5575d485dcff7658fcadfde31a211f), I changed the function signature to the following, to remove blob related logic from `encodedPayload`:

```
encodedPayloadFromElements(fieldElements []fr.Element, maxPayloadLength uint32)
```

api/clients/codecs/coeff_poly.go
litt3: When I made the suggested change, `type coefPoly []fr.Element`, everything started simplifying down, culminating in the removal of the type entirely.

api/clients/codecs/blob.go
litt3: @samlaf How do you feel about using the `serialize` / `deserialize` nomenclature? https://github.com/Layr-Labs/eigenda/pull/1120#discussion_r1954921604

api/clients/codecs/eval_poly.go
litt3: Type removed entirely

encoding/test_utils.go
litt3: Fixed

api/clients/codecs/blob.go
samlaf: I would personally prefer we get rid of symbols everywhere. But for consistency and minimal diff let’s just use symbols suffix here for now? We can chat during standup what people think of FE/symbols.

api/clients/codecs/blob.go
samlaf: Oh yes that’s even better!

api/clients/codecs/blob.go
litt3: [c028c00d](https://github.com/Layr-Labs/eigenda/pull/1120/commits/c028c00d6bd07cf6cf1e05a625285d019b323de8)

api/clients/codecs/blob.go
litt3: [f1a4e85f](https://github.com/Layr-Labs/eigenda/pull/1120/commits/f1a4e85f83ef84935aaba888cbb28dfb9e743612)

api/clients/codecs/coeff_poly.go
litt3: Done

api/clients/codecs/blob.go
samlaf: Ah. Didn't we discuss last week this very case and concluded that we would just force the relay to keep those 0 bytes around? I'm always in favor of simplifying logic, especiallywhen the "corner cases" are things that will tbh never happen. Let me finish reviewing the PR to get the full picture of this implementation though.

api/clients/codecs/blob.go
samlaf: Just realized now that golang has standard interfaces for serialization: MarshalJSON() and [MarshalBinary()](https://pkg.go.dev/encoding#BinaryMarshaler). Might be worth it to add a comment as to why we define our own function that isn't marshalBinary (perhaps we should just implement MarshalBinary() instead? so that people could encode our structs to gob if they need to for whatever reason).

Also the fact that we need that extra length param to unmarshal is another consequence of this weird "edge case that'll never happen". If we changed that requirement then this code would also be much simpler and more standard go. Again, we can discuss (perhaps with your brother).. I might be missing something.

api/clients/codecs/default_blob_codec.go
samlaf: I see... shouldnt we separate v1 and v2 codecs then? Why are they all in the same package? Or am I missing something?

api/clients/codecs/blob_codec.go
samlaf: I see! Think I'm missing the venn diagram of which code is used for v1, which is used for v2, and which is used for both. is it not possible to completely separate v1 and v2 stuff?

api/clients/codecs/blob_test.go
samlaf: > just to call a method that creates another instance of the object
In this case you would call a method which changes the object in place though.

`payloadForm` name does remove the ambiguity though, thanks!

cody-littley (approved):

api/clients/codecs/coeff_poly.go
samlaf: isn't best to keep the type around to have type safety when applying FFT/IFFT operations (can't by mistake apply FFT twice for eg)

api/clients/codecs/blob.go
litt3: I don't think there is any good way that we can force relays to keep these bytes around: it's a consequence of how our commitment scheme works. Let's assume we *were* to implement a policy that "forces" relays to NOT truncate trailing 0 bytes. Here a story which show the problem with this policy:

1. A dispersing client crafts a payload, which yields a blob with some trailing 0s
2. This dispersing client is malicious, and decides to truncate the trailing 0s before sending the blob to the disperser, just to cause trouble
3. The disperser receives the blob, everything goes smoothly, and the blob gets attested to by DA nodes
4. When a retrieving client later fetches the blob from a relay, the commitments verify correctly
5. When the retrieving client tries to decode the blob into a payload, though, it appears that this relay has maliciously truncated some zeros
6. The retrieving client therefore decides to try with another relay. But whaddayaknow, this relay *also* appears to have truncated some zeros
7. Etc., etc.. A malicious dispersing client is able to cause a situation where all commitments verify correctly, but it's not possible to retrieve a decodable blob from anywhere. Retrieving clients will spend lots of time and resources attempting to fetch the blob from every imaginable source, until finally timing out.

Aside from this attack vector, a policy which attempts to "force" relays not to truncate zeros also complicates retrieval logic. As currently implemented, if the bytes sent by a relay are successfully verified against the commitments, there is never a need to try fetching from a different relay: either the received bytes are decodable or not decodable. If we add a way that a relay might respond "maliciously", which *doesn't* result in failed commitment verification, that would mean a failed payload decode might require another fetch attempt, depending on the mode of failure.

api/clients/codecs/blob.go
litt3: I don't think it's worth implementing an interface just because it exists, without a concrete use case. I suppose it would be fine to add a comment saying why we don't implement the interface, but it doesn't seem necessary: throughout our whole codebase, there are no implementations of `BinaryMarshaler`: I think the implicit explanation for why it's not implemented is that we don't need it.

cody-littley: I don't feel strongly one way or another about the golang interfaces, but I'm generally in favor of avoiding code complexity that doesn't have a justification on the short-to-medium term roadmap.

Perhaps also relevant, the string `MarshalBinary` doesn't appear anywhere in the eigenDA repo right now, so this isn't a pattern we are currently using.

api/clients/codecs/default_blob_codec.go
litt3: Yeah, very good point, done in (3db92747)[https://github.com/Layr-Labs/eigenda/pull/1120/commits/3db927475247142dbcd7e715a04cc5d199ea1fde)

Unfortunately, I made the mistake of sharing code between v1 and v2 before I knew better. So, `polynomial_form.go` still lives in the `codecs` package. Everything else v2 related is now in the package `codecs/v2`, though

api/clients/codecs/blob_codec.go
litt3: [Mostly](https://github.com/Layr-Labs/eigenda/pull/1120#discussion_r1961810221) separated in [3db92747](https://github.com/Layr-Labs/eigenda/pull/1120/commits/3db927475247142dbcd7e715a04cc5d199ea1fde). Hopefully this makes things a bit clearer

api/clients/codecs/blob.go
samlaf: SGTM. Let's resolve this bike shedding haha. I just think that we should abide by golang conventions because otherwise it makes the code harder to read for people familiar with those conventions. The convention here being the use of `Marshal` instead of `Serialize`.

api/clients/codecs/default_blob_codec.go
samlaf: Nice, this looks cleaner! One last nit... I really don't like when /v2 is the last suffix, because it makes the package name "v2". So if we have that across the repo and import a bunch of "v2"s, they all have to be renamed. The convention is to put /v2 before the actual package name.

haven't verified that it makes sense here, but would .../v2/codecs/ work?

api/clients/codecs/blob_codec.go
samlaf: resolving in favor of that other thread. Thanks for the change!

samlaf (approved):
LGTM except for the codecs/v2 package name. If swapping those around makes sense, please do that before merging. Otherwise rest looks awesome, really happy with this new code!

api/clients/codecs/default_blob_codec.go
litt3: You're right, this is a much better configuration, obvious in hindsight [b8d43f59](https://github.com/Layr-Labs/eigenda/pull/1120/commits/b8d43f59ba3c666d9c8946770c87b11cb9c0c04e)

samlaf (approved):

PR #1125, opened 2025-01-17
Rename "node" to "validator" for v2 protobufs.

cody-littley:

Per previous discussions, we want to rebrand "DA Nodes" as "Validators". This renames the v2 protobufs. Refactoring the golang code will happen as future PRs.


jianoaix:
Wait, I thought it's "DA Operator" -> "Validator" rebranding, not DA Node.

The “Operator" or "Validator" is a role/persona in running the system, whereas "DA Node" is a piece of software ran by the role. So I am not sure naming "DA Node" software as "Validator" (or "Operator") makes sense.

jianoaix (approved):

PR #1126, opened 2025-01-17
refactor: protobuf payment header in v2 and salt

hopeyen:

protobuf structuring 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/clients/v2/accountant.go
```diff
@@ -109,7 +109,7 @@ func (a *Accountant) BlobPaymentInfo(ctx context.Context, numSymbols uint32, quo
 }
 
 // AccountBlob accountant provides and records payment information
-func (a *Accountant) AccountBlob(ctx context.Context, numSymbols uint32, quorums []uint8, salt uint32) (*core.PaymentMetadata, error) {
+func (a *Accountant) AccountBlob(ctx context.Context, numSymbols uint32, quorums []uint8) (*core.PaymentMetadata, error) {
```
samlaf, line 112: why are you removing the salt argument?

api/proto/common/v2/common_v2.proto
```diff
@@ -20,9 +20,11 @@ message BlobHeader {
   // commitment is the KZG commitment to the blob
   common.BlobCommitment commitment = 3;
   // payment_header contains payment information for the blob
-  common.PaymentHeader payment_header = 4;
+  PaymentHeader payment_header = 4;
   // signature over keccak hash of the blob_header that can be verified by blob_header.account_id
   bytes signature = 5;
+  // salt is used to ensure that the dispersal request is intentionally unique.
```
samlaf, line 26: add fact that this is (currently) only useful for reserved payments.

api/proto/common/v2/common_v2.proto
```diff
@@ -20,9 +20,11 @@ message BlobHeader {
...
+  uint32 salt = 6;
```
samlaf, line 27: I don't really like that we're adding stuff after the signature, because the signature signs all of the blob_header, except for itself.

In fact, @litt3 has raised the proposal that we move the signature out of the blob_header, and add is as a separate argument to dispersal requests.

samlaf: Thinking more about this though, in a future where people do self-dispersals and payments is verified by nodes, the signature should actually be part of the blob_header shouldn it?

@ian-shim does the blob_key = blob_header_hash also hash the signature?

core/data.go
```diff
@@ -492,11 +492,8 @@ type PaymentMetadata struct {
 
 	// ReservationPeriod represents the range of time at which the dispersal is made
 	ReservationPeriod uint32 `json:"reservation_period"`
-	// TODO: we are thinking the contract can use uint128 for cumulative payment,
-	// but the definition on v2 uses uint64. Double check with team.
+	// CumulativePayment represents the total amount of payment made by the user up to this point
```
samlaf, line 495: What is the payment unit? wei? Add this here as info

core/v2/serialization.go
```diff
@@ -129,6 +136,7 @@ func (b *BlobHeader) BlobKey() (BlobKey, error) {
 			},
 			DataLength: uint32(b.BlobCommitments.Length),
 		},
+		b.Salt,
```
samlaf, line 139: is it normal that blob_header_hash doesn't contain the payment_header nor the signature?

This also means that the salt was not even part of this hash until now? Seems like that was a bug/mistake?

core/v2/types.go
```diff
@@ -72,6 +72,9 @@ type BlobHeader struct {
 
 	// Signature is the signature of the blob header by the account ID
 	Signature []byte
+
+	// Salt is used to make blob intentionally unique when everything else is the same
+	Salt uint32
```
samlaf, line 77: move salt above signature. also make signature's comment more precise. What EXACLY is it signing over?

api/clients/v2/accountant.go
hopeyen: Since salt is not in payment header anymore we don't need to give it to the accountant

api/proto/common/v2/common_v2.proto
hopeyen: `blob_key` is packed with `blobHeaderHash` (hashed without signature) and `paymentMetadataHash`, then signer signs over the blob_key to make the signature, which is then written into blobHeader

I'm okay to move salt up but we can discuss if it makes more sense for blob_key to be changed to `(blobHeaderHash, paymentMetadataHash, Salt)`

core/v2/serialization.go
hopeyen: I think we agreed that `blob_header_hash` doesn't contain payment header nor the signature

Salt was part of the payment metadata hash, so it didn't need to be here

api/hashing/node_hashing.go
```diff
@@ -52,9 +54,8 @@ func hashBlobCommitment(hasher hash.Hash, commitment *commonv1.BlobCommitment) {
 	hashUint32(hasher, commitment.Length)
 }
 
-func hashPaymentHeader(hasher hash.Hash, header *commonv1.PaymentHeader) {
+func hashPaymentHeader(hasher hash.Hash, header *common.PaymentHeader) {
```
ian-shim, line 57: @cody-littley can you check if this change is good?

contracts/src/interfaces/IEigenDAStructs.sol
```diff
@@ -82,6 +82,7 @@ struct BlobHeaderV2 {
     bytes quorumNumbers;
     BlobCommitment commitment;
     bytes32 paymentHeaderHash;
+    uint32 salt;
```
ian-shim, line 85: nit: can we put salt before `paymentHeaderHash`?

ian-shim (approved):

contracts/src/interfaces/IEigenDAStructs.sol
hopeyen: Any particular reason? 
I think I can go ahead and switch the order, but checking with @0x0aa0 would you be concerned with making any other changes?

0x0aa0: no problems on my end with flipping

hopeyen: flipped!

core/v2/serialization.go
hopeyen: gonna go ahead and merge since I think we have agreed on the structures. please let me know if you have more concerns and I will create subsequent PRs addressing them

api/clients/v2/accountant.go
samlaf: Isn't that salt still needed for some things? Trying to understand where this salt is passed/used instead in the new code. Unless the accountant was just never using that salt anyways?

hopeyen: The salt was only there so the Accountant can include it in the PaymentMetadata that it later returns. It doesn't do anything special, and probably didn't need to take it in the first place if we added the salt to metadata after Accountant generates it

PR #1128, opened 2025-01-17
add api key bls remote signer

shrimalmadhur:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


node/config.go
```diff
@@ -186,6 +187,10 @@ func NewConfig(ctx *cli.Context) (*Config, error) {
 			return nil, fmt.Errorf("BLS key file and password is required if BLS remote signer is disabled")
 		}
 
+		if blsRemoteSignerEnabled && blsSignerAPIKey == "" {
+			return nil, fmt.Errorf("BLS signer API key is required if BLS remote signer is enabled")
+		}
```
pschork, line 192: Is this requirement predicated on a new version of Cerberus that also requires API key? I've got remote BLS running on my testnet now with Cerberus. When I upgrade my node, do I need to simultaneously upgrade Cerberus? If I upgrade Cerberus first, will my node fail to remote sign until upgraded with this change?

shrimalmadhur: yes basically you will have to clean import on the new version of cerberus - since it sets up a DB and then persists some metadata for the keys - if you are already running cerberus. You will have to upgrade both together
not ideal but given only few folks (2-3 folks outside us) are testing right now on testnet it's an okay process. wdyt?

pschork: Yea, as long as current testnet users are in the loop, seems ok. 

If possible might make sense to disable remote singing on node as first step - reverting to local BLS, to allow for Cerberus upgrade validation outside production path. Or maybe recommend users spin up new Cerberus alongside old Cerberus on new port so that they can import and validate new Cerberus is ready before pulling the switch. Then update `.env` and bump node release.

pschork (approved):

node/config.go
shrimalmadhur: that's a good point. I will write an upgrade guide in release notes with these options and share. good call on that.

PR #1129, opened 2025-01-17
Add support for range query for batches/attestations feed on dynamo

jianoaix:
This is to support both 1) batch feed; and 2) validator signing rate computing.

A batch/attestation feed applies to all batches that were dispersed to attestation, ordered by <AttestedAt> timestamp.

This PR is similar to https://github.com/Layr-Labs/eigenda/pull/1102 (for blob feed), but for batch/attestation.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/common/v2/blobstore/dynamo_metadata_store.go
+++ b/disperser/common/v2/blobstore/dynamo_metadata_store.go
@@ -381,13 +381,11 @@ func (s *BlobMetadataStore) GetBlobMetadataByRequestedAt(
 	return result, lastProcessedCursor, nil
 }
 
-// queryBucketAttestation returns attestations within a single bucket of time range
-// [start, end]. Results are ordered by AttestedAt in ascending order.
+// queryBucketAttestation returns attestations within a single bucket of time range [start, end]. Results are ordered by AttestedAt in
+// ascending order.
 //
-// The function handles DynamoDB's 1MB response size limitation by performing multiple queries
-// if necessary.
-// If there are more than numToReturn attestations in the bucket, returns numToReturn
-// attestations; otherwise returns all attestations in bucket.
+// The function handles DynamoDB's 1MB response size limitation by performing multiple queries  if necessary.
+// If there are more than numToReturn attestations in the bucket, returns numToReturn attestations; otherwise returns all attestations in bucket.
 func (s *BlobMetadataStore) queryBucketAttestation(ctx context.Context, bucket, start, end uint64, numToReturn int) ([]*corev2.Attestation, error) {
 	attestations := make([]*corev2.Attestation, 0)
 	var lastEvaledKey map[string]types.AttributeValue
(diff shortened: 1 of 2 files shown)


disperser/common/v2/blobstore/dynamo_metadata_store.go
```diff
@@ -400,6 +387,72 @@ func (s *BlobMetadataStore) GetBlobMetadataByRequestedAt(
 	return result, lastProcessedCursor, nil
 }
 
+// queryBucketAttestation queries a single bucket for attestations within the given time range.
+func (s *BlobMetadataStore) queryBucketAttestation(ctx context.Context, bucket, startTs, endTs uint64) ([]*corev2.Attestation, error) {
```
dmanc, line 391: what does Ts mean?

disperser/common/v2/blobstore/dynamo_metadata_store.go
```diff
@@ -400,6 +387,72 @@ func (s *BlobMetadataStore) GetBlobMetadataByRequestedAt(
...
+
+	startBucket, endBucket := getAttestedAtBucketIDRange(start, end)
+
+	result := make([]*corev2.Attestation, 0)
+	for bucket := startBucket; bucket <= endBucket; bucket++ {
+		if limit > 0 && len(result) >= limit {
+			break
+		}
+
+		bucketAttestation, err := s.queryBucketAttestation(ctx, bucket, start+1, end)
```
dmanc, line 440: Do we need to do pagination? not sure how big the attestation records are / how many of them we will query at most.

disperser/common/v2/blobstore/dynamo_metadata_store.go
jianoaix: Timestamp, it's quite common acronym

disperser/common/v2/blobstore/dynamo_metadata_store.go
jianoaix: I think it'd be simpler to cap the limit (which we have yet to check how large that number should be), than pagination nested in pagination

jianoaix: Dynamo can return only 1MB response at max (https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html)... This makes actually need pagination nested here

jianoaix: Ok, did math: assuming 200 nonsigners, the message size will 6588 bytes, which is large; and it can return only 151 results in one query. The nonsigners are certainly a big source of cost if there are many.

There are optimizations that can be done at upper level. But at blobstore level, this should be handled as well.

```
BatchHeader: 16 bytes
AttestedAt: 8 bytes
NonSignerPubKeys: 200 * 32 = 6400 bytes (200 G1Points)
APKG2: 64 bytes (one G2Point)
QuorumAPKs: 2 * 32 = 64 bytes (values only, plus map overhead)
Sigma: 32 bytes (one signature)
QuorumNumbers: 2 * 1 = 2 bytes (assuming QuorumID is uint8)
QuorumResults: 2 * 1 = 2 bytes (values only, plus map overhead)
```

dmanc (approved):

PR #1130, opened 2025-01-18
Wrap errors in relay

dmanc:

Should fix unknown errors.
image

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/relay/server.go
+++ b/relay/server.go
@@ -5,6 +5,7 @@ import (
 	"errors"
 	"fmt"
 	"net"
+	"strings"
 	"time"
 
 	"github.com/Layr-Labs/eigenda/api"
@@ -328,10 +329,13 @@ func (s *Server) GetChunks(ctx context.Context, request *pb.GetChunksRequest) (*
 
 	requiredBandwidth, err := computeChunkRequestRequiredBandwidth(request, mMap)
 	if err != nil {
-		return nil, api.NewErrorInvalidArg(fmt.Sprintf("error computing required bandwidth: %v", err))
+		return nil, api.NewErrorInternal(fmt.Sprintf("error computing required bandwidth: %v", err))
 	}
 	err = s.chunkRateLimiter.RequestGetChunkBandwidth(time.Now(), clientID, requiredBandwidth)
 	if err != nil {
+		if strings.Contains(err.Error(), "internal error") {
+			return nil, api.NewErrorInternal(err.Error())
+		}
 		return nil, api.NewErrorResourceExhausted(fmt.Sprintf("bandwidth limit exceeded: %v", err))
 	}
 	s.metrics.ReportChunkDataSize(requiredBandwidth)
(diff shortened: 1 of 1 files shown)


jianoaix (approved):
Yea, we should certainly do this!

I think the new Disperser API server and Node APIs will need a audit to make sure errors are canonicalized

relay/server.go
```diff
@@ -276,25 +277,25 @@ func (s *Server) GetChunks(ctx context.Context, request *pb.GetChunksRequest) (*
...
+		return nil, api.NewErrorInvalidArg(fmt.Sprintf(
+			"too many chunk requests provided, max is %d", s.config.MaxKeysPerGetChunksRequest))
 	}
 	s.metrics.ReportChunkKeyCount(len(request.ChunkRequests))
 
 	if s.authenticator != nil {
 		client, ok := peer.FromContext(ctx)
 		if !ok {
-			return nil, errors.New("could not get peer information")
+			return nil, api.NewErrorInternal("could not get peer information")
```
jianoaix, line 291: Is this invalid arg or internal?

relay/server.go
```diff
@@ -306,43 +307,43 @@ func (s *Server) GetChunks(ctx context.Context, request *pb.GetChunksRequest) (*
...
+			"error fetching metadata for blob, check if blob exists and is assigned to this relay: %v", err))
 	}
 
 	finishedFetchingMetadata := time.Now()
 	s.metrics.ReportChunkMetadataLatency(finishedFetchingMetadata.Sub(finishedAuthenticating))
 
 	requiredBandwidth, err := computeChunkRequestRequiredBandwidth(request, mMap)
 	if err != nil {
-		return nil, fmt.Errorf("error computing required bandwidth: %w", err)
+		return nil, api.NewErrorInvalidArg(fmt.Sprintf("error computing required bandwidth: %v", err))
```
jianoaix, line 331: Should this be internal?

relay/server.go
```diff
@@ -306,43 +307,43 @@ func (s *Server) GetChunks(ctx context.Context, request *pb.GetChunksRequest) (*
...
 	}
 	err = s.chunkRateLimiter.RequestGetChunkBandwidth(time.Now(), clientID, requiredBandwidth)
 	if err != nil {
```
jianoaix, line 334: Is it possible that rate limiter failed (not that there is no budget left)?

relay/server.go
```diff
@@ -306,43 +307,43 @@ func (s *Server) GetChunks(ctx context.Context, request *pb.GetChunksRequest) (*
...
 	frames, err := s.chunkProvider.GetFrames(ctx, mMap)
 	if err != nil {
-		return nil, fmt.Errorf("error fetching frames: %w", err)
+		return nil, api.NewErrorInternal(fmt.Sprintf("error fetching frames: %v", err))
 	}
 
 	bytesToSend, err := gatherChunkDataToSend(frames, request)
 	if err != nil {
-		return nil, fmt.Errorf("error gathering chunk data: %w", err)
+		return nil, api.NewErrorInvalidArg(fmt.Sprintf("error gathering chunk data: %v", err))
```
jianoaix, line 346: Again seems like an internal error

relay/server.go
dmanc: There is one case where it can be internal, will add that.
```
	limiter, ok := l.perClientBandwidthLimiter[requesterID]
	if !ok {
		return fmt.Errorf("internal error, unable to find bandwidth limiter for client ID %s", requesterID)
	}
```

relay/server.go
dmanc: Seems internal to me but not sure.

cody-littley (approved):

PR #1132, opened 2025-01-21
Update node socket registration format

ian-shim:
Decoupling v1 and v2 dispersal ports and integrating v2 port to the existing socket registration format (i.e. `<host>:<dispersalPort>;<retrievalPort>;<v2DispersalPort>`)

TODO: reachability check for v2 dispersal port
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/mock/state.go
+++ b/core/mock/state.go
@@ -26,11 +26,12 @@ var _ core.IndexedChainState = (*ChainDataMock)(nil)
 
 type PrivateOperatorInfo struct {
 	*core.IndexedOperatorInfo
-	KeyPair       *core.KeyPair
-	Signer        blssigner.Signer
-	Host          string
-	DispersalPort string
-	RetrievalPort string
+	KeyPair         *core.KeyPair
+	Signer          blssigner.Signer
+	Host            string
+	DispersalPort   string
+	RetrievalPort   string
+	V2DispersalPort string
 }
 
 type PrivateOperatorState struct {
@@ -138,7 +139,8 @@ func (d *ChainDataMock) GetTotalOperatorStateWithQuorums(ctx context.Context, bl
 		host := "0.0.0.0"
 		dispersalPort := fmt.Sprintf("3%03v", 2*i)
 		retrievalPort := fmt.Sprintf("3%03v", 2*i+1)
-		socket := core.MakeOperatorSocket(host, dispersalPort, retrievalPort)
+		v2DispersalPort := fmt.Sprintf("3%03v", 2*i+2)
+		socket := core.MakeOperatorSocket(host, dispersalPort, retrievalPort, v2DispersalPort)
 
 		indexed := &core.IndexedOperatorInfo{
 			Socket:   string(socket),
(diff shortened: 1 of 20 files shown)


node/config.go
```diff
@@ -228,12 +230,40 @@ func NewConfig(ctx *cli.Context) (*Config, error) {
...
+	_, err = strconv.Atoi(retrievalPort)
+	if err != nil {
+		return nil, fmt.Errorf("invalid retrieval port: %s", retrievalPort)
+	}
+
+	v2Enabled := ctx.GlobalBool(flags.EnableV2Flag.Name)
+
+	v2DispersalPort := ctx.GlobalString(flags.V2DispersalPortFlag.Name)
+	if v2DispersalPort == "" {
+		if v2Enabled {
```
ian-shim, line 250: making v2 dispersal port required if v2 is enabled

node/flags/flags.go
```diff
@@ -218,7 +224,7 @@ var (
 		Required: false,
 		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "ENABLE_GNARK_BUNDLE_ENCODING"),
 	}
-	EnableV2Flag = cli.BoolFlag{
+	EnableV2Flag = cli.BoolTFlag{
```
ian-shim, line 227: defaulting this to true

node/grpc/run.go
```diff
@@ -51,6 +51,38 @@ func RunServers(serverV1 *Server, serverV2 *ServerV2, config *node.Config, logge
...
+			return
+		}
+		for {
+			addr := fmt.Sprintf("%s:%s", localhost, config.V2DispersalPort)
+			listener, err := net.Listen("tcp", addr)
+			if err != nil {
+				logger.Fatalf("Could not start tcp listener: %v", err)
+			}
+
+			opt := grpc.MaxRecvMsgSize(config.GRPCMsgSizeLimitV2)
```
ian-shim, line 67: reducing msg size for v2 dispersal requests

core/state.go
```diff
@@ -12,35 +12,55 @@ import (
...
+		dispersalPort = s[1]
+
 		return
 	}
-	retrievalPort = s[1]
 
-	s = strings.Split(s[0], ":")
-	if len(s) != 2 {
-		err = fmt.Errorf("invalid socket address format: %s", socket)
+	if len(s) == 3 {
```
pschork, line 48: What if len > 3 or < 2?

ian-shim: Then the error throws in L63

pschork (approved):

core/mock/state.go
```diff
@@ -26,11 +26,12 @@ var _ core.IndexedChainState = (*ChainDataMock)(nil)
...
-	Signer        blssigner.Signer
-	Host          string
-	DispersalPort string
-	RetrievalPort string
+	KeyPair         *core.KeyPair
+	Signer          blssigner.Signer
+	Host            string
+	DispersalPort   string
+	RetrievalPort   string
+	V2DispersalPort string
```
cody-littley, line 34: Would it be possible to instead use `int` types for ports? We just need to change the flag type from `string` to `int` to support this. I've made this change in a few places and everything compiles and runs just fine.

ian-shim: Yes, I think that should have been the way to go, but I'm hesitant to change the flags from `string` to `int` because I'm not sure how every operator is configuring the values and if changing the type will be backward compatible

PR #1134, opened 2025-01-21
update middleware commit 

0x0aa0:
brings middleware to latest dev commit


ian-shim (approved):
do we need to redeploy preprod contracts?

PR #1136, opened 2025-01-22
Rename v2 structs

ian-shim:
Changes:
- `Data` -> `Blob` from disperser gRPC service 
- `BlobVerificationInfo` -> `BlobInclusionInfo`
- Dispersal request signature is moved from `BlobHeader` to `BlobCertificate` 
- `Relays` -> `RelayKeys` (in BlobCertificate)
- `BlobCertificate` hashing method (both offchain and onchain) includes `Signature` 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


contracts/src/interfaces/IEigenDAStructs.sol
0x0aa0, line 1: Do we also need to rename BlobVerificationProofV2 -> BlobInclusionInfo here?

ian-shim: I thought `BlobVerificationProofV2` in this context was fine. But curious if @samlaf has any takes

api/proto/common/v2/common_v2.proto
```diff
@@ -36,12 +34,14 @@ message BlobHeader {
...
-  // relays is the list of relays that are in custody of the blob.
+  // signature over keccak hash of the blob_header that can be verified by blob_header.payment_header.account_id
+  bytes signature = 2;
+  // relay_keys is the list of relay keys that are in custody of the blob.
   // The relays custodying the data are chosen by the Disperser to which the DisperseBlob request was submitted.
   // It needs to contain at least 1 relay number.
   // To retrieve a blob from the relay, one can find that relay's URL in the EigenDARelayRegistry contract:
   // https://github.com/Layr-Labs/eigenda/blob/master/contracts/src/core/EigenDARelayRegistry.sol
-  repeated uint32 relays = 2;
+  repeated uint32 relay_keys = 3;
```
jianoaix, line 44: it's fine to settle on "relay key" instead of "relay ID", but can you get all the "relay ID"s renamed to make it consistent, eg: https://github.com/Layr-Labs/eigenda/blob/9bc28688fbf5dbc11cdc0096549a0d34f75ab19e/relay/server.go#L72

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -35,14 +35,16 @@ message DisperseBlobRequest {
   // significant bits. The integer must stay in the valid range to be interpreted as a field element on the bn254 curve.
   // The valid range is 0 <= x < 21888242871839275222246405745257275088548364400416034343698204186575808495617.
   // If any one of the 32 bytes elements is outside the range, the whole request is deemed as invalid, and rejected.
-  bytes data = 1;
+  bytes blob = 1;
```
jianoaix, line 38: This is `encodedPayload` not `blob` per the latest naming definition.

api/proto/common/v2/common_v2.proto
```diff
@@ -25,8 +25,6 @@ message BlobHeader {
   // reserved payments when the same blob is submitted multiple times within the same reservation period. On-demand
   // payments already have unique cumulative_payment values for intentionally unique dispersal requests.
   uint32 salt = 5;
-  // signature over keccak hash of the blob_header that can be verified by blob_header.account_id
-  bytes signature = 6;
 }
 
 // BlobCertificate contains a full description of a blob and how it is dispersed. Part of the certificate
```
jianoaix, line 30: This comment (`a full description of a blob...`) gives a good hint about a better naming for BlobCertificate, it's `BlobDescriptor`. The `TBSBlobCertificate` is more accurate than `BlobCertificate`, but `BlobDescriptor` is further better than `TBSBlobCertificate` as a name.

api/proto/disperser/v2/disperser_v2.proto
ian-shim: I think we settled on still calling this blob, correct me if im wrong @samlaf

api/proto/common/v2/common_v2.proto
ian-shim: Agree this is not 100% accurate naming, but as we move `signature` from `BlobHeader` to `BlobCertificate`, this was more appropriate than before

api/proto/common/v2/common_v2.proto
```diff
@@ -36,12 +34,14 @@ message BlobHeader {
 message BlobCertificate {
   // blob_header contains data about the blob.
   BlobHeader blob_header = 1;
-  // relays is the list of relays that are in custody of the blob.
+  // signature over keccak hash of the blob_header that can be verified by blob_header.payment_header.account_id
```
hopeyen, line 37: Sam suggested a more detailed description for this on my PR for salt. Pasting my changes here for reference but it is outdated for the new structs here
```
// Signature is an ECDSA signature signed by the blob request signer's account ID over the BlobHeader's blobKey,
// which is a keccak hash of the serialized BlobHeader, and used to verify against blob dispersal request's account ID
```

disperser/apiserver/disperse_blob_v2.go
```diff
@@ -94,7 +95,11 @@ func (s *DispersalServerV2) StoreBlob(ctx context.Context, data []byte, blobHead
 }
 
 func (s *DispersalServerV2) validateDispersalRequest(ctx context.Context, req *pb.DisperseBlobRequest, onchainState *OnchainState) error {
-	data := req.GetData()
+	signature := req.GetSignature()
+	if len(signature) != 65 {
+		return api.NewErrorInvalidArg(fmt.Sprintf("signature is expected to be 65 bytes, but got %d bytes", len(signature)))
+	}
+	data := req.GetBlob()
```
hopeyen, line 102: nit: `blob := req.GetBlob()`?

api/proto/common/v2/common_v2.proto
jianoaix: I don't actually see it's more accurate, this "signature" is signed by the owner of this blob (*), not the validators. A TBSCertificate becomes only after sufficient validators have signed it.

(*) Can we could name it more specifically with a qualification so it's clear and not be confused with validators' signature?

contracts/src/interfaces/IEigenDAStructs.sol
samlaf: Yes please! Don't mind if its done here or in another PR but ideally all onchain names should match the onchain structs.

api/proto/common/v2/common_v2.proto
ian-shim: Let's settle this one as a follow up and not let it block this PR. 
Simple renaming can be done after audit starts, but want to get this structure change in before that

api/proto/disperser/v2/disperser_v2.proto
samlaf: Yes this is actually a `blob` now. @jianoaix can't be an encodedPayload because encodedPayload can either be interpreted as Eval or Coeffs, whereas a blob is always evaluated as coefficients by disperser:
image

We mentioned we were going to add an comment explanation that as a network optimization, the `blob` is allowed to be passed without suffix 0s and the disperser will pad to the next power of 2.

api/proto/common/v2/common_v2.proto
samlaf: Agree with @ian-shim here. Too bad @jianoaix you should have joined the discussion. Let's continue on slack, but basically I do think that Certificate makes sense now. A BlobCertificate is signed by the blob sender. A DACertificate is signed by the Validators. Both are certificates though.

api/proto/common/v2/common_v2.proto
```diff
@@ -25,8 +25,6 @@ message BlobHeader {
   // reserved payments when the same blob is submitted multiple times within the same reservation period. On-demand
   // payments already have unique cumulative_payment values for intentionally unique dispersal requests.
   uint32 salt = 5;
```
samlaf, line 27: Can we move the salt to position 3 above commitment?
Would make the hashing structure easier to draw out:
image

Unless the point is to have salt as the explicitly last field? But can we guarantee we'll never need to add a 6th field?

api/proto/common/v2/common_v2.proto
```diff
@@ -36,12 +34,15 @@ message BlobHeader {
 message BlobCertificate {
   // blob_header contains data about the blob.
   BlobHeader blob_header = 1;
-  // relays is the list of relays that are in custody of the blob.
+  // signature is an ECDSA signature signed by the blob request signer's account ID over the BlobHeader's blobKey,
+  // which is a keccak hash of the serialized BlobHeader, and used to verify against blob dispersal request's account ID
```
samlaf, line 38: How do you feel about removing blobKey everywhere and using the more explicit blobHeaderHash?
Explicit self-documenting names prevent having to add this kind of detail in every comment everywhere.

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -35,14 +35,16 @@ message DisperseBlobRequest {
...
   // The header contains metadata about the blob.
   //
   // This header can be thought of as an "eigenDA tx", in that it plays a purpose similar to an eth_tx to disperse a
   // 4844 blob. Note that a call to DisperseBlob requires the blob and the blobHeader, which is similar to how
   // dispersing a blob to ethereum requires sending a tx whose data contains the hash of the kzg commit of the blob,
   // which is dispersed separately.
   common.v2.BlobHeader blob_header = 2;
+  // signature over keccak hash of the blob_header that can be verified by blob_header.payment_header.account_id
```
samlaf, line 46: I'm fussy about this but I think it would be good to be super precise about the exact thing we are signing over. "keccak hash of blob_header" is not precise as it doesn't describe the serialization procedure (nor the fact that we use multi-level hashing).
image

Perhaps this is one reason to keep `BlobKey` around, and be super precise in its documentation that it refers to a very specific hashing/serialization combination of the BlobHeader? Then we can just use `BlobKey` everywhere and refer people to its documentation comment.

api/proto/disperser/v2/disperser_v2.proto
jianoaix: A blob has a power of 2 length, but this one doesn't have to

core/v2/types.go
```diff
@@ -170,6 +164,10 @@ type RelayKey = uint32
 type BlobCertificate struct {
 	BlobHeader *BlobHeader
 
+	// Signature is an ECDSA signature signed by the blob request signer's account ID over the BlobHeader's blobKey,
```
samlaf, line 167:
```suggestion
	// Signature is an ECDSA signature signed by the blob request signer's account ID over the blobKey,
```

samlaf: blobheader's blobkey sounds like blobkey is a field of blobheader

api/proto/common/v2/common_v2.proto
jianoaix: Go ahead for the separation of PRs.
Isn't the renaming "simple" as well?

I don't know what specific discussion you are talking about, but I don't think it's not too late to have discussion.

samlaf (commented):
Overall LGTM, thanks for making these changes!

2 last things I care about getting right that I don't see in this PR (perhaps they belong in a follow-up PR?) are:
1. renaming EigenDABlobVerifier.verifyBlobV2 function to verifyCertV2 (because its verifying a cert, not a blob!) cc @0x0aa0 
2. `blobKey` vs `blobHeaderHash`. I think I would still prefer we get rid of `blobKey` and use the more precise and self-documenting `blobHeaderHash` everywhere.

api/proto/disperser/v2/disperser_v2.proto
samlaf: Reread my last sentence. We agreed that the disperser endpoints accepts "conceptual blobs". Documentation has to be updated to reflect this. But basically to save network bandwidth the disperser allows you to send a `blob` with the last 0s removed (as long as you don't remove more then half), and itll pad the last zeros to the next power of 2 to turn it into an actual blob.

What you're dispersing is still a blob conceptually.

jianoaix: I'm confused at this: isn't all the effort exactly to bring it from conceptual level to actual semantics level?

If we are happy at conceptual level, there isn't even a need for payload/encodedPayload etc, they are all blobs conceptually anyway. The "what is a blob" confusion is exactly due to that conceptual blob is too vague.

api/proto/common/v2/common_v2.proto
ian-shim: Why does salt have to be the last field in the struct? 
Is this suggestion just to make the diagram prettier?

cody-littley (approved):

0x0aa0 (approved):

api/proto/common/v2/common_v2.proto
samlaf: doesn't have to be, was asking if that's what you were trying to achieve.
Yes.

samlaf: Well not just the diagram. I feel everything that gets hashed together should be next to each other. Just easier to not have bugs by forgetting one field that was far, for eg when replicating the new hashing structure onchain.

api/proto/common/v2/common_v2.proto
samlaf: Don't think it does, because it literally plays no purpose. It's a symlink that shouldn't exist. An indirection level that is not helpful. But I'm losing the battle here so will resolve.

PR #1137, opened 2025-01-22
docs: add style guide

samlaf:

Currently only added very bare bones style guide for golang. 
Solidity still TODO: I'm not the best person to fill this in.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/STYLE.md
+++ b/STYLE.md
@@ -1,0 +1,78 @@
+# Style Guide
+
+The eigenda repo consists majorly of Go and Solidity code, so we break it up into two sections:
+- [Go Style Guide](#go-style-guide)
+- [Solidity Style Guide](#solidity-style-guide)
+
+Our implicit rule is to defer to each language's respective style guides (see [resources](#resources) section). 
+However, some disagreements keep coming up in PRs, so we document our conventions here to prevent too much bike-shedding.
+This style guide is thus by definition perpetually a work in progress.
+
+## Go Style Guide
+
+This document outlines the coding standards and best practices for our Go projects.
+
+### Code Organization
+
+Follow the [Ben Johnson](https://medium.com/@benbjohnson/structuring-applications-in-go-3b04be4ff091) repo structuring method:
+- Write interfaces in a file at the root of each package, and
+- Write every implementation of that interface in its own package in a subdirectory.
+
+#### Function Organization
+- Keep functions small and focused
+- Limit function length to ~50 lines where possible
+- Place the most important functions at the top of the file
+- Public functions that aren't methods should be placed in files with `_utils.go` suffix
+
+### Error Handling
+
+- We follow Uber's [error-wrapping style](https://github.com/uber-go/guide/blob/master/style.md#error-wrapping).
+
+### Comments and Documentation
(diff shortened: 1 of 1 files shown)


STYLE.md
```diff
@@ -0,0 +1,78 @@
...
+- Place the most important functions at the top of the file
+- Public functions that aren't methods should be placed in files with `_utils.go` suffix
+
+### Error Handling
+
+- We follow Uber's [error-wrapping style](https://github.com/uber-go/guide/blob/master/style.md#error-wrapping).
+
+### Comments and Documentation
+
+- Document all exported functions, types, and variables.
```
litt3, line 33: I agree we should enforce docs for exported stuff, but I think we should also strongly encourage docs for ANYTHING non-trivial 

"Does the name of this thing tell me everything I could possibly need to know?"
-> yes: ok to skip doc
-> no: write a doc, regardless of whether it's exported

STYLE.md
```diff
@@ -0,0 +1,78 @@
...
+
+Follow the [Ben Johnson](https://medium.com/@benbjohnson/structuring-applications-in-go-3b04be4ff091) repo structuring method:
+- Write interfaces in a file at the root of each package, and
+- Write every implementation of that interface in its own package in a subdirectory.
+
+#### Function Organization
+- Keep functions small and focused
+- Limit function length to ~50 lines where possible
+- Place the most important functions at the top of the file
+- Public functions that aren't methods should be placed in files with `_utils.go` suffix
```
litt3, line 25: This isn't always the case, i.e. constructors. Perhaps reword to say something like:

"Public static functions that lack a tight coupling to a specific struct should be placed in files with `_utils.go` suffix"

STYLE.md
litt3, line 33: We should also enforce param docs for exported functions

STYLE.md
```diff
@@ -0,0 +1,78 @@
...
+
+Our implicit rule is to defer to each language's respective style guides (see [resources](#resources) section). 
+However, some disagreements keep coming up in PRs, so we document our conventions here to prevent too much bike-shedding.
+This style guide is thus by definition perpetually a work in progress.
+
+## Go Style Guide
+
+This document outlines the coding standards and best practices for our Go projects.
+
+### Code Organization
```
litt3, line 15: Consider adding a section about small *files*, in addition to small functions

Mega files that contain multiple structs are no good. If a struct is non-trivial, it ought to have a dedicated file

STYLE.md
cody-littley:
```
"Public static functions that lack a tight coupling to a specific struct (e.g. a constructor) should be placed in files with _utils.go suffix"
```

samlaf:
TODO: add guide for how configs and constructors should be created (use structs, pointers to struct, how to validate config and inject defaults when not present, etc). See [this discussion](https://github.com/Layr-Labs/eigenda/pull/1159#discussion_r1929337220)

samlaf:
Closing in favor of #1740

PR #1139, opened 2025-01-23
Misc documentation fix

jianoaix:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


core/v2/types.go
```diff
@@ -309,7 +309,7 @@ func BatchFromProtobuf(proto *commonpb.Batch) (*Batch, error) {
 type Attestation struct {
 	*BatchHeader
 
-	// AttestedAt is the time the attestation was made
+	// AttestedAt is the time the attestation was made in nanoseconds
```
samlaf, line 312: Why not change the variable name itself to `AttestedAtNano`

core/v2/types.go
```diff
@@ -321,7 +321,7 @@ type Attestation struct {
 	Sigma *core.Signature
 	// QuorumNumbers contains the quorums relevant for the attestation
 	QuorumNumbers []core.QuorumID
-	// QuorumResults contains the results of the quorum verification
+	// QuorumResults contains the operators' total signing percentage of the quorum
```
samlaf, line 324:
```suggestion
	// QuorumResults is a map from quorumID to the operators' total signing percentage for that quorum.
```

samlaf: which quorums have an entry? All the quorums found in the blobHeaders for the given batch?

disperser/dataapi/docs/v1/V1_docs.go
samlaf, line 1: Are these all auto-generated? Might be good to add to https://github.com/Layr-Labs/eigenda/blob/master/.gitattributes if so

disperser/dataapi/docs/v2/V2_docs.go
```diff
@@ -572,10 +572,6 @@ const docTemplateV2 = `{
                 "reservation_period": {
                     "description": "ReservationPeriod represents the range of time at which the dispersal is made",
                     "type": "integer"
-                },
-                "salt": {
-                    "description": "Allow same blob to be dispersed multiple times within the same reservation period",
-                    "type": "integer"
```
samlaf, line 578: Interesting. Why were these docs not updated as part of the other PRs? Are these just hand crated? If not, should we have a ci to check that they have been regenerated correctly?

disperser/dataapi/docs/v1/V1_docs.go
jianoaix: we are using swagger here

core/v2/types.go
jianoaix: each quorum has its own signing rate and different quorums may have different rates

disperser/dataapi/docs/v2/V2_docs.go
jianoaix: I think we should

core/v2/types.go
jianoaix: It's better to remain consistent with other timestamps in naming. I don't have objection if one wants to get all timestamps renamed by incorporating time units though.

samlaf: gotcha, agree its best to be consistent.

disperser/dataapi/docs/v2/V2_docs.go
jianoaix: It may be good starter task for infra, filed ticket: https://github.com/Layr-Labs/eigenda/issues/1142

jianoaix:
Ping to get this in?

samlaf (approved):

PR #1140, opened 2025-01-23
Relay client should limit max concurrency

cody-littley:

In preprod, we observe relays rate limiting concurrent `GetChunks()` requests. This is non-ideal in an environment where the relays are not under intentional attack or high load.

The following screenshot is a histogram of the logs reporting concurrency throttling.

Screenshot 2025-01-23 at 9 10 50 AM

This PR introduces two primary changes:
- increases the maximum permissible concurrency per relay client from 1 to 2
- adds logic to relay clients that causes them to avoid exceeding maximum permissible concurrency

### DIFF
--- a/api/clients/v2/relay_client.go
+++ b/api/clients/v2/relay_client.go
@@ -23,6 +23,8 @@ type RelayClientConfig struct {
 	UseSecureGrpcFlag bool
 	OperatorID        *core.OperatorID
 	MessageSigner     MessageSigner
+	// MaxConcurrentRequests is the maximum number of concurrent requests to a particular relay server.
+	MaxConcurrentRequests uint
 }
 
 type ChunkRequestByRange struct {
@@ -52,6 +54,17 @@ type RelayClient interface {
 	Close() error
 }
 
+// grpcClient encapsulates a gRPC client and a channel for limiting concurrent requests to that client.
+type grpcClient struct {
+	// client is the gRPC client used for sending gRPC requests.
+	client relaygrpc.RelayClient
+	// permits is a channel for limiting the number of concurrent requests to the client.
+	// when a request is initiated, a value is sent to the channel, and when the request is completed,
+	// the value is received. The channel has a buffer size of `MaxConcurrentRequests`, and will block
+	// if the number of concurrent requests exceeds this limit.
+	permits chan struct{}
+}
+
 type relayClient struct {
 	config *RelayClientConfig
 
@@ -62,7 +75,7 @@ type relayClient struct {
 	conns  sync.Map
 	logger logging.Logger
 
-	// grpcClients maps relay key to the gRPC client: `map[corev2.RelayKey]relaygrpc.RelayClient`
+	// grpcClients maps relay key to the gRPC client: `map[corev2.RelayKey]*grpcClient`
 	grpcClients sync.Map
 }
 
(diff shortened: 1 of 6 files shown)


api/clients/v2/relay_client.go
```diff
@@ -52,6 +54,20 @@ type RelayClient interface {
 	Close() error
 }
 
+// grpcClient encapsulates a gRPC client and a channel for limiting concurrent requests to that client.
+type grpcClient struct {
```
litt3, line 58: What do you think of the name `limitedClient`, or maybe `limitedConcurrencyClient`? 

Not sure if these are good suggestions, but `grpcClient` doesn't indicate the reason this struct exists

relay/cmd/flags/flags.go
```diff
@@ -199,7 +199,7 @@ var (
 		Usage:    "Max number of concurrent GetChunk operations per client",
 		Required: false,
 		EnvVar:   common.PrefixEnvVar(envVarPrefix, "MAX_CONCURRENT_GET_CHUNK_OPS_CLIENT"),
-		Value:    1,
+		Value:    2, // default value should stay in sync with the default value of node.Config.RelayConcurrency
```
ian-shim, line 202: can we increase this further like 8 or 16? (obviously this is just default value and real value will be configured in the cluster, but wondering what the most accommodating value should be for clients calling relay)

api/clients/v2/relay_client.go
```diff
@@ -52,6 +54,20 @@ type RelayClient interface {
...
+	// client is the gRPC client used for sending gRPC requests.
+	client relaygrpc.RelayClient
```
litt3, line 60: Style question:

Did you consider embedding the `relaygrpc.RelayClient` in this new struct? Doing so avoids the awkward `client.client` syntax brought on by the current setup

api/clients/v2/relay_client.go
```diff
@@ -95,7 +111,17 @@ func (c *relayClient) GetBlob(ctx context.Context, relayKey corev2.RelayKey, blo
 		return nil, err
 	}
 
-	res, err := client.GetBlob(ctx, &relaygrpc.GetBlobRequest{
+	select {
```
ian-shim, line 114: Why don't we return the error immediately (in case of rate limits at the relay, it will return ResourceExhausted error) instead of blocking and waiting? This would allow the user of this client to handle the error however it wants

api/clients/v2/relay_client.go
cody-littley: renamed to `limitedClient`

api/clients/v2/relay_client.go
cody-littley: I've refactored this... the concurrency limiting is now completely encapsulated inside the struct `limitedClient`.

relay/cmd/flags/flags.go
cody-littley: increased to 8

api/clients/v2/relay_client.go
cody-littley: Currently we don't have retry logic when fetching from a relay. The idea here was that it was better to slightly delay a request to a relay than it was to fail to fetch data (and potentially not sign a batch).

api/clients/v2/relay_client.go
```diff
@@ -248,7 +250,9 @@ func (c *relayClient) initOnceGrpcConnection(key corev2.RelayKey) error {
 			return
 		}
 		c.conns.Store(key, conn)
-		c.grpcClients.Store(key, relaygrpc.NewRelayClient(conn))
+
+		wrappedClient := newLimitedRelayClient(relaygrpc.NewRelayClient(conn), key, c.config.MaxConcurrentRequests)
```
litt3, line 254: mini nit: var name is still `wrappedClient`, could be changed to `limitedClient`

litt3 (approved):

api/clients/v2/relay_client.go
cody-littley: fixed

litt3 (approved):

api/clients/v2/relay_client.go
ian-shim: I think this choice should be made not inside the client but by the users of the client, and client being a thin wrapper for interacting with the relay. 
Should we add this functionality at the node side if we really do want this kind of limiting logic from the client side? 
I'm fine with not putting in this limiting mechanism at all from the client side (node) in favor of potentially retrying with backoff in case of ResourceExhausted error. Because the users of this relay client don't know the rate limiting configuration set by the relay, it's highly likely the client side accounting is not going to be correct.

PR #1147, opened 2025-01-23
Fallback ip

cody-littley:

If an IP provider doesn't work, try a fallback.

### DIFF
--- a/common/pubip/multi_provider.go
+++ b/common/pubip/multi_provider.go
@@ -1,0 +1,52 @@
+package pubip
+
+import (
+	"context"
+	"fmt"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	"strings"
+)
+
+var _ Provider = (*multiProvider)(nil)
+
+// An implementation of Provider that uses multiple providers. It attempts each provider in order until one succeeds.
+type multiProvider struct {
+	logger    logging.Logger
+	providers []Provider
+}
+
+func (m *multiProvider) Name() string {
+	sb := strings.Builder{}
+	sb.WriteString("multiProvider(")
+	for i, provider := range m.providers {
+		sb.WriteString(provider.Name())
+		if i < len(m.providers)-1 {
+			sb.WriteString(", ")
+		}
+	}
+	sb.WriteString(")")
+	return sb.String()
+}
+
+// NewMultiProvider creates a new multiProvider with the given providers.
+func NewMultiProvider(
+	logger logging.Logger,
+	providers ...Provider) Provider {
+
+	return &multiProvider{
+		logger:    logger,
+		providers: providers,
+	}
+}
+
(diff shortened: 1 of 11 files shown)


common/pubip/multi_provider.go
```diff
@@ -0,0 +1,52 @@
...
+	}
+}
+
+func (m *multiProvider) PublicIPAddress(ctx context.Context) (string, error) {
+	for _, provider := range m.providers {
+		ip, err := provider.PublicIPAddress(ctx)
+		if err == nil {
+			return ip, nil
+		}
+		m.logger.Errorf("failed to get public IP address from %s: %v", provider, err)
```
ian-shim, line 48: WARN?

node/flags/flags.go
```diff
@@ -150,7 +150,7 @@ var (
 	}
 	PubIPProviderFlag = cli.StringFlag{
```
ian-shim, line 151: we could use `StringSliceFlag` here

ian-shim (commented):
lgtm couple comments

ian-shim (approved):

common/pubip/multi_provider.go
cody-littley: changed to a warning

node/flags/flags.go
cody-littley: good suggestion, change made

litt3 (approved):

supriya-premkumar:
Might I suggest a simpler approach?

```
package main

import (
	"context"
	"fmt"
	"net"
	"time"
)


const openDNS = "208.67.222.222:53"

func main() {
	// Create a custom resolver that talks to openDNS resolvers
	r := &net.Resolver{
		PreferGo: true,
		Dial: func(ctx context.Context, network, address string) (net.Conn, error) {
			d := net.Dialer{Timeout: 3 * time.Second}
			return d.DialContext(ctx, network, openDNS)
		},
	}

	// This responds with your public IP using the custom resolver
	ips, err := r.LookupHost(context.Background(), "myip.opendns.com")
	if err != nil {
		fmt.Println("Error looking up public IP:", err)
		return
	}

	for _, ip := range ips {
		fmt.Println("Public IP:", ip)
	}
}

```
Here is why I think this might be a better approach

-  Uses only stdlib
- Doesn't open TCP socket, gets the response by very cheap DNS lookups, making it faster.
- Avoids all the http status checks and error handling
- Simpler response parsing
- Return response is a string parsed out of a native IP object
- Removes dependency on third party API schema

Let me know your thoughts.

cody-littley:
Repeating offline discussion for the record. @supriya-premkumar's suggestion is a good one, but since CI is currently broken and this fixes it, we are going to merge this PR and consider alternate solutions to resolving self IP addresses as follow up work.

PR #1151, opened 2025-01-24
Downgrade error logs

ian-shim:
Downgrading some error logs into warn level

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/relay/server.go
+++ b/relay/server.go
@@ -201,7 +201,7 @@ func NewServer(
 
 	return &Server{
 		config:           config,
-		logger:           logger,
+		logger:           logger.With("component", "RelayServer"),
 		metadataProvider: mp,
 		blobProvider:     bp,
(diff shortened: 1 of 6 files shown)


relay/server.go
```diff
@@ -233,6 +233,8 @@ func (s *Server) GetBlob(ctx context.Context, request *pb.GetBlobRequest) (*pb.G
 		return nil, api.NewErrorInvalidArg(fmt.Sprintf("invalid blob key: %v", err))
 	}
 
+	s.logger.Debug("GetBlob request received", "key", request.BlobKey)
```
dmanc, line 236: Think we should use .Hex() here

dmanc (approved):

cody-littley (approved):

PR #1152, opened 2025-01-24
Adds isolated v2 reachability check to operator node

pschork:
Creates v2 specific reachability check in v2 dataAPI
Isolates v1 & v2 node reachability checks
- ProbeV1OperatorPorts (calls dataAPI v1)
- ProbeV2OperatorPorts (calls dataAPI v2)

Adds RuntimeMode flag that accepts `v1-only` (default), `v2-only`, `v2-and-v1`

Fix startup logic prevented startup when old `--enableV2` was not set
```
2025/01/31 11:26:53 application failed: enable v2 flag needs to be set when v2 dispersal port (NODE_V2_DISPERSAL_PORT) is specified
```
but if V2 ports where unset, node would still fail because V2 ports were required
```
2025/01/31 12:13:41 application failed: Required flags "node.v2-dispersal-port, node.v2-retrieval-port" not set
```

Update dataAPI swagger

Add Makefile docker-node-group for testing local node
Add Makefile run target for running local node (requires local .env)
Add Makefile run-update-socket target (requires local .env)
Add Makefile run-list-quorums target (requires local .env)

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/common/v2/blob.go
+++ b/disperser/common/v2/blob.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 
 	pb "github.com/Layr-Labs/eigenda/api/grpc/disperser/v2"
-	core "github.com/Layr-Labs/eigenda/core/v2"
+	corev2 "github.com/Layr-Labs/eigenda/core/v2"
 	"github.com/Layr-Labs/eigenda/encoding"
 )
 
@@ -71,7 +71,7 @@ func BlobStatusFromProtobuf(s pb.BlobStatus) (BlobStatus, error) {
 
 // BlobMetadata is an internal representation of a blob's metadata.
 type BlobMetadata struct {
-	BlobHeader *core.BlobHeader
+	BlobHeader *corev2.BlobHeader
 	Signature  []byte
 
 	// BlobStatus indicates the current status of the blob
--- a/disperser/common/v2/blob.go
+++ b/disperser/common/v2/blob.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 
 	pb "github.com/Layr-Labs/eigenda/api/grpc/disperser/v2"
-	corev2 "github.com/Layr-Labs/eigenda/core/v2"
+	core "github.com/Layr-Labs/eigenda/core/v2"
 	"github.com/Layr-Labs/eigenda/encoding"
 )
 
@@ -71,7 +71,7 @@ func BlobStatusFromProtobuf(s pb.BlobStatus) (BlobStatus, error) {
 
 // BlobMetadata is an internal representation of a blob's metadata.
 type BlobMetadata struct {
-	BlobHeader *corev2.BlobHeader
+	BlobHeader *core.BlobHeader
 	Signature  []byte
 
 	// BlobStatus indicates the current status of the blob
--- a/disperser/dataapi/operator_handler.go
+++ b/disperser/dataapi/operator_handler.go
@@ -84,7 +84,7 @@ func NewOperatorHandler(logger logging.Logger, metrics *Metrics, chainReader cor
 	}
 }
 
-func (oh *OperatorHandler) ProbeOperatorHosts(ctx context.Context, operatorId string) (*OperatorPortCheckResponse, error) {
+func (oh *OperatorHandler) ProbeV2OperatorPorts(ctx context.Context, operatorId string) (*OperatorPortCheckResponse, error) {
 	operatorInfo, err := oh.subgraphClient.QueryOperatorInfoByOperatorId(ctx, operatorId)
 	if err != nil {
 		oh.logger.Warn("failed to fetch operator info", "operatorId", operatorId, "error", err)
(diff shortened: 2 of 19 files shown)


disperser/common/v2/blob.go
```diff
@@ -4,7 +4,7 @@ import (
 	"fmt"
 
 	pb "github.com/Layr-Labs/eigenda/api/grpc/disperser/v2"
-	core "github.com/Layr-Labs/eigenda/core/v2"
+	corev2 "github.com/Layr-Labs/eigenda/core/v2"
```
pschork, line 7: FYI this fixes swagger docs to resolve `v2.BlobHeader` (previously it was resolving v1 `core.BlobHeader`)

pschork: Moved to separate PR https://github.com/Layr-Labs/eigenda/pull/1199/files

disperser/dataapi/operator_handler.go
```diff
@@ -84,73 +84,90 @@ func NewOperatorHandler(logger logging.Logger, metrics *Metrics, chainReader cor
...
+		dispersalStatus = "v2 dispersal port is not registered"
 	} else {
-		v2RetrievalPortOpen := checkIsOperatorPortOpen(v2RetrievalSocket, 3, oh.logger)
-		if !v2RetrievalPortOpen {
-			v2RetrievalStatus = "port closed or unreachable"
-		} else {
-			v2RetrievalOnline, v2RetrievalStatus = checkServiceOnline(ctx, "node.v2.Retrieval", v2RetrievalSocket, 3*time.Second)
+		dispersalPortOpen := checkIsOperatorPortOpen(dispersalSocket, 3, oh.logger)
+		if dispersalPortOpen {
+			dispersalOnline, dispersalStatus = checkServiceOnline(ctx, "validator.Dispersal", dispersalSocket, 3*time.Second)
```
supriya-premkumar, line 114: I like removing the version from the service name. eventually we should move versioning to package levels to better support compat issues.

disperser/dataapi/operator_handler.go
```diff
@@ -340,7 +357,7 @@ func (oh *OperatorHandler) CreateOperatorQuorumIntervals(
 }
 
 func (oh *OperatorHandler) getOperatorQuorumEvents(
-	ctx context.Context,
+	_ context.Context,
```
supriya-premkumar, line 360: curious on why we are are dropping ctx here?

node/config.go
```diff
@@ -232,57 +233,39 @@ func NewConfig(ctx *cli.Context) (*Config, error) {
...
-		if err != nil {
+		if v2RetrievalPort == "" {
+			return nil, fmt.Errorf("v2 retrieval port (NODE_V2_RETRIEVAL_PORT) must be defined when v2 is enabled")
+		} else if core.ValidatePort(v2RetrievalPort) != nil {
 			return nil, fmt.Errorf("invalid v2 retrieval port: %s", v2RetrievalPort)
 		}
 	}
 
+	if v1Disabled && !v2Enabled {
+		return nil, fmt.Errorf("invalid configuration: both v1 and v2 are disabled")
```
supriya-premkumar, line 266: vetting nit: `Errorf` used without a format specifier. Can consider `errors.New("invalid configuration: both v1 and v2 are disabled")`

node/flags/flags.go
```diff
@@ -231,12 +232,18 @@ var (
...
 	}
-	EnableV2Flag = cli.BoolTFlag{
+	EnableV2Flag = cli.BoolFlag{
 		Name:     "enable-v2",
-		Usage:    "Enable V2 features",
+		Usage:    "Enable V2 features (default=false)",
 		Required: false,
 		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "ENABLE_V2"),
 	}
+	DisableV1Flag = cli.BoolFlag{
```
ian-shim, line 241: One is enable and the other is disable 🥲 Do we want to have these two flag control the mode? I thought we wanted to add a `mode` flag that chooses one of 3 options: `v1-only`, `v2-only`, `v1-and-v2`

disperser/dataapi/operator_handler.go
pschork: Lint. The `ctx` is not used.

disperser/dataapi/operator_handler.go
pschork: The renaming was done as part of documentation overhaul in https://github.com/Layr-Labs/eigenda/pull/1125

node/flags/flags.go
pschork: Added `RuntimeMode` flag. I added `Runtime` context so that env var looks like `NODE_RUMTIME_MODE` vs `NODE_MODE` which seems kinda lacking. But I don't have strong attachment if folks prefer `NODE_MODE`

supriya-premkumar (approved):
:shipit:

node/config.go
```diff
@@ -185,14 +187,14 @@ func NewConfig(ctx *cli.Context) (*Config, error) {
 		blsSignerAPIKey := ctx.GlobalString(flags.BLSSignerAPIKeyFlag.Name)
 
 		if blsRemoteSignerEnabled && (blsRemoteSignerUrl == "" || blsPublicKeyHex == "") {
-			return nil, fmt.Errorf("BLS remote signer URL and Public Key Hex is required if BLS remote signer is enabled")
+			return nil, errors.New("BLS remote signer URL and Public Key Hex is required if BLS remote signer is enabled")
```
ian-shim, line 190: 👨‍🍳

node/node.go
```diff
@@ -276,13 +279,16 @@ func (n *Node) Start(ctx context.Context) error {
...
+	}
 
 	if n.Config.EnableV2 {
 		go func() {
 			_ = n.RefreshOnchainState(ctx)
 		}()
+		go n.checkNodeReachability(v2CheckPath)
 	}
 
 	// Build the socket based on the hostname/IP provided in the CLI
```
ian-shim, line 294: should we disable the IP discovery/update in L344 if only v2 is enabled?

ian-shim (approved):

node/node.go
pschork: Good question. I'm gonna land this and think some more on IP discovery changes in new PR.

PR #1154, opened 2025-01-24
onchain renaming

0x0aa0:
- BlobVerificationProofV2 -> BlobInclusionInfo
- EigenDABlobVerifier -> EigenDACertVerifier
- verifyBlob() -> verifyDACert()


contracts/src/core/EigenDACertVerifier.sol
```diff
@@ -44,11 +44,11 @@ contract EigenDABlobVerifier is IEigenDABlobVerifier {
      * @param blobHeader The blob header to verify
      * @param blobVerificationProof The blob verification proof to verify the blob against
      */
-    function verifyBlobV1(
+    function verifyDACertV1(
```
samlaf, line 47: Need to update the docstrings everywhere, which still mention "verifies the blob" instead of cert

contracts/src/core/EigenDACertVerifier.sol
```diff
@@ -80,45 +80,45 @@ contract EigenDABlobVerifier is IEigenDABlobVerifier {
     /**
      * @notice Verifies a blob for the base required quorums and the default security thresholds
```
samlaf, line 81:
```suggestion
     * @notice Verifies a cert for the base required quorums and the default security thresholds
```

contracts/src/core/EigenDACertVerifier.sol
```diff
@@ -80,45 +80,45 @@ contract EigenDABlobVerifier is IEigenDABlobVerifier {
...
-     * @param blobVerificationProof The blob verification proof for the blob
+     * @param blobInclusionInfo The inclusion proof for the blob
      * @param nonSignerStakesAndSignature The nonSignerStakesAndSignature for the blob
      */
-    function verifyBlobV2(
+    function verifyDACertV2(
         BatchHeaderV2 calldata batchHeader,
-        BlobVerificationProofV2 calldata blobVerificationProof,
+        BlobInclusionInfo calldata blobInclusionInfo,
         NonSignerStakesAndSignature calldata nonSignerStakesAndSignature
```
samlaf, line 89: Could we take this change as an opportunity to create an explicit DACert struct, both onchain and offchain, that would hold these 3 things together? Would canonicalize the name and remove any ambiguities going forward as to what a DACert is.

contracts/src/core/EigenDACertVerifier.sol
```diff
@@ -80,45 +80,45 @@ contract EigenDABlobVerifier is IEigenDABlobVerifier {
...
     }
 
     /**
      * @notice Verifies a blob for the base required quorums and the default security thresholds
      * @param signedBatch The signed batch to verify the blob against
-     * @param blobVerificationProof The blob verification proof for the blob
+     * @param blobInclusionInfo The inclusion proof for the blob
      */
-    function verifyBlobV2FromSignedBatch(
+    function verifyDACertV2FromSignedBatch(
```
samlaf, line 108: Curious why the use of the `FromSignedBatch` suffix here. Doesn't solidity allow function polymorphism (2 functions with the same name if the argument number or types are different)?

contracts/src/core/EigenDACertVerifier.sol
samlaf: Also overall feel like we should keep the `SignedBatch` semantics, even when transforming the attestation into a NonSignerStakesAndSignature (because that's just a transformation of the exact same information, meant to be consumed by the contract differently).
image
Perhaps we could use `SignedBatch` and `SignedBatchWithNonsigners` or something? And then a DACert would be either {SignedBatch, BlobInclusionInfo} or {SignedBatchWithNonSigners, BlobInclusionInfo} and it would be more obvious that those two things are just one eth_call away, but have the same semantic. Right now people are always confused as to how the 2 verify functions we have are related, because syntactically they look very different.

contracts/src/interfaces/IEigenDAStructs.sol
```diff
@@ -82,8 +82,8 @@ struct BlobHeaderV2 {
     uint16 version;
     bytes quorumNumbers;
     BlobCommitment commitment;
-    uint32 salt;
     bytes32 paymentHeaderHash;
+    uint32 salt;
```
samlaf, line 86: Isn't this struct being called `BlobHeaderV2` a bit limitating? What if we ever in the future want to also reveal payments onchain? Then we'll have to create another struct with a weird name for that.

Perhaps we could keep `BlobHeaderV2` for the full struct to match offchain code, and use a more specific name here for this struct, like `BlobHeaderV2WithPaymentHash` (there's prob a better name)

contracts/src/core/EigenDACertVerifier.sol
```diff
@@ -80,45 +80,45 @@ contract EigenDABlobVerifier is IEigenDABlobVerifier {
...
+            blobInclusionInfo,
             nonSignerStakesAndSignature,
             getDefaultSecurityThresholdsV2(),
-            blobVerificationProof.blobCertificate.blobHeader.quorumNumbers
+            blobInclusionInfo.blobCertificate.blobHeader.quorumNumbers
         );
     }
 
     /**
      * @notice Verifies a blob for the base required quorums and the default security thresholds
```
litt3, line 104: The `@notice` tag still talks about verifying a "blob", instead of verifying a "cert"

(The other modified functions in this PR also have outdated docs)

contracts/src/core/EigenDACertVerifier.sol
```diff
@@ -147,7 +147,7 @@ contract EigenDABlobVerifier is IEigenDABlobVerifier {
         VersionedBlobParams memory blobParams,
         SecurityThresholds memory securityThresholds
     ) external view {
```
litt3, line 149: I think this function `verifyBlobSecurityParams` should also be renamed (in addition to the subsequent `verifyBlobSecurityParams` function)

contracts/src/libraries/EigenDACertVerificationUtils.sol
```diff
@@ -15,15 +15,15 @@ import {IEigenDARelayRegistry} from "../interfaces/IEigenDARelayRegistry.sol";
...
  * @author Layr Labs, Inc.
  */
-library EigenDABlobVerificationUtils {
+library EigenDACertVerificationUtils {
     using BN254 for BN254.G1Point;
 
     uint256 public constant THRESHOLD_DENOMINATOR = 100;
     
-    function _verifyBlobV1ForQuorums(
+    function _verifyDACertV1ForQuorums(
```
litt3, line 26: The error messages throughout still reference the old function names

contracts/src/interfaces/IEigenDAStructs.sol
```diff
@@ -82,8 +82,8 @@ struct BlobHeaderV2 {
...
 }
 
 struct BlobCommitment {
```
samlaf, line 89: Last name discrepancy (I think) between onchain<>offchain is `BlobCommitment.dataLength`. In the offchain code its just called `BlobCommitment.length`. Can we make them the same? Either `length` (to match `lengthProof` and `lengthCommitment`) or `blobLength`, but not `dataLength` because we've already moved away from the "data" naming to more precise naming.

contracts/src/core/EigenDACertVerifier.sol
```diff
@@ -1,17 +1,17 @@
...
 import {IEigenDASignatureVerifier} from "../interfaces/IEigenDASignatureVerifier.sol";
-import {EigenDABlobVerificationUtils} from "../libraries/EigenDABlobVerificationUtils.sol";
+import {EigenDACertVerificationUtils} from "../libraries/EigenDACertVerificationUtils.sol";
 import {OperatorStateRetriever} from "lib/eigenlayer-middleware/src/OperatorStateRetriever.sol";
 import {IRegistryCoordinator} from "lib/eigenlayer-middleware/src/RegistryCoordinator.sol";
 import {IEigenDARelayRegistry} from "../interfaces/IEigenDARelayRegistry.sol";
 import "../interfaces/IEigenDAStructs.sol";
 
-contract EigenDABlobVerifier is IEigenDABlobVerifier {
+contract EigenDACertVerifier is IEigenDACertVerifier {
```
litt3, line 14: Out of curiosity, when do the auto generated go bindings get updated?

contracts/src/interfaces/IEigenDACertVerifier.sol
```diff
@@ -22,31 +22,31 @@ interface IEigenDABlobVerifier is IEigenDAThresholdRegistry {
...
      * @param blobVerificationProofs The blob verification proofs to verify the blobs against
      */
-    function verifyBlobsV1(
+    function verifyDACertsV1(
         BlobHeader[] calldata blobHeaders,
         BlobVerificationProof[] calldata blobVerificationProofs
     ) external view;
 
     /**
      * @notice Verifies a blob for the required quorums and the default security thresholds
```
samlaf, line 31: Also just realized we use the wording "default security thresholds" everywhere (even the contract [storage slot name](https://github.com/Layr-Labs/eigenda/blob/77d4442aa1b37bdc275173a6b27d917cc161474c/contracts/src/core/EigenDAThresholdRegistryStorage.sol#L30)). 

"default" has a connotation of having multiple options that a user can pick from. But our recent discussions and decisions has moved towards these security thresholds being immutable. Think we should thus get rid of "default" everywhere, and comment that they are immutable.

contracts/src/core/EigenDACertVerifier.sol
0x0aa0: Yes solidity does allow for polymorphism so it is possible but my experience has been that differentiating functions by name has been preferred. In this instance I think it makes sense to keep the names distinct as one is more meant to be called onchain while this one is a helper meant to be called offchain

contracts/src/core/EigenDACertVerifier.sol
0x0aa0: bindings get updated by `make compile-el`

contracts/src/interfaces/IEigenDACertVerifier.sol
0x0aa0: Well by using the library for verification the security thresholds can be passed in by the user so there are other options

samlaf: We changed the code in the proxy and clients to cache the value though. When/how could they be changed?

0x0aa0: by using the verification function directly from the library: https://github.com/Layr-Labs/eigenda/blob/quaq/v2-renaming-onchain/contracts/src/libraries/EigenDACertVerificationUtils.sol#L169

0x0aa0: We also discussed having users be able to reference alternative thresholds even from the BlobVerifier so I think it makes sense to leave a default value here if they choose not to define that alternative

litt3 (approved):

ian-shim:
@samlaf let me know if there's any more outstanding changes needed in this PR. I'll generate bindings and add offchain changes on top of it when we're happy about the onchain updates

contracts/src/interfaces/IEigenDACertVerifier.sol
samlaf: Ok I'm still a bit confused by this whole thing. Created https://linear.app/eigenlabs/issue/DAINT-259/make-security-thresholds-either-rollup-customizable-or-immutable to discuss further, will resolve here.

samlaf:
> @samlaf let me know if there's any more outstanding changes needed in this PR. I'll generate bindings and add offchain changes on top of it when we're happy about the onchain updates

LGTM in this state. The two outstanding convos are nice-to-have but I'm fine to move on without them.

samlaf (approved):

PR #1155, opened 2025-01-24
Rename existing client to PayloadRetriever

litt3:

- This PR changes the name of the existing v2 client to `PayloadRetriever`
    - The decision was made to keep `GET` and `PUT` functionality separate, so a more specific name was required for the current struct

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/payload_retriever.go
+++ b/api/clients/v2/payload_retriever.go
@@ -250,9 +250,12 @@ func (pr *PayloadRetriever) verifyCertWithTimeout(
 //
 // This method should only be called once.
 func (pr *PayloadRetriever) Close() error {
-	relayClientErr := pr.relayClient.Close()
+	err := pr.relayClient.Close()
+	if err != nil {
+		return fmt.Errorf("close relay client: %w", err)
+	}
 
-	return fmt.Errorf("close relay client: %w", relayClientErr)
+	return nil
 }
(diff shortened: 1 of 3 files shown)


api/clients/v2/payload_retriever.go
```diff
@@ -17,29 +17,29 @@ import (
 	gethcommon "github.com/ethereum/go-ethereum/common"
 )
 
-// EigenDAClient provides the ability to get payloads from the relay subsystem, and to send new payloads to the disperser.
+// PayloadRetriever provides the ability to get payloads from the relay subsystem.
```
samlaf, line 20: Will we use the same struct later to also do distributed retrievals? Or will that be a different one? PayloadRetriever in my mind should be able to do both, in case the relays are all censoring.

litt3: I agree, distributed retrievals should be in this same struct.

I will amend to comment to declare the future intention

litt3: added [4fa78a99](https://github.com/Layr-Labs/eigenda/pull/1155/commits/4fa78a999dc01392d9b708090d913c72230ad760)

cody-littley (approved):

PR #1156, opened 2025-01-24
Add better error reporting for relay rate limits

cody-littley:

I am currently investigating a relay rate limiting issue. This PR adds more informative error messages when we hit a relay rate limit.

### DIFF
--- a/relay/server.go
+++ b/relay/server.go
@@ -336,7 +336,7 @@ func (s *Server) GetChunks(ctx context.Context, request *pb.GetChunksRequest) (*
 		if strings.Contains(err.Error(), "internal error") {
 			return nil, api.NewErrorInternal(err.Error())
 		}
-		return nil, api.NewErrorResourceExhausted(fmt.Sprintf("bandwidth limit exceeded: %v", err))
+		return nil, buildInsufficientGetChunksBandwidthError(request, mMap, err)
 	}
 	s.metrics.ReportChunkDataSize(requiredBandwidth)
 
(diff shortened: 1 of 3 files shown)


relay/server.go
```diff
@@ -469,6 +469,37 @@ func computeChunkRequestRequiredBandwidth(request *pb.GetChunksRequest, mMap met
...
+		if chunkRequest.GetByIndex() != nil {
+			prefix = "i"
+			chunkCount = len(chunkRequest.GetByIndex().ChunkIndices)
+		} else {
+			prefix = "r"
+			chunkCount = int(chunkRequest.GetByRange().EndIndex - chunkRequest.GetByRange().StartIndex)
+		}
+		chunkSize := int(mMap[v2.BlobKey(chunkRequest.GetByRange().GetBlobKey())].chunkSizeBytes)
+
+		sb.WriteString(fmt.Sprintf("%s(%d*%dbytes)", prefix, chunkCount, chunkSize))
```
jianoaix, line 494: It seems this can be a quite large message to be returned, not sure if it's good idea to do it in production mode.

ian-shim (approved):

relay/server.go
ian-shim: +1

litt3 (approved):

relay/server.go
cody-littley: Fair point. Changed to log that doesn't grow with the size of the request:

```
	return api.NewErrorResourceExhausted(fmt.Sprintf("unable to serve data (%d blobs, %d chunks, %d bytes): %v",
		blobCount, chunkCount, requiredBandwidth, originalError))
```

jianoaix: Alternatively, may add a "test_mode" flag (default to false) like the DA Node does, and enable it whenever you want to debug

PR #1159, opened 2025-01-24
Create payload disperser

litt3:

- This PR implements the `PUT` side of the v2 client
- In service of expediency in getting this PayloadDispersal API merged, I've decided to defer writing unit tests until a future PR
    - integration tests will provide better coverage anyway
    - I want to get this up for review ASAP, since @EthenNotEthan is starting with proxy integration, and it would be nice to have basic APIs set
    - I may end up adding some unit tests before this PR merges after all, depending on how long the review process stretches

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/clients/v2/payload_disperser.go
```diff
@@ -0,0 +1,222 @@
...
+					"blob key", blobKey,
+					"previous status", previousStatus,
+					"new status", newStatus)
+				previousStatus = newStatus
+			}
+
+			switch newStatus {
+			case dispgrpc.BlobStatus_CERTIFIED:
+				return blobStatusReply, nil
+			case dispgrpc.BlobStatus_QUEUED, dispgrpc.BlobStatus_ENCODED:
```
bxue-l2, line 215: should we also allow status to be encoded as a part of normal handling, 
https://github.com/Layr-Labs/eigenda/blob/master/api/grpc/disperser/v2/disperser_v2.pb.go#L30

api/clients/codecs/blob_codec.go
```diff
@@ -46,3 +46,26 @@ func GenericDecodeBlob(data []byte) ([]byte, error) {
 
 	return data, nil
 }
+
+// CreateCodec creates a new BlobCodec based on the defined polynomial form of payloads, and the desired
+// BlobEncodingVersion
+func CreateCodec(payloadPolynomialForm PolynomialForm, version BlobEncodingVersion) (BlobCodec, error) {
```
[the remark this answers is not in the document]
samlaf: iiuc codec will get removed completely by his other PR

api/clients/v2/payload_disperser.go
```diff
@@ -0,0 +1,222 @@
...
+	logger logging.Logger,
+	payloadDisperserConfig *PayloadDisperserConfig,
+	disperserClientConfig *DisperserClientConfig,
+	// signer to sign blob dispersal requests
+	signer core.BlobRequestSigner,
+	// prover is used to compute commitments to a new blob during the dispersal process
+	//
+	// IMPORTANT: it is permissible for the prover parameter to be nil, but operating with this configuration
+	// puts a trust assumption on the disperser. With a nil prover, the disperser is responsible for computing
+	// the commitments to a blob, and the PayloadDisperser doesn't have a mechanism to verify these commitments.
```
samlaf, line 40: hmm is this necessary? could we not instead just create a prover implementation that queries the disperser service instead, and force the prover to be non-nil?

api/clients/v2/payload_disperser.go
```diff
@@ -0,0 +1,222 @@
...
+type PayloadDisperser struct {
+	logger          logging.Logger
+	config          *PayloadDisperserConfig
+	codec           codecs.BlobCodec
+	disperserClient DisperserClient
+	certVerifier    verification.ICertVerifier
+}
+
+// BuildPayloadDisperser builds a PayloadDisperser from config structs
+func BuildPayloadDisperser(
```
samlaf, line 30: Prob also want a New constructor with dependency injection so we can test with a fake disperserClient or something?

api/clients/v2/payload_disperser.go
```diff
@@ -0,0 +1,222 @@
...
+) (*PayloadDisperser, error) {
+
+	codec, err := codecs.CreateCodec(
+		payloadDisperserConfig.PayloadPolynomialForm,
+		payloadDisperserConfig.BlobEncodingVersion)
+	if err != nil {
+		return nil, fmt.Errorf("create codec: %w", err)
+	}
+
+	disperserClient, err := NewDisperserClient(disperserClientConfig, signer, prover, accountant)
```
samlaf, line 57: this function feels like a mix between a dependency injection constructor and a builder. We have some configs but also need to inject 3 dependencies (signer, prover, accountant).

Feels like the DisperserClient could use a builder constructor

api/clients/v2/payload_disperser.go
```diff
@@ -0,0 +1,222 @@
...
+	//
+	// IMPORTANT: it is permissible for the prover parameter to be nil, but operating with this configuration
+	// puts a trust assumption on the disperser. With a nil prover, the disperser is responsible for computing
+	// the commitments to a blob, and the PayloadDisperser doesn't have a mechanism to verify these commitments.
+	//
+	// TODO: In the future, an optimized method of commitment verification using fiat shamir transformation will
+	//  be implemented. This feature will allow a PayloadDisperser to offload commitment generation onto the
+	//  disperser, but the disperser's commitments will be verifiable without needing a full-fledged prover
+	prover encoding.Prover,
+	accountant *Accountant,
```
samlaf, line 46: I looked quickly and you can pass in a nil accountant and that works... and its not documented. So we should also document that same behavior here. Maybe same for prover, not sure.

api/clients/v2/payload_disperser.go
```diff
@@ -0,0 +1,222 @@
...
+		codec:           codec,
+		disperserClient: disperserClient,
+		certVerifier:    certVerifier,
+	}, nil
+}
+
+// SendPayload executes the dispersal of a payload, with these steps:
+//
+//  1. Encode payload into a blob
+//  2. Disperse the blob
```
samlaf, line 84: 2. create a blob_header
3. sign the blob_header
4. disperser (blob, blob_header, signature)

api/clients/v2/payload_disperser.go
```diff
@@ -0,0 +1,222 @@
...
+//  3. Poll the disperser with GetBlobStatus until a terminal status is reached, or until the polling timeout is reached
+//  4. Construct an EigenDACert if dispersal is successful
+//  5. Verify the constructed cert with a call to an ethereum contract
```
samlaf, line 87:
```suggestion
//  5. Verify the constructed cert with an eth_call to the EigenDACertVerifier contract
```

api/clients/v2/payload_disperser.go
```diff
@@ -0,0 +1,222 @@
...
+//  6. Return the valid cert
+func (pd *PayloadDisperser) SendPayload(
+	ctx context.Context,
+	// payload is the raw data to be stored on eigenDA
+	payload []byte,
+	// salt is added while constructing the blob header
+	// This salt should be utilized if a blob dispersal fails, in order to retry dispersing the same payload under a
+	// different blob key.
```
samlaf, line 95: only true for reserved bandwidth payments.

api/clients/v2/payload_disperser.go
samlaf: oh I see these things are done by the disperser client? Is that really the separation that we want?

api/clients/v2/payload_disperser.go
```diff
@@ -0,0 +1,222 @@
...
+	}
+
+	return nil
+}
+
+// pollBlobStatus polls the disperser for the status of a blob that has been dispersed
+//
+// This method will only return a non-nil BlobStatusReply if the blob is reported to be CERTIFIED prior to the timeout.
+// In all other cases, this method will return a nil BlobStatusReply, along with an error describing the failure.
+func (pd *PayloadDisperser) pollBlobStatus(
```
samlaf, line 174: how about calling this method `pollBlobStatusUntilCertified` then?

api/clients/v2/payload_disperser.go
```diff
@@ -0,0 +1,222 @@
...
+		case <-ctx.Done():
+			return nil, fmt.Errorf(
+				"timed out waiting for %v blob status, final status was %v: %w",
+				dispgrpc.BlobStatus_CERTIFIED,
+				previousStatus,
+				ctx.Err())
+		case <-ticker.C:
+			// This call to the disperser doesn't have a dedicated timeout configured.
+			// If this call fails to return in a timely fashion, the timeout configured for the poll loop will trigger
+			blobStatusReply, err := pd.disperserClient.GetBlobStatus(ctx, blobKey)
```
samlaf, line 196: don't think this is true. call is blocking afaiu, so no other branches of the select loop can run in parallel. but the context will fail at some point once the context is done which will amount to the same thing

samlaf: also it does have a timeout, its the one configured and passed in to this function as part of ctx

api/clients/v2/payload_disperser.go
```diff
@@ -0,0 +1,222 @@
...
+		return nil, fmt.Errorf("get non signer stake and signature: %w", err)
+	}
+	pd.logger.Debug("Retrieved NonSignerStakesAndSignature", "blobKey", blobKey)
+
+	eigenDACert, err := verification.BuildEigenDACert(
+		blobStatusReply.GetBlobInclusionInfo(),
+		blobStatusReply.GetSignedBatch().GetHeader(),
+		nonSignerStakesAndSignature)
+	if err != nil {
+		return nil, fmt.Errorf("build eigen da cert: %w", err)
```
samlaf, line 140: can we create a function for this

api/clients/v2/payload_retriever_config.go
samlaf, line 1: Feel like all the configs should be in the same config.go file instead? Would be easier to compare them and understand the structure if I could see all 3 side by side.

api/clients/v2/verification/eigenda_cert.go
```diff
@@ -12,3 +17,30 @@ type EigenDACert struct {
 	BatchHeader                 contractEigenDABlobVerifier.BatchHeaderV2
 	NonSignerStakesAndSignature contractEigenDABlobVerifier.NonSignerStakesAndSignature
 }
+
+// BuildEigenDACert creates a new EigenDACert from a BlobInclusionInfo, BatchHeader, and NonSignerStakesAndSignature
+//
+// For convenience, this function accepts arguments as protobufs where applicable, since that's the form the caller
+// will have.
+func BuildEigenDACert(
```
samlaf, line 25: Feel like we should also have a `BuildEigenDACertFromBlobStatusReply` or something like that

api/clients/v2/payload_disperser.go
EthenNotEthan: +1

api/clients/v2/payload_disperser.go
```diff
@@ -0,0 +1,222 @@
...
+	previousStatus := initialStatus
+
+	ticker := time.NewTicker(pd.config.BlobStatusPollInterval)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil, fmt.Errorf(
+				"timed out waiting for %v blob status, final status was %v: %w",
```
EthenNotEthan, line 189: wouldn't the usage of `dispgrpc.BlobStatus_CERTIFIED` return the integer value 3?

api/clients/v2/payload_disperser.go
```diff
@@ -0,0 +1,222 @@
...
+				dispgrpc.BlobStatus_CERTIFIED,
+				previousStatus,
+				ctx.Err())
+		case <-ticker.C:
+			// This call to the disperser doesn't have a dedicated timeout configured.
+			// If this call fails to return in a timely fashion, the timeout configured for the poll loop will trigger
```
EthenNotEthan, line 195: is this actually true?

api/clients/v2/payload_disperser.go
```diff
@@ -0,0 +1,222 @@
...
+			if newStatus != previousStatus {
+				pd.logger.Debug(
+					"Blob status changed",
+					"blob key", blobKey,
+					"previous status", previousStatus,
+					"new status", newStatus)
+				previousStatus = newStatus
+			}
+
+			switch newStatus {
```
EthenNotEthan, line 212: let's add a TODO specifying that we'll need to add more in-depth response status processing to derive failover errors

api/clients/v2/verification/cert_verifier.go
```diff
@@ -104,3 +109,26 @@ func (cv *CertVerifier) VerifyCertV2(
...
+// contract, and returns the resulting NonSignerStakesAndSignature object.
+func (cv *CertVerifier) GetNonSignerStakesAndSignature(
+	ctx context.Context,
+	signedBatch *disperser.SignedBatch,
+) (*verifierBindings.NonSignerStakesAndSignature, error) {
+
+	signedBatchBinding, err := SignedBatchProtoToBinding(signedBatch)
+	if err != nil {
+		return nil, fmt.Errorf("convert signed batch: %s", err)
+	}
```
EthenNotEthan, line 123: shouldn't we be using `%w` instead of `%s` for error -> string formatting?

api/clients/v2/payload_disperser.go
```diff
@@ -0,0 +1,222 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("new disperser client: %s", err)
+	}
+
+	ethClient, err := geth.NewClient(ethConfig, gethcommon.Address{}, 0, logger)
+	if err != nil {
+		return nil, fmt.Errorf("new eth client: %w", err)
+	}
+
+	certVerifier, err := verification.NewCertVerifier(*ethClient, payloadDisperserConfig.EigenDACertVerifierAddr)
```
EthenNotEthan, line 67: should we update this constructor to take in `*ethClient` vs having to dereference to raw object when passing through?

api/clients/v2/payload_client_config.go
```diff
@@ -0,0 +1,43 @@
...
+	// commitment) then it will not be possible to create a commitment opening: the blob will need to be supplied in its
+	// entirety to perform a verification that any part of the data matches the KZG commitment.
+	PayloadPolynomialForm codecs.PolynomialForm
+
+	// The timeout duration for contract calls
+	ContractCallTimeout time.Duration
+
+	BlobVersion v2.BlobVersion
+
+	Quorums []core.QuorumID
```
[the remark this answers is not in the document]
EthenNotEthan: @samlaf wouldn't the quorums used be embedded into the cert?

api/clients/v2/payload_client_config.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/Layr-Labs/eigenda/core"
+	v2 "github.com/Layr-Labs/eigenda/core/v2"
+)
+
+// PayloadClientConfig contains configuration values that are needed by both PayloadRetriever and PayloadDisperser
+type PayloadClientConfig struct {
+	// The blob encoding version to use when writing and reading blobs
+	BlobEncodingVersion codecs.BlobEncodingVersion
+
+	// The Ethereum RPC URL to use for querying the Ethereum blockchain.
```
EthenNotEthan, line 16: knit
```suggestion
	// The Ethereum RPC URL to use for querying an Ethereum network
```

api/clients/v2/payload_disperser.go
```diff
@@ -0,0 +1,222 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("create codec: %w", err)
+	}
+
+	disperserClient, err := NewDisperserClient(disperserClientConfig, signer, prover, accountant)
+	if err != nil {
+		return nil, fmt.Errorf("new disperser client: %s", err)
+	}
+
+	ethClient, err := geth.NewClient(ethConfig, gethcommon.Address{}, 0, logger)
```
EthenNotEthan, line 62: what benefits does using our wrapper client provide vs e.g just using standard `go-ethereum/ethclient`? at first glance this client seems well optimized for write capabilities which we'd never do when verifying. Would there be any value in leveraging the `InstrumentedEthClient` since it provides telemetry wrappers which may be insightful to a customer running proxy? Or perhaps for better future proofing we can leverage the `MultiHomingClient` since it provides redundancy support with multiple RPC providers?

api/clients/codecs/blob_codec.go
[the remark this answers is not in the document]
litt3: Codec won't be removed entirely, but it will no longer be responsible for FFT and IFFT.

Since v2 is on a timeline, I'm just implementing everything using the codec system already in place. When there's time, I plan to fix the existing system, and will update the usage at that point

api/clients/v2/payload_disperser.go
```diff
@@ -0,0 +1,222 @@
...
+	disperserClient DisperserClient
+	certVerifier    verification.ICertVerifier
+}
+
+// BuildPayloadDisperser builds a PayloadDisperser from config structs
+func BuildPayloadDisperser(
+	logger logging.Logger,
+	payloadDisperserConfig *PayloadDisperserConfig,
+	disperserClientConfig *DisperserClientConfig,
+	// signer to sign blob dispersal requests
```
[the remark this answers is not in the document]
litt3: Right. Do you think this needs to be more clear?

api/clients/v2/payload_disperser.go
litt3: I'm not sure what you mean.

Status is permitted to be `ENCODED`, but my understanding is that we won't have a fully populated `BlobStatusReply` until the blob is `CERTIFIED`.

api/clients/v2/payload_disperser.go
litt3: This would be possible, but for now this is just working with how the disperser client is implemented.

When I have a chance, I want to go back and make some improvements like this one

api/clients/v2/payload_disperser.go
```diff
@@ -0,0 +1,222 @@
...
+	"github.com/Layr-Labs/eigenda/api/clients/v2/verification"
+	dispgrpc "github.com/Layr-Labs/eigenda/api/grpc/disperser/v2"
+	"github.com/Layr-Labs/eigenda/common/geth"
+	core "github.com/Layr-Labs/eigenda/core/v2"
+	"github.com/Layr-Labs/eigenda/encoding"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	gethcommon "github.com/ethereum/go-ethereum/common"
+)
+
+// PayloadDisperser provides the ability to disperse payloads to EigenDA
```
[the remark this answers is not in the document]
litt3: [2f4f4a88](https://github.com/Layr-Labs/eigenda/pull/1159/commits/2f4f4a88062c2bf18798ea233432898be7381c46)

api/clients/v2/payload_disperser.go
litt3: [2f4f4a88](https://github.com/Layr-Labs/eigenda/pull/1159/commits/2f4f4a88062c2bf18798ea233432898be7381c46)

api/clients/v2/payload_client_config.go
litt3: [2f4f4a88](https://github.com/Layr-Labs/eigenda/pull/1159/commits/2f4f4a88062c2bf18798ea233432898be7381c46)

api/clients/v2/payload_disperser.go
litt3: Good point, done. [ab2bf058](https://github.com/Layr-Labs/eigenda/pull/1159/commits/ab2bf0584a7307fa167383c057f844d322b56a4f)

api/clients/v2/payload_disperser.go
litt3: Done [a5bc6241](https://github.com/Layr-Labs/eigenda/pull/1159/commits/a5bc6241d0b595081a5efd320a98255f10644910)

api/clients/v2/payload_retriever_config.go
litt3: Done [887e5b92](https://github.com/Layr-Labs/eigenda/pull/1159/commits/887e5b9278af001cb65043721c2208b6ac62a8ed)

api/clients/v2/verification/eigenda_cert.go
litt3: I don't think there is any need to have both, construction is really only going to happen in one place in production. I changed the existing function signature to accept a `BlobStatusReply` instead of the component parts, since I don't have a preference here [32862cf3](https://github.com/Layr-Labs/eigenda/pull/1159/commits/32862cf30144ebc70c582a22385c95408b3d1785)

I'm assuming you didn't mean the create a function `BuildEigenDACertFromBlobStatusReply` that does the fetching of the `NonSignerStakesAndSignature` from the eth client? IMO that would not be a good change.

api/clients/v2/payload_client_config.go
```diff
@@ -0,0 +1,43 @@
...
+	// The chosen form also dictates how the KZG commitment made to the blob can be used. If the encoded payload starts
+	// in PolynomialFormEval (meaning the data WILL be IFFTed before computing the commitment) then it will be possible
+	// to open points on the KZG commitment to prove that the field elements correspond to the commitment. If the
+	// encoded payload starts in PolynomialFormCoeff (meaning the data will NOT be IFFTed before computing the
+	// commitment) then it will not be possible to create a commitment opening: the blob will need to be supplied in its
+	// entirety to perform a verification that any part of the data matches the KZG commitment.
+	PayloadPolynomialForm codecs.PolynomialForm
+
+	// The timeout duration for contract calls
+	ContractCallTimeout time.Duration
```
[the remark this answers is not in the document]
litt3: I think it might be worth adding some guidelines for configuration classes to the style guide.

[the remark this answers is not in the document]
litt3: @samlaf Could you clarify what config architecture you're proposing here? Creating a constructor for the struct? Creating a method which checks and sets defaults?

api/clients/v2/payload_disperser.go
litt3: > Feels like the DisperserClient could use a builder constructor

Agreed. I started implementing one based on this suggestion, but it requires broader changes (for example, creating a `DispserserClient` builder constructor would also entail creating an `AccountantConfig`)

What do you think of removing `BuildPayloadDisperser` and only providing a constructor with dependency injection for now? Doing so would remove the need to consider `Accountant` and `Prover` entirely from this struct, relying instead on that `DisperserClient` passed in.

api/clients/v2/payload_disperser.go
litt3: See my comment [here](https://github.com/Layr-Labs/eigenda/pull/1159#discussion_r1930740976). If we go this route, we can sidestep the issue for now

api/clients/v2/payload_disperser.go
litt3: Clarified [c08d16f8](https://github.com/Layr-Labs/eigenda/pull/1159/commits/c08d16f8e7b62275763a018dd9fc153284303f86)

api/clients/v2/verification/cert_verifier.go
litt3: Yes, thanks for catching [c08d16f8](https://github.com/Layr-Labs/eigenda/pull/1159/commits/c08d16f8e7b62275763a018dd9fc153284303f86)

api/clients/v2/payload_disperser.go
```diff
@@ -0,0 +1,222 @@
...
+	config          *PayloadDisperserConfig
+	codec           codecs.BlobCodec
+	disperserClient DisperserClient
+	certVerifier    verification.ICertVerifier
+}
+
+// BuildPayloadDisperser builds a PayloadDisperser from config structs
+func BuildPayloadDisperser(
+	logger logging.Logger,
+	payloadDisperserConfig *PayloadDisperserConfig,
```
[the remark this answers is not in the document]
litt3: Made configs not pointers for now. [01c81996](https://github.com/Layr-Labs/eigenda/pull/1159/commits/01c819961c2b25a17a85a9fc4e13707a2f2a4e74)

Config defaults etc. will be implemented pending result of [this discussion](https://github.com/Layr-Labs/eigenda/pull/1159#discussion_r1930711606)

api/clients/v2/payload_client_config.go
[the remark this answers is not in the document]
litt3: > otherwise how are the quorums used by the retriever client? are the quorums verified that be included?

Good point, they aren't. I'll the quorums down into the `PayloadDisperser` config.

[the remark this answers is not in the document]
litt3: > Think we always charge for quorums 0 and 1

The existing `DisperserClient` only checks that the quorums aren't empty. Do you know where in the code we are "always charging for quorums 0 and 1"?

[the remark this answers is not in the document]
litt3: I'm unclear on the distinction between `CustomQuorums` and simply `Quorums`.

Whatever is in this field will be directly put into the `BlobHeader`. If that is `CustomQuorums`, what would be referred to as `Quorums`?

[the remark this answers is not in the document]
litt3: Or are you saying we should only require users to put `Quorums` into this config field if the desired numbers != 0 or 1? And always auto populate with 0 and 1 under the hood?

api/clients/v2/payload_disperser.go
litt3: > also it does have a timeout, its the one configured and passed in to this function as part of ctx

This is what the comment is attempting to convey. The spot in the code intentionally doesn't enforce an *additional* timeout for the `GetBlobStatus` call, because the timeout configured and passed in as a parameter functions to break out of the polling loop, AND abort this call to the disperser if necessary.

Does that make sense? If so, do you have any suggestions for how to word this in a way that would be more clear?

api/clients/v2/payload_disperser.go
litt3: There is a timeout set on the context being passed into `pollBlobStatusUntilCertified`.

If that timeout triggers, it will abort this `GetBlobStatus` if necessary, *and* abort the polling loop, right?

api/clients/v2/payload_disperser.go
litt3: I'm not sure I understand what you mean by "derive failover errors".

Do you mean we will need to expand the types of status returned from the top level `SendPayload` method, beyond what's included in the existing `BlobStatus` enum?

api/clients/v2/payload_disperser.go
litt3: Good call out, I changed everything to log the names instead of the enum value [5150c5fe](https://github.com/Layr-Labs/eigenda/pull/1159/commits/5150c5fef18c466a9ce8ab1a82fbb2120e30059d)

api/clients/v2/payload_disperser.go
litt3: Done in commit `70c0b929`, but github is borked and this PR isn't picking up my new commit (stuck on `Checking for ability to merge automatically`).

api/clients/v2/payload_disperser.go
litt3: The wrapper client is need to pass into `NewContractEigenDABlobVerifierCaller`, in the constructor of `CertVerifier` https://github.com/litt3/eigenda/blob/70c0b92985bf53d5b8cef3542c2ee945d598f96e/api/clients/v2/verification/cert_verifier.go#L49

Regarding which flavor of eth client to use: I'm leaning toward just accepting an `EthClient` as a parameter, instead of constructing one here. That way the caller can just choose whichever type is best suited. Thoughts?

api/clients/v2/payload_disperser.go
```diff
@@ -0,0 +1,222 @@
...
+	// IMPORTANT: it is permissible for the prover parameter to be nil, but operating with this configuration
+	// puts a trust assumption on the disperser. With a nil prover, the disperser is responsible for computing
+	// the commitments to a blob, and the PayloadDisperser doesn't have a mechanism to verify these commitments.
+	//
+	// TODO: In the future, an optimized method of commitment verification using fiat shamir transformation will
+	//  be implemented. This feature will allow a PayloadDisperser to offload commitment generation onto the
+	//  disperser, but the disperser's commitments will be verifiable without needing a full-fledged prover
+	prover encoding.Prover,
+	accountant *Accountant,
+	ethConfig geth.EthClientConfig,
```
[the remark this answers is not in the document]
litt3: See my [comment below](https://github.com/Layr-Labs/eigenda/pull/1159#discussion_r1930946466), I'm thinking we should just accept a client directly instead of constructing one here

api/clients/v2/payload_client_config.go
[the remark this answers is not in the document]
litt3: Added doc for `BlobVersion`, and moved quorums into `PayloadDisperser` in commit [79026620](https://github.com/Layr-Labs/eigenda/pull/1159/commits/790266209257c5636548005d4bffa2dcee2fe549)

Awaiting resolution of discussions about Quorum before writing the doc

api/clients/v2/payload_disperser.go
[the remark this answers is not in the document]
bxue-l2: this is fine

api/clients/v2/payload_disperser.go
bxue-l2: So when we received a `ENCODED` temporarily, the newStatus will generate BlobStatus_ENCODED. Since there isn't an arm to handle that, it will go to default, and returning an error. Or do I miss somewhere?

litt3: Notice the second case in the switch is multi-case, we will `continue` if the status is either `QUEUED` or `ENCODED`

```
case dispgrpc.BlobStatus_QUEUED, dispgrpc.BlobStatus_ENCODED:
```

bxue-l2: ahh...

api/clients/v2/payload_client_config.go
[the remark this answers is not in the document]
samlaf: As we discussed just now offline, we can write a descrption for the quorums like: "Quorums that will be injected into the blobHeader. These quorums must meet the Thresholds in order for a cert to be valid." And you can add a TODO regarding the discussion of whether we should just move back to an interface like V1 where 0 and 1 were mandatory quorums always added by default.

api/clients/v2/payload_client_config.go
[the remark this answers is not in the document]
samlaf: Added TODO for me to add this to style guide: https://github.com/Layr-Labs/eigenda/pull/1137#issuecomment-2616865924

In terms of my personal preference, I like:
- no constructor for config structs (that's very much against the point of having extendability, which requires using go's default zero value initialization)
- having the constructor validate the config and inject default values over zero values (note: this requires care that zero values are always meaningless, which is not always the case!)
- validation + injecting defualt values can be part of a separate function if it's a large config struct that would overwhelm the rest of constructor logic

api/clients/v2/payload_disperser.go
samlaf: Think that would make sense. I would need to play with the code to myself get a feeling for what the best way to make this work would be, so I'll defer to your sense of aesthetics here :)

api/clients/v2/payload_disperser.go
samlaf: I see. That's fine with me for now. sorry for bike-shedding.

api/clients/v2/payload_client_config.go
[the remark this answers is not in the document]
litt3: As we discussed synchronously, I've added a doc for `Quorums`, along with a TODO to get further clarity. I'll follow up in slack to resolve this TODO

[0c366248](https://github.com/Layr-Labs/eigenda/pull/1159/commits/0c3662487afe3238d2ba91a25a0da67f8ad5911b)

api/clients/v2/verification/eigenda_cert.go
samlaf: Don't remember 100% but think that IS what I meant lol. But didn't see that we already have the payloadDisperser.buildEigenDACert function that does this.

samlaf (approved):
The changes that I had requested have been made, so approving.

api/clients/v2/payload_client_config.go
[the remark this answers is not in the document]
litt3: See my [comment](https://github.com/Layr-Labs/eigenda/pull/1159#discussion_r1931186947) above. I've added a TODO for now and will seek clarity

EthenNotEthan (approved):
LGTM

bxue-l2 (approved):

api/clients/v2/payload_disperser.go
```diff
@@ -0,0 +1,239 @@
...
+) (*verification.EigenDACert, error) {
+
+	timeoutCtx, cancel := context.WithTimeout(ctx, pd.config.ContractCallTimeout)
+	defer cancel()
+	nonSignerStakesAndSignature, err := pd.certVerifier.GetNonSignerStakesAndSignature(
+		timeoutCtx, blobStatusReply.GetSignedBatch())
+	if err != nil {
+		return nil, fmt.Errorf("get non signer stake and signature: %w", err)
+	}
+	pd.logger.Debug("Retrieved NonSignerStakesAndSignature", "blobKey", blobKey)
```
bxue-l2, line 230: add hex.EncodeToString or something similar to print hex string, as opposed to bytes. Very useful for debugging
Screenshot 2025-01-27 at 7 38 14 PM
Screenshot 2025-01-27 at 7 38 25 PM

litt3: Good point, done [a311f951](https://github.com/Layr-Labs/eigenda/pull/1159/commits/a311f9517b5c6fbc594ea45914e2017084669e57)

api/clients/v2/payload_client_config.go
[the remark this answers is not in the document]
litt3: > no constructor for config structs (that's very much against the point of having extendability, which requires using go's default zero value initialization)

Could you explain what you mean by this? What's the problem with having a constructor for `ConfigA`, and if `ConfigA` extends `ConfigB`, call the constructor for `ConfigB` in the constructor of `ConfigA`?

api/clients/v2/payload_disperser.go
litt3: Simplified for now [f4a6f804](https://github.com/Layr-Labs/eigenda/pull/1159/commits/f4a6f804daec74338cd6d9e242270c4fc2c185f1)

api/clients/v2/payload_disperser.go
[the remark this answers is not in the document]
litt3: I believe with the changes in [f4a6f804](https://github.com/Layr-Labs/eigenda/pull/1159/commits/f4a6f804daec74338cd6d9e242270c4fc2c185f1) this isn't relevant anymore.

api/clients/v2/payload_disperser.go
litt3: Hmm. Is it not true that _any_ error in dispersal would require resorting to failover? If not, what class of error would you expect to *not* result in failover, and what would be the strategy to handle it? Retries? (trying to understand likely outcomes, to be able to write a helpful TODO)

api/clients/v2/payload_client_config.go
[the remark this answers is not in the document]
litt3: I created some config utilities in [959e4564](https://github.com/Layr-Labs/eigenda/pull/1159/commits/959e4564cfb4705fa38d4f88e9e753a3d23bbecf). LMK what you think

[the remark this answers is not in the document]
samlaf: LGTM! Having a default constructor is totally fine, and is actually a good idea! What I meant is that users shouldn't be constructing the config by using a constructor of the type
```
func NewConfig(fieldA, fieldB, fieldC, fieldD,...) Config {
   return Config{
      fieldA,
      fieldB,
      ...
      }
}
```
Because this is dumb

api/clients/v2/payload_disperser.go
litt3: Added the TODO

PR #1164, opened 2025-01-26
[ejector] Don't create txn when there's no one to eject

ian-shim:
It should exit early when there is no one to eject

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/operators/ejector/ejector.go
+++ b/operators/ejector/ejector.go
@@ -24,6 +24,9 @@ const (
 	queryTickerDuration     = 3 * time.Second
 )
 
+// EjectionResponse encapsulates the response of an ejection request.
+// It contains the transaction hash of the ejection transaction.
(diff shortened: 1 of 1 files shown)


gpsanant (approved):
lgtm

operators/ejector/ejector.go
```diff
@@ -116,6 +116,14 @@ func (e *Ejector) Eject(ctx context.Context, nonsignerMetrics []*NonSignerMetric
 		}
 	}
 
+	if len(nonsigners) == 0 {
+		e.logger.Info("No operators to eject")
+		e.metrics.IncrementEjectionRequest(mode, codes.OK)
+		return &EjectionResponse{
+			TransactionHash: "",
```
jianoaix, line 123: can you add a comment to EjectionResponse that empty txn hash (when Eject has no error) means no nonsigners to eject?

jianoaix (approved):

PR #1167, opened 2025-01-27
Functional tests

cody-littley:

tests that run against preprod


common/logger_config.go
```diff
@@ -64,6 +64,20 @@ func DefaultLoggerConfig() LoggerConfig {
 	}
 }
 
+// DefaultTextLoggerConfig returns a LoggerConfig with the default settings for a text logger.
+// For use in tests or other scenarios where the logs are consumed by humans.
+func DefaultTextLoggerConfig() LoggerConfig {
```
dmanc, line 69: Could we add maybe `DefaultConsoleLoggerConfig". Only difference is NoColor = false. Then in cases where we run stuff locally we can use that, otherwise CI defaults to TextLogger?

dmanc (approved):

common/logger_config.go
cody-littley: added

PR #1172, opened 2025-01-28
Upgrade icicle library to v3.4.0

dmanc:

Upgrades icicle library to latests version. Didn't notice much improvement on GPU side but the CPU icicle backend made some improvements. Specifically seems their FFT is faster than the one we're using.

CPU comparison:
image
image
image

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


jianoaix:
So the existing Gnark is significantly faster than icicle on CPUs overall?

jianoaix (approved):

dmanc:
> So the existing Gnark is significantly faster than icicle on CPUs overall?

Only for MSM, the FFT is much faster on icicle 3.4.0. But overall there is still a gap due to the MSM performance but it's closing in.

PR #1174, opened 2025-01-28
Allow relay protos to be larger.

cody-littley:

Allow relay protobufs to be larger.

### DIFF
--- a/api/clients/v2/disperser_client.go
+++ b/api/clients/v2/disperser_client.go
@@ -3,6 +3,7 @@ package clients
 import (
 	"context"
 	"fmt"
+	"github.com/docker/go-units"
 	"sync"
 
 	"github.com/Layr-Labs/eigenda/api"
@@ -314,7 +315,7 @@ func (c *disperserClient) initOnceGrpcConnection() error {
 	var initErr error
 	c.initOnceGrpc.Do(func() {
 		addr := fmt.Sprintf("%v:%v", c.config.Hostname, c.config.Port)
-		dialOptions := getGrpcDialOptions(c.config.UseSecureGrpcFlag)
+		dialOptions := getGrpcDialOptions(c.config.UseSecureGrpcFlag, 4*units.MiB)
 		conn, err := grpc.NewClient(addr, dialOptions...)
 		if err != nil {
 			initErr = err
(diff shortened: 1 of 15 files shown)


api/clients/v2/disperser_client.go
```diff
@@ -314,7 +315,7 @@ func (c *disperserClient) initOnceGrpcConnection() error {
 	var initErr error
 	c.initOnceGrpc.Do(func() {
 		addr := fmt.Sprintf("%v:%v", c.config.Hostname, c.config.Port)
-		dialOptions := getGrpcDialOptions(c.config.UseSecureGrpcFlag)
+		dialOptions := getGrpcDialOptions(c.config.UseSecureGrpcFlag, 4*units.MiB)
```
ian-shim, line 318: isn't this already the default?

cody-littley: Yes, 4mb is the default. I added an argument to `getGrpcDialOptions()`, and so in this non-relay code I just preserved the prior behavior.

node/flags/flags.go
```diff
@@ -278,6 +279,13 @@ var (
 		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "DISPERSAL_AUTHENTICATION_TIMEOUT"),
 		Value:    time.Minute,
 	}
+	RelayMaxGRPCMessageSizeFlag = cli.IntFlag{
```
ian-shim, line 282: needs to be added to `optionalFlags` below

relay/cmd/flags/flags.go
```diff
@@ -66,7 +67,7 @@ var (
 		Usage:    "The size of the blob cache, in bytes.",
 		Required: false,
 		EnvVar:   common.PrefixEnvVar(envVarPrefix, "BLOB_CACHE_SIZE"),
-		Value:    1024 * 1024 * 1024,
+		Value:    units.GiB,
```
ian-shim, line 70: not sure what we're provisioning in prod, but these should be a lot bigger, no?

node/flags/flags.go
cody-littley: oops, added

relay/cmd/flags/flags.go
cody-littley: I set this to `1gb` when working on OOM problems. Now that the cache can correctly gauge its size, I'm ok changing it back higher again. It is now `8 * units.GiB` in this PR.

ian-shim (approved):

PR #1175, opened 2025-01-28
feat(api): relay registry caller

ethenotethan:
Relay information must be ingested via EigenDA proxy. Either the user can provide this manually or just have the proxy query the on-chain `EigenDARelayRegistry` during initialization to bootstrap `relayClient` config state. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/relay_registry_client.go
+++ b/api/clients/v2/relay_registry_client.go
@@ -1,0 +1,65 @@
+package clients
+
+import (
+	"fmt"
+
+	"github.com/Layr-Labs/eigenda/common/geth"
+	registry_binding "github.com/Layr-Labs/eigenda/contracts/bindings/EigenDARelayRegistry"
+	corev2 "github.com/Layr-Labs/eigenda/core/v2"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// Provides accessor methods for read interfacing with
+// relay registry contract (core/EigenDARelayRegistry.sol)
+type RelayRegistryCaller interface {
+	// Reads the current sockets or key->url mapping
+	// from the on-chain registry
+	GetSockets() (map[corev2.RelayKey]string, error)
+}
+
+// NewRelayRegistryClient constructs a RelayRegistryClient
+func NewRelayRegistryClient(
+	ethClient geth.EthClient,
+	relayRegistryAddr string,
+) (RelayRegistryClient, error) {
+
+	registryCaller, err := registry_binding.NewContractEigenDARelayRegistryCaller(
+		common.HexToAddress(relayRegistryAddr),
+		ethClient)
+
+	if err != nil {
+		return nil, fmt.Errorf("bind to relay registry contract at %s: %w", relayRegistryAddr, err)
+	}
+
+	return &relayRegistryClient{
+		caller: registryCaller,
+	}, nil
+}
+
+// TODO: Mitigate the risk of contract mutability where the on-chain registry
+// can be subject to new assertions or relay mapping changes.
+type relayRegistryClient struct {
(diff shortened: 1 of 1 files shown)


api/clients/v2/relay_registry_caller.go
```diff
@@ -0,0 +1,65 @@
...
+	corev2 "github.com/Layr-Labs/eigenda/core/v2"
+	"github.com/ethereum/go-ethereum/common"
+)
+
+// Provides accessor methods for read interfacing with
+// relay registry contract (core/EigenDARelayRegistry.sol)
+type RelayRegistryCaller interface {
+	// Reads the current sockets or key->url mapping
+	// from the on-chain registry
+	GetSockets() (map[corev2.RelayKey]string, error)
```
litt3, line 17: I don't think this should be called "GetSockets". I know that's the terminology used in the `relay_client`, but these are URLs, not sockets. Possible I'm wrong, but I've never heard a URL be called a "socket"

ian-shim:
Do we need a standalone client to do this? We probably shouldn't have a standalone client for every contract call.
We already have [this method](https://github.com/Layr-Labs/eigenda/blob/master/core/eth/reader.go#L956) in chain reader. Can we use that instead?

api/clients/v2/relay_registry_caller.go
```diff
@@ -0,0 +1,65 @@
...
+	}
+
+	return &relayRegistryCaller{
+		caller: registryCaller,
+	}, nil
+}
+
+func (rrc *relayRegistryCaller) GetSockets() (map[corev2.RelayKey]string, error) {
+	// read the # of relays by processing next key position
+	key, err := rrc.caller.NextRelayKey(nil)
```
litt3, line 47: nit: consider renaming `key` to `keyCount`, to make it more clear what it is

api/clients/v2/relay_registry_caller.go
samlaf: They use Socket to mean URL:port combination. Not sure whether relays register a full socket address or just a URL where default 443 port or something is used though.

litt3: I would call that an "endpoint". Socket really just seems like the wrong term

samlaf: https://www.ibm.com/docs/en/zvse/6.2?topic=considerations-socket-addresses

litt3: `Socket` != `SocketAddress`. *Maybe* you could argue this map contains socket addresses, but certainly not sockets

`SocketAddress` does indeed contain an address and a port, but it seems to me like the term is used internally on a machine, to reference the literal sockets that have been created locally.

cody-littley: `socket` generally has a technical meaning in networking, and an `IP_ADDRESS:PORT` string isn't a socket (although it does tell you how to open a socket). I think we should be careful about how we use the term `socket`.

samlaf: I also didn't like the term originally, and did a bunch of googling lol. Seems like its an accepted term? https://en.wikipedia.org/wiki/Network_socket#Socket_addresses

I personally find endpoint to not be precise enough, but agree that `socket` by itself is super confusing and nor is it anymore precise.

litt3: It's *an* accepted term, but I don't think in this context. From the linked article:

```
This combination is often known as a socket address. It is the network-facing access handle to the network socket.
```

It's used in a context specifically related to interactions with network sockets.

PR #1176, opened 2025-01-28
Update v2 swagger

pschork:
Update swagger
Fix batch feed group

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/dataapi/v2/server_v2.go
+++ b/disperser/dataapi/v2/server_v2.go
@@ -583,7 +583,7 @@ func (s *ServerV2) FetchBlobInclusionInfoHandler(c *gin.Context) {
 // FetchBatchFeedHandler godoc
 //
 //	@Summary	Fetch batch feed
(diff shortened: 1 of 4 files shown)


jianoaix (approved):

disperser/dataapi/v2/server_v2.go
```diff
@@ -583,7 +583,7 @@ func (s *ServerV2) FetchBlobInclusionInfoHandler(c *gin.Context) {
 // FetchBatchFeedHandler godoc
 //
 //	@Summary	Fetch batch feed
-//	@Tags		Blob
+//	@Tags		Batch
```
jianoaix, line 586: ah, this is the reason of 404?

jianoaix: actually it should just be the grouping of API paths

PR #1178, opened 2025-01-28
docs: proto on paymentHeader

hopeyen:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proto/common/v2/common_v2.proto
```diff
@@ -71,8 +71,14 @@ message PaymentHeader {
...
+  // The reservation period of the dispersal request. The dispersal client's account will set this value to 
+  // current timestamp divided by the on-chain configured reservation period interval. The disperser server validates
+  // this value to be the period they receive this request, or from the previous period.
   uint32 reservation_period = 2;
-  // The cumulative payment of the dispersal request. This field will be parsed as a big integer.
+  // The cumulative payment of the dispersal request. This field will be parsed as a big integer, and must represent
+  // the total amount of tokens paid over all previous requests. The disperser server validates this value against the 
+  // total on-chain deposit of the user's account, and the total amount of tokens paid over all previous requests. This 
+  // is not incremental but cumulative so the requests can arrive out of order but still has clear declaration of the 
+  // payment being used for the current request.
```
samlaf, line 82: not super clear what this means. adding a simple example would help.

api/proto/common/v2/common_v2.proto
```diff
@@ -71,8 +71,14 @@ message PaymentHeader {
   // The account ID of the disperser client. This account ID is an eth wallet address of the user,
   // corresponding to the key used by the client to sign the BlobHeader.
   string account_id = 1;
-  // The reservation period of the dispersal request. 
+  // The reservation period of the dispersal request. The dispersal client's account will set this value to 
+  // current timestamp divided by the on-chain configured reservation period interval. The disperser server validates
+  // this value to be the period they receive this request, or from the previous period.
   uint32 reservation_period = 2;
-  // The cumulative payment of the dispersal request. This field will be parsed as a big integer.
+  // The cumulative payment of the dispersal request. This field will be parsed as a big integer, and must represent
```
samlaf, line 78: Good to add a note that these are mutually exclusive options: only one of them should be set.

Or are they? Are there cases where both should be set in case reservation_period's bandwidth has been used and one wants to use the cumulative_payment?

samlaf: Either case more explanation as to how they work together will be helpful

api/proto/common/v2/common_v2.proto
```diff
@@ -71,8 +71,14 @@ message PaymentHeader {
   // The account ID of the disperser client. This account ID is an eth wallet address of the user,
   // corresponding to the key used by the client to sign the BlobHeader.
   string account_id = 1;
-  // The reservation period of the dispersal request. 
+  // The reservation period of the dispersal request. The dispersal client's account will set this value to 
+  // current timestamp divided by the on-chain configured reservation period interval. The disperser server validates
+  // this value to be the period they receive this request, or from the previous period.
```
samlaf, line 76: also need to mention that one must have already paid for reserved bandwidth in order to make use of this, etc

api/proto/common/v2/common_v2.proto
hopeyen: I wrote about this on [line](https://github.com/Layr-Labs/eigenda/pull/1178/files/c4f63c5287588f7e585121466b80be5dc8874ac2#diff-fee56aef46840667a14ec961b695984ae7e2f422d5ac74baab0759546259978bR65), lmk if it makes sense at all

api/proto/common/v2/common_v2.proto
hopeyen: added some explanations

api/proto/common/v2/common_v2.proto
hopeyen: added more explanations and an example

samlaf (approved):

api/proto/common/v2/common_v2.proto
samlaf: thanks, lgtm

PR #1181, opened 2025-01-29
V2 Retrieval Isolation

supriya-premkumar:

- Updates operator socket format to `host:v1DispersalPort;v1RetrievalPort;v2DispersalPort;V2RetrievalPort`
- Refactors parseOperatorSocket to handle v2 retrieval socket with strict port validation
- Adds ValidatePort() to core utils  
- Adds stricter validation for operator sockets. Valid formats:
    - `host:v1DispersalPort;v1RetrievalPort`
    - `host:v1DispersalPort;v1RetrievalPort;v2DispersalPort;V2RetrievalPort`
- Adds host validation supporting both FQDN and IP addr
- Registers node for the new V2 retrieval service

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Integration tests

### DIFF
--- a/core/state.go
+++ b/core/state.go
@@ -6,6 +6,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"math/big"
+	"net"
 	"slices"
 	"strings"
 )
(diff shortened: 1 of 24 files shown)


node/config.go
```diff
@@ -252,19 +253,33 @@ func NewConfig(ctx *cli.Context) (*Config, error) {
 			return nil, fmt.Errorf("v2 dispersal port (NODE_V2_DISPERSAL_PORT) must be specified if v2 is enabled")
 		}
 	} else {
-		_, err = strconv.Atoi(v2DispersalPort)
+		// TODO: should we mandate v2Enabled == true
```
supriya-premkumar, line 256: corner case: I think this and line:269 needs to check that `v2Enabled==true` This will avoid user errors of adding a v2 port in the flag without enabling v2.

cc: @ian-shim holding off on making this change to confirm intent.

ian-shim: I think that makes sense. Although we would need to update this logic when we update the flags that specify the `mode`

node/grpc/run.go
```diff
@@ -109,5 +109,34 @@ func RunServers(serverV1 *Server, serverV2 *ServerV2, config *node.Config, logge
 		}
 	}()
 
+	go func() {
```
ian-shim, line 112: we should remove serverV2 from L102 and update the server options in L95

node/config.go
supriya-premkumar: Agreed. Will add the check, that way we have a reference for all the valid v2Enable checks before refactoring.

core/state.go
```diff
@@ -19,48 +20,73 @@ func (s OperatorSocket) String() string {
...
 		return OperatorSocket(fmt.Sprintf("%s:%s;%s", nodeIP, dispersalPort, retrievalPort))
 	}
-	return OperatorSocket(fmt.Sprintf("%s:%s;%s;%s", nodeIP, dispersalPort, retrievalPort, v2DispersalPort))
+	return OperatorSocket(fmt.Sprintf("%s:%s;%s;%s;%s", nodeIP, dispersalPort, retrievalPort, v2DispersalPort, v2RetrievalPort))
 }
 
 type StakeAmount = *big.Int
 
-func ParseOperatorSocket(socket string) (host string, dispersalPort string, retrievalPort string, v2DispersalPort string, err error) {
+func ParseOperatorSocket(socket string) (host, v1DispersalPort, v1RetrievalPort, v2DispersalPort, v2RetrievalPort string, err error) {
```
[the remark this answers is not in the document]
supriya-premkumar: Totally agreed.

node/grpc/run.go
supriya-premkumar: Updating it.

core/state.go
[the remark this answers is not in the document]
supriya-premkumar: Let me make subsequent PR for this one.

node/config.go
```diff
@@ -252,19 +253,38 @@ func NewConfig(ctx *cli.Context) (*Config, error) {
 			return nil, fmt.Errorf("v2 dispersal port (NODE_V2_DISPERSAL_PORT) must be specified if v2 is enabled")
 		}
 	} else {
-		_, err = strconv.Atoi(v2DispersalPort)
+		if !v2Enabled {
+			return nil, fmt.Errorf("enable v2 flag needs to be set when v2 dispersal port (NODE_V2_DISPERSAL_PORT) is specified")
+		}
+		err = core.ValidatePort(v2DispersalPort)
```
ian-shim, line 259: should we also validate that `v2DispersalPort != v1DispersalPort` (and the same for retrieval port)?

ian-shim (approved):
lgtm

node/config.go
supriya-premkumar: Good point. Updating it.

PR #1183, opened 2025-01-29
fix(api/clients): Update v1 client to take in eigen-sdk logger

ethenotethan:
V2 clients take in the `eigensdk/logger` while v1 uses `go-ethereum/log`. We'll need to consolidate a single one within proxy or update the SDK's log interface to be interoperable with `go-ethereum/log`.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/eigenda_client_e2e_test.go
+++ b/api/clients/eigenda_client_e2e_test.go
@@ -4,12 +4,11 @@ import (
 	"context"
 	"flag"
 	"math/big"
-	"os"
 	"testing"
 	"time"
 
+	"github.com/Layr-Labs/eigensdk-go/logging"
 	"github.com/ethereum/go-ethereum/accounts/abi/bind"
-	"github.com/ethereum/go-ethereum/log"
 	"github.com/stretchr/testify/assert"
 )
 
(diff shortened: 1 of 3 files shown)


litt3 (approved):

api/clients/eigenda_client_e2e_test.go
```diff
@@ -29,7 +28,9 @@ func TestClientUsingTestnet(t *testing.T) {
 
 	t.Run("PutBlobWaitForConfirmationDepth0AndGetBlob", func(t *testing.T) {
 		t.Parallel()
-		logger := log.NewLogger(log.NewTerminalHandler(os.Stdout, true))
+		logger, err := logging.NewZapLogger(logging.Development)
```
samlaf, line 31: use slogger, not zaplogger (zaplogger should be deprecated! is it not @shrimalmadhur ?)

samlaf (commented):
We should use slog everywhere,as that is the golang standard since go1.21 https://go.dev/blog/slog

api/clients/eigenda_client_e2e_test.go
ethenotethan: there is slogger support in sdk: https://github.com/Layr-Labs/eigensdk-go/blob/dev/logging/slog_logger.go 

plugged zap through tests since it was easy to construct. On second glance though there is a `test_utils` function that generates a compatible logger specifically for testing. Using this instead

ethenotethan: [80bae61](https://github.com/Layr-Labs/eigenda/pull/1183/commits/80bae61cf5381bfc00edf173c8272ed3ad24b8b9)

samlaf: Nice thanks. I know there's slogger support in sdk, I wrote it :P
is the logging.NewSLogger not easy to use?
Didn't even notice this was just for tests. Agree its less important, but still better to be consistent and use slogger everywhere. Thanks for teh change

samlaf (approved):

PR #1184, opened 2025-01-29
swap g2 A0, A1 for X and Y

bxue-l2:
This PR addresses the inconsistency issue when serializing the blobHeader at the controller side and at the client side. 

When the controller computes a merkle tree for a batch, it serialize all Blobheader into leaves in the tree, and generate merkle proof for each blob. The merkle proof is given to the client, who computes the leaf locally and query the smart contract for the merkle verification.

Because two serialization are different, the merkle proof won't verify.

Note, the G2 serialization only affect G2 inside the blobCommitment. Not the one inside attestation.ApkG2 . The serialization in attestation.[ApkG2](https://github.com/bxue-l2/eigenda/blob/ba85f0bfb2c64644722fa47a6d637ed85c5e6035/inabox/tests/integration_v2_test.go#L192) is A1, A0. 

The BlobCommitments is not used for computation at all. So it is purely just standardizing the serialization format for G2. As it will be weird that we do both serialization in the codebase.

In the current integration test, it called VerifyDACertV2FromSignedBatch. 
But in the client, we used an upgraded version called VerifyDACertV2.

### DIFF
--- a/core/v2/serialization.go
+++ b/core/v2/serialization.go
@@ -132,22 +132,22 @@ func ComputeBlobKey(
 			},
 			LengthCommitment: abiG2Commit{
 				X: [2]*big.Int{
-					blobCommitments.LengthCommitment.X.A0.BigInt(new(big.Int)),
 					blobCommitments.LengthCommitment.X.A1.BigInt(new(big.Int)),
+					blobCommitments.LengthCommitment.X.A0.BigInt(new(big.Int)),
 				},
 				Y: [2]*big.Int{
-					blobCommitments.LengthCommitment.Y.A0.BigInt(new(big.Int)),
 					blobCommitments.LengthCommitment.Y.A1.BigInt(new(big.Int)),
+					blobCommitments.LengthCommitment.Y.A0.BigInt(new(big.Int)),
 				},
 			},
 			LengthProof: abiG2Commit{
 				X: [2]*big.Int{
-					blobCommitments.LengthProof.X.A0.BigInt(new(big.Int)),
 					blobCommitments.LengthProof.X.A1.BigInt(new(big.Int)),
+					blobCommitments.LengthProof.X.A0.BigInt(new(big.Int)),
 				},
 				Y: [2]*big.Int{
-					blobCommitments.LengthProof.Y.A0.BigInt(new(big.Int)),
 					blobCommitments.LengthProof.Y.A1.BigInt(new(big.Int)),
+					blobCommitments.LengthProof.Y.A0.BigInt(new(big.Int)),
 				},
 			},
 			DataLength: uint32(blobCommitments.Length),
--- a/core/v2/serialization.go
+++ b/core/v2/serialization.go
@@ -130,6 +130,10 @@ func ComputeBlobKey(
 				X: blobCommitments.Commitment.X.BigInt(new(big.Int)),
 				Y: blobCommitments.Commitment.Y.BigInt(new(big.Int)),
 			},
+			// Most crypptography library serializes a G2 point by having
+			// A0 followed by A1 for both X, Y field of G2. However, ethereum
+			// precompile assumes an ordering of A1, A0. Currently, we choose
+			// to conform with Ethereum order when serializing a blobHeaderV2
 			LengthCommitment: abiG2Commit{
 				X: [2]*big.Int{
 					blobCommitments.LengthCommitment.X.A1.BigInt(new(big.Int)),
@@ -140,6 +144,7 @@ func ComputeBlobKey(
 					blobCommitments.LengthCommitment.Y.A0.BigInt(new(big.Int)),
 				},
 			},
+			// Same as above
 			LengthProof: abiG2Commit{
 				X: [2]*big.Int{
 					blobCommitments.LengthProof.X.A1.BigInt(new(big.Int)),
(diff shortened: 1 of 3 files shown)


core/v2/serialization_test.go
```diff
@@ -110,8 +111,10 @@ func TestBlobCertHash(t *testing.T) {
 
 	hash, err := blobCert.Hash()
 	assert.NoError(t, err)
-	// 0x1db857aeead06422b8d727dc3972db6ceb04ceb87e194cb6c2389ac3015eda49 verified in solidity
-	assert.Equal(t, "1db857aeead06422b8d727dc3972db6ceb04ceb87e194cb6c2389ac3015eda49", hex.EncodeToString(hash[:]))
+
+	fmt.Printf("%x", hash)
+	// afa39b4c45197f0254f7e8e2127c797c74578357e9f077eab7a8aa62e1402bec has not verified in solidity
+	assert.Equal(t, "afa39b4c45197f0254f7e8e2127c797c74578357e9f077eab7a8aa62e1402bec", hex.EncodeToString(hash[:]))
```
samlaf, line 117: does this mean you computed the hash onchain by calling the smart contract by hand and then posted the code here? Shouldn't we just have an integration test instead that hashes offchain and onchain and makes sure they are equal?

core/v2/serialization.go
```diff
@@ -114,24 +114,29 @@ func (b *BlobHeader) BlobKey() (BlobKey, error) {
 				X: b.BlobCommitments.Commitment.X.BigInt(new(big.Int)),
 				Y: b.BlobCommitments.Commitment.Y.BigInt(new(big.Int)),
 			},
+			// Most crypptography library serializes a G2 point by having
```
litt3, line 117: "crypptography" -> "cryptography"

litt3 (approved):

core/v2/serialization_test.go
```diff
@@ -110,8 +111,10 @@ func TestBlobCertHash(t *testing.T) {
 
 	hash, err := blobCert.Hash()
 	assert.NoError(t, err)
-	// 0x1db857aeead06422b8d727dc3972db6ceb04ceb87e194cb6c2389ac3015eda49 verified in solidity
-	assert.Equal(t, "1db857aeead06422b8d727dc3972db6ceb04ceb87e194cb6c2389ac3015eda49", hex.EncodeToString(hash[:]))
+
+	fmt.Printf("%x", hash)
```
ian-shim, line 115: remove?

core/v2/serialization_test.go
ian-shim: These are done in chisel with a script that constructs the same struct. @bxue-l2 can you update this when you've done the validation?

ian-shim (approved):
lgtm, let's update `core/v2/serialization_test.go` when you've done the validation in solidity

core/v2/serialization.go
```diff
@@ -114,24 +114,29 @@ func (b *BlobHeader) BlobKey() (BlobKey, error) {
...
+			// A0 followed by A1 for both X, Y field of G2. However, ethereum
+			// precompile assumes an ordering of A1, A0. Currently, we choose
```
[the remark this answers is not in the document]
bxue-l2: will add later

core/v2/serialization_test.go
```diff
@@ -52,8 +53,8 @@ func TestBlobKeyFromHeader(t *testing.T) {
 	}
 	blobKey, err := bh.BlobKey()
 	assert.NoError(t, err)
-	// 0xac17ec01189d04e827ea282f49077b138f0aefa5177be146a74ae51135868c57 verified in solidity
-	assert.Equal(t, "ac17ec01189d04e827ea282f49077b138f0aefa5177be146a74ae51135868c57", blobKey.Hex())
+	// 0x2bac85c7fc4c21ad02538a7eb44b120efbc64d25b1691470273f84c8cf82187a has not verified in solidity
```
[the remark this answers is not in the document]
bxue-l2: will do

core/v2/serialization.go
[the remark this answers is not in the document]
bxue-l2: https://github.com/Layr-Labs/eigenda/pull/1184/commits/4a04c81f62a254752ffef05ea2d9f63935af8d42#diff-8760c5ade10e05a5071f897826c7b8540d7283b985b33680cb51dca6510ec0a7R137-R138

core/v2/serialization_test.go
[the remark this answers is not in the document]
bxue-l2: https://github.com/Layr-Labs/eigenda/pull/1184/commits/4a04c81f62a254752ffef05ea2d9f63935af8d42#diff-5df0a845790de8f5a44fa01bb1ef6c996b0e82800a8a382ecf7986faf59f5830R55

core/v2/serialization_test.go
bxue-l2: https://github.com/Layr-Labs/eigenda/pull/1184/commits/4a04c81f62a254752ffef05ea2d9f63935af8d42#diff-5df0a845790de8f5a44fa01bb1ef6c996b0e82800a8a382ecf7986faf59f5830R55

jianoaix (approved):
kudos to the pr description and code documentation, making the context clear

core/v2/serialization_test.go
samlaf: Still unclear as to how this is computed? How could I reproduce using chisel? Bowen did you really use chisel? Do you have to like load the contract and then call the function? Still feel like an automated test would be way better... do we have one?

samlaf (approved):
LGTM, although I reopened the chisel convo because I feel like we should have an automated test if we don't have one already.

PR #1186, opened 2025-01-29
Node api audit

jianoaix:
- Enhance the API validation
- Improve error handling with right gRPC categorization (better UX, more accurate accounting of API availability)
- Place heavy computing (e.g. auth) **after** cheap/light validation

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/errors.go
```diff
@@ -10,39 +11,96 @@ import (
...
+//     which is harder to debug and understand for users.
+//   - See https://github.com/googleapis/googleapis/blob/ba8ea80f25d19bde8501cd51f314391f8d39bde8/google/rpc/code.proto
+//     for the mapping of grpc error codes to HTTP status codes.
 func newErrorGRPC(code codes.Code, msg string) error {
 	return status.Error(code, msg)
 }
 
+// WrapAsGRPCError wraps an error with a gRPC code and message. If the error already has a gRPC code,
+// it retains the code but combines the messages. Otherwise, it uses the provided code and combined message.
+func WrapAsGRPCError(err error, code codes.Code, msg string) error {
```
samlaf, line 28: When should this happen? Feels weird to me to have to wrap a GRPC error in another one, because a grpc status code wrapping should be terminal and returned to the user. Otherwise don't we need to define the overwrite/precedence logic for which error to return to user (eg 400 comes before 500 or not?)

api/errors.go
```diff
@@ -10,39 +11,96 @@ import (
...
+	if err == nil {
+		return nil
```
samlaf, line 30: shouldnt this just use newErrorGRPC in this case?

api/errors.go
```diff
@@ -10,39 +11,96 @@ import (
...
+	}
+
+	combinedMessage := fmt.Sprintf("%s: %s", msg, err.Error())
+	if s, ok := status.FromError(err); ok {
+		// If it's already a gRPC error, use the same code but combine the messages
+		return status.Error(s.Code(), combinedMessage)
```
samlaf, line 36: why use the old code? Why not the new one?

api/errors.go
```diff
@@ -10,39 +11,96 @@ import (
...
+		return nil
+	}
+
+	combinedMessage := fmt.Sprintf("%s: %s", msg, err.Error())
+	if s, ok := status.FromError(err); ok {
+		// If it's a gRPC error, keep it as gRPC error but combine the messages
+		return status.Error(s.Code(), combinedMessage)
+	}
+
+	return errors.New(combinedMessage)
```
samlaf, line 56: this function has weird semantics, its called WrapGRPCError but can accept a non-grpc error and return a non-grpc error. Can we simplify it instead?

node/auth/authenticator.go
```diff
@@ -139,7 +140,7 @@ func (a *requestAuthenticator) getDisperserKey(
 	disperserID uint32) (*gethcommon.Address, error) {
 
 	if !a.disperserIDFilter(disperserID) {
-		return nil, fmt.Errorf("invalid disperser ID: %d", disperserID)
+		return nil, api.NewErrorInvalidArg(fmt.Sprintf("invalid disperser ID: %d", disperserID))
```
samlaf, line 143: there's no mention of GRPC in this method name, so I had to go check whether its wrapping a grpc error code or not. Should we change the name to api.NewGRPCErrorInvalidArg or something now that we have many different error handling types/functions in api.?

node/auth/authenticator.go
```diff
@@ -120,7 +121,7 @@ func (a *requestAuthenticator) AuthenticateStoreChunksRequest(
 
 	key, err := a.getDisperserKey(ctx, now, request.DisperserID)
 	if err != nil {
-		return fmt.Errorf("failed to get operator key: %w", err)
+		return api.WrapGRPCError(err, "failed to get operator key")
```
samlaf, line 124: do we know for sure that err here is a grpc error?

node/grpc/server_v2.go
```diff
@@ -126,23 +114,36 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
 
 	batch, err := s.validateStoreChunksRequest(in)
 	if err != nil {
-		return nil, err
+		return nil, api.NewErrorInvalidArg(fmt.Sprintf("failed to validate store chunk request: %v", err))
```
samlaf, line 117: Based on the PR description, I guess validateStoreChunksRequest is a cheaper operation than authenticating the request, hence the ordering change? Could you add a comment describing that.

api/errors.go
samlaf: Also should prob make private so force people to use the other specific ones you wrote for each error code?

samlaf (commented):
Like the approach overall, and big fan of having better error handling. That being said the api.errors functions are growing, lots of them are public, and they dont have super clear semantic, so I feel it could proliferate in the code base to very unsound error wrapping when people other than you who aren't as familiar with your abstraction try to use them and just lazy out and use whatever method they can find that looks ok.

Don't have a super clean proposal for making this better though... so will let you decide.

api/errors.go
jianoaix: When a server uses other services as backend, this can happen.

If the server calls `HandleFooWithServiceX(params) error`, and an error returned, this error by default should be INTERNAL error, but it's also possible the `params` are no valid which results in an error (the caller may not have sufficient information to validate `params` before calling).

In such case you may want to `WrapAsGRPCError(err, INTERNAL, "....")`: if `HandleFooWithServiceX` has an opinion about the error type, then we should use that; otherwise, it should be INTERNAL.

api/errors.go
jianoaix: explained above

api/errors.go
jianoaix: explained above

api/errors.go
jianoaix: That's exactly the desired semantics. If there is a better concise name I'm happy to use it.

node/auth/authenticator.go
jianoaix: The errors in `api` are intended for APIs, which is gRPC. Unless there are non gRPC APIs I think it should be clear.

node/auth/authenticator.go
jianoaix: The point of this wrapper is to make it easier for APIs to return canonical errors, without having to read all underlying code

node/grpc/server_v2.go
jianoaix: yes. we shouldn't spend resource on deep processing a request if it's invalid, i.e. reject invalid request with least amount of resource, which is helpful for security (e.g. spam requests)

jianoaix:
> Like the approach overall, and big fan of having better error handling. That being said the api.errors functions are growing, lots of them are public, and they dont have super clear semantic, so I feel it could proliferate in the code base to very unsound error wrapping when people other than you who aren't as familiar with your abstraction try to use them and just lazy out and use whatever method they can find that looks ok.
> 
> Don't have a super clean proposal for making this better though... so will let you decide.

Would it be better if we drop sugars like
```
func WrapAsInvalidArg(err error, msg string) error {
	return WrapAsGRPCError(err, codes.InvalidArgument, msg)
}
```
and let the caller directly use `WrapAsGRPCError(err, codes.InvalidArgument, msg)`?

api/errors.go
samlaf: WrapGRPCError indicates that only GRPC errors should be allowed to be wrapped, yet your documentation says that its fine to pass another error and itll just get wrapped normally. Error wrapping also typically means wrapping an error, not just extending the error with some other msg string. So it feels convoluted and unclear to me.

I like when functions have a single small, and clear purpose. Just like golang's errors api, only has .Is() and .As(), and errors.join() for combining errors, thats it. Forces users to use it in the same way, and not too much cognitive overload when trying to use it.

How about we mention that this should ONLY be used for wrapping grpc errors? I don't think any function should return either a GRPC error or not right?

node/auth/authenticator.go
samlaf: I personally think this is a bad idea though. It's better to be SUPER clear on the semantics on functions. If a function returns a grpc error, then ALL of its returned errors should be GRPC. If not, then the function is too large and should be broken down to respect this invariant. This way error handling semantics and documentation is simple, and you can't go wrong.

Otherwise if "everything is easy" people just don't know what they are dealing with, and to extend a function or return a new error code they need to go read the entire code path to understand what kind of errors they could be receiving in different cases.

samlaf: better to code for easy static checkability

api/errors.go
samlaf: I see! Thanks for the explanation.

So right, if an external service returns a 400, that should be, from our server's client's pov, converted to a 500 (we are doing something wrong).

So I feel like something simpler would be:
- rename this function something clearer (`convertExternalServiceError`?) and your explanation should be added as documentation comment)
- or just get rid of this function altogether, and force people to use the other methods (unwrap the error string from external service's error and use NewErrorInternal(err.Err()))

I personally like to keep any package's public method surface small, so would prefer to remove functions.

api/errors.go
samlaf: Don't really like this semantic though, would prefer at each call site the more explicit
```
if err != nil {
   WrapAsGRPCError(err, code, msg)
}
```
instead of wrapping by default and forcing reader to go make sure that the function returns nil if no error was present

api/errors.go
samlaf: This doesn't match your description above though?
If externalService returns 400, we want to return 500 to our users, not 400. So you should use the code passed to wrapAsGRPCError always I think?

samlaf:
> Would it be better if we drop sugars like
> 
> ```
> func WrapAsInvalidArg(err error, msg string) error {
> 	return WrapAsGRPCError(err, codes.InvalidArgument, msg)
> }
> ```
> 
> and let the caller directly use `WrapAsGRPCError(err, codes.InvalidArgument, msg)`?

I personally think so yes. Even the original ones
```
func NewErrorUnauthenticated(msg string) error {
	return newErrorGRPC(codes.Unauthenticated, msg)
}
```
I feel like I prefer just having a single function used everywhere, with call site specifying the status code.
Maybe we can define our own code enum to only allow the ones that we want, so the api would look like
```
api.newErrorGRPC(api.CodeUnauthenticated, msg)
```
?

node/grpc/server_v2.go
```diff
@@ -198,13 +201,24 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
 
 // validateStoreChunksRequest validates the StoreChunksRequest and returns deserialized batch in the request
 func (s *ServerV2) validateStoreChunksRequest(req *pb.StoreChunksRequest) (*corev2.Batch, error) {
+	if req.GetDisperserID() != api.EigenLabsDisperserID {
```
dmanc, line 204: Can we add a comment why this would result in an invalid store chunks request?

api/errors.go
jianoaix: it's not just external service error, it can be the validation with more info from external service failed

api/errors.go
jianoaix: The above explanation is basically "if lower level function has opinion of error, preserve it"

api/errors.go
jianoaix: golang's api needs to be small as it has no context

the functions are not expected to return grpc error, only the APIs do; and the wrapper here works no matter if a function returns grpc error or not.

node/auth/authenticator.go
jianoaix: the issue is you cannot ask all functions to return grpc error which fits only at interface level

jianoaix:
Taking a step back I think most (if not all) of the api/errors.go belongs to common/, they are relevant only for internal uses in error handling; users only needs to know the errors they'd see.

api/errors.go
samlaf: its not defensive coding... its the wrong semantics.

api/errors.go
samlaf: But that is just wrong...

samlaf:
@jianoaix what are you proposing here? I feel like we fundamentally agree about a lot of points, and are talking past each other. I propose we have a live discussion to sync on this.

samlaf (approved):

jianoaix:
> ~@jianoaix what are you proposing here? I feel like we fundamentally agree about a lot of points, and are talking past each other. I propose we have a live discussion to sync on this.~
> 
> grug update: let's move on, you're the owner of this code, we'll do it your way. Will stamp.

I was basically saying that how we manufacture errors is not a concern for the users, hence that part of code/logic is better to be moved out of `api/`. The users care about what errors are and what actions should be taken.

node/grpc/server_v2.go
jianoaix: This is better to have onchain information (i.e. what dispersers are registered), so dropped this, and leave it to authenticator.

jianoaix:
ping

samlaf (approved):

PR #1187, opened 2025-01-29
Extract a payload retriever interface

litt3:

- Rather than complicating the existing payload retriever with fetching data directly from eigen DA nodes, this PR creates a new interface that can be implemented by the existing retriever, and other future types of retrievers
- This also gives flexibility in how the proxy can fetch payloads:
    - maybe a rollup wants to implement a retriever backed by an s3 bucket
    - maybe a rollup wants freedom in the order that backends are attempted for retrieval

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/clients/v2/payload_retriever.go
```diff
@@ -2,283 +2,12 @@ package clients
...
-
-	blobCommitment, err := encoding.BlobCommitmentsFromProtobuf(
-		verification.BlobCommitmentBindingToProto(blobCommitmentBinding))
-
-	if err != nil {
-		return nil, fmt.Errorf("blob commitments from protobuf: %w", err)
-	}
-
-	return blobCommitment, nil
+// PayloadRetriever represents something that knows how to retrieve a payload from some backend using a verification.EigenDACert
```
samlaf, line 9: Explain why we need more than one backend: relay, nodes, and potentially s3 are the 3 main candidates

samlaf: Btw does our RetrievalClient implement this interface? Should we not modify it as part of this PR?

api/clients/v2/relay_payload_retriever.go
```diff
@@ -0,0 +1,289 @@
...
+	"github.com/Layr-Labs/eigenda/encoding"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	"github.com/consensys/gnark-crypto/ecc/bn254"
+	gethcommon "github.com/ethereum/go-ethereum/common"
+)
+
+// RelayPayloadRetriever provides the ability to get payloads from the relay subsystem.
+//
+// This struct is goroutine safe.
+type RelayPayloadRetriever struct {
```
samlaf, line 23: I personally MSB order (see https://github.com/tigerbeetle/tigerbeetle/blob/main/docs/TIGER_STYLE.md#naming-things), so that the implementations of PayloadRetriever would be:
- PayloadRetrieverRelay
- PayloadRetrieverS3
- PayloadRetrieverNodes

This makes it much easier to find them using simple ctrl-f, or just searching for `PayloadRetriever`

But not going to push super hard on this one if you don't like it

samlaf (commented):
Is this PR a good time to start some better package management practices? How about creating a package for payload_retriever with subpackages (subdirs) for all its implementations? Would make it obvious what the struct is (this is the ben johnson approach that I mention in the style guide)

litt3:
> Is this PR a good time to start some better package management practices? How about creating a package for payload_retriever with subpackages (subdirs) for all its implementations? Would make it obvious what the struct is (this is the ben johnson approach that I mention in the style guide)

@samlaf It's not immediately clear to me what the subpackage around the `PayloadRetrieverRelay` would be named. Looking at Ben Johnson's [article](https://www.gobeyond.dev/standard-package-layout/), it seems to suggest naming it `relay`, which doesn't seem to make sense. The only package name I can think of that would be understandable would be `payload_retriever_relay`, which also seems odd: like we're relegating the *real* struct name to the package

```
v2/
    payload_retriever.go
    payload_retriever_relay/
        payload_retriever.go
```

Is this really considered an improvement? Have I misunderstood the strategy?

litt3:
> Is this PR a good time to start some better package management practices? How about creating a package for payload_retriever with subpackages (subdirs) for all its implementations? Would make it obvious what the struct is (this is the ben johnson approach that I mention in the style guide)

@samlaf It's not immediately clear to me what the subpackage around the `PayloadRetrieverRelay` would be named. Looking at Ben Johnson's [article](https://www.gobeyond.dev/standard-package-layout/), it seems to suggest naming it `relay`, which doesn't seem to make sense. The only package name I can think of that would be understandable would be `payload_retriever_relay`, which also seems odd: like we're relegating the *real* struct name to the package

```
v2/
    payload_retriever.go
    payload_retriever_relay/
        payload_retriever.go
```

Is this really considered an improvement? Have I misunderstood the strategy?

api/clients/v2/payload_retriever.go
litt3: [95b2a3ee](https://github.com/Layr-Labs/eigenda/pull/1187/commits/95b2a3ee7c432b4de22eb618aa54a0459ed2ae3a)

litt3: > Btw does our RetrievalClient implement this interface? Should we not modify it as part of this PR?

I don't think so. Just as we have the `RelayPayloadRetriever`, which wraps around the `RelayClient`, I think we need to have a `DaNodePayloadRetriever`, which wraps around the `RetrievalClient`.

api/clients/v2/relay_payload_retriever.go
litt3: I'm torn on this one. In general I agree, but the name `PayloadRetrieverRelay` implies that this thing *IS* a relay of some kind. I think the potential confusion there outweighs the `ctrl-f` benefit

api/clients/v2/payload_retriever.go
cody-littley: Ideally, the an end user would have a single payload retriever that knows how to talk to relays and validators. The payload retriever would make a best effort attempt to fetch the data from the relays, only falling back to the validators if unsuccessful at getting the data from the relays.

api/clients/v2/relay_payload_retriever.go
cody-littley: > PayloadRetrieverS3

Do we have a payload retriever for S3? Is this purely an internal use tool? The general public shouldn't have direct access to S3, except as mitigated through the relays.

cody-littley: My personal preference on word ordering would be `RelayPayloadRetriever`.

api/clients/v2/relay_payload_retriever.go
```diff
@@ -0,0 +1,289 @@
...
+		certVerifier: certVerifier,
+		g1Srs:        g1Srs,
+	}, nil
+}
+
+// GetPayload iteratively attempts to fetch a given blob with key blobKey from relays that have it, as claimed by the
+// blob certificate. The relays are attempted in random order.
+//
+// If the blob is successfully retrieved, then the blob is verified. If the verification succeeds, the blob is decoded
+// to yield the payload (the original user data), and the payload is returned.
```
cody-littley, line 108: Does this utility chop off the extra padding at the end? Zeros sometimes get appended to the end of the data to make the blob a power of 2. Either way, would be nice to mention this in the godoc.

api/clients/v2/payload_retriever.go
litt3: > Ideally, the an end user would have a single payload retriever that knows how to talk to relays and validators.

Each individual rollup may have different configurations for how they want to retrieve payloads. Should you retry from relays before trying a fallback? Should you fallback to some S3 bucket instead of falling back to DA nodes? If our only wrapper for payload retrieval is opinionated here, then that limits options.

Practically speaking, the main "user" of this utility is our own proxy. Most "end users" will mainly interact with the proxy. But having this interface also lets more advanced end users provide the proxy with their own retrieval backends, if required.

api/clients/v2/relay_payload_retriever.go
litt3: > Do we have a payload retriever for S3

This isn't referring to any eigen DA controlled S3. Rather, some rollups might have their own home-rolled storage backends.

api/clients/v2/relay_payload_retriever.go
litt3: Using our most recently defined terms, `Payload` has a specific definition: it is precisely the bytes that the user wanted to store. Any padding or processing definitionally means the result is no longer the "Payload".

I'll expand the doc to mention that here. I have some [in-progress](https://github.com/Layr-Labs/eigenda/pull/1120/files) changes that will create a typesystem to make these definitions explicit, but it's back-burner for now.

litt3: Done

api/clients/v2/payload_retriever.go
samlaf: Good discussion. Resolving.

api/clients/v2/relay_payload_retriever.go
samlaf: I am outvoted. Resolving.

Regarding "PayloadRetrieverRelay implies that this thing IS a Relay". That just requires some brain restructuring, the same way reverse polish notation does. But I'm not going to fight this fight. Agree that it might make more sense in other scenarios that are less syntactically overloaded.

samlaf:
> > Is this PR a good time to start some better package management practices? How about creating a package for payload_retriever with subpackages (subdirs) for all its implementations? Would make it obvious what the struct is (this is the ben johnson approach that I mention in the style guide)
> 
> @samlaf It's not immediately clear to me what the subpackage around the `PayloadRetrieverRelay` would be named. Looking at Ben Johnson's [article](https://www.gobeyond.dev/standard-package-layout/), it seems to suggest naming it `relay`, which doesn't seem to make sense. The only package name I can think of that would be understandable would be `payload_retriever_relay`, which also seems odd: like we're relegating the _real_ struct name to the package
> 
> ```
> v2/
>     payload_retriever.go
>     payload_retriever_relay/
>         payload_retriever.go
> ```
> 
> Is this really considered an improvement? Have I misunderstood the strategy?

packages can be renamed when imported, so I think the structure is more important personally.
But what's wrong with a long package name like `payload_retriever_relay` literally?
We use a similar strategy for the backend stores in proxy: https://github.com/Layr-Labs/eigenda-proxy/tree/main/store
Curious what you'll think of that.

Also the structure would be
```
v2/
   payload_retriever/
     payload_retriever.go # contains interfaces
     relay/
       relay.go # or longer name
     nodes/
       node.go # or longer name
```

samlaf (approved):
LGTM like this. We can do package shuffling later.

litt3:
> But what's wrong with a long package name like payload_retriever_relay literally?

@samlaf Nothing wrong with a long package name from my perspective. The thing that I don't like is when the important name is the package name, and the filename ends up being something generic and unhelpful, or a duplicate of many other files in the codebase. I think it's a side effect of how I tend to navigate a codebase: I find it super convenient when file names are verbose and meaningful, because you can jump directly to a file via a shortcut, without ever having to look at the package structure.

```
v2/
   payload_retriever/
     payload_retriever.go # contains interfaces
     relay/
       relay.go # or longer name
     nodes/
       node.go # or longer name
   ```

I think a file name like `relay.go` in your example would be misleading: if I see a file called `relay.go`, I'm gonna expect it to contain a relay. I see you have the note `or longer name`: if the longer name is `relay_payload_retriever.go` then I wouldn't have any problem with the structure in your example.

The more I consider it, my only strong preference is that a file name descriptively informs you of what is contained in the file (verbosity is fine). But I might be fighting a losing battle there, since it seems standard go practice encourages pulling a lot of the meaningful naming into packages.

PR #1189, opened 2025-01-30
Added check in API server for commitment length

cody-littley:

Fail a dispersal request if the blob length exceeds the commitment length.


disperser/apiserver/disperse_blob_v2.go
```diff
@@ -153,6 +153,9 @@ func (s *DispersalServerV2) validateDispersalRequest(ctx context.Context, req *p
 	if commitmentLength == 0 || commitmentLength != encoding.NextPowerOf2(commitmentLength) {
 		return errors.New("invalid commitment length, must be a power of 2")
 	}
+	if commitmentLength < uint32(len(req.Blob)) {
+		return fmt.Errorf("commitment length %d is less than blob length %d", commitmentLength, len(req.Blob))
+	}
```
cody-littley, line 158: I don't think this check is correct... don't merge until fixed

anupsv: This is the length of the polynomial right ?

anupsv: Also would this handle the edge case of 32 bytes of 0 at the end of the blob.

cody-littley: updated based on slack conversation

Please double check my work here, I don't have a solid grasp of the math here 😅

disperser/apiserver/disperse_blob_v2.go
```diff
@@ -153,6 +157,10 @@ func (s *DispersalServerV2) validateDispersalRequest(ctx context.Context, req *p
 	if commitmentLength == 0 || commitmentLength != encoding.NextPowerOf2(commitmentLength) {
 		return errors.New("invalid commitment length, must be a power of 2")
 	}
+	lengthPowerOf2 := encoding.GetBlobLengthPowerOf2(uint(blobSize))
+	if lengthPowerOf2 > uint(commitmentLength) {
```
jianoaix, line 161: Commitment length is a fixed number that's determined by the curve we are using (bn254). I don't see it's comparable to the blob length. cc @bxue-l2

anupsv: Yes, per the comment:

```
// The length specifies the degree of the polynomial used to generate the blob commitment. The length
// must equal to the degree + 1, and it must be a power of 2.
```

cody-littley: I named the variable based on the function... should both be renamed? What name would you suggest?

jianoaix: Ok this is doing the right comparison. I think "commitmentLength" is the misleading part, it's more accurate to say "committedBlobLength" (or just "committedLength")

jianoaix (approved):

disperser/apiserver/disperse_blob_v2.go
jianoaix: Note L207 is recomputing the commitment and then comparing with the one from request, so that's a stronger validation. But it may still make sense to have this one as a quick sanity screening here.

PR #1192, opened 2025-01-31
chore: add heartbeat monitoring for controller

yujiezhu0:

- Implemented health checking mechanisms for the controller, including heartbeat monitoring, readiness probe, liveness probe, and health probe.
- These changes are essential for alerts and monitoring.
- Linear: EGDA-814.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests


disperser/cmd/controller/main.go
```diff
@@ -49,10 +55,21 @@ func main() {
 	if err != nil {
 		log.Fatalf("application failed: %v", err)
 	}
+
+	// Start HeartBeat Monitor
```
dmanc, line 59: nit: `Heatbeat` instead of `HeartBeat`

disperser/cmd/controller/main.go
```diff
@@ -34,6 +35,11 @@ var (
 	version   string
 	gitCommit string
 	gitDate   string
+
+	controllerReadinessProbePath string        = "/tmp/controller-ready"
+	controllerHealthProbePath    string        = "/tmp/controller-health"
+	controllerMaxStallDuration   time.Duration = 240 * time.Second
+	controllerLivenessChan                     = make(chan time.Time, 1)
```
dmanc, line 42: Doesn't seem like we ever send a liveness message to this channel in the code. See https://github.com/Layr-Labs/eigenda/blob/a1fceb9e015325746d5db9e14b5fbc515b731ca1/disperser/batcher/batcher.go#L450 for an example

disperser/cmd/controller/main.go
```diff
@@ -201,5 +218,41 @@ func RunController(ctx *cli.Context) error {
...
 
+	// Signal readiness once the controller starts successfully
+	if _, err := os.Create(controllerReadinessProbePath); err != nil {
+		log.Printf("Failed to create readiness file: %v at path %v \n", err, controllerReadinessProbePath)
+	}
+
 	return nil
 }
+
+// Function to process and send controller liveness probe to GoRoutine
```
dmanc, line 229: nit: `goroutine` instead of `GoRoutine`

disperser/cmd/controller/main.go
```diff
@@ -34,6 +36,12 @@ var (
 	version   string
 	gitCommit string
 	gitDate   string
+
+	controllerReadinessProbePath string        = "/tmp/controller-ready"
+	controllerHealthProbePath    string        = "/tmp/controller-health"
+	controllerMaxStallDuration   time.Duration = 240 * time.Second
+	controllerLivenessChan                     = make(chan time.Time, 1)
+	controllerHeartbeatChan                    = make(chan time.Time, 1)
```
dmanc, line 44: Not sure this is needed.

disperser/cmd/controller/main.go
```diff
@@ -201,5 +224,54 @@ func RunController(ctx *cli.Context) error {
 		}
 	}()
 
+	// Create readiness probe file once the controller starts successfully
+	if _, err := os.Create(controllerReadinessProbePath); err != nil {
+		logger.Warn("Failed to create readiness file", "error", err, "path", controllerReadinessProbePath)
+	}
+
+	// Signal heartbeat
+	signalHeartbeat(logger)
```
dmanc, line 233: We should add heartbeat signals in the main processing loops of both components:
1. Encoding manager: before `HandleBatch()`
2. Dispatcher: before `HandleBatch()`

Right now we're only signaling liveness at startup. But what we want is to have heartbeats for each control loop. Then our alerting system should alert if any of the two control loops fail and we should know which one failed.

disperser/cmd/controller/main.go
```diff
@@ -49,6 +56,14 @@ func main() {
 	if err != nil {
 		log.Fatalf("application failed: %v", err)
 	}
+
+	if _, err := os.Create(controllerHealthProbePath); err != nil {
```
ian-shim, line 60: why do we need to create the readiness file here if it's getting cleaned up and recreated in `RunController`?

disperser/cmd/controller/main.go
```diff
@@ -178,6 +201,7 @@ func RunController(ctx *cli.Context) error {
 		nodeClientManager,
 		logger,
 		metricsRegistry,
+		func() { signalHeartbeat(controllerLivenessChan, logger) },
```
ian-shim, line 204: if encoding manager and dispatcher are using the same signal, would heartbeat monitor detect if one of them is down?

disperser/controller/controller_test.go
```diff
@@ -180,3 +203,50 @@ func newBlob(t *testing.T, quorumNumbers []core.QuorumID) (corev2.BlobKey, *core
 	require.NoError(t, err)
 	return bk, bh
 }
+
+func TestHeartbeatMonitoring(t *testing.T) {
```
ian-shim, line 207: Not sure how this is testing the heartbeat reported by the controller. It doesn't seem to test the controller's actual behavior but just simulate what it theoretically should be doing?

disperser/controller/dispatcher.go
```diff
@@ -43,7 +43,8 @@ type Dispatcher struct {
 	logger            logging.Logger
 	metrics           *dispatcherMetrics
 
-	cursor *blobstore.StatusIndexCursor
+	cursor          *blobstore.StatusIndexCursor
+	signalHeartbeat func()
```
ian-shim, line 47: instead of passing in a function, can we pass in the channel instead?

disperser/cmd/controller/main.go
yujiezhu0: This is a different file being created here - the `health probe file`, which is updated continuously by the heartbeat monitor to indicate ongoing liveness. 
The file being removed upon RunController startup is the `readiness file`, and is created after all the components successfully start (at the end of RunController).

disperser/controller/dispatcher.go
yujiezhu0: Why do you think passing in the channel would be preferable?

I think passing in the function is preferable because heartbeat logic is encapsulated in controller main.go and decouples from components like encoding manager and dispatcher.

disperser/cmd/controller/main.go
yujiezhu0: You're right. I'm going to update `func signalHeartbeat()` to include a message indicating which component is being monitored. This way, if one component fails to send its heartbeat, we’ll know exactly which one is down.

disperser/controller/controller_test.go
yujiezhu0: Right. I realized that functions in main.go are not importable (making it impossible to test directly) and it's probably a better idea to move the heartbeat related functions out to util. I'll add more tests in dispatcher and encoding manager.

disperser/controller/dispatcher_test.go
```diff
@@ -607,13 +658,29 @@ func newDispatcherComponents(t *testing.T) *dispatcherComponents {
 	}
 	blobSet := &controller.MockBlobSet{}
 	blobSet.On("Size", mock.Anything).Return(0)
+
+	// Heartbeat tracking variables
```
ian-shim, line 662: If we call `SignalHeartbeat` directly inside the components, could we test the actual `SignalHeartbeat` method instead of only testing that the component called the passed in function?

common/healthcheck/heartbeat.go
```diff
@@ -0,0 +1,81 @@
...
+	// Map to keep track of last heartbeat per component
+	lastHeartbeats := make(map[string]time.Time)
+	// Create a timer that periodically checks for stalls
+	stallTimer := time.NewTimer(maxStallDuration)
+
+	for {
+		select {
+		case hb, ok := <-livenessChan:
+			if !ok {
+				log.Println("livenessChan closed, stopping health probe.")
```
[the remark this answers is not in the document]
yujiezhu0: liveness channel closing should be normal, but I'll update it to use logger here.

disperser/cmd/controller/main.go
```diff
@@ -35,6 +37,10 @@ var (
 	version   string
 	gitCommit string
 	gitDate   string
+
+	controllerReadinessProbePath = "/tmp/controller-ready"
+	controllerHealthProbePath    = "/tmp/controller-health"
```
ian-shim, line 42: should these be configurable (set by some flags) vs. hardcoded?

ian-shim (commented):
looks good to me. left last few questions.

common/healthcheck/heartbeat.go
```diff
@@ -0,0 +1,86 @@
...
+)
+
+type HeartbeatMessage struct {
+	Component string    // e.g., "encodingManager" or "dispatcher"
+	Timestamp time.Time // when the heartbeat was sent
+}
+
+// HeartbeatMonitor listens for heartbeat messages from different components, updates their last seen timestamps,
+// writes a summary to the specified file, and logs warnings if any component stalls.
+func HeartbeatMonitor(filePath string, maxStallDuration time.Duration, livenessChan <-chan HeartbeatMessage, logger logging.Logger) error {
```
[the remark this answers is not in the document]
yujiezhu0: `HeartbeatMonitor` was [batcher's convention](https://github.com/Layr-Labs/eigenda/blob/fc9a97144585186f25bcdc77134d02a6de9d53f0/disperser/cmd/batcher/main.go#L262). I think we can keep it for now. If in the future we want all the components to use this HeartbeatMonitor, we can remove the one from batcher and rename.

common/healthcheck/heartbeat.go
```diff
@@ -0,0 +1,86 @@
...
+			// Check for components that haven't sent a heartbeat recently
+			now := time.Now()
+			var staleComponents []string
+			for comp, ts := range lastHeartbeats {
+				if now.Sub(ts) > maxStallDuration {
+					staleComponents = append(staleComponents, fmt.Sprintf("Component %s: last heartbeat at %v", comp, ts))
+				}
+			}
+			if len(staleComponents) > 0 {
+				logger.Warn(
```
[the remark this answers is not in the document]
yujiezhu0: I think a “stalled” component isn’t a bug in the monitor. It just means nothing has come through in some time. The monitor should keep running and keep checking, so we wouldn’t want it to treat that as an error.

common/healthcheck/heartbeat.go
```diff
@@ -0,0 +1,86 @@
...
+	Timestamp time.Time // when the heartbeat was sent
+}
+
+// HeartbeatMonitor listens for heartbeat messages from different components, updates their last seen timestamps,
+// writes a summary to the specified file, and logs warnings if any component stalls.
+func HeartbeatMonitor(filePath string, maxStallDuration time.Duration, livenessChan <-chan HeartbeatMessage, logger logging.Logger) error {
+	// Map to keep track of last heartbeat per component
+	lastHeartbeats := make(map[string]time.Time)
+	// Create a timer that periodically checks for stalls
+	stallTimer := time.NewTimer(maxStallDuration)
```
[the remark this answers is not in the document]
yujiezhu0: Sounds good

disperser/cmd/controller/main.go
```diff
@@ -217,5 +230,22 @@ func RunController(ctx *cli.Context) error {
...
+
+	if _, err := os.Create(config.ControllerHealthProbePath); err != nil {
+		logger.Warn("Failed to create healthProbe file: %v", err)
+	}
+
+	// Start heartbeat monitor
+	go func() {
+		err := healthcheck.HeartbeatMonitor(config.ControllerHealthProbePath, controllerMaxStallDuration, controllerLivenessChan, logger)
+		if err != nil {
+			logger.Warn("Failed to start heartbeatMonitor: %v", err)
```
ian-shim, line 246: I think this case definitely deserves to fail more loudly (Error level)
If HeartbeatMonitor does not run but component keeps sending heartbeat, the component can eventually stall because the channel can be full. Is my understanding correct? 
We should fail loudly here and maybe close the channel as well

common/healthcheck/heartbeat.go
ian-shim, line 19: does it ever return error?

disperser/controller/dispatcher.go
```diff
@@ -158,6 +162,9 @@ func (d *Dispatcher) Start(ctx context.Context) error {
 }
 
 func (d *Dispatcher) HandleBatch(ctx context.Context) (chan core.SigningMessage, *batchData, error) {
+	// Signal Liveness to indicate no stall
+	healthcheck.SignalHeartbeat("dispatcher", d.controllerLivenessChan, d.logger)
```
ian-shim, line 166: ironically, this can stall the component if the channel is full, no?

ian-shim (approved):

PR #1194, opened 2025-01-31
Implement validator payload retrieval

litt3:

- This PR required a change to the signature of the retriever client `GetBlob` method. Specifically, the previous method signature required access to a `BlobHeader`, which cannot be constructed by a client, since they do not have the payment data.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/clients/v2/distributed_payload_retriever.go
```diff
@@ -0,0 +1,210 @@
...
+}
+
+var _ PayloadRetriever = &DistributedPayloadRetriever{}
+
+// BuildDistributedPayloadRetriever builds a DistributedPayloadRetriever from config structs.
+func BuildDistributedPayloadRetriever(
+	logger logging.Logger,
+	distributedPayloadRetrieverConfig DistributedPayloadRetrieverConfig,
+	ethConfig geth.EthClientConfig,
+	thegraphConfig thegraph.Config,
```
bxue-l2, line 40: I think we no longer require graph node for retrieving the DA node sockets. Instead, we can call a contract, https://github.com/Layr-Labs/eigenlayer-middleware/blob/fe5834371caed60c1d26ab62b5519b0cbdcb42fa/src/SocketRegistry.sol#L17, and use the mapping there to get the IP address, @0x0aa0 is that right?

mainnet: [0x5a3eD432f2De9645940333e4474bBAAB8cf64cf2](https://etherscan.io/address/0x5a3eD432f2De9645940333e4474bBAAB8cf64cf2#code)

preprod: [0x1747ef24dbbb52cB06382d323f455D48dE1AC7fd](https://holesky.etherscan.io/address/0x1747ef24dbbb52cb06382d323f455d48de1ac7fd#readProxyContract)

testnet: [0x25aFC8944f501545DDB7E7C4C8A0119965AAb166](https://holesky.etherscan.io/address/0x25aFC8944f501545DDB7E7C4C8A0119965AAb166#code)

api/clients/v2/distributed_payload_retriever.go
```diff
@@ -0,0 +1,210 @@
...
+
+		valid, err := verification.GenerateAndCompareBlobCommitment(pr.g1Srs, blobBytes, commitment.Commitment)
+		if err != nil {
+			pr.logger.Warn(
+				"generate and compare blob commitment",
+				"blobKey", blobKey.Hex(), "quorumID", quorumID, "error", err)
+			continue
+		}
+		if !valid {
+			pr.logger.Warn("cert is invalid", "blobKey", blobKey.Hex(), "quorumID", quorumID)
```
bxue-l2, line 169: would commitment is inconsistent be more concrete? invalid cert might mean wrong signature

api/clients/v2/relay_payload_retriever.go
```diff
@@ -154,11 +153,21 @@ func (pr *RelayPayloadRetriever) GetPayload(
...
+		valid, err := verification.GenerateAndCompareBlobCommitment(pr.g1Srs, blob, blobCommitments.Commitment)
 		if err != nil {
-			pr.log.Warn("verify blob from relay against cert: %w", err)
+			pr.log.Warn(
+				"generate and compare blob commitment",
+				"blobKey", blobKey.Hex(), "relayKey", relayKey, "error", err)
+			continue
+		}
+		if !valid {
+			pr.log.Warn("cert is invalid", "blobKey", blobKey.Hex(), "relayKey", relayKey)
```
bxue-l2, line 170: "commitments are different" is more descriptive

litt3: Changed to "generated commitment doesn't match cert commitment" [db338cd7](https://github.com/Layr-Labs/eigenda/pull/1194/commits/db338cd787368663147b821795360091e15b7c71)

api/clients/v2/distributed_payload_retriever.go
litt3: [db338cd7](https://github.com/Layr-Labs/eigenda/pull/1194/commits/db338cd787368663147b821795360091e15b7c71)

api/clients/v2/distributed_payload_retriever.go
```diff
@@ -0,0 +1,212 @@
...
+	config          DistributedPayloadRetrieverConfig
+	codec           codecs.BlobCodec
+	retrievalClient RetrievalClient
+	g1Srs           []bn254.G1Affine
+}
+
+var _ PayloadRetriever = &DistributedPayloadRetriever{}
+
+// BuildDistributedPayloadRetriever builds a DistributedPayloadRetriever from config structs.
+func BuildDistributedPayloadRetriever(
```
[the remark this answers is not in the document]
litt3: I chose the term "distributed", since the payload is being retrieved from a distributed set of DA nodes, which each provide chunks.

I don't think `MultiQuorumPayloadRetriever` would be a good substitute, since the property being highlighted isn't that it retries multiple quorums: I'm looking for a term to distinguish from a utility that retrieves payloads from a relay, or other backend. I'm not attached to using "distributed", though

While working on the best name for this utility, I think we should also consider what the best name for `retrieval client` is. I don't think that's a good name for the current utility, since there are multiple clients that do retrieval, and it's arbitrary to pick one of these and dub it *the* "retrieval client"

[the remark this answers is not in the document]
litt3: As we discussed synchronously, I'm changing the name to `ValidatorPayloadRetriever`

api/clients/v2/distributed_payload_retriever.go
0x0aa0: We will be able to retrieve onchain once all the operators have migrated over and set that in storage but this will not be instant so we will probably require the graph for a time while that migration takes place

api/clients/v2/config.go
```diff
@@ -154,6 +172,51 @@ func (rc *RelayPayloadRetrieverConfig) checkAndSetDefaults() error {
...
+	return &DistributedPayloadRetrieverConfig{
+		PayloadClientConfig: *getDefaultPayloadClientConfig(),
+		RetrievalTimeout:    20 * time.Second,
+		ConnectionCount:     10,
+	}
+}
+
+// checkAndSetDefaults checks an existing config struct and performs the following actions:
+//
+// 1. If a config value is 0, and a 0 value makes sense, do nothing.
```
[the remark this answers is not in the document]
litt3: What do you think about this rephrasing?

```
// checkAndSetDefaults checks an existing config struct. It performs one of the following actions for any contained 0 values:
//
// 1. If 0 is an acceptable value for the field, do nothing.
// 2. If 0 is NOT an acceptable value for the field, and a default value is defined, then set it to the default.
// 3. If 0 is NOT an acceptable value for the field, and a default value is NOT defined, return an error.
```

cody-littley (approved):

api/clients/v2/config.go
```diff
@@ -61,6 +61,24 @@ type RelayPayloadRetrieverConfig struct {
 	RelayTimeout time.Duration
 }
 
+// ValidatorPayloadRetrieverConfig contains an embedded PayloadClientConfig, plus all additional configuration values
+// needed by a ValidatorPayloadRetriever
+type ValidatorPayloadRetrieverConfig struct {
+	PayloadClientConfig
+
+	// The timeout duration for calls to retrieve blobs from a given quorum.
+	RetrievalTimeout time.Duration
```
samlaf, line 70: is this for each individual call to a single node? Or is this a global timeout for retrieving the entire blob from all nodes?

If for single calls might be more accurate to say "to retrieve chunks"

api/clients/v2/config.go
```diff
@@ -61,6 +61,24 @@ type RelayPayloadRetrieverConfig struct {
...
+
+	// The address of the BlsOperatorStateRetriever contract
+	BlsOperatorStateRetrieverAddr string
+
+	// The address of the EigenDAServiceManager contract
+	EigenDAServiceManagerAddr string
+
+	// The number of simultaneous connections to use when fetching chunks during validator retrieval
+	ConnectionCount uint
```
samlaf, line 79: shouldnt this be a max count instead?

api/clients/v2/verification/blob_verifier.go
```diff
@@ -0,0 +1,28 @@
...
+	"errors"
+	"fmt"
+
+	"github.com/Layr-Labs/eigenda/encoding"
+)
+
+// CheckBlobLength accepts bytes representing a blob, and a claimed length in symbols. Note that claimed length is an
+// upper bound, not a precise length. Two length checks are performed:
+//
+// 1. Blob doesn't have length 0
```
samlaf, line 13: this still trips me a bit. Why are we not fine with a blob of length 0 but we're fine with a blob of length 2?

api/clients/v2/distributed_payload_retriever.go
bxue-l2: @0x0aa0 in V2, does operator inherit registration in V1, or have to register independently in V2? then it looks like migration is not a constraint for V2. But it forces us to support two version for retrieval.

api/clients/v2/validator_payload_retriever.go
```diff
@@ -0,0 +1,212 @@
...
+}
+
+// GetPayload iteratively attempts to retrieve a given blob from the quorums listed in the EigenDACert.
+//
+// If the blob is successfully retrieved, then the blob verified against the EigenDACert. If the verification succeeds,
+// the blob is decoded to yield the payload (the original user data, with no padding or any modification), and the
+// payload is returned.
+func (pr *ValidatorPayloadRetriever) GetPayload(
+	ctx context.Context,
+	eigenDACert *verification.EigenDACert,
```
samlaf, line 122: cert might need to be moved to some common/core package instead now that we also want to have a bindings version to be used for the contracts?

samlaf (commented):
LGTM. Small nits/questions below

api/clients/v2/validator_payload_retriever.go
```diff
@@ -0,0 +1,212 @@
...
+		logger,
+		ethClient,
+		validatorPayloadRetrieverConfig.BlsOperatorStateRetrieverAddr,
+		validatorPayloadRetrieverConfig.EigenDAServiceManagerAddr)
+	if err != nil {
+		return nil, fmt.Errorf("new reader: %w", err)
+	}
+
+	chainState := eth.NewChainState(reader, ethClient)
+	indexedChainState := thegraph.MakeIndexedChainState(thegraphConfig, chainState, logger)
```
ethenotethan, line 63: does this feature assume the use of a graph node?

api/clients/v2/distributed_payload_retriever.go
litt3: @0x0aa0 Theoretically we should be able to make this change right away for this code, since operators will have to make the required updates to be able to use V2 anyway?

api/clients/v2/config.go
litt3: This is the amount of time we will wait while trying to retrieve chunks from a given quorum, and reassemble the blob (so, the first option you said)

I will modify this doc to be clearer

litt3: [8be41aa7](https://github.com/Layr-Labs/eigenda/pull/1194/commits/8be41aa7b9f639e44e894f4d44669aa441910ccd)

api/clients/v2/config.go
litt3: [3f6f092c](https://github.com/Layr-Labs/eigenda/pull/1194/commits/3f6f092c87fa167c8b5d1462e2e6ffaa4c456785)

api/clients/v2/validator_payload_retriever.go
litt3: I think that makes sense, but it's not clear to me exactly where it ought to be moved.

api/clients/v2/config.go
```diff
@@ -54,6 +54,24 @@ type RelayPayloadRetrieverConfig struct {
...
+// DistributedPayloadRetrieverConfig contains an embedded PayloadClientConfig, plus all additional configuration values
+// needed by a DistributedPayloadRetriever
+type DistributedPayloadRetrieverConfig struct {
+	PayloadClientConfig
+
+	// The timeout duration for calls to retrieve blobs from a given quorum.
+	RetrievalTimeout time.Duration
+
+	// The address of the BlsOperatorStateRetriever contract
+	BlsOperatorStateRetrieverAddr string
```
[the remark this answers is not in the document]
litt3: I don't think so. Looking at [reader.go](https://github.com/Layr-Labs/eigenda/blob/77d4442aa1b37bdc275173a6b27d917cc161474c/core/eth/reader.go#L67), `BlsOperatorStateRetrieverAddr` and `eigenDAServiceManagerHexAddr` are both required arguments. Though I'm not sure why that's the case

api/clients/v2/verification/blob_verifier.go
litt3: @bxue-l2 I feel like we may have discussed this in the past, but I'm not sure.

What are your thoughts here: do we need to check if the blob is length 0 here?

api/clients/v2/config.go
```diff
@@ -194,7 +194,7 @@ func GetDefaultValidatorPayloadRetrieverConfig() *ValidatorPayloadRetrieverConfi
 	return &ValidatorPayloadRetrieverConfig{
 		PayloadClientConfig: *getDefaultPayloadClientConfig(),
 		RetrievalTimeout:    20 * time.Second,
-		ConnectionCount:     10,
+		MaxConnectionCount:  10,
```
samlaf, line 197: shouldn't this be like 100? With 200 validators, assuming each call takes 1 sec, just hitting all the validators already takes 20 seconds.

api/clients/v2/validator_payload_retriever.go
samlaf: yep lets. keep as is for now

samlaf (approved):
LGTM, although would be good to decide what to do with the graph node requirement, and also to get an answer for @bxue-l2 as to why we need that blobLength != 0 check.

api/clients/v2/config.go
litt3: Good point. I increased the connection count to `100`, and also increased retrieval timeout to 30 seconds, to add a little buffer.

```
		RetrievalTimeout:    30 * time.Second,
		MaxConnectionCount:  100,
```

api/clients/v2/distributed_payload_retriever.go
litt3: I created [this issue](https://github.com/Layr-Labs/eigenda/issues/1228) to address in the near future

api/clients/v2/verification/blob_verifier.go
litt3: I created an issue for this https://github.com/Layr-Labs/eigenda/issues/1230

bxue-l2: For proxy, we added a header, so the length is at least 1.
For non-proxy, the data length itself is literally 0. Then I don't think it makes sense to do any erasure coding (i.e there is no polynomial, since there is no data)

litt3: So, are you saying we _should_ enforce a non zero length here?

bxue-l2: yes

samlaf: But this case would have a blob length of 1 in the blob_commitment no? So even if relay returns 0 bytes (because header is all 0s and he truncated all of them), client can just pad it back to 32 and then decode no? Why does it matter if relay sends nothing back? That should just mean the whole header is 0s afaiu?

litt3: > So even if relay returns 0 bytes (because header is all 0s and he truncated all of them), client can just pad it back to 32 and then decode no?

Ah, I think you are right. This considered, I'm in favor of removing the 0 length check.

bxue-l2: I feel this is a protocol problem, like should we accept empty blob?

EigenDA disperser rejects blobs that has size 0, https://github.com/Layr-Labs/eigenda/blob/master/api/proto/disperser/v2/disperser_v2.proto#L32. 

Not sure if this is already restricted by protocol. I.e. DA node signs only sign if chunk size >0

If we know that is for sure, then we should just stop early, when relay returns an empty bytes.

litt3: > I feel this is a protocol problem, like should we accept empty blob?

The argument is that we shouldn't do a 0 length check, because a blob with *non-zero* length, but that contains exclusively `0x0` bytes, *could* be successfully dispersed. But a relay might return a zero-length response (truncating all trailing 0x0s) which successfully verifies against the commitment.

PR #1195, opened 2025-01-31
Disable auth cache

cody-littley:

In testing, I found that auth caching does not work as we intended. This is because traffic is sometimes routed through things like load balancers, which make the IP addresses of different requesters appear to be the same.

### DIFF
--- a/relay/cmd/flags/flags.go
+++ b/relay/cmd/flags/flags.go
@@ -231,7 +231,7 @@ var (
 	AuthenticationDisabledFlag = cli.BoolFlag{
 		Name:     common.PrefixFlag(FlagPrefix, "authentication-disabled"),
 		Usage:    "Disable GetChunks() authentication",
-		Required: false,
+		Required: true, // TODO(cody-littley) remove this feature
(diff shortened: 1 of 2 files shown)


jianoaix:
Is authentication required or optional for each request?

The LB may be configured to pass along the client IP, may be worth check it.

anupsv:
> Is authentication required or optional for each request?
> 
> The LB may be configured to pass along the client IP, may be worth check it.

Are you referring to sticky sessions from aws LB ?

relay/cmd/flags/flags.go
```diff
@@ -231,7 +231,7 @@ var (
 	AuthenticationDisabledFlag = cli.BoolFlag{
 		Name:     common.PrefixFlag(FlagPrefix, "authentication-disabled"),
 		Usage:    "Disable GetChunks() authentication",
-		Required: false,
+		Required: true, // TODO(cody-littley) remove this feature
```
ian-shim, line 234: As sync'd offline, need to update this to only disable the caching

ian-shim (approved):

PR #1196, opened 2025-01-31
docs: quorum number requirement for payment method

hopeyen:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/proto/common/v2/common_v2.proto
+++ b/api/proto/common/v2/common_v2.proto
@@ -11,11 +11,13 @@ message BlobHeader {
   // blob's availability.
   uint32 version = 1;
   // quorum_numbers is the list of quorum numbers that the blob is part of.
-  // All quorums must be specified (including required quorums).
+  // All quorums must be specified.
   //
-  // The following quorums are currently required:
+  // On-demand dispersal is currently limited to use the following quorums:
   // - 0: ETH
   // - 1: EIGEN
+  // 
+  // If the dispersal is using reservations, the quorum numbers are limited to the specification on-chain for that particular reservation.
   repeated uint32 quorum_numbers = 2;
   // commitment is the KZG commitment to the blob
   common.BlobCommitment commitment = 3;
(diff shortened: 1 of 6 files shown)


api/proto/common/v2/common_v2.proto
```diff
@@ -11,11 +11,13 @@ message BlobHeader {
   // blob's availability.
   uint32 version = 1;
   // quorum_numbers is the list of quorum numbers that the blob is part of.
-  // All quorums must be specified (including required quorums).
+  // All quorums must be specified.
   //
-  // The following quorums are currently required:
+  // On-demand dispersal is currently limited to use the following quorums:
```
samlaf, line 16:
```suggestion
  // On-demand dispersal is currently limited to using a subset of the following quorums:
```

api/proto/common/v2/common_v2.proto
```diff
@@ -11,11 +11,13 @@ message BlobHeader {
...
   // - 0: ETH
   // - 1: EIGEN
+  // 
+  // If the dispersal is using reservations, the quorum numbers are limited to the specification on-chain for that particular reservation.
```
samlaf, line 20:
```suggestion
  // Reserved-bandwidth dispersal is free to use multiple quorums, however those must be reserved ahead of time. The quorum_numbers specified here must be a subset of the ones allowed by the on-chain reservation.
```

samlaf: Can we also point to the exact place in the contract where to look this up?

hopeyen: Like the struct interface? https://github.com/Layr-Labs/eigenda/blob/1430d56258b4e814b388e497320fd76354bfb478/contracts/src/interfaces/IPaymentVault.sol#L10

samlaf (approved):
LGTM. Thanks for the changes

PR #700, opened 2024-08-13
[2/N][zero serialization] Make Batcher operate on chunks without ser/deser

jianoaix:
## Why are these changes needed?
This eliminates both the deserialization and serialization of chunks at the Batcher.

This matters because the cost on just the serialization of chunks can be significant: e.g. a large operator took 61s to serialize the dispersal request

Such perf/cost is on the critical path of E2E flow, so by eliminating them (as well as eliminating the cost to deserialize chunks from Encoder) will contribute directly the E2E perf.

Screenshot 2024-08-13 at 2 58 56 PM

### Testing
In addition to the unit test, also tested it in Preprod, with following setup
- Original chunks are in Gob
- The 3 Nodes are running with: 1) v0.7.4; 2) v0.8.0+ with Gnark disabled (i.e. using Gob); 3) v0.8.0+ with Gnark enabled

And the Batcher was tested in these two cases:
- Batcher shipping Gob chunks to Node: chunks will be shipped as it is
- Batcher shipping Gnark chunks to Node: chunks will be converted to Gnark (since they were in Gob) and then shipped to Nodes

This shows the compatibility/safety of:
1) interaction with the ongoing Gob->Gnark migration is correct
2) the migration of frames (deserialized) -> ChunksData (zero ser/deser) itself is correct

## Checks

- [x] I've made sure the lint is passing in this PR.
- [x] I've made sure the tests are passing. Note that there might be a few flaky tests, in that case, please comment that they are not relevant.
- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests
   - [ ] This PR is not tested :(

PR #1197, opened 2025-01-31
Comapct relay chunks

cody-littley:
Instead of deserializing chunk data on the relay, keep it in binary form. This is much more space and CPU efficient.


jianoaix:
This seems the same as what we did for Batcher before: https://github.com/Layr-Labs/eigenda/pull/700
And repeating a same issue in relay. But not too late to fix.

cody-littley:
@jianoaix 

> This seems the same as what we did for Batcher before

I was able to reuse the `ChunksData` struct, and deleted my equivalent implementation. I do still need some new serialization/deserialization logic, but less than I would have otherwise needed without `ChunksData`.

The core reason the new serialization/deserialization logic is needed is because the v2 encoder uploads the proofs and coefficients in separate files. I need to be able to serialize each individually, and then parse and recombine them into `ChunksData` on the relay (without fully deserializing).

encoding/rs/bundle_builder.go
```diff
@@ -0,0 +1,33 @@
...
+	"github.com/Layr-Labs/eigenda/core"
+)
+
+// BuildChunksData creates a binary core.ChunksData object from the given proofs and coefficients.
+func BuildChunksData(
+	proofs [][]byte,
+	chunkLen int,
+	coefficients [][]byte) (*core.ChunksData, error) {
+
+	if len(proofs) != len(coefficients) {
```
jianoaix, line 14: does chunkLen need to validate?

encoding/rs/frame_coeffs.go
```diff
@@ -0,0 +1,112 @@
+package rs
+
+import (
+	"encoding/binary"
+	"fmt"
+	"github.com/Layr-Labs/eigenda/encoding"
+	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
+)
+
+// FrameCoeffs is a slice of coefficients (i.e. an encoding.Frame object without the proofs).
```
jianoaix, line 10: Where is "encoding.Frame" defined now?

relay/chunkstore/chunk_reader.go
```diff
@@ -9,19 +9,33 @@ import (
...
+
 	// GetChunkCoefficients reads a slice of frames from the chunk store. The metadata parameter
 	// should match the metadata returned by PutChunkCoefficients.
 	GetChunkCoefficients(
 		ctx context.Context,
 		blobKey corev2.BlobKey,
-		fragmentInfo *encoding.FragmentInfo) ([]*rs.Frame, error)
+		fragmentInfo *encoding.FragmentInfo) ([]rs.FrameCoeffs, error)
+
+	// GetBinaryChunkCoefficients reads a slice of frames from the chunk store, similar to GetChunkCoefficients.
```
jianoaix, line 32: Why is this called chunk coeff not frame coeff?

encoding/rs/bundle_builder.go
cody-littley: `chunkLen` is the size of the `[]fr.Element` for each chunk. It's a required piece of data that gets serialized in the bundle format. It's size is independent of `len(proofs)` and `len(coefficients)`.

encoding/rs/frame_coeffs.go
cody-littley: `encoding.Frame` is different than `rs.Frame` (renamed in this PR to `rs.FrameCoeffs`). The old naming pattern was really confusing... these objects had the same name but different data. The location where `encoding.Frame` is defined did not change. All things that used the old `rs.Frame` have been renamed to use `rs.FrameCoeffs`.

relay/chunkstore/chunk_reader.go
cody-littley: Renamed methods to use `frame` terminology.

encoding/rs/frame_coeffs.go
```diff
@@ -0,0 +1,112 @@
...
+	}
+
+	return coeffs, nil
+}
+
+// SplitSerializedFrameCoeffs splits data as serialized by SerializeFrameCoeffsSlice into a slice of byte slices.
+// Each byte slice contains the serialized data for a single FrameCoeffs object as serialized by FrameCoeffs.Serialize.
+// Also returns ElementCount, the number of elements in each FrameCoeffs object.
+func SplitSerializedFrameCoeffs(serializedData []byte) (elementCount uint32, binaryFrameCoeffs [][]byte, err error) {
+	elementCount = binary.BigEndian.Uint32(serializedData)
```
jianoaix, line 90: same here

relay/chunkstore/chunk_reader.go
bxue-l2: A Frame is a chunk + a proof

common/util/size_of.go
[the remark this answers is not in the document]
cody-littley: moved

relay/chunkstore/chunk_reader.go
```diff
@@ -9,19 +9,33 @@ import (
 	"github.com/Layr-Labs/eigenda/encoding"
 	"github.com/Layr-Labs/eigenda/encoding/rs"
 	"github.com/Layr-Labs/eigensdk-go/logging"
-	"github.com/consensys/gnark-crypto/ecc/bn254"
 )
 
 // ChunkReader reads chunks written by ChunkWriter.
 type ChunkReader interface {
-	// GetChunkProofs reads a slice of proofs from the chunk store.
-	GetChunkProofs(ctx context.Context, blobKey corev2.BlobKey) ([]*encoding.Proof, error)
```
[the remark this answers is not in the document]
cody-littley: Renamed all methods using `Chunk` instead of `Frame`.

relay/chunkstore/chunk_reader.go
```diff
@@ -9,19 +9,33 @@ import (
...
-	// GetChunkCoefficients reads a slice of frames from the chunk store. The metadata parameter
-	// should match the metadata returned by PutChunkCoefficients.
-	GetChunkCoefficients(
+
+	// GetFrameProofs reads a slice of proofs from the chunk store.
+	GetFrameProofs(ctx context.Context, blobKey corev2.BlobKey) ([]*encoding.Proof, error)
```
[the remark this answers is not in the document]
cody-littley: moved to helper functions

encoding/rs/frame_proofs.go
```diff
@@ -0,0 +1,89 @@
+package rs
+
+import (
+	"fmt"
+	"github.com/Layr-Labs/eigenda/encoding"
+	"github.com/consensys/gnark-crypto/ecc/bn254"
+)
+
+const SerializedProofLength = bn254.SizeOfG1AffineCompressed
```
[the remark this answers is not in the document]
cody-littley: Yes, when files get too big I like extracting functionality to smaller and more focused files. There is some new code here, but it is rigorously unit tested.

encoding/rs/frame_proofs.go
```diff
@@ -0,0 +1,89 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to serialize proof: %w", err)
+		}
+	}
+	return bytes, nil
+}
+
+// DeserializeFrameProof deserializes an encoding.Proof. Only the first proof is deserialized
+// from the first SerializedProofLength bytes of the input array.
+func DeserializeFrameProof(bytes []byte) (*encoding.Proof, error) {
```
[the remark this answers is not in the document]
cody-littley: Done. There was originally a reason why I allowed longer byte arrays (I was deserializing a single proof out of a large byte slice that contained many proofs), but that reason no longer exists.

encoding/rs/frame_coeffs.go
cody-littley: added

jianoaix (approved):

PR #1199, opened 2025-02-02
Fix swagger doc gen to resolve v2.BlobHeader

pschork:
Swagger was including the (v1) `core.BlobHeader` instead of the `v2.BlobHeader` in the `v2.BlobFeedResponse`

```
v2.BlobFeedResponse{
blobs	[v2.BlobInfo{
blob_key	[...]
blob_metadata	v2.BlobMetadata{
blobHeader	core.BlobHeader{
accountID	[...] <-- accountID is inside the BlobHeader object
```

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/common/v2/blob.go
+++ b/disperser/common/v2/blob.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 
 	pb "github.com/Layr-Labs/eigenda/api/grpc/disperser/v2"
-	core "github.com/Layr-Labs/eigenda/core/v2"
+	corev2 "github.com/Layr-Labs/eigenda/core/v2"
 	"github.com/Layr-Labs/eigenda/encoding"
 )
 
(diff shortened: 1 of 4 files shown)


disperser/common/v2/blob.go
```diff
@@ -71,7 +71,7 @@ func BlobStatusFromProtobuf(s pb.BlobStatus) (BlobStatus, error) {
 
 // BlobMetadata is an internal representation of a blob's metadata.
 type BlobMetadata struct {
-	BlobHeader *core.BlobHeader
+	BlobHeader *corev2.BlobHeader
```
jianoaix, line 74: ah, this is not just a doc issue, it's a real bug. Our test seems weak.

jianoaix (approved):

disperser/common/v2/blob.go
pschork: Not sure. The go compiler does the right thing with `core "github.com/Layr-Labs/eigenda/core/v2"`

PR #1204, opened 2025-02-03
Wait for internal eth client to reach block num

litt3:

- Verification calls previously assumed that the internal eth client was advanced far enough that the most recent block number > the block number in the cert
- This isn't necessarily the case, which can cause verification failures
- This PR introduces logic to wait for the internal eth client to catch up to the necessary block to perform the verification

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/config.go
+++ b/api/clients/v2/config.go
@@ -38,6 +38,13 @@ type PayloadClientConfig struct {
 	// The timeout duration for contract calls
 	ContractCallTimeout time.Duration
 
+	// BlockNumberPollInterval is how frequently to check latest block number when waiting for the internal eth client
+	// to advance to a certain block.
+	//
+	// If this is configured to be <= 0, then contract calls which require the internal eth client to have reached a
+	// certain block height will fail if the internal client is behind.
+	BlockNumberPollInterval time.Duration
+
 	// The BlobVersion to use when creating new blobs, or interpreting blob bytes.
 	//
 	// BlobVersion needs to point to a version defined in the threshold registry contract.
@@ -89,10 +96,11 @@ type PayloadDisperserConfig struct {
 // NOTE: EthRpcUrl and EigenDACertVerifierAddr do not have defined defaults. These must always be specifically configured.
 func getDefaultPayloadClientConfig() *PayloadClientConfig {
 	return &PayloadClientConfig{
-		BlobEncodingVersion:   codecs.DefaultBlobEncoding,
-		PayloadPolynomialForm: codecs.PolynomialFormEval,
-		ContractCallTimeout:   5 * time.Second,
-		BlobVersion:           0,
+		BlobEncodingVersion:     codecs.DefaultBlobEncoding,
+		PayloadPolynomialForm:   codecs.PolynomialFormEval,
+		ContractCallTimeout:     5 * time.Second,
+		BlockNumberPollInterval: 1 * time.Second,
+		BlobVersion:             0,
 	}
 }
 
@@ -119,6 +127,8 @@ func (cc *PayloadClientConfig) checkAndSetDefaults() error {
 	if cc.ContractCallTimeout == 0 {
 		cc.ContractCallTimeout = defaultConfig.ContractCallTimeout
 	}
+
+	// BlockNumberPollInterval may be 0, so don't do anything
 
 	// BlobVersion may be 0, so don't do anything
 
--- a/api/clients/v2/verification/cert_verifier.go
+++ b/api/clients/v2/verification/cert_verifier.go
@@ -40,9 +40,9 @@ type CertVerifier struct {
 	ethClient          *geth.EthClient
 	pollInterval       time.Duration
 	// storage shared between goroutines, containing the most recent block number observed by calling ethClient.BlockNumber()
-	latestBlockNumber *atomic.Uint64
+	latestBlockNumber atomic.Uint64
 	// atomic bool, so that only a single goroutine is polling the internal client with BlockNumber() calls at any given time
-	pollingActive *atomic.Bool
+	pollingActive atomic.Bool
 }
 
 var _ ICertVerifier = &CertVerifier{}
@@ -79,16 +79,11 @@ func NewCertVerifier(
 			"pollInterval", pollInterval)
 	}
 
-	latestBlockNumber := atomic.Uint64{}
-	pollingActive := atomic.Bool{}
-
 	return &CertVerifier{
 		logger:             logger,
 		certVerifierCaller: verifierCaller,
 		ethClient:          ethClient,
 		pollInterval:       pollInterval,
-		latestBlockNumber:  &latestBlockNumber,
-		pollingActive:      &pollingActive,
 	}, nil
 }
 
--- a/api/clients/v2/verification/cert_verifier.go
+++ b/api/clients/v2/verification/cert_verifier.go
@@ -3,6 +3,7 @@ package verification
 import (
 	"context"
 	"fmt"
+	"sync/atomic"
 	"time"
 
 	disperser "github.com/Layr-Labs/eigenda/api/grpc/disperser/v2"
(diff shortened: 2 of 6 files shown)


api/clients/v2/verification/cert_verifier.go
```diff
@@ -30,18 +33,31 @@ type ICertVerifier interface {
 //
 // The cert verifier contract is located at https://github.com/Layr-Labs/eigenda/blob/master/contracts/src/core/EigenDACertVerifier.sol
 type CertVerifier struct {
+	logger logging.Logger
 	// go binding around the EigenDACertVerifier ethereum contract
 	certVerifierCaller *verifierBindings.ContractEigenDACertVerifierCaller
+	ethClient          geth.EthClient
```
ethenotethan, line 39: why not reference `ethClient` by * instead?

api/clients/v2/verification/cert_verifier.go
```diff
@@ -130,3 +188,42 @@ func (cv *CertVerifier) GetNonSignerStakesAndSignature(
...
+			if err != nil {
+				return fmt.Errorf("get block number: %w", err)
+			}
+
+			if actualBlockNumber >= targetBlockNumber {
+				return nil
+			}
+
+			cv.logger.Debug(
+				"local client is behind the target block number",
```
ethenotethan, line 224: knit - can we use "reference block" instead?

api/clients/v2/verification/cert_verifier.go
```diff
@@ -120,6 +173,11 @@ func (cv *CertVerifier) GetNonSignerStakesAndSignature(
 		return nil, fmt.Errorf("convert signed batch: %w", err)
 	}
 
+	err = cv.waitForBlockNumber(ctx, signedBatch.GetHeader().GetReferenceBlockNumber())
```
ethenotethan, line 176: knit -`maybeWaitForBlockNumber`

api/clients/v2/verification/cert_verifier.go
```diff
@@ -130,3 +188,42 @@ func (cv *CertVerifier) GetNonSignerStakesAndSignature(
 
 	return &nonSignerStakesAndSignature, nil
 }
+
+// waitForBlockNumber waits until the internal eth client has advanced to a certain targetBlockNumber.
+//
+// This method will check the current block number of the internal client every CertVerifier.pollInterval duration.
+// It will return nil if the internal client advances to (or past) the targetBlockNumber. It will return an error
+// if the input context times out, or if any error occurs when checking the block number of the internal client.
+func (cv *CertVerifier) waitForBlockNumber(ctx context.Context, targetBlockNumber uint64) error {
```
ethenotethan, line 197: think you may be missing something in your context usage:
- https://golang.cafe/blog/golang-context-with-timeout-example.html

api/clients/v2/verification/cert_verifier.go
cody-littley: +1, `geth.EthClient` is a struct not an interface

api/clients/v2/verification/cert_verifier.go
```diff
@@ -30,18 +33,31 @@ type ICertVerifier interface {
...
 
 var _ ICertVerifier = &CertVerifier{}
 
 // NewCertVerifier constructs a CertVerifier
 func NewCertVerifier(
-	ethClient geth.EthClient, // the eth client, which should already be set up
-	certVerifierAddress string, // the hex address of the EigenDACertVerifier contract
+	logger logging.Logger,
+	// the eth client, which should already be set up
+	ethClient geth.EthClient,
```
cody-littley, line 49: should also be a pointer

api/clients/v2/verification/cert_verifier.go
cody-littley: nit, "nit" 😜

api/clients/v2/verification/cert_verifier.go
litt3: [342f6832](https://github.com/Layr-Labs/eigenda/pull/1204/commits/342f6832de75ed9eebd5635d43f6cd1d170e4a91)

api/clients/v2/verification/cert_verifier.go
litt3: > This function doesn't add the timeout itself, but it plays nicely if one is configured.

Right, this is the intention. @ethenotethan Is there something else you're referring to here?

api/clients/v2/config.go
```diff
@@ -120,6 +128,8 @@ func (cc *PayloadClientConfig) checkAndSetDefaults() error {
 		cc.ContractCallTimeout = defaultConfig.ContractCallTimeout
 	}
 
+	// BlockNumberPollInterval may be 0, so don't do anything
+
 	// BlobVersion may be 0, so don't do anything
```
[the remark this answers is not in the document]
litt3: Yes- there are 2 fields here, both of which are intentionally ignored (but I want it to be explicit that they weren't forgotten)

api/clients/v2/verification/cert_verifier.go
litt3: fixed

api/clients/v2/verification/cert_verifier.go
litt3: On second thought, if we're making this method public, I don't think "reference block" is appropriate. Someone could use this method in a context which doesn't have anything to do with the reference block in one of our certs.

Thoughts?

api/clients/v2/verification/cert_verifier.go
```diff
@@ -36,8 +37,12 @@ type CertVerifier struct {
 	logger logging.Logger
 	// go binding around the EigenDACertVerifier ethereum contract
 	certVerifierCaller *verifierBindings.ContractEigenDACertVerifierCaller
-	ethClient          geth.EthClient
+	ethClient          *geth.EthClient
 	pollInterval       time.Duration
+	// storage shared between goroutines, containing the most recent block number observed by calling ethClient.BlockNumber()
+	latestBlockNumber *atomic.Uint64
+	// atomic bool, so that only a single goroutine is polling the internal client with BlockNumber() calls at any given time
+	pollingActive *atomic.Bool
```
cody-littley, line 45: Nit, if you make these non-pointers, then the zero value is actually a sane instantiated object, I think.

cody-littley (approved):

api/clients/v2/verification/cert_verifier.go
litt3: Done

api/clients/v2/verification/cert_verifier.go
```diff
@@ -130,3 +188,42 @@ func (cv *CertVerifier) GetNonSignerStakesAndSignature(
...
+
+	var actualBlockNumber uint64
+	for {
+		select {
+		case <-ctx.Done():
+			return fmt.Errorf(
+				"timed out waiting for block number %d (latest block number observed was %d): %w",
+				targetBlockNumber, actualBlockNumber, ctx.Err())
+		case <-ticker.C:
+			actualBlockNumber, err := cv.ethClient.BlockNumber(ctx)
```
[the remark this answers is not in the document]
litt3: Concurrency logic added

api/clients/v2/config.go
```diff
@@ -89,10 +96,11 @@ type PayloadDisperserConfig struct {
 // NOTE: EthRpcUrl and EigenDACertVerifierAddr do not have defined defaults. These must always be specifically configured.
 func getDefaultPayloadClientConfig() *PayloadClientConfig {
```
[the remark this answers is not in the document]
litt3: Done

api/clients/v2/verification/cert_verifier.go
litt3: Thusly renamed

api/clients/v2/verification/cert_verifier.go
ethenotethan: right but where are we configuring that timeout? Shouldn't that be something isolated within the `certVerifier` itself?

ethenotethan (approved):
LGTM - we may wanna test the concurrency logic to ensure safe multi routine accesz

api/clients/v2/verification/cert_verifier.go
litt3: Since the methods on this struct accept a `ctx` parameter, isn't the default implication that the caller of the public methods will set a timeout (since it isn't stated otherwise)?

I wouldn't be strongly opposed to defining the timeout when constructing the `CertVerifier`, but I think doing so might actually increase complexity. Since this struct provides different utilities (verification, *and* waiting for a block), I think we would need to have multiple timeout values configured: one for how long we should wait for a block, and one for how long we should wait for verification.

As it currently stands, the caller of any public method is responsible for configuring a single timeout, and that timeout covers whatever stack of underlying actions need to be performed to serve the purpose of the method being called.

My question would be whether this additional degree of configuration would be beneficial to any use cases?

litt3:
> LGTM - we may wanna test the concurrency logic to ensure safe multi routine accesz

@ethenotethan Test added

PR #1207, opened 2025-02-03
CertVerifier fix

0x0aa0:
Fixes edge case when blob quorums are subset of batch quorums
- Signature check within v2 cert verification now uses the batch level quorum numbers ("signedQuorums") specified in the Attestation of the SignedBatch instead of blob level quorum numbers
- getNonSignerStakesAndSignatures now returns NonSignerStakesAndSignatures and corresponding batch level signedQuroums
- signedQuorums must now be passed to verifyCertV2() and the corresponding ZK wrapper

### DIFF
--- a/contracts/src/core/EigenDACertVerifier.sol
+++ b/contracts/src/core/EigenDACertVerifier.sol
@@ -98,11 +98,13 @@ contract EigenDACertVerifier is IEigenDACertVerifier {
      * @param batchHeader The batch header of the blob 
      * @param blobInclusionInfo The inclusion proof for the blob cert
      * @param nonSignerStakesAndSignature The nonSignerStakesAndSignature to verify the blob cert against
+     * @param signedQuorumNumbers The signed quorum numbers corresponding to the nonSignerStakesAndSignature
      */
     function verifyDACertV2(
         BatchHeaderV2 calldata batchHeader,
         BlobInclusionInfo calldata blobInclusionInfo,
-        NonSignerStakesAndSignature calldata nonSignerStakesAndSignature
+        NonSignerStakesAndSignature calldata nonSignerStakesAndSignature,
+        bytes memory signedQuorumNumbers
     ) external view {
         EigenDACertVerificationUtils._verifyDACertV2ForQuorums(
             eigenDAThresholdRegistry,
@@ -112,7 +114,8 @@ contract EigenDACertVerifier is IEigenDACertVerifier {
             blobInclusionInfo,
             nonSignerStakesAndSignature,
             securityThresholdsV2,
-            quorumNumbersRequiredV2
+            quorumNumbersRequiredV2,
+            signedQuorumNumbers
         );
     }
 
@@ -145,11 +148,13 @@ contract EigenDACertVerifier is IEigenDACertVerifier {
      * @param batchHeader The batch header of the blob 
      * @param blobInclusionInfo The inclusion proof for the blob cert
      * @param nonSignerStakesAndSignature The nonSignerStakesAndSignature to verify the blob cert against
+     * @param signedQuorumNumbers The signed quorum numbers corresponding to the nonSignerStakesAndSignature
      */
     function verifyDACertV2ForZKProof(
         BatchHeaderV2 calldata batchHeader,
         BlobInclusionInfo calldata blobInclusionInfo,
-        NonSignerStakesAndSignature calldata nonSignerStakesAndSignature
+        NonSignerStakesAndSignature calldata nonSignerStakesAndSignature,
+        bytes memory signedQuorumNumbers
     ) external view returns (bool) {
         try EigenDACertVerificationUtils.verifyDACertV2ForQuorumsExternal(
             eigenDAThresholdRegistry,
@@ -159,7 +164,8 @@ contract EigenDACertVerifier is IEigenDACertVerifier {
             blobInclusionInfo,
             nonSignerStakesAndSignature,
             securityThresholdsV2,
-            quorumNumbersRequiredV2
+            quorumNumbersRequiredV2,
+            signedQuorumNumbers
         ) {
             return true;
         } catch {
@@ -175,7 +181,7 @@ contract EigenDACertVerifier is IEigenDACertVerifier {
      */
     function getNonSignerStakesAndSignature(
         SignedBatch calldata signedBatch
-    ) external view returns (NonSignerStakesAndSignature memory) {
+    ) external view returns (NonSignerStakesAndSignature memory, bytes memory) {
         return EigenDACertVerificationUtils._getNonSignerStakesAndSignature(
             operatorStateRetriever, 
             registryCoordinator, 
--- a/contracts/src/libraries/EigenDACertVerificationUtils.sol
+++ b/contracts/src/libraries/EigenDACertVerificationUtils.sol
@@ -167,7 +167,8 @@ library EigenDACertVerificationUtils {
         BlobInclusionInfo memory blobInclusionInfo,
         NonSignerStakesAndSignature memory nonSignerStakesAndSignature,
         SecurityThresholds memory securityThresholds,
-        bytes memory requiredQuorumNumbers
+        bytes memory requiredQuorumNumbers,
+        bytes memory signedQuorumNumbers
     ) internal view {
         require(
             Merkle.verifyInclusionKeccak(
@@ -184,7 +185,7 @@ library EigenDACertVerificationUtils {
             bytes32 signatoryRecordHash
         ) = signatureVerifier.checkSignatures(
             EigenDAHasher.hashBatchHeaderV2(batchHeader),
-            blobInclusionInfo.blobCertificate.blobHeader.quorumNumbers,
+            signedQuorumNumbers,
             batchHeader.referenceBlockNumber,
             nonSignerStakesAndSignature
         );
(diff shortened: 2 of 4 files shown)


contracts/src/core/EigenDACertVerifier.sol
```diff
@@ -145,11 +148,13 @@ contract EigenDACertVerifier is IEigenDACertVerifier {
...
      * @param blobInclusionInfo The inclusion proof for the blob cert
      * @param nonSignerStakesAndSignature The nonSignerStakesAndSignature to verify the blob cert against
+     * @param signedQuorumNumbers The signed quorum numbers corresponding to the nonSignerStakesAndSignature
      */
     function verifyDACertV2ForZKProof(
         BatchHeaderV2 calldata batchHeader,
         BlobInclusionInfo calldata blobInclusionInfo,
-        NonSignerStakesAndSignature calldata nonSignerStakesAndSignature
+        NonSignerStakesAndSignature calldata nonSignerStakesAndSignature,
+        bytes memory signedQuorumNumbers
```
bxue-l2, line 157: intention is to let proxy to set those value for now, and later pulled from rollup config?

samlaf (commented):
Let's discuss during standup whether this approach is best for what we are trying to achieve. It's still not clear to me what exactly the problem is.

Feels like a problem with our DACert struct. Would prefer to modify that instead of forcing another ad-hoc parameter (batchSignedQuorums) to be passed for verification. 
image

When looking at this diagram, which information is missing, and why? Where do we get it from? e.g. how is a call to verifyDACertV2 supposed to be made now that it requires an extra `signedQuorumNumbers` param? Where is that supposed to be fetched from?

contracts/src/libraries/EigenDACertVerificationUtils.sol
```diff
@@ -184,7 +185,7 @@ library EigenDACertVerificationUtils {
             bytes32 signatoryRecordHash
         ) = signatureVerifier.checkSignatures(
             EigenDAHasher.hashBatchHeaderV2(batchHeader),
-            blobInclusionInfo.blobCertificate.blobHeader.quorumNumbers,
+            signedQuorumNumbers,
```
bxue-l2, line 188: Is it right that signedQuorumNumber is based on quorum's from a batch, which contains all quorum for all blobs.

contracts/src/core/EigenDACertVerifier.sol
bxue-l2: looks like it is not the case, the signedQuorumNumbers is actually the all quorum for all blobs a batch, I think it makes sense.

contracts/src/libraries/EigenDACertVerificationUtils.sol
bxue-l2: would be good to put a comment

contracts/src/core/EigenDACertVerifier.sol
samlaf: That's what it currently is, but it doesn't NEED to be. As we discussed with Robert the GetBlobStatus endpoint could eventually be optimized to return a different signature for each blob request. I vote to rename `signedQuorumNumbers` to `sigmaQuorumNumbers` so that the sigma name matches the sigma inside the NonSignerStakesAndSignature (sigma = signature)

contracts/src/libraries/EigenDACertVerificationUtils.sol
samlaf: Right now yes, but doesn't necessarily have to be that way. sigma and this sigmaQuorumNumbers could be a subset of the actual batch signers, as long as it contains the quorums that the blob requested. see https://github.com/Layr-Labs/eigenda/pull/1207#discussion_r1945846839

samlaf:
@0x0aa0 please address the comments and make the CI pass before re-requesting for review. it's very hard to review otherwise.

api/clients/v2/payload_disperser.go
```diff
@@ -269,7 +269,11 @@ func (pd *PayloadDisperser) buildEigenDACert(
 
 	timeoutCtx, cancel := context.WithTimeout(ctx, pd.config.ContractCallTimeout)
 	defer cancel()
-	nonSignerStakesAndSignature, err := pd.certVerifier.GetNonSignerStakesAndSignature(
+
+	// TODO (litt3): @reviewers I think this method (buildEigenDACert) is the main end user of GetNonSignerStakesAndSignature,
+	//  and we are just discarding the signedQuorumNumbers returned from GetNonSignerStakesAndSignature. IMO this points
+	//  to returning signedQuorumNumbers being the wrong API choice
+	nonSignerStakesAndSignature, _, err := pd.certVerifier.GetNonSignerStakesAndSignature(
```
samlaf, line 276: TODO: fix this

contracts/src/libraries/EigenDACertVerificationUtils.sol
```diff
@@ -210,16 +211,24 @@ library EigenDACertVerificationUtils {
...
         }
 
         require(
             BitmapUtils.isSubsetOf(
-                BitmapUtils.orderedBytesArrayToBitmap(requiredQuorumNumbers),
+                BitmapUtils.orderedBytesArrayToBitmap(blobInclusionInfo.blobCertificate.blobHeader.quorumNumbers),
                 confirmedQuorumsBitmap
             ),
-            "EigenDACertVerificationUtils._verifyDACertV2ForQuorums: required quorums are not a subset of the confirmed quorums"
+            "EigenDACertVerificationUtils._verifyDACertV2ForQuorums: blob quorums are not a subset of the confirmed quorums"
```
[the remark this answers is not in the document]
samlaf: Too big of a change right now. Let's roll without this.

contracts/src/libraries/EigenDACertVerificationUtils.sol
```diff
@@ -201,7 +202,7 @@ library EigenDACertVerificationUtils {
 
         uint256 confirmedQuorumsBitmap;
 
-        for (uint i = 0; i < blobInclusionInfo.blobCertificate.blobHeader.quorumNumbers.length; i++) {
+        for (uint i = 0; i < signedQuorumNumbers.length; i++) {
```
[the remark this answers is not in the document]
samlaf: arch says this has been solved by taking the bit subset a few lines below.

samlaf:
Action items:
1. don't return signedQuorumNumbers in GetNonSignerStakesAndSignature
2. add comments to contracts (which ones are v1, v2, shared by both)

0x0aa0:
> Action items:
> 
> 1. don't return signedQuorumNumbers in GetNonSignerStakesAndSignature
> 2. add comments to contracts (which ones are v1, v2, shared by both)

https://github.com/Layr-Labs/eigenda/pull/1207/commits/8ac09e7e445a2e1634d4f409b46fdd4835d02f11

api/clients/v2/payload_disperser.go
```diff
@@ -270,12 +270,8 @@ func (pd *PayloadDisperser) buildEigenDACert(
...
-
-	// TODO (litt3): @reviewers I think this method (buildEigenDACert) is the main end user of GetNonSignerStakesAndSignature,
-	//  and we are just discarding the signedQuorumNumbers returned from GetNonSignerStakesAndSignature. IMO this points
-	//  to returning signedQuorumNumbers being the wrong API choice
-	nonSignerStakesAndSignature, _, err := pd.certVerifier.GetNonSignerStakesAndSignature(
-		timeoutCtx, blobStatusReply.GetSignedBatch())
+	nonSignerStakesAndSignature, err := pd.certVerifier.GetNonSignerStakesAndSignature(
+		timeoutCtx, certVerifierAddress, blobStatusReply.GetSignedBatch())
 	if err != nil {
 		return nil, fmt.Errorf("get non signer stake and signature: %w", err)
```
samlaf, line 276: @litt3 we can do this in another PR but would be good for now on to start adopting some standard error construction for eth_calls like we discussed during standup this week. Otherwise our clients don't know when they see an error that it might be coming from their own rpc provider, as opposed to being a bug on our side.

contracts/src/core/EigenDACertVerifier.sol
```diff
@@ -18,12 +18,24 @@ import "../interfaces/IEigenDAStructs.sol";
...
 
+    /// @notice The EigenDAThresholdRegistry contract address
     IEigenDAThresholdRegistry public immutable eigenDAThresholdRegistry;
+
+    /// @notice The EigenDABatchMetadataStorage contract address
+    /// @dev On L1 this contract is the EigenDA Service Manager contract
     IEigenDABatchMetadataStorage public immutable eigenDABatchMetadataStorage;
+
+    /// @notice The EigenDASignatureVerifier contract address
+    /// @dev On L1 this contract is the EigenDA Service Manager contract
```
samlaf, line 29: Do these notices show up somewhere in some auto-generated documentation or something? I'm generally very much against these kind of useless comments that provide absolutely no information, other than restating the var name + type. 

@notice in my mind should provide external context. We had talked about adding information about which contracts are for V1 vs V2. You did not add this information. Why?

api/clients/v2/verification/conversion_utils.go
```diff
@@ -307,3 +300,23 @@ func BlobCommitmentsBindingToInternal(
 
 	return blobCommitment, nil
 }
+
+// QuorumNumbersUint32ToUint8 accepts an array of uint32 quorum numbers, and converts it into an array of uint8 quorum
+// numbers.
+//
+// Returns an error if any quorum number is too large to fit into a uint8
+func QuorumNumbersUint32ToUint8(inputQuorums []uint32) ([]uint8, error) {
```
samlaf, line 308: Outside the scope for this PR but @litt3 why are we doing this? Shouldn't the offchain code use uint8s everywhere? Why are we using uint32s?

samlaf (approved):

litt3 (approved):

api/clients/v2/payload_disperser.go
litt3: My assumption has been that whatever `err` returned would be informative in some way, but I haven't tested that assumption. Agree, if the error returned from the eth call isn't informative enough, we need to find a way to add context

PR #1208, opened 2025-02-04
Additional correctness tests

cody-littley:

This PR does the following:
- adds additional test case
- adds additional checks to existing tests
- cleans up code


test/v2/test_setup.go
```diff
@@ -0,0 +1,131 @@
...
+		TestDataPath:                  "~/.test-v2",
+		DisperserHostname:             "disperser-preprod-holesky.eigenda.xyz",
+		DisperserPort:                 443,
+		EthRPCURLs:                    []string{"https://ethereum-holesky-rpc.publicnode.com"},
+		BLSOperatorStateRetrieverAddr: "0x93545e3b9013CcaBc31E80898fef7569a4024C0C",
+		EigenDAServiceManagerAddr:     "0x54A03db2784E3D0aCC08344D05385d0b62d4F432",
+		EigenDACertVerifierAddress:    "0xe2C7AfB3c47B800b439b0a3d8EA40ca79759B245",
+		SubgraphURL:                   "https://subgraph.satsuma-prod.com/51caed8fa9cb/eigenlabs/eigenda-operator-state-preprod-holesky/version/v0.7.0/api",
+		SRSOrder:                      268435456,
+		SRSNumberToLoad:               2097152,
```
litt3, line 24: I *think* you can decrease `SRSNumberToLoad` to `MaxBlobSize` / 32 (since each SRS point loaded is 32 bytes)

This change would decrease spin-up time for these tests by a lot. 

@bxue-l2 Could you please check my understanding here?

litt3 (approved):

test/v2/test_setup.go
bxue-l2: Right, Client in theory can store considerably less. But we bounded it with max blob size. In V2, since number of chunks is fixed say, 4096. Then in theory, a node only needs maxBlobSize/32/4096*8 number of SRS.

cody-littley: change made, now loads much faster!

PR #1211, opened 2025-02-04
docs: add minimum symbols to proto api explanation

hopeyen:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -29,6 +29,8 @@ message DisperseBlobRequest {
   // The blob to be dispersed.
   //
   // The size of this byte array may be any size as long as it does not exceed the maximum length of 16MiB.
+  // While there is no minimum size, the blob size charged against the payment method will be rounded up to
+  // the nearest multiple of `minNumSymbols` defined by the payment vault contract (https://github.com/Layr-Labs/eigenda/blob/1430d56258b4e814b388e497320fd76354bfb478/contracts/src/payments/PaymentVaultStorage.sol#L9).
   // (In the future, the 16MiB limit may be increased, but this is not guaranteed to happen.)
```
jianoaix, line 34: Drop this line?

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -29,6 +29,8 @@ message DisperseBlobRequest {
   // The blob to be dispersed.
   //
   // The size of this byte array may be any size as long as it does not exceed the maximum length of 16MiB.
+  // While there is no minimum size, the blob size charged against the payment method will be rounded up to
```
jianoaix, line 32: Can it be 0? len(blob)>0 is required right?

jianoaix (approved):

api/proto/disperser/v2/disperser_v2.proto
hopeyen: `// (In the future, the 16MiB limit may be increased, but this is not guaranteed to happen.)`
is this not true anymore?

hopeyen: removed the line but still curious

api/proto/disperser/v2/disperser_v2.proto
hopeyen: you are right, fixed the sentence

api/proto/disperser/v2/disperser_v2.proto
jianoaix: Just not quite useful to mention a "maybe" future that doesn't have any specifics at the API.

When we actually increase it, we'll announce it loud and clear anyway (and update this documentation).

PR #1212, opened 2025-02-04
feat: metrics on meterer usage

hopeyen:

added a `disperseBlobSymbolsCharged` gauge vector metrics, to be compared against the normal disperseBlobSize (we expect `disperseBlobSymbolsCharged` to round up and be multiple of 128kb for each dispersal

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/apiserver/disperse_blob_v2.go
+++ b/disperser/apiserver/disperse_blob_v2.go
@@ -28,7 +28,7 @@ func (s *DispersalServerV2) DisperseBlob(ctx context.Context, req *pb.DisperseBl
 	if onchainState == nil {
 		return nil, api.NewErrorInternal("onchain state is nil")
 	}
-	if err := s.validateDispersalRequest(ctx, req, onchainState); err != nil {
+	if err := s.validateDispersalRequest(req, onchainState); err != nil {
 		return nil, api.NewErrorInvalidArg(fmt.Sprintf("failed to validate the request: %v", err))
 	}
 
@@ -39,8 +39,6 @@ func (s *DispersalServerV2) DisperseBlob(ctx context.Context, req *pb.DisperseBl
 
 	finishedValidation := time.Now()
 	s.metrics.reportValidateDispersalRequestLatency(finishedValidation.Sub(start))
-
-	s.metrics.reportDisperseBlobSize(len(req.GetBlob()))
 
 	blob := req.GetBlob()
 	blobHeader, err := corev2.BlobHeaderFromProtobuf(req.GetBlobHeader())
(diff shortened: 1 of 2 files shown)


disperser/apiserver/disperse_blob_v2.go
```diff
@@ -137,6 +136,7 @@ func (s *DispersalServerV2) validateDispersalRequest(
 	}
 	blob := req.GetBlob()
 	blobSize := len(blob)
+	s.metrics.reportDisperseBlobSize(blobSize)
```
ian-shim, line 139: should we export this metric only for valid requests?

disperser/apiserver/metrics_v2.go
```diff
@@ -88,6 +89,15 @@ func newAPIServerV2Metrics(registry *prometheus.Registry, metricsConfig disperse
 		[]string{},
 	)
 
+	disperseBlobMeteredBytes := promauto.With(registry).NewGaugeVec(
```
ian-shim, line 92: can we make this a counter so that we can track the increments?

disperser/apiserver/disperse_blob_v2.go
```diff
@@ -122,12 +120,13 @@ func (s *DispersalServerV2) checkPaymentMeter(ctx context.Context, req *pb.Dispe
 	if err != nil {
```
ian-shim, line 120: agree that it would be a better idea to return the symbols charges as a return value from `MeterRequest`

disperser/apiserver/metrics_v2.go
hopeyen: `dispersalBlobSize` is a gauge and not a counter, so I think this should be the same type in order for these two throughputs to compare. We can sum over the gauge over a range to look at the increments, right?

disperser/apiserver/disperse_blob_v2.go
hopeyen: I'll move this so it has the same reasoning as before. Was thinking we should record the size even if the request end up being invalid, or maybe add a metric that count different invalid categories, but probably a different issue

disperser/apiserver/disperse_blob_v2.go
hopeyen: did some minor related refactoring:)

disperser/apiserver/metrics_v2.go
ian-shim: `dispersalBlobSize` should probably also be a counter. The problem with making it a gauge is that one request will replace the previous value. Instead of this, we want subsequent requests to accumulate over one counter.

hopeyen: okay, make sense, I've updated both to counters!

ian-shim (approved):

PR #1213, opened 2025-02-05
Disable v1 blobstore dynamo collector in v2

pschork:
Disable v1 blobstore dynamo collector in v2

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests


ian-shim (approved):
Looks like we're going to share the same `Metrics` in both v1 and v2.
Maybe not in this PR, but can we have the collector implementation for v2 as well? We would need to export the blob count in QUEUED status for example

dmanc (approved):

PR #1215, opened 2025-02-05
chore(api/clients/v2): Reference prover via interface vs raw struct

ethenotethan:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


samlaf (commented):
Why is this change needed? ~~The prover is passed to the NewDisperserClietn constructor, which works fine with a nil pointer. In fact, by passing an interface in, I think you break the actual logic inside the DisperserClient (the nil check will evaluate to false) even when kzgConfig == nil~~

samlaf (approved):
I was wrong. This fix is indeed a fix. The problem is in the DisperserClient. It checks for `if c.prover == nil` which checks if the interface is nil.
But here we were passing a nil pointer, which in the constructor was being wrapped in a NON-NIL interface that pointed to the nil pointer struct. So the `if c.prover == nil` was not working.

We can merge this as a fix, but @litt3 I would prefer if we actually fix the DisperserClient. An ugly solution is to check in the constructor if prover is an interface that points to a nil struct and set the prover to be nil in those cases. A better solution is what I had asked for originally to instead not rely on `c.prover == nil` and use a proper NoProver implementation of the interface. What was the reason for this not working again?

ethenotethan:
+1 to using Noop prover

litt3:
> What was the reason for this not working again?

@samlaf It's not that the solution won't work. The `DisperserClient` needs to be fixed, it just hasn't been undertaken yet

I want to still do a deep dive exploration of alternate strategies, but I agree the Noop path sounds like the likely solution

litt3 (approved):

PR #1217, opened 2025-02-05
chore(contracts): Add binding check

ethenotethan:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/.github/workflows/test-contracts.yml
+++ b/.github/workflows/test-contracts.yml
@@ -13,7 +13,7 @@ concurrency:
   cancel-in-progress: true
 
 jobs:
-  check:
+  forge-tests:
     name: Foundry Project
     runs-on: ubuntu-22.04
     steps:
@@ -37,3 +37,23 @@ jobs:
       - name: Run snapshot
         run: forge snapshot
         working-directory: ./contracts
+
+  binding-verify:
+    name: Verify bindings are updated
+    runs-on: ubuntu-22.04
+    steps:
+      - uses: actions/checkout@v3
+        with:
+          submodules: recursive
+
+      - name: Install Foundry
+        uses: foundry-rs/foundry-toolchain@v1
+        with:
+          version: nightly
+      
+      - name: Install forge dependencies
+        run: forge install
+        working-directory: ./contracts
+
+      - name: Bindings diff check
+        run: make compile-contracts && git diff --exit-code
--- a/.github/workflows/test-contracts.yml
+++ b/.github/workflows/test-contracts.yml
@@ -46,6 +46,11 @@ jobs:
         with:
           submodules: recursive
 
+      - name: Setup Go 1.21.13
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.21.13'
+
       - name: Install Foundry
         uses: foundry-rs/foundry-toolchain@v1
         with:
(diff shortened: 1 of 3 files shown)


.github/workflows/test-contracts.yml
```diff
@@ -37,3 +37,31 @@ jobs:
       - name: Run snapshot
         run: forge snapshot
         working-directory: ./contracts
+
+  binding-verify:
```
litt3, line 41: Would this run in CI then?

If so, why have the bindings committed to the repo at all, if we generate them in CI?

litt3 (commented):
If we are going to check that the bindings match in CI,

jianoaix (approved):

jianoaix:
The PR lgtm, but i think we had some discussion before regarding if requiring the latest version of binding, if anyone remembers it, may bring the context here so we make sure this is what we decided to go

.github/workflows/test-contracts.yml
samlaf: the CI is just recreating the bindings and making sure that no new artifacts are created (by making sure git status is clean). If it fails it reminds the PR creator that they forgot to update the bindings

litt3: It's not something to address for this PR, but my point is that I don't think we should have generated files checked into git at all. CI should generate, and local builds should also generate. Tracking via git is just clutter

ethenotethan: @litt3 How would package artifacts be importable under this model? iiuc this would make EigenDA entirely self contained. Also generating them once ensures that everyone is pinned to the same code.

litt3: > How would package artifacts be importable under this model?

The files would be generated as part of the build process- so they would be importable just like they currently are, after doing a build. And if any local changes were to be made to the contracts, then the locally generated artifacts would be updated during the next build execution.

> Also generating them once ensures that everyone is pinned to the same code.

Is the concern here that the artifact generation might not be deterministic? If we *do* trust the determinism of the generation, then we are guaranteed that everyone is using the same generated artifacts, as long as they are using the same source files.

I suppose committing the generated files to git provides a check that the CI artifact generation precisely matches the artifact generation of the development machine used to commit the files to git. This doesn't seem like the most robust mechanism to test deterministic generation, though.

jianoaix: I find it useful to view generate golang code, e.g. you could see clearly what's the function signature in golang is like when making a call.

litt3: With my proposed strategy, you would still be able to view the generated golang code. Everything would be exactly the same on disk, the only thing that would change is what's committed to git

jianoaix: The use case is code viewing on github

ethenotethan: @litt3 ideally these artifacts can be imported from a third-party go project without having to submodule and build our monorepo. Also different users could be on alternative versions of abigen which could cause differing local views. OP generates bindings globally for their monorepo which make importing and using it very seamless - e.g we leverage [parts](https://github.com/Layr-Labs/eigenda-proxy/blob/main/e2e/optimism_test.go) of their testing framework in proxy E2E tests.

litt3 (approved):

0x0aa0 (approved):

PR #1235, opened 2025-02-06
Split out blob dispersal from status polling in `PayloadDisperser`

litt3:

- For the sake of benchmarking, it would be very useful to be able to have a method to disperse a blob, without forcing the caller to wait for the blob to be complete
- This functionality could also feasibly be useful to an end user, if they are using the `PayloadDisperser` directly
- This PR splits apart the existing `SendPayload` method into two parts:
    - `SendDispersalRequest`, which calls the `DisperseBlob` rpc, and returns immediately with the response
    - `SendPayload`, which leverages `SendDispersalRequest`, and then polls for blob status until the blob is COMPLETE

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


litt3:
Might redo at a later date, but this isn't urgent, so closing.

PR #1236, opened 2025-02-06
Encoder API robustness audit/fix

jianoaix:
Enhance Encoder robustness

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/encoder/server_v2.go
+++ b/disperser/encoder/server_v2.go
@@ -2,13 +2,13 @@ package encoder
 
 import (
 	"context"
-	"errors"
 	"fmt"
 	"log"
 	"net"
 	"sync"
 	"time"
 
+	"github.com/Layr-Labs/eigenda/api"
 	"github.com/Layr-Labs/eigenda/common/healthcheck"
 	corev2 "github.com/Layr-Labs/eigenda/core/v2"
 	"github.com/Layr-Labs/eigenda/disperser"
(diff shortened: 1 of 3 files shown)


disperser/encoder/server_v2.go
```diff
@@ -224,11 +223,18 @@ func (s *EncoderServerV2) validateAndParseRequest(req *pb.EncodeBlobRequest) (co
...
+	if req.EncodingParams.ChunkLength&(req.EncodingParams.ChunkLength-1) != 0 {
+		return blobKey, params, status.Error(codes.InvalidArgument, "chunk length must be power of 2")
+	}
 
 	if req.EncodingParams.NumChunks == 0 {
 		return blobKey, params, status.Error(codes.InvalidArgument, "number of chunks must be greater than zero")
 	}
 
+	if req.BlobSize == 0 || uint64(encoding.GetBlobLength(uint(req.BlobSize))) > req.EncodingParams.ChunkLength*req.EncodingParams.NumChunks {
+		return blobKey, params, status.Error(codes.InvalidArgument, "blob size is invalid")
```
dmanc, line 235: Seems like all of these return invalid argument, should we just wrap the upper error instead?

dmanc (approved):

disperser/encoder/server_v2.go
jianoaix: yea. I'm thinking a proposal to improve error handling across the codebase. It's currently vague, leaky (like this rate limiting error in Encoder, or wrong (like the above GetBlobStatus); these should be addressed comprehensively.

PR #1242, opened 2025-02-07
A round of misc improvements to dataapi

jianoaix:
- Make handlers naming consistent (not mixed with `FetchXX` and `FetchXXHandler`)
- Reduce handlers' naming redundancy (`FetchXX` is sufficient, the `Handler` suffix is unnecessary)
- Make latency computing a bit cleaner (`f * 1000` is ugly)
- Account the latency better (account the success request latency -- failed requests, like invalid args, will have a super low latency and skewed the view)
- API path naming fix: `operators/nodeinfo` -> `operators/node-info` which is more conventional 
- API path naming improvement: `operators/reachablity` -> `operators/liveness`
- Remove junk code (junk fields in `Metric`)

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/dataapi/v2/server_v2.go
+++ b/disperser/dataapi/v2/server_v2.go
@@ -164,10 +164,6 @@ type (
 
 	Metric struct {
 		Throughput float64 `json:"throughput"`
-		CostInGas  float64 `json:"cost_in_gas"`
-		// deprecated: use TotalStakePerQuorum instead. Remove when the frontend is updated.
-		TotalStake          *big.Int                   `json:"total_stake"`
-		TotalStakePerQuorum map[core.QuorumID]*big.Int `json:"total_stake_per_quorum"`
 	}
 
 	Throughput struct {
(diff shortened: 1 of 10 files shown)


jianoaix:
@merli-eigenlabs please note the small changes in API path / naming.

disperser/dataapi/v2/server_v2.go
```diff
@@ -164,10 +163,6 @@ type (
 
 	Metric struct {
 		Throughput float64 `json:"throughput"`
```
dmanc, line 165: Are we still using this in dataapi v2?

disperser/dataapi/v2/server_v2_test.go
```diff
@@ -1397,9 +1394,9 @@ func TestCheckOperatorsReachability(t *testing.T) {
 	operatorId := "0xa96bfb4a7ca981ad365220f336dc5a3de0816ebd5130b79bbc85aca94bc9b6ab"
 	mockSubgraphApi.On("QueryOperatorInfoByOperatorIdAtBlockNumber").Return(operatorInfoV2, nil)
 
-	r.GET("/v2/operators/reachability", testDataApiServerV2.CheckOperatorsReachability)
+	r.GET("/v2/operators/liveness", testDataApiServerV2.CheckOperatorsLiveness)
```
pschork, line 1397: Need to update node v2 reachability checkPath
https://github.com/Layr-Labs/eigenda/blob/master/node/node.go#L52

disperser/dataapi/v2/server_v2.go
jianoaix: Yes, it's returned

disperser/dataapi/v2/server_v2_test.go
jianoaix: Done

disperser/dataapi/v2/server_v2.go
dmanc: How is the throughput being computed for V2?

jianoaix: same as v1, prometheus

dmanc: Ok it seems we should update the query it's using since it looks like it's just returning the v1 throughput.

node/flags/flags.go
```diff
@@ -294,45 +294,6 @@ var (
 		Value:    units.GiB, // intentionally large for the time being
 	}
 
-	// Test only, DO NOT USE the following flags in production
```
jianoaix, line 297: This is a drop-by fix, as we happen to need the `OverrideStoreDurationBlocksFlag` below.

`DO NOT USE the following flags` is not correct as other production flags were added it below

Fixed it, and made it even more clear about the test flag section

disperser/dataapi/v2/server_v2.go
jianoaix: Good point.

pschork (approved):

PR #1243, opened 2025-02-07
test: make reservation test less flaky (due to time)

hopeyen:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/meterer/meterer_test.go
+++ b/core/meterer/meterer_test.go
@@ -133,9 +133,9 @@ func setup(_ *testing.M) {
 	accountID1 = crypto.PubkeyToAddress(privateKey1.PublicKey)
 	accountID2 = crypto.PubkeyToAddress(privateKey2.PublicKey)
 	accountID3 = crypto.PubkeyToAddress(privateKey3.PublicKey)
-	account1Reservations = &core.ReservedPayment{SymbolsPerSecond: 100, StartTimestamp: now - 120, EndTimestamp: now + 180, QuorumSplits: []byte{50, 50}, QuorumNumbers: []uint8{0, 1}}
-	account2Reservations = &core.ReservedPayment{SymbolsPerSecond: 200, StartTimestamp: now - 120, EndTimestamp: now + 180, QuorumSplits: []byte{30, 70}, QuorumNumbers: []uint8{0, 1}}
-	account3Reservations = &core.ReservedPayment{SymbolsPerSecond: 200, StartTimestamp: now + 120, EndTimestamp: now + 180, QuorumSplits: []byte{30, 70}, QuorumNumbers: []uint8{0, 1}}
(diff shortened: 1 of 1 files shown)


core/meterer/meterer.go
```diff
@@ -73,7 +73,7 @@ func (m *Meterer) Start(ctx context.Context) {
 
 // MeterRequest validates a blob header and adds it to the meterer's state
 // TODO: return error if there's a rejection (with reasoning) or internal error (should be very rare)
-func (m *Meterer) MeterRequest(ctx context.Context, header core.PaymentMetadata, numSymbols uint, quorumNumbers []uint8) (uint32, error) {
+func (m *Meterer) MeterRequest(ctx context.Context, header core.PaymentMetadata, numSymbols uint, quorumNumbers []uint8, receivedAt time.Time) (uint32, error) {
```
litt3, line 76: Could you explain why these changes fix the error I saw [here](https://github.com/Layr-Labs/eigenda/actions/runs/13549489206/job/37869380020?pr=1328)?

```
panic: interface conversion: types.AttributeValue is nil, not *types.AttributeValueMemberS [recovered]
panic: interface conversion: types.AttributeValue is nil, not *types.AttributeValueMemberS
```

I'm not understanding how passing a "received time" down into these methods is related?

hopeyen: My reasoning for why these tests are flaky is because the behavior largely depends on time. Based on how these tests span across time, they might be counted into different reservation intervals than we expected it to. With the received time passed in instead of having meterer grab a time within itself, the behavior of meterer is more predictable. 
For the error you shared, I believe it is because the meterer stored records into the previous reservation period of the current one, so no item was found, thus `nil`
Does that sound reasonable?

litt3: > I believe it is because the meterer stored records into the previous reservation period of the current one, so no item was found, thus nil

If this is the case, isn't it true that some misalignment with time in a production environment could cause a panic? It seems like there needs to be an additional check somewhere so that panic isn't possible, regardless of what timestamp is passed in

litt3: Based on my reading of the code, it doesn't appear that these meter requests are being buffered up or anything. Is that correct?

If it's true that requests aren't being buffered, it doesn't seem like the time discrepancy between the previous implementation and the new implementation would be all that different. Maybe a few milliseconds at most? It seems like there might be a deeper issue lurking here (I'm not confident that's the case for sure)

hopeyen: I think the panic comes from the test itself and the production code uses a [helper](https://github.com/Layr-Labs/eigenda/blob/99792bf38b035b3d67d82d56b67b4950f4a5bdb1/core/meterer/offchain_store.go#L252) method that should be more safe. We can't really call that function within this meterer test though since offchain_store isn't exposed.

litt3 (approved):
Changes LGTM, but I suggest getting an additional review, since I don't feel well equipped to evaluate the implications of this change

ian-shim (approved):

PR #1247, opened 2025-02-08
Report hashing error from v1 batcher dispatcher

ian-shim:
Error should be propagated to the channel before returning

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/batcher/grpc/dispatcher.go
+++ b/disperser/batcher/grpc/dispatcher.go
@@ -3,6 +3,7 @@ package dispatcher
 import (
 	"context"
 	"errors"
+	"fmt"
 	"time"
 
 	commonpb "github.com/Layr-Labs/eigenda/api/grpc/common"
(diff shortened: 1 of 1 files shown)


disperser/batcher/grpc/dispatcher.go
```diff
@@ -55,6 +56,13 @@ func (c *dispatcher) sendAllChunks(ctx context.Context, state *core.IndexedOpera
 			hasAnyBundles := false
 			batchHeaderHash, err := batchHeader.GetBatchHeaderHash()
 			if err != nil {
+				update <- core.SigningMessage{
```
jianoaix, line 59: Did we hit this issue recently?

jianoaix (approved):

disperser/batcher/grpc/dispatcher.go
ian-shim: I don't think so. Still good to have it for completeness

PR #1252, opened 2025-02-10
Blob feed iteration with dynamo pagination

jianoaix:
The blob feed needs to use pagination to get all blob, because Dynamo has a 1MB limit on the response size, so if there are more blobs to fetch, it won't be able to get them all.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/common/v2/blobstore/dynamo_metadata_store.go
+++ b/disperser/common/v2/blobstore/dynamo_metadata_store.go
@@ -293,7 +293,7 @@ func (s *BlobMetadataStore) GetBlobMetadataByStatus(ctx context.Context, status 
 // queryBucketBlobMetadata returns blobs (as metadata) within range [startKey, endKey] from a single bucket.
 // Results are ordered by <RequestedAt, Bobkey> in ascending order.
 //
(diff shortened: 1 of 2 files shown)


disperser/common/v2/blobstore/dynamo_metadata_store.go
```diff
@@ -290,35 +290,61 @@ func (s *BlobMetadataStore) GetBlobMetadataByStatus(ctx context.Context, status
 	return metadata, nil
 }
 
-// queryBucketBlobMetadata queries a single bucket for blob metadata within the given key range.
+// queryBucketBlobMetadata returns blobs (as metadata) within range [startKey, endKey] from a single bucket.
+// Results are ordered by <RequestedAt, Bobkey> in ascending order.
+//
+// The function handles DynamoDB's 1MB response size limitation by performing multiple queries if necessary.
```
dmanc, line 296: Isn't `GetBlobMetadataByRequestedAt` already doing pagination?

dmanc (approved):

PR #1255, opened 2025-02-10
Refactor the dataapi server v2 file into modular files

jianoaix:
The current `server_v2.go` is getting too large. It can be cleanly organized into:
- blobs
- batches
- operators
- metrics

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


disperser/dataapi/v2/batches.go
```diff
@@ -0,0 +1,148 @@
...
+//	@Param		end			query		string	false	"Fetch batches up to the end time (ISO 8601 format: 2006-01-02T15:04:05Z) [default: now]"
+//	@Param		interval	query		int		false	"Fetch batches starting from an interval (in seconds) before the end time [default: 3600]"
+//	@Param		limit		query		int		false	"The maximum number of batches to fetch. System max (1000) if limit <= 0 [default: 20; max: 1000]"
+//	@Success	200			{object}	BatchFeedResponse
+//	@Failure	400			{object}	ErrorResponse	"error: Bad request"
+//	@Failure	404			{object}	ErrorResponse	"error: Not found"
+//	@Failure	500			{object}	ErrorResponse	"error: Server error"
+//	@Router		/batches/feed [get]
+func (s *ServerV2) FetchBatchFeed(c *gin.Context) {
+	handlerStart := time.Now()
```
dmanc, line 29: Since these are handlers could we put them all under a `handlers` directory or do you prefer keeping it at top level.

dmanc (approved):

disperser/dataapi/v2/batches.go
jianoaix: Yea I considered it, just feel the nesting is quite deep already. We may see if there are more other files coming up under `v2`, and if so it may make more sense to group these handlers in under `handlers`.

jianoaix: AI's comments:
dataapi handlers

jianoaix: and:
grouping

PR #1257, opened 2025-02-11
Force consistent read from blob metadata store

ian-shim:
Force consistent read from v2 blob metadata store

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


common/aws/dynamodb/client.go
```diff
@@ -353,13 +362,15 @@ func (c *client) QueryIndexWithPagination(ctx context.Context, tableName string,
 			KeyConditionExpression:    aws.String(keyCondition),
 			ExpressionAttributeValues: expAttributeValues,
 			Limit:                     &limit,
+			ConsistentRead:            aws.Bool(consistentRead),
```
dmanc, line 365: Are we able to do consistent reads on any GSI?

dmanc: hmm says we can't https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/HowItWorks.ReadConsistency.html

ian-shim: Nope. Need to come up with another approach

PR #1260, opened 2025-02-11
Load generator docker

cody-littley:

Set up new load generation utility to run in docker


test/v2/client/test_client_setup.go
```diff
@@ -24,6 +23,13 @@ var (
 
 const (
 	PreprodEnv = "../config/environment/preprod.json"
+
+	//G1URL         = "https://eigenda.s3.amazonaws.com/srs/g1.point"
+	//G2URL         = "https://eigenda.s3.amazonaws.com/srs/g2.point"
+	//G2PowerOf2URL = "https://eigenda.s3.amazonaws.com/srs/g2.point.powerOf2"
+	G1URL         = "https://srs-mainnet.s3.amazonaws.com/kzg/g1.point"
```
dmanc, line 30: Should we make `https://srs-mainnet.s3.amazonaws.com` a configuration value?

test/v2/client/test_client_setup.go
```diff
@@ -3,14 +3,13 @@ package client
 import (
 	"encoding/json"
 	"fmt"
+	"github.com/Layr-Labs/eigenda/common"
+	"github.com/Layr-Labs/eigensdk-go/logging"
```
dmanc, line 7: nit: Does your IDE use a different import ordering scheme?

test/v2/client/test_client_setup.go
```diff
@@ -24,6 +23,13 @@ var (
 
 const (
 	PreprodEnv = "../config/environment/preprod.json"
+
+	//G1URL         = "https://eigenda.s3.amazonaws.com/srs/g1.point"
```
dmanc, line 27: nit: remove comments

dmanc (dismissed):

test/v2/client/test_client_setup.go
cody-littley: deleted

test/v2/client/test_client_setup.go
cody-littley: I'm not sure this makes sense as a config value. The purpose of these is to download the SRS files on a developer laptop the first time they run the test suite. Once those files are present, these constants aren't used for anything. So in kubernetes where we are using a persistent volume, these files are ignored.

test/v2/client/test_client_setup.go
cody-littley: I think it's just ordering them alphabetically. Is there a different schema we should be using?

dmanc: https://pkg.go.dev/golang.org/x/tools/cmd/goimports

test/v2/client/test_client_setup.go
dmanc: Sounds good, I don't think this URL will change anyway.

test/v2/client/test_client_setup.go
cody-littley: My IDE uses the `gofmt` standards, I'm pretty sure. The difference here is that I deleted the newline between `"testing"` and `"github.com/Layr-Labs/eigenda/common"`, which caused them to be ordered alphabetically with everything else. If we want to keep github dependencies in a separate block, I'm fairly sure it requires hand maintenance.

cody-littley: reformatted

litt3 (approved):

PR #1261, opened 2025-02-12
Reset cursors and deduplicate blobs in controller

ian-shim:
- `GetBlobMetadataByStatusPaginated`: Cursor is reset to nil when all records have been enumerated. This will ensure that any blobs that have not been enumerated will be eventually picked up (in controller's encoding manager and dispatcher)
- `EncodingManager` and `Dispatcher` keep track of blobs in flight and deduplicate them
- `Dispatcher` transitions blobs to `GatheringSignatures` status and creates an empty attestation object before sending payloads to DA network. These blobs are finalized to `Complete` or `Failed` status after the dispersal is complete (has received all signatures or timed out)
- Enable consistent read on blob cert retrieval 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/proto/disperser/v2/disperser_v2.proto
+++ b/api/proto/disperser/v2/disperser_v2.proto
@@ -157,8 +157,6 @@ enum BlobStatus {
   // and in doing so requesting that the validators sign to acknowledge receipt of the blob.
   // Requests that timeout or receive errors are resubmitted to DA nodes for some period of time set by the disperser,
   // after which the BlobStatus becomes COMPLETE.
-  //
-  // Note: this status is not currently implemented, and is a placeholder for future functionality.
   GATHERING_SIGNATURES = 3;
 
   // COMPLETE means the blob has been dispersed to DA nodes, and the GATHERING_SIGNATURES period of time has completed.
--- a/disperser/controller/dispatcher.go
+++ b/disperser/controller/dispatcher.go
@@ -23,6 +23,8 @@ import (
 
 var errNoBlobsToDispatch = errors.New("no blobs to dispatch")
 
+type BlobCallback func(blobKey corev2.BlobKey) error
+
 type DispatcherConfig struct {
 	PullInterval time.Duration
 
@@ -45,6 +47,12 @@ type Dispatcher struct {
 	metrics           *dispatcherMetrics
 
 	cursor *blobstore.StatusIndexCursor
+	// beforeDispatch function is called before dispatching a blob
+	beforeDispatch BlobCallback
+	// blobSet keeps track of blobs that are being dispatched
+	// This is used to deduplicate blobs to prevent the same blob from being dispatched multiple times
+	// Blobs are removed from the queue when they are in a terminal state (Complete or Failed)
+	blobSet BlobSet
 }
 
 type batchData struct {
@@ -64,6 +72,8 @@ func NewDispatcher(
 	nodeClientManager NodeClientManager,
 	logger logging.Logger,
 	registry *prometheus.Registry,
+	beforeDispatch func(blobKey corev2.BlobKey) error,
+	blobSet BlobSet,
 ) (*Dispatcher, error) {
 	if config == nil {
 		return nil, errors.New("config is required")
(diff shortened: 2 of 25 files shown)


api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -157,8 +157,6 @@ enum BlobStatus {
   // and in doing so requesting that the validators sign to acknowledge receipt of the blob.
   // Requests that timeout or receive errors are resubmitted to DA nodes for some period of time set by the disperser,
   // after which the BlobStatus becomes COMPLETE.
-  //
-  // Note: this status is not currently implemented, and is a placeholder for future functionality.
   GATHERING_SIGNATURES = 3;
```
cody-littley, line 160: Is this now implemented?

disperser/controller/blob_queue.go
```diff
@@ -0,0 +1,54 @@
+package controller
+
+import (
+	"sync"
+
+	v2 "github.com/Layr-Labs/eigenda/core/v2"
+)
+
+type BlobQueue interface {
```
cody-littley, line 9: Nit, can you add a one or two sentence godoc to this interface to describe its purpose?

cody-littley: Perhaps `BlobSet` would be a better name? This struct doesn't implement a queue.

disperser/controller/dispatcher.go
cody-littley, line 1: Thinking through potential interleavings, and I'm concerned about the following race condition.

# Race Condition

The following two methods can be executing concurrently:
- `HandleBatch()`: this method gets blobs from dynamo, and may accidentally fetch duplicates. Duplicate blobs are detected and discarded.
- `HandleSignatures()`: this method looks at new signatures. If enough time has passed, it will update the blob status and remove the blob from the ones we are tracking. It also removes the blob from the set used to detect duplicates.

Scenario:
- `HandleBatch()` reads blob `X` from dynamo. `X` is not a duplicate, and so the dispatcher begins tracking blob `X`.
- Later on, `HandleBatch()` and `HandleSignatures()` are being processed in parallel.
    - `HandleBatch()` reads blob `X`. It has not yet discarded `X` as a duplicate.
    - `HandleSignatures()` identifies blob `X` as having become complete, and removes blob `X` from the set used to detect duplicates.
    - `HandleBatch()` checks to see if `X` is in the duplicate set, but does not see it. It starts tracking `X` again.

# Possible Fixes

## Option 1: fewer threads

Instead of calling `HandleSignatures()` in a gouroutine, call it directly in the dispatcher's main loop. It may be the case that this parallelism is not strictly necessary.

## Option 2: one thread manages the set tracking duplicates

- only the `HandleBatch()` goroutine is allowed to make updates to set tracking duplicates
- when `HandleSignatures()` decides that a blob is complete/failed, instead of removing it from the set tracking duplicates, it puts the blob key into a channel. It is important to do this AFTER the blob has been marked as complete/failed in dynamo.
- Every time `HandleBatch()` runs, it removes blob keys from the channel, and removes them from the duplciate set. It is important that it does this BEFORE it queries dynamo.
    - Every blob removed from the set tracking duplicates is gauranteed to be marked as complete/failed in dynamo.

ian-shim: Good point. 

`HandleSignatures` first moves it to `GatheringSignatures` state before spending many seconds waiting for signatures, after which it moves blob to `Complete/Failed`, after which it removes the blob from the duplicate set. 
The dedup logic runs on pure CPU, no IO. For this to interfere with `HandleSignatures` like the scenario, this dedup logic needs to take more than the entire time `HandleSignatures` takes (10-30s). 

I don't think option 1 is reasonable given the e2e latency requirements. I don't think option 2 resolves this race condition either, actually i think it makes this problem worse because the finished blob isn't removed from the duplicate set as soon as possible. 

If this race condition ever happens, the worst thing is a blob gets dispersed twice (included 2 batches), which is  manageable. I'd rather keep this logic simple if there aren't other concerns. Wdyt?

api/proto/disperser/v2/disperser_v2.proto
ian-shim: It moves a blob to this state as soon as it is being sent to the DA network. It doesn't periodically update the attestation results though. 
cc @litt3

disperser/controller/blob_queue.go
ian-shim: will update

cody-littley (approved):

PR #1266, opened 2025-02-12
Provide signer and nonsigner operator IDs in blob attestation-info API

jianoaix:
- The clients will want to work with operator IDs of nonsigners, instead of their pubkeys.
- The clients will also want to know the signers of a blob (not  just nonsigners)

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests


jianoaix:
`codeql-scanning` failure is not relevant to this PR (it's separately fixed: https://github.com/Layr-Labs/eigenda/pull/1267)

dmanc (approved):

PR #1270, opened 2025-02-13
Fix overflow on node block stale measure & store duration blocks override values when test-mode is enabled

pschork:
If `NODE_ENABLE_TEST_MODE=true` the node will evaluate `NODE_OVERRIDE_BLOCK_STALE_MEASURE` and `NODE_OVERRIDE_STORE_DURATION_BLOCKS` which default to -1 (disabled) and get stored in Config as `int64`. 

Before evaluating if the overrides should be applied, the override values were cast to `uint32` to match return type of `tx.GetBlockStaleMeasure`. This resulted in the default value -1 (disabled) being cast to `18446744073709551615`.

This change updates the flag type to `Uint64Flag` with default value of 0

~~This change also adds the override values to the grpc `NodeInfo` response if overrides are enabled. This will allow us to observe if any nodes are explicitly load shedding / withholding data via overrides.~~

This bug explains why, after overriding my preprod node's `NODE_OVERRIDE_STORE_DURATION_BLOCKS=10`, my disk consumption did not plateau - since the `NODE_OVERRIDE_BLOCK_STALE_MEASURE` was not set and was defaulting to `18446744073709551615` blocks.

image

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


cody-littley (dismissed):

api/proto/validator/node_v2.proto
```diff
@@ -107,4 +107,8 @@ message GetNodeInfoReply {
   uint32 num_cpu = 4;
   // The amount of memory on the node in bytes.
   uint64 mem_bytes = 5;
+  // The store duration blocks value if overridden.
```
jianoaix, line 110: I don't think it's a good idea to add test flags to a production API

pschork: @jianoaix Can you elaborate on your concern? These flags are already displayed in `node --help`, they are not secret. The grpc reply will only include these values if the values are actively overridden. My motivation to add this was for us to be able to observe if an operator is overriding the protocol's settings. Why is that not a good idea?

jianoaix: Such test flag should never be used in prod, it's very odd to surface such flags from a public production API.

pschork: I agree with you, but my conclusion is different. From my perspective we have no idea if any operator *is* doing this in prod but they could be. Exposing this via NodeInfo allows us to observe it. I don't think that exposing this in NodeInfo advertises this any more than current `node --help` exposes this.

pschork: Gonna drop the overrides from the proto. After getting some additional feedback, I think the more robust approach here would be to completely remove test mode + associated flags from "release" builds.

jianoaix: Users can run any code they want, even if you remove these flags it won't prevent it, release builds or not.
This API isn't for monitor whether they have used test flags or not. The point is also not about whether it advertises the test flags more than cmd line -- it's just bad to randomly dump test stuff to prod api.

pschork: Overflow bugs are bad. Violating the storage protocol is bad. Exposing operators who violate the storage protocol via an observability API isn't bad, nor random, it's just knowledge. We've chosen to be blind to this knowledge because extending an observability API with this information would pollute the pristine landscape of the prod API with test stuff. And besides, people can run any code they want, so it's turtles all the way down.

ian-shim (approved):

PR #1272, opened 2025-02-13
Handle edge case with negative request size

cody-littley:

Handle an edge case with relay rate limiting where the requester sends a request with a negative size.


litt3 (approved):

relay/limiter/chunk_rate_limiter.go
```diff
@@ -149,15 +149,15 @@ func (l *ChunkRateLimiter) FinishGetChunkOperation(requesterID string) {
...
+func (l *ChunkRateLimiter) RequestGetChunkBandwidth(now time.Time, requesterID string, bytes uint32) error {
 	if l == nil {
 		// If the rate limiter is nil, do not enforce rate limits.
 		return nil
 	}
 
 	// no lock needed here, as the bandwidth limiters themselves are thread-safe
 
-	allowed := l.globalBandwidthLimiter.AllowN(now, bytes)
+	allowed := l.globalBandwidthLimiter.AllowN(now, int(bytes))
```
anupsv, line 160: interesting, is there a case where the bandwidth limiter behaves different with negative values ?

cody-littley: The underlying utility I'm using is a `rate.Limter` from `golang.org/x/time`. Passing it negative values will refill the rate limit, so in theory somebody doing this could artificially increase what the rate limiter is willing to permit.

ian-shim (approved):

PR #1275, opened 2025-02-14
Disable auto IP discovery if not all ports are configured

ian-shim:
If it auto registers the socket with missing ports, it can render the node unreachable. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -349,12 +349,33 @@ func (n *Node) Start(ctx context.Context) error {
 
 	n.CurrentSocket = socket
 	// Start the Node IP updater only if the PUBLIC_IP_PROVIDER is greater than 0.
-	if n.Config.PubIPCheckInterval > 0 {
+	if n.Config.PubIPCheckInterval > 0 && n.allPortsConfigured() {
 		go n.checkRegisteredNodeIpOnChain(ctx)
 		go n.checkCurrentNodeIp(ctx)
 	}
 
 	return nil
+}
+
+// allPortsConfigured checks if all the ports are configured in the config.
+func (n *Node) allPortsConfigured() bool {
+	if err := core.ValidatePort(n.Config.DispersalPort); err != nil {
+		return false
+	}
+
+	if err := core.ValidatePort(n.Config.V2DispersalPort); err != nil {
+		return false
+	}
+
+	if err := core.ValidatePort(n.Config.RetrievalPort); err != nil {
+		return false
+	}
+
+	if err := core.ValidatePort(n.Config.V2RetrievalPort); err != nil {
+		return false
+	}
+
+	return true
 }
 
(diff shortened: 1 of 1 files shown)


pschork (approved):

node/node.go
```diff
@@ -349,14 +349,35 @@ func (n *Node) Start(ctx context.Context) error {
 
 	n.CurrentSocket = socket
 	// Start the Node IP updater only if the PUBLIC_IP_PROVIDER is greater than 0.
-	if n.Config.PubIPCheckInterval > 0 {
+	if n.Config.PubIPCheckInterval > 0 && n.allPortsConfigured() {
```
jianoaix, line 352: It seems unnecessary, configs is validated already https://github.com/Layr-Labs/eigenda/blob/master/node/config.go#L250-L274

ian-shim: This makes sure the feature is disabled if its only running v1 and missing v2 ports for example

jianoaix: Then the condition is `if v2Enabled {...}` ?

ian-shim: That would cover most of the cases and is probably fine. Checking the ports is the closest condition we can check here imo, for example, this allows the feature if node is only running v1 but does have the v2 ports configured.

jianoaix: why v1 should disable this feature again?

pschork: Its possible to run the node in `v1-only` mode without defining v2 ports.

The other concern is chance of split brain problem when running 2 nodes in isolation mode on 2 separate hosts (potentially in different subnets/AZs) the `.env` port configs could diverge and/or the egress IP could be different - then both nodes start flip flopping with different socket updates.

So the idea is to only enable IP discovery updates when the node is in v1-and-v2 runtime mode

pschork: > Then the condition is `if v2Enabled {...}` ?

`if n.Config.v1Enabled && n.Config.v2Enabled {...}`

jianoaix: > Its possible to run the node in `v1-only` mode without defining v2 ports.
This sounds different than "node is only running v1 but does have the v2 ports"

> 
> The other concern is chance of split brain problem when running 2 nodes in isolation mode on 2 separate hosts (potentially in different subnets/AZs) the `.env` port configs could diverge and/or the egress IP could be different - then both nodes start flip flopping with different socket updates.
> 
> So the idea is to only enable IP discovery updates when the node is in v1-and-v2 runtime mode

Even it's running v1 and v2 together, they can still run two machines that both have v1 and v2 running, v1-and-v2 mode doesn't seem safe as well

jianoaix (approved):

pschork (approved):

PR #1286, opened 2025-02-18
Make blob feed API into two-way APIs

jianoaix:
To make it easy and reliable to traverse the feed backward and forward in time

This PR changes the APIs to support this.

Note: the existing API is forward iteration, so `FetchBlobFeedForward` works here already; but `FetchBlobFeedBackward` will need a change in DB layer (in a followup later)

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/dataapi/v2/blobs.go
+++ b/disperser/dataapi/v2/blobs.go
@@ -10,65 +10,51 @@ import (
 
 	"github.com/Layr-Labs/eigenda/core"
 	corev2 "github.com/Layr-Labs/eigenda/core/v2"
+	v2 "github.com/Layr-Labs/eigenda/disperser/common/v2"
 	"github.com/Layr-Labs/eigenda/disperser/common/v2/blobstore"
 	"github.com/Layr-Labs/eigenda/disperser/dataapi"
 	"github.com/gin-gonic/gin"
 )
 
-// FetchBlobFeed godoc
+// FetchBlobFeedForward godoc
 //
-//	@Summary	Fetch blob feed
-//	@Tags		Blobs
-//	@Produce	json
-//	@Param		end					query		string	false	"Fetch blobs up to the end time (ISO 8601 format: 2006-01-02T15:04:05Z) [default: now]"
-//	@Param		interval			query		int		false	"Fetch blobs starting from an interval (in seconds) before the end time [default: 3600]"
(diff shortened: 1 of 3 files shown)


disperser/dataapi/v2/blobs.go
```diff
@@ -10,129 +10,201 @@ import (
...
+	untilCursor := blobstore.BlobFeedCursor{
+		RequestedAt: uint64(untilTime.UnixNano()),
 	}
-	c.Writer.Header().Set(cacheControlParam, fmt.Sprintf("max-age=%d", maxFeedBlobAge))
-	s.metrics.IncrementSuccessfulRequestNum("FetchBlobFeed")
-	s.metrics.ObserveLatency("FetchBlobFeed", time.Since(handlerStart))
-	c.JSON(http.StatusOK, response)
+
+	// TODO(jianxiao): this is just a placeholder as GetBlobMetadataByRequested is doing forward retrieval
+	blobs, nextCursor, err := s.blobMetadataStore.GetBlobMetadataByRequestedAt(
```
jianoaix, line 196: This one is TBD in followup

disperser/dataapi/v2/server_v2.go
```diff
@@ -263,7 +263,8 @@ func (s *ServerV2) Start() error {
 	{
 		blobs := v2.Group("/blobs")
 		{
-			blobs.GET("/feed", s.FetchBlobFeed)
+			blobs.GET("/feed/forward", s.FetchBlobFeedForward)
```
dmanc, line 266: Didn't take a deep look into logic but is this simpler than just adding a query param to determine the direction?

jianoaix: It looks more clear to have separate APIs, but am open to merge them if these params can have clear semantics when they are used for both directions.

jianoaix: thinking it more, maybe using `before` and `after` naturally addresses this, like
```
- direction: string  // "forward" or "backward"
- after: string     // exclusive lower bound
- before: string    // exclusive upper bound
- cursor: string    // if present, must be within (after, before)
- limit: int        // max items to return
```
For backward: fetches (after, cursor) in DESC order
For forward: fetches (cursor, before) in ASC order

jianoaix:
ping?

ian-shim (approved):

dmanc (approved):

PR #1289, opened 2025-02-19
Support dynamic cert verifier address

litt3:

- Current code assumes that the cert verifier address doesn't ever change
- In reality, rollups will need to change cert verifier address on the fly
- This PR adds the cert verifier address as a parameter to the necessary methods, to support a dynamically configured address

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


cody-littley (approved):

api/clients/v2/verification/cert_verifier.go
```diff
@@ -118,7 +109,16 @@ func (cv *CertVerifier) VerifyCertV2FromSignedBatch(
 		return fmt.Errorf("wait for block number: %w", err)
 	}
 
-	err = cv.certVerifierCaller.VerifyDACertV2FromSignedBatch(
+	// don't try to bind to the address until AFTER waiting for the block number. if you try to bind too early, the
+	// contract might not exist yet
+	certVerifierCaller, err := verifierBindings.NewContractEigenDACertVerifierCaller(
```
samlaf, line 114: hmm this feels a bit arbitrary given that you're waiting for the batch's RBN block, which in theory should have nothing to do with the deployment address of the cert verifier.... I guess this kind of error happens in tests only? Is there another way to mitigate this?

samlaf (approved):
Mostly LGTM! Do want to point out that this change seems like its needed right now, but we'd probably need to revert this once we implement proper DACertVerifier versioning like outlined in https://app.excalidraw.com/s/1XPZRMVbRNH/7CGfvnrkKDN, where there would always be a single EigenDACertVerifierRouter entrypoint that woud route to the correct Certverifier.

image

api/clients/v2/verification/cert_verifier.go
litt3: I don't think we can safely define a mapping from `RBN` to `CertVerifierAddress` in the `EigenDACertVerifierRouter` without imposing a relationship between a given `RBN` entry and the eth block where the entry was created. Specifically, the eth transaction which adds a new `CertVerifierAddress` entry to the `EigenDACertVerifierRouter` MUST be included in eth block <= the `RBN` key of the entry. If this isn't enforced, then it would be possible for a rollup to retroactively add a `CertVerifierAddress` entry, long after a blob has been verified using the previous entry.

Therefore, I think the ordering currently enforced is correct: when a verifier encounters a blob with a given RBN, there needs to be a way for the verifier to know *for certain* that no update to the `EigenDACertVerifierRouter` is incoming, which would change the correct `CertVerifierAddress` to use for verification. I think the obvious way to guarantee this is to say "if there isn't a new entry in the `EigenDACertVerifierRouter` by the time your eth node has handled the block with blocknum == RBN, then you are safe to use the latest existing entry at that block"

litt3:
> Mostly LGTM! Do want to point out that this change seems like its needed right now, but we'd probably need to revert this once we implement proper DACertVerifier versioning like outlined in https://app.excalidraw.com/s/1XPZRMVbRNH/7CGfvnrkKDN, where there would always be a single EigenDACertVerifierRouter entrypoint that woud route to the correct Certverifier.

@samlaf I'm not sure this is true. The way I imagined us implementing the `EigenDACertVerifierRouter` involves the proxy fetching the correct `CertVerifierAddress` from the `EigenDACertVerifierRouter`, based on the `RBN` in a given cert, and passing the correct address into these client methods.

I personally like this strategy, since it means that using an `EigenDACertVerifierRouter` is totally optional. A given rollup could decide that it will never want to update the `CertVerifierAddress`, or that it would like to manage it's own upgrade process a different way, like by using config files directly in the proxy. Such a rollup could avoid using an `EigenDACertVerifierRouter` entirely.

api/clients/v2/payload_disperser.go
```diff
@@ -167,15 +167,15 @@ func (pd *PayloadDisperser) SendPayload(
 	}
 	pd.logger.Debug("Blob status CERTIFIED", "blobKey", blobKey.Hex())
 
-	eigenDACert, err := pd.buildEigenDACert(ctx, blobKey, blobStatusReply)
+	eigenDACert, err := pd.buildEigenDACert(ctx, certVerifierAddress, blobKey, blobStatusReply)
```
anupsv, line 170: I was thinking if we should check if given `certVerifierAddress` is a valid eth address. But a client providing bad value would essentially shoot themselves in the foot so i'm assuming its fine? Just want to confirm.

anupsv (approved):

api/clients/v2/payload_disperser.go
litt3: > But a client providing bad value would essentially shoot themselves in the foot

That's correct- choosing an invalid address at this stage would just cause the `buildEigenDACert` method to fail

samlaf:
> @samlaf I'm not sure this is true. The way I imagined us implementing the `EigenDACertVerifierRouter` involves the proxy fetching the correct `CertVerifierAddress` from the `EigenDACertVerifierRouter`, based on the `RBN` in a given cert, and passing the correct address into these client methods.
> 
> I personally like this strategy, since it means that using an `EigenDACertVerifierRouter` is totally optional. A given rollup could decide that it will never want to update the `CertVerifierAddress`, or that it would like to manage its own upgrade process a different way, like by using config files directly in the proxy. Such a rollup could avoid using an `EigenDACertVerifierRouter` entirely.

I see. That makes sense. We'll have to see what's the best way to implement this optionality then, because anything we push  out as "rollups could decide" actually has to be implemented in proxy where the rollup's decision is just to turn on a flag (for eg are we going to have flags for `EigenDACertVerifier` and `EigenDACertVerifierRouter` in proxy, and enforce that only one is turned on? That could work actually..

api/clients/v2/verification/cert_verifier.go
samlaf: makes sense.

samlaf (approved):

anupsv (approved):

PR #1290, opened 2025-02-19
Check that disperser uses correct blob length

litt3:

- closes https://github.com/Layr-Labs/eigenda/issues/1291

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/clients/v2/disperser_client.go
```diff
@@ -175,6 +175,18 @@ func (c *disperserClient) DisperseBlob(
 			return nil, [32]byte{}, fmt.Errorf("error deserializing blob commitments: %w", err)
 		}
 		blobCommitments = *deserialized
+
+		// We need to check that the disperser used the correct length. Even once checking the commitment from the
+		// disperser has been implemented, there is still an edge case where the disperser could truncate trailing 0s,
+		// yielding the wrong blob length, but not causing commitment verification to fail. It is important that the
```
anupsv, line 181: is there a change on the disperser side to make sure this case is handled ?

anupsv (approved):

api/clients/v2/disperser_client.go
litt3: I don't think so- this check is intended to protect against a disperser maliciously/incorrectly truncating 0s

PR #1295, opened 2025-02-19
Allow bytes from relay to be of 0 length

litt3:

- As discussed [here](https://github.com/Layr-Labs/eigenda/pull/1194#discussion_r1943397899), we have come to agreement that we shouldn't check that what a retrieving client receives is greater than 0 length
- This PR removes the 0 length check, and since the remaining method contents were only a single comparison, I removed the helper method and class entirely, in favor of simply doing the comparison

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/clients/v2/relay_payload_retriever.go
```diff
@@ -140,9 +141,13 @@ func (pr *RelayPayloadRetriever) GetPayload(
 			continue
 		}
 
-		err = verification.CheckBlobLength(blob, blobCommitments.Length)
-		if err != nil {
-			pr.log.Warn("check blob length", "blobKey", blobKey.Hex(), "relayKey", relayKey, "error", err)
+		if uint(len(blob)) > blobCommitments.Length*encoding.BYTES_PER_SYMBOL {
```
samlaf, line 144: this is the kind of code that I'm trying to avoid by forcing unit suffixes everywhere :\
It's basically impossible for me to know whether this check is correct without an IDE, which I don't like that much.

samlaf: EDIT: I actually even checkout out the code on vscode and still didn't have enough context to verify this inequality. Can you add comments to BlobCommitments struct (https://github.com/Layr-Labs/eigenda/blob/9f4d8be6ce264e2caf9757ed6cc383a896680a58/encoding/data.go#L27) and also fix the comment (BlomCommitments -> BlobCommitments)

samlaf: Also is it normal that blob is an array of bytes? Is this something that your new type system fixes? Aren't blobs arrays of Fr.Elements now?

litt3: >  Is this something that your new type system fixes?

Correct. Now that the type system PR has merged, I will be putting up a PR which updates these clients to use the new types.

litt3: > this is the kind of code that I'm trying to avoid by forcing unit suffixes everywhere 

💯 agree. I'm on board with adopting this standard, so definitely point out if you see me forgetting and generating any new code without a proper suffix

> Can you add comments to BlobCommitments struct

Sure, I will add some comments there

api/clients/v2/relay_payload_retriever.go
```diff
@@ -140,9 +141,13 @@ func (pr *RelayPayloadRetriever) GetPayload(
...
+			pr.log.Warn(
+				"received length is greater than claimed blob length",
+				"blobKey", blobKey.Hex(),
+				"relayKey", relayKey,
+				"receivedLengthBytes", len(blob),
+				"claimedLengthBytes", blobCommitments.Length*encoding.BYTES_PER_SYMBOL)
 			continue
```
anupsv, line 151: It's fine to skip over because the disperser returns the length which includes any "0" value elements at the end correct?

anupsv (approved):

api/clients/v2/relay_payload_retriever.go
litt3: Right. If a relay returns anything > blob length in the commitment, we know reflexively that the bytes cannot be correct

api/clients/v2/relay_payload_retriever.go
samlaf: Awesome thanks for the updates. For the BlobCommitment comments, you can look at the proto comments, they should prob just match: https://github.com/Layr-Labs/eigenda/blob/master/api/proto/common/common.proto#L13-L27

The state of comment drift across proto/backend/contract is one thing that I really had a way to solve. I think CUElang has some really interesting approach to this, which I want to look deeper into. But otherwise for now, perhaps we can just add a link to proto file from the backend struct to prevent comment drift?

samlaf (approved):
Happy with current state

api/clients/v2/relay_payload_retriever.go
litt3: Yeah, I think it would be really good to have a way to keep comments consistent. Next time I touch this area of code, I'll add a link as suggested

PR #1305, opened 2025-02-21
Use payload and blob structs

litt3:

- This PR puts into use the `Payload` and `Blob` structs that were recently implemented

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


cody-littley (approved):

api/clients/v2/codecs/blob.go
```diff
@@ -27,6 +28,13 @@ type Blob struct {
 
 // DeserializeBlob initializes a Blob from bytes
 func DeserializeBlob(bytes []byte, blobLengthSymbols uint32) (*Blob, error) {
```
samlaf, line 30: Might be good to document why blobLengthSymbols is needed. We've had this discussion 2-3 times already and I still can't remember the exact reason :\

api/clients/v2/codecs/blob.go
```diff
@@ -27,6 +28,13 @@ type Blob struct {
...
+	if uint32(len(bytes)) > blobLengthSymbols*encoding.BYTES_PER_SYMBOL {
+		return nil, fmt.Errorf(
+			"length (%d bytes) is greater than claimed blob length (%d bytes)",
+			len(bytes),
+			blobLengthSymbols*encoding.BYTES_PER_SYMBOL)
```
samlaf, line 35: do we also want to add a check if len(bytes) < blobLengthSymbols*encoding.BYTES_PER_SYMBOL ?

api/clients/v2/codecs/blob.go
```diff
@@ -66,17 +74,18 @@ func (b *Blob) ToPayload(payloadForm codecs.PolynomialForm) (*Payload, error) {
 	return payload, nil
 }
 
+// GetBlobLengthSymbols returns the length of the blob, in symbols
+func (b *Blob) GetBlobLengthSymbols() uint32 {
```
samlaf, line 78:
```suggestion
func (b *Blob) BlobLengthSymbols() uint32 {
```

api/clients/v2/codecs/blob.go
```diff
@@ -93,6 +102,11 @@ func (b *Blob) toEncodedPayload(payloadForm codecs.PolynomialForm) (*encodedPayl
 		return nil, fmt.Errorf("invalid polynomial form")
 	}
 
+	maxPermissiblePayloadLength, err := codec.GetMaxPermissiblePayloadLength(b.blobLengthSymbols)
```
samlaf, line 105: have you been using `Get` prefixes everywhere? I tend to prefer not having those, but as long as we're consistent everywhere.

api/clients/v2/payload_disperser.go
```diff
@@ -137,14 +128,12 @@ func (pd *PayloadDisperser) SendPayload(
 	ctx context.Context,
 	certVerifierAddress string,
 	// payload is the raw data to be stored on eigenDA
-	payload []byte,
+	payload *codecs.Payload,
```
samlaf, line 131: why is payload defined in `codecs`? Looks a bit weird no? It feels like its something that's purely related to some codec process, whereas payload is super fundamental.

api/clients/v2/payload_disperser.go
```diff
@@ -137,14 +128,12 @@ func (pd *PayloadDisperser) SendPayload(
...
 ) (*verification.EigenDACert, error) {
-
-	blobBytes, err := pd.codec.EncodeBlob(payload)
+	blob, err := payload.ToBlob(pd.config.PayloadPolynomialForm)
```
samlaf, line 133: this is super nice, like it now

api/clients/v2/relay_payload_retriever.go
```diff
@@ -130,7 +118,9 @@ func (pr *RelayPayloadRetriever) GetPayload(
 	for _, val := range indices {
 		relayKey := relayKeys[val]
 
-		blob, err := pr.getBlobWithTimeout(ctx, relayKey, blobKey)
+		blobLengthSymbols := eigenDACert.BlobInclusionInfo.BlobCertificate.BlobHeader.Commitment.Length
+
+		blob, err := pr.getBlobWithTimeout(ctx, relayKey, blobKey, blobLengthSymbols)
```
samlaf, line 123: here's another reason why I don't like `get`ter prefixes. Because then its unclear that this `getBlobWithTimeout` here is fetching over the network (the ctx gives it away but still).

would `retrieveBlobWithTimeout` make more sense?

api/clients/v2/relay_payload_retriever.go
```diff
@@ -141,17 +131,14 @@ func (pr *RelayPayloadRetriever) GetPayload(
...
-				"receivedLengthBytes", len(blob),
-				"claimedLengthBytes", blobCommitments.Length*encoding.BYTES_PER_SYMBOL)
-			continue
-		}
-
-		valid, err := verification.GenerateAndCompareBlobCommitment(pr.g1Srs, blob, blobCommitments.Commitment)
+		// TODO (litt3): eventually, we should make GenerateAndCompareBlobCommitment accept a blob, instead of the
+		//  serialization of a blob. Commitment generation operates on field elements, which is how a blob is stored
+		//  under the hood, so it's actually duplicating work to serialize the blob here. I'm declining to make this
+		//  change now, to limit the size of the refactor PR.
```
samlaf, line 137: so you're serializing and unserializing inside the function right now?

api/clients/v2/relay_payload_retriever.go
```diff
@@ -165,10 +152,10 @@ func (pr *RelayPayloadRetriever) GetPayload(
...
 
-		payload, err := pr.codec.DecodeBlob(blob)
+		payload, err := blob.ToPayload(pr.config.PayloadPolynomialForm)
 		if err != nil {
 			pr.log.Error(
-				`Cert verification was successful, but decode blob failed!
+				`Cert verification was successful, but conversion from blob to payload failed!
 					This is likely a problem with the local blob codec configuration,
 					but could potentially indicate a maliciously generated blob certificate.
 					It should not be possible for an honestly generated certificate to verify
```
samlaf, line 161: can it really be a maliciously generated blob cert? isn't it moreso a bug on the client side?

samlaf: unless you meant maliciously generated blobHeader?

api/clients/v2/relay_payload_retriever.go
```diff
@@ -187,12 +174,24 @@ func (pr *RelayPayloadRetriever) GetPayload(
 func (pr *RelayPayloadRetriever) getBlobWithTimeout(
 	ctx context.Context,
 	relayKey core.RelayKey,
-	blobKey *core.BlobKey) ([]byte, error) {
+	blobKey *core.BlobKey,
+	blobLengthSymbols uint32,
```
samlaf, line 178: can you add documentation as to where blobLengthSymbols typically comes from when calling this function? Gotten from the cert right?

api/clients/v2/relay_payload_retriever.go
```diff
@@ -187,12 +174,24 @@ func (pr *RelayPayloadRetriever) GetPayload(
...
 	timeoutCtx, cancel := context.WithTimeout(ctx, pr.config.RelayTimeout)
 	defer cancel()
 
-	return pr.relayClient.GetBlob(timeoutCtx, relayKey, *blobKey)
+	blobBytes, err := pr.relayClient.GetBlob(timeoutCtx, relayKey, *blobKey)
+	if err != nil {
+		return nil, fmt.Errorf("get blob from relay: %w", err)
+	}
+
+	blob, err := codecs.DeserializeBlob(blobBytes, blobLengthSymbols)
```
samlaf, line 189: should relayClient not returned a deserializedBlob? Or is this also one of the TODOs that you have somewhere in here?

api/clients/v2/test/relay_payload_retriever_test.go
```diff
@@ -20,22 +20,24 @@ import (
...
 	prover2 "github.com/Layr-Labs/eigenda/encoding/kzg/prover"
+	"github.com/Layr-Labs/eigenda/encoding/utils/codec"
 	"github.com/consensys/gnark-crypto/ecc/bn254"
 	"github.com/stretchr/testify/mock"
 	"github.com/stretchr/testify/require"
 )
 
 const g1Path = "../../../../inabox/resources/kzg/g1.point"
-const payloadLength = 100
+const maxPayloadBytes = 1025
```
samlaf, line 33: Why 1025? Feel like any surprising number should be documented. Aka if its not a power of 2, I feel like a reason should be given, because I naturally wonder: "is the +1 meaningful here? what is it used for? if I want to change it what will break? etc"

api/clients/v2/codecs/blob.go
litt3: This is documented on the `blobLengthSymbols` field of the struct:

```
	// blobLengthSymbols must be a power of 2, and should match the blobLength claimed in the BlobCommitment
	//
	// This value must be specified, rather than computed from the length of the coeffPolynomial, due to an edge case
	// illustrated by the following example: imagine a user disperses a very small blob, only 64 bytes, and the last 40
	// bytes are trailing zeros. When a different user fetches the blob from a relay, it's possible that the relay could
	// truncate the trailing zeros. If we were to say that blobLengthSymbols = nextPowerOf2(len(coeffPolynomial)), then the
	// user fetching and reconstructing this blob would determine that the blob length is 1 symbol, when it's actually 2.
```

api/clients/v2/codecs/blob.go
litt3: Added a comment in [7b9e48e0](https://github.com/Layr-Labs/eigenda/pull/1305/commits/7b9e48e05d0d50dccb7af22a61da6f59e124fc60), addressing why we assert byte length is <=, rather than asserting equality

api/clients/v2/codecs/blob.go
litt3: [91b80248](https://github.com/Layr-Labs/eigenda/pull/1305/commits/91b80248a3ac2edf3cd91e6691082e20419bde12)

api/clients/v2/codecs/blob.go
litt3: I tend to use `Get`, but don't have a strong preference. I will start leaving it off in the future (though universal consistency is gonna prove elusive, since I think we have instances of both naming strategies throughout)

api/clients/v2/payload_disperser.go
litt3: Makes sense, I think the `codecs` package containing `Payload`, `Blob`, and `encodedPayload` could just be renamed in its entirety. Any naming suggestions?

`Payload` and `Blob` are both fundamental, and `encodedPayload` feels like it just belongs wherever `Payload` and `Blob` end up.

api/clients/v2/relay_payload_retriever.go
litt3: It should, add TODOs to indicate places I previously missed, where future work is needed [76dac208](https://github.com/Layr-Labs/eigenda/pull/1305/commits/76dac208bb98f9572610a72f75a0dad20b95335d)

api/clients/v2/test/relay_payload_retriever_test.go
litt3: Did you notice that this is a test file? IMO an unusual upper bound on a test like this has implied documentation, i.e. "I want to test approximately this range, but I want to avoid implying that it must be a power of 2". It's changed in this PR, since I noticed I had used a hardcoded length instead of randomizing, and since I was already making changes, it seemed like 100 bytes was low for the tests.

If we need a comment for `1025`, wouldn't we also need a comment for `1024`? Since the same logic applies, that a reader won't know specifically why `1024` was chosen, and if that specific value is relevant. They might even wrongly assume that a power of 2 is necessary, and be misled.

I guess my question would be, are you saying you think every chosen test value needs a comment?

api/clients/v2/payload_disperser.go
samlaf: `types`, `datatypes`, `core`?

litt3: Went with mashup `coretypes` in [650f6800](https://github.com/Layr-Labs/eigenda/pull/1305/commits/650f6800eeafe2fc5105fed0a4432cc016f4f24c), since it seems a good balance of specificity. LMK if you're not a fan

samlaf: sgtm. I guess the only possible issue is that we've been using `core` to mean stuff in eigenda-core, as opposed to integration, and these seem like more integration-like types. But unclear to me if we want this kind of distinction going forward (especially as we bring everything into monorepo). Let's roll with this for now but we can keep thinking about it.

samlaf (approved):

anupsv (commented):
lgtm.

PR #1309, opened 2025-02-21
rewards router

0x0aa0:
Adds a ServiceManagerRewardsRouter contract that allows governance to permission multiple rewards initiators to push operator rewards


samlaf:
@pakim249CAL can we close this?

PR #1311, opened 2025-02-21
Add environment variable to hold eth RPC urls.

cody-littley:

Make it so that the eth RPC URLs can be loaded from an environment variable.

### DIFF
--- a/test/v2/client/test_client_config.go
+++ b/test/v2/client/test_client_config.go
@@ -23,7 +23,13 @@ type TestClientConfig struct {
 	// The disperser's port
 	DisperserPort int
 	// The URL(s) to point the eth client to
+	//
+	// Either this or EthRPCURLsVar must be set. If both are set, EthRPCURLs is used.
 	EthRPCURLs []string
+	// The environment variable that contains the URL(s) to point the eth client to. Use a comma-separated list.
+	//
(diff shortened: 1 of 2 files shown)


test/v2/client/test_client_config.go
```diff
@@ -23,7 +23,13 @@ type TestClientConfig struct {
 	// The disperser's port
 	DisperserPort int
 	// The URL(s) to point the eth client to
+	//
+	// Either this or EthRPCURLsVar must be set. If both are set, EthRPCURLs is used.
 	EthRPCURLs []string
+	// The environment variable that contains the URL(s) to point the eth client to. Use a comma-separated list.
+	//
+	// Either this or EthRPCURLs must be set. If both are set, EthRPCURLsVar is used.
```
litt3, line 31: > If both are set, EthRPCURLsVar is used.

This contradicts the comment above, which states "If both are set, EthRPCURLs is used."

litt3 (approved):

test/v2/client/test_client_config.go
cody-littley: Good catch. The first comment was correct, this one was wrong. Fixed.

litt3 (approved):

PR #1314, opened 2025-02-22
Support backward iteration for blob feed

jianoaix:
This actually adds support for backward blob feed API (the API added in https://github.com/Layr-Labs/eigenda/pull/1286).

This will make it simple and deterministic to support use cases that need to retrieve blobs backward along the timeline.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/common/v2/blobstore/dynamo_metadata_store.go
+++ b/disperser/common/v2/blobstore/dynamo_metadata_store.go
@@ -354,7 +354,7 @@ func (s *BlobMetadataStore) queryBucketBlobMetadata(
 }
 
 // GetBlobMetadataByRequestedAtForward returns blobs (as BlobMetadata) in cursor range
-// (after, until) (both exclusive). Blobs are ordered by <RequestedAt, BlobKey>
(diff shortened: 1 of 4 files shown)


disperser/common/v2/blobstore/dynamo_metadata_store.go
```diff
@@ -354,7 +354,7 @@ func (s *BlobMetadataStore) queryBucketBlobMetadata(
 }
 
 // GetBlobMetadataByRequestedAtForward returns blobs (as BlobMetadata) in cursor range
-// (after, until] (after exclusive, until inclusive). Blobs are ordered by <RequestedAt, BlobKey>
+// (after, until) (both exclusive). Blobs are ordered by <RequestedAt, BlobKey>
```
jianoaix, line 357: Making both ends exclusive, because:
- it's easier for the feed API implementation
- this is at nanosecond level (the highest resolution), so inclusive use cases can just make a `[L-1, R+1]` adjustment which will work correctly (at nanosecond, there is no time points between `L-1` and `L`)

merli-eigenlabs (approved):

PR #1323, opened 2025-02-25
refactor: use block number for all payment vault param calls

hopeyen:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/eth/reader.go
+++ b/core/eth/reader.go
@@ -854,12 +854,13 @@ func (t *Reader) GetOnDemandPaymentByAccount(ctx context.Context, accountID geth
 	}, nil
 }
 
-func (t *Reader) GetGlobalSymbolsPerSecond(ctx context.Context) (uint64, error) {
+func (t *Reader) GetGlobalSymbolsPerSecond(ctx context.Context, blockNumber uint32) (uint64, error) {
 	if t.bindings.PaymentVault == nil {
 		return 0, errors.New("payment vault not deployed")
 	}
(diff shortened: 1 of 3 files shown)


ian-shim (approved):

core/eth/reader.go
```diff
@@ -854,64 +854,70 @@ func (t *Reader) GetOnDemandPaymentByAccount(ctx context.Context, accountID geth
...
 
-func (t *Reader) GetGlobalRatePeriodInterval(ctx context.Context) (uint32, error) {
+func (t *Reader) GetGlobalRatePeriodInterval(ctx context.Context, blockNumber uint32) (uint32, error) {
 	if t.bindings.PaymentVault == nil {
 		return 0, errors.New("payment vault not deployed")
 	}
 	globalRateBinInterval, err := t.bindings.PaymentVault.GlobalRatePeriodInterval(&bind.CallOpts{
-		Context: ctx,
+		Context:     ctx,
+		BlockNumber: big.NewInt(int64(blockNumber)),
```
anupsv, line 877: why `int64` here ?

hopeyen: callOps require a big.Int but blockNumber has to be cast into `int64` to create the int

anupsv: I see. Looks like something we don't control to change the receiver also to uint64?

anupsv (approved):

core/eth/reader.go
hopeyen: yep, `CallOpts` is part of go-ethereum

PR #1326, opened 2025-02-25
refactor: Structify Opeartor Socket

supriya-premkumar:
Changes:
	Defines a concrete type OperatorSocket
	Internal state is unexported to prevent unintented changes from the importing package.
	Defines Get functions to access methods

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/node_client.go
+++ b/api/clients/node_client.go
@@ -85,8 +85,9 @@ func (c client) GetChunks(
 	quorumID core.QuorumID,
 	chunksChan chan RetrievedChunks,
 ) {
+
 	conn, err := grpc.NewClient(
-		core.OperatorSocket(opInfo.Socket).GetV1RetrievalSocket(),
+		opInfo.Socket.GetV1RetrievalSocket(),
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
 	)
 	if err != nil {
--- a/core/eth/reader.go
+++ b/core/eth/reader.go
@@ -509,10 +509,16 @@ func (t *Reader) GetOperatorStakesWithSocketForQuorums(ctx context.Context, quor
 		state[quorumID] = make(map[core.OperatorIndex]core.OperatorStakeWithSocket, len(result.Operators[i]))
 		for j, op := range result.Operators[i] {
 			operatorIndex := core.OperatorIndex(j)
+			operatorSocket, err := core.ParseOperatorSocket(result.Sockets[i][j])
+			if err != nil {
+				t.logger.Errorf("failed to parse operator socket from: %v, err: %v", result.Sockets[i][j], err)
+				return nil, fmt.Errorf("failed to parse operator socket from: %v, err: %v", result.Sockets[i][j], err)
+
+			}
 			state[quorumID][operatorIndex] = core.OperatorStakeWithSocket{
 				Stake:      op.Stake,
 				OperatorID: op.OperatorId,
-				Socket:     core.OperatorSocket(result.Sockets[i][j]),
+				Socket:     operatorSocket,
 			}
 		}
 	}
--- a/core/state.go
+++ b/core/state.go
@@ -13,78 +13,84 @@ import (
 
 // Operators
 
-// OperatorSocket is formatted as "host:dispersalPort;retrievalPort;v2DispersalPort"
-type OperatorSocket string
+// OperatorSocket is formatted as follows:
+// v1-only 	  "host:dispersalPort;retrievalPort"
+// v1-and-v2  "host:dispersalPort;retrievalPort;v2DispersalPort;v2RetrievalPort"
+// v2-only    "host:dispersalPort;retrievalPort;v2DispersalPort;v2RetrievalPort" (currently v2 only mandates running v1)
 
-func (s OperatorSocket) String() string {
-	return string(s)
+type OperatorSocket struct {
+	host            string
+	v1DispersalPort string
+	v1RetrievalPort string
+	v2DispersalPort string
+	v2RetrievalPort string
 }
 
-func MakeOperatorSocket(nodeIP, dispersalPort, retrievalPort, v2DispersalPort, v2RetrievalPort string) OperatorSocket {
-	//TODO:  Add config checks for invalid v1/v2 configs -- for v1, both v2 ports must be empty and for v2, both ports must be valid, reject any other combinations.
-	if v2DispersalPort == "" && v2RetrievalPort == "" {
-		return OperatorSocket(fmt.Sprintf("%s:%s;%s", nodeIP, dispersalPort, retrievalPort))
+// NewOperatorSocket returns new OperatorSocket struct
+func NewOperatorSocket(host, dispersalPort, retrievalPort, v2DispersalPort, v2RetrievalPort string) OperatorSocket {
+	return OperatorSocket{
+		host:            host,
+		v1DispersalPort: dispersalPort,
+		v1RetrievalPort: retrievalPort,
+		v2DispersalPort: v2DispersalPort,
+		v2RetrievalPort: v2RetrievalPort,
 	}
-	return OperatorSocket(fmt.Sprintf("%s:%s;%s;%s;%s", nodeIP, dispersalPort, retrievalPort, v2DispersalPort, v2RetrievalPort))
 }
+
+func (s OperatorSocket) Encode() string {
+	if s.v2DispersalPort == "" && s.v2RetrievalPort == "" {
+		return fmt.Sprintf("%s:%s;%s", s.host, s.v1DispersalPort, s.v1RetrievalPort)
+	}
+	return fmt.Sprintf("%s:%s;%s;%s;%s", s.host, s.v1DispersalPort, s.v1RetrievalPort, s.v2DispersalPort, s.v2RetrievalPort)
+}
+
 
 type StakeAmount = *big.Int
 
-func ParseOperatorSocket(socket string) (host, v1DispersalPort, v1RetrievalPort, v2DispersalPort, v2RetrievalPort string, err error) {
+func ParseOperatorSocket(opSocketStr string) (opSocket OperatorSocket, err error) {
 
-	s := strings.Split(socket, ";")
+	operatorSocket := OperatorSocket{}
+	s := strings.Split(opSocketStr, ";")
 
-	host, v1DispersalPort, err = net.SplitHostPort(s[0])
+	host, v1DispersalPort, err := net.SplitHostPort(s[0])
 	if err != nil {
-		err = fmt.Errorf("invalid host address format in %s: it must specify valid IP or host name (ex. 0.0.0.0:32004;32005;32006;32007)", socket)
-
+		err = fmt.Errorf("invalid host address format in %s: it must specify valid IP or host name (ex. 0.0.0.0:32004;32005;32006;32007)", opSocketStr)
 		return
 	}
 	if _, err = net.LookupHost(host); err != nil {
-		//Invalid host
-		host, v1DispersalPort, v1RetrievalPort, v2DispersalPort, v2RetrievalPort, err =
-			"", "", "", "", "",
-			fmt.Errorf("invalid host address format in %s: it must specify valid IP or host name (ex. 0.0.0.0:32004;32005;32006;32007)", socket)
+		err = fmt.Errorf("invalid host address format in %s: unable to lookup host)", opSocketStr)
 		return
 	}
 	if err = ValidatePort(v1DispersalPort); err != nil {
-		host, v1DispersalPort, v1RetrievalPort, v2DispersalPort, v2RetrievalPort, err =
-			"", "", "", "", "",
-			fmt.Errorf("invalid v1 dispersal port format in %s: it must specify valid v1 dispersal port (ex. 0.0.0.0:32004;32005;32006;32007)", socket)
+		err = fmt.Errorf("invalid v1 dispersal port format in %s: it must specify valid v1 dispersal port (ex. 0.0.0.0:32004;32005;32006;32007)", opSocketStr)
 		return
 	}
 
 	switch len(s) {
 	case 4:
-		v2DispersalPort = s[2]
-		if err = ValidatePort(v2DispersalPort); err != nil {
-			host, v1DispersalPort, v1RetrievalPort, v2DispersalPort, v2RetrievalPort, err =
-				"", "", "", "", "",
-				fmt.Errorf("invalid v2 dispersal port format in %s: it must specify valid v2 dispersal port (ex. 0.0.0.0:32004;32005;32006;32007)", socket)
+		if err = ValidatePort(s[2]); err != nil {
+			err = fmt.Errorf("invalid v2 dispersal port format in %s: it must specify valid v2 dispersal port (ex. 0.0.0.0:32004;32005;32006;32007)", opSocketStr)
 			return
 		}
 
-		v2RetrievalPort = s[3]
-		if err = ValidatePort(v2RetrievalPort); err != nil {
-			host, v1DispersalPort, v1RetrievalPort, v2DispersalPort, v2RetrievalPort, err =
-				"", "", "", "", "",
-				fmt.Errorf("invalid v2 retrieval port format in %s: it must specify valid v2 retrieval port (ex. 0.0.0.0:32004;32005;32006;32007)", socket)
+		if err = ValidatePort(s[3]); err != nil {
+			err = fmt.Errorf("invalid v2 retrieval port format in %s: it must specify valid v2 retrieval port (ex. 0.0.0.0:32004;32005;32006;32007)", opSocketStr)
 			return
 		}
+		operatorSocket.v2DispersalPort = s[2]
+		operatorSocket.v2RetrievalPort = s[3]
 		fallthrough
 	case 2:
-		// V1 Parsing
-		v1RetrievalPort = s[1]
-		if err = ValidatePort(v1RetrievalPort); err != nil {
-			host, v1DispersalPort, v1RetrievalPort, v2DispersalPort, v2RetrievalPort, err =
-				"", "", "", "", "",
-				fmt.Errorf("invalid v1 retrieval port format in %s: it must specify valid v1 retrieval port (ex. 0.0.0.0:32004;32005;32006;32007)", socket)
+		operatorSocket.host = host
+		operatorSocket.v1DispersalPort = v1DispersalPort
+		if err = ValidatePort(s[1]); err != nil {
+			err = fmt.Errorf("invalid v1 retrieval port format in %s: it must specify valid v1 retrieval port (ex. 0.0.0.0:32004;32005;32006;32007)", opSocketStr)
+			return
 		}
-		return
+		operatorSocket.v1RetrievalPort = s[1]
+		return operatorSocket, nil
 	default:
-		host, v1DispersalPort, v1RetrievalPort, v2DispersalPort, v2RetrievalPort, err =
-			"", "", "", "", "",
-			fmt.Errorf("invalid socket address format %s: it must specify v1 dispersal/retrieval ports, or v2 dispersal/retrieval ports (ex. 0.0.0.0:32004;32005;32006;32007)", socket)
+		err = fmt.Errorf("invalid opSocketStr address format %s: it must specify v1 dispersal/retrieval ports, or v2 dispersal/retrieval ports (ex. 0.0.0.0:32004;32005;32006;32007)", opSocketStr)
 		return
 	}
 }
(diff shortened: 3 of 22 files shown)


core/state.go
```diff
@@ -13,78 +13,78 @@ import (
...
+// OperatorSocket is formatted as "host:dispersalPort;retrievalPort;v2DispersalPort;v2RetrievalPort"
+type OperatorSocket struct {
+	host            string
+	v1DispersalPort string
+	v1RetrievalPort string
+	v2DispersalPort string
+	v2RetrievalPort string
+}
 
 func (s OperatorSocket) String() string {
```
jianoaix, line 25: Marshal/Unmarshal or Serialize/Deserialize? String() feels like producing something for logging, debugging or display.

core/state.go
```diff
@@ -13,78 +13,78 @@ import (
...
+		return fmt.Sprintf("%s:%s;%s", s.host, s.v1DispersalPort, s.v2RetrievalPort)
+	}
+	return fmt.Sprintf("%s:%s;%s;%s;%s", s.host, s.v1DispersalPort, s.v1RetrievalPort, s.v2DispersalPort, s.v2RetrievalPort)
 }
 
-func MakeOperatorSocket(nodeIP, dispersalPort, retrievalPort, v2DispersalPort, v2RetrievalPort string) OperatorSocket {
-	//TODO:  Add config checks for invalid v1/v2 configs -- for v1, both v2 ports must be empty and for v2, both ports must be valid, reject any other combinations.
-	if v2DispersalPort == "" && v2RetrievalPort == "" {
-		return OperatorSocket(fmt.Sprintf("%s:%s;%s", nodeIP, dispersalPort, retrievalPort))
+func NewOperatorSocket(nodeIP, dispersalPort, retrievalPort, v2DispersalPort, v2RetrievalPort string) OperatorSocket {
```
jianoaix, line 32: nit: move factory method right above (before String())

disperser/dataapi/queried_operators_handlers.go
```diff
@@ -198,7 +198,11 @@ func checkIsOnlineAndProcessOperator(operatorStatus OperatorOnlineStatus, operat
 	var isOnline bool
 	var socket string
 	if operatorStatus.IndexedOperatorInfo != nil {
-		socket = core.OperatorSocket(operatorStatus.IndexedOperatorInfo.Socket).GetV1RetrievalSocket()
+		operatorSocket, err := core.ParseOperatorSocket(operatorStatus.IndexedOperatorInfo.Socket)
+		if err != nil {
+			logger.Warn("Failed to parse operator socket", "err", err)
+		}
+		socket = operatorSocket.GetV1RetrievalSocket()
```
jianoaix, line 205: If `err != nil`, how can this still get v1 retrieval socket?

disperser/dataapi/operator_handler.go
```diff
@@ -95,7 +95,11 @@ func (oh *OperatorHandler) ProbeV2OperatorPorts(ctx context.Context, operatorId
 		return &OperatorPortCheckResponse{}, err
 	}
 
-	operatorSocket := core.OperatorSocket(operatorInfo.Socket)
+	operatorSocket, err := core.ParseOperatorSocket(operatorInfo.Socket)
+	if err != nil {
+		oh.logger.Warn("failed to parse operator socket", "operatorId", operatorId, "error", err)
```
jianoaix, line 100: In what cases it'll fail to parse? Is that just a warning issue?

api/clients/node_client.go
```diff
@@ -40,8 +40,12 @@ func (c client) GetBlobHeader(
 	batchHeaderHash [32]byte,
```
ian-shim, line 40: instead of taking `socket` parameter as string, can we take it as `OperatorSocket`?

core/state.go
```diff
@@ -13,78 +13,78 @@ import (
...
+	v1DispersalPort string
+	v1RetrievalPort string
+	v2DispersalPort string
+	v2RetrievalPort string
+}
 
 func (s OperatorSocket) String() string {
-	return string(s)
+	if s.v2DispersalPort == "" && s.v2RetrievalPort == "" {
+		return fmt.Sprintf("%s:%s;%s", s.host, s.v1DispersalPort, s.v2RetrievalPort)
```
ian-shim, line 27: s/`s.v2RetrievalPort`/`s.v1RetrievalPort`

should it also check if v1 ports are there?

core/state.go
```diff
@@ -13,78 +13,78 @@ import (
...
+	}
+	return fmt.Sprintf("%s:%s;%s;%s;%s", s.host, s.v1DispersalPort, s.v1RetrievalPort, s.v2DispersalPort, s.v2RetrievalPort)
 }
 
-func MakeOperatorSocket(nodeIP, dispersalPort, retrievalPort, v2DispersalPort, v2RetrievalPort string) OperatorSocket {
-	//TODO:  Add config checks for invalid v1/v2 configs -- for v1, both v2 ports must be empty and for v2, both ports must be valid, reject any other combinations.
-	if v2DispersalPort == "" && v2RetrievalPort == "" {
-		return OperatorSocket(fmt.Sprintf("%s:%s;%s", nodeIP, dispersalPort, retrievalPort))
+func NewOperatorSocket(nodeIP, dispersalPort, retrievalPort, v2DispersalPort, v2RetrievalPort string) OperatorSocket {
+	return OperatorSocket{
```
ian-shim, line 33: should we add a validation here? (i.e. host & v1 ports cannot be empty, if `v2DispersalPort` exists, `v2RetrievalPort` must exist)

disperser/dataapi/queried_operators_handlers.go
supriya-premkumar: There is an existing implementation issue where the expectation is to call get sockets method to return the zero value on error conditions. This will be a longer pull refactor.

Ideally we should not rely on zero values as errors, instead anytime we get an *OperatorSocket struct and an error we need to handle error and return before calling its implemented methods. Will get this addressed in the next PR

api/clients/node_client.go
supriya-premkumar: yes. However there are two other functions that we need to unify. Currently this implementation is split between `MakeNewOperatorSocket` and `ParseOperatorSocket` Having a unified definition here will remove some confusion and guarantee better type safety. 

Can I make this change in the subsequent PR? Tracking it here EGDA-1061

core/state.go
supriya-premkumar: Ack. Good call.

core/state.go
supriya-premkumar: good catch. was a typo!

core/state.go
supriya-premkumar: Agreed. The error handling will come in as a part of EGDA-1061.

ian-shim: Not sure if doing it as a separate PR makes sense unless you're looking to merge them together. I think this change is relevant for this PR

ian-shim (commented):
This change should be included in the v0.9.0-rc.1 release. Can we get this in this week?

core/state.go
supriya-premkumar: Thinking more about this, a Marshal/Serialize  and Unmarshal/Deserialize might not be accurate because what we want is the socket string representation. Plus we need only the stringified output, without converting this to serializables, i.e there is no need for implementing deserialize unmarshals.

Was originally overriding the String() method. To avoid confusion for logging, debug use case. I will call this `Normalize` instead, as we are using custom logic to represent socket string in different runtime modes.

core/state.go
supriya-premkumar: Wanted to do both newer error handling and unification of NewOperatorSocket and ParseOperatorSocket in a different PR. As this PR was a refactor, was planning not to add newer logic.

But happy to add the error handling in ParseOperatorSocket.

supriya-premkumar: Actually we are already handling these cases in ParseOperatorSocket. Added [new unit tests](https://github.com/Layr-Labs/eigenda/pull/1326/files#diff-8c20029f49ea6a96942c0358cc7ec5eded6578f512948aecc7d6d48f2933c55dR343) to ensure that all invalid string socket string errors are correctly handled.

disperser/dataapi/operator_handler.go
jianoaix: Here and below: useful to wrap this error with operatorId information, so it's more clear what went wrong

node/plugin/cmd/main.go
```diff
@@ -135,12 +136,36 @@ func pluginOps(ctx *cli.Context) {
...
-	_, dispersalPort, retrievalPort, v2DispersalPort, v2RetrievalPort, err := core.ParseOperatorSocket(config.Socket)
+	opSocket, err := core.ParseOperatorSocket(config.Socket)
 	if err != nil {
 		log.Printf("Error: failed to parse operator socket: %v", err)
 		return
 	}
 
+	_, dispersalPort, err := net.SplitHostPort(opSocket.GetV1DispersalSocket())
+	if err != nil {
+		log.Printf("Error: failed to get dispersal port from the socket: %s, err: %s", opSocket.Normalize(), err)
```
jianoaix, line 147: It may be better to log the `config.Socket`, let users see which param they supplied is not right

core/state.go
```diff
@@ -13,78 +13,80 @@ import (
...
+	v1DispersalPort string
+	v1RetrievalPort string
+	v2DispersalPort string
+	v2RetrievalPort string
+}
 
-func (s OperatorSocket) String() string {
-	return string(s)
+// NewOperatorSocket returns new OperatorSocket struct
+func NewOperatorSocket(nodeIP, dispersalPort, retrievalPort, v2DispersalPort, v2RetrievalPort string) OperatorSocket {
```
jianoaix, line 26: nodeIP -> host, it may not be IP but an URL

core/state.go
```diff
@@ -13,78 +13,80 @@ import (
 
 // Operators
 
-// OperatorSocket is formatted as "host:dispersalPort;retrievalPort;v2DispersalPort"
-type OperatorSocket string
+// OperatorSocket is formatted as "host:dispersalPort;retrievalPort;v2DispersalPort;v2RetrievalPort"
```
jianoaix, line 16: This comment is not quite true: when v2 ports are empty, it's not in this format

core/state.go
```diff
@@ -13,78 +13,80 @@ import (
...
+		v2DispersalPort: v2DispersalPort,
+		v2RetrievalPort: v2RetrievalPort,
+	}
 }
 
-func MakeOperatorSocket(nodeIP, dispersalPort, retrievalPort, v2DispersalPort, v2RetrievalPort string) OperatorSocket {
-	//TODO:  Add config checks for invalid v1/v2 configs -- for v1, both v2 ports must be empty and for v2, both ports must be valid, reject any other combinations.
-	if v2DispersalPort == "" && v2RetrievalPort == "" {
-		return OperatorSocket(fmt.Sprintf("%s:%s;%s", nodeIP, dispersalPort, retrievalPort))
+func (s OperatorSocket) Normalize() string {
```
jianoaix, line 36: Given the opposite function is `ParseOperatorSocket`, it may be clear to use something like `Encode()` (Parse/Encode pair is clear)?

node/plugin/cmd/main.go
```diff
@@ -135,12 +136,36 @@ func pluginOps(ctx *cli.Context) {
...
 	}
 
-	_, dispersalPort, retrievalPort, v2DispersalPort, v2RetrievalPort, err := core.ParseOperatorSocket(config.Socket)
+	opSocket, err := core.ParseOperatorSocket(config.Socket)
 	if err != nil {
 		log.Printf("Error: failed to parse operator socket: %v", err)
 		return
 	}
 
+	_, dispersalPort, err := net.SplitHostPort(opSocket.GetV1DispersalSocket())
```
jianoaix, line 145: Actually these look odd to me: with `ParseOperatorSocket` you already parsed the socket, but then you use `GetV1DispersalSocket` to encode it back to string, and then use `SplitHostPort` to decode again.

core/eth/reader.go
```diff
@@ -509,10 +509,14 @@ func (t *Reader) GetOperatorStakesWithSocketForQuorums(ctx context.Context, quor
 		state[quorumID] = make(map[core.OperatorIndex]core.OperatorStakeWithSocket, len(result.Operators[i]))
 		for j, op := range result.Operators[i] {
 			operatorIndex := core.OperatorIndex(j)
+			operatorSocket, err := core.ParseOperatorSocket(result.Sockets[i][j])
+			if err != nil {
+				t.logger.Errorf("failed to parse operator socket from: %v, err: %v", result.Sockets[i][j], err)
```
jianoaix, line 514: This error handling seems not right: Seeing a malformed socket is internal error (it shouldn't be accepted when registering the socket in the first place), and L519 still uses this socket and later returns the value to the caller.

disperser/batcher/grpc/dispatcher.go
```diff
@@ -120,13 +120,18 @@ func (c *dispatcher) sendAllChunks(ctx context.Context, state *core.IndexedOpera
 
 func (c *dispatcher) sendChunks(ctx context.Context, blobs []*core.EncodedBlobMessage, batchHeader *core.BatchHeader, op *core.IndexedOperatorInfo) (*core.Signature, error) {
 	// TODO Add secure Grpc
+	operatorSocket, err := core.ParseOperatorSocket(op.Socket)
+	if err != nil {
+		c.logger.Warn("Disperser failed to parse socket", "socket", op.Socket, "err", err)
```
jianoaix, line 125: Here and below: this should log error

disperser/common/semver/semver.go
```diff
@@ -28,7 +28,10 @@ func ScanOperators(operators map[core.OperatorID]*core.IndexedOperatorInfo, oper
 	operatorChan := make(chan core.OperatorID, len(operators))
 	worker := func() {
 		for operatorId := range operatorChan {
-			operatorSocket := core.OperatorSocket(operators[operatorId].Socket)
+			operatorSocket, err := core.ParseOperatorSocket(operators[operatorId].Socket)
+			if err != nil {
+				logger.Warnf("Failed to parse operator socket: %v", err)
```
jianoaix, line 33: Error handling seems not right: the "operatorSocket" is still being used below even if this is error

core/eth/reader.go
supriya-premkumar: Agreed. I think this issue existed before. We were directly casting result.Sockets[i][j] into a string without any verification. This PR makes this a typed struct. 

Was trying to keep the focus of this PR to just the refactor while minimizing adding new behavior. Let me return the err here.

core/state.go
supriya-premkumar: Ack. We internally handle this as a host. ParseOperator also does a net.Lookup of this param to provide better err handling.

disperser/batcher/grpc/dispatcher.go
supriya-premkumar: Ack

node/plugin/cmd/main.go
supriya-premkumar: We don't want to be exporting internal fields of the struct and have well defined methods to access variables. 
This allows us to have defensive coding mechanisms where we don't let pkg importee modify internal state of operator socket. 

As we want to have access to read only value of dispersal port, didn't want to make these exportable fields in the struct def.

node/plugin/cmd/main.go
supriya-premkumar: Ack. Good call.

supriya-premkumar: Parsing errors are logged on line 141

core/state.go
```diff
@@ -13,78 +13,84 @@ import (
 
 // Operators
 
-// OperatorSocket is formatted as "host:dispersalPort;retrievalPort;v2DispersalPort"
-type OperatorSocket string
+// OperatorSocket is formatted as follows:
+// v1-only 	  "host:dispersalPort;retrievalPort"
+// v1-and-v2  "host:dispersalPort;retrievalPort;v2DispersalPort;v2RetrievalPort"
+// v2-only    "host:dispersalPort;retrievalPort;v2DispersalPort;v2RetrievalPort" (currently v2 only mandates running v1)
```
jianoaix, line 19: This isn't a valid setup though, as v1 is required

disperser/controller/dispatcher.go
```diff
@@ -174,6 +176,19 @@ func (d *Dispatcher) HandleBatch(ctx context.Context) (chan core.SigningMessage,
 			continue
 		}
 
+		host, v2DispersalPort, err := net.SplitHostPort(operatorSocket.GetV2DispersalSocket())
+		if err != nil {
+			d.logger.Warn("failed to get node client; incorrect operator socket format", "operator", opID.Hex(), "host", host, "v2DispersalPort", v2DispersalPort, "err", err)
```
jianoaix, line 181: This is an error instead of warning. There may be other spots still using warning in this PR.

node/plugin/cmd/main.go
jianoaix: I think it's idiomatic to make fields public if they need to be accessed outside in golang. If you really want read-only struct (does it provide that much value here?), then adding getters should be fiine

disperser/common/semver/semver.go
jianoaix: I think this needs to be fixed, a return value is undefined when there is error

node/plugin/cmd/main.go
supriya-premkumar: It adds value by not letting anybody outside modify endpoint state. Essentially the struct wraps network information. This should be abuse proof outside of the package instantiating it.

disperser/common/semver/semver.go
supriya-premkumar: The return values are not undefined. These are empty strings, which is the zero value for strings.
The code currently (before structify PR) mandates this behavior. Unless we clean up this everywhere else, which I believe is not the scope of the PR. 

I think your concern is valid if we were returning a pointer and the dereferencing them errors would lead to panic. 
We don't do that here. The only use of a hypothetically errored out `operatorSocket` is to call `GetV1RetrievalSocket()` or `GetV1DispersalSocket()` which returns the zero value for this.

core/state.go
supriya-premkumar: We need a v2-only flag to support isolation, i.e allowing validators to run v1 and v2 on separate instances.

disperser/dataapi/operator_handler.go
```diff
@@ -144,7 +148,12 @@ func (oh *OperatorHandler) ProbeV1OperatorPorts(ctx context.Context, operatorId
...
```
jianoaix, line 153: error log here

disperser/dataapi/queried_operators_handlers.go
```diff
@@ -198,7 +198,11 @@ func checkIsOnlineAndProcessOperator(operatorStatus OperatorOnlineStatus, operat
 	var isOnline bool
 	var socket string
 	if operatorStatus.IndexedOperatorInfo != nil {
-		socket = core.OperatorSocket(operatorStatus.IndexedOperatorInfo.Socket).GetV1RetrievalSocket()
+		operatorSocket, err := core.ParseOperatorSocket(operatorStatus.IndexedOperatorInfo.Socket)
+		if err != nil {
+			logger.Warn("Failed to parse operator socket", "err", err)
```
jianoaix, line 203: error log here

node/plugin/cmd/main.go
jianoaix: I think it's minor. If you really want it, I'd add getters. It's convoluted to do this parse-encode-parse loop here.

disperser/common/semver/semver.go
jianoaix: I don't think this looks good. If you still want to proceed upon error, you may define and use an empty or zero value constant instead of using returned value.

disperser/dataapi/queried_operators_handlers.go
jianoaix: Here as well

disperser/common/semver/semver.go
supriya-premkumar: Again, this is the existing behavior, as this is a refactor I wouldn't add new functionality here. The easy solution here is to not make ParseOperatorSocket return an err, but that would be worse than what we have now.

jianoaix: Sure this is existing, but why not make it less crazy (by having a `EmptyOperatorSocket` instead of using a return value upon error)

node/plugin/cmd/main.go
jianoaix: Unresolved: I don't see this is addressed

supriya-premkumar: Making the fields public when not necessary would be a code smell. It allows for unnecessary modification outside of package. I don't see an issue with keeping this as is.

PR #1327, opened 2025-02-25
avoid unnecessary casting

hopeyen:

https://linear.app/eigenlabs/issue/EGDA-996/eda2-10-[medium]-avoid-lossy-and-unnecessary-casting

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/clients/v2/accountant.go
```diff
@@ -67,9 +67,9 @@ func NewAccountant(accountID string, reservation *core.ReservedPayment, onDemand
 // indicating on-demand payment.
 // These generated values are used to create the payment header and signature, as specified in
 // api/proto/common/v2/common_v2.proto
-func (a *Accountant) BlobPaymentInfo(ctx context.Context, numSymbols uint32, quorumNumbers []uint8, timestamp int64) (*big.Int, error) {
+func (a *Accountant) BlobPaymentInfo(ctx context.Context, numSymbols uint64, quorumNumbers []uint8, timestamp int64) (*big.Int, error) {
 	currentReservationPeriod := meterer.GetReservationPeriodByNanosecond(timestamp, a.reservationWindow)
-	symbolUsage := uint64(a.SymbolsCharged(numSymbols))
+	symbolUsage := a.SymbolsCharged(numSymbols)
```
anupsv, line 72: Is the a max value for numSymbols that we should be checking ?

api/clients/v2/accountant.go
```diff
@@ -128,22 +128,22 @@ func (a *Accountant) AccountBlob(ctx context.Context, timestamp int64, numSymbol
...
 	}
 	// Round up to the nearest multiple of MinNumSymbols
-	return uint32(core.RoundUpDivide(uint(numSymbols), uint(a.minNumSymbols))) * a.minNumSymbols
+	return core.RoundUpDivide(numSymbols, a.minNumSymbols) * a.minNumSymbols
 }
 
-func (a *Accountant) GetRelativePeriodRecord(index uint32) *PeriodRecord {
-	relativeIndex := index % a.numBins
+func (a *Accountant) GetRelativePeriodRecord(index uint64) *PeriodRecord {
+	relativeIndex := uint32(index % uint64(a.numBins))
```
anupsv, line 146: No loss of precision here ?

anupsv (approved):

api/clients/v2/accountant.go
hopeyen: Not enforced by payments, I think it will be checked [in validation](https://github.com/Layr-Labs/eigenda/blob/d14567918f036284279cd9b805d587fa31eb3104/disperser/apiserver/disperse_blob_v2.go#L145) before it gets to meterer. If you think it is within the meterer scope and should double check, I can add config to the meterer for a maximum number of blobs and add a check

api/clients/v2/accountant.go
hopeyen: In theory the `relativeIndex` would never be greater than uint32 max, since we are doing a mod of `numBins` which is an uint32

api/clients/v2/accountant.go
anupsv: If it's enforced somewhere else and it's the only call path then no need.

anupsv (approved):

ian-shim (approved):

PR #1328, opened 2025-02-25
Implement relay url provider

litt3:

- This PR changes how the relay client is interacted with: rather than constructing a new relay client every time there are updates, the relay client now is able to create new connections on the fly, as they are added
- This PR is needed to support [blazar clients](https://linear.app/eigenlabs/issue/EGDA-827/payloadretriever-should-dynamically-fetch-relay-urls), which currently don't have a way to receive relay URL updates
- In addition to implications for the blazar clients, this changeset has implications for DA validator nodes, which should be considered when reviewing this PR

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


cody-littley (approved):

api/clients/v2/relay/key_lock.go
```diff
@@ -0,0 +1,41 @@
...
+// KeyLock is a utility that provides a way to lock access to a given key of type T
+//
+// This utility is useful in situations where you want to synchronize operations for something that doesn't exist
+// in a concrete form. For example, perhaps you only want to create connections with a given peer on a single
+// thread of execution, but the new peer could appear simultaneously in concurrent operations. This utility allows
+// the first thread which encounters the new peer to perform necessary initialization tasks, and store generated
+// artifacts in a central location for subsequent callers to access.
+type KeyLock[T comparable] struct {
+	// Map from key T to a mutex that corresponds to that key
+	keyMutexMap map[T]*sync.Mutex
```
anupsv, line 16: It's upto the code which acquired the lock to release it as well ?

anupsv (approved):

api/clients/v2/relay/key_lock.go
litt3: Yes- I added a comment to state this explicitly:

```
// The caller MUST eventually invoke the returned unlock function, or all future calls with the same key will block
// indefinitely
```

api/clients/v2/relay/default_relay_url_provider.go
```diff
@@ -0,0 +1,57 @@
...
+
+	"github.com/Layr-Labs/eigenda/common"
+	relayRegistryBindings "github.com/Layr-Labs/eigenda/contracts/bindings/EigenDARelayRegistry"
+	v2 "github.com/Layr-Labs/eigenda/core/v2"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	gethcommon "github.com/ethereum/go-ethereum/common"
+)
+
+// DefaultRelayUrlProvider provides relay URL strings, based on relay key.
+type DefaultRelayUrlProvider struct {
```
ian-shim, line 15: `DefaultRelayUrlProvider` sounds like there should be other relay URL provider implementations. Maybe we can just call it `relayURLProvider` unexported? 
`NewRelayURLProvider` can return the struct pointer as RelayURLProvider type.

api/clients/v2/relay_client.go
```diff
@@ -234,53 +242,69 @@ func (c *relayClient) getClient(key corev2.RelayKey) (relaygrpc.RelayClient, err
...
+		// that means that a different caller did the necessary work already
+		return nil
+	}
+
+	// TODO (litt3): immediately storing `true` for key mirrors the previous implementation, where a failed init
+	//  is NOT retried, and the connection to that relay will just be broken forever. Consider implementing
+	//  logic to retry initialization after a period of time in case of failure
+	c.relayInitializationStatus.Store(key, true)
+
+	relayUrl, err := c.relayUrlProvider.GetRelayUrl(ctx, key)
```
ian-shim, line 271: What happens if this fails? `relayInitializationStatus` for this key has been set to true and `grpcRelayClients.Load` will continue failing, right?

api/clients/v2/relay/default_relay_url_provider.go
ian-shim: nit: not sure how useful this abstraction is. It simply wraps the eth client to fetch relay info from chain. Could we just use EthClient straight from the relay client?

api/clients/v2/relay_client.go
litt3: That's how the current implementation behaves, yes. 

Do you think it makes more sense to rework the method, such that a failed fetch doesn't result in the initialization flag being set to `true`?

Callers would therefore infinitely retry the initialization (lazily), with the rate of retries being limited by the timeout configured on the context.

litt3: Or, alternatively, I could implement more specific retry logic to limit the rate.

api/clients/v2/relay/default_relay_url_provider.go
litt3: > nit: not sure how useful this abstraction is

There are two reasons I thought it makes sense to create this abstraction:
1. it allows for a test implementation of the functionality
2. In the (near?) future, the relay registry will need to be updated with functionality to remove and update entries. At that time, there will be much more logic, and making this split now means that future changes to the `RelayURLProvider` won't require changes to the relay client

litt3: > DefaultRelayUrlProvider sounds like there should be other relay URL provider implementations

Did you notice that there is a `TestRelayUrlProvider` implementation? That was the reason I chose the `Default` prefix, curious if that affects your opinion at all

api/clients/v2/relay_client.go
litt3: Sounds good- done

api/clients/v2/relay/default_relay_url_provider.go
litt3: I renamed to `relayURLProvider`, and moved the test implementation into the `test` package

api/clients/v2/relay/default_relay_url_provider.go
```diff
@@ -0,0 +1,58 @@
+package relay
```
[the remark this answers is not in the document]
litt3: Fixed

cody-littley (approved):
LGTM

PR #1335, opened 2025-02-27
Type `accountID` as Address instead of string

ian-shim:
Typing `accountID` as Address (`[20]byte`) instead of string 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/apiserver/disperse_blob_v2.go
+++ b/disperser/apiserver/disperse_blob_v2.go
@@ -15,6 +15,7 @@ import (
 	dispv2 "github.com/Layr-Labs/eigenda/disperser/common/v2"
 	"github.com/Layr-Labs/eigenda/encoding"
 	"github.com/Layr-Labs/eigenda/encoding/rs"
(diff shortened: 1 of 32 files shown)


disperser/apiserver/disperse_blob_v2.go
```diff
@@ -114,7 +115,7 @@ func (s *DispersalServerV2) checkPaymentMeter(ctx context.Context, req *pb.Dispe
 	accountID := blobHeaderProto.GetPaymentHeader().GetAccountId()
```
hopeyen, line 115: is it worth doing the same check here as you did for req.AccountId?
```
	if !gethcommon.IsHexAddress(req.AccountId) {
		return nil, api.NewErrorInvalidArg("invalid account ID")
	}
```

ian-shim: Good catch. Updated!

hopeyen (approved):

anupsv (approved):

PR #1338, opened 2025-02-27
Make sure each relay `GetChunks` request is either by index or by range

ian-shim:
With protobuf's `oneof` field, it's possible neither `ByIndex` or `ByRange` is set. This PR adds a validation to ensure either field is set. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/relay/server.go
+++ b/relay/server.go
@@ -211,6 +211,30 @@ func (s *Server) GetBlob(ctx context.Context, request *pb.GetBlobRequest) (*pb.G
 	return reply, nil
 }
 
+func (s *Server) validateGetChunksRequest(request *pb.GetChunksRequest) error {
+	if request == nil {
+		return api.NewErrorInvalidArg("request is nil")
+	}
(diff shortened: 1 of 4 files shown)


cody-littley (approved):

relay/server.go
```diff
@@ -211,6 +211,30 @@ func (s *Server) GetBlob(ctx context.Context, request *pb.GetBlobRequest) (*pb.G
 	return reply, nil
 }
 
+func (s *Server) validateGetChunksRequest(request *pb.GetChunksRequest) error {
+	if request == nil {
+		return api.NewErrorInvalidArg("request is nil")
+	}
+	if len(request.OperatorId) == 0 {
```
jianoaix, line 218: If it's unauthenticated request, this may be empty

ian-shim: Good call

jianoaix (approved):

PR #1341, opened 2025-02-28
Check multiplication overflow in `ValidateEncodingParams`

ian-shim:
Ensure that `params.ChunkLength*params.NumChunks` doesn't result in overflow

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/encoding/params.go
+++ b/encoding/params.go
@@ -3,6 +3,7 @@ package encoding
 import (
 	"errors"
 	"fmt"
+	"math"
 
 	"golang.org/x/exp/constraints"
 )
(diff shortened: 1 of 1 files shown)


encoding/params.go
```diff
@@ -61,6 +63,9 @@ func GetNumSys(dataSize uint64, chunkLen uint64) uint64 {
 
 // ValidateEncodingParams takes in the encoding parameters and returns an error if they are invalid.
 func ValidateEncodingParams(params EncodingParams, SRSOrder uint64) error {
+	if params.ChunkLength != 0 && params.NumChunks > maxUint64/params.ChunkLength {
```
anupsv, line 66: Is there no relation between the blob size we allow and the max num of chunks based on coding ratio ? calculating upto max uint seems excessive?

ian-shim: There should be, but they're configured parameters and can change. This is just to ensure that overflow does not happen

anupsv (approved):

PR #1346, opened 2025-02-28
Calculate Charge at the payment

yujiezhu0:
Why are these changes needed?

- Calculate the charge at the time of payment and replace dataLength with the computed charge, storing it in DynamoDB.
- Resolves the issue where updates to pricePerSymbol would otherwise impact existing payments.
- Linear: EGDA-990

Checks
[✔] I've made sure the tests are passing. Note that there might be a few flaky tests, in that case, please comment that they are not relevant.
[✔] I've checked the new test coverage and the coverage percentage didn't drop.
Testing Strategy
[✔] Unit tests
Integration tests
This PR is not tested :(


core/meterer/meterer.go
```diff
@@ -214,14 +214,20 @@ func (m *Meterer) ServeOnDemandRequest(ctx context.Context, header core.PaymentM
 		return fmt.Errorf("invalid quorum for On-Demand Request: %w", err)
 	}
 
+	// Compute charge at payment time
```
hopeyen, line 217: is it possible to move this block into the deleted `PaymentCharged` helper fn? Would like to keep those unit tests if possible?

hopeyen (commented):
looking good!

core/meterer/meterer.go
```diff
@@ -214,14 +214,20 @@ func (m *Meterer) ServeOnDemandRequest(ctx context.Context, header core.PaymentM
...
+	charge := m.PaymentCharged(big.NewInt(int64(symbolsCharged)), big.NewInt(int64(m.ChainPaymentState.GetPricePerSymbol())))
```
hopeyen, line 218: Hmm, looking at the function `PaymentCharged`, I don't think you need to create a separate big.NewInt for the two inputs. 
And `m.ChainPaymentState.GetPricePerSymbol()` can be accessed within `PaymentCharged` because it is part of the `meterer` function so you don't need to pass it in. Alternatively can make this function standalone from `meterer` struct so accountant can use this helper as well

core/meterer/meterer_test.go
```diff
@@ -323,10 +323,11 @@ func TestMetererOnDemand(t *testing.T) {
 
 	// test duplicated cumulative payments
 	symbolLength := uint64(100)
-	priceCharged := mt.PaymentCharged(symbolLength)
+	symbolsCharged := mt.SymbolsCharged(symbolLength)
+	priceCharged := mt.PaymentCharged(big.NewInt(int64(symbolsCharged)), big.NewInt(int64(mt.ChainPaymentState.GetPricePerSymbol())))
```
hopeyen, line 327: similar comment as the above and for the other priceCharged calculations

hopeyen (commented):
very minor comments:)

core/meterer/meterer.go
yujiezhu0: Thx. Will go with making it standalone.

anupsv: Could we change the variable name to something more descriptive. `charge` is quite hard to read and understand at first go 😅

yujiezhu0: Do you have suggestions? How about `paymentCharged`?

core/meterer/meterer.go
```diff
@@ -214,14 +214,20 @@ func (m *Meterer) ServeOnDemandRequest(ctx context.Context, header core.PaymentM
...
+	if charge.Cmp(big.NewInt(0)) <= 0 {
+		return fmt.Errorf("invalid charge value: must be positive, got %s", charge.String())
+	}
```
anupsv, line 221: Interesting check. Can the `charge` value be less than 0 for any case ?

yujiezhu0: Actually, it can be removed, since uint64 will not be negative

hopeyen (approved):
lgtm

anupsv (approved):

PR #1348, opened 2025-02-28
chore: move autogenerated api protobuf mdfiles to mdbook

samlaf:

Closes EGDA-821.

Updated the protobuf generation script to generate the protobuf md files in the mdbook src dir instead of in ./api/docs.
Added `make mdbook-serve` command to generate a preview of the mdbook.
Book currently looks like:
image

Might want to group together the v1 and v2 files...? I'm just not sure what to do with common.md since v2 also uses some of those msgs afaiu.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


jianoaix (approved):
Can we add a simple documentation (maybe README) on what to run/update when one changes the `.proto` files?

samlaf:
> Can we add a simple documentation (maybe README) on what to run/update when one changes the `.proto` files?

Added README documentation in [be124e1](https://github.com/Layr-Labs/eigenda/pull/1348/commits/be124e1919d879939553dbdeb7e9f310373637c8)

cody-littley (commented):
Can you also update the `.gitattributes` file in the root of the repo? Currently it ignores `.html` and `.md` files in `api/docs`, and I'm not sure if the current patterrn will match to the other directories.

samlaf:
> Can you also update the `.gitattributes` file in the root of the repo? Currently it ignores `.html` and `.md` files in `api/docs`, and I'm not sure if the current patterrn will match to the other directories.

Done: [baa3677](https://github.com/Layr-Labs/eigenda/pull/1348/commits/baa3677baaadaa0cd6cc3d0106b357e1fbadf3a6)

PR #1350, opened 2025-03-03
reservation period index

hopeyen:

Proposed fix for https://linear.app/eigenlabs/issue/EGDA-994/eda2-08-%5Bmedium%5D-changes-in-reservation-period-interval-a%EF%AC%80ects-current

I didn't like any of the proposed fixes by the audit. The approach I'm taking here is to change reservation period index to the start of the reservation period timestamp, by getting the nearest lower multiple of the period interval from the payment header timestamp. I've described my reasonings in the linear ticket

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


core/meterer/meterer.go
```diff
@@ -147,9 +147,10 @@ func (m *Meterer) ValidateReservationPeriod(reservation *core.ReservedPayment, r
 	reservationWindow := m.ChainPaymentState.GetReservationWindow()
 	currentReservationPeriod := GetReservationPeriod(receivedAt.Unix(), reservationWindow)
 	// Valid reservation periodes are either the current bin or the previous bin
```
ian-shim, line 149: nit: typo `periodes`

core/meterer/meterer.go
```diff
@@ -147,9 +147,10 @@ func (m *Meterer) ValidateReservationPeriod(reservation *core.ReservedPayment, r
...
-	isCurrentOrPreviousPeriod := requestReservationPeriod == currentReservationPeriod || requestReservationPeriod == (currentReservationPeriod-1)
+	isCurrentOrPreviousPeriod := requestReservationPeriod == currentReservationPeriod || requestReservationPeriod == (currentReservationPeriod-reservationWindow)
```
ian-shim, line 150: should this be 
```
isCurrentOrPreviousPeriod := (currentReservationPeriod-reservationWindow) <=  requestReservationPeriod && requestReservationPeriod < currentReservationPeriod
```
?

hopeyen: hmm, could you please explain why you think so?
The check here is to make sure the request's reservation period is either the current one or the previous one. I think the check you proposed is for the request's reservation period to be the previous reservation window or even older?

ian-shim: If `reservationWindow` changed after `requestReservationPeriod` was calculated, would the periods (start times) align? 
Could `requestReservationPeriod` be calculated using one `reservationWindow` and `currentReservationPeriod` be calculated using a different `reservationWindow`?

hopeyen: Whether the periods would align depends on the window change, if timestamp is 6 and window changed from 2 to 3, then the period start time wouldn't align. This means meterer will be using a fresh bucket, and I think that's reasonable when a `reservationWindow` change go through

The second question is a good point! I've committed changes that read reservationWindow only once throughout a reservation metering so only 1 window will be used

ian-shim (approved):
I think ci should pass if you rebase

PR #1351, opened 2025-03-03
Fix how relay/validators compute hashes.

cody-littley:

Fix how relays & validators hash various messages for blazar release.

### DIFF
--- a/api/hashing/node_hashing.go
+++ b/api/hashing/node_hashing.go
@@ -1,6 +1,7 @@
 package hashing
 
 import (
+	"fmt"
 	"hash"
 
 	commonv1 "github.com/Layr-Labs/eigenda/api/grpc/common"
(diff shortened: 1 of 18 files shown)


api/proto/relay/relay.proto
```diff
@@ -42,15 +42,19 @@ message GetChunksRequest {
   // All integers are encoded as unsigned 4 byte big endian values.
   //
   // Perform a keccak256 hash on the following data in the following order:
+  // 1. the length of the operator ID in bytes
```
litt3, line 45: Numbering is off: duplicate `1.`s, and the sublists each have duplicate `ii.`s

litt3 (approved):

api/proto/relay/relay.proto
cody-littley: fixed

litt3: Nope, looks like there are still two `1.` entries

api/hashing/node_hashing.go
```diff
@@ -12,49 +13,109 @@ import (
...
+func hashBlobCertificate(hasher hash.Hash, blobCertificate *common.BlobCertificate) error {
+	err := hashBlobHeader(hasher, blobCertificate.GetBlobHeader())
+	if err != nil {
+		return fmt.Errorf("failed to hash blob header: %w", err)
+	}
+	err = hashByteArray(hasher, blobCertificate.GetSignature())
+	if err != nil {
+		return fmt.Errorf("failed to hash signature: %w", err)
 	}
+	err = hashUint32Array(hasher, blobCertificate.GetRelayKeys())
```
ian-shim, line 47: do we need to hash the length of the relay keys too?

api/hashing/relay_hashing.go
```diff
@@ -1,38 +1,49 @@
...
-			hasher.Write(iByte)
-			hasher.Write(getByIndex.BlobKey)
-			for _, index := range getByIndex.ChunkIndices {
-				hashUint32(hasher, index)
+			hashChar(hasher, 'i')
+			err = hashByteArray(hasher, getByIndex.BlobKey)
+			if err != nil {
+				return nil, fmt.Errorf("failed to hash blob key: %w", err)
+			}
+			err = hashUint32Array(hasher, getByIndex.ChunkIndices)
```
ian-shim, line 32: here as well?

api/hashing/node_hashing.go
cody-littley: `hashUint32Array()` includes a hash of the length.

```
// hashUint32Array hashes the given uint32 array.
func hashUint32Array(hasher hash.Hash, values []uint32) error {
	if len(values) > math.MaxUint32 {
		return fmt.Errorf("uint32 array is too long: %d", len(values))
	}

	err := hashLength(hasher, values)
	if err != nil {
		return fmt.Errorf("failed to hash length: %w", err)
	}
	for _, value := range values {
		hashUint32(hasher, value)
	}

	return nil
}
```

PR #1355, opened 2025-03-03
Introduce feed cache

jianoaix:
We have three APIs at dataapi that will operate on feed: blob feed, batch feed, and signing-rate

The reasons that feed caching matters are:
- Performance: at blob rate up to 100 blobs/sec, attestation rate up to 1 attestation/sec, it’ll have a lot of data for 24h, not to mention 14 days (lifetime of data in EigenDA).
- Scaling: dataapi will not only be the backend for Blob Explorer, but also third-party Apps that integrate with EigenLabs-operated Disperser, which will demand higher scale than before (v1) — we have to drive down the cost of serving request in order to serve larger scale traffic.
- Robustness: more importantly, the dataapi is operating on the same database backend as disperser so we want to make sure this won’t affect the availability of disperser as dataapi scales.

The design is to:
- Maintain the longest (up to cache size limit) most recent consecutive interval of results in cache
- Evicts the oldest data (the left endpoint of the interval) when cache limit is reached

Our key hypothesis is that recent data is most often requested, and a longer segment of recent data provides a higher chance of cache hit. (effectiveness)

This approach is simple as it maintains just one segment (but will extend the segment).

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


dmanc (approved):
Looks great. Probably we should add some metrics for cache misses and hits.

jianoaix:
> Looks great. Probably we should add some metrics for cache misses and hits.

yea these are needed, will be in a followup

PR #1365, opened 2025-03-05
Replay protection

cody-littley:

Prevents replay attacks for signed messages.


common/replay/replay_guardian.go
```diff
@@ -0,0 +1,169 @@
...
+	// If we've reached this point, then the request will still be in the observedHashes set if it is a replay.
+	if _, ok := r.observedHashes[string(requestHash)]; ok {
+		return fmt.Errorf("request hash %x has been observed before", requestHash)
+	}
+
+	// The request is not a replay. Add the hash to the observedHashes set and the expirationQueue.
+	r.observedHashes[string(requestHash)] = struct{}{}
+	r.expirationQueue.Enqueue(&hashWithExpiration{
+		hash:       string(requestHash),
+		expiration: now.Add(r.maxTimeInPast),
```
litt3, line 122: It seems like defining the expiration based on `maxTimeInPast` is kinda arbitrary? Couldn't it just as well be `maxTimeInFuture`, or `maxTimeInPast + maxTimeInFuture`? Or a separately configured value entirely? Why `maxTimeInPast`

common/replay/replay_guardian.go
```diff
@@ -0,0 +1,169 @@
...
+// ReplayGuardian ensures that the same request is not processed more than once. It can be used to do things such
+// as protecting against replay attacks or accidental duplicate requests.
+type ReplayGuardian interface {
+
+	// VerifyRequest verifies that a request with the given hash and timestamp is not a replay
+	// of a previous request. If it cannot be determined if a request is a replay or not,
+	// then the request is rejected. Only if it can be guaranteed that the request is not a replay
+	// will this method return nil.
+	//
+	// In order to be a verified unique request, the following conditions must be met:
```
litt3, line 20: It seems this description after the first paragraph is implementation specific - given that there is an interface defined, this doc might be better placed alongside the implementation the contains this logic

node/auth/authenticator.go
```diff
@@ -3,27 +3,26 @@ package auth
...
-	"time"
 )
 
 // RequestAuthenticator authenticates requests to the DA node. This object is thread safe.
+// Returns the hash of the request and an error if the request is invalid.
 //
 // This class has largely been future-proofed for decentralized dispersers, with the exception of the
 // preloadCache method, which will need to be updated to handle decentralized dispersers.
 type RequestAuthenticator interface {
 	// AuthenticateStoreChunksRequest authenticates a StoreChunksRequest, returning an error if the request is invalid.
```
litt3, line 21: doc wasn't updated to contain the meaning of the returned bytes

node/auth/authenticator_test.go
```diff
@@ -516,8 +325,12 @@ func TestKeyCacheSize(t *testing.T) {
 	require.NoError(t, err)
 	request.Signature = signature
 
-	err = authenticator.AuthenticateStoreChunksRequest(context.Background(), "0", request, start)
+	hash, err = authenticator.AuthenticateStoreChunksRequest(context.Background(), request, start)
+	require.NoError(t, err)
+	require.NoError(t, err)
```
litt3, line 330: duplicate no error check

litt3 (approved):

common/replay/replay_guardian.go
```diff
@@ -0,0 +1,169 @@
...
+// maxTimeInFuture is the maximum amount of time that a request's timestamp can be ahead of the local wall clock time.
+// Increasing this value permits more leniency in the timestamp of incoming requests, at the potential cost of a higher
+// memory overhead. In theory, if requests are sent exactly at the maximum age, this utility will remember them for
+// that amount of time. If this is extremely large, then an attacker may be able to cause this utility to be forced
+// to remember a very large amount of data.
+//
+// maxTimeInPast is the maximum amount of time that a request's timestamp can be behind the local wall clock time.
+// Increasing this value permits more leniency in the timestamp of incoming requests, at the potential cost of a higher
+// memory overhead.
+func NewReplayGuardian(
```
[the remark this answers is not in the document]
cody-littley: Changed as we discussed

common/replay/replay_guardian.go
```diff
@@ -0,0 +1,169 @@
...
+	VerifyRequest(
+		requestHash []byte,
+		requestTimestamp time.Time) error
+}
+
+// NewReplayGuardian creates a new ReplayGuardian.
+//
+// maxTimeInFuture is the maximum amount of time that a request's timestamp can be ahead of the local wall clock time.
+// Increasing this value permits more leniency in the timestamp of incoming requests, at the potential cost of a higher
+// memory overhead. In theory, if requests are sent exactly at the maximum age, this utility will remember them for
```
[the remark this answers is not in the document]
cody-littley: Updated wording. Let me know if it is still unclear.

```
// maxTimeInFuture is the maximum amount of time that a request's timestamp can be ahead of the local wall clock time.
// Increasing this value permits more leniency in the timestamp of incoming requests, at the potential cost of a higher
// memory overhead. In theory, if requests are sent exactly at the maximum time in the future, this utility will
// remember them for a total of (maxTimeInFuture + maxTimeInPast), since that is the amount of time that will need to
// elapse locally before the request exceeds the maximum age. If maxTimeInFuture is extremely large, then an attacker 
// may be able to cause this utility to be forced to remember a very large amount of data.
```

common/replay/replay_guardian.go
```diff
@@ -0,0 +1,169 @@
...
+		if timeInPast > r.maxTimeInPast {
+			return fmt.Errorf("request timestamp %v is too far in the past. "+
+				"Current time is %v, maximum time in the past is %v",
+				requestTimestamp, now, r.maxTimeInPast)
+		}
+	}
+	return nil
+}
+
+// pruneObservedHashes removes any hashes from the observedHashes set that have expired. A hash is considered to have
```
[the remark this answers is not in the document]
cody-littley: Rephrased.

```
// pruneObservedHashes removes any hashes from the observedHashes set that have expired. A hash is considered to have
// expired if its expiration time is before the current wall clock time.
```

common/replay/replay_guardian.go
cody-littley: Fixed based on our discussion

common/replay/replay_guardian.go
cody-littley: Makes sense, the second paragraph has been removed from the interface.

node/auth/authenticator.go
cody-littley: Updated method doc and removed the description from the top level interface.

node/auth/authenticator_test.go
cody-littley: I just want to be REALLY sure there wasn't an error. Perhaps I should spin here in a for loop just in case. (fixed)

common/replay/replay_guardian_impl.go
```diff
@@ -0,0 +1,166 @@
...
+	// The maximum amount of time that a request's timestamp can be ahead of the local wall clock time.
+	maxTimeInFuture time.Duration
+
+	// The maximum amount of time that a request's timestamp can be behind the local wall clock time.
+	maxTimeInPast time.Duration
+
+	// A set of hashes that have been observed within the time window.
+	observedHashes map[string]struct{}
+
+	// A queue of observed hashes, ordered by expiration time. Used to prune old hashes.
```
litt3, line 27: > ordered by expiration time

This is technically true, due to how expiration time directly relates to request time. But it would probably be clearer to describe this queue as ordered by the `requestTimestamp`

common/replay/replay_guardian_impl.go
```diff
@@ -0,0 +1,166 @@
...
+
+// VerifyRequest verifies that a request with the given hash and timestamp is not a replay
+// of a previous request. If it cannot be determined if a request is a replay or not,
+// then the request is rejected. Only if it can be guaranteed that the request is not a replay
+// will this method return nil.
+//
+// In order to be a verified unique request, the following conditions must be met:
+// - the request's timestamp must be no more than X minutes ahead of the local wall clock time
+// - the request's timestamp must be no more than Y minutes behind the local wall clock time
+// - the request's hash must not have been previously observed (hashes are remembered for Y minutes)
```
litt3, line 89: > (hashes are remembered for Y minutes)

I don't think this is precisely correct: hashes are remembered for `0 <= t <= X + Y` minutes. Or, put another way, until `t + Y < now`

node/auth/authenticator_test.go
```diff
@@ -321,117 +209,35 @@ func TestKeyExpiry(t *testing.T) {
...
 
 	// The key should not yet have been fetched again.
 	chainReader.Mock.AssertNumberOfCalls(t, "GetDisperserAddress", 1)
 
 	// Move time forward to just after the key expires.
 	now = now.Add(2 * time.Second)
-	err = authenticator.AuthenticateStoreChunksRequest(context.Background(), "localhost", request, now)
+	hash, err = authenticator.AuthenticateStoreChunksRequest(context.Background(), request, now)
 	require.NoError(t, err)
+	require.NoError(t, err)
```
litt3, line 234: here is another duplicate `require.NoError(t, err)`

litt3 (approved):

common/replay/replay_guardian_impl.go
cody-littley:
```
	// A queue of observed hashes, ordered by request timestamp. Used to prune old hashes.
```

common/replay/replay_guardian_impl.go
cody-littley: How about this?
```
// - the request's hash must not have been previously observed (hashes are remembered until they are Y in the past)
```

node/auth/authenticator_test.go
cody-littley: fixed

litt3 (approved):

PR #1368, opened 2025-03-06
Implement cert verifier address provider

litt3:
Closes EGDA-979

- As described in [this issue](https://linear.app/eigenlabs/issue/EGDA-979/payload-dispersal-must-check-selected-rbn-to-see-if-a-new-cert), a single cert verifier address isn't enough for payload dispersal: the active address may change after the disperser chooses an RBN, once dynamic cert verifier changes are possible
- To future proof the client APIs for when dynamic cert verifier addresses are supported, this PR makes necessary API changes, so that a `CertVerifierAddressProvider` is used throughout the clients, instead of manually passing in an address for each call
    - This change represents a nice simplification of the APIs, in addition to the benefit of supporting future efforts
- This PR also does a lot of work organizing code into packages. These changes were necessary, so that the new structures could be put into reasonable packages without introducing circular dependencies. IMO the changes in this PR represent a substantial improvement to the package structure of the clients.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/clients/v2/payloaddispersal/payload_disperser_config.go
```diff
@@ -0,0 +1,66 @@
...
+		BlobStatusPollInterval: 1 * time.Second,
+		ContractCallTimeout:    5 * time.Second,
+	}
+}
+
+// checkAndSetDefaults checks an existing config struct. It performs one of the following actions for any contained 0 values:
+//
+// 1. If 0 is an acceptable value for the field, do nothing.
+// 2. If 0 is NOT an acceptable value for the field, and a default value is defined, then set it to the default.
+// 3. If 0 is NOT an acceptable value for the field, and a default value is NOT defined, return an error.
```
cody-littley, line 45: I'm guessing this part is in case you add a value without a sane default in the future? Doesn't currently have any such values. `and a default value is NOT defined, return an error.`

api/clients/v2/verification/cert_verifier.go
```diff
@@ -219,105 +168,82 @@ func (cv *CertVerifier) GetNonSignerStakesAndSignature(
...
-	if cv.pollInterval <= 0 {
-		// don't wait for the internal client to advance
-		return nil
-	}
-
-	if cv.latestBlockNumber.Load() >= targetBlockNumber {
-		// immediately return if the local client isn't behind the target block number
-		return nil
+// This method caches ContractEigenDACertVerifierCaller instances, since their construction requires acquiring a lock
+// and parsing json, and is therefore not trivially inexpensive.
```
cody-littley, line 221: nit, `non-trivially`?

cody-littley (approved):

api/clients/v2/payloaddispersal/payload_disperser_config.go
litt3: I simplified these comments to the following:

```
// checkAndSetDefaults checks an existing config struct. If a given field is 0, and 0 is not an acceptable value, then
// this method sets it to the default.
```

api/clients/v2/verification/cert_verifier.go
litt3: Changed to "non-trivially expensive" (which I believe carries the same meaning of "not trivially inexpensive", but is easier to parse)

cody-littley (approved):

api/clients/v2/cert_verifier_address_provider.go
```diff
@@ -0,0 +1,10 @@
+package clients
+
+// CertVerifierAddressProvider defines an object which can translate block number to cert verifier address
+//
+// This provider uses block number as a key, since updates to a cert verifier address in a running system are
+// coordinated by defining the block number at which a new cert verifier address takes effect.
+type CertVerifierAddressProvider interface {
+	// GetCertVerifierAddress returns the EigenDACertVerifierAddress that is active at the input block number
+	GetCertVerifierAddress(blockNumber uint64) (string, error)
```
samlaf, line 9:
```suggestion
	GetCertVerifierAddress(blockNumber uint64) (common.Address, error)
```

samlaf: would using geth's Address type make sense here? Typically prefer this over a string

api/clients/v2/payloaddispersal/payload_disperser_config.go
```diff
@@ -0,0 +1,63 @@
...
+	// GetBlobStatus.
+	BlobStatusPollInterval time.Duration
+
+	// The timeout duration for contract calls
+	ContractCallTimeout time.Duration
+}
+
+// GetDefaultPayloadDisperserConfig creates a PayloadDisperserConfig with default values
+func GetDefaultPayloadDisperserConfig() *PayloadDisperserConfig {
+	return &PayloadDisperserConfig{
```
samlaf, line 32: if this is only used in checkAndSetDefaults below, then prob can keep private?

api/clients/v2/payloadretrieval/relay_payload_retriever_test.go
```diff
@@ -35,18 +35,18 @@ const maxPayloadBytes = 1025 // arbitrary value
 
 type RelayPayloadRetrieverTester struct {
 	Random                *testrandom.TestRandom
-	RelayPayloadRetriever *clients.RelayPayloadRetriever
+	RelayPayloadRetriever *RelayPayloadRetriever
 	MockRelayClient       *clientsmock.MockRelayClient
 	G1Srs                 []bn254.G1Affine
-	Config                clients.RelayPayloadRetrieverConfig
+	Config                RelayPayloadRetrieverConfig
```
samlaf, line 41: Is this config used for anything? Or is it just passed to the RelayPayloadRetriever constructor? Why keep it around?

api/clients/v2/payloaddispersal/payload_disperser.go
```diff
@@ -67,18 +60,17 @@ func NewPayloadDisperser(
...
+) (*coretypes.EigenDACert, error) {
 	blob, err := payload.ToBlob(pd.config.PayloadPolynomialForm)
 	if err != nil {
 		return nil, fmt.Errorf("convert payload to blob: %w", err)
 	}
 
 	timeoutCtx, cancel := context.WithTimeout(ctx, pd.config.ContractCallTimeout)
 	defer cancel()
-	requiredQuorums, err := pd.requiredQuorumsStore.GetQuorumNumbersRequired(timeoutCtx, certVerifierAddress)
+	requiredQuorums, err := pd.certVerifier.GetQuorumNumbersRequired(timeoutCtx)
```
anupsv, line 73: This utilizes the cache righT?

test/v2/correctness/correctness_test.go
```diff
@@ -37,25 +37,27 @@ func getEnvironmentName(environment string) string {
 	return environmentName
 }
 
+// checkAndSetCertVerifierAddress checks whether the input address string is empty, and skips the test if it is
+//
+// If the input address is not empty, this method configures the test client to use the input address as the cert
+// verifier address.
+func checkAndSetCertVerifierAddress(t *testing.T, c *client.TestClient, certVerifierAddress string) {
+	if certVerifierAddress == "" {
+		t.Skip("Requested cert verifier address is not configured")
```
samlaf, line 46: first time I see this kind of pattern of skipping a test inside some function. Is this something you've used before? Curious if there would be another way to skip the test earlier and inside the test itself.

api/clients/v2/verification/cert_verifier.go
```diff
@@ -15,58 +16,29 @@ import (
...
-	latestBlockNumber atomic.Uint64
-	// atomic bool, so that only a single goroutine is polling the internal client with BlockNumber() calls at any given time
-	pollingActive atomic.Bool
+	logger                      logging.Logger
+	ethClient                   common.EthClient
+	blockNumberProvider         *BlockNumberProvider
+	certVerifierAddressProvider clients.CertVerifierAddressProvider
 	// maps contract address to a ContractEigenDACertVerifierCaller object
 	verifierCallers sync.Map
+	// maps cert verifier address string to set of required quorums specified in the contract at that address
```
samlaf, line 30: nit: use same doc for requiredQuorums and verifierCallers

`maps contract address`
vs
`maps cert verifier address`

to make it obvious they are the same. consider also potentially merging into a single map whose value is a struct containing both caller + requiredQuorums?

api/clients/v2/verification/cert_verifier.go
```diff
@@ -149,18 +111,10 @@ func (cv *CertVerifier) VerifyCertV2FromSignedBatch(
...
-	// the hex address of the EigenDACertVerifier contract
-	certVerifierAddress string,
-	eigenDACert *EigenDACert,
-) error {
-	err := cv.MaybeWaitForBlockNumber(ctx, uint64(eigenDACert.BatchHeader.ReferenceBlockNumber))
-	if err != nil {
-		return fmt.Errorf("wait for block number: %w", err)
-	}
+func (cv *CertVerifier) VerifyCertV2(ctx context.Context, eigenDACert *coretypes.EigenDACert) error {
+	blockNumber := uint64(eigenDACert.BatchHeader.ReferenceBlockNumber)
```
samlaf, line 115:
```suggestion
	referenceBlockNumber := uint64(eigenDACert.BatchHeader.ReferenceBlockNumber)
```

api/clients/v2/verification/cert_verifier.go
```diff
@@ -189,21 +143,16 @@ func (cv *CertVerifier) VerifyCertV2(
...
+	signedBatchBinding, err := coretypes.SignedBatchProtoToBinding(signedBatch)
 	if err != nil {
 		return nil, fmt.Errorf("convert signed batch: %w", err)
 	}
 
-	err = cv.MaybeWaitForBlockNumber(ctx, signedBatch.GetHeader().GetReferenceBlockNumber())
-	if err != nil {
-		return nil, fmt.Errorf("wait for block number: %w", err)
-	}
+	blockNumber := signedBatch.GetHeader().GetReferenceBlockNumber()
```
samlaf, line 153:
```suggestion
	referenceBlockNumber := signedBatch.GetHeader().GetReferenceBlockNumber()
```

samlaf: RBN as a concept is very meaningful to most eigenda engineers, so good to use it as the variable name

api/clients/v2/verification/block_number_provider.go
```diff
@@ -0,0 +1,136 @@
...
+// configured pollInterval is <= 0, in which case this method will NOT wait for the internal client to advance.
+//
+// This method will check the current block number of the internal client every pollInterval duration.
+// It will return nil if the internal client advances to (or past) the targetBlockNumber. It will return an error
+// if the input context times out, or if any error occurs when checking the block number of the internal client.
+//
+// This method is synchronized in a way that, if called by multiple goroutines, only a single goroutine will actually
+// poll the internal eth client for the most recent block number. The goroutine responsible for polling at a given time
+// updates an atomic integer, so that all goroutines may check the most recent block without duplicating work.
+func (bnp *BlockNumberProvider) MaybeWaitForBlockNumber(ctx context.Context, targetBlockNumber uint64) error {
```
anupsv, line 57: The `maybe` is quite confusing until i went through the logic twice. Idk if we can rename it or soemthing.

api/clients/v2/verification/cert_verifier.go
```diff
@@ -219,105 +168,82 @@ func (cv *CertVerifier) GetNonSignerStakesAndSignature(
...
+// VerifyDACertV2 and verifyDACertV2FromSignedBatch.
+//
+// This method will return required quorum numbers from an internal cache if they are already known for the currently
+// active cert verifier. Otherwise, this method will query the required quorum numbers from the currently active
+// cert verifier, and cache the result for future use.
+func (cv *CertVerifier) GetQuorumNumbersRequired(ctx context.Context) ([]uint8, error) {
+	blockNumber, err := cv.blockNumberProvider.FetchLatestBlockNumber(ctx)
 	if err != nil {
-		return nil, fmt.Errorf("get verifier caller: %w", err)
+		return nil, fmt.Errorf("fetch latest block number: %w", err)
```
samlaf, line 180: iiuc this will be called from the proxy in the PUT route right?

Wouldn't it be best for this method to take an actual block number instead of fetching the latest block number? I tend to prefer libraries to be deterministic as much as possible, and "latest block" to be fetched from the outer layer. I don't have full context though so this comment might not make sense..

api/clients/v2/verification/cert_verifier.go
```diff
@@ -219,105 +168,82 @@ func (cv *CertVerifier) GetNonSignerStakesAndSignature(
...
 
-	ticker := time.NewTicker(cv.pollInterval)
-	defer ticker.Stop()
-
-	polling := false
-	if cv.pollingActive.CompareAndSwap(false, true) {
-		// no other goroutine is currently polling, so assume responsibility
-		polling = true
-		defer cv.pollingActive.Store(false)
+	certVerifierAddress, err := cv.certVerifierAddressProvider.GetCertVerifierAddress(blockNumber)
```
samlaf, line 231: shouldnt the MaybeWaitForBlockNumber be done inside the certVerifierAddressProvider? Otherwise what if these 2 providers use a different underlying eth client that arent at the same block as an example

api/clients/v2/verification/block_number_provider.go
```diff
@@ -0,0 +1,136 @@
...
+	"github.com/Layr-Labs/eigenda/common"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+)
+
+// BlockNumberProvider is a utility for interacting with the ethereum block number
+type BlockNumberProvider struct {
+	logger    logging.Logger
+	ethClient common.EthClient
+	// duration of interval when periodically polling the block number
+	pollIntervalDuration time.Duration
```
samlaf, line 18: add comment that this can be negative. btw are we sure we need this behavior of alowing negative numbers here?

samlaf (approved):
Looks like very nice cleanup. Have a bunch of nits but honestly would be happy merging in this state.

api/clients/v2/payloaddispersal/payload_disperser.go
litt3: Correct. The cache is now internal to the cert verifier, rather than being managed by the payload disperser.

api/clients/v2/verification/block_number_provider.go
litt3: Renamed to `WaitForBlockNumber` [ceee94b9](https://github.com/Layr-Labs/eigenda/pull/1368/commits/ceee94b935feaad6bb0cce01dc23bfd5d59053cd)

api/clients/v2/cert_verifier_address_provider.go
litt3: Yeah, I think that makes sense. [a01d8fed](https://github.com/Layr-Labs/eigenda/pull/1368/commits/a01d8fedb7c0a0abcd6b32a4e2aef5c3df7b4912)

api/clients/v2/payloaddispersal/payload_disperser_config.go
litt3: [f48dea34](https://github.com/Layr-Labs/eigenda/pull/1368/commits/f48dea34ea9e3a1b0364c347e2e1f9306a84420b)

api/clients/v2/verification/cert_verifier.go
litt3: Modified comment in [947a1b7e](https://github.com/Layr-Labs/eigenda/pull/1368/commits/947a1b7e45155cf039725d1b8a5c2bb1c52112a8)

I think it would increase the complexity of the logic to combine these maps into one: depending on the needs of the user, it's possible that they might end up populating the verifierCallers, but not the required quorums (if they are verifying, but not dispersing blobs). By keeping these maps separate, it's possible to infer from map membership whether the needed value exists, without needing to check the map value.

api/clients/v2/verification/cert_verifier.go
litt3: [a8e5cb09](https://github.com/Layr-Labs/eigenda/pull/1368/commits/a8e5cb09e4ba0f8371f3ea80fe590949d728862a)

api/clients/v2/verification/cert_verifier.go
litt3: > iiuc this will be called from the proxy in the PUT route right?

Not directly: `GetQuorumNumbersRequired` is used by `PayloadDisperser::SendPayload`. As currently organized, the `CertVerifier` manages all interactions with the cert verifier contract for the `PayloadDisperser`: the `PayloadDisperser` doesn't actually have access to an ethereum client.

From the perspective of the `PayloadDisperser`, I'm not able to think of a situation where we'd want anything other than the latest block number. Your point about deterministic libraries makes sense, but I'm leaning towards keeping the complexity hidden from the `PayloadDisperser` in this case.

Do you think it would make sense to rename the method to `GetLatestQuorumNumbersRequired`, to better convey that the result is dependent on the current state of the blockchain?

api/clients/v2/verification/block_number_provider.go
litt3: Fair point - I've added checks so that duration <= simply causes errors [9db13f7f](https://github.com/Layr-Labs/eigenda/pull/1368/commits/9db13f7ff69160b739241e6b8171a5db07cb7f6d)

api/clients/v2/verification/block_number_monitor.go
```diff
@@ -10,8 +10,11 @@ import (
 	"github.com/Layr-Labs/eigensdk-go/logging"
 )
 
-// BlockNumberProvider is a utility for interacting with the ethereum block number
-type BlockNumberProvider struct {
+// BlockNumberMonitor is a utility for waiting for a certain ethereum block number
+//
+// TODO: this utility is not currently in use, but DO NOT delete it. It will be necessary for the upcoming
```
cody-littley, line 15:
```suggestion
// TODO(litt): this utility is not currently in use, but DO NOT delete it. It will be necessary for the upcoming
```

cody-littley (approved):

api/clients/v2/verification/cert_verifier.go
samlaf: Ah I see, thanks for context. I guess you're right makes sense for the CertVerifier to manage contract interactions given our choice that the certVerification logic is actually stored onchain in a contract. I do think `GetLatestuorumNumbersRequired` would help here!

api/clients/v2/verification/block_number_monitor.go
samlaf: which litt? ;)

api/clients/v2/verification/cert_verifier.go
samlaf: Very nice. I definitely did not have such an complex change in mind but glad it helped you find this new abstraction :)

api/clients/v2/payloadretrieval/relay_payload_retriever_test.go
litt3: > Guessing buildBlobAndCert is a test function you use for testing only and need this config?

Correct. To make this less confusing, I've removed the config from the test harness, and simply put the single piece of info there that is actually needed by the test

[37c48c56](https://github.com/Layr-Labs/eigenda/pull/1368/commits/37c48c564e76b0b74c0442ace759efac024caf1c)

api/clients/v2/verification/block_number_monitor.go
litt3: Litt specified, and disambiguated

api/clients/v2/verification/cert_verifier.go
litt3: Hmm, so I've given this some more thought, and I've found a good reason to *not* change the name of this method to `GetLatestQuorumNumbersRequired`:

Once we implement the `CertVerifierRouter`, the values returned from this method will *not necessarily* represent the quorum set from the latest cert verifier contract. Rather, it will represent the union of quorums from all near-past, present, and near-future cert verifier contracts, to mitigate the race condition we've discussed in the dynamic case.

Therefore, I think the current, more generic `GetQuorumNumbersRequired` naming correctly represents what this method is doing: it's simply getting whatever quorums ought to be used by the `PaylodDisperser`, all factors considered.

samlaf: oh nice, that makes sense

test/v2/correctness/correctness_test.go
samlaf: we talked offline. Would prefer in the future that we don't have this kind of dynamic skipping of tests. But let's move on for now, more tests is always better than no tests!

samlaf (approved):

PR #1375, opened 2025-03-07
fix: use consistent query params across pages

jianoaix:
Manually narrowing down the range is unnecessary, i.e. no perf gains since DynamoDB handles the sortkeys efficiently already. Removing these makes it simpler/cleaner.

In addition, just updating the `start` will have issues for descending order retrieval.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


jianoaix:
ping

ian-shim (approved):

PR #1377, opened 2025-03-10
refactor: better check for existing payments

hopeyen:

Use conditional put to check that existing payment doesn't exist already. Improved from the previous get then put

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/meterer/offchain_store.go
+++ b/core/meterer/offchain_store.go
@@ -123,29 +123,24 @@ func (s *OffchainStore) UpdateGlobalBin(ctx context.Context, reservationPeriod u
 }
 
 func (s *OffchainStore) AddOnDemandPayment(ctx context.Context, paymentMetadata core.PaymentMetadata, paymentCharged *big.Int) error {
-	result, err := s.dynamoClient.GetItem(ctx, s.onDemandTableName,
-		commondynamodb.Item{
-			"AccountID":          &types.AttributeValueMemberS{Value: paymentMetadata.AccountID.Hex()},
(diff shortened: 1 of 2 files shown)


ian-shim (commented):
can we add a test?

hopeyen:
> can we add a test?

There's a test https://github.com/Layr-Labs/eigenda/blob/4af69d98ae102db62993cc57fca2798021f50250/core/meterer/meterer_test.go#L335
are you thinking of something else? just in offchain_store?

ian-shim (approved):

core/meterer/offchain_store.go
```diff
@@ -123,29 +123,24 @@ func (s *OffchainStore) UpdateGlobalBin(ctx context.Context, reservationPeriod u
...
+	item := commondynamodb.Item{
+		"AccountID":          &types.AttributeValueMemberS{Value: paymentMetadata.AccountID.Hex()},
+		"CumulativePayments": &types.AttributeValueMemberN{Value: paymentMetadata.CumulativePayment.String()},
+		"PaymentCharged":     &types.AttributeValueMemberN{Value: paymentCharged.String()},
 	}
-	if result != nil {
+
+	// Use a conditional put with attribute_not_exists to ensure atomicity
+	// This ensures we don't overwrite an existing record with the same key
+	condition := "attribute_not_exists(AccountID) AND attribute_not_exists(CumulativePayments)"
```
ian-shim, line 134: What should happen is an ondemand payment record already exists with a different cumulative payments? Should it allow the put or revert?

hopeyen: It should revert with the error `exact payment already exists`

PR #1386, opened 2025-03-13
fix: ondemand scheme

hopeyen:

addressing payment audit critical issue 2

- relax invariant to be always sufficient increment, doesn't allow out of order on-demand payments
- reduce the ondemand table scheme to be 1 record per account, record only tracks the largest cumulative payment
- conditional write to the table by checking existing record + new charge <= new cumulative payment

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/common/aws/dynamodb/client.go
+++ b/common/aws/dynamodb/client.go
@@ -50,6 +50,7 @@ type Client interface {
 	DeleteTable(ctx context.Context, tableName string) error
 	PutItem(ctx context.Context, tableName string, item Item) error
 	PutItemWithCondition(ctx context.Context, tableName string, item Item, condition string, expressionAttributeNames map[string]string, expressionAttributeValues map[string]types.AttributeValue) error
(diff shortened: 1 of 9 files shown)


common/aws/dynamodb/client.go
```diff
@@ -154,6 +155,33 @@ func (c *client) PutItemWithCondition(
 	return nil
 }
 
+// PutItemWithConditionAndReturn puts an item in the table with a condition and returns the old item if it exists
+func (c *client) PutItemWithConditionAndReturn(
```
ian-shim, line 159: would be nice to have a test

ian-shim (approved):

core/meterer/util.go
```diff
@@ -57,24 +57,6 @@ func CreateGlobalReservationTable(clientConfig commonaws.ClientConfig, tableName
 				KeyType:       types.KeyTypeHash,
 			},
 		},
-		GlobalSecondaryIndexes: []types.GlobalSecondaryIndex{
```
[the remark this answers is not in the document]
hopeyen: yep!

common/aws/dynamodb/client.go
hopeyen: added some unit tests

PR #1394, opened 2025-03-14
Add metrics to feed cache

jianoaix:
Provide visibility to the feed cache.

**TESTED**

**cache hit rate**
Screenshot 2025-03-14 at 9 03 04 PM

**cached time range**
Screenshot 2025-03-14 at 9 03 21 PM

**cache start/end time**
Screenshot 2025-03-14 at 9 04 06 PM

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/dataapi/feed_cache_metrics.go
+++ b/disperser/dataapi/feed_cache_metrics.go
@@ -1,0 +1,96 @@
+// This file should have been placed under disperser/dataapi/v2.
+// The reason it's placed here in "dataapi" package is to avoid circular dependency
+// (the "v2" already has dependency on "dataapi").
+// Note the reason there is a "v2" package in the first place is to enable the separation of
+// swagger docs for v1 and v2 APIs.
+
+package dataapi
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+const namespace = "eigenda_dataapi"
+
+type FeedCacheMetrics struct {
+	// Time range metrics
+	cacheTimeRangeSeconds      prometheus.Gauge
+	cacheSegmentStartTimestamp prometheus.Gauge
+	cacheSegmentEndTimestamp   prometheus.Gauge
+
+	// Cache hit metrics
+	cacheHitRatePercent prometheus.Gauge
+}
+
+func NewFeedCacheMetrics(name string, registry *prometheus.Registry) *FeedCacheMetrics {
+	cacheTimeRangeSeconds := promauto.With(registry).NewGauge(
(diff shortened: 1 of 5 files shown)


disperser/dataapi/feed_cache_metrics.go
```diff
@@ -0,0 +1,96 @@
...
+	cacheSegmentEndTimestamp   prometheus.Gauge
+
+	// Cache hit metrics
+	cacheHitRatePercent prometheus.Gauge
+}
+
+func NewFeedCacheMetrics(name string, registry *prometheus.Registry) *FeedCacheMetrics {
+	cacheTimeRangeSeconds := promauto.With(registry).NewGauge(
+		prometheus.GaugeOpts{
+			Namespace: namespace,
```
dmanc, line 32: I learned recently that you can use "Subsystem" here. So the namespace is `eigenda` and the subsystem is `dataapi`.

disperser/dataapi/feed_cache_metrics.go
```diff
@@ -0,0 +1,96 @@
...
+	return &FeedCacheMetrics{
+		cacheTimeRangeSeconds:      cacheTimeRangeSeconds,
+		cacheSegmentStartTimestamp: cacheSegmentStartTimestamp,
+		cacheSegmentEndTimestamp:   cacheSegmentEndTimestamp,
+		cacheHitRatePercent:        cacheHitRatePercent,
+	}
+}
+
+// UpdateHitRate updates the hit rate metric based on accumulated hits and misses.
+func (m *FeedCacheMetrics) UpdateHitRate(hits, misses int) {
```
dmanc, line 71: Should we track hits and misses as separate metrics and compute the percentage using prometheus query?

disperser/dataapi/metrics.go
```diff
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/Layr-Labs/eigenda/disperser"
+
```
dmanc, line 10: nit: rm newline

dmanc (approved):

disperser/dataapi/metrics.go
jianoaix: removed

disperser/dataapi/feed_cache_metrics.go
jianoaix: It may not that useful to know these two numbers separately. They are about the same request, so recording a hit percentage is more accurate than computing from prometheus query

disperser/dataapi/feed_cache_metrics.go
jianoaix: Ok, it looks more conventional to use subsystem, and may make it easier to filter metrics on dashboard.

PR #1395, opened 2025-03-15
Adds support for simulating operator ejection on v1/v2 metrics

pschork:
This change breaks out ejector's ejection scoring logic into a separate function so that we can "dry run" ejections

Updates the ejections tool to evaluate v1, v2 and v1+v2 combined ejections

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


tools/ejections/cmd/main.go
```diff
@@ -229,3 +263,308 @@ func RunScan(ctx *cli.Context) error {
...
+				metric.PerfScore,
+				metric.IsViolatingSLA,
+				metric.IsViolatingThreshold,
+				metric.NeedsEjection,
+			})
+		}
+		fmt.Println(nonSigningTableV2.Render())
+
+		// Merge non-signing metrics from V1 and V2
+		nonSigningMetricsMerged := e.MergeNonSigningMetrics(nonsigningRateV1, nonsigningRateV2, quorumIDs, logger)
```
supriya-premkumar, line 424: Do we need a check to call `MergeNonSigningMetrics` only in hybrid mode?
Because as it is returning metrics with worse signing rate in `v2-only` mode v1 signing rate will always be 0.

pschork: There is no `v2-only` ejection mode. V1 signing rate is always checked. If v2-ejections are enabled, they will be included and `MergeNonSigningMetrics` will pick the worst signing rate.

PR #1396, opened 2025-03-16
Verifiable Contract Deployment Script

pakim249CAL:

There is a desire to have a deployment script in a secure, verifiable manner. The script in this PR is my first attempt at making a verifiable deployment of the new EigenDA contracts.

## Proposed Deployment Procedure

* Fill in the config file and commit it. This commit should be used later for verification and multisig calldata generation
* Any deployer can run the deployment script with `forge script DeployVerifiable` with the DEPLOY_CONFIG_PATH env variable pointing to the config file. The deployer should make sure that the contracts are verified on etherscan. What the script does is detailed separately below.
* A transaction will be proposed to the owning multisig that calls initializeDeployment(params) on the DeploymentInitializer.
* This tx and all contracts that this tx interacts with will be verified. Separate section on this below.
* After the team is satisfied from the verification process, execute the tx and the contracts are deployed!

## What the Script Does

* Deploy a proxy admin
* Reads a config file to get all initialization parameters and constructor arguments necessary to deploy the contracts.
* Deploy proxies for each deployed contract with mock/inert implementations.
* Deploy implementations that will be upgraded to for each deployed proxy
* Creates a DeploymentInitializer contract which is programmed to atomically upgrade all proxies to their respective implementation contracts and initialize them in a single function call, only ONCE, and only by the intended OWNER when the initializeDeployment function is called. 
* Transfer proxy admin ownership to the DeploymentInitializer. This contract will transfer proxy admin ownership to the OWNER when initializeDeployment is executed.

## Verification

What we want to verify in a deployment is chiefly the following:

* Does the bytecode of all deployed contracts match what a specific github commit, with a specific compilation target?
* Were the contracts constructed with the correct arguments, and in the correct order if there is a stateful dependency?
* Were the contracts initialized with the correct arguments in the correct order?
* Is all the initial state what we expect?

Matching bytecode to a commit is straightforward, and tools such as etherscan can do a lot of the work here. If the deployer verifies the contracts on etherscan as they should, then we can easily know for any specific contract what code and compilation target generated the deployed bytecode.

Constructor arguments can be examined in the calldata of the deployment transactions, through the deployed bytecode, and through tools like tenderly.

Initializations in this procedure are all done through the DeploymentInitializer contract. All initialization parameters that are not dynamically sized are immutable public variables that can be queried on chain. All other parameters are submitted in calldata, which will need to be verified off-chain. 

Initial state can be tested with a foundry test. TODO

## Some Additional Notes

* There is a stateful dependency in the constructor of the BLSSignatureChecker contract which is inherited by the EigenDAServiceManager as it queries the RegistryCoordinator and StakeManager contracts. This is problematic if we are trying to deploy inert implementation contracts. I made an issue [here](https://github.com/Layr-Labs/eigenlayer-middleware/issues/434) about it, but for now this PR works around it by deploying mock implementations for these contracts that return the values that would be returned once the proxies are initialized.
* We can possibly make the deployment procedure simpler (and therefore safer) by having ALL initialization arguments on-chain rather than submitting them as calldata. There is a tradeoff to be made here though, because dynamically sized initialization parameters cannot be stored as immutable, therefore necessitating the use of storage. Depending on the intended size of these parameters, this can become very expensive.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


pakim249CAL:
@ethenotethan @ian-shim @samlaf Just pinging for a review around the ideas. I have some cleanup to do, but I'd appreciate if you could read through what I wrote up above here and let me know if anything seems off or confusing about it.

contracts/script/deploy/sepolia/DeploymentInitializer.sol
```diff
@@ -0,0 +1,202 @@
...
+import {IEigenDARelayRegistry, EigenDARelayRegistry} from "src/core/EigenDARelayRegistry.sol";
+import {PaymentVault} from "src/payments/PaymentVault.sol";
+import {IEigenDADisperserRegistry, EigenDADisperserRegistry} from "src/core/EigenDADisperserRegistry.sol";
+import {EigenDAServiceManager} from "src/core/EigenDAServiceManager.sol";
+import {IRewardsCoordinator} from
+    "lib/eigenlayer-middleware/lib/eigenlayer-contracts/src/contracts/interfaces/IRewardsCoordinator.sol";
+import {ProxyAdmin, TransparentUpgradeableProxy} from "@openzeppelin/contracts/proxy/transparent/ProxyAdmin.sol";
+import "./DeploymentTypes.sol";
+
+/**
```
ethenotethan, line 18: just for understanding - the intent of this contract is for providing an on-chain initialization step where dependency injection happens as a precursor tx which allows a deployer to then verify before actually invoking the deploy? 

curious about the following:
1. is this only used for bootstrapping/ genesis deployment of the contracts or will it also be leveraged for future upgrades? If for bootstrapping could we not achieve the same effect through a forge deployment script which verifies non-eigenda contract dependencies through as set of invariants before beginning the initialization steps for each eigenda contract?

contracts/script/deploy/sepolia/Deploy.s.sol
```diff
@@ -0,0 +1,255 @@
...
+    CalldataRegistryCoordinatorParams,
+    CalldataThresholdRegistryParams,
+    CalldataServiceManagerParams
+} from "./DeploymentInitializer.sol";
+
+import "forge-std/Script.sol";
+import "forge-std/StdJson.sol";
+
+contract DeploySepolia is Script {
+    string deployConfigPath = string(bytes("./script/deploy/sepolia/config/placeholder.config.json"));
```
samlaf, line 50: Do you not like forge's best practice of having a single deploy file that works with config files for different networks? See #7 in https://book.getfoundry.sh/guides/best-practices#scripts

Do you think its safer to have a Deploy Script per network with hardcoded config path like this?

contracts/script/deploy/sepolia/Deploy.s.sol
```diff
@@ -0,0 +1,255 @@
...
+    address initialOwner;
+
+    ProxyAdmin proxyAdmin;
+
```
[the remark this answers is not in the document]
samlaf: +1 @pakim249CAL add as much of your smart contract wisdom as possible in comments. This will help us a lot when reviewing these.

contracts/script/deploy/sepolia/Deploy.s.sol
samlaf, line 1: Is this script meant to replace https://github.com/Layr-Labs/eigenda/blob/master/contracts/script/EigenDADeployer.s.sol?

contracts/script/deploy/sepolia/config/placeholder.config.json
samlaf, line 1: Could we add a bit more structure to this file? Perhaps put all the eigenlayer contracts under an "eigenalyer_contracts" field, the multisigs/owners under some field, the eigenda contracts under an "eigenda_contracts" field etc.

contracts/script/deploy/sepolia/DeploymentInitializer.sol
```diff
@@ -0,0 +1,202 @@
...
+ * @title DeploymentInitializer
+ * @author Layr Labs, Inc.
+ * @notice This contract is intended to be used by a multisig to verify a deployment.
+ * It accomplishes this by storing all the parameters needed to initialize the contracts that are not dynamically sized.
```
samlaf, line 22:
```suggestion
 * It accomplishes this by storing all the statically-sized parameters needed to initialize the contracts.
```

samlaf (commented):
First pass; not going to claim that I understand what's happening. Will need to re-read in the morning.

contracts/script/deploy/sepolia/Deploy.s.sol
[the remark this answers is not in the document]
pakim249CAL: Yes, and this connects with a below comment that we can maybe make this whole thing chain agnostic. I'll be working on that.

contracts/script/deploy/sepolia/DeploymentInitializer.sol
pakim249CAL: Your understanding is basically correct. There's no reason this framework cannot be used for upgrades too.

contracts/script/deploy/sepolia/Deploy.s.sol
pakim249CAL: As @ethenotethan mentions, it should be possible to make this all chain agnostic. I'll work on that.

contracts/script/deploy/sepolia/Deploy.s.sol
pakim249CAL: As a source material, I was using Holesky_Deploy.s.sol, so I think there is some duplicated effort in the repo currently that may need to be cleaned up.

.github/workflows/test-contracts.yml
```diff
@@ -19,7 +19,7 @@ concurrency:
 jobs:
   forge-tests:
     name: Foundry Project
-    runs-on: ubuntu-22.04
+    runs-on: ubuntu-24.04
```
pakim249CAL, line 22: This change is needed because of a GLIBC dependency issue

contracts/script/deploy/sepolia/config/placeholder.config.json
pakim249CAL: Updated

contracts/script/deploy/sepolia/Deploy.s.sol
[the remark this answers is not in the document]
pakim249CAL: Made this chain agnostic, but I think I will reserve the actual config for another PR because I'm actually not quite sure yet what I should put into the configs. I think we might need to wait for some deployments to happen on Sepolia.

.github/workflows/test-contracts.yml
samlaf: PR with this change has been merged to master already

contracts/script/deploy/sepolia/config/placeholder.config.json
samlaf: I'm a big stickler for this but could you try to make each comment addressal in a separate commit and then tag the commit link as a response. For example: [a0e8f1e](https://github.com/Layr-Labs/eigenda/pull/1396/commits/a0e8f1ee87b646fd9d464d3315e19ea1a6bee58c) (just a random commit as example). I have to review a lot of different PRs in diffeent contexts throughout the day and this makes my job a lot easier.

contracts/script/deploy/sepolia/Deploy.s.sol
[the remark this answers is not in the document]
pakim249CAL: Included in this commit that also reorganizes the variables for better clarity https://github.com/Layr-Labs/eigenda/pull/1396/commits/c41455dda6afbd2a711ab8c7f9d580d9811a8724

contracts/script/deploy/sepolia/config/placeholder.config.json
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1396/commits/e74d1470a7f9519dc030b825668f91cfa68ddb46

contracts/script/deploy/sepolia/DeploymentInitializer.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1396/commits/ae611d80e48ef04efe3011bc607b541d58e7f457

contracts/script/deploy/sepolia/Deploy.s.sol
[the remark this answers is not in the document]
samlaf: Nice thanks! Curious will you add a runbook README somewhere to explain the process you described in this PR description. Feel like it would be useful. For example when I look at "initialOwner" I assume that its the initialOwner at the start of this script, and that it passes ownership to another address by the end of the script. So might be good to have that detailed that the change of ownership is done by the multisig yadi yada (I already forgot how exactly it works :\)

contracts/script/deploy/sepolia/config/placeholder.config.json
samlaf: ty!

samlaf (approved):
LGTM in this state. What are the future steps? Are we ready to test the deployment? Excited to run the whole procedure and learn from you how to verify the deployment using tenderly.

Some potential future TODOS (maybe worth creating issues?):
1. you have a TODO left in the PR description to write test script. Is that important?
2. we talked about this offline but I would personally love to replace all json everywhere with toml (which [forge can also parse](https://book.getfoundry.sh/cheatcodes/parse-toml)) because then we can add comments. My POV is that config files should be understandable without needing to go read the code
3. Would be nice to add a runbook (maybe just part of the contracts/README?) to describe the procedure you detailed in the PR description. Perhaps after we tried it and are happy with the whole procedure?

Nit: you mention in PR description that one just needs to `forge script DeployVerifiable`, but I think you added an env var which also needs to be set now to point to the config file right?

pakim249CAL:
> LGTM in this state. What are the future steps? Are we ready to test the deployment? Excited to run the whole procedure and learn from you how to verify the deployment using tenderly.

I will definitely spend some time this week writing a more detailed and organized rundown on the entire procedure. AKA a runbook like you suggested, which will be focused on the specific actions and checks each role needs to do to verify everything.

> you have a TODO left in the PR description to write test script. Is that important?

The TODO in the test script should be populated with a call to a test that check all the initial state. I have some requirements that I would like this test to satisfy, so I am not sure if I should include it in this PR specifically. But the gist of it is that I would like to code ONE test that can both be run in the script, as well as separately outside of the script to test all the initial state. I think I will make a separate PR for this. Issue in linear.

> we talked about this offline but I would personally love to replace all json everywhere with toml (which [forge can also parse](https://book.getfoundry.sh/cheatcodes/parse-toml)) because then we can add comments. My POV is that config files should be understandable without needing to go read the code

Agree. Will try this out in another PR. Issue in linear.

> Nit: you mention in PR description that one just needs to forge script DeployVerifiable, but I think you added an env var which also needs to be set now to point to the config file right?

Will fix.

PR #1401, opened 2025-03-18
add trailing option to read g2 file

bxue-l2:
This PR adds a new flag to the encoding prover cli interface. The flag allows user to specify a file containing only a trailing portion of g2 points.

This PR is driven by the requirement of generating blob length proof on the proxy. It avoids downloading the 2^28 G2 points (having 16GiB) bytes

The encoder service hosted by eigenlabs can safely ignore the new flag, and continue to run with the current configuration.

### DIFF
--- a/encoding/kzg/cli.go
+++ b/encoding/kzg/cli.go
@@ -37,7 +37,7 @@ func CLIFlags(envPrefix string) []cli.Flag {
 		},
 		cli.StringFlag{
 			Name:     G2TrailingPathFlagName,
-			Usage:    "Path to trailing G2 SRS file. Its intended purpose is to allow local generation the blob length proof. If you already downloaded the entire G2 SRS file which contains 268435456 G2 points with total size 16GiB. With this G2TrailingPathFlag, user can use a smaller file that contains only the trailing end of the whole G2 SRS file. Ignoring this flag, the program assumes the entire G2 SRS file is provided. With this flag, the size of the provided file must be at least SRSLoadingNumberFlagName * 64 Bytes.",
+			Usage:    "Path to trailing G2 SRS file. Its intended purpose is to allow local generation the blob length proof. If you already downloaded the entire G2 SRS file which contains 268435456 G2 points with total size 16GiB, this flag is not needed. With this G2TrailingPathFlag, user can use a smaller file that contains only the trailing end of the whole G2 SRS file. Ignoring this flag, the program assumes the entire G2 SRS file is provided. With this flag, the size of the provided file must be at least SRSLoadingNumberFlagName * 64 Bytes.",
 			Required: false,
 			EnvVar:   common.PrefixEnvVar(envPrefix, "G2_TRAILING_PATH"),
 		},
--- a/encoding/kzg/cli.go
+++ b/encoding/kzg/cli.go
@@ -10,6 +10,7 @@ import (
 const (
 	G1PathFlagName            = "kzg.g1-path"
 	G2PathFlagName            = "kzg.g2-path"
+	G2TrailingPathFlagName    = "kzg.g2-trailing-path"
 	CachePathFlagName         = "kzg.cache-path"
 	SRSOrderFlagName          = "kzg.srs-order"
 	NumWorkerFlagName         = "kzg.num-workers"
@@ -33,6 +34,12 @@ func CLIFlags(envPrefix string) []cli.Flag {
 			Usage:    "Path to G2 SRS. Either this flag or G2_POWER_OF_2_PATH needs to be specified. For operator node, if both are specified, the node uses G2_POWER_OF_2_PATH first, if failed then tries to G2_PATH",
 			Required: false,
 			EnvVar:   common.PrefixEnvVar(envPrefix, "G2_PATH"),
+		},
+		cli.StringFlag{
+			Name:     G2TrailingPathFlagName,
+			Usage:    "Path to trailing G2 SRS file. User can specify a file that contains the trailing end of the G2 SRS points. Using this flag has the advantage to download less G2 SRS file in order to compute the length proof. If you already downloaded the entire G2 SRS file, feel free to ignore this option.",
+			Required: false,
+			EnvVar:   common.PrefixEnvVar(envPrefix, "G2_TRAILING_PATH"),
 		},
 		cli.StringFlag{
 			Name:     CachePathFlagName,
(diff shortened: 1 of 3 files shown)


encoding/kzg/cli.go
```diff
@@ -34,6 +35,12 @@ func CLIFlags(envPrefix string) []cli.Flag {
 			Required: false,
 			EnvVar:   common.PrefixEnvVar(envPrefix, "G2_PATH"),
 		},
+		cli.StringFlag{
+			Name:     G2TrailingPathFlagName,
+			Usage:    "Path to trailing G2 SRS file. User can specify a file that contains the trailing end of the G2 SRS points. Using this flag has the advantage to download less G2 SRS file in order to compute the length proof. If you already downloaded the entire G2 SRS file, feel free to ignore this option.",
```
samlaf, line 40: can you expand on what the assumption for "trailing" are, aka we assume the SRSOrder is hardcoded to 16GiB.

samlaf: Also can you be clearer with what happens if this option is not set (instead of just saying "feel free to ignore"). What happens if it is not set (default): the assumption is that G2Path contains the entire 16GiB file and the trailing part will be read from it

encoding/kzg/kzgrs.go
```diff
@@ -3,6 +3,7 @@ package kzg
 type KzgConfig struct {
 	G1Path          string
 	G2Path          string
+	G2TrailingPath  string
```
samlaf, line 6: explain the relationship between these 2 fields. What happens when G2TrailingPath is not set. Explain their interdependency, etc.

encoding/kzg/prover/prover.go
```diff
@@ -66,12 +66,39 @@
...
-		)
+		hasG2TrailingFile := len(kzgConfig.G2TrailingPath) != 0
+		if hasG2TrailingFile {
+			fileStat, err := os.Stat(kzgConfig.G2TrailingPath)
+			if err != nil {
+				return nil, err
+			}
+			fileSizeByte := fileStat.Size()
+			if fileSizeByte%64 != 0 {
+				return nil, errors.New("corrupted g2 points from G2TrailingPath.")
```
litt3, line 77: more info would be helpful: what was the actual size? why is that size not valid (i.e. it's not a multiple of 64)

encoding/kzg/prover/prover.go
```diff
@@ -66,12 +66,39 @@ func NewProver(kzgConfig *kzg.KzgConfig, encoderConfig *encoding.Config) (*Prove
...
+			}
+			// get the size
+			numG2point := uint64(fileSizeByte / kzg.G2PointBytes)
+			if numG2point < kzgConfig.SRSNumberToLoad {
+				return nil, errors.New("Insufficent number of g2 points from G2TrailingPath")
```
litt3, line 82: more info would be helpful here, too. how many g2 points exist, vs how many are being attempted to load?

litt3 (commented):
Your PR fixes the G2 reading for the prover, but the verifier suffers from the same problem

https://github.com/Layr-Labs/eigenda/blob/217365e143148896addd8ce54440bfebf10d6a76/encoding/kzg/verifier/verifier.go#L62

litt3:
> > Your PR fixes the G2 reading for the prover, but the verifier suffers from the same problem
> > https://github.com/Layr-Labs/eigenda/blob/217365e143148896addd8ce54440bfebf10d6a76/encoding/kzg/verifier/verifier.go#L62
> 
> Does proxy run verifiers? Usually da nodes run verifier.
> 
> The naming is pretty bad there, da node will default LoadG2Points=false. So that path is triggered.

Yes, the proxy uses a verifier for V1

encoding/kzg/cli.go
bxue-l2: https://github.com/Layr-Labs/eigenda/pull/1401/commits/08c7c57430ac685af7d88426880ccb3c0167031a

encoding/kzg/kzgrs.go
bxue-l2: https://github.com/Layr-Labs/eigenda/pull/1401/commits/08c7c57430ac685af7d88426880ccb3c0167031a

encoding/kzg/prover/prover.go
bxue-l2: https://github.com/Layr-Labs/eigenda/pull/1401/commits/e435f13aeb156a9417f2aee9a541294414b9fde4

encoding/kzg/cli.go
```diff
@@ -34,6 +35,12 @@ func CLIFlags(envPrefix string) []cli.Flag {
 			Required: false,
 			EnvVar:   common.PrefixEnvVar(envPrefix, "G2_PATH"),
 		},
+		cli.StringFlag{
+			Name:     G2TrailingPathFlagName,
+			Usage:    "Path to trailing G2 SRS file. Its intended purpose is to allow local generation the blob length proof. If you already downloaded the entire G2 SRS file which contains 268435456 G2 points with total size 16GiB. With this G2TrailingPathFlag, user can use a smaller file that contains only the trailing end of the whole G2 SRS file. Ignoring this flag, the program assumes the entire G2 SRS file is provided. With this flag, the size of the provided file must be at least SRSLoadingNumberFlagName * 64 Bytes.",
```
dmanc, line 40: nit: `If you already downloaded the entire G2 SRS file which contains 268435456 G2 points with total size 16GiB.` incomplete sentence?

bxue-l2: https://github.com/Layr-Labs/eigenda/pull/1401/commits/b9e160838ba805ec14fbc1bbcd346364cb9e6e0a

dmanc (approved):

encoding/kzg/prover/prover.go
```diff
@@ -66,12 +66,39 @@ func NewProver(kzgConfig *kzg.KzgConfig, encoderConfig *encoding.Config) (*Prove
...
+				return nil, err
+			}
+			fileSizeByte := fileStat.Size()
+			if fileSizeByte%64 != 0 {
+				return nil, fmt.Errorf("corrupted g2 point from the G2TrailingPath. The size of the file on the provided path has size that is not multiple of 64, which is %v. It indicates there is an incomplete g2 point.", fileSizeByte)
+			}
+			// get the size
+			numG2point := uint64(fileSizeByte / kzg.G2PointBytes)
+			if numG2point < kzgConfig.SRSNumberToLoad {
+				return nil, fmt.Errorf("insufficent number of g2 points from G2TrailingPath. Requested %v, Acutal %v", kzgConfig.SRSNumberToLoad, numG2point)
```
litt3, line 82: Acutal -> Actual

bxue-l2: https://github.com/Layr-Labs/eigenda/pull/1401/commits/8946b6fb0088d5dfd9f59cdaa797dc0e55f8268b

litt3 (approved):

encoding/kzg/kzgrs.go
samlaf: think it should be explained in the config as well personally. Because some people might want to import this and use it directly without the flags.

samlaf (approved):
LGTM although I still think config fields should be commented as well, not only flags

encoding/kzg/kzgrs.go
bxue-l2: I don't think it is too bad an assumption that the person will trace back to the cli, since that is the only place it was set.

PR #1405, opened 2025-03-18
Don't block forever in EncodeBlob() if the context is cancelled.

cody-littley:

Fixes an issue pointed out by auditors. Don't block forever if the context is cancelled.

### DIFF
--- a/disperser/encoder/server_v2.go
+++ b/disperser/encoder/server_v2.go
@@ -127,13 +127,23 @@ func (s *EncoderServerV2) EncodeBlob(ctx context.Context, req *pb.EncodeBlobRequ
 		s.queueLock.Unlock()
 	default:
 		s.metrics.IncrementRateLimitedBlobRequestNum(int(blobSize))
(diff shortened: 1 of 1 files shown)


disperser/encoder/server_v2.go
```diff
@@ -127,13 +127,23 @@ func (s *EncoderServerV2) EncodeBlob(ctx context.Context, req *pb.EncodeBlobRequ
...
-	defer s.popRequest()
+
+	select {
+	case s.runningRequests <- struct{}{}:
+		defer s.popRequest()
+	case <-ctx.Done():
+		s.metrics.IncrementCanceledBlobRequestNum(int(blobSize))
+		return nil, status.Error(codes.Canceled, "request was canceled")
+	}
+
```
dmanc, line 146: Do we still need the below ctx.Err() check?

dmanc (approved):

litt3 (approved):

PR #1406, opened 2025-03-18
Fix eda2-22

cody-littley:

Don't change payment state if the request fails.

### DIFF
--- a/api/clients/v2/accountant.go
+++ b/api/clients/v2/accountant.go
@@ -97,6 +97,7 @@ func (a *Accountant) BlobPaymentInfo(
 		if err := QuorumCheck(quorumNumbers, a.reservation.QuorumNumbers); err != nil {
 			return big.NewInt(0), err
 		}
+		overflowPeriodRecord.Usage += relativePeriodRecord.Usage - binLimit
 		return big.NewInt(0), nil
 	}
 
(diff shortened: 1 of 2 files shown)


api/clients/v2/accountant.go
```diff
@@ -89,30 +94,39 @@ func (a *Accountant) BlobPaymentInfo(ctx context.Context, numSymbols uint64, quo
...
+	resultingPayment := big.NewInt(0)
+	resultingPayment.Add(a.cumulativePayment, incrementRequired)
+	if resultingPayment.Cmp(a.onDemand.CumulativePayment) <= 0 {
 		if err := QuorumCheck(quorumNumbers, requiredQuorums); err != nil {
 			return big.NewInt(0), err
 		}
+		a.cumulativePayment.Add(a.cumulativePayment, incrementRequired)
 		return a.cumulativePayment, nil
 	}
+	a.cumulativePayment.Add(a.cumulativePayment, incrementRequired)
```
hopeyen, line 118: Ohh wait, second thought. Could we remove increment in this case? I changed the behavior on the server side that if the payment is exceeding on-chain deposit, the server will not meter the request

cody-littley: removed

PR #1411, opened 2025-03-19
[Breaking Changes] Add signature replay protection for AuthenticatePaymentStateRequest

yujiezhu0:
[Breaking Changes] Audit fix implementation to add signature replay protection for AuthenticatePaymentStateRequest 

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Integration tests


api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -120,6 +120,11 @@ message GetPaymentStateRequest {
   string account_id = 1;
   // Signature over the account ID
   bytes signature = 2;
+  // Timestamp of the request in seconds since the Unix epoch. If too far out of sync with the server's clock,
+  // request may be rejected.
+  uint32 timestamp = 3;
+  // A client-generated nonce to ensure the request is unique.
+  bytes nonce = 4;
```
ian-shim, line 127: why do we need a timestamp _and_ a nonce? could we just use timestamp and expire it after some threshold?

yujiezhu0: Timestamp is recorded in seconds. If multiple requests are sent in the same second, we need an additional layer (nonce) to ensure each request is unique.

disperser/cmd/apiserver/flags/flags.go
```diff
@@ -161,6 +161,20 @@ var (
...
+		Value:    30 * time.Second,
+		EnvVar:   common.PrefixEnvVar(envVarPrefix, "AUTH_PMT_REQUEST_MAX_PAST_AGE"),
+	}
+	AuthPmtStateRequestMaxFutureAge = cli.DurationFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "auth-pmt-state-request-max-future-age"),
+		Usage:    "The maximum age of an AuthPaymentState request in the future that the disperser accepts",
+		Required: false,
+		Value:    30 * time.Second,
+		EnvVar:   common.PrefixEnvVar(envVarPrefix, "AUTH_PMT_REQUEST_MAX_FUTURE_AGE"),
+	}
```
yujiezhu0, line 177: We probably want to discuss what a reasonable bound would be. Currently, this setup allows a 1-min window.

core/auth/v2/authenticator.go
```diff
@@ -48,12 +56,15 @@ func (*authenticator) AuthenticateBlobRequest(header *core.BlobHeader, signature
...
+	sig := request.GetSignature()
 	// Ensure the signature is 65 bytes (Recovery ID is the last byte)
 	if len(sig) != 65 {
 		return fmt.Errorf("signature length is unexpected: %d", len(sig))
 	}
 
+	timestamp := request.GetTimestamp()
+
 	// Verify the signature
 	hash := sha256.Sum256(accountAddr.Bytes())
```
ian-shim, line 69: the signature should be over the hash of the request instead of just the account addr
The [corresponding signing logic](https://github.com/Layr-Labs/eigenda/blob/master/core/auth/v2/signer.go#L53) also needs to be updated so that it signs the request hash when it creates the request

core/auth/v2/authenticator.go
```diff
@@ -4,16 +4,24 @@ import (
...
 )
 
-type authenticator struct{}
+type authenticator struct {
+	replayGuardian replay.ReplayGuardian
+}
 
-func NewAuthenticator() *authenticator {
-	return &authenticator{}
+func NewAuthenticator(maxTimeInPast, maxTimeInFuture time.Duration) *authenticator {
```
ian-shim, line 21: The inputs to the constructor and replayGuardian are only necessary for `AuthenticatePaymentStateRequest`. These inputs shouldn't be needed for `AuthenticateBlobRequest`.

api/hashing/payment_state_hashing.go
```diff
@@ -0,0 +1,32 @@
...
+	}
+
+	// Hash the timestamp
+	hashUint64(hasher, timestamp)
+
+	return hasher.Sum(nil), nil
+}
+
+// HashGetPaymentStateRequestFromRequest hashes the given GetPaymentStateRequest from request
+func HashGetPaymentStateRequestFromRequest(request *pb.GetPaymentStateRequest) ([]byte, error) {
```
ian-shim, line 27: Do we need both of these methods? They seem to be doing to same thing. Maybe we could keep just one?

core/auth/v2/authenticator.go
```diff
@@ -46,16 +64,23 @@ func (*authenticator) AuthenticateBlobRequest(header *core.BlobHeader, signature
...
 	// Ensure the signature is 65 bytes (Recovery ID is the last byte)
 	if len(sig) != 65 {
 		return fmt.Errorf("signature length is unexpected: %d", len(sig))
 	}
 
+	requestHash, err := hashing.HashGetPaymentStateRequestFromRequest(request)
+	if err != nil {
+		return fmt.Errorf("failed to hash request: %w", err)
+	}
+	accountAddrWithHash := append(accountAddr.Bytes(), requestHash...)
```
ian-shim, line 80: same here as https://github.com/Layr-Labs/eigenda/pull/1411/files#r2005624478

api/hashing/payment_state_hashing.go
yujiezhu0: They take different params. `HashGetPaymentStateRequestFromFields()` is just for the signer since the signer won't be able to pass in a request; in other cases, passing in the request is a lot simpler

ian-shim: > in other cases, passing in the request is a lot simpler

They could also pass in individual fields as well, right? Sharing this logic in one method will help prevent any logic changes from being made in one method vs. both methods

cody-littley (approved):

PR #1412, opened 2025-03-19
chore: updating subgraphs deps

anupsv:

Update the dependencies.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


jianoaix (approved):
Is this an optional upgrade?

anupsv:
> Is this an optional upgrade?

So dependabot was complaining a lot about old deps and vukbs from it. So figured I'd update the deps directly and then see how much is resolved.

ian-shim (approved):

PR #1413, opened 2025-03-19
enable sync writes for levelDB

cody-littley:

Enable sync writes for level DB. Can be reverted to old behavior with settings changes.


node/config.go
```diff
@@ -289,6 +293,10 @@ func NewConfig(ctx *cli.Context) (*Config, error) {
 		ReachabilityPollIntervalSec:         reachabilityPollIntervalSec,
 		EnableTestMode:                      testMode,
 		OverrideBlockStaleMeasure:           ctx.GlobalUint64(flags.OverrideBlockStaleMeasureFlag.Name),
+		LevelDBDisableSeeksCompactionV1:     !ctx.GlobalBool(flags.LevelDBEnableSeeksCompactionV1Flag.Name),
+		LevelDBSyncWritesV1:                 !ctx.GlobalBool(flags.LevelDBDisableSyncWritesV1Flag.Name),
+		LevelDBDisableSeeksCompactionV2:     !ctx.GlobalBool(flags.LevelDBEnableSeeksCompactionV2Flag.Name),
+		LevelDBSyncWritesV2:                 !ctx.GlobalBool(flags.LevelDBDisableSyncWritesV2Flag.Name),
```
cody-littley, line 299: Note to the reviewer: The framework we are using doesn't support boolean configuration options with a default value of `true`, only a default value of `false`. So I had to invert these flags, since by default we want to make these variables `true`.

jianoaix: https://pkg.go.dev/github.com/urfave/cli#BoolTFlag is true by default

cody-littley: neat, I didn't know that. Fixed.

common/kvstore/tablestore/config.go
```diff
@@ -36,18 +36,27 @@ type Config struct {
 	Schema []string
 	// Optional Prometheus registry for metrics collection. If nil, metrics collection is disabled.
 	MetricsRegistry *prometheus.Registry
+
+	// If true, levelDB will disable seeks compaction. Default is true. This setting is ignored if the base store is
```
jianoaix, line 40: When this table store was suggested, I made a point of leaky abstraction. This looks getting started.

jianoaix (approved):

common/kvstore/tablestore/config.go
cody-littley: I agree, this is ugly.

Perhaps a better way to handle this sort of custom configuration for different impls is to make the base DB object something that gets passed into the constructor of the wrapper object. Not something for this PR of course, but perhaps something to consider in the future.

PR #1414, opened 2025-03-19
clean up ReceiveSignatures()

cody-littley:

This PR does not make any functional changes. The goal of this PR is to make the `RecieveSignatures()` function easier to read. It was originally a very long and complex function, and so I broke it apart into several smaller functions that are individually easier to groc.

### DIFF
--- a/core/aggregation.go
+++ b/core/aggregation.go
@@ -42,8 +42,8 @@ type QuorumAttestation struct {
 	QuorumAggPubKey map[QuorumID]*G1Point
 	// SignersAggPubKey is the aggregated public key for all of the operators that signed the message by each quorum
 	SignersAggPubKey map[QuorumID]*G2Point
-	// AggSignature is the aggregated signature for all of the operators that signed the message for each quorum, mirroring the
-	// SignersAggPubKey.
+	// AggSignature is the aggregated signature for all of the operators that signed the message for each quorum,
+	// mirroring the SignersAggPubKey.
 	AggSignature map[QuorumID]*Signature
 	// QuorumResults contains the quorum ID and the amount signed for each quorum
 	QuorumResults map[QuorumID]*QuorumResult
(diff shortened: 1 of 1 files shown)


core/aggregation.go
```diff
@@ -115,171 +130,283 @@ func (a *StdSignatureAggregator) ReceiveSignatures(ctx context.Context, state *I
...
+	operatorQuorums []uint8) []uint8 {
+
+	// Get stake amounts for operator
+	ops := state.Operators[quorumID]
+	opInfo, ok := ops[signingMessage.Operator]
+	// If operator is not in quorum, skip
+	if !ok {
+		return operatorQuorums
+	}
+	operatorQuorums = append(operatorQuorums, quorumID)
```
ian-shim, line 183: The fact that the method returns updated `operatorQuorums` seems to imply that it should not mutate the input `operatorQuorums`. Should we return a new copy instead of mutating the input?

core/aggregation.go
```diff
@@ -115,171 +130,283 @@ func (a *StdSignatureAggregator) ReceiveSignatures(ctx context.Context, state *I
...
+}
 
-		op, found := state.IndexedOperators[r.Operator]
-		if !found {
-			a.Logger.Error("Operator not found in state", "operatorID", operatorIDHex, "operatorAddress", operatorAddr, "socket", socket)
-			continue
-		}
+// collectSignatureInfoForQuorum collects the signature information for the operator in the quorum, updating the
+// attestation and stake signed. Returns the updated operatorQuorums.
+func (a *StdSignatureAggregator) collectSignatureInfoForQuorum(
```
ian-shim, line 166: Not really against factoring out these methods, but also not sure if this actually makes the code more readable given that it's taking a large number of arguments and directly mutating input values

ian-shim (approved):
lgtm!

PR #1415, opened 2025-03-19
feat: Verifiable Deployment Calldata script

pakim249CAL:

Adds a script to get calldata that would be submitted to a multisig, and adds supporting refactors to minimize code duplication and vectors for misconfiguration. Also adds tests for what is testable on-chain, but there is still some verification that will need to be done offchain such as verifying the contracts on etherscan, making sure proxies have a clean history, etc.

The refactors mainly are:

* cfg string is referenced and used instead of putting init vars in script storage
* cfg library is made so that multiple scripts can read the same cfg

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


.github/workflows/test-contracts.yml
```diff
@@ -47,11 +47,11 @@ jobs:
         working-directory: ./contracts
 
       - name: Run tests
-        run: forge test -vvv
+        run: forge test --no-match-path script/**/*.sol -vvv
```
ethenotethan, line 50: should we also capture this exclusion regex in the `package.json`'s test command definition? also what's the rationale for not running the script tests as part of CI?

contracts/script/deploy/verifiable/DeployVerifiable.s.sol
```diff
@@ -116,24 +95,49 @@ contract DeployVerifiable is Script {
...
+    function _initConfig() internal virtual {
+        cfg = vm.readFile(vm.envString("DEPLOY_CONFIG_PATH"));
     }
 
-    function _doTests(string memory cfg) internal {
-        vm.startPrank(initialOwner);
-        CalldataInitParams memory params = CalldataInitParamsLib.getCalldataInitParams(cfg);
+    function _logs() internal virtual {
+        console2.log("Deployment addresses:");
+        console2.log("Deployment Initializer: ", address(deploymentInitializer));
```
ethenotethan, line 112: nit - inconsistent formatting with other log stmts

contracts/script/deploy/verifiable/DeployVerifiable.s.sol
```diff
@@ -116,24 +95,49 @@ contract DeployVerifiable is Script {
...
+        console2.log("Empty Contract Implementation", emptyContract);
+        console2.log("Mock Stake Registry Implementation", mockStakeRegistry);
+        console2.log("Mock Registry Coordinator Implementation", mockRegistryCoordinator);
+
+        console2.log(
+            "\n\nAll other relevant deployment addresses should be queried from the DeploymentInitializer contract."
+        );
+    }
+
+    function _doTests() internal {
```
ethenotethan, line 122: is this not a redundant simulation with just running the script with `--broadcast` enabled?

.github/workflows/test-contracts.yml
pakim249CAL: The test I wrote to test the deployment is meant to be ran after a deployment by a verifier (which may not be the deployer!), ingesting environment variables input by the verifier. This would obviously fail in CI without contracts already deployed. And yes, I guess it would need to be excluded in package.json too.

Maybe there is a better way to do this, but I think in general there should be no need to put tests that are in the script/ folder into CI. If we put tests there meant for CI, they should be in the test/ folder.

contracts/script/deploy/verifiable/DeploymentTypes.sol
```diff
@@ -74,3 +75,141 @@ struct ImmutablePaymentVaultParams {
 struct ImmutableServiceManagerParams {
     address rewardsInitiator;
 }
+
+library InitParamsLib {
```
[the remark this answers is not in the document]
pakim249CAL: More accurately, it would be a json config reader. But yes, point taken. Maybe something like DeploymentJSONConfigLib?

.github/workflows/test-contracts.yml
pakim249CAL: I guess by default, I could have the test run after running the deployment script using mock values. But then the test would not fail if the correct environment variables are not supplied, which seems like a way that the verifier might make a mistake.

ethenotethan: +1 to having it as a part of a script and not a test

contracts/script/deploy/verifiable/DeployVerifiable.s.sol
ethenotethan: ah I see that makes sense - any chance we could capture this role differentiation in a smol comment? At first glance the behavior is a bit hard to decipher

contracts/script/deploy/verifiable/DeployVerifiable.s.sol
```diff
@@ -116,24 +95,49 @@ contract DeployVerifiable is Script {
...
+    /// @dev override this if you don't want to use the environment to get the config path
+    function _initConfig() internal virtual {
+        cfg = vm.readFile(vm.envString("DEPLOY_CONFIG_PATH"));
     }
 
-    function _doTests(string memory cfg) internal {
-        vm.startPrank(initialOwner);
-        CalldataInitParams memory params = CalldataInitParamsLib.getCalldataInitParams(cfg);
+    function _logs() internal virtual {
+        console2.log("Deployment addresses:");
```
ethenotethan, line 111: just to learn - how does `console2` differ from just regular `console`?

pakim249CAL: From the foundry [doc](https://book.getfoundry.sh/reference/forge-std/),

"console2.sol contains patches to console.sol that allow Forge to decode traces for calls to the console, but it is not compatible with Hardhat."

contracts/script/deploy/verifiable/DeploymentTypes.sol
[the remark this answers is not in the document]
pakim249CAL: But then you would have to track a contract address that store configData instead of just the configData string itself, right? That seems more complex to me...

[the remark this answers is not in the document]
ethenotethan: haha yeah its a nit for a reason - feel free to resolve

.github/workflows/test-contracts.yml
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1415/commits/e1e14498b97626ef09b9b39e2d222a10713a8ffa

contracts/script/deploy/verifiable/DeployVerifiable.s.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1415/commits/acf9096d69f2e4f805e50c023de91269b4cde81b

contracts/script/deploy/verifiable/DeployVerifiable.s.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1415/commits/acf9096d69f2e4f805e50c023de91269b4cde81b

ethenotethan (approved):
LGTM!

PR #1416, opened 2025-03-19
chore: update eigenlayer-middleware commit

hopeyen:

update to the commit with operator state retriever with socket

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


ian-shim:
@pakim249CAL do we need to redeploy OperatorStateRetriever?

ian-shim (approved):

pakim249CAL:
There should be no need to redeploy it, and it can very quickly be redeployed if there is any problem with it because the contract actually has no constructor arguments, and is completely stateless.

pakim249CAL (approved):

ian-shim:
@pakim249CAL But the state retriever would get a new address and if offchain code points to the existing state retriever, it won't be able to access the updated method, right?

PR #1419, opened 2025-03-20
ci: Add job to check if PR title is using conventional commits

dmanc:

Adds workflow for enforcing [conventional commit](https://www.conventionalcommits.org/en/v1.0.0/) specification for our PR titles. This should make it easier to create release notes since the structure of the titles is standardized. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/.github/workflows/pr-title.yaml
+++ b/.github/workflows/pr-title.yaml
@@ -1,0 +1,31 @@
+name: PR Title Linting
+
+on:
+  pull_request:
+    types: [opened, edited, synchronize]
+
+jobs:
+  lint-pr-title:
+    runs-on: ubuntu-latest
+    name: Validate PR Title
+    steps:
+      - name: Fetch PR Title
+        run: |
+          PR_TITLE=$(jq -r '.pull_request.title' "$GITHUB_EVENT_PATH")
+          echo "PR title: $PR_TITLE"
+          
+          # Define the valid pattern (supports conventional commit format with breaking changes)
+          if [[ ! "$PR_TITLE" =~ ^(feat|fix|chore|docs|refactor|test|style|ci|perf)(\([a-z0-9-]+\))?(!)?:\ .* ]]; then
+            echo "❌ Invalid PR title: '$PR_TITLE'"
+            echo "Expected format: 'type[(scope)][!]: description'"
+            echo "Allowed types: feat, fix, chore, docs, refactor, test, style, ci, perf."
+            echo ""
+            echo "Examples of valid PR titles:"
+            echo "- feat: add user authentication"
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-title.yaml
```diff
@@ -0,0 +1,31 @@
+name: PR Title Linting
+
+on:
+  pull_request:
+    types: [opened, edited, synchronize]
```
samlaf, line 5: is this needed?

samlaf (approved):

.github/workflows/pr-title.yaml
dmanc: Yeah it triggers when the workflow will run. So maybe synchronize is not needed but I see no harm keeping it.

jianoaix (commented):
If the motivation is to create release, I don't think this can be used as the basis to increase the release version, or even be treated same in release note. Breaking something experimental is different than breaking something aphla or beta or GA.

dmanc:
> If the motivation is to create release, I don't think this can be used as the basis to increase the release version, or even be treated same in release note. Breaking something experimental is different than breaking something aphla or beta or GA.

Do you have a proposal?

jianoaix:
> > If the motivation is to create release, I don't think this can be used as the basis to increase the release version, or even be treated same in release note. Breaking something experimental is different than breaking something aphla or beta or GA.
> 
> Do you have a proposal?

yes: https://eigenlabs.slack.com/archives/C06JZQHN5R7/p1728002369961579

.github/workflows/pr-title.yaml
samlaf: I guess my question was: what is the default types value? Is it smaller or bigger than what is here? I would assume the default would just trigger on all of these 3 types, plus maybe more?

dmanc:
> > > If the motivation is to create release, I don't think this can be used as the basis to increase the release version, or even be treated same in release note. Breaking something experimental is different than breaking something aphla or beta or GA.
> > 
> > 
> > Do you have a proposal?
> 
> yes: https://eigenlabs.slack.com/archives/C06JZQHN5R7/p1728002369961579

That looks good to me. But this conventional commit is intended as a helper not a comprehensive solution for API changes. I'm going to go ahead and move forward with it.

.github/workflows/pr-title.yaml
dmanc: Defaults are opened, synchronize, and reopened. https://docs.github.com/en/actions/writing-workflows/choosing-when-your-workflow-runs/events-that-trigger-workflows#pull_request. So we would like edited as well

PR #1423, opened 2025-03-21
fix: separate the interference of Dispersal Response to Request index

jianoaix:
req

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/common/v2/blobstore/dynamo_metadata_store.go
+++ b/disperser/common/v2/blobstore/dynamo_metadata_store.go
@@ -897,7 +897,7 @@ func (s *BlobMetadataStore) GetDispersalRequestByDispersedAt(
 			OperatorDispersalIndexName,
 			"OperatorID = :pk AND DispersedAt BETWEEN :start AND :end",
 			commondynamodb.ExpressionValues{
-				":pk":    &types.AttributeValueMemberS{Value: operatorId.Hex()},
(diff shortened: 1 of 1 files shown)


pschork (approved):

disperser/common/v2/blobstore/dynamo_metadata_store.go
```diff
@@ -1606,7 +1614,7 @@ func MarshalDispersalRequest(req *corev2.DispersalRequest) (commondynamodb.Item,
 
 	fields["PK"] = &types.AttributeValueMemberS{Value: dispersalKeyPrefix + hashstr}
 	fields["SK"] = &types.AttributeValueMemberS{Value: fmt.Sprintf("%s%s", dispersalRequestSKPrefix, req.OperatorID.Hex())}
-	fields["OperatorID"] = &types.AttributeValueMemberS{Value: req.OperatorID.Hex()}
+	fields["OperatorID"] = &types.AttributeValueMemberS{Value: fmt.Sprintf("%s%s", dispersalRequestSKPrefix, req.OperatorID.Hex())}
```
jianoaix, line 1617: I'm using `dispersalResponseSKPrefix` here, which is also used above for `SK`. The `OperatorID` is used for secondary index's PK, so it seems fine to reuse this prefix.

PR #1425, opened 2025-03-21
perf: add KV cache for lookup requests in data apis

jianoaix:
Cache the lookup requests will improve latency, and more importantly reduce the traffic hitting the database (which is shared between dataapi and disperser so will need protection from unnecessary interference).

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/dataapi/v2/batches.go
+++ b/disperser/dataapi/v2/batches.go
@@ -194,18 +194,26 @@ func (s *ServerV2) FetchBatch(c *gin.Context) {
 		errorResponse(c, errors.New("invalid batch header hash"))
 		return
 	}
-	batchHeader, attestation, err := s.blobMetadataStore.GetSignedBatch(c.Request.Context(), batchHeaderHash)
-	if err != nil {
-		s.metrics.IncrementFailedRequestNum("FetchBatch")
-		errorResponse(c, err)
-		return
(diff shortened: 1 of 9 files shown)


disperser/dataapi/v2/batches.go
```diff
@@ -194,18 +194,28 @@ func (s *ServerV2) FetchBatch(c *gin.Context) {
 		errorResponse(c, errors.New("invalid batch header hash"))
 		return
 	}
-	batchHeader, attestation, err := s.blobMetadataStore.GetSignedBatch(c.Request.Context(), batchHeaderHash)
-	if err != nil {
-		s.metrics.IncrementFailedRequestNum("FetchBatch")
-		errorResponse(c, err)
-		return
+	signedBatch, ok := s.signedBatchCache.Get(batchHeaderHashHex)
+	if !ok {
```
pschork, line 198: I would rename `ok`->`cached` here and elsewhere. It makes the cache check easier to read `if !cached {`

pschork (commented):
LGTM. Only minor var name change request.

disperser/dataapi/v2/batches.go
jianoaix: Sg

PR #1427, opened 2025-03-24
feat: deploy contract toml config

pakim249CAL:
Use TOML config in deployment script. Enables comments mainly, and I think TOML is just generally preferred over JSON when possible since it can be cleaner.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


contracts/script/deploy/verifiable/PrintMultisigCalldata.s.sol
```diff
@@ -36,7 +34,6 @@ contract PrintMultisigCalldata is Script {
 
     /// @dev override this if you don't want to use the environment to get the config path
     function _initConfig() internal virtual {
-        deploymentInitializer = vm.envAddress("DEPLOYMENT_INITIALIZER");
```
samlaf, line 39: was this just not used? removing this doesn't seem related to PR title/description.

samlaf (approved):
LGTM, very much in favor of json->toml

contracts/script/deploy/verifiable/PrintMultisigCalldata.s.sol
pakim249CAL: Yes, It was just something I caught while I was working on the PR. I can separate it out, but I think this PR is still acceptably small.

PR #1429, opened 2025-03-24
feat: littDB validator integration

cody-littley:

Enables littDB as the validator database.


api/clients/eigenda_client_e2e_test.go
```diff
@@ -22,9 +22,9 @@ func init() {
 // We don't test waiting for finality because that adds 12 minutes to the test, and is not necessary
 // because we already test for this in the unit tests using a mock disperser which is much faster.
 func TestClientUsingTestnet(t *testing.T) {
-	if !runTestnetIntegrationTests {
-		t.Skip("Skipping testnet integration test")
-	}
+	//if !runTestnetIntegrationTests { // TODO
+	//}
+	t.Skip("Skipping testnet integration test")
```
cody-littley, line 27: This test is currently failing on multiple different branches.

litt3: This test randomly started working for me on a different branch, I wonder if testnet was just having trouble

node/config.go
```diff
@@ -108,6 +108,21 @@ type Config struct {
...
+
+	// The size of the thread pool used to download chunks from the network is computed by the following formula:
+	// (number of CPUs) * DownloadPoolMultiplier + DownloadPoolConstant
+	DownloadPoolMultiplier int
+
+	// The size of the thread pool used to download chunks from the network is computed by the following formula:
+	// (number of CPUs) * DownloadPoolMultiplier + DownloadPoolConstant
+	DownloadPoolConstant int
+
+	// If true, use littDB instead of levelDB for v2 storage. Note than in its current form, no data migration is
```
litt3, line 120: > Note than in its current form, no data migration is performed when this setting is enabled.

Is this true? Seems there are migration considerations in `validator_store.go`

node/validator_store.go
```diff
@@ -0,0 +1,540 @@
...
+			return nil, fmt.Errorf("failed to get key builder for migration: %v", err)
+		}
+		migrationKey := migrationKeyBuilder.Key([]byte("migrationCompleteTime"))
+
+		if !littDBExists {
+			// This is the first time we are starting up with littDB enabled and there is potentially data still
+			// in the levelDB. Start the data migration from levelDB to littDB.
+
+			migrationComplete = timeSource().Add(ttl)
+			logger.Infof("Begining data migration from levelDB to littDB. Migration will be completed at %s",
```
litt3, line 202: Begining -> Beginning

node/validator_store.go
```diff
@@ -0,0 +1,540 @@
...
+			LevelDBDisableSeeksCompaction: config.LevelDBDisableSeeksCompactionV2,
+			LevelDBSyncWrites:             config.LevelDBSyncWritesV2,
+		})
+		if err != nil {
+			return nil, fmt.Errorf("failed to create new tablestore: %w", err)
+		}
+	}
+
+	// Set up migration if necessary.
+	var migrationComplete time.Time
```
litt3, line 187: Can you explain the logic behind `migrationComplete`? Having not yet reviewed the migration code below, it's confusing to me why this is needed

node/validator_store.go
```diff
@@ -0,0 +1,540 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to create new litt config: %w", err)
+		}
+
+		littDB, err = littbuilder.NewDB(littConfig)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create new litt store: %w", err)
+		}
+
+		chunkTable, err = littDB.GetTable("chunks")
```
litt3, line 254: Any reason you hardcode "chunks" here, instead of having a const defined above, like for the other table names?

node/validator_store.go
```diff
@@ -0,0 +1,540 @@
...
+
+	// GetBundleData returns the chunks of a blob with the given bundle key.
+	// The returned chunks are encoded in bundle format.
+	GetBundleData(bundleKey []byte) ([]byte, error)
+
+	// Stop stops the store.
+	Stop() error
+}
+
+type validatorStore struct {
```
litt3, line 69: Would it make more sense to separate the levelDB impl from the littDB impl of `ValidatorStore`?

node/validator_store.go
```diff
@@ -0,0 +1,540 @@
...
+		if s.levelDB == nil {
+			// There is no data in levelDB.
+			return nil, fmt.Errorf("failed to get chunks: not found")
+		}
+
+		s.migrationLock.RLock()
+		defer s.migrationLock.RUnlock()
+
+		if s.levelDB == nil {
+			// The migration completed while we were waiting for the lock.
```
litt3, line 442: would it make sense to reattempt a `get` from littdb in this case?

common/stage_timer.go
```diff
@@ -78,13 +94,29 @@ func (p *SequenceProbe) SetStage(stage string) {
 	}
 
 	now := time.Now()
-	elapsed := now.Sub(p.currentStageStart)
-	p.stageTimer.stageLatency.WithLabelValues(p.currentStage).Observe(ToMilliseconds(elapsed))
+	p.stageTimer.stageCount.WithLabelValues(stage).Inc()
+
+	if !p.initialized {
```
ian-shim, line 99: is this thread safe?

node/store.go
```diff
@@ -51,7 +51,7 @@ func NewLevelDBStore(
 	syncWrites bool,
 	storeDurationBlocks uint32) (*Store, error) {
 
-	// Create the db at the path. This is currently hardcoded to use
+	// Create the levelDB at the path. This is currently hardcoded to use
 	// levelDB.
```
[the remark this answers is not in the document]
litt3: I suspect an IDE rename that exceeded its mandate

node/node.go
```diff
@@ -104,12 +105,13 @@ func NewNode(
...
 
-	eigenMetrics := metrics.NewEigenMetrics(AppName, fmt.Sprintf(":%d", config.MetricsPort), reg, logger.With("component", "EigenMetrics"))
+	socketAddr := fmt.Sprintf(":%d", config.MetricsPort)
+	eigenMetrics := metrics.NewEigenMetrics(AppName, socketAddr, reg, logger.With("component", "EigenMetrics"))
 
 	// Make sure config folder exists.
 	err := os.MkdirAll(config.DbPath, os.ModePerm)
 	if err != nil {
-		return nil, fmt.Errorf("could not create db directory at %s: %w", config.DbPath, err)
+		return nil, fmt.Errorf("could not create levelDB directory at %s: %w", config.DbPath, err)
```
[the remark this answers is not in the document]
litt3: I suspect an IDE rename that exceeded its mandate

litt/littbuilder/db_impl.go
```diff
@@ -184,7 +184,8 @@ func (d *db) DropTable(name string) error {
 
 	table, ok := d.tables[name]
 	if !ok {
-		return fmt.Errorf("table %s does not exist", name)
+		// Table does not exist, nothing to do.
```
ian-shim, line 187: nit: a WARN/ERROR log would be useful here

node/grpc/server_v2.go
```diff
@@ -143,27 +142,100 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
 		}
 	}
 
-	s.logger.Info("new StoreChunks request", "batchHeaderHash", hex.EncodeToString(batchHeaderHash[:]), "numBlobs", len(batch.BlobCertificates), "referenceBlockNumber", batch.BatchHeader.ReferenceBlockNumber)
-	operatorState, err := s.node.ChainState.GetOperatorStateByOperator(ctx, uint(batch.BatchHeader.ReferenceBlockNumber), s.node.Config.ID)
+	probe.SetStage("get_operator_state")
```
ian-shim, line 145: will this stage capture not just `GetOperatorStateByOperator` but also DownloadBundles and validateAndStoreV2Chunks?

node/grpc/server_v2.go
```diff
@@ -143,27 +142,100 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
...
+	if err != nil {
+		return nil, api.NewErrorInternal(fmt.Sprintf("failed to sign batch: %v", err))
+	}
+
+	return &pb.StoreChunksReply{
+		Signature: sig,
+	}, nil
+}
+
+func (s *ServerV2) validateAndStoreV2Chunks(
```
ian-shim, line 179: nit: `validateAndStoreChunks` since the struct already indicates V2

node/grpc/server_v2.go
```diff
@@ -173,41 +245,58 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
...
+	ctx context.Context,
+	batch *corev2.Batch,
+	blobShards []*corev2.BlobShard,
+	batchData []*node.BundleToStore,
+	operatorState *core.OperatorState,
+	batchHeaderHash [32]byte,
+	probe *common.SequenceProbe,
+) error {
+	probe.SetStage("validate")
+	err := s.node.ValidateBatchV2(ctx, batch, blobShards, operatorState)
```
ian-shim, line 286: don't we want to parallelize storage and validation on LittDB?

node/grpc/server_v2_test.go
```diff
@@ -320,6 +322,7 @@ func TestV2StoreChunksStorageFailure(t *testing.T) {
 func TestV2StoreChunksValidationFailure(t *testing.T) {
 	config := makeConfig(t)
 	config.EnableV2 = true
+	config.LittDBEnabled = true
```
ian-shim, line 325: does this remove the test case for levelDB? shouldn't we run the same test cases for both DBs?

node/node.go
```diff
@@ -195,10 +197,28 @@ func NewNode(
...
+		"churnerUrl", config.ChurnerUrl,
+		"quorumIDs", fmt.Sprint(config.QuorumIDList),
+		"registerNodeAtStart", config.RegisterNodeAtStart,
+		"pubIPCheckInterval", config.PubIPCheckInterval,
+		"eigenDAServiceManagerAddr", config.EigenDAServiceManagerAddr,
+		"blockStaleMeasure", blockStaleMeasure,
+		"storeDurationBlocks", storeDurationBlocks,
+		"enableGnarkBundleEncoding", config.EnableGnarkBundleEncoding)
+
+	numCPUs := runtime.NumCPU()
```
ian-shim, line 216: I don't think this gives an accurate number of vCPUs when you run it inside a deployment

node/node.go
```diff
@@ -195,10 +197,28 @@ func NewNode(
...
+	downloadPoolSize := config.DownloadPoolMultiplier*numCPUs + config.DownloadPoolConstant
```
ian-shim, line 217: What's the rationale behind this equation? Download doesn't consume a lot of CPU cycles. I wonder why it scales with the # CPUs

ian-shim (commented):
initial skim, will review validator store by EOD

anupsv (approved):

common/index_lock.go
```diff
@@ -0,0 +1,30 @@
...
+func NewIndexLock(numLocks uint32) *IndexLock {
+	locks := make([]sync.Mutex, numLocks)
+	return &IndexLock{locks: locks}
+}
+
+// Lock locks the given index. Two calls to Lock with the same index will attempt to acquire the same lock.
+// Two calls to Lock with different indices may or may not acquire the same lock. After calling lock,
+// the caller must eventually also call Unlock.
+func (i *IndexLock) Lock(index uint64) {
+	lockIndex := index % uint64(len(i.locks))
```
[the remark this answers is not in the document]
cody-littley:
```// Internally, an IndexLock keeps an array of mutexes. Each index is mapped onto one of these mutexes in the array,
// such that the same index always maps to the same mutex. Note that due to this mapping, otherwise unrelated indices
// may end up using the same mutex. Increasing the number of locks will decrease the probability of unrelated indices
// contending for the same lock, but will also increase memory usage.
```

node/grpc/metrics_v2.go
```diff
@@ -64,7 +55,7 @@ func NewV2Metrics(logger logging.Logger, registry *prometheus.Registry) (*Metric
 		prometheus.SummaryOpts{
 			Namespace:  namespace,
 			Name:       "get_chunks_latency_ms",
-			Help:       "The latency of a GetChunks() RPC call.",
+			Help:       "The latency of a GetBundleData() RPC call.",
```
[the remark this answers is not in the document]
cody-littley: Unintentional, fixed.

node/grpc/metrics_v2.go
```diff
@@ -74,19 +65,21 @@ func NewV2Metrics(logger logging.Logger, registry *prometheus.Registry) (*Metric
 		prometheus.GaugeOpts{
 			Namespace: namespace,
 			Name:      "get_chunks_data_size_bytes",
-			Help:      "The size of the data requested to be retrieved by GetChunks() RPC calls.",
+			Help:      "The size of the data requested to be retrieved by GetBundleData() RPC calls.",
```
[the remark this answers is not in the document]
cody-littley: unintentional, fixed

node/grpc/server_v2.go
```diff
@@ -143,27 +142,100 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
...
+	batch *corev2.Batch,
+	blobShards []*corev2.BlobShard,
+	rawBundles []*node.RawBundles,
+	operatorState *core.OperatorState,
+	batchHeaderHash [32]byte,
+	probe *common.SequenceProbe,
+) error {
+
+	batchData := make([]*node.BundleToStore, 0, len(rawBundles))
+	for _, bundles := range rawBundles {
```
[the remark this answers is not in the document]
cody-littley: There is an inner for loop that iterates over something called a "bundle":

```
for quorum, bundle := range bundles.Bundles {
```

Honestly, I'm not really sure why these things are called the things they are. I mostly just carried forward the existing terminology. We should certainly revisit this naming schema, but probably in a future PR.

node/grpc/server_v2.go
```diff
@@ -143,27 +142,100 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
...
+		return s.validateAndStoreV2ChunksLittDB(
+			ctx,
+			batch,
+			blobShards,
+			batchData,
+			operatorState,
+			batchHeaderHash,
+			probe)
+	} else {
+		probe.SetStage("validate_and_store")
```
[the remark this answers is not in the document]
cody-littley: The problem with this in the levelDB implementation is that storage and validation happen in parallel to each other, and the stage timer utility is not good at reporting on the timing of parallel operations.

Although I could in theory spend more time on v1 metrics here, I'm kind of tempted to leave it to a later date. The plan is to enable littDB by default, so if all goes to plan we will never execute this v2 levelDB pathway outside of testing.

node/grpc/server_v2.go
```diff
@@ -173,41 +245,58 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
...
+	batchData []*node.BundleToStore,
+	operatorState *core.OperatorState,
+	batchHeaderHash [32]byte,
+	probe *common.SequenceProbe,
+) error {
+	probe.SetStage("validate")
+	err := s.node.ValidateBatchV2(ctx, batch, blobShards, operatorState)
 	if err != nil {
-		return nil, api.NewErrorInternal(fmt.Sprintf("failed to sign batch: %v", err))
+		return api.NewErrorInternal(fmt.Sprintf("failed to validate batch %s: %v", batchHeaderHash, err))
```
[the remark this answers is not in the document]
cody-littley: done

node/config.go
```diff
@@ -108,6 +108,21 @@ type Config struct {
...
+	// performed when this setting is enabled. (Migration is a feature we will need prior to deployment to mainnet.)
+	LittDBEnabled bool
+
+	// A special test only setting. If true, then littDB will throw an error if the same data is written twice.
+	LittDBDoubleWriteProtection bool
```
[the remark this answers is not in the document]
cody-littley: Double write protection is a test only feature that would warn us if we use an illegal littDB write pattern. I use it in `TestDoubleInsertionLittDB()` in `validator_store_test.go`.

node/node.go
```diff
@@ -195,10 +197,28 @@ func NewNode(
...
+		"retrievalPort", config.RetrievalPort,
+		"v2RetrievalPort", config.V2RetrievalPort,
+		"churnerUrl", config.ChurnerUrl,
+		"quorumIDs", fmt.Sprint(config.QuorumIDList),
+		"registerNodeAtStart", config.RegisterNodeAtStart,
+		"pubIPCheckInterval", config.PubIPCheckInterval,
+		"eigenDAServiceManagerAddr", config.EigenDAServiceManagerAddr,
+		"blockStaleMeasure", blockStaleMeasure,
+		"storeDurationBlocks", storeDurationBlocks,
+		"enableGnarkBundleEncoding", config.EnableGnarkBundleEncoding)
```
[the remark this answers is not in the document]
cody-littley: Unfortunately I'm running low on neck grease, and it hurts when I have to turn my head so far to see the end of a line like this.

node/node_test.go
```diff
@@ -94,14 +95,15 @@ func newComponents(t *testing.T, operatorID [32]byte) *components {
...
+		Config:       config,
+		Logger:       logger,
+		KeyPair:      keyPair,
+		Metrics:      nil,
+		Store:        store,
+		ChainState:   chainState,
+		Validator:    mockVal,
+		Transactor:   tx,
+		DownloadPool: workerpool.New(1),
 	}
```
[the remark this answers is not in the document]
cody-littley: As far as I can tell, these tests are exercising things that aren't DB related, so I didn't think it would add much.

node/store.go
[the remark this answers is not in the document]
cody-littley:
```
// Create the DB at the path.
```

node/validator_store.go
```diff
@@ -0,0 +1,540 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+const (
+	BatchHeaderTableName     = "batch_headers"
+	BlobCertificateTableName = "blob_certificates"
+	BundleTableName          = "bundles"
+	MigrationTableName       = "migration"
+
+	// LevelDBPath is he path where the levelDB database is stored.
```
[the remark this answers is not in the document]
cody-littley: fixed

node/config.go
cody-littley: Reduced comment to `// If true, use littDB instead of levelDB for v2 storage.`. When I originally wrote this comment, I hadn't yet implemented migration. Once I did add migration, I forgot to circle back.

node/validator_store.go
cody-littley: fixed

node/validator_store.go
cody-littley:
```
	// Set up migration if necessary. The migrationComplete time is the time when all data in the old levelDB instance
	// has exceeded its TTL. If the TTL is 2 weeks, then the migrationComplete time will be 2 weeks from now. At that
	// moment, it becomes safe to permanently stop and delete the levelDB database.
```

node/validator_store.go
cody-littley: No reason, moved magic strings into constants.

node/validator_store.go
cody-littley: Nope. This condition can only trigger in the following circumstances:

- we've looked for the data, and it isn't in littDB
- the data may not be in levelDB either, in which case we do no harm saying "we can't find it"
- the data may be in levelDB
    - the data is about to expire
    - the data is some of the last data in levelDB to expire
    - when the user asks for the data, it has not yet expired
    - before the data is returned to the user, the last data expires in levelDB, and we delete the entire database

Migration is not a true "migration", in that we aren't moving data from levelDB into littDB. We are actually just waiting for all of the levelDB data to exceed its TTL, after which we delete the database in its entirety.

common/stage_timer.go
cody-littley:
```
// A SequenceProbe is not thread safe. It is intended for use in measuring a linear sequence of operations. Do not call
// SetStage or End from multiple goroutines at the same time.
```

litt/littbuilder/db_impl.go
cody-littley:
```
d.logger.Infof("table %s does not exist, cannot drop", name)
```

node/grpc/server_v2.go
cody-littley: The `probe` object is passed into both `DownloadBundles()` and `validateAndStoreV2Chunks()`, and those functions appropriately set the stage before doing any nontrivial work.

node/grpc/server_v2.go
cody-littley: renamed

node/grpc/server_v2.go
cody-littley: Two factors at play here.

- LittDB doesn't let you delete data once you write it, so there isn't a way we can back data out of the DB if we later decide it isn't valid.
- Parallelizing these things stands to shave ~100ms off of the end-to-end latency. At the current performance profile, this isn't high bang for the buck (although we may want to revisit it later).

Screenshot 2025-04-22 at 12 01 18 PM

Screenshot 2025-04-22 at 12 01 34 PM

Screenshot 2025-04-22 at 12 01 51 PM

node/grpc/server_v2_test.go
cody-littley: Good point, added a test to cover levelDB.

node/node.go
cody-littley: I've changed this to a static setting that doesn't scale with number of CPUs.

node/node.go
cody-littley: I've changed this to a static setting that doesn't scale with number of CPUs.

node/grpc/server_v2_test.go
```diff
@@ -365,7 +368,9 @@ func TestV2StoreChunksValidationFailure(t *testing.T) {
 	require.Nil(t, reply.GetSignature())
 	requireErrorStatus(t, err, codes.Internal)
 
-	c.store.AssertCalled(t, "DeleteKeys", mock.Anything, mock.Anything)
+	if !config.LittDBEnabled {
```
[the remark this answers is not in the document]
cody-littley: Originally both littDB and levelDB were tested in this same file, now they aren't. Removed if statement.

node/validator_store.go
```diff
@@ -0,0 +1,559 @@
...
+	if !config.LittDBEnabled {
+		logger.Warn("WARNING: This node is running with littDB disabled. " +
+			"This is a deprecated mode of operation, and will not be supported in future versions.")
+	}
+
+	littDBPath := path.Join(config.DbPath, LittDBPath)
+	levelDBPath := path.Join(config.DbPath, LevelDBPath)
+	levelDBDeletionPath := path.Join(config.DbPath, LevelDBDeletionPath)
+
+	// If we previously made an attempt at deleting the levelDB database but it was interrupted, delete it now.
```
ian-shim, line 138: This migration logic seems worth having test cases for

node/validator_store.go
```diff
@@ -0,0 +1,559 @@
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to create new tablestore: %w", err)
+		}
+	}
+
+	// Set up migration if necessary. The migrationComplete time is the time when all data in the old levelDB instance
+	// has exceeded its TTL. If the TTL is 2 weeks, then the migrationComplete time will be 2 weeks from now. At that
+	// moment, it becomes safe to permanently stop and delete the levelDB database.
+	var migrationComplete time.Time
+	if config.LittDBEnabled && levelDBExists {
```
ian-shim, line 201: Is it handling a case where it turns LittDB on/off back and forth?
i.e. LittDB is enabled -> sets the migration complete time -> LittDB is disabled (new data is written in levelDB) -> LittDB is enabled again. It doesn't seem like it will update the migration complete time again. 
We should either support this use case or make it clear that enabling LittDB is one way door?

node/validator_store.go
```diff
@@ -0,0 +1,559 @@
...
+		littConfig.MetricsEnabled = true
+		littConfig.MetricsRegistry = registry
+		littConfig.MetricsNamespace = littDBMetricsPrefix
+		littConfig.Logger = logger
+		littConfig.DoubleWriteProtection = config.LittDBDoubleWriteProtection
+		if err != nil {
+			return nil, fmt.Errorf("failed to create new litt config: %w", err)
+		}
+
+		littDB, err = littbuilder.NewDB(littConfig)
```
ian-shim, line 262: Will it load the existing LittDB if it has been created before?

node/validator_store.go
```diff
@@ -0,0 +1,559 @@
...
+	// LevelDBPath is the path where the levelDB database is stored.
+	LevelDBPath = "chunk_v2"
+	// LevelDBDeletionPath is the path where the levelDB database is stored while it is being deleted (for atomicity).
+	LevelDBDeletionPath = "chunk_v2_deleted"
+	// LittDBPath is the path where the littDB database is stored.
+	LittDBPath = "chunk_v2_litt"
+
+	// The name of the littDB table containing chunk data.
+	chunksTableName = "chunks"
+	// A legacy littDB table, this will exist until all old implementations are migrated and delete this table.
```
ian-shim, line 42: do we need to handle this? What's the legacy LittDB table if it hasn't been rolled out yet?

node/validator_store.go
```diff
@@ -0,0 +1,559 @@
...
+
+	return store, nil
+}
+
+func (s *validatorStore) StoreBatch(batchHeaderHash []byte, batchData []*BundleToStore) ([]kvstore.Key, uint64, error) {
+	if len(batchData) == 0 {
+		return nil, 0, fmt.Errorf("no batch data")
+	}
+
+	if s.littDB != nil {
```
ian-shim, line 313: should it also check `LittDBEnabled`? It looks like `s.littDB != nil` if and only if `LittDBEnabled`, but if the initialization logic changes, it seems it's possible that we still write to LittDB even when `LittDBEnabled` is false

ian-shim (commented):
Looks good! Few questions. Also, not necessarily in this PR, but we should update the integration tests (and potentially add more test cases)

node/validator_store.go
cody-littley: The backwards migration is intentionally not supported. Added the following to the godoc:

```
	// If true, use littDB instead of levelDB for v2 storage. Once enabled, cannot be disabled. That is,
	// once littDB is turned on, turning it off again is not supported.
	LittDBEnabled bool
```

node/validator_store.go
cody-littley: Yes. `NewDB()` is just the contstructor. You use the same constructor when starting a new database and when loading an existing one from disk.

node/validator_store.go
cody-littley: Earlier iterations of this branch created this table, and we deployed that to preprod. Since we don't want to reset the drives on preprod, I wanted to make sure I cleaned that environment up. This cleanup becomes a no-op if run in environments without this legacy table.

node/validator_store.go
cody-littley: `LittDBEnabled` is not saved in this struct's state, and is only available in the constructor. Do you think it's worth it to store the boolean flag in the `validatorStore` struct?

ian-shim: Got it. I think it's fine as long as we keep it consistent and dont end up doing things like `s.config.LittDBEnabled` outside the constructor.

ian-shim (approved):
lgtm! I still think we should have tests for DB migration logic and updated integration tests using LittDB

PR #1431, opened 2025-03-24
perf: apply query limit to the feed fetch

jianoaix:
We can actually apply this limit, based on documentation https://docs.aws.amazon.com/amazondynamodb/latest/APIReference/API_Query.html

`A single Query operation will read up to the maximum number of items set (if using the Limit parameter) or a maximum of 1 MB of data and then apply any filtering to the results using FilterExpression.`

That is, if we are not using filter expression, which is the case here, it'll return exactly `limit` items (if there are actually at least `limit` items and it doesn't go above 1MB).

(this nuance wasn't understood well when the code was written initially)

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


jianoaix:
ping?

dmanc (approved):

PR #1437, opened 2025-03-31
chore(proto): push protos to buf BSR registry

samlaf:

Closes DAINT-395.

Goal was just to add rust bindings, but ended up (re)discovering buf's protobuf registry [BSR](https://buf.build/product/bsr).
This is actually super nice, full-fledged product offering (they did raise 68M+..), which is also free for open-source repos.
We get nicer documentation, linting, auto-gen, sdks in every language for free, and more.
So this not only solves the original problem ([generating rust bindings](https://buf.build/blog/bsr-generated-sdks-for-rust)), but versions them, and also generates golang sdks which we could eventually migrate to.

Created an eigenda org here: https://buf.build/layrlabs/eigenda
Tested the rust bindings which work fine; added a README documenting a basic v1 dispersal example in https://buf.build/layrlabs/eigenda/sdks/main:community/neoeinstein-tonic

Also added a CI to buf lint, format, breaking (will catch breaking upgrades to our proto files and force a version upgrade).
Following https://buf.build/docs/ci-cd/github-actions/
Still missing a `secrets.BUF_TOKEN` auth token, which I can't add to this repo because I don't have admin access on github. Asked @anupsv to create that token.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/.github/workflows/buf-proto.yaml
+++ b/.github/workflows/buf-proto.yaml
@@ -1,0 +1,33 @@
+# Copied from https://buf.build/docs/ci-cd/github-actions/
+#
+# With the default configuration above, the Action does the following:
+# - On a pull request, it runs all checks (buf build, buf lint, buf format, and buf breaking) and posts a summary comment on the PR.
+# - When you push a Git commit, tag, or branch to GitHub, it pushes named modules to the BSR using buf push.
+# - When you delete a Git branch or tag, it archives the corresponding label on the BSR.
+#
+# This workflow could be merged with compile-protobufs.yaml, but we keep them separate for now,
+# given that this workflow requires PR write permissions.
+# If buf works well I think we should eventually migrate all manual protoc stuff to using buf.
+# This would give us versioned golang bindings, as well as better documentation, and a bunch of other stuff.
+name: Buf Proto
+on:
+  push:
+  pull_request:
+    types: [opened, synchronize, reopened, labeled, unlabeled]
+  delete:
+permissions:
+  contents: read
+  pull-requests: write
+jobs:
+  buf:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - uses: bufbuild/buf-action@v1
+        with:
+          token: ${{ secrets.BUF_TOKEN }}
+          # Change setup_only to true if you only want to set up the Action and not execute other commands.
+          # Otherwise, you can delete this line--the default is false.
+          setup_only: false
+          # Optional GitHub token for API requests. Ensures requests aren't rate limited.
+          github_token: ${{ secrets.GITHUB_TOKEN }}
--- a/api/buf.yaml
+++ b/api/buf.yaml
@@ -1,0 +1,15 @@
+# We currently only push the eigenda proto files to the buf registry
+# in order for it to generate rust prost/tonic bindings for us.
+# We also get nice documentation and versioning: see https://buf.build/layrlabs/eigenda.
+
+# For details on buf.yaml configuration, visit https://buf.build/docs/configuration/v2/buf-yaml
+version: v2
+modules:
+  - path: proto
+    name: buf.build/layrlabs/eigenda
+lint:
+  use:
+    - STANDARD
+breaking:
+  use:
+    - FILE


samlaf:
@cody-littley made format and push CIs pass. Lint and Breaking (finds breaking changes) are failing though. 2 ways to fix:
1. I turn them off
2. I make the required changes. Might be some breaking changes in there... I'm not sure. Or maybe its just simple name changes that will easily be propagated to golang changes which I can fix.

I think the main difficulty is making breaking pass. `breaking` ci would be super nice moving forward because it catches any breaking api changes, and forces us to release a new protobuf version if so. One way that we've shot ourselves in the foot is that we've named our protofiles v2, which might (?) conflict with actual protobuf versioning (v2_v2). @jianoaix I feel like you might like this though if we fix this (perhaps we can chat). You've been meaning to add breaking change release docs and do better maintaining of API breaking changes right? This might be the perfect tool.

jianoaix:
> https://buf.build/layrlabs/eigenda

Yea it sounds nice to catch people breaking the API silently (maybe until it's in production and breaking users which is bad).

I didn't know Buf before:
Is it like the pkg.go.dev for protobuf?
Is there risk that this will stop being free?
If we have the docs there, would it have users go there and use it (will it have utility)?

samlaf:
@jianoaix sorry was travelling just found time to finish this now.
Ended up not fixing any of the lint warnings as that would create breaking changes. Added ignore warnings throughout the buf files instead (see [f96f2b3](https://github.com/Layr-Labs/eigenda/pull/1437/commits/f96f2b3a95ac1ced35d6582b4e83d2c66fac5d60)) . Future messages/services will be forced to follow the lint suggestions though, which I think will be helpful.

This PR is ready to merge.The `breaking` CI is failing because its trying to compare against the master commit, which didn't have a buf.yaml file, so the import paths are not working properly. I'm not sure how to fix this... I suggest just merging without fixing this, and future PRs will work because they will target a commit which contains a buf.yaml file, and will prevent creating breaking API changes.

> I didn't know Buf before: Is it like the pkg.go.dev for protobuf? 

Not quite. Think you meant to ask whether its a central package registry. Buf is a protobuf ecosystem tooling project, which contains a protobuf registry called BSR (buf schema registry). This is like rust's crates.io, ghcr for docker images, npm for node packages, etc.

Golang actually works different from most languages, in that it doesn't have a central registry. pkg.go.dev is actually just a package crawler. Its more of a search engine. The actual packages are stored and hosted in VCS websites like github, gitlab, etc. This is actually a pretty cool feature of the golang ecosystem. See for example https://2019.jsconf.eu/c-j-silverio/the-economics-of-open-source.html for the downsides of central registries.

> Is there risk that this will stop being free? 

I guess there is, but I would guess that it will follow the freemium model that it currently does, and most people seem to do, meaning that its free for open-source public repos. If BSR becomes not free and we don't want to pay for it, we could always just generate the bindings locally and push them to the repo, the same way that we do for the golang bindings right now.

> If we have the docs there, would it have users go there and use it (will it have utility)?

Same response as above. I think it would be quite nice to send users there as its a LOT more readable. But if we are scared of buf somehow dissapearing or starting to charge exorbitant fee, we don't have to, and can point our users to our ugly html or md files instead lol.

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -1,18 +1,28 @@
...
 package disperser.v2;
+
 import "common/common.proto";
 import "common/v2/common_v2.proto";
+
 option go_package = "github.com/Layr-Labs/eigenda/api/grpc/disperser/v2";
 
 // Disperser defines the public APIs for dispersing blobs.
+//
+// buf:lint:ignore SERVICE_SUFFIX // Skip this rule to not create a breaking change.
```
jianoaix, line 11: here and below: why do we skip so many rules?

samlaf: Because I didn't want to create breaking changes. Otherwise the names of the packages/services/requests/responses would have to change to meet the proto code guidelines, but that would mean any people using those would now have to deal with the breaking change.

3 options here, which do you prefer?:
1. we leave as is here skipping any previous names that don't follow standard. We'll only follow the linter for future packages/services/requests/responses
2. we make the breaking changes
3. we keep these old names, mark them as deprecated, and create new ones that follow the convention

samlaf: The only thing we can do afaiu is ignore entire files from being linted, by adding them to `buf.yaml`.
I opted against that option in case we decide to add new messages/services to the current files, since then they would also not be linted.
What we could do though is to ignore all of our current files, and make sure any new additions are added to new files

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -158,29 +174,42 @@ enum BlobStatus {
...
+  // buf:lint:ignore ENUM_VALUE_PREFIX // Skip this rule to not create a breaking change.
+  // buf:lint:ignore ENUM_ZERO_VALUE_SUFFIX // Skip this rule to not create a breaking change.
   UNKNOWN = 0;
 
   // QUEUED means that the blob has been queued by the disperser for processing.
   // The DisperseBlob API is asynchronous, meaning that after request validation, but before any processing,
   // the blob is stored in a queue of some sort, and a response immediately returned to the client.
+  //
+  // buf:lint:ignore ENUM_VALUE_PREFIX // Skip this rule to not create a breaking change.
   QUEUED = 1;
```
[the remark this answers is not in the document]
samlaf: You can see the justification for this rule here: https://buf.build/docs/lint/rules/#enum_value_prefix

TLDR is that enums are package-scoped, so 2 different enums couldn't both have a QUEUED value, since you would have a conflict. This is actually same as golang. I always add enum name as a prefix in my golang enums, so I think its absolutely fine to do same here.

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -1,18 +1,28 @@
...
 // Disperser defines the public APIs for dispersing blobs.
+//
+// buf:lint:ignore SERVICE_SUFFIX // Skip this rule to not create a breaking change.
 service Disperser {
   // DisperseBlob accepts blob to disperse from clients.
   // This executes the dispersal asynchronously, i.e. it returns once the request
   // is accepted. The client could use GetBlobStatus() API to poll the the
   // processing status of the blob.
+  //
+  // buf:lint:ignore RPC_REQUEST_STANDARD_NAME // Skip this rule to not create a breaking change.
```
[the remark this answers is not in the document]
samlaf: Yes you might be right. I can remove this if you decide to go with this approach of keeping ignore statements everywhere, instead of the ignoring-entire-files appraoch (see https://github.com/Layr-Labs/eigenda/pull/1437#discussion_r2038795797).

api/proto/disperser/v2/disperser_v2.proto
jianoaix: Ah looks like this comes just a bit late, as it may catch the v2 proto changes.
I personally prefer not to be too intrusive on existing protos like adding a lot ignore comments (they are visible to readers as we generate docs), unless we have strong reasons to do so.

samlaf: Ok removed all of them and added to `ignore_only` section in buf.yaml

jianoaix (approved):

PR #1441, opened 2025-04-02
feat: add metrics to payload disperser

cody-littley:

Adds optional metrics to the payload disperser. Fixes performance bottleneck exposed by new metrics.


api/clients/v2/payloaddispersal/payload_disperser.go
```diff
@@ -20,21 +22,24 @@ type PayloadDisperser struct {
 	config          PayloadDisperserConfig
 	disperserClient clients.DisperserClient
 	certVerifier    clients.ICertVerifier
+	stageTimer      *common.StageTimer
 }
 
 // NewPayloadDisperser creates a PayloadDisperser from subcomponents that have already been constructed and initialized.
+// If the registry is nil then no metrics will be collected.
```
litt3, line 29: this doc would seem better placed with the `registry` param

api/clients/v2/disperser_client.go
```diff
@@ -128,10 +145,44 @@ func (c *disperserClient) DisperseBlob(
...
+		c.accountantLock.Unlock()
+		return nil, [32]byte{}, fmt.Errorf("error accounting blob: %w", err)
+	}
+
+	if payment.CumulativePayment.Sign() == 0 {
+		// This request is using reserved bandwidth, no need to prevent parallel dispersal.
+		c.accountantLock.Unlock()
+	} else {
+		// This request is using on-demand bandwidth, current implementation requires sequential dispersal.
+		defer c.accountantLock.Unlock()
```
litt3, line 180: So... what happens if someone runs two clients at the same time?

common/stage_timer.go
```diff
@@ -0,0 +1,100 @@
...
+
+// It may be helpful to generalize this utility to be used in other parts of the codebase. For a future PR, perhaps.
+
+// StageTimer encapsulates metrics to help track the time spent in each stage of the payload dispersal process.
+type StageTimer struct {
+	stageCount   *prometheus.GaugeVec
+	stageLatency *prometheus.SummaryVec
+}
+
+// SequenceProbe tracks the timing of a single sequence of operations. Multiple sequences can be tracked concurrently.
```
litt3, line 18: > can be tracked concurrently

Presumably with this you mean "can be tracked concurrently *by the same StageTimer*"?

common/stage_timer.go
```diff
@@ -0,0 +1,100 @@
...
+		prometheus.SummaryOpts{
+			Namespace:  prefix,
+			Name:       name + "_stage_latency_ms",
+			Help:       "the latency of each type of operation",
+			Objectives: map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001},
+		},
+		[]string{"stage"},
+	)
+
+	statusCount := promauto.With(registry).NewGaugeVec(
```
litt3, line 41: mixed terminology, sometimes "statusCount", sometimes "stageCount". Is there any reason that both "stage" and "status" need to exist in this file, it seems both refer to the same concept

common/stage_timer.go
```diff
@@ -0,0 +1,100 @@
...
+func (s *StageTimer) NewSequence(initialStatus string) *SequenceProbe {
+	if s == nil {
+		return nil
+	}
+
+	s.stageCount.WithLabelValues(initialStatus).Inc()
+	return &SequenceProbe{
+		stageTimer:        s,
+		currentStage:      initialStatus,
+		currentStageStart: time.Now(),
```
litt3, line 66: any reason not to store a time source in the stage timer, for easier testing?

feel free to ignore if this should be part of future generalization effort

disperser/common/v2/blobstore/dynamo_metadata_store.go
```diff
@@ -1596,7 +1596,7 @@ func UnmarshalOperatorID(item commondynamodb.Item) (*core.OperatorID, error) {
 	operatorIDStr := obj.OperatorID
 	if strings.HasPrefix(operatorIDStr, dispersalRequestSKPrefix) {
 		operatorIDStr = strings.TrimPrefix(operatorIDStr, dispersalRequestSKPrefix)
-	} else if strings.HasPrefix(operatorIDStr, dispersalResponseSKPrefix) {
+	} else {
```
litt3, line 1599: did you intend to include this change in this PR?

api/clients/v2/payloaddispersal/payload_disperser.go
cody-littley:
```
	// if nil, then no metrics will be collected
	registry *prometheus.Registry,
```

api/clients/v2/disperser_client.go
cody-littley: 🤷 

Not any worse than it was before, I guess.

common/stage_timer.go
cody-littley: Yes. Updated.

```
// SequenceProbe can be used to track the amount of time that a particular operation spends doing particular 
// sub-operations (i.e. stages). Multiple instances of a particular operation can be tracked concurrently by the same 
// StageTimer. For each operation, the StageTimer builds a SequenceProbe. Each SequenceProbe the lifecycle of a single 
// iteration of an operation.
```

common/stage_timer.go
cody-littley: `status` -> `stage`

common/stage_timer.go
cody-littley: This is necessary to support measurement of concurrent operations. Each operation needs its own start time, so we can't store state inside the parent `StageTimer`.

disperser/common/v2/blobstore/dynamo_metadata_store.go
cody-littley: Yes. lint is failing locally unless I make this change, I'm not sure why my local lint is not doing the same thing as the one in the CI pipeline. But it seems like a harmless change to make.

api/clients/v2/disperser_client.go
```diff
@@ -128,42 +145,67 @@ func (c *disperserClient) DisperseBlob(
...
+	c.accountantLock.Lock()
+
+	probe.SetStage("prepare_for_dispersal")
+
+	err := c.initOncePopulateAccountant(ctx)
 	if err != nil {
 		return nil, [32]byte{}, api.NewErrorFailover(err)
 	}
-	err = c.initOncePopulateAccountant(ctx)
+	err = c.initOnceGrpcConnection()
```
litt3, line 182: Why are you making this `initOnceGrpcConnection` call while you have the accountant lock? It seems like you'd want to do this outside the lock.

Side note, `initOnceGrpcConnection` is also called within `initOncePopulateAccountant`, so the call in its current place isn't *actually* doing anything, and would need to be placed prior to acquiring the lock if we want the grpc connection initialized outside the lock.

api/clients/v2/disperser_client.go
```diff
@@ -128,42 +145,67 @@ func (c *disperserClient) DisperseBlob(
...
-	for _, q := range quorums {
-		if q > corev2.MaxQuorumID {
-			return nil, [32]byte{}, api.NewErrorInvalidArg("quorum number must be less than 256")
-		}
+	if payment.CumulativePayment == nil || payment.CumulativePayment.Sign() == 0 {
+		// This request is using reserved bandwidth, no need to prevent parallel dispersal.
+		c.accountantLock.Unlock()
+	} else {
+		// This request is using on-demand bandwidth, current implementation requires sequential dispersal.
+		defer c.accountantLock.Unlock()
```
litt3, line 199: Just a thought- if this ends up presenting a performance issue, I think some easy optimization could be done to minimize lock time: some of the payload/blob/commitment work is quite expensive, and could be done prior to acquiring the accountant lock

common/stage_timer.go
```diff
@@ -0,0 +1,101 @@
...
+
+// StageTimer encapsulates metrics to help track the time spent in each stage of the payload dispersal process.
+type StageTimer struct {
+	stageCount   *prometheus.GaugeVec
+	stageLatency *prometheus.SummaryVec
+}
+
+// SequenceProbe can be used to track the amount of time that a particular operation spends doing particular
+// sub-operations (i.e. stages). Multiple instances of a particular operation can be tracked concurrently by the same
+// StageTimer. For each operation, the StageTimer builds a SequenceProbe. Each SequenceProbe the lifecycle of a single
```
litt3, line 18: > Each SequenceProbe the lifecycle of a single iteration of an operation.

sentence is missing a word

api/clients/v2/disperser_client.go
cody-littley: Moved above accountant lock.

api/clients/v2/disperser_client.go
cody-littley: Good point. In the initial metrics I've collected, the time spent in this part of the code is very small. But if it ever starts to become an issue, we can circle back on it.

common/stage_timer.go
cody-littley: fixed

api/clients/v2/disperser_client.go
litt3: Do any of your tests use on-demand payments?

cody-littley: no, I'm unsure how to configure on-demand payments

litt3: Resolving for now

PR #1443, opened 2025-04-02
Check signing percentage range during aggregation

ian-shim:
Make sure signing rate is within [0, 100]

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/aggregation.go
+++ b/core/aggregation.go
@@ -19,10 +19,11 @@ import (
 const maxNumOperatorAddresses = 300
 
 var (
-	ErrPubKeysNotEqual     = errors.New("public keys are not equal")
-	ErrInsufficientEthSigs = errors.New("insufficient eth signatures")
-	ErrAggPubKeyNotValid   = errors.New("aggregated public key is not valid")
-	ErrAggSigNotValid      = errors.New("aggregated signature is not valid")
+	ErrPubKeysNotEqual         = errors.New("public keys are not equal")
+	ErrInsufficientEthSigs     = errors.New("insufficient eth signatures")
+	ErrAggPubKeyNotValid       = errors.New("aggregated public key is not valid")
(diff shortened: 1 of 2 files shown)


core/aggregation.go
```diff
@@ -361,16 +367,18 @@ func GetStakeThreshold(state *OperatorState, quorum QuorumID, quorumThreshold ui
...
+		return 0, nil
 	}
 
 	stakeAmount = stakeAmount.Mul(stakeAmount, new(big.Int).SetUint64(percentMultiplier))
 	quorumThresholdBig := stakeAmount.Div(stakeAmount, totalStake)
 
 	quorumThreshold := uint8(quorumThresholdBig.Uint64())
-
-	return quorumThreshold
+	if quorumThreshold > 100 {
```
jianoaix, line 380: why this needs to check? It's just computed locally and it has fully context about the calculation, which the caller may not know even if the error is returned

core/aggregation.go
```diff
@@ -361,16 +367,18 @@ func GetStakeThreshold(state *OperatorState, quorum QuorumID, quorumThreshold ui
...
 	totalStake := state.Totals[quorum].Stake
 	if totalStake.Cmp(big.NewInt(0)) == 0 {
-		return 0
+		return 0, nil
 	}
 
 	stakeAmount = stakeAmount.Mul(stakeAmount, new(big.Int).SetUint64(percentMultiplier))
 	quorumThresholdBig := stakeAmount.Div(stakeAmount, totalStake)
 
 	quorumThreshold := uint8(quorumThresholdBig.Uint64())
```
jianoaix, line 379: this doesn't look like quorum threshold (which is a constant) and match function name (where signed percentage is accurate)

cody-littley (approved):
LGTM once Jian's comments are addressed.

PR #1452, opened 2025-04-07
feat: reservation update in payment vault

hopeyen:

https://www.notion.so/eigen-labs/Protocol-Design-Confined-Reservations-updates-1b913c11c3e080fe9cf8e250b97a4465

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/contracts/src/interfaces/IPaymentVault.sol
+++ b/contracts/src/interfaces/IPaymentVault.sol
@@ -7,8 +7,7 @@ interface IPaymentVault {
         uint64 symbolsPerSecond; // Number of symbols reserved per second
         uint64 startTimestamp;   // timestamp of epoch where reservation begins
         uint64 endTimestamp;     // timestamp of epoch where reservation ends
-        bytes quorumNumbers;     // quorum numbers in an ordered bytes array
-        bytes quorumSplits;      // quorum splits in a bytes array that correspond to the quorum numbers
+        uint64 quorumNumber;     // quorum number for the reservation
     }
 
     struct OnDemandPayment {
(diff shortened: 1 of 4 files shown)


contracts/src/interfaces/IPaymentVault.sol
```diff
@@ -7,16 +7,21 @@ interface IPaymentVault {
         uint64 symbolsPerSecond; // Number of symbols reserved per second
         uint64 startTimestamp;   // timestamp of epoch where reservation begins
         uint64 endTimestamp;     // timestamp of epoch where reservation ends
-        bytes quorumNumbers;     // quorum numbers in an ordered bytes array
-        bytes quorumSplits;      // quorum splits in a bytes array that correspond to the quorum numbers
+        uint64 quorumNumber;     // quorum number for the reservation
```
pakim249CAL, line 10: Just note that we can't upgrade the vault like this, but if redeploying is okay then this is fine.

pakim249CAL: Sorry, I guess I should elaborate more on the options here. First, I'll explain the problem and then some potential solutions.

Because reservations are stored for each reservation in contract storage in a mapping, all contract storage rules apply. e.g. it is generally safe to "add" variables, but you cannot delete a variable and put a new one there without a storage conflict.

So one potential solution is to just start including quorumNumber, and no longer use the bytes versions by just not using it in the code, though it must stay in the struct to maintain storage alignment. This implies very unpleasant UX though since some members of the struct are essentially deprecated, with all the confusion that may cause. And that is not even to address the question of how existing reservations would be used by the new code.

Another potential solution is to transform all existing reservations to the new data schema when the contract upgrade happens. This would only be possible if there are very few reservations, and is not exactly the most kosher way of doing things in general.

I think a simpler and cleaner solution for this is simply to redeploy if this is possible. But the alternative, messy solutions do exist.

hopeyen:
closing as Patrick started on the new design

PR #1453, opened 2025-04-07
feat: disperser and relay registry updates

hopeyen:

https://www.notion.so/eigen-labs/TDD-Disperser-Relay-registry-update-1a613c11c3e0802db1c2ca00a9924fa9

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/contracts/src/core/EigenDADisperserRegistry.sol
+++ b/contracts/src/core/EigenDADisperserRegistry.sol
@@ -87,4 +87,12 @@ contract EigenDADisperserRegistry is OwnableUpgradeable, EigenDADisperserRegistr
     function disperserKeyToAddress(uint32 _key) external view returns (address) {
         return disperserKeyToInfo[_key].disperserAddress;
     }
+    
+    function disperserKeysToAddresses(uint32[] memory _keys) external view returns (address[] memory) {
+        address[] memory addresses = new address[](_keys.length);
+        for (uint32 i = 0; i < _keys.length; i++) {
+            addresses[i] = disperserKeyToInfo[_keys[i]].disperserAddress;
+        }
+        return addresses;
+    }
 }
--- a/contracts/src/core/EigenDADisperserRegistry.sol
+++ b/contracts/src/core/EigenDADisperserRegistry.sol
@@ -9,6 +9,9 @@ import "../interfaces/IEigenDAStructs.sol";
 /**
  * @title Registry for EigenDA disperser info
  * @author Layr Labs, Inc.
+ * @notice Permissionless registration with a deposit on registration and each update.
+ *         Deregistration refunds the accumulated deposit minus a flat percentage tax.
+ *         All users, including the owner, must pay deposits and taxes.
  */
 contract EigenDADisperserRegistry is OwnableUpgradeable, EigenDADisperserRegistryStorage, IEigenDADisperserRegistry {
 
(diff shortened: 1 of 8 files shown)


contracts/src/core/EigenDADisperserRegistry.sol
```diff
@@ -22,11 +25,65 @@ contract EigenDADisperserRegistry is OwnableUpgradeable, EigenDADisperserRegistr
...
+            
+            // All users must pay additional deposit for updates
+            require(msg.value == REGISTRATION_DEPOSIT, "Additional deposit of 0.1 ether required for update");
+            disperserDeposits[_disperserKey] += msg.value;
+        }
+        
+        // Set disperser info
         disperserKeyToInfo[_disperserKey] = _disperserInfo;
+        
+        // If the sender is the owner, mark this disperser as owner-created
```
pakim249CAL, line 54: What is the point of this marking? Is it used anywhere? If you want to give extra privileges offchain to an owner owned disperser key, maybe a better design would be to have an owner function that can mark keys as privileged.

contracts/src/core/EigenDADisperserRegistry.sol
```diff
@@ -22,11 +25,65 @@ contract EigenDADisperserRegistry is OwnableUpgradeable, EigenDADisperserRegistr
...
+        uint256 refund = totalDeposit - tax;
+
+        // Clear registration data
+        delete disperserRegistrants[_disperserKey];
+        delete disperserDeposits[_disperserKey];
+        delete disperserKeyToInfo[_disperserKey];
+        delete isOwnerCreatedDisperser[_disperserKey];
+
+        emit DisperserRemoved(_disperserKey, msg.sender);
+        payable(msg.sender).transfer(refund);
```
pakim249CAL, line 84: Do you not want to keep a running total of the tax accrued so that the owner can transfer these funds at some point?

contracts/src/core/EigenDADisperserRegistry.sol
hopeyen: Yeah, the intend is to give extra privileges offchain. Your approach makes sense to me

contracts/src/core/EigenDADisperserRegistry.sol
hopeyen: I assumed that there's balance tracked per contract automatically. I think the funds should be kept in the contract as long as the disperser is registered, and when they deregister they get 1-tax_percentage back. For the taxed funds, I think we can straight-up burn it or send it to rewards

pakim249CAL: I'm not sure you only want to keep track of the balance in the contract, as you need to be able to differentiate between funds that are taxes, and funds that need to be kept around as a refund for each registered disperser.

hopeyen: I see, this makes sense to me. I think we should keep the registration amount as long as the disperser is registered (i.e., owner shouldn't transfer funds out, and disperser "stake" is not tax until they deregister. Perhaps we can use `fee` instead of `tax` as `tax` might be more sensitive and misleading.) The owner can keep a running total of `fee` accured in storage whenever someone deregister, and can transfer the funds out

hopeyen:
closing as Patrick started on the new design

PR #1458, opened 2025-04-09
docs: Add SRS docs

litt3:
- Adds SRS files in an easy-to-find place
- Creates a script for download SRS files
- Creates README with SRS file basics


resources/srs/README.md
```diff
@@ -0,0 +1,80 @@
+# Structured Reference String (SRS) Files
+
+This directory contains the Structured Reference String (SRS) files required for KZG commitments and proofs in EigenDA.
+
+## File Information
+
+| File Name          | Size   | Number of Points | Point Size | SHA256 Hash                                                      |
+|--------------------|--------|------------------|------------|------------------------------------------------------------------|
+| g1.point           | 16 MB  | 524,288          | 32 bytes   | 8f18b9c04ed4bddcdb73001fb693703197328cecabdfa9025f647410b0c50d7f |
+| g2.point           | 32 MB  | 524,288          | 64 bytes   | a6942684aa751b4ec7873e2edb4660ac5c4516adb3b310441802cc0d489f645a |
```
anupsv, line 10: Can we add the difference between g2point and the trailing points as well.

anupsv (approved):

resources/srs/README.md
litt3: To clarify, do you mean add a description of how they differ?

tools/download-srs.sh
```diff
@@ -0,0 +1,225 @@
...
+        fi
+    done
+    
+    # Generate the hash file
+    {
+        echo "# SRS files hashes for blob size ${size_bytes} bytes"
+        echo "# Generated on $(date)"
+        echo "# Format: SHA256 (filename)"
+        echo ""
+        sha256sum "${output_dir}/g1.point" | sed "s|${output_dir}/||"
```
dmanc, line 154: Should we store the sha value in the repo?

dmanc (approved):

tools/download-srs.sh
litt3: It's included in the README for the "standard" file size.

Are you saying you think it _shouldn't_ be there, or just didn't see that it was?

dmanc: I was thinking something like this https://github.com/Layr-Labs/eigenda-operator-setup/blob/v0.9.0-rc.1/resources/srssha256sums.txt

dmanc: I see I missed that it was in the README.md

litt3: I'll commit the sha file in addition to leaving it in the readme 👍

resources/srs/README.md
litt3: @anupsv

tools/download-srs.sh
```diff
@@ -0,0 +1,225 @@
...
+    local missing_deps=0
+    for cmd in curl sha256sum awk grep tr; do
+        if ! command -v "$cmd" &> /dev/null; then
+            echo "Error: Required command '$cmd' not found"
+            missing_deps=1
+        fi
+    done
+    
+    if [ $missing_deps -eq 1 ]; then
+        echo "Please install missing dependencies and try again."
```
samlaf, line 23: I would error inside the for loop and print the missing dependency instead.

tools/download-srs.sh
samlaf, line 1: any reason why this is written in bash and not in golang? This is a pretty long bash script, has external dependencies, most people in the team are much more comfortable with golang than bash, etc.

Also more generally I feel more and more than we should have an `eigenda` CLI in golang that includes a bunch of tooling, like this (download srs), doing the stuff kzgpad (which is super badly named) is doing, etc.

tools/download-srs.sh
```diff
@@ -0,0 +1,225 @@
...
+    echo "Verification hashes:"
+    cat "${hash_file}"
+    
+    return 0
+}
+
+# Validate command line arguments
+function validate_arguments() {
+    if [ $# -ne 1 ]; then
+        echo "Usage: $0 <size_in_bytes>"
```
samlaf, line 174:
```suggestion
        echo "Usage: $0 <blob_size_in_bytes>"
```

samlaf: was unclear to me if the size was referring to G1 or G2 file. Explicitly saying its blob makes this more obvious what this file is doing imo

resources/srs/README.md
```diff
@@ -0,0 +1,80 @@
...
+## Security Considerations
+
+Using the correct SRS files is essential for the proper functioning of any software interacting with EigenDA. 
+If a user has incorrect or tampered SRS files, the following would occur:
+
+1. **Verification failures**: The user would be unable to successfully verify KZG commitments and proofs, making it 
+   impossible to validate blob data from the network.
+
+2. **Network incompatibility**: A node using incorrect SRS files would be unable to meaningfully interact with the 
+   EigenDA network, as it would consistently fail to verify certificates from honest nodes.
```
samlaf, line 72: eigenda nodes don't verify certs right afaiu? what "node" are you referring to?

resources/srs/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+3. **Self-isolation**: Rather than creating a security vulnerability, having incorrect SRS files simply results in 
+   self-isolation from the network's consensus.
+
+It's important to understand that this isn't a security concern for the broader network. If a disperser attempted 
+to use incorrect SRS files, honest nodes would immediately detect this during certificate verification. The network's 
+security properties rely on the fact that honest nodes using the correct SRS can detect and reject improperly generated 
+certificates.
```
samlaf, line 80: oh I see the cert you are talking about here is the `blobCertificate` right? When I see `certificates` I immediately think of `DACerts`. 

Also this is v1 focused I think, where disperser is generating the commitment? For v2 its the proxy that generates it (and its in the `blobHeader`, not `blobCertificate`. So prob better to take a broader perspective and not refer to cert here?

tools/download-srs.sh
samlaf: Just checked and all our other tools/ are in golang as well.

resources/srs/README.md
```diff
@@ -0,0 +1,80 @@
...
+
+### Using the Download Script
+
+The EigenDA repository provides a convenience script located at `tools/download-srs.sh` for downloading and 
+generating SRS file hashes.
+
+Usage:
+```bash
+# Download SRS files for a specified blob size (in bytes)
+./tools/download-srs.sh <size_in_bytes>
```
samlaf, line 36:
```suggestion
curl -O https://raw.githubusercontent.com/Layr-Labs/eigenda/master/tools/download-srs.sh
chmod +x download-srs.sh
# Download SRS files for a specified blob size (in bytes)
./download-srs.sh <size_in_bytes>
```

samlaf: think its better if we don't force users to clone the repo

resources/srs/README.md
```diff
@@ -0,0 +1,80 @@
...
+1. Download the appropriate portions of g1.point and g2.point files
+2. Calculate the correct byte range for g2.trailing.point
+3. Save all files to a "srs-files" directory
+4. Generate SHA256 hashes for verification
+5. Create a hash file with labeled hashes for each downloaded file
+
+## SRS Verification and Alternative Retrieval Method
+
+For users who need to verify the integrity of SRS files, please refer to the
+[SRS Utilities README](/tools/srs-utils/README.md) for detailed instructions. This tool provides:
```
samlaf, line 52: didn't know we already have this golang cli. Shouldnt this download-srs utility have been added to that golang CLI instead? Why have 2 separate tools

resources/srs/README.md
samlaf: I suggested the edit. You can just curl the script itself.

litt3: Makes sense, but unfortunately won't be possible anymore, as I'm refactoring the tool to be written in go

tools/download-srs.sh
litt3: No longer relevant, after migration to go cli tool, instead of bash

tools/download-srs.sh
litt3: Rewritten in golang [90cf5c4](https://github.com/Layr-Labs/eigenda/pull/1458/commits/90cf5c4d9f024aff7eca514bb8c2b09211928353)

tools/download-srs.sh
litt3: Golang implementation has clearer names, so that it's obvious we are talking about blob size

resources/srs/README.md
litt3: I was using "node" in a more general sense, which is confusing. 

I've rewritten this section of the doc in [b0bc39e7](https://github.com/Layr-Labs/eigenda/pull/1458/commits/b0bc39e7da8e6a87e3c783eb2698f8e0f828a54c)

resources/srs/README.md
litt3: I've reworded to section, to try and make it more clear [b0bc39e7](https://github.com/Layr-Labs/eigenda/pull/1458/commits/b0bc39e7da8e6a87e3c783eb2698f8e0f828a54c)

resources/srs/README.md
litt3: Rewritten in go, and added to the existing SRS cli

resources/srs/README.md
samlaf: In that case you just use `go install ...` which is even cleaner. It does require golang being installed, but I think its fair to assume most people have golang installed...? :D

One reason why having a single CLI for all our tooling would be nice is that we could dockerize it and push it to ghcr for people to use without any dependencies

tools/srs-utils/README.md
```diff
@@ -1,41 +1,102 @@
...
 `go run main.go help`
 
-###  How to parse G1, G2 points from the challenge file.
+### Downloading SRS Files
+
+The simplest way to get the required SRS files is to download the pre-processed files directly from the EigenDA
+S3 bucket:
+
+```bash
+go run main.go download --blob-size-bytes 16777216
```
samlaf, line 39: I prefer `go install github.com/Layr-Labs/eigenda/tools/srs-utils@latest` and then just run `srs-utils <ARGS>`.
I don't like when instructions force me to clone the repo to run a little CLI binary.

samlaf (approved):

PR #1460, opened 2025-04-09
Batch timeout

cody-littley:

Add a global timeout to the attestation collection phase of v1 batches.

### DIFF
--- a/core/aggregation.go
+++ b/core/aggregation.go
@@ -42,8 +42,8 @@ type QuorumAttestation struct {
 	QuorumAggPubKey map[QuorumID]*G1Point
 	// SignersAggPubKey is the aggregated public key for all of the operators that signed the message by each quorum
 	SignersAggPubKey map[QuorumID]*G2Point
-	// AggSignature is the aggregated signature for all of the operators that signed the message for each quorum, mirroring the
-	// SignersAggPubKey.
+	// AggSignature is the aggregated signature for all of the operators that signed the message for each quorum, 
+	// mirroring the SignersAggPubKey.
 	AggSignature map[QuorumID]*Signature
 	// QuorumResults contains the quorum ID and the amount signed for each quorum
 	QuorumResults map[QuorumID]*QuorumResult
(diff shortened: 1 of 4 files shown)


core/aggregation.go
```diff
@@ -132,7 +157,19 @@ func (a *StdSignatureAggregator) ReceiveSignatures(ctx context.Context, state *I
...
+
+		select {
+		case r = <-messageChan:
+		case <-attestationCtx.Done():
+			remainingReplies := numOperators - numReply
+			a.Logger.Warnf(
+				"batch attestation time exceeded, no further signatures will be accepted for batch %x. "+
+					"There are %d uncollected signatures.", message, remainingReplies)
+			break
+		}
```
cody-littley, line 171: Note to the reviewers: this is the primary change in this PR. Most of the rest of the diff is fixing line wrapping in these files (I find code harder to read with really long lines).

disperser/cmd/batcher/main.go
```diff
@@ -84,6 +84,10 @@ func RunBatcher(ctx *cli.Context) error {
 		return err
 	}
 
+	// TODO remove this log statement
```
dmanc, line 87: Maybe we could leave it as trace or debug?

core/aggregation.go
```diff
@@ -151,7 +157,23 @@ func (a *StdSignatureAggregator) ReceiveSignatures(
...
-		r := <-messageChan
+
+		var r SigningMessage
+
+		var contextExpired bool
+		select {
+		case r = <-messageChan:
+		case <-attestationCtx.Done():
+			remainingReplies := numOperators - numReply
+			a.Logger.Errorf(
```
dmanc, line 168: Should it be an error or would warning make more sense?

dmanc (approved):

disperser/cmd/batcher/main.go
cody-littley: Set to debug level.

core/aggregation.go
cody-littley: changed to a warning

PR #1472, opened 2025-04-14
refactor!: Refactor Cert Verifier

pakim249CAL:

This is a proposed CertVerifier refactor to accommodate as many use cases as possible, improve extensibility, and maintain a reasonable amount of backwards compatibility. 

### Breaking
Note that this new version of the CertVerifier does not bother complying with the IThresholdRegistry interface. The threshold registry already does that, and the cert verifier already contains a getter for the threshold registry. I currently estimate the impact of this to be low/none, but LMK if you think otherwise.

@ethenotethan expressed that the RBN should be at a known position in calldata. This PR reorders the BatchHeader struct. This is a breaking change for current integrators.

### Overall refactor design

See below comment for a description about the overall refactor and why it might be needed.

### Update forge-std

Also updates forge-std to use a newer partialRevert feature. This is used to replicate the current testing setup where only a static string error was tested for, but I set up errors to return parameters now. I would like to rewrite all the tests at some point, so I'll leave that for another PR. No testing granularity is lost in this PR.

### Move security thresholds check

The security thresholds check is moved to be in the CertVerifier constructor, and this check is removed from the overall verification flow. There is no reason to do this check if the security thresholds are set as immutable on contract construction.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests
   - [ ]


pakim249CAL:
_Generated by Claude AI, reviewed and edited by me._

# EigenDA Certificate Verification Refactoring

## Overview

This document details the refactoring of EigenDA's certificate verification system. The refactoring separates verification functionality into distinct V1 and V2 implementations while maintaining backward compatibility. The new design improves code quality, maintainability, and introduces a cleaner error handling system.

## Key Changes

### 1. Separation of Concerns

The original implementation mixed V1 and V2 verification logic in a single library and contract. The refactoring:

- Creates separate libraries (`EigenDACertVerificationV1Lib` and `EigenDACertVerificationV2Lib`)
- Creates separate contract implementations (`EigenDACertVerifierV1` and `EigenDACertVerifierV2`)
- Provides a unified contract (`EigenDACertVerifier`) that inherits from both implementations

### 2. Error Handling Improvements

The original implementation used generic `require` statements with string error messages. The refactoring:

- Introduces specific error types with meaningful parameters
- Implements an error code system with detailed error parameters
- Provides clear, gas-efficient error reporting
- Uses a two-step validation pattern (validation + error handling)

### 3. Code Modularity

The original implementation had large monolithic functions. The refactoring:

- Breaks down verification into smaller, focused functions
- Creates a clear flow of validation steps
- Provides better function naming for clarity

### 4. Performance Optimizations

The refactoring includes several performance improvements:

- Reduced gas costs through optimized error handling
- Better memory management by limiting duplicate parameter passing
- Structured error codes for efficient gas usage
- Direct calldata access where possible

### 5. Enhanced Security

The refactoring improves security through:

- More explicit validation steps
- Clearer error messages for debugging
- Consistent security parameter validation
- Separation of security-critical code paths

## Detailed Changes

### Libraries

### EigenDACertVerificationV1Lib

- Extracted from original `EigenDACertVerificationUtils` library
- Focused solely on V1 certificate verification
- Introduces proper error types and error code system
- Breaks down verification into clear steps

### EigenDACertVerificationV2Lib

- Extracted from original `EigenDACertVerificationUtils` library
- Dedicated to V2 certificate verification
- Maintains consistent error handling pattern with V1
- Improves verification flow and readability

### Contracts

### EigenDACertVerifierV1

- Implements V1-specific verification
- Uses the new V1 library
- Virtual methods for extensibility

### EigenDACertVerifierV2

- Implements V2-specific verification
- Uses the new V2 library
- Virtual methods for extensibility

### EigenDACertVerifier

- Inherits from both V1 and V2 verifiers
- Maintains identical interface to the original contract
- Ensures backward compatibility
- Properly overrides functions to prevent ambiguity

## Error Handling System

One of the most significant improvements is the new error handling system:

1. Each library defines specific error types with meaningful parameters
2. Functions return error codes and parameters rather than immediately reverting
3. A separate error handler function interprets error codes and reverts with the appropriate error

Benefits of this approach:

- Better gas efficiency (avoid string encoding in require statements)
- More informative error messages with context
- Ability to handle errors without reverting (useful for ZK proofs)
- Consistent error handling across the system

## Technical Considerations

### Gas Efficiency

The refactoring may slightly increase deployment costs due to additional code but should reduce runtime gas costs for verification through:

- More efficient error handling
- Better organization of validation logic
- Eliminating redundant checks

### Contract Size

The separation increases the combined contract size but improves maintainability. If contract size becomes an issue, consider:

- Deploying V1 and V2 verifiers as separate contracts
- Using delegation patterns for verification
- Further optimizing library code

### Future Extensions

The new architecture makes it easier to:

- Add V3 verification in the future
- Modify individual verification paths
- Introduce new security parameters
- Add specialized verification methods

contracts/test/unit/EigenDACertVerifierV2Unit.t.sol
```diff
@@ -130,23 +132,10 @@ contract EigenDACertVerifierV2Unit is MockEigenDADeployer {
         eigenDACertVerifier.verifyDACertV2FromSignedBatch(signedBatch, blobInclusionInfo);
     }
 
-    function test_verifyDACertV2_revert_BadSecurityParams(uint256 pseudoRandomNumber) public {
```
pakim249CAL, line 133: This is not necessary since this check is done on CertVerifier construction now.

contracts/src/periphery/EigenDACertVerifierV2.sol
```diff
@@ -0,0 +1,260 @@
...
+    /// @notice The EigenDA middleware OperatorStateRetriever contract address
+    OperatorStateRetriever public immutable operatorStateRetrieverV2;
+
+    /// @notice The EigenDA middleware RegistryCoordinator contract address
+    IRegistryCoordinator public immutable registryCoordinatorV2;
+
+    /// @notice Security thresholds for V2 certificate verification
+    /// Storing individual fields as immutable instead of the struct
+    uint8 public immutable securityThresholdsConfirmation;
+    uint8 public immutable securityThresholdsAdversary;
```
samlaf, line 39: prob good to add small description of what these are and how they are used

contracts/src/periphery/EigenDACertVerifierV2.sol
```diff
@@ -0,0 +1,260 @@
...
+    /**
+     * @notice Thin try/catch wrapper around verifyDACertV2 that returns false instead of reverting
+     * @dev Useful for ZK proof systems that require a return value
+     * @param batchHeader The batch header of the blob
+     * @param blobInclusionInfo The inclusion proof for the blob cert
+     * @param nonSignerStakesAndSignature The nonSignerStakesAndSignature to verify the blob cert against
+     * @param signedQuorumNumbers The signed quorum numbers corresponding to the nonSignerStakesAndSignature
+     * @return success True if verification succeeded
+     */
+    function verifyDACertV2ForZKProof(
```
samlaf, line 131: don't like this ZKProof suffix (also doesnt match V1 which is ZkProof). See my comment in v1 contract

contracts/src/periphery/EigenDACertVerifierV1_V2.sol
```diff
@@ -0,0 +1,88 @@
...
+     * @dev Overrides both V1 and V2 implementations to use a single registry for consistency
+     */
+    function _thresholdRegistry()
+        internal
+        view
+        override(EigenDACertVerifierV1, EigenDACertVerifierV2)
+        returns (IEigenDAThresholdRegistry)
+    {
+        // This contract enforces that V1 and V2 use the same registry on construction. So we just choose V2.
+        return eigenDAThresholdRegistryV2;
```
samlaf, line 72: is inheritance really the right pattern here? Seems kind of bug prone to me with related asserts/checks in different non-local places like this.

samlaf:
Hmm seems like a duplicate PR of https://github.com/Layr-Labs/eigenda/pull/1477 ?

PR #1475, opened 2025-04-16
feat: network signing rate

jianoaix:
Networking signing rate is crucial information about the EigenDA, as it indicates the liveness of the network.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests


disperser/dataapi/prometheus_client.go
```diff
@@ -12,6 +12,9 @@ import (
 const (
 	// maxNumOfDataPoints is the maximum number of data points that can be queried from Prometheus based on latency that this API can provide
 	maxNumOfDataPoints = 3500
+
+	// Calculate the average over this number of minutes for signing rate
+	siningRateRangeVectorMinutes = 9
```
dmanc, line 17: typo 
```suggestion
	signingRateRangeVectorMinutes = 9
```

dmanc, line 17: Why the number 9?

dmanc (approved):

disperser/dataapi/prometheus_client.go
jianoaix: Done.

jianoaix: it's just copied the existing throughput time series query, but I'm not sure what was the calculation behind it:) it looks arbitrary to me

for signing rate, we may probably want to be a bit more responsive, so we could try 5m

PR #1477, opened 2025-04-17
refactor!: cert verifier 

pakim249CAL:
To reviewers: review through individual commits! Though the commits are largely my work, I had AI organize the commits to separate concerns. You can review changes to respective codebases individually.

This is a proposed CertVerifier refactor to accommodate as many use cases as possible, improve extensibility, and maintain a reasonable amount of backwards compatibility. 

### Breaking
Note that this new version of the CertVerifier does not bother complying with the IThresholdRegistry interface. The threshold registry already does that, and the cert verifier already contains a getter for the threshold registry. I currently estimate the impact of this to be low/none, but LMK if you think otherwise.

### Overall refactor design

See below comment for a description about the overall refactor and why it might be needed.

### Update forge-std

Also updates forge-std to use a newer partialRevert feature. This is used to replicate the current testing setup where only a static string error was tested for, but I set up errors to return parameters now. I would like to rewrite all the tests at some point, so I'll leave that for another PR. No testing granularity is lost in this PR.

### Move security thresholds check

The security thresholds check is moved to be in the CertVerifier constructor, and this check is removed from the overall verification flow. There is no reason to do this check if the security thresholds are set as immutable on contract construction.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests
   - [ ]


pakim249CAL:
_Generated by Claude AI, reviewed and edited by me._

# EigenDA Certificate Verification Refactoring

## Overview

This document details the refactoring of EigenDA's certificate verification system. The refactoring separates verification functionality into distinct V1 and V2 implementations while maintaining backward compatibility. The new design improves code quality, maintainability, and introduces a cleaner error handling system.

## Key Changes

### 1. Separation of Concerns

The original implementation mixed V1 and V2 verification logic in a single library and contract. The refactoring:

- Creates separate libraries (`EigenDACertVerificationV1Lib` and `EigenDACertVerificationV2Lib`)
- Creates separate contract implementations (`EigenDACertVerifierV1` and `EigenDACertVerifierV2`)
- Provides a unified contract (`EigenDACertVerifier`) that inherits from both implementations

### 2. Error Handling Improvements

The original implementation used generic `require` statements with string error messages. The refactoring:

- Introduces specific error types with meaningful parameters
- Implements an error code system with detailed error parameters
- Provides clear, gas-efficient error reporting
- Uses a two-step validation pattern (validation + error handling)

### 3. Code Modularity

The original implementation had large monolithic functions. The refactoring:

- Breaks down verification into smaller, focused functions
- Creates a clear flow of validation steps
- Provides better function naming for clarity

### 4. Performance Optimizations

The refactoring includes several performance improvements:

- Reduced gas costs through optimized error handling
- Better memory management by limiting duplicate parameter passing
- Structured error codes for efficient gas usage
- Direct calldata access where possible

### 5. Enhanced Security

The refactoring improves security through:

- More explicit validation steps
- Clearer error messages for debugging
- Consistent security parameter validation
- Separation of security-critical code paths

## Detailed Changes

### Libraries

### EigenDACertVerificationV1Lib

- Extracted from original `EigenDACertVerificationUtils` library
- Focused solely on V1 certificate verification
- Introduces proper error types and error code system
- Breaks down verification into clear steps

### EigenDACertVerificationV2Lib

- Extracted from original `EigenDACertVerificationUtils` library
- Dedicated to V2 certificate verification
- Maintains consistent error handling pattern with V1
- Improves verification flow and readability

### Contracts

### EigenDACertVerifierV1

- Implements V1-specific verification
- Uses the new V1 library
- Virtual methods for extensibility

### EigenDACertVerifierV2

- Implements V2-specific verification
- Uses the new V2 library
- Virtual methods for extensibility

### EigenDACertVerifier

- Inherits from both V1 and V2 verifiers
- Maintains identical interface to the original contract
- Ensures backward compatibility
- Properly overrides functions to prevent ambiguity

## Error Handling System

One of the most significant improvements is the new error handling system:

1. Each library defines specific error types with meaningful parameters
2. Functions return error codes and parameters rather than immediately reverting
3. A separate error handler function interprets error codes and reverts with the appropriate error

Benefits of this approach:

- Better gas efficiency (avoid string encoding in require statements)
- More informative error messages with context
- Ability to handle errors without reverting (useful for ZK proofs)
- Consistent error handling across the system

## Technical Considerations

### Gas Efficiency

The refactoring may slightly increase deployment costs due to additional code but should reduce runtime gas costs for verification through:

- More efficient error handling
- Better organization of validation logic
- Eliminating redundant checks

### Contract Size

The separation increases the combined contract size but improves maintainability. If contract size becomes an issue, consider:

- Deploying V1 and V2 verifiers as separate contracts
- Using delegation patterns for verification
- Further optimizing library code

### Future Extensions

The new architecture makes it easier to:

- Add V3 verification in the future
- Modify individual verification paths
- Introduce new security parameters
- Add specialized verification methods

contracts/compile.sh
```diff
@@ -39,7 +39,9 @@ contracts="PaymentVault \
   IEigenDAServiceManager \
   MockRollup \
   EjectionManager \
-  EigenDACertVerifier \
+  EigenDACertVerifierV1 \
+  EigenDACertVerifierV2 \
+  EigenDACertVerifierV1_V2 \
```
ethenotethan, line 44: will the decoupling of V1 cert verification from V2 require existing rollups leveraging the V1 cert verifier we've already deployed to deploy a new one?

contracts/src/periphery/EigenDACertVerifierV1.sol
```diff
@@ -0,0 +1,152 @@
...
+    /**
+     * @notice Verifies a blob cert and returns result without reverting
+     * @param blobHeader Pointer to the blob header in calldata
+     * @param blobVerificationProof Pointer to the blob cert verification proof in calldata
+     * @return success True if verification succeeded
+     */
+    function verifyDACertV1ForZkProof(
+        BlobHeader calldata blobHeader,
+        BlobVerificationProof calldata blobVerificationProof
+    ) external view returns (bool success) {
```
ethenotethan, line 87: technically we'd use this for fraud proofs too so the suffix `ZkProof` may need to be updated to something else

contracts/compile.sh
pakim249CAL: Nah, they can just use whatever already works for them. If they want to transition, they can deploy a new one at their leisure. The new cert verifiers should remain functionally the same, as this PR strives NOT to change verification logic, exposed function selectors for verification, and behaviors.

contracts/src/periphery/EigenDACertVerifierV1.sol
pakim249CAL: Agreed it's not a great name, but it is consistent with the naming for the V2 Cert verifier which is taken from master.

samlaf:
Ehhh what is this PR? I commented on https://github.com/Layr-Labs/eigenda/pull/1472. Are these the same?

contracts/src/periphery/EigenDACertVerifierV1.sol
samlaf: Also think we should change both. Made the same comment here https://github.com/Layr-Labs/eigenda/pull/1472#discussion_r2048232747

ethenotethan: iiuc the interface will have to change given that we plan to change the verification functions to also take in a structured EigenDA cert type?

pakim249CAL:
> Ehhh what is this PR? I commented on #1472. Are these the same?

Hey sorry, I guess I should explain 😅. This PR is a reorganization of the commits in #1472, with help from claude. I was paranoid about claude doing it so I made a new branch that I could diff check with #1472. I'll copy over and respond to your comments from that PR.

contracts/src/periphery/EigenDACertVerifierV1.sol
pakim249CAL: @ethenotethan I should mention that I discovered there are obstacles to reordering fields. The problem is that the batch header hash calculation depends on the field order in a batch, and so any reordering of these fields is a VERY breaking change. I think this should be done in another PR with a full discussion of implications.

contracts/script/GenerateUnitTestHashes.s.sol
```diff
@@ -14,7 +14,7 @@ contract GenerateHashes is Script {
     string deployConfigPath = "script/input/eigenda_deploy_config.json";
 
     // deploy all the EigenDA contracts. Relies on many EL contracts having already been deployed.
-    function run() external view {
+    function run() external pure {
```
pakim249CAL, line 17: from @samlaf  #1472 

comment is wrong. Also is this script useful? Looks like its just console.logging, which I'm guessing was used to manually create test cases for some other script..? There's def a better way to do this...

pakim249CAL: I only made this change to fix a compiler warning which blocks CI with the updated forge-std. I don't think fixing the comment would be in scope of this PR

contracts/src/periphery/EigenDACertVerifierV1.sol
```diff
@@ -0,0 +1,152 @@
...
+import {EigenDACertVerificationV1Lib as CertLib} from "src/periphery/libraries/EigenDACertVerificationV1Lib.sol";
+import "src/interfaces/IEigenDAStructs.sol";
+
+/**
+ * @title CertVerifierV1 - EigenDA V1 certificate verification
+ * @author Layr Labs, Inc.
+ * @notice Contains all V1-specific verification functionality
+ */
+contract EigenDACertVerifierV1 is IEigenDACertVerifierV1 {
+    /// @notice Thrown when there is a length mismatch
```
pakim249CAL, line 16: @samlaf #1472 

length mismatch between what and what? cert fields and some stuff stored onchain? or..

pakim249CAL: This error is currently only used in verifyDaCertsV1, since we need to verify that the length of blob headers and proofs is the same. I'll add this to the comment.

contracts/src/periphery/EigenDACertVerifierV1.sol
```diff
@@ -0,0 +1,152 @@
...
+    error LengthMismatch();
+
+    /// @notice The EigenDAThresholdRegistry contract address
+    IEigenDAThresholdRegistry public immutable eigenDAThresholdRegistryV1;
```
pakim249CAL, line 20: @samlaf #1472 

I personally prefer not to include comments like this that just restate the information already available in the variable name/type. wdyt?

I can agree with that. I'll cut out the obvious ones.

pakim249CAL: I can agree with that. I'll cut out the obvious ones.

contracts/src/periphery/EigenDACertVerifierV1.sol
```diff
@@ -0,0 +1,152 @@
...
+
+    /// @notice The EigenDABatchMetadataStorage contract address
+    /// @dev On L1 this contract is the EigenDA Service Manager contract
+    IEigenDABatchMetadataStorage public immutable eigenDABatchMetadataStorageV1;
```
pakim249CAL, line 24: @samlaf #1472 

what is it on not L1? We don't know yet I'm guessing?

pakim249CAL: I should probably remove that part, as there is nothing we don't know yet about how V1 will work, as it is going to be deprecated AFAICT

contracts/src/periphery/EigenDACertVerifierV1.sol
```diff
@@ -0,0 +1,152 @@
...
+    function _quorumNumbersRequired() internal view virtual returns (bytes memory) {
+        return _thresholdRegistry().quorumNumbersRequired();
+    }
+
+    /**
+     * @notice Returns the quorum confirmation threshold percentages
+     * @return bytes The confirmation threshold percentages for each quorum
+     * @dev Can be overridden by derived contracts
+     */
+    function _quorumConfirmationThresholdPercentages() internal view virtual returns (bytes memory) {
```
pakim249CAL, line 134: @samlaf #1472 

> does bytes behave the same as []uint8 in solidity? If so we can change the type

The behavior is actually quite different, but I am a fan of changing the type anyway. However, this is a threshold registry level change, so should probably be part of another PR.

contracts/src/periphery/EigenDACertVerifierV1.sol
```diff
@@ -0,0 +1,152 @@
...
+     * @dev Can be overridden by derived contracts
+     */
+    function _quorumNumbersRequired() internal view virtual returns (bytes memory) {
+        return _thresholdRegistry().quorumNumbersRequired();
+    }
+
+    /**
+     * @notice Returns the quorum confirmation threshold percentages
+     * @return bytes The confirmation threshold percentages for each quorum
+     * @dev Can be overridden by derived contracts
```
pakim249CAL, line 132: @samlaf #1472 

> "can be override by derived contracts"
Is inheritance an accepted practice in smart contract development? In general purpose languages there's a huge movement against OOP/inheritance practices, as they are more bug prone and harder to reason about.
Quick googling found https://medium.com/consensys-diligence/a-case-against-inheritance-in-smart-contracts-d7f2c738f78e, curious for your thoughts.

> The alternative here being that quorumConfirmationThresholdPercentages would be called in a separate contact which is linked at constructor time (composition style as opposed to inheritance). Perhaps the problem for smart contracts is gas costs because this prevents inlining?

> Also do you have an example of why a deriving contract would want to override this? What else would they use, other than calling the threshold registry?

I actually am totally on board with reducing inheritance, and I also have a personal vendetta against inheritance. However, this PR focuses on replicating how the current CertVerifier functions, which essentially is inheriting the functionality of both V1 and V2, and there's no other way to write this unless we start using a new storage pattern.

The default solidity storage layout is the biggest obstacle to getting rid of inheritance in contracts. And I do not intend for this PR to introduce a new storage layout schema. That will be future work.

If we want to remove inheritance, I think we should simply remove the V1_V2 contract and only use named versions.

pakim249CAL: >"can be override by derived contracts"
Is inheritance an accepted practice in smart contract development? In general purpose languages there's a huge movement against OOP/inheritance practices, as they are more bug prone and harder to reason about.
Quick googling found https://medium.com/consensys-diligence/a-case-against-inheritance-in-smart-contracts-d7f2c738f78e, curious for your thoughts.

I actually am totally on board with reducing inheritance, and I also have a personal vendetta against inheritance. However, this PR focuses on replicating how the current CertVerifier functions, which essentially is inheriting the functionality of both V1 and V2, and there's no other way to write this unless we start using a new storage pattern.

The default solidity storage layout is the biggest obstacle to getting rid of inheritance in contracts. And I do not intend for this PR to introduce a new storage layout schema. That will be future work.

If we want to remove inheritance, I think we should simply remove the V1_V2 contract and only use named versions.

pakim249CAL: > The alternative here being that quorumConfirmationThresholdPercentages would be called in a separate contact which is linked at constructor time (composition style as opposed to inheritance). Perhaps the problem for smart contracts is gas costs because this prevents inlining?

Yes, your proposal is definitely less efficient. I don't think that deploying another contract is at all necessary here. The inheritance IMO is acceptable here as it is only one layer. I'm open to changing it, but I think it would be strange to need multiple contracts for a CertVerifier deployment.

pakim249CAL: > Also do you have an example of why a deriving contract would want to override this? What else would they use, other than calling the threshold registry?

It is so that derived contracts can ensure consistency between the threshold registry they use across CertVerifier versions in case they do not want the threshold registry to be immutable.

contracts/src/periphery/EigenDACertVerifierV1.sol
```diff
@@ -0,0 +1,152 @@
...
+        );
+        CertLib.revertOnError(err, errParams);
+    }
+
+    /**
+     * @notice Verifies a batch of blob certs for the required quorums
+     * @param blobHeaders Pointer to array of blob headers in calldata
+     * @param blobVerificationProofs Pointer to array of blob cert verification proofs in calldata
+     */
+    function verifyDACertsV1(BlobHeader[] calldata blobHeaders, BlobVerificationProof[] calldata blobVerificationProofs)
```
pakim249CAL, line 63: > I don't think anyone is using this afaiu. Should we just remove? This also feels like something that can be provided by some outer layer like a multicall contract right?

I agree. However, I think removals can be done in another PR as another breaking change, to limit the already pretty large scope of this PR.

pakim249CAL: I agree. However, I think removals can be done in another PR as another breaking change, to limit the already pretty large scope of this PR.

contracts/src/periphery/EigenDACertVerifierV1.sol
```diff
@@ -0,0 +1,152 @@
...
+        eigenDAThresholdRegistryV1 = _eigenDAThresholdRegistryV1;
+        eigenDABatchMetadataStorageV1 = _eigenDABatchMetadataStorageV1;
+    }
+
+    /**
+     * @notice Verifies a the blob cert is valid for the required quorums
+     * @param blobHeader Pointer to the blob header in calldata
+     * @param blobVerificationProof Pointer to the blob cert verification proof in calldata
+     */
+    function verifyDACertV1(BlobHeader calldata blobHeader, BlobVerificationProof calldata blobVerificationProof)
```
pakim249CAL, line 44: @samlaf #1472 

> add a notice that this reverts

contracts/src/periphery/EigenDACertVerifierV2.sol
```diff
@@ -0,0 +1,260 @@
...
+    /// @notice The EigenDA middleware OperatorStateRetriever contract address
+    OperatorStateRetriever public immutable operatorStateRetrieverV2;
+
+    /// @notice The EigenDA middleware RegistryCoordinator contract address
+    IRegistryCoordinator public immutable registryCoordinatorV2;
+
+    /// @notice Security thresholds for V2 certificate verification
+    /// Storing individual fields as immutable instead of the struct
+    uint8 public immutable securityThresholdsConfirmation;
+    uint8 public immutable securityThresholdsAdversary;
```
pakim249CAL, line 39: @samlaf #1472 

prob good to add small description of what these are and how they are used

contracts/src/periphery/EigenDACertVerifierV1_V2.sol
```diff
@@ -0,0 +1,88 @@
...
+     * @dev Overrides both V1 and V2 implementations to use a single registry for consistency
+     */
+    function _thresholdRegistry()
+        internal
+        view
+        override(EigenDACertVerifierV1, EigenDACertVerifierV2)
+        returns (IEigenDAThresholdRegistry)
+    {
+        // This contract enforces that V1 and V2 use the same registry on construction. So we just choose V2.
+        return eigenDAThresholdRegistryV2;
```
pakim249CAL, line 72: @samlaf #1472 

is inheritance really the right pattern here? Seems kind of bug prone to me with related asserts/checks in different non-local places like this.

pakim249CAL: I dislike it too! However, this is the only well understood way to do it in solidity with the default solidity storage layout.

contracts/src/periphery/EigenDACertVerifierV1.sol
pakim249CAL, line 20: @samlaf #1472

>is this contract maintained by us for V1? What are the currently set thresholds? Who can change them? Mantle is thinking of doing pessimistic cert verification in their op integration, so would be good to capture and explain all of these concerns/explanations as a comment here.

pakim249CAL: I was thinking about making those parameters more flexible. However, I think that is out of scope of this PR, which aims to replicate the current CertVerifier and focuses on refactoring the structure of verification and separate concerns while keeping behavior changes at a minimum.

contracts/src/periphery/EigenDACertVerifierV1_V2.sol
pakim249CAL: In a future PR, I might actually just propose that we remove combined version verification contracts entirely. This is kind of invasive to the current state of CertVerifier test setups though, which I don't want to deal with in this PR

contracts/src/periphery/EigenDACertVerifierV1.sol
pakim249CAL: Sure, I can remove this, and this would also remove length mismatch too.

pakim249CAL:
My worklog with Claude for addressing review comments:

https://gist.github.com/pakim249CAL/f6579c1ac894c10488298507ec417e2f

contracts/src/periphery/EigenDACertVerifierV1.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1477/commits/33e719bddee8e5b701c314100d1ceb1f1ca33faf

contracts/src/periphery/EigenDACertVerifierV1.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1477/commits/c7963d237e43d0cb722f1bad8bbe4fef35d933a4

contracts/src/periphery/EigenDACertVerifierV1.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1477/commits/96fda95576fc07c3045029a01d4df814b6eff3f8

contracts/src/periphery/EigenDACertVerifierV1.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1477/commits/96fda95576fc07c3045029a01d4df814b6eff3f8

contracts/src/periphery/EigenDACertVerifierV1.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1477/commits/c7963d237e43d0cb722f1bad8bbe4fef35d933a4

contracts/src/periphery/EigenDACertVerifierV1.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1477/commits/5f4c14a531d297dc936940c7b45b36bd20c6cc30

contracts/src/periphery/EigenDACertVerifierV2.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1477/commits/5f4c14a531d297dc936940c7b45b36bd20c6cc30

anupsv:
I have a concern here. I agree the refactor looks good.

But these contracts were audited and it will be bad if we split this up to this extent now because the audit report will now be pointing to files that don't even exist. The change needed for arbitrum, is something we definitely need to do and it's something we can get by with minimal changes. But i think we need to separate that change with the refactor overall because the audit will be a public pdf and anyone asking to verify will raise concerns here. 

If this wasn't audited, it wouldn't be a problem. This can probably be included for the next Eigenda proxy audit or similar. 
@pakim249CAL

pakim249CAL:
Changes will likely be required for arbitrum anyway though, which would necessitate an audit, right?

anupsv:
> Changes will likely be required for arbitrum anyway though, which would necessitate an audit, right?

Not necessarily, it's minimal so we can do it internally.

pakim249CAL:
Well in that case, what if I just added the new CertVerifier in periphery? e.g. I don't have to delete the old one. It can stay as is. But I don't want to maintain the old one at all, deploy the old one in the future, provide support for it, etc.. The contract is not structured in a way that makes traceability easy at all. I think it is worth another audit if needed. 

For now, maybe I'll just stick this onto another branch though. I don't think it's super urgent, but this kind of redesign is IMO going to prove to be very important for maintaining the contracts into the future, and for providing better support to our integrators (and maybe obviating the need to provide support entirely)

contracts/script/GenerateUnitTestHashes.s.sol
samlaf: Why not? Just delete the comment, its clearly wrong and was copy-pasted from somewhere else. I think its fine to include this kind of small change as part of a larger PR; otherwise it'll never be done because we'll just forget.

contracts/src/periphery/EigenDACertVerifierV1.sol
samlaf: I'm confused by the change. You deleted the function?

contracts/src/periphery/EigenDACertVerifierV1.sol
samlaf: Wait so a uint8 array actually uses 32 bytes per uint8??

contracts/src/periphery/EigenDACertVerifierV1.sol
samlaf: thanks for the detailed answers, useful

contracts/src/periphery/EigenDACertVerifierV1_V2.sol
samlaf: sgtm!

contracts/src/periphery/EigenDACertVerifierV1.sol
samlaf: Can't tell if I am reading Patrick Kim or Claude for half of these comments :P

How would you make them more flexible? You haven't answered my most important question: is this contract maintained by us, and shared by all rollups? Who has permission to change the thresholds?

samlaf:
@anupsv I have to agree with Patrick here. These contracts really need to be reworked in order to be maintainable. Patrick's work will also probably make them easier to audit in the future. Also we'll need full audits for op/arbitrum secure integrations I feel.
Also if Patrick can show that the bytecode between this refactor and the previous one is the same, isnt that enough to convince an auditor that this was just a readability/maintainability refactor?

anupsv:
> @anupsv I have to agree with Patrick here. These contracts really need to be reworked in order to be maintainable. Patrick's work will also probably make them easier to audit in the future. Also we'll need full audits for op/arbitrum secure integrations I feel. Also if Patrick can show that the bytecode between this refactor and the previous one is the same, isnt that enough to convince an auditor that this was just a readability/maintainability refactor?
> 
> @pakim249CAL what's the approach we're taking for these contract changes re branching/versioning? Is this really being merged into master? Is that because the bytecode is equivalent to what's deployed? Doesn't this still deserve a redeploy (with etherscan verification) before we merge to master?

I definitely agree that the refactor is worth it and makes sense. There's a distinction i'm making, which is, if we are not going to deploy the current audited version at all, and wait for these refactored code to be audited then deployed that's fine. But if we are going to deploy the refactored code today, this will not correspond to the audit that was done. Bytecode aside, there's definitely some confidence that is drawn that the same version or a version with minimal changes was deployed to mainnet. I also wouldn't be able to say that the code is audited if someone asked because with the interfaces, peripheral contracts etc, it would give them way more pause to agree and understand.

pakim249CAL:
@samlaf The bytecode won't be the same. However, all current test cases for the CertVerifier pass except one check that is moved to a constructor instead of verification. Admittedly though, there are not that many test cases which might have something to do with the monolithic structure of the current CertVerifier. 

I'll return this to a draft since I probably won't be merging this PR this week. Maybe I'll start a branch that will collect contract changes from PRs that I want to make.

contracts/src/periphery/EigenDACertVerifierV1.sol
pakim249CAL: in memory, yes.

contracts/src/periphery/EigenDACertVerifierV1.sol
pakim249CAL: to be clear, im not answering questions in this PR with claude haha. The focus of this PR is only to refactor the CertVerifier with minimal opinionated changes, which can be extended with opinionated changes after this PR is merged. And so I think it is appropriate to defer answering this question until another PR that builds on top of this one. 

CertVerifiers, as they currently are, are meant to be deployed by a rollup with their desired quorum numbers and security parameters. This is true for the current contracts, as well as the new contracts in this PR. There is no change in behavior on that front in this PR. The CertVerifier is a non-upgradeable contract that stores unchangeable threshold parameters that cannot be changed even by the owner.

I do not actually know the answer to your questions to answer with confidence here. However, I believe this design makes creating new CertVerifiers that fit the needs of each specific rollup MUCH easier than before, and maybe with the design of the library, we do not need a "canonical" CertVerifier contract.

contracts/script/GenerateUnitTestHashes.s.sol
pakim249CAL: If that is the reason, I must insist even further that it belongs in another PR since this one will likely not be merged for a while since I anticipate it will be hard to get consensus on it. I will put the fix in [this PR](https://github.com/Layr-Labs/eigenda/pull/1482) which we should be able to merge right away, and remove this changeset from this PR entirely.

ethenotethan:
@pakim249CAL would prefer if we prioritized immediate short-term business blockers on the cert verifier (i.e, status code interpretation, cert field orderings). Tbh its very hard to confidently review this change-set given its incredibly large size and scope of changes. Also agree with @anupsv that we shouldn't deviate too hard given this contract already has production audit guarantees which could very well be invalidated here. In the future it'd be nice to have corresponding proposals/specs which receive team consensus to avoid potential throwaway work. That said I do agree with the structure of this refactor - but also do feel like unblocking V2 on mainnet for integrations is the highest priority item for us at the moment.

pakim249CAL:
> @pakim249CAL would prefer if we prioritized immediate short-term business blockers on the cert verifier (i.e, status code interpretation, cert field orderings). Tbh its very hard to confidently review this change-set given its incredibly large size and scope of changes. Also agree with @anupsv that we shouldn't deviate too hard given this contract already has production audit guarantees which could very well be invalidated here. In the future it'd be nice to have corresponding proposals/specs which receive team consensus to avoid potential throwaway work. That said I do agree with the structure of this refactor - but also do feel like unblocking V2 on mainnet for integrations is the highest priority item for us at the moment.

Not sure what you mean by status code interpretation, but cert field orderings are ultimately not going to be viable without some serious upgrade consequences. Any changes in field ordering in a cert's batch header (which is where RBN is), as I noted in another comment, is going to break all previous cert verifications. This is due to the fact that the verifications rely on the hash of the batch header which would obviously not hash the same with different field orders.

pakim249CAL:
> @pakim249CAL what's the approach we're taking for these contract changes re branching/versioning? Is this really being merged into master? Is that because the bytecode is equivalent to what's deployed? Doesn't this still deserve a redeploy (with etherscan verification) before we merge to master?

I think it is enough to simply just have this on a separate branch. Though there are a lot of modifications here, I think I can minimize the changeset to be something that can cleanly live on another branch, being just the new implementation of CertVerifier without deleting and replacing the old one. Then we can have flexibility in deciding how we want to release it with minimal updates on this PR needed.

The bytecode is definitely NOT going to be equivalent, but the overall logic should be the same.

The CertVerifier is NOT a core contract. It is ultimately just a contract that reads state from other contracts and computes a result, and is meant to be deployed by a rollup, which means that there is no such thing currently as a "canonical" CertVerifier. As such, I do not think that the CertVerifier needs a release process of its own. This is also why I chose to move the CertVerifier from "core" to "periphery" in this PR.

I do like the idea of redeploying the new CertVerifier and verifying it. However, I don't think I'll end up merging to master soon. For now, I think I will create a contract feature branch that includes this new design so that there is no pressure to merge this changeset to master.

PR #1481, opened 2025-04-18
feat: support batch->blobs at metadata store

jianoaix:
We have use cases to get the comprising blobs of a given batch

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests


jianoaix:
ping?

disperser/controller/dispatcher.go
```diff
@@ -554,9 +554,10 @@ func (d *Dispatcher) NewBatch(ctx context.Context, referenceBlockNumber uint64)
 
 	copy(batchHeader.BatchRoot[:], tree.Root())
 
-	batchHeaderHash, err := batchHeader.Hash()
```
dmanc, line 557: What was the reason for moving this after the timer ends?

dmanc (approved):

dmanc:
Any infra change needed?

disperser/controller/dispatcher.go
jianoaix: It's more close to batch header handling, as the stuff above is building merkle tree and measuring it's latency.

jianoaix:
> Any infra change needed?

Should not have, this doesn't change schema list

PR #1492, opened 2025-04-23
refactor: separate v1 v2 cert contracts

pakim249CAL:

Separates out V1 and V2 cert verifier contracts. Top level cert verifier still has the same interface as before.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


contracts/src/periphery/EigenDACertVerifierV1.sol
```diff
@@ -0,0 +1,117 @@
...
+        eigenDAThresholdRegistryV1 = _eigenDAThresholdRegistryV1;
+        eigenDABatchMetadataStorageV1 = _eigenDABatchMetadataStorageV1;
+    }
+
+    /**
+     * @notice Verifies a the blob cert is valid for the required quorums
+     * @param blobHeader The blob header to verify
+     * @param blobVerificationProof The blob cert verification proof to verify
+     */
+    function verifyDACertV1(BlobHeader calldata blobHeader, BlobVerificationProof calldata blobVerificationProof)
```
ethenotethan, line 43: nit - why have additional redundancy with V1 suffix given versioning is already declared in contract name?

ethenotethan (approved):
LGTM - very clean and straightforward imo. just one nit

contracts/src/periphery/EigenDACertVerifierV1.sol
pakim249CAL: This is to maintain compatibility with the old CertVerifier interface. Note that the CertVerifier inherits CertVerifierV1 and CertVerifierV2. In this way, the interface of the CertVerifier contract does not change, and so this PR is not a breaking change.

pakim249CAL: Note that if the suffix was not here, aka it was just called verifyDACert, then this function selector would need to be exposed in inheriting contracts! which is a problem if we are trying to inherit and expose functions from both V1 and V2.

contracts/src/periphery/EigenDACertVerifierV2.sol
```diff
@@ -0,0 +1,218 @@
...
+
+    /**
+     * @notice Verifies a blob cert using the immutable required quorums and security thresholds set in the constructor
+     * @param batchHeader The batch header of the blob
+     * @param blobInclusionInfo The inclusion proof for the blob cert
+     * @param nonSignerStakesAndSignature The nonSignerStakesAndSignature to verify the blob cert against
+     * @param signedQuorumNumbers The signed quorum numbers corresponding to the nonSignerStakesAndSignature
+     */
+    function verifyDACertV2(
+        BatchHeaderV2 calldata batchHeader,
```
bxue-l2, line 83: Are we still planning to consolidate the four fields into one?
- BatchHeaderV2
- BlobInclusionInfo
- NonSignerStakesAndSignature
- signedQuorumNumbers

contracts/src/periphery/EigenDACertVerifierV2.sol
```diff
@@ -0,0 +1,218 @@
...
+     * @param nonSignerStakesAndSignature The nonSignerStakesAndSignature to verify the blob cert against
+     * @param signedQuorumNumbers The signed quorum numbers corresponding to the nonSignerStakesAndSignature
+     */
+    function verifyDACertV2ForZKProof(
+        BatchHeaderV2 calldata batchHeader,
+        BlobInclusionInfo calldata blobInclusionInfo,
+        NonSignerStakesAndSignature calldata nonSignerStakesAndSignature,
+        bytes memory signedQuorumNumbers
+    ) external view returns (bool) {
+        (CertV2Lib.StatusCode status,) = CertV2Lib.checkDACertV2(
```
bxue-l2, line 138: Would it be safer to have a catch all statement, in case somewhere in the dependency went wrong? 
Do you think no-reversion is just a basic thing we have to (or simple) to ensure

bxue-l2 (commented):
very new to contract development, it is a learning for me. But here are some questions

contracts/src/core/EigenDACertVerifier.sol
```diff
@@ -50,164 +28,19 @@ contract EigenDACertVerifier is IEigenDACertVerifier {
...
-            CertV2Lib.getNonSignerStakesAndSignature(operatorStateRetriever, registryCoordinator, signedBatch);
-        return nonSignerStakesAndSignature;
-    }
-
-    /**
-     * @notice Verifies the security parameters for a blob cert
-     * @param blobParams The blob params to verify
-     * @param securityThresholds The security thresholds to verify against
-     */
     function verifyDACertSecurityParams(
```
litt3, line 44: Did you intentionally delete the function docs from all these functions?

contracts/src/interfaces/IEigenDACertVerifierV2.sol
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+import {IEigenDAThresholdRegistry} from "./IEigenDAThresholdRegistry.sol";
+import "./IEigenDAStructs.sol";
+
+interface IEigenDACertVerifierV2 {}
```
litt3, line 7: Did you not add any functions here because the signatures to use are still slated to change?

contracts/src/periphery/EigenDACertVerifierV1.sol
```diff
@@ -0,0 +1,117 @@
...
+import {EigenDACertVerificationV1Lib as CertV1Lib} from "src/libraries/EigenDACertVerificationV1Lib.sol";
+import {EigenDACertVerificationV2Lib as CertV2Lib} from "src/libraries/EigenDACertVerificationV2Lib.sol";
+import {OperatorStateRetriever} from "../../lib/eigenlayer-middleware/src/OperatorStateRetriever.sol";
+import {IRegistryCoordinator} from "../../lib/eigenlayer-middleware/src/RegistryCoordinator.sol";
+import {IEigenDARelayRegistry} from "../interfaces/IEigenDARelayRegistry.sol";
+import "../interfaces/IEigenDAStructs.sol";
+
+/**
+ * @title A CertVerifier is an immutable contract that is used by a consumer to verify EigenDA blob certificates
+ * @notice For V2 verification this contract is deployed with immutable security thresholds and required quorum numbers,
```
litt3, line 17: You mention `V2`, but this is the `V1` verifier. Shouldn't `V2` specific docs live there?

contracts/src/periphery/EigenDACertVerifierV2.sol
```diff
@@ -0,0 +1,218 @@
...
+/**
+ * @title A CertVerifier is an immutable contract that is used by a consumer to verify EigenDA blob certificates
+ * @notice For V2 verification this contract is deployed with immutable security thresholds and required quorum numbers,
+ *         to change these values or verification behavior a new CertVerifier must be deployed
+ */
+contract EigenDACertVerifierV2 is IEigenDACertVerifierV2 {
+    error InvalidSecurityThresholds();
+
+    /// @notice The EigenDAThresholdRegistry contract address
+    IEigenDAThresholdRegistry public immutable eigenDAThresholdRegistryV2;
```
litt3, line 30: function names need the version suffix for backwards compatibility, but do these fields? Can't this be named like e.g. `eigenDAThresholdRegistry`?

contracts/src/interfaces/IEigenDACertVerifierV2.sol
pakim249CAL: Yes, I basically just wanted to make a place for the interface without actually defining it yet.

contracts/src/libraries/EigenDACertVerificationV2Lib.sol
```diff
@@ -64,7 +64,7 @@ library EigenDACertVerificationV2Lib {
     /// @param relayKey The relay key that is not set
     error RelayKeyNotSet(uint32 relayKey);
 
-    /// @notice Error codes for certificate verification results
+    /// @notice Status codes for certificate verification results
```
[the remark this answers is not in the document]
pakim249CAL: Correct. I will fix this in another PR.

contracts/src/periphery/EigenDACertVerifierV2.sol
pakim249CAL: Yes, in another PR.

contracts/src/periphery/EigenDACertVerifierV2.sol
pakim249CAL: Yes, I think that would be a good change to make. I will include it in another PR.

contracts/src/core/EigenDACertVerifier.sol
pakim249CAL: No, but this contract is slated for deletion in a future PR anyway, as well as most functions in this contract in general. With that context, do you think it would be necessary to fix this?

contracts/src/periphery/EigenDACertVerifierV2.sol
pakim249CAL: Only if it were private! It's probably not ideal, but this contract is meant to be inherited to mirror the current interface of the CertVerifier, which means that if any other inherited contract declared the same variables, the compiler would not accept this. I plan to delete the CertVerifier in a future PR, which would allow us to remove this annoyance, but this PR prioritizes replicating the previous CertVerifier behavior by splitting up V1 and V2 and relying on inheritance.

contracts/src/periphery/EigenDACertVerifierV1.sol
pakim249CAL: It is even worse than you say, as the entire comment is pretty much incorrect and does not apply to V1 verification. I will update this.

contracts/src/core/EigenDACertVerifier.sol
ethenotethan: I'd argue its a quick fix and worth not introducing a regression here if possible

contracts/src/periphery/EigenDACertVerifierV1.sol
ethenotethan: I'd argue this contract is literally irrelevant now since we have a V1 cert verifier pinned at release 0.1.0 which has production audit guarantees

contracts/src/core/EigenDACertVerifier.sol
samlaf, line 1: Would be nice to add some tooling to catch breaking contract API changes. I trust you but hard to know right now whether this contract actually keeps the same ABI as the previous contract.
I think arbitrum has something like this @ethenotethan ? Or is it only for storage slot diffs?
Maybe what we could do is just commit the generated ABIs to some folder, and then have CI check that those files are up to date. So any breaking ABI changes would show up as commited files in PRs, and would be easy to catch? Also would help us enforce correct semver upgrades.

samlaf: compiled both versions and there are some diffs in the ABI (mostly contract name changes in storage), see https://www.diffchecker.com/JMkOLhvB/
For example
image

Do we care about these? Prob not but raising in case.

samlaf (approved):
LGTM although there are small API breaking changes. Think they are bening but overall would still like us at some point to develop tooling to automatically catch those in PRs, the same way we've added for flags in proxy cc @ethenotethan @litt3

contracts/src/core/EigenDACertVerifier.sol
pakim249CAL: Ultimately I believe this is what interface files are for. Ideally, off-chain integrators would only use the minimal interfaces we define, rather than having them use the entire contract's ABI to leave room for me to make changes like this. I think we are making good progress in doing this with our future plans for the cert verifier. With this format, we can enforce that the expected interfaces for integrators do not break at compile time.

samlaf: But then why not just make those storage slots private? I don't think a contract implementer should rely on people using the interface he published. People should and will write the interface file that they need/want. If methods/storage slots need to not be public, then they shouldnt be no?

pakim249CAL: I'm of the view that it is generally desirable for every variable to have some getter for it. This is simply because getters are much more convenient for reviewing parameters than examining constructor/init arguments.

I believe the interface is where things should be much more specific, and integrators should generally only interact with interfaces that we define for them.

contracts/src/periphery/EigenDACertVerifierV2.sol
```diff
@@ -0,0 +1,218 @@
...
+            blobInclusionInfo,
+            _securityThresholds(),
+            _quorumNumbersRequired()
+        );
+    }
+
+    /**
+     * @notice Thin try/catch wrapper around verifyDACertV2 that returns false instead of panicing
+     * @dev The Steel library (https://github.com/risc0/risc0-ethereum/tree/main/crates/steel)
+     *      currently has a limitation that it can only create zk proofs for functions that return a value
```
anupsv, line 126: nit: Probably true for SP1 too. Do we care enough to change the wording ?

contracts/src/periphery/EigenDACertVerifierV1.sol
pakim249CAL: I agree with the irrelevance of this contract, but IMO it is better to avoid removal in this PR to make it easier to see that no important functionality was removed in this PR.

pakim249CAL: @litt3 comment removed

contracts/src/core/EigenDACertVerifier.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1492/commits/82caf341a8ebb281cf9149fd6008f998856c8a1a

contracts/src/periphery/EigenDACertVerifierV2.sol
```diff
@@ -0,0 +1,218 @@
...
+     * @param batchHeader The batch header of the blob
+     * @param blobInclusionInfo The inclusion proof for the blob cert
+     * @param nonSignerStakesAndSignature The nonSignerStakesAndSignature to verify the blob cert against
+     * @param signedQuorumNumbers The signed quorum numbers corresponding to the nonSignerStakesAndSignature
+     */
+    function verifyDACertV2ForZKProof(
```
anupsv, line 132: nit: probably true to say "with return bool" or something. I guess when i first read it i thought there was an underlying call to the groth verifier. Or may be it's just me.

anupsv (approved):

contracts/src/periphery/EigenDACertVerifierV2.sol
pakim249CAL: Later!

contracts/src/core/EigenDACertVerifier.sol
samlaf: I see, that makes sense for having getters.
"integrators should generally only interact with interfaces that we define for them" feels very anti-interface.
Feels like there should be better mechanism to deal with this, but I guess contracts don't really have the notion of module-visiblity like programming languages... so maybe what you're saying makes sense. Feel like this should be documented in the natspec though, since just hoping people use the interface we defined is wishful thinking (they can after all just write a simple sol! macro to interact with whatever function they want).

But I just learned that in C bunch of librairies define both an implementer header (interior), and a public facing header (exterior), like:
https://github.com/endlesssoftware/sqlite3/blob/master/sqlite3ext.h
https://github.com/endlesssoftware/sqlite3/blob/master/sqliteint.h
so maybe would make sense to do something similar. Then of course user is always allowed to define a subset of the external interface file. Feel like there should be better tooling to do this.

pakim249CAL: Of course integrators are free to use whatever they want, and we obviously cannot restrict them from interfacing with the contract in their own way. However, I don't know of a good way to share minimal ABIs without the use of interfaces, and it is ultimately necessary IMO to limit the number of functions that we support for integrators if we want to have queryable contracts.. I think we should make clear that we only really support specific interfaces that we define for integrators, and everything else is subject to change.

litt3 (approved):

PR #1493, opened 2025-04-23
refactor: unnecessary cert checks

pakim249CAL:

cert verifier no longer checks relay registry, nor does it contain a getter for it. the rationale is that relays are not trusted in the first place, so there is no point in verifying them. Also, the relays are included in a struct that is hashed and checked for inclusion.

also removes a check that confirmation threshold > adversary threshold from verification, instead favoring doing this check at construction.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


contracts/src/periphery/EigenDACertVerifierV2.sol
```diff
@@ -32,9 +32,6 @@ contract EigenDACertVerifierV2 is IEigenDACertVerifierV2 {
     /// @notice The EigenDASignatureVerifier contract address
     IEigenDASignatureVerifier public immutable eigenDASignatureVerifierV2;
 
-    /// @notice The EigenDARelayRegistry contract address
-    IEigenDARelayRegistry public immutable eigenDARelayRegistryV2;
-
     /// @notice The EigenDA middleware OperatorStateRetriever contract address
     OperatorStateRetriever public immutable operatorStateRetrieverV2;
```
samlaf, line 36: Btw unlreated to this PR but is there not a naming convention in solidity contracts for global variables (contract storage slots). I find it hard when reading a function to just see a bunch of variables, and I don't know which ones were passed as arguments vs which one are contract variables (and on top of that which of the contract variables are mutable vs immutable). Feels like a naming convention could help here?

contracts/src/libraries/EigenDACertVerificationV2Lib.sol
```diff
@@ -237,12 +215,6 @@ library EigenDACertVerificationV2Lib {
...
     {
-        if (securityThresholds.confirmationThreshold <= securityThresholds.adversaryThreshold) {
-            return (
-                StatusCode.INVALID_THRESHOLD_PERCENTAGES,
-                abi.encode(securityThresholds.confirmationThreshold, securityThresholds.adversaryThreshold)
-            );
-        }
         uint256 gamma = securityThresholds.confirmationThreshold - securityThresholds.adversaryThreshold;
         uint256 n = (10000 - ((1_000_000 / gamma) / uint256(blobParams.codingRate))) * uint256(blobParams.numChunks);
         uint256 minRequired = blobParams.maxNumOperators * 10000;
```
samlaf, line 220: Most likely unrelated to this PR, but I'm a bit puzzled as to this checkSecurityParams function overall, since it related VersioninBlobParams (which are eigenda set) and securityThresholds (which are rollup set). So certs could randomly start not verifying because of a mismatch between the options chosen by rollup and eigenda independently it seems..

samlaf (approved):
PR itself LGTM. I have 2 important questions that Ive left as comment that I would like to discuss though, here or elsewhere.

contracts/src/libraries/EigenDACertVerificationV2Lib.sol
```diff
@@ -60,26 +55,19 @@ library EigenDACertVerificationV2Lib {
...
 
     }
 
     function verifyDACertV2(
         IEigenDAThresholdRegistry eigenDAThresholdRegistry,
         IEigenDASignatureVerifier signatureVerifier,
-        IEigenDARelayRegistry relayRegistry,
         BatchHeaderV2 memory batchHeader,
         BlobInclusionInfo memory blobInclusionInfo,
         NonSignerStakesAndSignature memory nonSignerStakesAndSignature,
```
samlaf, line 73: on second thought this changes the function signature, and hence the abi of the contract (hence the `!` in the PR title I believe?)

Wondering whether we shouldn't at least for now keep the function signature the same even if we don't use the arguments. Until we've settled on a way to version/upgrade this contract.

pakim249CAL: it does not change the ABI for the cert verifier. It only changes the inputs to the library. The CertVerifier itself would just no longer have a relay registry getter, and I highly doubt that this was used in the first place. (Because there's no reason to use it from the CertVerifier!)

contracts/src/periphery/EigenDACertVerifierV2.sol
pakim249CAL: Eventually I'd like to transition to using unstructured storage only, not declaring any storage as public, etc., because I hate dealing with these kinds of questions haha. But for the codebase as it currently is, I think this naming makes the most sense for now. Seems like a separate issue to tackle.

samlaf (approved):

anupsv (approved):

contracts/src/periphery/EigenDACertVerifierV2.sol
```diff
@@ -49,23 +45,23 @@ contract EigenDACertVerifierV2 is IEigenDACertVerifierV2 {
...
         IEigenDASignatureVerifier _eigenDASignatureVerifierV2,
-        IEigenDARelayRegistry _eigenDARelayRegistryV2,
         OperatorStateRetriever _operatorStateRetrieverV2,
         IRegistryCoordinator _registryCoordinatorV2,
         SecurityThresholds memory _securityThresholdsV2,
         bytes memory _quorumNumbersRequiredV2
     ) {
+        if (_securityThresholdsV2.confirmationThreshold <= _securityThresholdsV2.adversaryThreshold) {
+            revert InvalidSecurityThresholds();
+        }
```
ethenotethan, line 62: can we add a comment explaining why this check is required?

pakim249CAL: I don't actually have a comment to add, as I am not sure myself why the check is required. But it is a check that existed in the old CertVerifier, so I think this issue should be out of scope here.

api/clients/v2/verification/cert_verifier.go
```diff
@@ -184,26 +184,6 @@ func (cv *CertVerifier) GetQuorumNumbersRequired(ctx context.Context) ([]uint8,
 	return quorumNumbersRequired, nil
 }
 
-// GetRelayRegistryAddress returns the address of the EigenDARelayRegistry contract
-func (cv *CertVerifier) GetRelayRegistryAddress(ctx context.Context) (*gethcommon.Address, error) {
```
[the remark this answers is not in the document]
pakim249CAL: I am not sure what contracts proxy already interfaces with, but the relay registry address is also available from the service manager contract. Would that work as a replacement?

litt3 (approved):
LGTM

bxue-l2:
https://github.com/Layr-Labs/eigenda/pull/1494/files was going to merge into this. 

Restore for now, to see the diff

PR #1494, opened 2025-04-23
refactor: remove operator state retriever

pakim249CAL:

The operator state retriever is stateless. We can just incorporate the logic of that contract into the library rather than having it be an external dependency.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


contracts/src/periphery/EigenDACertVerifierV2.sol
```diff
@@ -32,9 +32,6 @@ contract EigenDACertVerifierV2 is IEigenDACertVerifierV2 {
     /// @notice The EigenDASignatureVerifier contract address
     IEigenDASignatureVerifier public immutable eigenDASignatureVerifierV2;
 
-    /// @notice The EigenDA middleware OperatorStateRetriever contract address
-    OperatorStateRetriever public immutable operatorStateRetrieverV2;
```
samlaf, line 36: I guess the original thinking with this contract is that we'd have a single one shared by all AVSs so we'd save on deployment gas costs by not having every single contract deploy it as an embedded library.

But if its immutable pointer then agree that it doesn't make a lot of sense, unless we're reaching the bytesize limit of the contract. So I think I'm fine with this change although I don't know the solidity best practices.

contracts/src/libraries/EigenDACertVerificationV2Lib.sol
```diff
@@ -353,28 +348,112 @@ library EigenDACertVerificationV2Lib {
...
+
+    struct CheckSignaturesIndices {
+        uint32[] nonSignerQuorumBitmapIndices;
+        uint32[] quorumApkIndices;
+        uint32[] totalStakeIndices;
+        uint32[][] nonSignerStakeIndices; // nonSignerStakeIndices[quorumNumberIndex][nonSignerIndex]
+    }
+
+    /**
+     * @notice copied from OperatorStateRetriever.sol (which should probably be a library in the first place)
```
samlaf, line 385: I'm fine with a little bit of copying, but is the idea here that we're moving away from using the middleware repo as much as we can?

Should we instead put in the extra effort to do the refactor into a library in the middleware repo and import it from there?

samlaf (approved):
LGTM but I'd like to understand how you plan to merge this stack of PRs you have going. Prob best to wait until they are all approved before merging any of them?

contracts/src/core/EigenDACertVerifier.sol
```diff
@@ -32,7 +31,6 @@ contract EigenDACertVerifier is EigenDACertVerifierV1, EigenDACertVerifierV2 {
         EigenDACertVerifierV2(
             _eigenDAThresholdRegistry,
             _eigenDASignatureVerifier,
-            _operatorStateRetriever,
             _registryCoordinator,
             _securityThresholdsV2,
             _quorumNumbersRequiredV2
```
samlaf, line 36: on second thought this changes the function signature, and hence the abi of the contract. So PR title should be `refactor!` right? What's your semantic for using `!` or not?
Wondering whether we shouldn't at least for now keep the function signature the same even if we don't use the arguments. Until we've settled on a way to version/upgrade this contract.

contracts/src/periphery/EigenDACertVerifierV2.sol
pakim249CAL: I think it is important to reduce dependencies on "deployed" contracts, as this is a prerequisite to lessening the number of contracts we have to maintain overall.

contracts/src/libraries/EigenDACertVerificationV2Lib.sol
pakim249CAL: I think it would be great if middleware had these things as libraries. However, I do not expect that I would be able to merge changes like this into middleware very quickly. There's a lot of middleware that I would like to have as libraries, so maybe that can be a more focused and separate task.

contracts/src/core/EigenDACertVerifier.sol
pakim249CAL: Only the operator state retriever getter is gone, which I highly doubt was being used. Contract constructor arguments are also different, but that is all. The interface to verify remains the same.

pakim249CAL:
> LGTM but I'd like to understand how you plan to merge this stack of PRs you have going. Prob best to wait until they are all approved before merging any of them?

I don't think the PRs necessarily need to be merged atomically. I think it is better that PRs get merged as soon as they are satisfactory and it is verified by relevant parties that they do not break anything. Waiting for all the PRs would likely lead to lots of time wasted rebasing for minor reasons. Nitpicks and such can always be tackled in another PR on top, as long as a PR is an improvement off of the current work and there are no regressions.

contracts/src/core/EigenDACertVerifier.sol
samlaf: Ah right I'm dumb this is just a breaking change on constructor, which is not important (you've updated the deployer script to work with this new constructor. So LGTM, thanks for explanation

samlaf (approved):
Changes LGTM, though will need to be rebased on top of latest stuff

contracts/src/libraries/EigenDACertVerificationV2Lib.sol
pakim249CAL: The thing is though that logic changes in middleware cannot happen without us upgrading our contracts. So I think this is still acceptable, and I have some plans to re-implement some key parts of middleware here so that we are not beholden to their implementation details. Fetching operator state is one of those implementation details that we should control, rather than relying on middleware IMO

contracts/src/interfaces/IEigenDACertVerifierV2.sol
```diff
@@ -0,0 +1,7 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+import {IEigenDAThresholdRegistry} from "./IEigenDAThresholdRegistry.sol";
+import "./IEigenDAStructs.sol";
+
+interface IEigenDACertVerifierV2 {}
```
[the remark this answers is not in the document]
pakim249CAL: Right, I guess I should just delete it for now since we will be defining this in a PR soon.

contracts/src/libraries/EigenDACertVerificationV2Lib.sol
ethenotethan: right but we already rely on an OperatorStateRetriever binding across the core protocol functions (e.g, decentralized retrieval). A stateful operator retriever is already revered as a first class citizen within the protocol itself so having it as a dependency on the contracts isn't overly burdening and better aligns onchain/offchain interactions. 

Biggest concern with this PR is we'd be fixing a small regression while introducing a large one. I.e we're not eliminating middleware as a dependency while copy pasting some of its logic. So now if we bump the submodule then the copied sublogic would have potentially be updated manually with it. This is done iiuc to remove a stateful dependency for doing retrievals against - but I'd argue this won't be net positive until we can get the changes put into middleware directly

pakim249CAL: I acknowledge your points, but I feel the need to point out that it is not just a small regression we are fixing here. The operator state retriever is, for some reason, a configuration parameter. This is not correct in my view, and is a vector for misconfiguration and questions. Especially if CertVerifiers are to be deployed frequently, it would be really silly to ever have to deal with an issue just because the operator state retriever address was wrong. (And maybe they might question why they need one in the first place!) I am really most concerned with that aspect of this dependency. 

Unfortunately I don't think updating the submodule is a great choice at this point in time. Personally, I think it is better to simply remove our dependencies in middleware as much as possible. Ultimately, we are responsible for the contracts we deploy, and so I don't feel comfortable importing commits from middleware wholesale. It is better IMO if we just duplicate the logic and change it ourselves if we ever need to upgrade.

Also, it is my opinion that implementing middleware ourselves will actually be necessary to implement many features that EigenDA wants without a huge expansion in scope of contracts needed to support them. For example, role management of quorums by quorum owners will be much more manageable if we just stop using middleware since they essentially hardlock the authority mechanisms you must use.

pakim249CAL:
Closing for now.

PR #1495, opened 2025-04-24
refactor!: remove old cert verifier. create new eigendacertv2 struct

pakim249CAL:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


pakim249CAL:
I feel like this PR should have broken SOMETHING, but nothing broke, so I am concerned. I'll take a further look tomorrow.

contracts/src/core/EigenDACertVerifier.sol
```diff
@@ -1,79 +0,0 @@
-// SPDX-License-Identifier: MIT
-pragma solidity ^0.8.9;
-
-import {EigenDACertVerifierV1} from "src/periphery/EigenDACertVerifierV1.sol";
-import {EigenDACertVerifierV2} from "src/periphery/EigenDACertVerifierV2.sol";
-import {IEigenDAThresholdRegistry} from "../interfaces/IEigenDAThresholdRegistry.sol";
-import {IEigenDABatchMetadataStorage} from "../interfaces/IEigenDABatchMetadataStorage.sol";
-import {IEigenDASignatureVerifier} from "../interfaces/IEigenDASignatureVerifier.sol";
```
ethenotethan, line 8: we may need this contract - or at least its binding in the interim given we'll potentially have to support cross compatibility with the older cert verifier interface within the V2 clients.... I think. 

thoughts @samlaf @litt3 ?

samlaf: Conduit just upgraded all their rollups to V2, and are using our currently deployed EigenDACertVerifier for v2. See https://eigenlabs.slack.com/archives/C05LFNLDN3W/p1745589330959239

samlaf: I'm not following you guys' discussion here btw. So let's discuss during standup. I agree with ethen that we need to be careful as we already have people like conduit running on testnet with the current CertVerifier, so proxy will have to keep bindings for it somewhere/somehow. Perhaps this should be part of larger discussion to move to a VerifyDACert signature that only takes bytes, and what's the best way to deal with this re version management, bindings generation, etc.

api/clients/v2/verification/cert_verifier.go
```diff
@@ -184,26 +184,6 @@ func (cv *CertVerifier) GetQuorumNumbersRequired(ctx context.Context) ([]uint8,
 	return quorumNumbersRequired, nil
 }
 
-// GetRelayRegistryAddress returns the address of the EigenDARelayRegistry contract
-func (cv *CertVerifier) GetRelayRegistryAddress(ctx context.Context) (*gethcommon.Address, error) {
```
ethenotethan, line 188: this will require a change in the proxy when doing dependency bump: http://github.com/Layr-Labs/eigenda-proxy/blob/100801d1dcb4cb3d1d36455da3fad5ef6f3fe34f/store/storage_manager_builder.go#L226-L229

contracts/test/unit/EigenDACertVerifierV2Unit.t.sol
```diff
@@ -17,108 +17,46 @@ contract EigenDACertVerifierV2Unit is MockEigenDADeployer {
...
-        nonSignerStakesAndSignature.totalStakeIndices = nssas.totalStakeIndices;
-        nonSignerStakesAndSignature.nonSignerStakeIndices = nssas.nonSignerStakeIndices;
-
-        (NonSignerStakesAndSignature memory _nonSignerStakesAndSignature, bytes memory signedQuorumNumbers) =
-            CertV2Lib.getNonSignerStakesAndSignature(registryCoordinator, signedBatch);
-        bool zk = eigenDACertVerifier.verifyDACertV2ForZKProof(
-            signedBatch.batchHeader, blobInclusionInfo, _nonSignerStakesAndSignature, signedQuorumNumbers
-        );
+        EigenDACertV2 memory cert = _generateDACert(pseudoRandomNumber, 0);
+        bool zk = eigenDACertVerifier.checkDACert(cert);
```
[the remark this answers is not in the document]
pakim249CAL: Honestly I will probably just rewrite the entire test suite for the CertVerifier soon. So I'd prefer to defer this.

api/clients/v2/verification/cert_verifier.go
pakim249CAL: It's just a compilation issue though, right? Not actually being used for anything?

contracts/test/unit/EigenDACertVerifierV2Unit.t.sol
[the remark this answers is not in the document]
ethenotethan: all good its a nit for a reason

contracts/src/core/EigenDACertVerifier.sol
samlaf: @ethenotethan is this contract only used by arbitrum? I don't think any of eigenda v1 tests use it right..? Maybe mantle will need it but they can just copy-paste the "audited" version. So I think I agree we just need to keep the interface around to we can build bindings for it.

api/clients/v2/verification/cert_verifier.go
samlaf: Actually, thinking about bindings some more, shouldn't we be manually generating MINIMAL interfaces?
I just started wondering why we generated ABIs for the entire contract when individual call sites typically only need to make 1 or 2 calls, and hence only need a small subset of the ABI.
This is basically the golang interface philosophy. Define the interface at each call site. Also protects from breaking changes to the contract that I don't care about.

There's def some tooling missing to make this approach secure, but I feel like even the full-ABI generation is not super secure the way its used right now (see https://x.com/samlafer/status/1917127140090515556)

contracts/src/interfaces/IEigenDACertVerifierV2.sol
```diff
@@ -0,0 +1,9 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+import {IEigenDAThresholdRegistry} from "./IEigenDAThresholdRegistry.sol";
+import "./IEigenDAStructs.sol";
+
+interface IEigenDACertVerifierV2 {
+    function verifyDACertV2(EigenDACertV2 calldata cert) external view;
+}
```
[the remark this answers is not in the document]
pakim249CAL: Not really IMO. I think this is still a proper, discrete step towards our desired entry point.

api/clients/v2/verification/cert_verifier.go
pakim249CAL: Right! I touch on that topic [here](https://github.com/Layr-Labs/eigenda/pull/1492#discussion_r2066151413)

pakim249CAL:
The scope has changed, so closing

PR #1497, opened 2025-04-24
feat: periodically return attestation with updated signatures

litt3:
- closes EGDA-1008
    - [link](https://linear.app/eigenlabs/issue/EGDA-1008/update-attestation-status-while-in-gathering-signatures-status)


api/clients/v2/payloaddispersal/payload_disperser.go
```diff
@@ -224,12 +224,7 @@ func (pd *PayloadDisperser) pollBlobStatusUntilSigned(
 				var thresholdNotMetErr *thresholdNotMetError
 				if !errors.As(err, &thresholdNotMetErr) {
 					// an error occurred which was unrelated to an unmet threshold: something went wrong while checking!
-					//
-					// TODO (litt3): with the current disperser implementation, it's expected that the blobStatusReply
-					//  will be nil while gathering signatures, so this case will *always* be triggered. Once the
-					//  disperser has been updated to return a non-nil blobStatusReply while gathering signatures,
-					//  this case will no longer be expected to occur, so a warning should be logged here.
-					continue
+					pd.logger.Warnf("check thresholds: %v", err)
```
cody-littley, line 227:
```suggestion
					pd.logger.Warnf("error checking thresholds: %v", err)
```

disperser/controller/dispatcher.go
```diff
@@ -341,37 +349,58 @@ func (d *Dispatcher) HandleSignatures(
...
-		if quorumResult.PercentSigned > 0 {
-			nonZeroQuorums = append(nonZeroQuorums, quorumID)
-			quorumResults[quorumID] = quorumResult.PercentSigned
+	// keep track of the final attestation, since that's the attestation which will determine to final batch status
+	finalAttestation := &core.QuorumAttestation{}
+	// continue receiving attestations from the channel until it's closed
+	for receivedQuorumAttestation := range attestationChan {
+		err := d.submitAttestation(ctx, batchData, receivedQuorumAttestation)
+		if err != nil {
+			d.logger.Warnf("submit attestation for batch %s: %v", batchHeaderHash, err)
```
cody-littley, line 387:
```suggestion
			d.logger.Warnf("error submitting attestation for batch %s: %v", batchHeaderHash, err)
```

disperser/controller/dispatcher.go
```diff
@@ -381,64 +410,90 @@ func (d *Dispatcher) HandleSignatures(
...
 	err = d.blobMetadataStore.PutAttestation(ctx, attestation)
-	putAttestationFinished := time.Now()
-	d.metrics.reportPutAttestationLatency(putAttestationFinished.Sub(aggregateSignaturesFinished))
 	if err != nil {
-		dbErr := d.failBatch(ctx, batchData)
-		if dbErr != nil {
-			return fmt.Errorf("failed to update blob statuses for batch %s to failed: %w", batchHeaderHash, dbErr)
-		}
-		return fmt.Errorf("failed to put attestation for batch %s: %w", batchHeaderHash, err)
+		return fmt.Errorf("put attestation: %w", err)
```
cody-littley, line 462:
```suggestion
		return fmt.Errorf("error calling put attestation: %w", err)
```

disperser/controller/dispatcher.go
```diff
@@ -341,37 +349,58 @@ func (d *Dispatcher) HandleSignatures(
 	}
 	err := d.blobMetadataStore.PutAttestation(ctx, attestation)
 	if err != nil {
-		return fmt.Errorf("failed to put attestation for batch %s: %w", batchHeaderHash, err)
+		// this error isn't fatal: a subsequent PutAttestation attempt might succeed
+		// TODO: this used to cause the HandleSignatures method to fail entirely. Is it ok to continue trying here?
```
cody-littley, line 353: reminder to resolve this TODO

disperser/controller/dispatcher.go
```diff
@@ -341,37 +349,58 @@ func (d *Dispatcher) HandleSignatures(
...
+		d.logger.Error("put attestation",
```
cody-littley, line 354:
```suggestion
		d.logger.Error("failed to put attestation",
```

core/signature_receiver.go
```diff
@@ -0,0 +1,329 @@
...
+	quorumIDs          []QuorumID
+}
+
+// ReceiveSignatures receives SigningMessages over the signingMessageChan, and yields QuorumAttestations produced
+// from these SigningMessages.
+//
+// The yielded QuorumAttestations contain aggregate signing data from all SigningMessages received thus far. Each
+// QuorumAttestation will have incorporated more SigningMessages than the previously yielded QuorumAttestation.
+//
+// The caller should not manually close the returned QuorumAttestation chan. It will be closed automatically, when:
```
cody-littley, line 38: It's ok to leave this comment. Just pointing out that there is no requirement to close a golang channel when you are done using it. In general, channels are only closed when doing so sends some sort of signal you want to transmit (e.g. "we're done, go away now"). So even without this comment, in general I wouldn't think that closing this channel from the outer context would be something people would be likely to accidentally do.

core/signature_receiver.go
```diff
@@ -0,0 +1,329 @@
...
+		aggregateSignatures:       aggregateSignatures,
+		signerMap:                 signerMap,
+		aggregateSignersG2PubKeys: aggregateSignersG2PubKeys,
+		stakeSigned:               stakeSigned,
+		batchHeaderHash:           batchHeaderHash,
+		signingMessageChan:        signingMessageChan,
+		quorumIDs:                 sortedQuorumIDs,
+	}
+
+	attestationChan := make(chan *QuorumAttestation)
```
cody-littley, line 78: This creates a channel with capacity 0 (i.e. the goroutine that inserts a thing into the channel blocks until another goroutine removes the thing from the channel). If we know the maximum number of messages that will be passed, we can avoid the unnecessary blocking by setting the capacity equal to the maximum number of messages.

```suggestion
	attestationChan := make(chan *QuorumAttestation, len(signerMap))
```

core/signature_receiver.go
```diff
@@ -0,0 +1,329 @@
...
+	go receiver.receiveSigningMessages(ctx, attestationChan)
+
+	return attestationChan, nil
+}
+
+// receiveSigningMessages receives SigningMessages, and sends QuorumAttestations to the input attestationChan
+func (sr *signatureReceiver) receiveSigningMessages(ctx context.Context, attestationChan chan *QuorumAttestation) {
+	// this ticker causes QuorumAttestations to be sent periodically
+	ticker := time.NewTicker(1 * time.Second)
```
cody-littley, line 87: Can this be made into a setting? (default value of 1 second is fine)

api/clients/v2/payloaddispersal/payload_disperser.go
litt3: done

disperser/controller/dispatcher.go
litt3: done

disperser/controller/dispatcher.go
litt3: done

disperser/controller/dispatcher.go
litt3: I'm not sure the answer here yet. Let's leave this conversation open until I can get to the bottom it it

disperser/controller/dispatcher.go
litt3: done

core/signature_receiver.go
litt3: I simplified the comment, to avoid saying something redundant:

```
This channel will be closed when one of the following conditions is met:
```

core/signature_receiver.go
litt3: done

disperser/controller/dispatcher.go
samlaf: No need to revert as this is just a nit but overall @cody-littley I prefer not to encode redundant information in errors that are getting wrapped. This follows the uber convention: https://github.com/uber-go/guide/blob/master/style.md#error-wrapping

I'm totally onboard with adding the "error" prefix in logs though! Because those are not wrapped and are printed as is, so extra context is useful

litt3: I agree with not prefixing errors this way- I misread the line and thought this was a log when applying this change.

Reverted for consistency

core/aggregation.go
```diff
@@ -99,240 +85,37 @@ type SignatureAggregator interface {
...
-			}
-		}
-		a.Logger.Info("received signature from operator",
-			"operatorID", operatorIDHex,
-			"operatorAddress", operatorAddr,
-			"socket", socket,
-			"quorumIDs", fmt.Sprint(operatorQuorums),
-			"batchHeaderHash", batchHeaderHashHex,
-			"attestationLatencyMs", r.AttestationLatencyMs)
+	// tick interval is very long, since we don't actually need periodic attestation updates in this case. we only
```
samlaf, line 106: what does "this case" mean? V1?

disperser/controller/dispatcher.go
```diff
@@ -304,16 +309,20 @@ func (d *Dispatcher) HandleBatch(ctx context.Context) (chan core.SigningMessage,
 	return sigChan, batchData, nil
 }
 
-// HandleSignatures receives signatures from operators, validates, and aggregates them
+// HandleSignatures receives signatures from operators, validates, and aggregates them.
+//
+// This method submits Attestations to the blobMetadataStore, containing signing data from the SigningMessages received
+// through the sigChan. It periodically submits Attestations, as signatures are gathered.
 func (d *Dispatcher) HandleSignatures(
```
samlaf, line 316: whats the termination criterion for this function? does it just loop forever? until some timer ends? is it run in a goroutine?

disperser/controller/dispatcher.go
```diff
@@ -323,12 +332,15 @@ func (d *Dispatcher) HandleSignatures(
...
-			d.logger.Error("failed to update blob status for blob %s to gathering signatures",
+			d.logger.Error("failed to update blob status to 'gathering signatures'",
 				"blobKey", key.Hex(),
+				"batchHeaderHash", batchHeaderHash,
 				"err", err)
 		}
 	}
 
+	// submit an empty attestation before starting to gather signatures.
+	// a new attestation will be periodically resubmitted as signatures are gathered.
```
samlaf, line 343: curious why do you use the term "(re)submit" when talking about writing to dynamoDB? why not just say "updated" or "overwritten"? Is there some daemon process that manages the db that we "submit" write requests to?

disperser/controller/dispatcher.go
```diff
@@ -341,35 +353,57 @@ func (d *Dispatcher) HandleSignatures(
...
-	quorumResults := make(map[core.QuorumID]uint8)
-	for quorumID, quorumResult := range quorumAttestation.QuorumResults {
-		d.logger.Debug("quorum attestation results", "quorumID", quorumID, "result", quorumResult)
-		if quorumResult.PercentSigned > 0 {
-			nonZeroQuorums = append(nonZeroQuorums, quorumID)
-			quorumResults[quorumID] = quorumResult.PercentSigned
+	// keep track of the final attestation, since that's the attestation which will determine the final batch status
+	finalAttestation := &core.QuorumAttestation{}
+	// continue receiving attestations from the channel until it's closed
+	for receivedQuorumAttestation := range attestationChan {
```
samlaf, line 389: who/how/when is this channel closed?

core/signature_receiver.go
```diff
@@ -0,0 +1,365 @@
...
+					"attestationLatencyMs", signingMessage.AttestationLatencyMs)
+				continue
+			}
+
+			if seen := sr.signatureMessageReceived[signingMessage.Operator]; seen {
+				sr.logger.Warn("duplicate message from operator",
+					"batchHeaderHash", hex.EncodeToString(sr.batchHeaderHash[:]),
+					"operatorID", signingMessage.Operator.Hex(),
+					"attestationLatencyMs", signingMessage.AttestationLatencyMs)
+				continue
```
anupsv, line 150: do we not close the connection with that operator once they send the signature ?

core/signature_receiver.go
samlaf: Right also afaiu it's always (?) the sender's job to close the channel. It's actually very dangerous for a receiver to close a channel because a write on a closed channel panics, so unless the receiver has communicated with the sender through some other means and is damn sure that the sender won't close it (can't think of a situation where this would be the case, since if you've communicated with the sender, then might as well ask him to close the channel for you), then better to leave sender close channel.

core/signature_receiver.go
```diff
@@ -0,0 +1,365 @@
...
+		QuorumAggPubKey:  quorumAggPubKeyCopy,
+		SignersAggPubKey: aggregateSignersG2PubKeysCopy,
+		AggSignature:     aggregateSignaturesCopy,
+		QuorumResults:    quorumResults,
+		SignerMap:        validSignerMapCopy,
+	}
+}
+
+// computeQuorumResult creates a QuorumResult for a given quorum
+func (sr *signatureReceiver) computeQuorumResult(
```
anupsv, line 300: i assume none of the logic for verification was changed from before, just repurposed in a new function ?

samlaf (commented):
leaving a very superficial review, after having read through for ~30 mins. Logic overall seems good, but have not delved into specifics. I don't have enough context and would take me too much time to understand it all, so will leave it to others who have more context.

disperser/controller/dispatcher.go
```diff
@@ -323,12 +332,15 @@ func (d *Dispatcher) HandleSignatures(
...
 		if err != nil {
-			d.logger.Error("failed to update blob status for blob %s to gathering signatures",
+			d.logger.Error("failed to update blob status to 'gathering signatures'",
 				"blobKey", key.Hex(),
+				"batchHeaderHash", batchHeaderHash,
 				"err", err)
 		}
 	}
 
+	// submit an empty attestation before starting to gather signatures.
```
anupsv, line 342: Why submit an empty attestation ? is it just to show that nothing has been received yet?

disperser/controller/dispatcher.go
```diff
@@ -341,35 +353,57 @@ func (d *Dispatcher) HandleSignatures(
...
+			"err", err,
+			"batchHeaderHash", batchHeaderHash)
 	}
 
-	// TODO(ian-shim): periodically update the attestation with intermediate quorum results
-
-	quorumAttestation, err := d.aggregator.ReceiveSignatures(
-		ctx,
+	// This channel will remain open until the attestationTimeout triggers, or until signatures from all validators
+	// have been received and processed. It will periodically yield QuorumAttestations with the latest set of received
```
anupsv, line 364: received and processed ? I assume because they're part of the same function if not we could have quit just after receiving ?

core/aggregation.go
litt3: It occurred to me that it's easily doable in this PR to leave V1 functionality entirely unchanged, and to just reimplement with the necessary changes for V2.

I reverted changes that affect V1 functionality in 67d8e3c5aac5adb255ed65d42eb5364782e4408b, so the comment you're referring to is now gone entirely

disperser/controller/dispatcher.go
cody-littley: In that case, I'm ok if you don't prefix the error string (here as well as elsewhere in this PR)

cody-littley (approved):

disperser/controller/dispatcher.go
litt3: > Is there some daemon process that manages the db that we "submit" write requests to?

Nope, just a poor choice of term on my part. Fixed in afa7304f62d00a2e4689b56ca23e471237c27725

disperser/controller/dispatcher.go
litt3: This function doesn't loop, it's called once per batch. I reworked this doc in f3b5f28bf17e140835357144ea35be82e44d7e28 to improve clarity

```
// HandleSignatures receives SigningMessages from operators for a given batch through the input sigChan. The signatures
// are validated, aggregated, and used to put an Attestation for the batch into the blobMetadataStore. The Attestation
// is periodically updated as additional signatures are gathered.
//
// This method will continue gathering signatures until a SigningMessage has been received from every operator, or until
// the global attestationCtx times out.
```

core/signature_receiver.go
litt3: Correct, this is just a reorganization

core/signature_receiver.go
litt3: I believe you're correct, but I think we should leave this as a sanity check: If we remove it, then the `signature_receiver` would be dependent on the current dispatcher logic, and we could have a silent vulnerability if the `sigChan` logic ever changes

disperser/controller/dispatcher.go
litt3: I removed the empty attestation submission for now, because I'm pretty sure it isn't necessary.

Leaving this thread open until I can get confirmation.

disperser/controller/dispatcher.go
```diff
@@ -350,26 +352,7 @@ func (d *Dispatcher) HandleSignatures(
...
-	}
-	err := d.blobMetadataStore.PutAttestation(ctx, attestation)
-	if err != nil {
-		// this error isn't fatal: a subsequent PutAttestation attempt might succeed
-		// TODO: this used to cause the HandleSignatures method to fail entirely. Is it ok to continue trying here?
-		d.logger.Error("error calling PutAttestation",
-			"err", err,
-			"batchHeaderHash", batchHeaderHash)
-	}
+	// TODO: I removed the initial empty attestation update. Is that ok?
```
cody-littley, line 355: reminder to resolve TODO

cody-littley (approved):

disperser/controller/signature_receiver.go
```diff
@@ -0,0 +1,472 @@
...
+			if !found {
+				sr.logger.Warn("operator not found in state",
+					"batchHeaderHash", hex.EncodeToString(sr.batchHeaderHash[:]),
+					"operatorID", signingMessage.Operator.Hex(),
+					"attestationLatencyMs", signingMessage.AttestationLatencyMs)
+				continue
+			}
+
+			if seen := sr.signatureMessageReceived[signingMessage.Operator]; seen {
+				sr.logger.Warn("duplicate message from operator",
```
ian-shim, line 191: should this be Error?

disperser/controller/signature_receiver.go
```diff
@@ -0,0 +1,472 @@
...
+					"This shouldn't be possible.",
+					signingMessage.Operator.Hex(),
+					hex.EncodeToString(sr.batchHeaderHash[:]))
+			} else if sr.metrics != nil {
+				sr.metrics.reportSigningMessageChannelLatency(time.Since(signingMessage.TimeReceived))
+			}
+
+			indexedOperatorInfo, found := sr.indexedOperatorState.IndexedOperators[signingMessage.Operator]
+			if !found {
+				sr.logger.Warn("operator not found in state",
```
ian-shim, line 183: should this be Error?

disperser/controller/signature_receiver.go
```diff
@@ -0,0 +1,472 @@
...
+		}
+	}()
+
+	if signingMessage.Err != nil {
+		return fmt.Errorf("signingMessage contained error: %w", signingMessage.Err)
+	}
+
+	operatorPubkey := indexedOperatorInfo.PubkeyG2
+	if !signingMessage.Signature.Verify(operatorPubkey, sr.batchHeaderHash) {
+		return fmt.Errorf("signature verification with pubkey %s", hex.EncodeToString(operatorPubkey.Serialize()))
```
ian-shim, line 273: nit: signature verification "failed" with pubkey...?

disperser/controller/signature_receiver.go
```diff
@@ -0,0 +1,472 @@
...
+		}
+	}
+
+	if sr.aggregateSignersG2PubKeys[quorumID] == nil {
+		return nil, errors.New("nil aggregate signer G2 public key")
+	}
+
+	ok, err := aggregateSignersG1PubKey.VerifyEquivalence(sr.aggregateSignersG2PubKeys[quorumID])
+	if err != nil {
+		return nil, fmt.Errorf("verify aggregate G1 and G2 pubkey equivalence: %w", err)
```
ian-shim, line 398: nit: failed to verify...

disperser/controller/signature_receiver.go
```diff
@@ -0,0 +1,472 @@
...
+	quorumIDs []core.QuorumID
+
+	// tickInterval determines how frequently intermediate attestations are yielded
+	tickInterval time.Duration
+
+	// attestationUpdateStart is initialized when we first start receiving signatures, and is updated each time an
+	// attestation is yielded. This is used to track how long it takes to yield each attestation.
+	attestationUpdateStart time.Time
+
+	// significantSigningThresholdPercentage is a configurable "important" signing threshold. Right now, it's being
```
ian-shim, line 53: what does "important" and "special handling" mean?
looks like its there to record time at which the threshold is met. how is that timestamp used?

ian-shim (commented):
looks good to me! few questions

disperser/controller/signature_receiver.go
litt3: @samlaf has been driving an effort to shorten error messages by omitting prefixed like "failed to". See [this conversation above](https://github.com/Layr-Labs/eigenda/pull/1497#discussion_r2065533785)

disperser/controller/signature_receiver.go
litt3: "Important" for now means "whatever the default signing threshold is", so the current default is `55`.

Special handling in the future *could* be something like immediately emitting an attestation once quorums have reached the configured threshold. This is an optimization that isn't necessary now, but could shave off a couple hundred milliseconds of dispersal latency in the future, once we've improved performance enough for this delta to be significant.

Right now, this value is only being used to understand the profile of the signature gathering phase. i.e. `how long after crossing the 'standard' signing threshold do we keep on gathering signatures?`

disperser/controller/signature_receiver.go
litt3: See [above](https://github.com/Layr-Labs/eigenda/pull/1497#discussion_r2070265324)

disperser/controller/dispatcher.go
litt3: removed

cody-littley (approved):

ian-shim (approved):
lgtm, please resolve the ci failures

disperser/controller/dispatcher.go
```diff
@@ -323,53 +345,61 @@ func (d *Dispatcher) HandleSignatures(
...
-		QuorumNumbers:    nil,
-		QuorumResults:    nil,
-	}
-	err := d.blobMetadataStore.PutAttestation(ctx, attestation)
-	if err != nil {
-		return fmt.Errorf("failed to put attestation for batch %s: %w", batchHeaderHash, err)
-	}
-
-	// TODO(ian-shim): periodically update the attestation with intermediate quorum results
+	// TODO: I removed the initial empty attestation update. Is that ok?
```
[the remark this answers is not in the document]
litt3: After struggling with CI for a while, I decided to just keep the empty attestation `put` for now, to unblock this PR.

cody-littley (approved):

PR #1504, opened 2025-04-25
feat: create a calculator for eigenda requirements

cody-littley:

Create a calculator for basic EigenDA requirements.

### DIFF
--- a/tools/calculator/calculator.html
+++ b/tools/calculator/calculator.html
@@ -1,0 +1,422 @@
+<!DOCTYPE html>
+<html lang="en">
+<head>
+    <meta charset="UTF-8">
+    <meta name="viewport" content="width=device-width, initial-scale=1.0">
+    <title>EigenDA Resource Calculator</title>
+    <style>
+        body {
+            font-family: Arial, sans-serif;
+            max-width: 800px;
+            margin: 0 auto;
+            padding: 20px;
+            line-height: 1.6;
+        }
+        h1 {
+            color: #333;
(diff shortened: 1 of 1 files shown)


tools/calculator/calculator.html
```diff
@@ -0,0 +1,422 @@
...
+                <span class="tooltip-text">The maximum throughput supported by EigenDA.</span>
+            </div>
+        </div>
+        
+        <div class="input-group">
+            <label for="sumOfStake">Sum of Stake Across All Quorums:</label>
+            <input type="number" id="sumOfStake" value="0.125" step="0.001" min="0" max="1">
+            <div class="tooltip">
+                <div class="tooltip-icon">?</div>
+                <span class="tooltip-text">The sum of an operator's stake percentage across all quorums. For example, if an operator has 10% stake in 3 quorums, the sum of stake across all quorums is 30%.</span>
```
litt3, line 145: This mentions percentage, but the field actually accepts a fraction (0–1)

litt3 (approved):

tools/calculator/calculator.html
cody-littley:
```
<span class="tooltip-text">The sum of an operator's stake across all quorums. 1.0 means 100% stake for one quorum, 2.0 means 100% stake across two quorums. For example, if an operator has 10% stake in 3 quorums, the sum of stake across all quorums is 30%, and so you'd use 0.3 for field.</span>
```

PR #1502, opened 2025-04-25
ci: only run buf-proto on protobuf file changes

samlaf:
## Why are these changes needed?

This CI is running on every single PR right now, even those that don't change proto, and the buf bot publishes a message which just clutters PRs:
image
https://github.com/Layr-Labs/eigenda/pull/1493#issuecomment-2825440837

Changed such that it only runs on changes to protobuf files, or to the buf.yaml config file itself.

## Checks

- [ ] I've made sure the tests are passing. Note that there might be a few flaky tests, in that case, please comment that they are not relevant.
- [ ] I've checked the new test coverage and the coverage percentage didn't drop.
- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests
   - [ ] This PR is not tested :(

PR #1505, opened 2025-04-28
chore: move encoder protos under api dir

samlaf:

Not sure why we had these protos in another dir. See discussion started in https://github.com/Layr-Labs/eigenda/pull/1502#issuecomment-2833619633
Consolidated encoder protos under api/proto like all other protos.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


dmanc (approved):

jianoaix:
encoder is not a user-facing component. Just being a proto or using gRPC doesn't mean it should be in api folder, which is intended for users.

samlaf:
> encoder is not a user-facing component. Just being a proto or using gRPC doesn't mean it should be in api folder, which is intended for users.

I see the intention now. We should def remove the encoder from the generated documentation. We can always revert this PR, but I still think having them in disparate places is not the right way. Also encoder being alone under disperser felt weird. How do I know there’s not more random places in the repo where other protos live? 

I think they should all be under the same dir personally. Maybe we can do something like this:
image

PR #1529, opened 2025-05-06
feat: validator node authenticate and validate blob headers

hopeyen:

https://linear.app/eigenlabs/issue/EGDA-1195/validator-node-checks-individual-dispersal-requests

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/node/grpc/server_v2.go
+++ b/node/grpc/server_v2.go
@@ -148,9 +148,12 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
 		}
 	}
 	if s.blobAuthenticator != nil {
+		// TODO: check the latency of request validation later; could be parallelized to avoid significant
+		// impact to the request latency
 		for _, blob := range batch.BlobCertificates {
 			_, err = s.validateDispersalRequest(blob)
 			if err != nil {
+				// TODO: Blacklist the disperser if there's an invalid dispersal request
 				return nil, api.NewErrorInvalidArg(fmt.Sprintf("failed to validate blob request: %v", err))
 			}
 		}
@@ -387,6 +390,7 @@ func (s *ServerV2) GetChunks(ctx context.Context, in *pb.GetChunksRequest) (*pb.
 // - no encoding prover GetCommitmentsForPaddedLength check
 // - directly take blob lengths (no blob data yet)
 // - doesn't check every 32 bytes is a valid field element
+// Node cannot make these checks because the checks require the blob data
 func (s *ServerV2) validateDispersalRequest(
 	blobCert *corev2.BlobCertificate,
 ) (*corev2.BlobHeader, error) {
@@ -398,7 +402,7 @@ func (s *ServerV2) validateDispersalRequest(
 		return nil, fmt.Errorf("failed to authenticate blob request: %v", err)
 	}
 
-	//this is the length in SYMBOLS (32 byte field elements) of the blob. it must be a power of 2
+	// this is the length in SYMBOLS (32 byte field elements) of the blob. it must be a power of 2
 	commitedBlobLength := blobCert.BlobHeader.BlobCommitments.Length
 	if commitedBlobLength == 0 {
 		return nil, errors.New("blob size must be greater than 0")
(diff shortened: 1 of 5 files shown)


node/grpc/server_v2.go
```diff
@@ -367,3 +380,62 @@ func (s *ServerV2) GetChunks(ctx context.Context, in *pb.GetChunksRequest) (*pb.
...
 	}, nil
 }
+
+// validateDispersalRequest validates the DisperseBlobRequest and returns the blob header
+// Differences between this and the DispersalServerV2 are:
+// - Takes *corev2.BlobCertificate instead of DisperseBlobRequest
+// - no encoding prover GetCommitmentsForPaddedLength check
+// - directly take blob lengths (no blob data yet)
+// - doesn't check every 32 bytes is a valid field element
+func (s *ServerV2) validateDispersalRequest(
```
jianoaix, line 390: Interesting we didn't validate the blob certificates from the request before

node/grpc/server_v2.go
```diff
@@ -367,3 +380,62 @@ func (s *ServerV2) GetChunks(ctx context.Context, in *pb.GetChunksRequest) (*pb.
...
+) (*corev2.BlobHeader, error) {
+	if len(blobCert.Signature) != 65 {
+		return nil, fmt.Errorf("signature is expected to be 65 bytes, but got %d bytes", len(blobCert.Signature))
+	}
+	err := s.blobAuthenticator.AuthenticateBlobRequest(blobCert.BlobHeader, blobCert.Signature)
+	if err != nil {
+		return nil, fmt.Errorf("failed to authenticate blob request: %v", err)
+	}
+
+	//this is the length in SYMBOLS (32 byte field elements) of the blob. it must be a power of 2
```
jianoaix, line 401: nit: add a space

node/grpc/server_v2.go
```diff
@@ -367,3 +380,62 @@ func (s *ServerV2) GetChunks(ctx context.Context, in *pb.GetChunksRequest) (*pb.
 		ChunkEncodingFormat: pb.ChunkEncodingFormat_GNARK,
 	}, nil
 }
+
+// validateDispersalRequest validates the DisperseBlobRequest and returns the blob header
+// Differences between this and the DispersalServerV2 are:
+// - Takes *corev2.BlobCertificate instead of DisperseBlobRequest
+// - no encoding prover GetCommitmentsForPaddedLength check
+// - directly take blob lengths (no blob data yet)
+// - doesn't check every 32 bytes is a valid field element
```
jianoaix, line 389: Is it useful to check these items at validator here as well?

node/grpc/server_v2.go
```diff
@@ -141,7 +147,14 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
 			return nil, api.NewErrorInvalidArg(fmt.Sprintf("failed to verify request: %v", err))
 		}
 	}
-
+	if s.blobAuthenticator != nil {
+		for _, blob := range batch.BlobCertificates {
```
jianoaix, line 151: It may be worth checking the latency of request validation later, and this part could be parallelized if it takes a meaningful amount of time (the time spent here is contributing directly to the overall request latency)

jianoaix (approved):
It seems this is mostly an enhancement of the validator request validation, is it for detecting noncompliant dispersers later?

node/grpc/server_v2.go
hopeyen: It would be, but not possible, because those checks are done to the blob directly while we shouldn't make validators validate request against the blob itself

node/grpc/server_v2.go
hopeyen: yeah I think we just trusted the disperser for those checks:p

hopeyen:
> It seems this is mostly an enhancement of the validator request validation, is it for detecting noncompliant dispersers later?

yes! in theory nodes could blacklist a disperser as soon as a blob fails here. I think we can later utilize these for slashing dispersers

node/grpc/server_v2.go
jianoaix: Where is it checked?
It's generally better to check at earlier stage if there's sufficient information so we can reject invalid requests fast and cheaply.

hopeyen: It isn't checked by the validators. I don't think it is possible because it requires the blob data itself:(

anupsv (approved):

PR #1530, opened 2025-05-06
fix: Quorum threshold checking

litt3:
## Bug Description

- The previous code was asserting that the number of quorums defined in the blob header must match the number of quorums in the batch
    - This is not a correct assertion, since the batch may include more quorums than the blob
- The previous code was also asserting that every quorum in the batch meets the defined threshold, but we should only assert that the quorums in the blob header meet the threshold

## Fix

I removed the incorrect assertion, and fixed the reporting for unmet thresholds


cody-littley (approved):

anupsv (commented):
Lgtm

PR #1533, opened 2025-05-07
docs(mdbook): update secure integration RBN recency validation explanation

samlaf:

Our Secure Integration page was out of date on a few things. Updated the links and validation rules that have changed (for example we no longer check relayURL in verifyEigenDACertV2).

Most importantly, the cert->batcher-inbox timing check (which we've so far referred to as Punctuality check) was super under-specified. Thinking through its requirement more made me realize that "punctuality" checking was not actually needed, because that is already enforced by (most? all?) rollup frameworks. Instead, what this check does of importance is enforce that the RBN is not too far back in the past, so I have opted to rename `punctuality check` to `RBN recency check`.

I don't think I have fully figured out this design, so opening this PR as a discussion starter.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,29 +1,44 @@
...
 
-The validation process is:
+### 1. RBN Recency Validation
 
-1. Validate the Cert (against state on ethereum)
-2. Validate the Blob (against the Cert)
+First, we assume that every rollup stack has a mechanism to force blobs to be included in the rollup’s batcher-inbox in a timely matter, such that the cert's blob is still available on EigenDA for long enough to cover the fraud proof duration (typically 7 days), or long enough to be downloadable by zk rollup participants.
 
-### Cert Validation
+> For example, optimism has a [sequencerWindow](https://docs.optimism.io/stack/rollup/derivation-pipeline#sequencer-window) which forces batches to land onchain in a timely fashion (12h). This indirectly also forces EigenDA certs to land onchain in a timely fashion. Therefore, it is safe to assume that the certs land onchain < SequencerWindowSize after having been signed.
```
samlaf, line 17: Do most rollups have something like sequencerWindow?
- @ethenotethan what about arb?
- @anupsv what about zksync?

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,29 +1,44 @@
...
 
-Cert validation is done inside the EigenDACertVerifier contract, which EigenDA deploys as-is, but is also available for rollups to modify and deploy on their own. Specifically, [verifyDACertV2](https://github.com/Layr-Labs/eigenda/blob/98e21397e3471d170f3131549cdbc7113c0cdfaf/contracts/src/core/EigenDACertVerifier.sol#L86) is the entry point for validation. This could either be called during a normal eth transaction (either for pessimistic “bridging” like EigenDA V1 used to do, or when uploading a Blob Field Element to a one-step-proof’s [preimage contract](https://specs.optimism.io/fault-proof/index.html#pre-image-oracle)), or be zk proven using a library like [Steel](https://github.com/risc0/risc0-ethereum/blob/main/crates/steel/docs/what-is-steel.md).
+This guarantee alone however is not enough, because the EigenDA disperser could have chosen a reference block number (RBN) that is very old, such that the stake of operators does not represent the true current value. Because EigenDA V2 doesn't pessimistically bridge certs onchain like EigenDA V1, enforcing RBN to not be too old is thus the responsibility of the rollup.
```
samlaf, line 21: Could EigenDA validators enforce this check? Might not be possible given we don't have our own consensus for validators to agree on time. cc @bxue-l2 @mooselumph

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,29 +1,44 @@
...
 
-The [cert verification](https://github.com/Layr-Labs/eigenda/blob/master/contracts/src/libraries/EigenDABlobVerificationUtils.sol#L162) logic consists of:
+> Rollups must enforce that `cert.RBN` > `cert.L1InclusionBlock` - `RecencyWindowSize`
 
-1. [merkleInclusion](https://github.com/Layr-Labs/eigenda/blob/c3d1ff2a9a1ec39fc78c4936eabbc2443df38571/contracts/src/libraries/EigenDABlobVerificationUtils.sol#L173): 
+We suggest using the same value as the `SequencerWindowSize` for the `RecencyWindowSize`, namely 12h.
```
litt3, line 25: Hmm, are we sure about this?

When the disperser chooses an `RBN` for a batch, it intentionally chooses one that is slightly in the past (see the [finalization delay](https://github.com/Layr-Labs/eigenda/blob/fc9a97144585186f25bcdc77134d02a6de9d53f0/disperser/controller/dispatcher.go#L32))

If `RecencyWindowSize == SequencerWindowSize`, in a practical sense you're decreasing the `SequencerWindowSize` by the disperser's `finalization delay`, since a batch created at `now - SequencerWindowSize + 1` will be invalid, even though it's a perfectly normal batch, and was submitted within the required sequencer window.

Would it make more sense for `RecencyWindowSize` to be some moderate amount > `SequencerWindowSize`?

docs/spec/src/integration/spec/6-secure-integration.md
samlaf: Just asked Dima for zksync, they have something similar. Would need to ask for links to docs where this is written down, but basically once a batch is "sealed", it has only 3 days to be submitted to L1, otherwise the proof is invalid. Hence, iiuc, they have 3 days to send to EigenDA, and submit the cert to L1.

samlaf: I don't think so? The way I understood it is they have to seal before they send to DA, and once its sealed they have 3 days to post to L1.

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: This is actually fine, in the worst case, this is a liveness problem

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: So rollup recency issue now is coupled with eigenda punctuality check. For example, they need to understand the recency guaranteed by EigenDA. If the sequencing window are changed to be smaller than what DA validator guaranteed

samlaf: > If the sequencing window are changed to be smaller than what DA validator guaranteed
@bxue-l2 you mean longer? Yes indeed. But SW of 14 days seems absurd.

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: I am not convinced at that high level, it achieves what we want. Consider the following scenario

A batch (B) is created with a RBN at T1, and suppose DA nodes verify the RBN is within 2 hours, when it was signed, say it is T2=T1+2h. Now the malicious batcher can at most hold it for an additional 12h, otherwise the derivation pipeline restarts. 

But say the batch (B) is just some gibberish data. The malicious batcher dispersed and held it for 14days. Then it was submitted to the L1 maybe interleaved it among legitimate EigenDA certs.

The Derivation pipeline works from l1 traversal -> l1 retrieval -> altda -> frame queue -> channel bank -> batch queue (where sequence_window_size is checked). Now since altda isn’t available it could get to the batch queue where it can invalidate the batch

bxue-l2: https://specs.optimism.io/protocol/derivation.html#batch-queue

samlaf: Good callout! So basically the idea is that we don't want malicious crated certs to stall the pipeline. Or actually any garbage posted to batcher_inbox:
1. If batcher posts 0x010100 with no cert, or jibberish after, then our CertVerifier call will return an error and we'll get rid of it right?
2. But if CertVerifier passes, RBN could still be super old, meaning the data is no longer available for download, so your point is we def don't want to move on to atttempt to try to retrieve the blob otherwise we'll just stall the pipeline.

So we do need to validate the RBN in the L1Retrieval stage. Kinda weird that RBN is just a "proxy" to "when the blob is actually made available":
RBN -------- blob_made_available -------- blob_no_longer_available (validator dependent) ---->

bxue-l2: 1. yes, it should. But I think proxy currently will just stall. But once we have the error code, the deriviation pipeline can move on.
2. in hokulea, we will check punctuality before verifying the blob, or even before retrieving it.

Though, why would a batcher want to stall its liveness for 12h. So practically, it makes system more robust.

But more important, if someone submit a wrong l2 root, no one will be able to challenge it

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,29 +1,44 @@
...
+### 1. RBN Recency Validation
 
-1. Validate the Cert (against state on ethereum)
-2. Validate the Blob (against the Cert)
+First, we assume that every rollup stack has a mechanism to force blobs to be included in the rollup’s batcher-inbox in a timely matter, such that the cert's blob is still available on EigenDA for long enough to cover the fraud proof duration (typically 7 days), or long enough to be downloadable by zk rollup participants.
 
-### Cert Validation
+> For example, optimism has a [sequencerWindow](https://docs.optimism.io/stack/rollup/derivation-pipeline#sequencer-window) which forces batches to land onchain in a timely fashion (12h). This indirectly also forces EigenDA certs to land onchain in a timely fashion. Therefore, it is safe to assume that the certs land onchain < SequencerWindowSize after having been signed.
+>
+> OP stack rollups thus need to make sure that `SequencerWindow` << 14 days.
```
ethenotethan, line 19: Why are we setting the `RecencyWindow=SequencerWindow`? iiuc there's a liveness x safety tradeoff for the value assumed for the `RecencyWindow` where increasing the window increases liveness while decreasing safety. Ie, for optimistic rollups you'd need the data available for 14 days for ensuring successful fraud proofs by an honest party. This creates an attack zone boundary starting at `MaxChallengeWindow + MaxResolveWindow - RecencyWindow` and ending at `MaxChallengeWindow + MaxResolveWindow`. Curious how 12 hours is the optimal expression given we could put it down to e.g, 30-minutes to one hour while still ensuring liveness at the benefit of greater safety

samlaf: TLDR is just that I don't think it matters that much, and this way we don't need to introduce a new variable in the rollup.json config. So we keep our fork minimal. Also RecencyWindow and SequencerWindow are related, I think RecencyWindow needs to be <= SequencerWindow (since anyways if you wait longer the batches contained inside the blob will get discarded by the batch_queue stage as bowen pointed out in https://github.com/Layr-Labs/eigenda/pull/1533#discussion_r2078526064

docs/spec/src/integration/spec/6-secure-integration.md
samlaf: But so... seems like we need to make `RecencyWindow = SequencerWindowSize + how_far_back_disperser_can_set_RBN`... and even if we do that there's nothing stopping disperser from changing its behavior and setting RBN to be `RecencyWindowSize` in the past...

Actually I think we can solve this by changing the validity equation to be:
```
cert.ReferenceBlockNumber < cert.L1InclusionBlock <= max(SequencerWindowSize, cert.ReferenceBlockNumber + eigenDAInclusionWindowSize)
```
WDYT?

litt3: Couple questions:

- When you say `eigenDAInclusionWindowSize`, do you actually mean `recencyWindowSize`?
- The first param of your `max` function is `SequencerWindowSize`, but I assume you meant to add something to it?

samlaf: Yes and yes. Sorry copy-pasted from somewhere else and was careless. updated with correct variable names.

litt3 (approved):

docs/spec/src/integration/spec/6-secure-integration.md
samlaf: Ok so think the overall TLDR of this discussion we had last week was "yes it makes sequencerWindowSize a bit shorter, but who cares". If you're happy with this feel free to resolve this discussion.

docs/spec/src/integration/spec/6-secure-integration.md
samlaf: Updated doc with these details: [011e807](https://github.com/Layr-Labs/eigenda/pull/1533/commits/011e8077cd1b08156adc83f0173dcea73c5ff2b0)

bxue-l2:
> in order to be downloadable if needed during a rollup fraud proof window.

We should be more general including zk fault proof and zk validity proof. "Challenge and resolution window." roughly capture the idea

litt3 (commented):
LGTM

litt3 (approved):

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -12,17 +12,24 @@ When deriving a rollup chain by running its derivation pipeline, only EigenDA ce
...
 
-We suggest using the same value as the `SequencerWindowSize` for the `RecencyWindowSize`, namely 12h.
+> To give a concrete example with a rollup stack, optimism has a [sequencerWindow](https://docs.optimism.io/stack/rollup/derivation-pipeline#sequencer-window) which forces batches to land onchain in a timely fashion (12h). This filtering however, happens in the [BatchQueue](https://specs.optimism.io/protocol/derivation.html#batch-queue) stage of the derivation pipeline (DP), and doesn't prevent the DP being stalled in the [L1Retrieval](https://specs.optimism.io/protocol/derivation.html#l1-retrieval) stage by an old cert having been submitted whose blob is no longer available on EigenDA. To prevent this, we need the recencyWindow filtering to happen during the L1Retrieval stage of the DP.
+>
+> Despite its semantics being slightly different, sequencerWindow and recencyWindow are related concepts, and in order to not force another config change on op altda forks, we suggest using the same value as the `SequencerWindowSize` for the `RecencyWindowSize`, namely 12h.
```
bxue-l2, line 32: In the diagram below, the two time scale are pretty confusing, it looks like 12h contains 3 l1 block and 24 seconds also contain 3 l1 blocks

bxue-l2: Maybe use dashed line

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -12,17 +12,24 @@ When deriving a rollup chain by running its derivation pipeline, only EigenDA ce
...
->
-> OP stack rollups thus need to make sure that `SequencerWindow` << 14 days.
+![](../../assets/integration/recency-window-timeline.png)
+
+Looking at the timing diagram above, in order for the blobs to be available during the entire challenge period, we need the fraud proof game to start < 7days after the cert is posted to the rollup's batcher inbox. In order to uphold this guarantee, what we need to do is simply to have rollups' derivation pipelines reject certs whose DA availability period started a long time ago. However, from the cert itself, there is no way to know when the cert was signed and made available. The only information available on the cert itself is cert.RBN, the reference block number chosen by the disperser at which to anchor operator stakes. But that happens to be before validators sign, so it is enough to bound how far that can be from the cert's inclusion block.
```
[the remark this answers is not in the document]
samlaf: [d8fe518](https://github.com/Layr-Labs/eigenda/pull/1533/commits/d8fe518c14cf201634990ba0d282b54d6508a927)

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -12,17 +12,24 @@ When deriving a rollup chain by running its derivation pipeline, only EigenDA ce
 
 ### 1. RBN Recency Validation
 
-First, we assume that every rollup stack has a mechanism to force blobs to be included in the rollup’s batcher-inbox in a timely matter, such that the cert's blob is still available on EigenDA for long enough to cover the fraud proof duration (typically 7 days), or long enough to be downloadable by zk rollup participants.
+This check is related to time guarantees. Timing is especially important for optimistic rollups where 
+the blobs need to be available during an interactive fraud proof window in order to
```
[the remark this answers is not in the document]
samlaf: updated in [181f1a0](https://github.com/Layr-Labs/eigenda/pull/1533/commits/181f1a02bacc486d90b9fc53d5320dd27053dacb) and [1f43c1c](https://github.com/Layr-Labs/eigenda/pull/1533/commits/1f43c1c813a0ec5d833a6650a9d160af0ffbb2eb)

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -12,11 +12,13 @@ When deriving a rollup chain by running its derivation pipeline, only EigenDA ce
...
+Looking at the timing diagram above, we need the EigenDA availability period to overlap the ~7days challenge and/or resolution period. In order to uphold this guarantee, what we need to do is simply to have rollups' derivation pipelines reject certs whose DA availability period started a long time ago. However, from the cert itself, there is no way to know when the cert was signed and made available. The only information available on the cert itself is `cert.RBN`, the reference block number chosen by the disperser at which to anchor operator stakes. But that happens to be before validators sign, so it is enough to bound how far that can be from the cert's inclusion block.
```
bxue-l2, line 21: why and/or 

overlap the worst case ~7days challenge window.

docs/spec/src/integration/spec/6-secure-integration.md
samlaf: oops real fix: [8701e98](https://github.com/Layr-Labs/eigenda/pull/1533/commits/8701e989a171700bc3e427c9bfdad24bdd3fb4da)

PR #1536, opened 2025-05-07
feat: reserved-only disperser apiserver

hopeyen:

Users running API server automatically only serve reserved usage, as encouraged. EigenLabs disperser would take care of setting the flag to false. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


disperser/cmd/apiserver/flags/flags.go
```diff
@@ -189,6 +189,13 @@ var (
 		Value:    5 * time.Minute,
 		EnvVar:   common.PrefixEnvVar(envVarPrefix, "NTP_SYNC_INTERVAL"),
 	}
+	ReservedOnly = cli.BoolTFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "reserved-only"),
+		Usage:    "if true, only reserved dispersal requests are served; on-demand requests are rejected (default: true)",
+		Required: false,
+		EnvVar:   common.PrefixEnvVar(envVarPrefix, "RESERVED_ONLY"),
+		Hidden:   false,
+	}
```
anupsv, line 198: should probably set the default to false ?

anupsv (approved):

disperser/cmd/apiserver/flags/flags.go
hopeyen: I think default being true is better, for all the permissionless dispersers. 
Reserved AND on-demand is only for specific dispersers, like EigenLabsDisperser. Since we have more context into the configs, I figured we will be mindful in setting this to false to allow for on-demand payments.

disperser/apiserver/disperse_blob_v2.go
```diff
@@ -178,6 +178,10 @@ func (s *DispersalServerV2) validateDispersalRequest(
 		return nil, errors.New("payment metadata is required")
 	}
 
+	if s.ReservedOnly && blobHeader.PaymentMetadata.CumulativePayment.Sign() != 0 {
```
jianoaix, line 181: Do you need to check that the reservation_period (or timestamp lately) is not set?
Otherwise, it'll be treated as reserved, even though cumulative_payment is also set (reserved takes precedence).

disperser/cmd/apiserver/flags/flags.go
jianoaix: What if the 3P dispersers set it to false? Will anything break?

disperser/apiserver/disperse_blob_v2.go
hopeyen: I don't think the check is needed, because I've been thinking adding timestamp to ondemand payments as well for smarter global ondemand ratelimiting, then the only clear separation between ondemand and reservation is `cumulative_payment`. This is also currently how meterer determines the type of requests

disperser/cmd/apiserver/flags/flags.go
hopeyen: Nope. if 3P dispersers set it to false and attempt to send an on-demand request, validators should reject due to the change in #1539

disperser/apiserver/disperse_blob_v2.go
jianoaix: It doesn't match what the API says: https://github.com/Layr-Labs/eigenda/blob/master/api/proto/common/v2/common_v2.proto#L66-L73

hopeyen: ooof thanks for pointing it out. The API docs need updating

jianoaix: Can you update it to make them consistent?

jianoaix (approved):

disperser/apiserver/disperse_blob_v2.go
hopeyen: yeah, I overhauled the section this afternoon, pls lmk if it looks better https://github.com/Layr-Labs/eigenda/pull/1570

PR #1539, opened 2025-05-08
feat: validator authenticate on-demand request by disperser key

hopeyen:

Validators should only accept ondemand requests from the EigenLabs disperser. 
The key has already been authenticated at the step of the check. 
If batch request contains any blob with ondemand payment but the batch didn't come from EigenLabs disperser, then validator rejects. 
- this check will be removed when we support decentralized ondemand usage

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


node/grpc/server_v2.go
```diff
@@ -140,6 +138,15 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
...
 		}
+		// TODO: move to blob authenticator later
+		for _, blob := range batch.BlobCertificates {
+			if meterer.IsOnDemandPayment(&blob.BlobHeader.PaymentMetadata) {
+				// Batch contains on-demand payments, so the chunk must be from EigenLabsDisperser
+				if in.DisperserID != api.EigenLabsDisperserID {
+					return nil, api.NewErrorInvalidArg("on-demand payments are only allowed for EigenLabsDisperser")
+				}
+			}
+		}
```
anupsv, line 149: Before this, do we check that the ondemand blobs we're from EigenLabsDisperser by checking the signature or something ?

anupsv (dismissed):

node/grpc/server_v2.go
hopeyen: yep! in `AuthenticateStoreChunksRequest` on L131 we authenticate the signature first to make sure it is a registered disperser

core/meterer/meterer.go
```diff
@@ -277,3 +277,7 @@ func (m *Meterer) IncrementGlobalBinUsage(ctx context.Context, symbolsCharged ui
 func (m *Meterer) GetReservationBinLimit(reservation *core.ReservedPayment, reservationWindow uint64) uint64 {
 	return reservation.SymbolsPerSecond * reservationWindow
 }
+
+func IsOnDemandPayment(paymentMetadata *core.PaymentMetadata) bool {
+	return paymentMetadata.CumulativePayment.Cmp(big.NewInt(0)) > 0
```
jianoaix, line 282: The reservation_period should be unset as well?

node/grpc/server_v2.go
```diff
@@ -140,6 +138,15 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
 		if err != nil {
 			return nil, api.NewErrorInvalidArg(fmt.Sprintf("failed to verify request: %v", err))
 		}
+		// TODO: move to blob authenticator later
+		for _, blob := range batch.BlobCertificates {
+			if meterer.IsOnDemandPayment(&blob.BlobHeader.PaymentMetadata) {
+				// Batch contains on-demand payments, so the chunk must be from EigenLabsDisperser
+				if in.DisperserID != api.EigenLabsDisperserID {
+					return nil, api.NewErrorInvalidArg("on-demand payments are only allowed for EigenLabsDisperser")
```
jianoaix, line 146: would be helpful to pinpoint the blob that's on-demand but via a non-eigen disperser in error log

hopeyen: yes! would be. I will add

disperser/apiserver/server_v2.go
```diff
@@ -170,6 +170,7 @@ func (s *DispersalServerV2) Start(ctx context.Context) error {
 		for {
 			select {
 			case <-ticker.C:
+				s.logger.Debug("Refreshing onchain quorum state", "onchainState", s.onchainState.Load())
```
ian-shim, line 173: nit: Is this intentionally logging the state prior to the refresh? A bit confusing to me as I'd interpret this as the state after the refresh if this showed up in the logs

node/grpc/server_v2.go
```diff
@@ -145,6 +143,15 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
 		if err != nil {
 			return nil, api.NewErrorInvalidArg(fmt.Sprintf("failed to verify request: %v", err))
 		}
+		// TODO: move to blob authenticator later
+		for _, blob := range batch.BlobCertificates {
```
ian-shim, line 147: Should metering logic be gated behind the permissionless disperser feature flag?

ian-shim:
Need to resolve CI

node/grpc/server_v2.go
hopeyen: There's no metering logic, just a filter for whether the node should serve the ondemand request from a certain disperser. If the disperser is not EigenLabs, it will get rejected regardless of whether there's permissionless dispersers or not, so I didn't think this block should be gated by other flags

disperser/apiserver/server_v2.go
hopeyen: Ah, makes sense. moved down

node/grpc/server_v2.go
```diff
@@ -145,6 +143,15 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
...
+			if meterer.IsOnDemandPayment(&blob.BlobHeader.PaymentMetadata) {
+				// Batch contains on-demand payments, so the chunk must be from EigenLabsDisperser
+				if in.DisperserID != api.EigenLabsDisperserID {
+					return nil, api.NewErrorInvalidArg(fmt.Sprintf("on-demand payments are only allowed for EigenLabsDisperser; receiving disperser ID: %d", in.DisperserID))
```
dmanc: Should this be a forbidden error?
  - PermissionDenied (Code 7) → HTTP 403: "the caller does not have permission to execute the specified operation"

hopeyen: makes sense, [updated](https://github.com/Layr-Labs/eigenda/pull/1539/commits/d77c666e3afeff91e68623d3de498d1d66a6f033)!

core/meterer/meterer.go
anupsv: So on demand payment will not have cumulative payment ?

node/grpc/server_v2.go
anupsv: Arguably, we can move this check in the `AuthenticateStoreChunksRequest` right ? Cause it has the registered disperser and signature check.

dmanc (approved):

PR #1554, opened 2025-05-11
feat: experimental WASM build target for relay

pschork:
These changes allow us to compile the relay to WASM

Cloudflare Zero-Fee Egress would require hosting the relay as a [CloudFlare Worker](https://workers.cloudflare.com/) via WASM (go is not supported)

WASM is itself a sandbox and does not allow for OS-level or related system calls. This means Inabox/dockertest is not supported. See annotations to `chunkstore/chunk_store_test.go` and `relay_test_utils.go` to exclude them from WASM build target.

Running `relay.wasm` on terminal requires a nodejs wrapper defined under `wasm/run_wasm.js`. Hosted WASM services (ie on Cloudflare) would not require this as they provide their own WASM runtime.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/relay/Makefile
+++ b/relay/Makefile
@@ -9,10 +9,16 @@ build: clean
 	go mod tidy
 	go build -o ./bin/relay ./cmd
 
-build-wasm: clean
-	go mod tidy
-	GOOS=js GOARCH=wasm go build -o ./bin/relay.wasm ./cmd
-
 # Run the light node.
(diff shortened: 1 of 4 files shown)


pschork:
Cloudflare web workers are limited to 128MB memory. Closing.

PR #1560, opened 2025-05-13
feat: adds batch header hash to blobs/feed and blobs/blob_key response

pschork:
Resolves EGDA-1465

### DIFF
--- a/disperser/dataapi/v2/blobs.go
+++ b/disperser/dataapi/v2/blobs.go
@@ -211,10 +211,11 @@ func (s *ServerV2) FetchBlob(c *gin.Context) {
 		BlobSizeBytes: metadata.BlobSize,
 	}
 
-	// Get batch header hash from attestation info if available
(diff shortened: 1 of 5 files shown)


disperser/dataapi/v2/blobs.go
```diff
@@ -447,6 +457,15 @@ func (s *ServerV2) sendBlobFeedResponse(
 		}
 		blobInfo[i].BlobKey = bk.Hex()
 		blobInfo[i].BlobMetadata = createBlobMetadata(blobs[i])
+
+		// Get batch header hash from attestation info if available
+		attestationInfo, err := s.blobMetadataStore.GetBlobAttestationInfo(c.Request.Context(), bk)
```
jianoaix, line 462: This is inefficient, as it's in blob feed and it'll make hundreds of calls to dynamoDB

pschork: Profiled this change and it is ~36x slower vs without the batch header hash.

PR #1562, opened 2025-05-14
fix: littdb read metrics

cody-littley:

Fix broken LittDB read metrics. Currently, the DB doesn't report reads that only hit the cache.

### DIFF
--- a/litt/cache/cached_table.go
+++ b/litt/cache/cached_table.go
@@ -6,6 +6,7 @@ import (
 
 	"github.com/Layr-Labs/eigenda/common/cache"
 	"github.com/Layr-Labs/eigenda/litt"
+	"github.com/Layr-Labs/eigenda/litt/metrics"
 	"github.com/Layr-Labs/eigenda/litt/types"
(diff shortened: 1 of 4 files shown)


litt/cache/cached_table.go
```diff
@@ -114,17 +119,28 @@ func (c *cachedTable) CacheAwareGet(
...
+		return value, exists, hot, nil
 	} else {
 		value, exists = c.readCache.Get(stringKey)
 		if exists {
 			// The value was recently read
-			return value, true, true, nil
+			hot = true
+			return value, exists, hot, nil
 		}
 	}
```
litt3, line 146: There is still one return case where you manually specify bools, instead of using `exists` and `hot`. Is that intentional?

litt3 (approved):

litt/cache/cached_table.go
cody-littley: There was a method to the madness, but since it wasn't obvious to the reader, probably not a good pattern. I now always return the variables in this function.

PR #1564, opened 2025-05-14
feat!: LittDB multi-drive support

cody-littley:
Enable validators to run with multiple drives. Removes support for levelDB in v2.


node/validator_store.go
```diff
@@ -61,12 +41,7 @@ type ValidatorStore interface {
 
 	// StoreBatch stores a batch and its raw bundles in the database. Returns the keys of the stored data
```
litt3, line 42: > Returns the keys
 
outdated

node/validator_store.go
```diff
@@ -128,211 +80,97 @@ type validatorStore struct {
...
-	levelDBPath := path.Join(config.DbPath, LevelDBPath)
-	levelDBDeletionPath := path.Join(config.DbPath, LevelDBDeletionPath)
-
-	// If we previously made an attempt at deleting the levelDB database but it was interrupted, delete it now.
-	exists, err := util.Exists(levelDBDeletionPath)
-	if err != nil {
-		return nil, fmt.Errorf("failed to stat path %s: %v", levelDBDeletionPath, err)
+	stringBuilder := strings.Builder{}
+	if len(config.LittDBStoragePaths) > 1 {
+		stringBuilder.WriteString("s")
```
litt3, line 95: this is confusing when reading this function linearly, consider writing `Using littDB at path` to the string builder, and then just doing an `logger.Info(stringBuilder.String())` below

litt3 (approved):

node/flags/flags.go
```diff
@@ -439,6 +421,12 @@ var (
 		Value:    0.05,
 		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "LITT_DB_READ_CACHE_SIZE_FRACTION"),
 	}
+	LittDBStoragePathsFlag = cli.StringSliceFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "litt-db-storage-paths"),
+		Usage:    "Comma separated list of paths to store the LittDB data files. At least one path must be provided.",
+		Required: true,
+		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "LITT_DB_STORAGE_PATHS"),
```
dmanc, line 428: What's the behavior of changing this flag / what is supported? Some example scenarios:

1. Running with 1 path and I add another path.
2. Running with 2 paths and I remove a path
3. Running with 1 path then i change that path.

pschork: Can this be optional and fallback to legacy `NODE_DB_PATH` if not set?
Currently the `NODE_DB_PATH=/data/operator/db` and is mounted via docker volume defined from `NODE_DB_PATH_HOST=${EIGENDA_HOME}/db`. Is the migration as simple as setting `NODE_LITT_DB_STORAGE_PATHS=/data/operator/db` or does this path expect something like `NODE_LITT_DB_STORAGE_PATHS=/data/operator/db/chunks_v2` when migrating?

node/validator_store.go
cody-littley: fixed

node/validator_store.go
cody-littley: makes sense, done

node/flags/flags.go
cody-littley: @pschork I've added the requested fallback. If you are still using it in a year, I'm going to rugpull you. 😜 

```
	if len(config.LittDBStoragePaths) == 0 {
		logger.Warnf("WARNING: setting NODE_DB_PATH is deprecated and will be removed in a future version. " +
			"Please use NODE_LITT_DB_STORAGE_PATHS instead.")
		config.LittDBStoragePaths = []string{
			config.DbPath + "/chunk_v2_litt",
		}
	}
```

litt3 (approved):

PR #1566, opened 2025-05-15
fix: replaces hard coded quorum IDs on dataapi handlers

pschork:

DataAPI liveness/node-info/signing-info handlers used a hard coded quorum IDs slice `{0,1,2}` which caused them to fail against sepolia testnet.
This fix dynamically determines quorum IDs to support testnet-sepolia which lacks quorum 2

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/cmd/dataapi/main.go
+++ b/disperser/cmd/dataapi/main.go
@@ -12,6 +12,7 @@ import (
 	"github.com/Layr-Labs/eigenda/common/aws/dynamodb"
 	"github.com/Layr-Labs/eigenda/common/aws/s3"
 	"github.com/Layr-Labs/eigenda/common/geth"
+	"github.com/Layr-Labs/eigenda/core/eth"
 	coreeth "github.com/Layr-Labs/eigenda/core/eth"
 	"github.com/Layr-Labs/eigenda/core/thegraph"
 	"github.com/Layr-Labs/eigenda/disperser/cmd/dataapi/flags"
@@ -91,6 +92,17 @@ func RunDataApi(ctx *cli.Context) error {
 		return err
 	}
 
+	// Determine valid set of quorum IDs at startup
+	currentBlock, err := tx.GetCurrentBlockNumber(context.Background())
+	if err != nil {
+		return err
+	}
+	quorumCount, err := tx.GetQuorumCount(context.Background(), uint32(currentBlock))
+	if err != nil {
+		return err
+	}
+	quorumIds := eth.GetAllQuorumIDs(quorumCount)
+
 	var (
 		promClient        = dataapi.NewPrometheusClient(promApi, config.PrometheusConfig.Cluster)
 		blobMetadataStore = blobstore.NewBlobMetadataStore(dynamoClient, logger, config.BlobstoreConfig.TableName, 0)
@@ -108,6 +120,7 @@ func RunDataApi(ctx *cli.Context) error {
 				DisperserHostname:  config.DisperserHostname,
 				ChurnerHostname:    config.ChurnerHostname,
 				BatcherHealthEndpt: config.BatcherHealthEndpt,
+				QuorumIds:          quorumIds,
 			},
 			sharedStorage,
 			promClient,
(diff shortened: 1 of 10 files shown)


disperser/dataapi/operator_handler.go
```diff
@@ -100,7 +98,15 @@ func (oh *OperatorHandler) ProbeV2OperatorsLiveness(ctx context.Context, operato
 	if err != nil {
 		return nil, err
 	}
-	state, err := oh.indexedChainState.GetIndexedOperatorState(ctx, uint(currentBlock), defaultQuorumIds)
+
+	// Get quorum count and generate all quorum IDs
+	quorumCount, err := oh.chainReader.GetQuorumCount(ctx, uint32(currentBlock))
```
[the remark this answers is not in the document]
pschork: Considering this handler does potentially hundreds of RPC fan out to probe validator liveness, I think we can deal with one additional RPC to determine quorums.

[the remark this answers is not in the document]
pschork: Profiled this change vs hardcoded quorums against sepolia testnet and response times where consistently the same 2.46s. The quorum query has no meaningful impact on liveness response time - the majority of the response time is spent waiting for RPC fan out to validators to return - not to mention the need to absorb potential RPC timeouts if a validator is down.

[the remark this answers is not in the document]
jianoaix: I think: 1) this isn't the only API that uses this; 2) I'd minimize the cost, and the chance of security issue, by not making unnecessary calls

And there are ways to achieve it, by fetch it once for the entire server lifetime; or make it fetch periodically (which may not necessary given how infrequent it changes)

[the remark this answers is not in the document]
pschork: Fetches quorum IDs at startup. PTAL

disperser/cmd/dataapi/main.go
```diff
@@ -108,6 +120,7 @@ func RunDataApi(ctx *cli.Context) error {
 				DisperserHostname:  config.DisperserHostname,
 				ChurnerHostname:    config.ChurnerHostname,
 				BatcherHealthEndpt: config.BatcherHealthEndpt,
+				QuorumIds:          quorumIds,
```
jianoaix, line 123: It seems not natural to have this as part of dataapi config, compared to e.g. having the quorum count fetch inside the NewOperatorHandler, which makes code change more contained.

pschork: NewOperatorHandler does not return err, so I moved the quorum ID query into main. I can refactor to keep it more contained.

pschork: Updated. PTAL

disperser/dataapi/operator_handler.go
```diff
@@ -84,23 +83,36 @@ func (o *OperatorList) GetID(address string) (core.OperatorID, bool) {
...
 }
 
-func NewOperatorHandler(logger logging.Logger, metrics *Metrics, chainReader core.Reader, chainState core.ChainState, indexedChainState core.IndexedChainState, subgraphClient SubgraphClient) *OperatorHandler {
+func NewOperatorHandler(logger logging.Logger, metrics *Metrics, chainReader core.Reader, chainState core.ChainState, indexedChainState core.IndexedChainState, subgraphClient SubgraphClient) (*OperatorHandler, error) {
+	// Determine valid set of quorum IDs at startup
+	currentBlock, err := chainReader.GetCurrentBlockNumber(context.Background())
+	if err != nil {
+		return nil, err
+	}
+	quorumCount, err := chainReader.GetQuorumCount(context.Background(), uint32(currentBlock))
```
jianoaix, line 92: If the onchain state changes for quorum count, it seems still correct, just incomplete (e.g. it gives results from quorum 0 to 2, and missing for quorum 3)

jianoaix (approved):

disperser/dataapi/operator_handler.go
pschork: meaning it requires a restart if quorum is added? yes

jianoaix: more about what will go wrong if it hasn't restarted to catch up the onchain state

PR #1578, opened 2025-05-20
feat: versioned blob params v2

pakim249CAL:

Versioned Blob params struct update

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


contracts/src/periphery/cert/libraries/EigenDACertVerificationLib.sol
```diff
@@ -186,17 +188,23 @@ library EigenDACertVerificationLib {
      * @return errParams Additional error parameters
      */
     function checkSecurityParams(
```
mooselumph, line 190: Just double-checking my understanding: Because this is part of the library, it is deployed as part of the CertVerifier; therefore, any CertVerifiers which are already deployed will still be executing the old version of the code.

pakim249CAL: yes, that is correct. If the library changes and a new cert verifier is deployed, it is as if the code in the contract changes as well.

contracts/src/periphery/cert/libraries/EigenDACertVerificationLib.sol
```diff
@@ -29,10 +29,12 @@ library EigenDACertVerificationLib {
     error InvalidInclusionProof(uint256 blobIndex, bytes32 blobHash, bytes32 rootHash);
 
     /// @notice Thrown when security assumptions are not met
-    /// @param gamma The difference between confirmation and adversary thresholds
-    /// @param n The calculated security parameter
-    /// @param minRequired The minimum required value for n
-    error SecurityAssumptionsNotMet(uint256 gamma, uint256 n, uint256 minRequired);
+    /// @param confirmationThreshold The confirmation threshold
+    /// @param adversaryThreshold The adversary threshold
+    /// @param reconstructionThreshold The reconstruction threshold
```
samlaf, line 34: I feel very strongly about and would really appreciate if you tried to write meaningful and useful comments. To me these comments are worse than not having comments. If you don't understand the params well enough to describe them, then its fine to tag someone else to provide those explanations. Let's all help each other make the documentation useful.

contracts/src/core/libraries/v2/EigenDATypesV2.sol
```diff
@@ -56,4 +56,12 @@ library EigenDATypesV2 {
...
         uint32[] quorumNumbers;
     }
+
+    struct VersionedBlobParams {
+        uint32 codingRate;
+        uint32 reconstructionThreshold;
+        uint32 numChunks;
+        uint32 numUnits;
+        uint32 samplesPerUnit;
+    }
```
samlaf, line 66: Please document these. cc @mooselumph can you help with this?

contracts/src/core/interfaces/IEigenDAThresholdRegistry.sol
```diff
@@ -47,3 +47,11 @@ interface IEigenDAThresholdRegistry {
     /// @notice Returns the blob params for a given blob version
     function getBlobParams(uint16 version) external view returns (DATypesV1.VersionedBlobParams memory);
 }
+
+interface IEigenDAThresholdRegistry is IEigenDAThresholdRegistryBase {
```
samlaf, line 51: Why is this separation between Base and notBase needed? Base doesn't seem to be a Base given that its got both v1 and v2 stuff, and then this interface is adding more V2 stuff on top. Pretty confusing. Can we describe the interface and why its organized this way?

contracts/src/core/interfaces/IEigenDAServiceManager.sol
```diff
@@ -4,10 +4,10 @@ pragma solidity ^0.8.9;
 import {IServiceManager} from "lib/eigenlayer-middleware/src/interfaces/IServiceManager.sol";
 import {BLSSignatureChecker} from "lib/eigenlayer-middleware/src/BLSSignatureChecker.sol";
 import {BN254} from "lib/eigenlayer-middleware/src/libraries/BN254.sol";
-import {IEigenDAThresholdRegistry} from "src/core/interfaces/IEigenDAThresholdRegistry.sol";
+import {IEigenDAThresholdRegistryBase} from "src/core/interfaces/IEigenDAThresholdRegistry.sol";
 import {EigenDATypesV1 as DATypesV1} from "src/core/libraries/v1/EigenDATypesV1.sol";
 
-interface IEigenDAServiceManager is IServiceManager, IEigenDAThresholdRegistry {
+interface IEigenDAServiceManager is IServiceManager, IEigenDAThresholdRegistryBase {
```
samlaf, line 10: prob good to document why this extends Base instead of ThresholdRegistry

contracts/src/core/interfaces/IEigenDAThresholdRegistry.sol
samlaf: Is that goal that once we get rid of V1, then `Base` will be deleted and we'll only be left with this minimal `IEigenDAThresholdRegistry` interface? Why does it need to extend the Base interface btw? All good things to document

samlaf (changes requested):
Going to start being a stickler and enforce proper documentation in our contracts. We've had a lot of issues with this in the past, and I really want to change this, as knowledge dissemination only scales with proper documentation, and proper documentation leads to more eyes looking at the contract, and more eyes means more bugs/improvement getting found. @mooselumph can you help document the important struct and verification changes?

contracts/src/core/interfaces/IEigenDAThresholdRegistry.sol
pakim249CAL: No, the goal here is simply to accommodate the fact that the ServiceManager should not be upgraded, but right now it inherits the Threshold Registry. It is a mess.

contracts/src/core/interfaces/IEigenDAServiceManager.sol
pakim249CAL: Will do

contracts/src/periphery/cert/libraries/EigenDACertVerificationLib.sol
pakim249CAL: Okay, then I'm tagging in @mooselumph because IDK at all really.

contracts/src/core/interfaces/IEigenDAServiceManager.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1578/commits/ddf56dce74d3b8d4eeebb610bea3a7b152c37362

pakim249CAL:
cancelled

PR #1579, opened 2025-05-20
fix: separate cert verifier interface

pakim249CAL:

Separates the cert verifier interface from its bases as off-chain prefers.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


contracts/src/periphery/cert/interfaces/IEigenDACertVerifier.sol
```diff
@@ -9,7 +9,7 @@ import {IEigenDASignatureVerifier} from "src/core/interfaces/IEigenDASignatureVe
 
 import {EigenDATypesV1 as DATypesV1} from "src/core/libraries/v1/EigenDATypesV1.sol";
 
-interface IEigenDACertVerifier is IEigenDACertVerifierBase, IVersionedEigenDACertVerifier {
+interface IEigenDACertVerifier {
```
samlaf, line 12:
```suggestion
// IEigenDACertVerifier provides the getters necessary to transform a BlobStatusReply received after a dispersal into a Cert that can be verified by the EigenDACertVerifier that implements this interface version.
interface IEigenDACertVerifier {
```

samlaf (commented):
This separation makes sense to me generically, but I haven't tried actually working with it offchain, so let's wait for @ethenotethan to stamp this as working with what he has in mind for the offchain integration.

litt3 (dismissed):
LGTM, let's wait for Ethen's approval, as mentioned by Sam

PR #1581, opened 2025-05-21
Create interface for offchain metering store

dmanc:

So we can support different storage backends in the future. See https://github.com/Layr-Labs/eigenda/pull/1448 for example. Also renamed `OffchainStore` to `MeteringStore` but open to other names.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


hopeyen (approved):
💯

PR #1585, opened 2025-05-21
fix(contracts): Update V2 cert verifier deploy script to use latest verifier

ethenotethan:
…

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/contracts/script/deploy/certverifier/README.md
+++ b/contracts/script/deploy/certverifier/README.md
@@ -1,6 +1,6 @@
-## EigenDA Cert Verfier Deployer
+## EigenDA V2 Cert Verfier Deployer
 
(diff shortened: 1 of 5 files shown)


litt3 (approved):
LGTM

contracts/script/deploy/certverifier/README.md
```diff
@@ -23,7 +20,7 @@ To set up the deployment, a config json should be placed in the `config/` folder
 }
 ```
 
-Two sample configs are provided in the `config/` folder for preprod and testnet environments.
+Three sample configs are provided in the `config/` folder for preprod and testnet environments on holesky as well as testnet environment on sepolia.
```
pakim249CAL, line 23: IMO the number of configs should not be specified here since it is a double bookkeeping

ethenotethan: Would prefer to just leave as is if possible

PR #1586, opened 2025-05-21
feat: adds account reservation usage timeseries to dataapi

pschork:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


disperser/dataapi/v2/accounts.go
```diff
@@ -102,3 +103,96 @@ func (s *ServerV2) FetchAccountBlobFeed(c *gin.Context) {
...
+// AccountReservationUsageResponse represents the response for account reservation usage
+type AccountReservationUsageResponse struct {
+	AccountId string         `json:"account_id"`
+	Records   []PeriodRecord `json:"records"`
+}
+
+// PeriodRecord represents a single period's usage record
+type PeriodRecord struct {
+	ReservationPeriod uint32 `json:"reservation_period"`
+	UsageBytes        uint64 `json:"usage_bytes"`
```
hopeyen, line 197: usage is metered in unit of symbols (32bytes)

core/meterer/dynamodb_metering_store.go
```diff
@@ -332,3 +332,32 @@ func parsePeriodRecord(bin map[string]types.AttributeValue) (*pb.PeriodRecord, e
 		Usage: uint64(binUsageValue),
 	}, nil
 }
+
+func (s *DynamoDBMeteringStore) GetReservationBinUsage(ctx context.Context, accountID gethcommon.Address, startPeriod uint64, limit int32) ([]*pb.PeriodRecord, error) {
```
hopeyen, line 336: function looks good. wondering if it is worth refactoring the existing `GetPeriodRecords` to take in a limit, instead of adding a method, since the functionality is largely the same

disperser/dataapi/v2/accounts.go
pschork: removed symbols -> bytes conversion

core/meterer/dynamodb_metering_store.go
pschork: refactored

core/meterer/dynamodb_metering_store.go
```diff
@@ -241,18 +241,18 @@ func (s *DynamoDBMeteringStore) GetPeriodRecords(ctx context.Context, accountID
 			":reservationPeriod": &types.AttributeValueMemberN{Value: strconv.FormatUint(reservationPeriod, 10)},
 		},
 		ScanIndexForward: aws.Bool(true),
-		Limit:            aws.Int32(MinNumBins),
+		Limit:            aws.Int32(limit),
 	}
 	bins, err := s.dynamoClient.QueryWithInput(ctx, queryInput)
 	if err != nil {
-		return [MinNumBins]*pb.PeriodRecord{}, fmt.Errorf("failed to query payments for account: %w", err)
+		return make([]*pb.PeriodRecord, MinNumBins), fmt.Errorf("failed to query payments for account: %w", err)
```
hopeyen, line 248: should this be length of `limit` instead of `MinNumBins`? also for below ones

pschork: A valid query w/limit 1000 might only return 100 valid results - in which case it returns 100 PeriodRecords. But if it doesn't find anything, it would then return 1000 empty PeriodRecords - that seems odd. My original impl returned empty slice, but when I refactored GetPeriodRecords, it always returned `MinNumBins` size slice - which I don't totally understand why. So I chose to not break any caller assumptions and return `MinNumBins` size slice when the query fails.

hopeyen: I see! It was for the clients to always be able to create the MinNumBins even if they are empty. the reasoning for dataapi makes sense to me

hopeyen (approved):
Kind of wonder if adding a meterer to the dataapi is the correct approach, like if we should instead expose an internal apiserver endpoint for un-authed payment state, or that we remove auth entirely... I think we didn't reach any conclusion on that topic, but I'm okay to proceed like this

PR #1588, opened 2025-05-22
feat: Validator blacklisting dispersers

anupsv:

Adding blacklisting of disperser by validators

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


.github/workflows/golangci-lint.yml
```diff
@@ -13,15 +13,24 @@ jobs:
     name: Linter
     runs-on: ubuntu-latest
     steps:
+<<<<<<< Updated upstream
```
hopeyen: lets remove the merge conflict lines

inabox/deploy/config.go
```diff
@@ -769,6 +769,7 @@ func (env *Config) GenerateAllVariables() {
 		if err != nil {
 			log.Panicf("Error: %s", err.Error())
 		}
+		fmt.Println("composeFile", string(composeFile))
```
hopeyen: is this print helpful in debugging inabox deploy or should we remove it?

node/blacklist.go
```diff
@@ -0,0 +1,50 @@
...
+	Reason    string `json:"reason"`
+}
+
+// ToBytes serializes the Blacklist to JSON bytes
+func (b *Blacklist) ToBytes() ([]byte, error) {
+	return json.Marshal(b)
+}
+
+// FromBytes deserializes JSON bytes into the current Blacklist
+func (b *Blacklist) FromBytes(data []byte) error {
```
hopeyen, line 35: is it correct that every blacklist operation serializes or deserializes the entire entry list? maybe in the future we can both cache the list memory and write to db? I'm mostly worried about db taking a while for validating if a key has been blacklisted

common/aws/kms_fuzz_test.go
```diff
@@ -27,7 +27,7 @@ func generateValidSignature() (*ecdsa.PublicKey, []byte, []byte, error) {
 
 	// Define a message and compute its SHA-256 hash.
 	message := "Test message for ECDSA signature"
-	hash := sha256.Sum256([]byte(message))
+	hash := 56.Sum256([]byte(message))
```
dmanc: Think it's a typo

node/blacklist_store.go
```diff
@@ -0,0 +1,184 @@
...
+
+	blacklist.AddEntry(disperserId, contextId, reason)
+
+	data, err := blacklist.ToBytes()
+	if err != nil {
+		return fmt.Errorf("failed to serialize blacklist: %w", err)
+	}
+
+	s.logger.Info("Adding entry to blacklist", "disperserId", disperserId, "contextId", contextId, "reason", reason)
+	disperserIdHash := sha256.Sum256(fmt.Appendf(nil, "%d", disperserId))
```
dmanc, line 158: What's the point of hashing? Don't all the dispersers get a unique id already?

inabox/deploy/config.go
anupsv: we can remove it now that inabox is gonna get reworked.

node/blacklist_store.go
anupsv: It's for a uniform standard way for storing in the db. Incase we change something in the id format or allow something unique or other cases, the sha hash will form a proper set of bytes to store.

node/blacklist.go
anupsv: I was thinking of this interms of cleanup of old records as well. But normal functioning, the disperser can get banned 1 in a day smallest, and due to no cleanup of old records, it might take longer for that disperser only. So they'd they affecting themselves but not others since we first check if the disperser actually exists in the blacklist.

node/blacklist_store.go
```diff
@@ -0,0 +1,199 @@
...
+	HasDisperserID(ctx context.Context, disperserId uint32) bool
+
+	// GetByDisperserID retrieves a blacklist by disperser ID
+	GetByDisperserID(ctx context.Context, disperserId uint32) (*Blacklist, error)
+
+	// DeleteByDisperserID deletes a blacklist by disperser ID
+	DeleteByDisperserID(ctx context.Context, disperserId uint32) error
+
+	// Get retrieves a blacklist by key
+	Get(ctx context.Context, key []byte) (*Blacklist, error)
```
hopeyen, line 28: for the get and updates, do we need a lock to prevent concurrent disperser requests going through when it is getting blacklisted?

hopeyen (approved):
📓

node/blacklist_store.go
anupsv: discussed offline.

PR #1595, opened 2025-05-23
ci: add env and update readme to include the correct mdBook

bxue-l2:
Publish md book in CI

### DIFF
--- a/.github/workflows/mdbook.yaml
+++ b/.github/workflows/mdbook.yaml
@@ -8,6 +8,10 @@ on:
 jobs:
   deploy:
     runs-on: ubuntu-latest
+    environment:
+      name: github-pages
+      url: ${{ steps.deploy.outputs.page_url }}
+
     permissions:
       contents: write  # To push a branch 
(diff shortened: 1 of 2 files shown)


.github/workflows/mdbook.yaml
```diff
@@ -8,6 +8,10 @@ on:
 jobs:
   deploy:
     runs-on: ubuntu-latest
+    environment:
+      name: github-pages
+      url: ${{ steps.deploy.outputs.page_url }}
```
samlaf, line 13: why are you defining this? it doesn't seem like its used?

.github/workflows/mdbook.yaml
```diff
@@ -20,7 +24,7 @@ jobs:
         run: make build # also installs deps
         working-directory: docs/spec
       - name: Setup Pages
-        uses: actions/configure-pages@v4
+        uses: actions/configure-pages@v5
```
samlaf, line 27: why is this bump needed?

samlaf (approved):

PR #1599, opened 2025-05-26
feat: validator node multi account metering

hopeyen:

we are not adding db requirements to validator nodes anymore for MVP.
- validators are not distributed, so they don't need DB for keeping data in sync within the node
- validators only need to track recent reservation usages, so they don't need usage persistence. By the design of reservations, validator losing the cache means losing a 15minutes of offchain usage tracked. This implies validators may do more work for the current reservation duration since they forgot they have already done work in the previous session. 
- decentralized on-demand isn't part of MVP, so validators don't need to persist ondemand payment records

This PR takes an approach more similar to the client's accountant implementation, which calculates and validates off-chain usage in cache. We noted that a reservation user cannot target a particular validator, and a particular validator providing more reservation bandwidth than on-chain agreement does not affect overall network liveness and safety. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


node/flags/flags.go
```diff
@@ -470,6 +470,14 @@ var (
 		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "GC_SAFETY_BUFFER_SIZE_GB"),
 	}
 
+	ReservationNumBinsFlag = cli.IntFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "reservation-num-bins"),
+		Usage:    "The number of bins to track for each account's reservation usage. The minimum value is 3.",
+		Required: false,
+		Value:    3,
```
dmanc: Whats the reason for ever changing this value from 3 to some bigger number?

node/grpc/server_v2.go
```diff
@@ -76,6 +79,13 @@ func NewServerV2(
 		config.StoreChunksRequestMaxPastAge,
 		config.StoreChunksRequestMaxFutureAge)
 
+	// TODO(hopeyen): BatchMeterer initialization is left as a placeholder.
```
dmanc: Does this mean if the PR is merged it's not gonna affect how the validator currently operates?

node/grpc/server_v2.go
```diff
@@ -157,6 +168,19 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
 			}
 		}
 	}
+
+	// Batch metering validation
+	if s.batchMeterer != nil {
+		// If the batch meterer is configured, use it to validate the batch
+		err = s.batchMeterer.MeterBatch(ctx, batch, time.Now())
```
dmanc: Should we use in.Timestamp instead?

node/batch_meterer.go
```diff
@@ -0,0 +1,482 @@
...
+		}
+		quorumIDs := make([]core.QuorumID, 0, len(quorumSet))
+		for quorumID := range quorumSet {
+			quorumIDs = append(quorumIDs, quorumID)
+		}
+		slices.Sort(quorumIDs)
+		b.logger.Info("Sorted quorum IDs", "quorumIDs", quorumIDs)
+
+		// Get active reservations for this account's quorums
+		reservations, err := b.ChainPaymentState.GetReservedPaymentByAccountAndQuorums(ctx, accountID, quorumIDs)
```
dmanc, line 198: Would it make sense to cache this or prefetch it with some timer? Maybe a future optimization?

node/batch_meterer.go
```diff
@@ -0,0 +1,482 @@
...
+	if accountUsage.PeriodRecords[quorumID][relativeIndex].Index < uint32(period) {
+		accountUsage.PeriodRecords[quorumID][relativeIndex].Index = uint32(period)
+		accountUsage.PeriodRecords[quorumID][relativeIndex].Usage = 0
+	}
+
+	return accountUsage.PeriodRecords[quorumID][relativeIndex]
+}
+
+// symbolsCharged returns the number of symbols charged for a given data length
+func (b *BatchMeterer) symbolsCharged(numSymbols uint64) uint64 {
```
dmanc: This code also exists in the clients, does it make sense to make some shared common library for accounting related stuff?

Also for my learning, what's the rationale for having the symbolsCharged be multiples of minSymbols? Does it simplify some accounting logic?

node/batch_meterer.go
```diff
@@ -0,0 +1,482 @@
...
+	minSymbols := b.ChainPaymentState.GetMinNumSymbols()
+	if numSymbols <= minSymbols {
+		return minSymbols
+	}
+	// Round up to the nearest multiple of MinNumSymbols
+	return core.RoundUpDivide(numSymbols, minSymbols) * minSymbols
+}
+
+// validateReservationPeriod checks if the provided reservation period is valid
+// Based on meterer.ValidateReservationPeriod
```
dmanc: If these are pretty much the same function why not import?

node/batch_meterer.go
```diff
@@ -0,0 +1,482 @@
...
+		}
+		b.logger.Info("Got reservations", "reservations", reservations)
+
+		// Process each update for this account
+		for _, update := range accountUpdates {
+			b.logger.Debug("Processing update", "update", update)
+			reservation, ok := reservations[update.quorumID]
+			if !ok {
+				b.rollbackUpdates(successfulUpdates)
+				return fmt.Errorf("account %s has no reservation for quorum %d", accountID.Hex(), update.quorumID)
```
dmanc, line 209: Do we rollback in these cases because we assume that the disperser messed up / is malicious?

node/batch_meterer.go
```diff
@@ -0,0 +1,482 @@
...
+		"accountID", accountID.Hex(),
+		"quorumID", quorumID,
+		"prevUsage", prevUsage,
+		"newUsage", usage,
+		"binLimit", binLimit,
+		"currentPeriod", currentPeriod,
+	)
+
+	// Write-first: increment usage
+	periodRecord.Usage += usage
```
dmanc, line 288: Would this need a lock?

node/flags/flags.go
hopeyen: If the node wants to track a longer history than just 3 reservation window, they can increment this number

node/grpc/server_v2.go
hopeyen: yep. I want to enable this when there's integration tests on the batch meterer behavior

node/batch_meterer.go
hopeyen: `ChainPaymentState` comes with a cache ([line](https://gitkraken.dev/link/dnNjb2RlOi8vZWFtb2Rpby5naXRsZW5zL2xpbmsvci8wZThmZDhkM2UyNDA4NWI1NmY4ODg0MmNiMWJmMmNhMjQ5MDRjMjBiL2YvY29yZS9tZXRlcmVyL29uY2hhaW5fc3RhdGUuZ28%2FdXJsPWh0dHBzJTNBJTJGJTJGZ2l0aHViLmNvbSUyRkxheXItTGFicyUyRmVpZ2VuZGEuZ2l0?origin=gitlens))

node/batch_meterer.go
hopeyen: Yes, this code also exists in the apiserver 😬 I agree on having a common library. I can create a separate PR that dedicate to DRY some of these logic. 

`minSymbols` is the minimum charge unit. I don't think there's any impact of the logic itself, but it will enforce the symbols charged to be some multiple of $$2^x$$, similar to `minSymbols` and `blobSize`

node/batch_meterer.go
hopeyen: `meterer.ValidateReservationPeriod` needs a `meterer` and batch meterer is a different struct 😞 
I can DRY out `ValidateReservationPeriod` and other functions so it doesn't depend one having the struct, and ask each to pass in the necessary fields

node/batch_meterer.go
hopeyen: disperser can be honest and still run into these problems if a client uses multiple dispersers within a reservation window and overall exceed their ratelimit. In this case, we won't blacklist the disperser, but allow the disperser to remove that client's requests and resubmit the batch.
In the happy path with honest client and disperser, the disperser will not submit batches that contains a client reaching their ratelimit. In the unhappy path, validator can blacklist a malicious disperser, and disperser can reject an abusive client

node/batch_meterer.go
dmanc: We can refactor in some future PR, maybe add todo so we don't forget

node/flags/flags.go
dmanc: Why would the validators want/care to track a longer history? Usually the validators will just stick to the default values.

node/batch_meterer.go
```diff
@@ -0,0 +1,458 @@
...
+// MeterBatch meters a batch of blobs directly from a corev2.Batch
+// Returns an error if any part of the batch is invalid
+func (b *BatchMeterer) MeterBatch(
+	ctx context.Context,
+	batch *corev2.Batch,
+	batchReceivedAt time.Time,
+) error {
+	b.logger.Info("MeterBatch", "batch", batch, "batchReceivedAt", batchReceivedAt)
+	// Convert the batch to a map of account IDs to quorum IDs to symbols usage, aggregated over the batch's blobs
+	usageMap, err := b.BatchRequestUsage(batch)
```
dmanc: nit: usageMap is not a map right?

api/clients/v2/accountant.go
```diff
@@ -91,7 +91,7 @@ func (a *Accountant) BlobPaymentInfo(
 		return big.NewInt(0), nil
 	}
 
-	overflowPeriodRecord := a.GetRelativePeriodRecord(currentReservationPeriod + 2)
+	overflowPeriodRecord := a.GetOrRefreshRelativePeriodRecord(currentReservationPeriod+2*reservationWindow, reservationWindow)
```
anupsv: should probably say why +2.

node/flags/flags.go
hopeyen: yeah, the default works, but can configure if they wanted to track more usage. I'm okay to remove this since it seems confusing for both clients/validators

node/batch_meterer.go
hopeyen: cool. this should be done in https://github.com/Layr-Labs/eigenda/pull/1653

api/clients/v2/accountant.go
samlaf: +1

node/flags/flags.go
samlaf: If we're pretty sure that no one is using this feature, then I would definitely opt for removing any unneeded complexity.

hopeyen: yeah, I'm removing these from both the accountant and the validator node

api/clients/v2/accountant.go
hopeyen: docs were added

samlaf: which commit? I'm confused accountant is not even part of the files changed in this PR anymore

hopeyen: it was refactored [here](https://github.com/Layr-Labs/eigenda/pull/1653/files)
The PR never meant to have anything to do with the accountant, so the changes were cleaned up to other PRs

samlaf:
@litt3 can we close this?

PR #1607, opened 2025-05-28
fix: Update retrieval clients to take in cert sum type interface

ethenotethan:
…

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/clients/v2/examples/example_relay_retrieval_test.go
```diff
@@ -48,14 +46,9 @@ func Example_relayPayloadRetrieval() {
...
-
-	eigenDAV3Cert, ok := eigenDACert.(*coretypes.EigenDACertV3)
-	if !ok {
-		panic(fmt.Sprintf("expected EigenDACertV3, got %T", eigenDACert))
-	}
-
+	
 	// Retrieve the payload using the certificate
-	_, err = payloadRetriever.GetPayload(ctx, eigenDAV3Cert)
+	_, err = payloadRetriever.GetPayload(ctx, eigenDACert)
```
samlaf, line 51: I feel like this example is cheating/incomplete after re-reading it. Because we are taking the interface returned by the payload disperser, and feeding it directly into the payload retriever. Whereas in a real world scenario (meaning any of our users who will look at this example), they will have to serialize the retrieved cert, to be sent to some p2p network, or in a batcher inbox, etc. And then they will have to parse the network into the interface in order to be able to call `GetPayload`.

Really sorry I keep going back and forth on this interface... but now I'm thinking we might want to add `Serialize()` back to the Cert interface? Haven't thought through everything this affects so there might be a better way to do this? Though I feel letting the user do a manual type switch just to serialize is prob very bad UX for them, given that this is something EVERYONE of them will need (my hunch at least).

And we should also have a Parse() function somewhere in core repo to parse bytes into a cert interface (sum-type). This was actually a ticket assigned to me (https://linear.app/eigenlabs/issue/DAINT-465/certrbn-parsing-golang-library). Wasn't sure how/when/where we would need this, but I can work on this if we don't already have this somewhere.

api/clients/v2/payloadretrieval/relay_payload_retriever.go
```diff
@@ -65,25 +66,57 @@ func NewRelayPayloadRetriever(
...
+	var blobKey *core.BlobKey
+	var relayKeys []core.RelayKey
+	var blobCommitments *encoding.BlobCommitments
+	var err error
+
+	switch eigenDACert.Version() {
+	case coretypes.VersionTwoCert:
+		eigenDAV2Cert, ok := eigenDACert.(*coretypes.EigenDACertV2)
+		if !ok {
+			return nil, fmt.Errorf("expected eigenDACert to be of type *coretypes.EigenDAV2Cert, got %T", eigenDACert)
```
samlaf, line 80: I think we can switch directly on the type? Which would remove the need for this `!ok` case (which really should never happen...). We'd get better type safety and cleaner code. Haven't tried to do it though

api/clients/v2/examples/example_relay_retrieval_test.go
samlaf: Perhaps we also want `ComputeBlobKey` on the interface? Realizing I might have pushed the sum-type minimal interface too far... :\ If there are things that we are 95% sure won't change, and will dramatically reduce boilerplate switch bloat code that our users will have to write... then I think that's probably a good tradeoff? (and once in a while its prob fine to break them?)

api/clients/v2/payloadretrieval/relay_payload_retriever.go
```diff
@@ -65,25 +66,57 @@ func NewRelayPayloadRetriever(
 // verified prior to calling this method.
 func (pr *RelayPayloadRetriever) GetPayload(
 	ctx context.Context,
-	eigenDACert *coretypes.EigenDACertV3) (*coretypes.Payload, error) {
+	eigenDACert coretypes.EigenDACert) (*coretypes.Payload, error) {
```
samlaf, line 69: I think (?) the cleanest way to implement the clients might be to define 2 cert interfaces: a minimal "sum-type" one that we return to our users (because used externally) on `DispersePayload`, and a larger one (the one you had before... apologies again) for `GetPayload`. 

We shouldn't need to write the below switch statement in (both!) validator and relay retrieval functions because we own the cert types, and we can implement them (and in the future update them!) to be compatible with a larger interface argument accepted by GetPayload.

This is basically an instance of the [Robustness/Postel principle](https://en.wikipedia.org/wiki/Robustness_principle): "Among programmers, to produce [compatible functions](https://en.wikipedia.org/wiki/Liskov_substitution_principle), the principle is also known in the form: be contravariant in the input type and covariant in the output type."

samlaf (commented):
Lots of churn here.. partially my fault (see comments below). I think we'll eventually land on the correct/robust api, but don't think we're there yet.

api/clients/v2/payloadretrieval/validator_payload_retriever.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/Layr-Labs/eigenda/api/clients/v2/coretypes"
 	"github.com/Layr-Labs/eigenda/api/clients/v2/validator"
 	"github.com/Layr-Labs/eigenda/api/clients/v2/verification"
+	core "github.com/Layr-Labs/eigenda/core/v2"
```
litt3, line 11: You're importing `"github.com/Layr-Labs/eigenda/core/v2"` twice, under two different aliases.

Curious that go with all of its opinionated import rules doesn't complain about this

litt3 (approved):

api/clients/v2/payloadretrieval/validator_payload_retriever.go
samlaf: I'm also surprised actually. I think go vet or some linter (revive?) will catch this though. Which I added as part of https://github.com/Layr-Labs/eigenda/pull/1604 (revive is not turned on though.. because it was complaining about too many things that I didn't feel like fixing lol)

ethenotethan: [97e1641](https://github.com/Layr-Labs/eigenda/pull/1607/commits/97e1641e24de5dbfeb50f7ff41cb6e23529f4ab9)

api/clients/v2/payloadretrieval/relay_payload_retriever.go
ethenotethan: [97e1641](https://github.com/Layr-Labs/eigenda/pull/1607/commits/97e1641e24de5dbfeb50f7ff41cb6e23529f4ab9)

api/clients/v2/payloadretrieval/relay_payload_retriever.go
ethenotethan: [97e1641](https://github.com/Layr-Labs/eigenda/pull/1607/commits/97e1641e24de5dbfeb50f7ff41cb6e23529f4ab9)

api/clients/v2/coretypes/eigenda_cert.go
```diff
@@ -95,6 +95,18 @@ type EigenDACert interface {
 	Version() CertificateVersion
 }
 
+
+type RetrievableEigenDACert interface {
```
litt3, line 99: Given that there is both `EigenDACert` and `RetrievableEigenDACert`, I think a doc explaining the purpose of each would be helpful. i.e. why do these need to be separate?

TBH I don't understand the benefit of having two interfaces here I read @samlaf 's [comment](https://github.com/Layr-Labs/eigenda/pull/1607#discussion_r2111549815), but I'm not fully following. Why can't we just have an `EigenDACert` interface, which has all of these methods?

I'm approving, since I don't see anything necessary wrong with the current approach, so if you guys have agreement then that works for me too

litt3 (approved):

api/clients/v2/examples/example_relay_retrieval_test.go
samlaf: You're just converting the interface to the other interface, not serializing and deserializing. How will this work in proxy, where we receive an rlp serialized cert for example? We'll need a function to parse those bytes to a RetrievableCert interface right?

api/clients/v2/coretypes/eigenda_cert.go
ethenotethan: Can we please just get this through and think about a more optimised solution down the line?

samlaf: We can merge this. I'll make the proposed fix I'm thinking in another PR. I'll approve once you resolve the conflicts. @litt3 let's discuss on monday

PR #1611, opened 2025-05-28
chore: add proxy to monorepo

samlaf:
Closes DAINT-537
Closes DAINT-586

Note: We should merge this PR as a normal merge (not squash merge) to keep proxy history.
One nice consequence of this that I just realized is that we keep the git blame history:
image

About time we merge proxy into monorepo. This will greatly help with testing, releases, maintainability.
Pulled in ("vendoring merge") the proxy history using
```
git subtree add --prefix=api/proxy https://github.com/Layr-Labs/eigenda-proxy.git main
```
so that now the proxy codebase sits under `api/proxy` in the monorepo.
We can merge the last few open PRs in proxy and still pull them in using `git subtree pull --prefix=api/proxy https://github.com/Layr-Labs/eigenda-proxy.git main` feature, but after that we should sunset the proxy repo and migrate everything here. 

Note: could have used `--squash` option to squash the proxy's history... but decided to keep the history. Not sure what we want here. Both will affect our next release differently, in ways that I'm not super clear on.

## Note to reviewers

This PR contains 150+ files changed, but most of them are just pulled from proxy. I doubt I got everything right, so there might be a bit of "please trust me, and let's fix issues as the come after merging this", but would suggest reviewers to:
1. look at the diff between `api/proxy` and eigenda-proxy repo: `git diff refs/remotes/eigenda-proxy/main chore--add-proxy-to-monorepo:api/proxy` (not as small as I would have liked because lots of deleted stuff and package name changes etc... :\)
2. scrutinize every single file outside of `api/proxy` (there's only ~10 of them). Please nitpick the hell out of them, to make sure I haven't broken anything: makefile, ci, tests, etc.

The changes TLDR are:
1. moved proxy's github workflows to the root workflows dir. Kept proxy CI separate for now to keep diff of this PR small, but eventually would want to merge all CI together
2. got rid of proxy's Dockerfile, and instead added proxy as a target to the shared Dockerfile used to build every other service in the monorepo (`make docker-release-build` now also builds the proxy).
3. copied the linter configuration from proxy over to monorepo, which had a LOT more linters enabled than the default 4/5 that monorepo was using. Surprisingly not that many things required changes, but this explains the few golang files outside of api/proxy that are changed in this PR.


dmanc:
If we merge this I think all the commits will get squashed so might lose the entire history. Not sure if that's okay or not

dmanc:
> > If we merge this I think all the commits will get squashed so might lose the entire history. Not sure if that's okay or not
> 
> Oh good point! We don't HAVE to squash and merge though right? We can just do a normal merge.

cc: @anupsv I think we may have to change some github setting

anupsv:
> > > > If we merge this I think all the commits will get squashed so might lose the entire history. Not sure if that's okay or not
> > > 
> > > 
> > > Oh good point! We don't HAVE to squash and merge though right? We can just do a normal merge.
> > 
> > 
> > cc: @anupsv I think we may have to change some github setting
> 
> I don't see that restriction being enforced on branch protection. Let me see if there's other places.

Ok, i fixed it. You might need to reapprove.

samlaf:
> +100 to doing this but we should definitely file some issues for tracking for fast follow ups; eg:
> 
> * Conforming proxy to the service resource schema used for other eigenda services
> * Test incorporation
> * E2E tests and integration test coalescing

Created some issues for this in https://linear.app/eigenlabs/project/move-proxy-to-monorepo-5d31c0b7c4b2/issues

api/proxy/resources/g1.point
samlaf, line 1: Do we want these in here? Or do we have these resources somewhere else in the monorepo?

api/proxy/cmd/server/entrypoint.go
```diff
@@ -0,0 +1,100 @@
...
+
+	if err := cfg.Check(); err != nil {
+		return err
+	}
+	configString, err := cfg.StoreBuilderConfig.ToString()
+	if err != nil {
+		return fmt.Errorf("convert config json to string: %w", err)
+	}
+
+	log.Infof("Initializing EigenDA proxy server with config (\"*****\" fields are hidden): %v", configString)
```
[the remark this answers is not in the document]
samlaf: We already have this so dismissing. We should probably move to another model where all secrets are in secret and not in config.... but should do that as part of a future PR.

dmanc:
Something that's a bit awkward is this but not sure we can fix it 
image

.dockerignore
```diff
@@ -1,6 +1,11 @@
...
+# because our main shared Dockerfile uses a `COPY . .` in the base-builder stage.
 node_modules
 testdata
 data
 bin
 ./contracts/out
 ./contracts/cache
+**/*.md
+.dockerignore
+**/Dockerfile
```
dmanc: I noticed there's some pngs and pdfs introduced, should we add that to dockerignore?

Dockerfile
```diff
@@ -111,7 +111,17 @@ RUN --mount=type=cache,target=/go/pkg/mod \
     go build -ldflags="-X main.version=${SEMVER} \
                        -X main.gitCommit=${GITCOMMIT} \
                        -X main.gitDate=${GITDATE}" \
-      -o ./bin/blobapi ./cmd/blobapi
+        -o ./bin/blobapi ./cmd/blobapi
+
+# Proxy build stage
+FROM common-builder AS proxy-builder
```
dmanc, line 117: One concern I have with this dockerfile is that we do `COPY . .`. When I built the proxy it was 100MB image due to including the points, so I think it's going to make all the docker images 100MB now.

disperser/dataapi/prometheus/api.go
```diff
@@ -29,7 +29,7 @@ var _ Api = (*prometheusApi)(nil)
 func NewApi(config Config) (*prometheusApi, error) {
 	var err error
 	clientOnce.Do(func() {
-		roundTripper := promconfig.NewBasicAuthRoundTripper(config.Username, promconfig.Secret(config.Secret), "", "", api.DefaultRoundTripper)
+		roundTripper := promconfig.NewBasicAuthRoundTripper(promconfig.NewInlineSecret(config.Username), promconfig.NewInlineSecret(config.Secret), api.DefaultRoundTripper)
```
dmanc, line 32: What does this do?

Dockerfile
samlaf: That's only for the builder image. The actual images produced at the end just contain the binary which they copy from their respective builder image, so they are slim.

disperser/dataapi/prometheus/api.go
samlaf: I'm actually not 100% sure why it happened (maybe some new dependency from proxy?) but the prometheus module was bumped to a more recent version, and the method NewBasicAuthRoundTripper was different. Now their usernames are kept as secrets.
I'm like 95% sure this doesn't change the underlying semantics, and is just a syntax change.

.dockerignore
samlaf: [92da33a](https://github.com/Layr-Labs/eigenda/pull/1611/commits/92da33afaf0a60259f583f16240475b7824fe66b)

Eventually we prob want a reverse ignore pattern? Like ignore everything and then un-ignore only the stuff we want (go files... and that's it?)

Dockerfile
dmanc: oh i see, got it

bxue-l2:
Why it Is under api/

litt3 (approved):

PR #1613, opened 2025-05-29
feat: protobuf payment state api quorum specific variant

hopeyen:

Added a new request type for payment state. This is to avoid the breaking change complain from buf formats, and later we will migrate the current call to the new payment state API with quorum specifications. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -141,6 +164,30 @@ message GetPaymentStateReply {
   bytes onchain_cumulative_payment = 5;
 }
 
+// GetQuorumSpecificPaymentStateReply contains the payment state of an account.
+message GetQuorumSpecificPaymentStateReply {
+  // payment vault parameters with per-quorum configurations
+  PaymentVaultParams payment_vault_params = 1;
+  // off-chain account reservation usage records
+  map<uint32, PeriodRecord> period_records = 2;
```
samlaf, line 172: what is this uint32? Guessing its a quorum number. needs to be properly documented

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -34,6 +34,16 @@ service Disperser {
...
+
+  // GetQuorumSpecificPaymentState is a utility method to get the payment state of a given account, at a given disperser.
+  // EigenDA's payment system for v2 is currently centralized, meaning that each disperser does its own accounting.
+  // A client wanting to disperse a blob would thus need to synchronize its local accounting state with that of the disperser.
+  // That typically only needs to be done once, and the state can be updated locally as the client disperses blobs.
+  // The accounting rules are simple and can be updated locally, but periodic checks with the disperser can't hurt.
+  //
+  // For an example usage, see how our disperser_client makes a call to this endpoint to populate its local accountant struct:
+  // https://github.com/Layr-Labs/eigenda/blob/6059c6a068298d11c41e50f5bcd208d0da44906a/api/clients/v2/disperser_client.go#L298
+  rpc GetQuorumSpecificPaymentState(GetQuorumSpecificPaymentStateRequest) returns (GetQuorumSpecificPaymentStateReply) {}
```
samlaf, line 46: Think renaming to `GetPaymentStateQuorumSpecific` might be cleanest. Any IDE (or maybe even grpcurl reflection) that will list rpc endpoints alphabetically will show GetPaymentState and GetPaymentStateQuorumSpecific next to each other, which isn't true for GetQuorumSpecificPaymentState.

api/proto/disperser/v2/disperser_v2.proto
hopeyen: added documentation

disperser/apiserver/server_v2.go
```diff
@@ -367,3 +367,8 @@ func (s *DispersalServerV2) GetPaymentState(ctx context.Context, req *pb.GetPaym
 	}
 	return reply, nil
 }
+
+// TODO(hopeyen): separate this into a subsequent PR
+func (s *DispersalServerV2) GetQuorumSpecificPaymentState(ctx context.Context, req *pb.GetQuorumSpecificPaymentStateRequest) (*pb.GetQuorumSpecificPaymentStateReply, error) {
+	return nil, api.NewErrorUnimplemented()
```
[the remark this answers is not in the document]
hopeyen: added description in protobuf def and PR description

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -27,13 +27,30 @@ service Disperser {
...
+
+  // GetPaymentStateQuorumSpecific is a utility method to get the payment state of a given account, at a given disperser.
+  // EigenDA's dispersers and validators each does its own accounting for reservation usages, indexed by the account and quorum id.
+  // A client wanting to disperse a blob would thus need to synchronize its local accounting state with the disperser it disperse from.
+  // That typically only needs to be done once, and the state can be updated locally as the client disperses blobs.
+  // The accounting rules are simple and can be updated locally, but periodic checks with the disperser can't hurt.
+  //
+  // For an example usage, see how our disperser_client makes a call to this endpoint to populate its local accountant struct:
+  // https://github.com/Layr-Labs/eigenda/blob/6059c6a068298d11c41e50f5bcd208d0da44906a/api/clients/v2/disperser_client.go#L298
+  rpc GetPaymentStateQuorumSpecific(GetPaymentStateQuorumSpecificRequest) returns (GetPaymentStateQuorumSpecificReply) {}
```
ian-shim, line 53: nit: What do you think about `GetPaymentStateByQuorum` or `GetQuorumPaymentState`?

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -141,6 +171,30 @@ message GetPaymentStateReply {
   bytes onchain_cumulative_payment = 5;
 }
 
+// GetPaymentStateQuorumSpecificReply contains the payment state of an account.
+message GetPaymentStateQuorumSpecificReply {
+  // payment vault parameters with per-quorum configurations
+  PaymentVaultParams payment_vault_params = 1;
+  // period_records maps quorum IDs to the off-chain account reservation usage records for the current and next two periods
+  map<uint32, PeriodRecords> period_records = 2;
```
ian-shim, line 179: Isn't the request filtering the payment state by quorum? What does it return results for all quorums?

api/proto/disperser/v2/disperser_v2.proto
samlaf: I asked to rename to this here: https://github.com/Layr-Labs/eigenda/pull/1613#discussion_r2113455285
I don't like `GetQuorumPaymentState` for the reasons listed in that comment, but I'm fine with `GetPaymentStateByQuorum`

api/proto/disperser/v2/disperser_v2.proto
samlaf: please add commit links, makes the reviewers' job much much easier (and the PR author's job not that much more harder.. take 15 seconds more)

api/proto/disperser/v2/disperser_v2.proto
hopeyen: The intended behavior is to return all the quorum usages they have reservations for, not filter by quorum. The `QuorumSpecific` really means that the results returned are quorum specific, not for a specific quorum. I can see how this leads to misunderstanding, so I'm okay to use a different name
There's arguments you can make for either behavior, but I think this is the easiest on the client side, so they don't need to know quorum numbers before they request for the state or make separate requests; and this behavior is closer to the behavior of the current `GetPaymentState`

disperser/apiserver/server_v2.go
```diff
@@ -8,6 +8,7 @@ import (
 	"sync/atomic"
 	"time"
 
+	gethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/prometheus/client_golang/prometheus"
```
dmanc, line 13: nit: can we remove this new line

core/auth/v2/authenticator.go
```diff
@@ -4,9 +4,10 @@ import (
 	"crypto/sha256"
 	"errors"
 	"fmt"
+	"time"
+
 	pb "github.com/Layr-Labs/eigenda/api/grpc/disperser/v2"
 	"github.com/Layr-Labs/eigenda/api/hashing"
-	"time"
```
dmanc, line 11: nit: can we also remove this new line

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -27,13 +27,30 @@ service Disperser {
 
   // GetPaymentState is a utility method to get the payment state of a given account, at a given disperser.
   // EigenDA's payment system for v2 is currently centralized, meaning that each disperser does its own accounting.
+  // As reservation moves to be quorum specific and served by permissionless dispersers, GetPaymentState will soon be deprecate
```
dmanc, line 30:
```suggestion
  // As reservation moves to be quorum specific and served by permissionless dispersers, GetPaymentState will soon be deprecated
```

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -27,13 +27,30 @@ service Disperser {
...
+  // in replacement of GetPaymentStateQuorumSpecific to include more specifications. During the endpoint migration time, the response
+  // take quorum 0 for the global parameters, and the most retrictive reservation parameters of a user across quorums. For
```
dmanc, line 32:
```suggestion
  // uses quorum 0 for the global parameters, and the most retrictive reservation parameters of a user across quorums. For
```

api/proto/disperser/v2/disperser_v2.proto
dmanc, line 30: Do we have a procedure for deprecating grpc methods?

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -27,13 +27,30 @@ service Disperser {
...
   // That typically only needs to be done once, and the state can be updated locally as the client disperses blobs.
   // The accounting rules are simple and can be updated locally, but periodic checks with the disperser can't hurt.
   //
   // For an example usage, see how our disperser_client makes a call to this endpoint to populate its local accountant struct:
   // https://github.com/Layr-Labs/eigenda/blob/6059c6a068298d11c41e50f5bcd208d0da44906a/api/clients/v2/disperser_client.go#L298
   rpc GetPaymentState(GetPaymentStateRequest) returns (GetPaymentStateReply) {}
+
+  // GetPaymentStateQuorumSpecific is a utility method to get the payment state of a given account, at a given disperser.
+  // EigenDA's dispersers and validators each does its own accounting for reservation usages, indexed by the account and quorum id.
+  // A client wanting to disperse a blob would thus need to synchronize its local accounting state with the disperser it disperse from.
```
dmanc, line 47:
```suggestion
  // A client wanting to disperse a blob would thus need to synchronize its local accounting state with the disperser it dispersed from.
```

api/proto/disperser/v2/disperser_v2.proto
dmanc: What about `GetPaymentStateFromQuorums`?

core/auth/v2/auth_test.go
```diff
@@ -2,13 +2,14 @@ package v2_test
...
-	"github.com/Layr-Labs/eigenda/api/hashing"
-	"github.com/Layr-Labs/eigenda/common/replay"
 	"math/big"
 	"testing"
 	"time"
 
+	disperser_rpc "github.com/Layr-Labs/eigenda/api/grpc/disperser/v2"
+	"github.com/Layr-Labs/eigenda/api/hashing"
+	"github.com/Layr-Labs/eigenda/common/replay"
+
```
dmanc, line 12: nit: remove new line

api/proto/disperser/v2/disperser_v2.proto
hopeyen: Since we are not allowing users to specify quorums in this request, maybe we should remove that from the name altogether. what about `GetPaymentStateVerbose`?

core/auth/v2/auth_test.go
hopeyen: ty! above nits and grammar errors addressed in [commit](https://github.com/Layr-Labs/eigenda/pull/1613/commits/7193742c451cd73dd126aaa42cb102f59a5420b1)

api/proto/disperser/v2/disperser_v2.proto
hopeyen: no written procedure afaik, so the current plan is for disperser to serve both old and new endpoints. I think we should deprecate this before releasing permissionless dispersers. I can bring this up in the standup for making a standard procedure

core/auth/v2/authenticator.go
```diff
@@ -67,20 +67,31 @@ func (*authenticator) AuthenticateBlobRequest(header *core.BlobHeader, signature
...
 // See implementation of BlobRequestSigner.SignPaymentStateRequest for more details
 func (a *authenticator) AuthenticatePaymentStateRequest(accountAddr common.Address, request *pb.GetPaymentStateRequest) error {
-	sig := request.GetSignature()
+	return a.authenticatePaymentRequest(accountAddr, request.GetSignature(), request.GetTimestamp())
+}
+
+// AuthenticatePaymentStateQuorumSpecificRequest verifies the signature of the quorum specific payment state request
+// The signature is signed over the byte representation of the account ID and requestHash
+// See implementation of BlobRequestSigner.SignPaymentStateRequest for more details
+func (a *authenticator) AuthenticatePaymentStateQuorumSpecificRequest(accountAddr common.Address, request *pb.GetPaymentStateQuorumSpecificRequest) error {
```
anupsv: This fn name will also need to change based on above comments.

anupsv (dismissed):
looks good over all. Need the naming conflict to resolve.

core/auth/v2/authenticator.go
hopeyen: yeah, I don't think there's an agreement on the name yet:(

ian-shim (dismissed):

api/proto/disperser/v2/disperser_v2.proto
samlaf: How about `GetPaymentStateForAllQuorums` then?

core/auth/v2/authenticator.go
```diff
@@ -67,20 +67,31 @@ func (*authenticator) AuthenticateBlobRequest(header *core.BlobHeader, signature
 // The signature is signed over the byte representation of the account ID and requestHash
 // See implementation of BlobRequestSigner.SignPaymentStateRequest for more details
 func (a *authenticator) AuthenticatePaymentStateRequest(accountAddr common.Address, request *pb.GetPaymentStateRequest) error {
```
dmanc, line 70: https://go.dev/wiki/Deprecated, maybe we add in the go doc something like:

Deprecated: Use AuthenticatePaymentStateForAllQuorumsRequest instead.

dmanc (approved):

api/proto/disperser/v2/disperser_v2.proto
hopeyen: updated to `GetPaymentStateForAllQuorums` ([commit](https://github.com/Layr-Labs/eigenda/pull/1613/commits/0cefc62b4d88d74535b5a35f57bfd8d4e80b0ff7))

core/auth/v2/authenticator.go
hopeyen: added in [commit](https://github.com/Layr-Labs/eigenda/pull/1613/commits/6fbbfe85965c0eca4eb56febabf58dffe21fd799) and [commit](https://github.com/Layr-Labs/eigenda/pull/1613/commits/936c279d50fa5f71e5cbeb27bd51819a17571236) for GetPaymentState API reques

dmanc (approved):

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -27,13 +27,30 @@ service Disperser {
 
   // GetPaymentState is a utility method to get the payment state of a given account, at a given disperser.
   // EigenDA's payment system for v2 is currently centralized, meaning that each disperser does its own accounting.
+  // As reservation moves to be quorum specific and served by permissionless dispersers, GetPaymentState will soon be deprecated
```
samlaf, line 30: why soon? Why not now? When are we planning to deprecate?

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -27,13 +27,30 @@ service Disperser {
...
+  // in replacement of GetPaymentStateForAllQuorums to include more specifications. During the endpoint migration time, the response
+  // uses quorum 0 for the global parameters, and the most retrictive reservation parameters of a user across quorums. For
+  // OnDemand, EigenDA disperser is the only allowed disperser, so it will provide real values for off-chain payments.
+  // For other dispersers, they will provide 0 for offchain payments. A client using non-EigenDA dispersers should not worry about
+  // the zero values for onDemand records.
```
samlaf: It's a bit hard to follow what you are talking about here, as we don't all have the same context that you have in your brain.

Would prob be best to move these details next to the actual Reply value that you are talking about. Unclear what "provide 0 for offchain payments" means here.

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -27,13 +27,30 @@ service Disperser {
...
   // That typically only needs to be done once, and the state can be updated locally as the client disperses blobs.
   // The accounting rules are simple and can be updated locally, but periodic checks with the disperser can't hurt.
   //
   // For an example usage, see how our disperser_client makes a call to this endpoint to populate its local accountant struct:
   // https://github.com/Layr-Labs/eigenda/blob/6059c6a068298d11c41e50f5bcd208d0da44906a/api/clients/v2/disperser_client.go#L298
   rpc GetPaymentState(GetPaymentStateRequest) returns (GetPaymentStateReply) {}
+
+  // GetPaymentStateForAllQuorums is a utility method to get the payment state of a given account, at a given disperser.
+  // EigenDA's dispersers and validators each does its own accounting for reservation usages, indexed by the account and quorum id.
+  // A client wanting to disperse a blob would thus need to synchronize its local accounting state with the disperser it dispersed from.
```
samlaf:
```suggestion
  // A client wanting to disperse a blob would thus need to synchronize its local accounting state with the disperser it plans to disperse to.
```

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -127,6 +144,17 @@ message GetPaymentStateRequest {
...
 
+// GetPaymentStateForAllQuorumsRequest contains parameters to query the payment state of an account.
+message GetPaymentStateForAllQuorumsRequest {
+  // The ID of the account being queried. This account ID is an eth wallet address of the user.
+  string account_id = 1;
+  // Timestamp of the request in nanoseconds since the Unix epoch. If too far out of sync with the server's clock,
+  // request may be rejected.
+  uint64 timestamp = 2;
+  // Signature over the payment account ID and timestamp.
+  bytes signature = 3;
```
samlaf, line 155: why was GetPaymentStateRequest.signature only signing over account_id? Why is this one signing over the timestamp as well?

api/clients/v2/accountant_test.go
```diff
@@ -467,3 +468,128 @@ func isRotation(arrA, arrB []uint64) bool {
...
+		assert.Equal(t, uint32(1), accountant.periodRecords[0].Index)
+		assert.Equal(t, uint64(150), accountant.periodRecords[0].Usage)
+		assert.Equal(t, uint32(0), accountant.periodRecords[1].Index)
+		assert.Equal(t, uint64(0), accountant.periodRecords[1].Usage)
+		assert.Equal(t, uint32(0), accountant.periodRecords[2].Index)
+		assert.Equal(t, uint64(0), accountant.periodRecords[2].Usage)
+	})
+
+	t.Run("successful set payment state with minimal fields", func(t *testing.T) {
+		state := &disperser_rpc.GetPaymentStateReply{
```
samlaf, line 563: this is the deprecated proto type. shouldnt you be using the new one? Why are you adding these tests now? Did you just now notice that they were missing? Shouldnt we also add tests for the new proto types?

hopeyen: yes, I should be using the new one but the API isn't actually implemented in the server yet and the accountant doesn't call out to it either. I noticed these unit tests were missing so wanted to add them, and planned on updating more in this upcoming [pr](https://github.com/Layr-Labs/eigenda/pull/1619/files). Do you prefer if I remove these and just add tests in the next pr?

api/proto/disperser/v2/disperser_v2.proto
hopeyen: Planning to deprecate this after the traffic of this method goes to zero. Alternatively can deprecate this in a later PR after the client has been updated
[Relevant comment](https://github.com/Layr-Labs/eigenda/pull/1613#discussion_r2116693848)

It is a separate PR because I was told to split PR into smaller pieces for easier review process. This PR is meant to contain the new Protobuf API

api/proto/disperser/v2/disperser_v2.proto
hopeyen: `AuthenticatePaymentStateRequest` is also signing over timestamp but wasn't properly documented. I updated its documentation ([commit](https://github.com/Layr-Labs/eigenda/pull/1613/commits/6278524bff68140db2f81b5ffee8e8be8a94dac6)). just didn't want to touch it since it would get deprecated soon

api/proto/disperser/v2/disperser_v2.proto
hopeyen: Okay, I moved these explanations to `GetPaymentStateForAllQuorumsReply` and rephrase a bit. [commit](https://github.com/Layr-Labs/eigenda/pull/1613/commits/6278524bff68140db2f81b5ffee8e8be8a94dac6)

api/proto/disperser/v2/disperser_v2.proto
samlaf: Makes sense. I think it should be deprecated as soon as we have the new client available. Deprecating is a signal for people to switch over. If they don't then likely the traffic would never go to zero.
We shouldn't deprecate after traffic goes to zero. We should deprecate for traffic to go to zero, and then just remove the rpc method entirely.

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -169,7 +169,10 @@ message GetPaymentStateReply {
   bytes onchain_cumulative_payment = 5;
 }
 
-// GetPaymentStateForAllQuorumsReply contains the payment state of an account.
+// GetPaymentStateForAllQuorumsReply contains the payment state of an account. EigenLabs disperser is the only disperser that allows
+// for ondemand usages, and it will provide the latest on-demand offchain payment records for the request account.
+// Other dispersers will refuse to serve ondemand requests and serve 0 for off-chain on-demand payment usage (`cumulative_payment`). A client using
```
samlaf: what is a "payment records"? Do you mean `period_records`? I don't see payment record anywhere. Also wdym by "serve 0 for off-chain on-demand payment usage"? What is 0? Is it all fields in the Reply that are set to 0? Or only certain ones? Better to talk about the exact fields on this struct to be precise.

api/clients/v2/accountant_test.go
samlaf: no its fine

api/proto/disperser/v2/disperser_v2.proto
hopeyen: iterated based on the questions [commit](https://github.com/Layr-Labs/eigenda/pull/1613/commits/19d6f265633405685ad191668f5faf37e10a2bd8)

samlaf: Much much clearer. Thanks hope :)

PR #1617, opened 2025-05-29
feat: add structured errors to verifier client

samlaf:
This is needed to return proper errors in proxy. For example we want to return a 418 HTTP error to rollup nodes when cert verification fails, such that they can filter out the invalid certs from their derivation pipeline.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/verification/errors.go
+++ b/api/clients/v2/verification/errors.go
@@ -1,0 +1,42 @@
+package verification
+
+import "github.com/Layr-Labs/eigenda/api/clients/v2/coretypes"
+
+// CertVerifierInputError represents a 4xx-like error (invalid input, misuse, unsupported, etc.)
+type CertVerifierInputError struct {
+	Msg string
+}
+
+func (e *CertVerifierInputError) Error() string {
+	return e.Msg
+}
+
+// CertVerifierInternalError represents a 5xx-like error (unexpected, internal, infra, etc.)
+type CertVerifierInternalError struct {
+	Msg string
+	// Err is optional and only present if an underlying error is available.
+	// Note that we only provide this as a convenience for logging and debugging.
+	// Error is NOT part of our public API, so don't match on internal errors,
+	// as these errors may change in the future.
+	Err error
+}
+
+func (e *CertVerifierInternalError) Error() string {
+	if e.Err != nil {
+		return e.Msg + ": " + e.Err.Error()
+	}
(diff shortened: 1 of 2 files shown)


litt3 (dismissed):

anupsv (dismissed):

api/clients/v2/verification/errors.go
```diff
@@ -0,0 +1,42 @@
+package verification
+
+import "github.com/Layr-Labs/eigenda/api/clients/v2/coretypes"
+
+// CertVerifierInputError represents a 4xx-like error (invalid input, misuse, unsupported, etc.)
+type CertVerifierInputError struct {
+	Msg string
```
ethenotethan, line 7: is there any merit in capturing the status code enum in this type?

samlaf: wdym by status code? like 400 vs 404 vs 403 etc? I don't think we should return http code from library function calls personally. Perhaps I shouldnt have used 4xx and just sticked to "wrong input error type"?
In any case, I do agree we could go a lot more fine grained than 4xx/5xx/certInvalid coarse grained errors that I defined here, but for now the only thing we really need to know in proxy is whether a cert is (in)valid, so I (at this point) didn't see value in going more fine grained. If ever we do want more fine grained detail to type switch on and perform different behaviors inside proxy (for example retry 5xx errors?), then we should! Not sure if there is any value for 4xx errors though...? Haven't thought about it super deeply yet

litt3 (approved):

litt3 (approved):

api/clients/v2/verification/errors.go
bxue-l2: but if recency check is moved to the client from proxy, we do have more error case

samlaf: @bxue-l2 you're looking at the wrong error. This error is when an input is wrong. Like we weren't even able to call the CertVerifier contract or check rbn recency because what was sent is not even a cert for example. This error represents a bug, not an invalid cert.

You want to look at `CertVerificationFailedError`, which does include the StatusCode returned by the contract. I regret naming it that way.. probably confused you guys. How about we rename it to `InvalidCertError` or something of that sort? I suggest we do this after we move proxy to monorepo (can also move RBN check to client once we do that)

PR #1620, opened 2025-05-29
feat: offchain quorum period record getter

hopeyen:

Add a dynamodb getter for period records with quorum specifications 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


core/meterer/dynamodb_metering_store.go
```diff
@@ -256,30 +256,50 @@ func (s *DynamoDBMeteringStore) RollbackOnDemandPayment(ctx context.Context, acc
...
-	records := [MinNumBins]*pb.PeriodRecord{}
-	for i := 0; i < len(bins) && i < int(MinNumBins); i++ {
-		periodRecord, err := parsePeriodRecord(bins[i])
+	records := make(map[core.QuorumID]*pb.PeriodRecords)
+	for _, item := range items {
+		quorumNumber, periodRecord, err := parsePeriodRecord(item)
 		if err != nil {
-			return [MinNumBins]*pb.PeriodRecord{}, fmt.Errorf("failed to parse bin %d record: %w", i, err)
+			s.logger.Debug("Failed to parse period record", "err", err)
+			continue
```
pschork: Seems like the level should be Error. Why don't we propagate the error any more?

core/meterer/metering_store.go
```diff
@@ -27,8 +27,9 @@ type MeteringStore interface {
 	// RollbackOnDemandPayment rolls back a payment to the previous value
 	RollbackOnDemandPayment(ctx context.Context, accountID gethcommon.Address, newPayment, oldPayment *big.Int) error
 
-	// GetPeriodRecords fetches period records for the given account ID and reservation period
-	GetPeriodRecords(ctx context.Context, accountID gethcommon.Address, reservationPeriod uint64) ([MinNumBins]*pb.PeriodRecord, error)
+	// GetPeriodRecords fetches period records for the given account ID and reservation period across multiple quorums
+	// Consider alternative: singular quorumId and reservationPeriod as an input, require multiple calls
```
dmanc: nit: Is the comment here about the alternative needed?

disperser/apiserver/server_v2.go
```diff
@@ -304,11 +304,24 @@ func (s *DispersalServerV2) GetPaymentState(ctx context.Context, req *pb.GetPaym
 
 	// off-chain account specific payment state
 	now := time.Now().Unix()
-	currentReservationPeriod := meterer.GetReservationPeriod(now, reservationWindow)
-	periodRecords, err := s.meterer.MeteringStore.GetPeriodRecords(ctx, accountID, currentReservationPeriod)
-	if err != nil {
-		s.logger.Debug("failed to get reservation records, use placeholders", "err", err, "accountID", accountID)
+	quorumIds := s.onchainState.Load().getAllQuorumIds()
+	// Constructing the strictest period records across all quorums; a client should migrate to GetPaymentStateForAllQuorums for more precise state
+	// TODO(hopeyen): remove this in a subsequent PR. The logic here is complicated and only temporary
```
dmanc, line 309: What's the condition for removing this logic?

core/meterer/dynamodb_metering_store.go
```diff
@@ -256,30 +256,50 @@ func (s *DynamoDBMeteringStore) RollbackOnDemandPayment(ctx context.Context, acc
...
+		}
+		records[quorumNumber] = &pb.PeriodRecords{
+			Records: []*pb.PeriodRecord{periodRecord},
```
dmanc: The comment for the function says `// Returns an array of PeriodRecords up to numBins in length, with records for each requested quorum.
` but seems we only return a size one array?

core/meterer/dynamodb_metering_store.go
hopeyen: you are right. updated to propagate ([commit](https://github.com/Layr-Labs/eigenda/pull/1620/commits/4ce821b2be755c3bd1f53c369ca144cd379adba0))

core/meterer/metering_store.go
hopeyen: this is like an alternative replacement. Since this API might be used by DataAPI, I was curious if they would like an alternative type

hopeyen: this comment is just considering if other clients, specificially DataAPI, want a different interface; or if other engineers would think the alternative is more performant/optimized in some situations. the one defined here is suited for how disperser would use it

disperser/apiserver/server_v2.go
hopeyen: it is removed in this draft [pr](https://github.com/Layr-Labs/eigenda/pull/1619)

core/meterer/dynamodb_metering_store.go
hopeyen: ah nice catch! fixed in [commit](https://github.com/Layr-Labs/eigenda/pull/1620/commits/ddb8256f4ec865ffe4bd9fa44ae8a8eb52151372) and added some tests

core/meterer/metering_store.go
dmanc: Ah I see, but I guess the comment feels more like a todo to validate if that call pattern is sufficient for DataAPI rather than documentation about how GetPeriodRecords works?

core/meterer/dynamodb_metering_store.go
```diff
@@ -256,30 +256,53 @@ func (s *DynamoDBMeteringStore) RollbackOnDemandPayment(ctx context.Context, acc
...
-	for i := 0; i < len(bins) && i < int(MinNumBins); i++ {
-		periodRecord, err := parsePeriodRecord(bins[i])
+	records := make(map[core.QuorumID]*pb.PeriodRecords)
+	for _, item := range items {
+		quorumNumber, periodRecord, err := parsePeriodRecord(item)
 		if err != nil {
-			return [MinNumBins]*pb.PeriodRecord{}, fmt.Errorf("failed to parse bin %d record: %w", i, err)
+			return nil, fmt.Errorf("failed to parse period record: %w", err)
+		}
+		if existingRecords, exists := records[quorumNumber]; exists {
```
dmanc, line 299: Does it make sense to do initializing upfront like

```
records := make(map[core.QuorumID]*pb.PeriodRecords)
for _, quorum := range quorumNumbers {
    records[quorum] = &pb.PeriodRecords{
        Records: make([]*pb.PeriodRecord, 0, numBins),
    }
}
```

then we only need to append in this loop 

```
    records[quorumNumber].Records = append(records[quorumNumber].Records, periodRecord)
```

core/meterer/dynamodb_metering_store_test.go
```diff
@@ -47,11 +46,6 @@ func setupTest(t *testing.T) *testContext {
 	err = meterer.CreateGlobalReservationTable(clientConfig, tc.globalBinTable)
 	require.NoError(t, err)
 
-	// Register cleanup to remove tables after test completes
-	t.Cleanup(func() {
```
dmanc, line 51: Wouldn't this affect the other tests that depend on it?

core/meterer/dynamodb_metering_store.go
hopeyen: I think so, it is a good optimization we can do. it is just that we are not guaranteed to have `numBins` of items to fill it in with and then client will have to determine check for nil. I was thinking maybe it is better to just return the non-nil results?

core/meterer/dynamodb_metering_store_test.go
hopeyen: none of the current tests depend on each other at the moment; they will each get a setup and cleanup. i think it is generally better to keep each tests isolated

core/meterer/metering_store.go
hopeyen: hmm yeah it is a todo. maybe @pschork can answer this? I can update within this PR to accommodate the usage

core/meterer/dynamodb_metering_store_test.go
dmanc: But wouldn't a test like `TestGetLargestCumulativePayment` no longer get cleaned up after it runs?

core/meterer/dynamodb_metering_store.go
dmanc: ah gotcha

core/meterer/dynamodb_metering_store_test.go
hopeyen: I see, added cleanup to all the tests in this file

dmanc (approved):

core/meterer/metering_store.go
pschork: I think dataAPI can work with current multi quorum results.

pschork (approved):

PR #1622, opened 2025-05-30
chore: documentation for database path configuration

cody-littley:

Create more detailed documentation on configuring database paths.


node/database-paths.md
```diff
@@ -0,0 +1,173 @@
...
+
+The V2 protocol's disk footprint depends on how it is configured.
+
+### Deprecated Configuration: only `NODE_DB_PATH` set
+
+If only `NODE_DB_PATH` is set and `NODE_LITT_DB_STORAGE_PATHS` is not set (not recommended!), then the V2 protocol
+will store its data like this:
+
+```
+${NODE_DB_PATH}
```
rootwarp, line 53: As you mentioned I can see directories like below,

```
${NODE_DB_PATH}
.
├── chunk
└── chunk_v2_litt
    └── chunks
        ├── keymap
        │   └── data
        └── segments
```

Do you think it would help clarify things if you point out the `chunk` directory, which appears to be used for V1?

node/database-paths.md
```diff
@@ -0,0 +1,173 @@
...
+- Stop the validator node.
+- Copy/move the contents of the old `NODE_DB_PATH` to the new intended `NODE_DB_PATH`, e.g. `mv /old/path/ /new/path/`.
+- Update the `NODE_DB_PATH` environment variable to point to the new path.
+- Restart the validator node.
+
+# Changing `NODE_LITT_DB_STORAGE_PATHS`
+
+## Adding a Path
+
+It's possible to add additional paths to `NODE_LITT_DB_STORAGE_PATHS`. This might be useful if want to add
```
litt3: This might be useful if *you* want

node/database-paths.md
```diff
@@ -0,0 +1,173 @@
...
+
+If using the legacy fallback `NODE_DB_PATH` for V2 data storage, the validator software stores its data at
+`${NODE_DB_PATH}/chunk_v2_litt/`. The `chunk_v2_litt` directory is hard coded and always added to the path.
+But if you are using the new `NODE_LITT_DB_STORAGE_PATHS` flag, the `chunk_v2_litt` directory is NOT added to the path.
+
+In order to remedy this, you will need to move the contents of `${NODE_DB_PATH}/chunk_v2_litt/` to the location where
+you want to store the V2 data. For example, if you want to store the V2 data in `${volume1}`, then you would
+do `mv ${NODE_DB_PATH}/chunk_v2_litt/ ${volume1}/`.
+
+- Stop the validator node.
```
dmanc, line 170: Wouldn't the validator get ejected if this procedure took longer than the SLA for signing (how likely is that)?

dmanc (approved):

node/database-paths.md
cody-littley: This is described in the section above titled `## V1 Protocol`.

```
## V1 Protocol

The V1 protocol's disk footprint looks something like this:

${NODE_DB_PATH}
├── chunk
│   ├── 000001.log
│   ├── CURRENT
│   ├── LOCK
│   ├── LOG
│   └── MANIFEST-000000

The `chunk` directory is created by the V1 software inside the directory specified by `NODE_DB_PATH`. Inside
the `chunk` directory are files maintained by the V1 data storage engine (i.e. `LevelDB`).
```

node/database-paths.md
cody-littley: fixed

node/database-paths.md
cody-littley: Yes. I'm not sure how long the window is. Presumably operators will need to restart their nodes every once in a while for maintenance purposes, so we should have a little bit of a grace period baked into ejection.

LittDB doesn't take long to load from disk, so hopefully it's possible to do this sort of restart very rapidly. Although this is something I don't think we've specifically measured yet.

PR #1623, opened 2025-05-30
feat: slow traffic ramp up

cody-littley:

When the traffic generator starts, slowly ramp up traffic.


cody-littley:
Closing this one, will cherry pick instead

PR #1630, opened 2025-06-02
Traffic slow startup rc4

cody-littley:

Add a ramp up period for the traffic generator.

### DIFF
--- a/common/variable_ticker.go
+++ b/common/variable_ticker.go
@@ -230,7 +230,7 @@ func (t *VariableTicker) computePeriod() {
 }
 
 // computeInflectionFrequency handles an edge case when starting from a very low frequency. Suppose we start at 0.0 hz
-// and are accelerating. At the moment we start accelerating, the frequency is zero period is infinite (1/0=∞),
+// and are accelerating. At the moment we start accelerating, the frequency is zero and the period is infinite (1/0=∞),
 // which is obviously not what we want. The "inflection frequency" is an adjusted frequency that will cause us to sleep
 // for a more reasonable time. Specifically, it causes us to sleep long enough so that at the moment we wake up,
 // the frequency at the moment we wake up will produce a period equal to the time we just slept.
--- a/common/variable_ticker.go
+++ b/common/variable_ticker.go
@@ -63,12 +63,17 @@ func NewVariableTickerWithPeriod(ctx context.Context, period time.Duration) (*Va
 func NewVariableTickerWithFrequency(ctx context.Context, frequency float64) *VariableTicker {
 	ctx, cancel := context.WithCancel(ctx)
 
+	currentPeriod := time.Duration(0)
+	if frequency > 0 {
+		currentPeriod = time.Duration(float64(time.Second) / frequency)
+	}
+
 	ticker := &VariableTicker{
 		ctx:              ctx,
 		close:            cancel,
 		acceleration:     0.0,
 		currentFrequency: frequency,
-		currentPeriod:    time.Duration(1.0 / frequency),
+		currentPeriod:    currentPeriod,
 		targetFrequency:  frequency,
 		tickChan:         make(chan struct{}),
 		controlChan:      make(chan any, 2),
@@ -124,6 +129,8 @@ func (t *VariableTicker) run() {
 	timer := time.NewTimer(t.currentPeriod)
 	defer timer.Stop()
 
+	previousTickTime := time.Now()
+
 	for {
 		// Check for control messages to update the ticker's configuration.
 		select {
@@ -133,8 +140,10 @@ func (t *VariableTicker) run() {
 			// No control message received, continue with the current configuration.
 		}
 
-		if t.currentFrequency == 0 {
-			// If the current frequency is zero, never tick.
+		t.computePeriod()
+		if t.currentPeriod == 0 {
+			// Period 0 is a proxy for an infinite period, do not tick.
+			fmt.Printf("VariableTicker: current period is 0, not ticking\n") // TODO
 			continue
 		}
 
@@ -145,8 +154,17 @@ func (t *VariableTicker) run() {
 			return
 		}
 
-		// Wait until it is time to send the next tick.
-		sleepTime := t.computeSleepTime()
+		now := time.Now()
+		elapsed := now.Sub(previousTickTime)
+		sleepTime := t.currentPeriod - elapsed
+		previousTickTime = now
+		if sleepTime < 0 {
+			// If ticks are requested less often than the configured frequency, no need to sleep.
+			continue
+		}
+
+		fmt.Printf("frequency: %f, period: %v, sleep time: %v\n", t.currentFrequency, t.currentPeriod, sleepTime) // TODO
+
 		timer.Reset(sleepTime)
 		select {
 		case <-timer.C:
@@ -177,10 +195,11 @@ func (t *VariableTicker) handleControlMessage(msg any) {
 	t.acceleration = acceleration
 }
 
-// computeSleepTime calculates the time to sleep until the next tick based on the current and target periods.
-func (t *VariableTicker) computeSleepTime() time.Duration {
+// computePeriod updates the current period based on configured frequency and acceleration
+func (t *VariableTicker) computePeriod() {
 	if t.currentFrequency == t.targetFrequency {
-		return t.currentPeriod
+		// shortcut, don't recompute period if the period is already correct
+		return
 	}
 
 	elapsedSinceAnchorTime := time.Since(t.anchorTime)
@@ -204,6 +223,9 @@ func (t *VariableTicker) computeSleepTime() time.Duration {
 		}
 	}
 
-	t.currentPeriod = time.Duration(1.0 / currentFrequency)
-	return t.currentPeriod
+	if currentFrequency == 0 {
+		t.currentPeriod = 0
+	} else {
+		t.currentPeriod = time.Duration(float64(time.Second) / currentFrequency)
+	}
 }
--- a/common/variable_ticker.go
+++ b/common/variable_ticker.go
@@ -215,7 +215,7 @@ func (t *VariableTicker) computePeriod() {
 		}
 	} else {
 		// We are above the target frequency.
-		t.currentFrequency = t.targetFrequency - (t.acceleration * elapsedSinceAnchorTime.Seconds())
+		t.currentFrequency = t.currentFrequency - (t.acceleration * elapsedSinceAnchorTime.Seconds())
 		if t.currentFrequency < t.targetFrequency {
 			// If we over shoot, adopt the target frequency.
 			t.currentFrequency = t.targetFrequency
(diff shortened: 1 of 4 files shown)


common/variable_ticker.go
```diff
@@ -0,0 +1,267 @@
...
+	acceleration float64
+
+	// The current frequency of the ticker, in HZ.
+	currentFrequency float64
+
+	// Matches currentFrequency. currentFrequency is the "source of truth", but we cache the period to avoid
+	// recomputing it each tick.
+	currentPeriod time.Duration
+
+	// The previous period held by this ticker the last time its configuration was changed.
```
litt3: doc says "period", member is named "frequency"

common/variable_ticker.go
```diff
@@ -0,0 +1,267 @@
...
+	// The channel that produces an output every time the ticker ticks.
+	tickChan chan struct{}
+
+	// The channel used to send control messages to main ticker loop.
+	controlChan chan any
+}
+
+// frequencyUpdate is a control message to update the target frequency of the ticker.
+type frequencyUpdate struct {
+	// The target period to move towards.
```
litt3: doc says "period", field is "frequency"

common/variable_ticker.go
```diff
@@ -0,0 +1,267 @@
...
+		select {
+		case msg := <-t.controlChan:
+			t.handleControlMessage(msg)
+		default:
+			// No control message received, continue with the current configuration.
+		}
+
+		t.computePeriod()
+		if t.currentPeriod == 0 {
+			// Period 0 is a proxy for an infinite period, do not tick.
```
litt3: you explicitly forbid period of 0 in `SetTargetPeriod`. Is this comment accurate?

common/variable_ticker.go
```diff
@@ -0,0 +1,267 @@
...
+			// If we over shoot, adopt the target frequency.
+			t.currentFrequency = t.targetFrequency
+		} else {
+			// When speeding up, substitute the current frequency with the inflection frequency.
+			// This is to avoid sleeping for a very long time when starting from a low frequency.
+			t.currentFrequency = t.computeInflectionFrequency()
+		}
+	} else {
+		// We are above the target frequency.
+		t.currentFrequency = t.currentFrequency - (t.acceleration * elapsedSinceAnchorTime.Seconds())
```
litt3: Shouldn't this be `t.anchorFrequency - (t.acceleration * elapsedSinceAnchorTime.Seconds())`?

common/variable_ticker.go
```diff
@@ -0,0 +1,267 @@
...
+
+	t.controlChan <- &frequencyUpdate{
+		targetFrequency: frequency,
+	}
+
+	return nil
+}
+
+func (t *VariableTicker) SetTargetFrequency(frequency float64) {
+	t.controlChan <- &frequencyUpdate{
```
litt3, line 122: Shouldn't you check that set frequency is > 0?

common/variable_ticker.go
```diff
@@ -0,0 +1,267 @@
...
+		// Check for control messages to update the ticker's configuration.
+		select {
+		case msg := <-t.controlChan:
+			t.handleControlMessage(msg)
+		default:
+			// No control message received, continue with the current configuration.
+		}
+
+		t.computePeriod()
+		if t.currentPeriod == 0 {
```
litt3, line 159: won't this result in a busy loop?

common/variable_ticker.go
```diff
@@ -0,0 +1,267 @@
...
+// move towards the target period at the rate of acceleration per second. If acceleration is zero,
+// the ticker will immediately adopt the target period.
+func (t *VariableTicker) SetTargetPeriod(period time.Duration) error {
+	if period <= 0 {
+		return fmt.Errorf("invalid period %v, period must be positive", period)
+	}
+
+	frequency := float64(time.Second) / float64(period)
+
+	t.controlChan <- &frequencyUpdate{
```
litt3: Are you sure you want these control messages to block? What if the tick consumer dies or is slow? This could block indefinitely, if the run loop is stuck waiting for a tick consumer, and never handes the control messages

common/variable_ticker.go
```diff
@@ -0,0 +1,267 @@
...
+		elapsed := time.Since(startOfTick)
+		sleepTime := t.currentPeriod - elapsed
+		if sleepTime < 0 {
+			// If ticks are requested less often than the configured frequency, no need to sleep.
+			continue
+		}
+
+		timer.Reset(sleepTime)
+		select {
+		case <-timer.C:
```
litt3, line 200: claude complains about a race condition edge case, if the following order happens:

1. ctx is cancelled
2. timer fires
3. deferred timer.Stop() executes

The result being that the timer value sits in the channel forever

Not sure if you think this is worth patching

common/variable_ticker.go
```diff
@@ -0,0 +1,267 @@
...
+		if t.currentFrequency < t.targetFrequency {
+			// If we over shoot, adopt the target frequency.
+			t.currentFrequency = t.targetFrequency
+		}
+	}
+
+	if t.currentFrequency == 0 {
+		t.currentPeriod = 0
+	} else {
+		t.currentPeriod = time.Duration(float64(time.Second) / t.currentFrequency)
```
litt3, line 263: I think you need to check for overflow here. If frequency is very small, `currentPeriod` may overflow

common/variable_ticker.go
cody-littley: fixed

common/variable_ticker.go
cody-littley: fixed

common/variable_ticker.go
cody-littley: This is accurate, although the comment is sparse on details. 

Period 0 can only be set internally, not by the external caller. The user is not allowed to set a period of 0, since that would yield an infinite frequency (`freq = 1.0/0.0`). I overload this variable internally, and instead interpret period 0 to mean an infinitely long period. An infinitely long period means a frequency of `0`, which is a case this utility handles.

Updated comment as follows:

```
		if t.currentPeriod == 0 {
			// Period==0 is overloaded, and is used as a proxy for an infinitely long period (i.e. a frequency of 0). 
			// In that case, do not tick.
			// 
			// Only internal logic can set the period to 0. A user is unable to directly set the period to 0,
			// since if we interpret a period of 0 literally, it would require us to tick infinitely fast,
			continue
		}
```

common/variable_ticker.go
cody-littley: Ah, yes. You are correct. Fixed.

common/variable_ticker.go
cody-littley: check added

common/variable_ticker.go
cody-littley: Good point, added the following:

```
			time.Sleep(time.Millisecond) // to avoid busy waiting.
			continue
```

common/variable_ticker.go
cody-littley: Refactored so that we can handle control messages when waiting to send a tick.

common/variable_ticker.go
cody-littley: I'm not sure I understand this race condition...

Suppose we ignore the context for a moment and just think about the timer. If a timer fires but nobody reads the value from the channel and then `timer.Stop()` is called, then we will have a similar issue. If this is actually a real leak, then it seems like the timer API is flawed, since it's basically impossible to guarantee that a timer hasn't just fired when you go to stop it.

If it's simply an object left in a channel, this is not a real leak, since the channel will be GCd once nothing references the channel. It is not necessary to clean up channels, as long as we make sure we don't leave any goroutines stuck in a blocked state.

common/variable_ticker.go
cody-littley: I've set a minimum and maximum frequency that should protect us against overflow.

```
// Any frequency at or below this value will be interpreted as a frequency of 0 Hz. Needed to avoid overflow.
// The factor of 2 is to take care of floating point precision issues.
const MinimumFrequency = float64(time.Nanosecond/math.MaxInt64) * 2.0

// Any frequency above this value will be interpreted as a frequency of MaximumFrequency Hz. Needed to avoid overflow.
const MaximumFrequency = float64(time.Nanosecond)
```

common/variable_ticker.go
```diff
@@ -0,0 +1,315 @@
...
+
+// SetTargetPeriod sets the target period for the ticker. If acceleration is non-zero, the ticker will
+// move towards the target period at the rate of acceleration per second. If acceleration is zero,
+// the ticker will immediately adopt the target period.
+func (t *VariableTicker) SetTargetPeriod(period time.Duration) error {
+	if period <= 0 {
+		return fmt.Errorf("invalid period %v, period must be positive", period)
+	}
+
+	frequency := float64(time.Second) / float64(period)
```
litt3, line 106: Any reason not to just pass this `frequency` value into `SetTargetFrequency`, instead of duplicating the min and max checks?

test/v2/load/load_generator.go
```diff
@@ -143,9 +145,28 @@ func (l *LoadGenerator) Stop() {
 
 // run runs the load generator.
 func (l *LoadGenerator) run() {
-	ticker := time.NewTicker(l.submissionPeriod)
+
+	// Start with frequency 0.
+	ticker, err := common.NewVariableTickerWithFrequency(l.ctx, 0)
+	if err != nil {
+
```
litt3: empty block?

test/v2/load/load_generator.go
```diff
@@ -143,9 +145,28 @@ func (l *LoadGenerator) Stop() {
...
+	}
+
+	defer ticker.Close()
+	// Set acceleration prior to setting target frequency, since acceleration 0 allows "infinite" acceleration.
+	err := ticker.SetAcceleration(l.config.FrequencyAcceleration)
```
litt3: I'm confused how this is passing CI, my IDE has an error `no new variables on left side of :=`

common/variable_ticker.go
litt3: Makes sense, probably just a bad analysis from claude

common/variable_ticker.go
cody-littley: good point, simplified

test/v2/load/load_generator.go
cody-littley: fixed

test/v2/load/load_generator.go
cody-littley: Also confusing to me. Fixed though.

samlaf: We need to change https://github.com/Layr-Labs/eigenda/blob/510291b9be38cacbed8bc62125f6f9a14bd604e4/.github/workflows/golangci-lint.yml#L6
I assumed the default was that it runs for PRs targetting ALL branches, but it seems like it only runs on PRs targetting master (its passing CI because the lint workflow didnt even run)

litt3 (approved):

PR #1632, opened 2025-06-02
feat: use u8 for cert version

pakim249CAL:

Use a u8 for the certificate version, for off-chain convenience purposes.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


pakim249CAL:
> Changes LGTM. Any idea why all the bindings are being regenerated with a diff though?? I tried to fix it in #1635 by pinning forge/abigen versions using mise, but still ci is giving me a diff error... :(
> 
> When we look at the BIN diff in the generated abi, it doesn't even seem to have anything different. I'm very puzzled.

I don't think using bindings the way we are is common practice precisely because this is going to be a constant issue. I think we need to cut down on what is actually generated and kept in the bindings. Binary changes for example should not be reflected in a changeset to the bindings. Maybe we need to use abi jsons specifically rather than the whole contract artifacts.

see --extra-ouput [here](https://book.getfoundry.sh/forge/reference/forge-build/#compiler-options):

api/clients/v2/verification/cert_verifier.go
```diff
@@ -251,7 +251,7 @@ func (cv *CertVerifier) GetCertVersion(ctx context.Context, referenceBlockNumber
 	// if the version for the active cert verifier address has already been cached, return it immediately
 	cachedVersion, ok := cv.versions.Load(certVerifierAddress)
 	if ok {
-		castVersion, ok := cachedVersion.(uint64)
+		castVersion, ok := cachedVersion.(uint8)
 		if !ok {
 			return 0, fmt.Errorf("expected version to be uint64")
```
litt3, line 256: error message is outdated, should say `uint8`

litt3 (approved):

PR #1639, opened 2025-06-03
feat: eigen da ejection

pakim249CAL:

Adds an EigenDA Ejection Manager, which is a contract that replaces our current Ejection Manager with one that supports ejection delays and cancellations.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


contracts/src/periphery/ejection/EigenDAEjectionManager.sol
```diff
@@ -0,0 +1,187 @@
...
+        operator.completeChurn(quorums);
+        _tryEjectOperator(operator, quorums);
+        _returnDeposit(msg.sender);
+    }
+
+    /// OPERATOR FUNCTIONS
+
+    /// @notice Cancels the ejection process initiated by the operator. Transfers the deposit to the operator.
+    function cancelEjection() external {
+        msg.sender.cancelEjection();
```
Lulu-Zhou-EigenLabs, line 129: TODO: To allow the validator to sign with their BLS key and submit the transaction from any address, we need to implement the eligibility check for cancellation transactions as follows: verify that the input is a valid BLS signature over `OperatorProceedingParams`, using the operator’s BLS key. The mapping between validator addresses and their corresponding BLS public keys can be found [here](https://github.com/Layr-Labs/eigenlayer-middleware/blob/6797f3821db92c2214aaa6f137d94c603011ac2a/src/BLSApkRegistryStorage.sol).

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
```diff
@@ -0,0 +1,226 @@
...
+import {AccessControlConstants} from "src/core/libraries/v3/access-control/AccessControlConstants.sol";
+import {IAccessControl} from "@openzeppelin/contracts/access/IAccessControl.sol";
+
+contract EigenDAEjectionManager {
+    using AddressDirectoryLib for string;
+    using EigenDAEjectionLib for address;
+    using SafeERC20 for IERC20;
+
+    address internal immutable _depositToken;
+    uint256 internal immutable _depositAmount;
```
Lulu-Zhou-EigenLabs: The deposit is currently a fixed amount, but we originally planned to base it on the block’s `BASEFEE`. Will that logic still be implemented?
Also, are we planning to add per-ejector balance support, so that an ejector doesn't need to deposit funds every time it tries to eject a validator?

pakim249CAL: If we want ejector balance support, then there is probably no point in a depositAmount anymore, right?

pakim249CAL: [Updated](https://github.com/Layr-Labs/eigenda/pull/1639/commits/b0f6c5aeabe4e28fb7690687d37cdc92a7999283)

contracts/src/periphery/ejection/libraries/EigenDAEjectionLib.sol
```diff
@@ -0,0 +1,122 @@
...
+
+        deleteEjection(operator);
+        emit EjectionCompleted(operator, quorums);
+    }
+
+    /// @notice Helper function to clear an ejection.
+    function deleteEjection(address operator) internal {
+        EigenDAEjectionTypes.EjectionParams storage operatorParams = s().ejectionParams[operator];
+        operatorParams.ejector = address(0);
+        operatorParams.quorums = hex"";
```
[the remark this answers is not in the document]
pakim249CAL: This is the intended design. The cooldown is intended to limit how often an operator would need to respond to an ejection initiation from malicious ejectors. It is not meant to protect against having another ejection initiated after a cancellation.  I will clarify in the comments that this field is meant to be omitted from deletion, which is part of the reason why I chose to internalize the deletion function.

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
```diff
@@ -0,0 +1,260 @@
...
+    /// EJECTOR FUNCTIONS
+
+    /// @inheritdoc IEigenDAEjectionManager
+    function addEjectorBalance(uint256 amount) external onlyEjector(msg.sender) {
+        msg.sender.addEjectorBalance(amount);
+        IERC20(_depositToken).safeTransferFrom(msg.sender, address(this), amount);
+    }
+
+    /// @inheritdoc IEigenDAEjectionManager
+    function withdrawEjectorBalance(uint256 amount) external onlyEjector(msg.sender) {
```
Lulu-Zhou-EigenLabs, line 81: Should we check that the ejector has no on-going ejection here, or is this logic implemented at other places?

pakim249CAL: There is no need to check for an ejection, because the balance is already subtracted from when an ejector initiates an ejection. Any ejector balance they can withdraw is therefore safe to withdraw.

contracts/bindings/EigenDAEjectionManager/binding.go
samlaf, line 1: do we really need this binding, or does IEigenDAEjectionManager suffice?

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
```diff
@@ -0,0 +1,260 @@
...
+    using SafeERC20 for IERC20;
+
+    address internal immutable _depositToken;
+    uint256 internal immutable _depositAmount;
+    address internal immutable _addressDirectory;
+    uint256 internal immutable _estimatedGasUsedWithoutSig;
+    uint256 internal immutable _estimatedGasUsedWithSig;
+
+    bytes32 internal constant CANCEL_EJECTION_TYPEHASH = keccak256(
+        "CancelEjection(address operator, uint64 proceedingTime, uint64 lastProceedingInitiated, bytes quorums, address recipient)"
```
samlaf: shouldnt this only contain the types and not the variable names? also what about spaces?

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
samlaf: maybe rename function to `withdrawUnusedEjectorBalance`?

contracts/bindings/EigenDAEjectionManager/binding.go
pakim249CAL: Hmm, I think this would be remedied by this [pr](https://github.com/Layr-Labs/eigenda/pull/1906)? Because this file is not a compile target anymore.

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
pakim249CAL: You are correct to point out the spaces, but the names seem to be used in foundry's documentation and openzeppelin contracts at least.

https://getfoundry.sh/guides/eip712/

I'm not really sure if it matters in this case though, as we are not complying with EIP712. It should suffice to only have a unique typehash that can be shared, as this would be part of the message "signed" via BLS signature and verified in the contract. We can change this typehash to anything, and maybe it would be good to change it to avoid any confusion with an EIP712 typehash which is generally NOT used with bls signatures.

WDYT? any suggestions?

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
pakim249CAL: Does it not suffice to document that this is the case? Like [here](https://github.com/Layr-Labs/eigenda/blob/f213d70379ebf018863d338b82fda257440e2616/contracts/src/periphery/ejection/IEigenDAEjectionManager.sol#L10)?

samlaf: Curious why you prefer not to rename. Self-describing functions always better.
Or are you expecting the implementation to change and then the name of the function could be outdated?

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
samlaf: Missing a lot of context here, and not even super familiar with EIP712.
Isn't this typehash something we would eventually use offchain? And hence having access to eip712 utilities/libraries would be a good thing? Whereas a custom hash could cause confusion and bugs for offchain implementers?

If its only ever used onchain then I guess anything works indeed... but should probably document the reasoning.

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
pakim249CAL: Personally I think it is already self describing, as it should be implied that the contract will not let you withdraw any balance that you are not supposed to be able to withdraw. It would be like if a lending protocol named their borrow function "borrowOnlyIfCollateralized"

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
pakim249CAL: No, and I guess the case becomes stronger that we must use a separate scheme because this is NOT 712 related at all. It is a simple cryptographic principle that messages that are signed via any signature scheme be unique and target a specific domain. The typehash accomplishes that in EIP712, but we are not using EIP712, as the request was for a BLS signature to be used. Not the ECDSA curve, which is what EIP712 expects. This signature scheme does not satisfy any EIP, and should be considered unrelated.

contracts/src/periphery/ejection/libraries/EigenDAEjectionTypes.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
```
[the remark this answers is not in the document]
pakim249CAL: No, because this is a library. Libraries and interfaces should generally version as low as possible as long as there are no compatibility issues, so that other projects with different solidity versions can still use the files.

contracts/test/unit/EigenDAEjectionManager.t.sol
```diff
@@ -0,0 +1,113 @@
...
+    EigenDADirectory directory;
+    EigenDAAccessControl accessControl;
+    EigenDAEjectionManager ejectionManager;
+    ERC20Mintable token;
+
+    uint256 constant DEPOSIT_AMOUNT = 1e18;
+    uint256 constant CANCEL_EJECTION_WITHOUT_SIG_GAS_REFUND = 39128;
+    uint256 constant CANCEL_EJECTION_WITH_SIG_GAS_REFUND = 70000;
+    /// TODO: PLACEHOLDER UNTIL GAS COST FOR SIGNATURES IS KNOWN
+
```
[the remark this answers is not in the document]
pakim249CAL: It's okay. It's just a test.

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
```diff
@@ -0,0 +1,260 @@
...
+            weights[i] = IStakeRegistry(stakeRegistry).weightOfOperatorForQuorum(quorumNumber, operator);
+        }
+    }
+
+    function _refundGas(address receiver, uint256 estimatedGasUsed) internal virtual {
+        uint256 estimatedRefund = estimatedGasUsed * block.basefee;
+        IERC20(_depositToken).safeTransfer(
+            receiver, estimatedRefund > _depositAmount ? _depositAmount : estimatedRefund
+        );
+    }
```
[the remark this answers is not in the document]
pakim249CAL: Priority fee is intentionally not taken into account because the priority fee can be set by the caller.

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1639/commits/f1456b9c26870d567ee61442e526b771fab8529a

contracts/src/periphery/ejection/IEigenDAEjectionManager.sol
```diff
@@ -0,0 +1,73 @@
...
+    function getDepositAmount() external view returns (uint256);
+
+    /// @notice Returns the address of the ejector for a given operator.
+    function getEjector(address operator) external view returns (address);
+
+    /// @notice Returns whether an ejection process has been initiated for a given operator.
+    function ejectionTime(address operator) external view returns (uint64);
+
+    /// @notice Returns the timestamp of the last ejection proceeding initiated for a given operator.
+    function lastEjectionInitiated(address operator) external view returns (uint64);
```
cody-littley, line 60: Previously, there was a function `ejectionInitiated()` that returned a boolean value, depending on if there was currently an ejection in progress or not, but I don't see that function in this PR. Does `lastEjectionInitiated()` replace `ejectionInitiated()`? 

What is the behavior of this function if there was a previous ejection, but that ejection was canceled and is no longer active? Will it still return a timestamp?

contracts/src/periphery/ejection/IEigenDAEjectionManager.sol
```diff
@@ -0,0 +1,73 @@
...
+    /// @param recipient The address to which the gas refund will be sent.
+    function cancelEjectionWithSig(
+        address operator,
+        BN254.G2Point memory apkG2,
+        BN254.G1Point memory sigma,
+        address recipient
+    ) external;
+
+    /// @notice Cancels the ejection process for the message sender. Refunds gas to the caller.
+    function cancelEjection() external;
```
cody-littley, line 48: I'm confused why there are two methods for cancelling the ejection. When does an operator use one vs. the other? Does the offchain code that cancels ejections need to be capable of canceling with both strategies?

pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1639#discussion_r2128077516

contracts/src/periphery/ejection/IEigenDAEjectionManager.sol
pakim249CAL: The function was replaced with just getEjector(). This should suffice because there is a 1:1 correspondence between an ejection being initiated, and the ejector being a non-zero address. And so the previous boolean value can simply be replaced with getEjector(operator) != address(0)

cody-littley: Can you document that behavior in the `getEjector()` doc?

pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1639/commits/369158e6b0dd2648dc3c68dc1ad02993c8c7c1c6

samlaf (dismissed):
The issues I had raised have been addressed so giving my stamp. But I have not reviewed this PR in detail at all, so counting on others to do so.

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
Lulu-Zhou-EigenLabs: Should we set the `_depositAmount` as `MAX_GAS_SPIKE_MULTIPLIER * _estimatedGasUsedWithSig * base_fee`, where  `MAX_GAS_SPIKE_MULTIPLIER = 7`?

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
```diff
@@ -0,0 +1,268 @@
...
+    function _tryEjectOperator(address operator, bytes memory quorums) internal {
+        address registryCoordinator = IEigenDADirectory(_addressDirectory).getAddress(
+            AddressDirectoryConstants.REGISTRY_COORDINATOR_NAME.getKey()
+        );
+        try IRegistryCoordinator(registryCoordinator).ejectOperator(operator, quorums) {} catch {}
+    }
+
+    /// @notice Defines a unique identifier for a cancel ejection message to be signed by an operator for the purpose of authorizing a cancellation.
+    function _cancelEjectionMessageHash(address operator, address recipient) internal view returns (bytes32) {
+        return keccak256(
```
Lulu-Zhou-EigenLabs, line 229: Should also include ChainID to prevent replay attack.

Lulu-Zhou-EigenLabs (approved):
lgtm, left 2 comments.

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
```diff
@@ -0,0 +1,268 @@
...
+import {IAccessControl} from "@openzeppelin/contracts/access/IAccessControl.sol";
+import {IEigenDASemVer} from "src/core/interfaces/IEigenDASemVer.sol";
+
+contract EigenDAEjectionManager is IEigenDAEjectionManager, IEigenDASemVer {
+    using AddressDirectoryLib for string;
+    using EigenDAEjectionLib for address;
+    using SafeERC20 for IERC20;
+
+    address internal immutable _depositToken;
+    uint256 internal immutable _depositAmount;
```
anupsv: I'm wondering if this might come back to bite us. To change this amount, we'll need to deploy a new contract with the updated value and point the proxy to that yes ?

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
```diff
@@ -0,0 +1,268 @@
...
+    /// EJECTOR FUNCTIONS
+
+    /// @inheritdoc IEigenDAEjectionManager
+    function addEjectorBalance(uint256 amount) external onlyEjector(msg.sender) {
+        msg.sender.addEjectorBalance(amount);
+        IERC20(_depositToken).safeTransferFrom(msg.sender, address(this), amount);
+    }
+
+    /// @inheritdoc IEigenDAEjectionManager
+    function withdrawEjectorBalance(uint256 amount) external onlyEjector(msg.sender) {
```
anupsv, line 81: Should we give an option to take the amount or Would it be simpler to just return everything they have ?

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
```diff
@@ -0,0 +1,268 @@
...
+        address operator,
+        BN254.G2Point memory apkG2,
+        BN254.G1Point memory sigma,
+        address recipient
+    ) external {
+        address blsApkRegistry =
+            IEigenDADirectory(_addressDirectory).getAddress(AddressDirectoryConstants.BLS_APK_REGISTRY_NAME.getKey());
+
+        (BN254.G1Point memory apk,) = IBLSApkRegistry(blsApkRegistry).getRegisteredPubkey(operator);
+        _verifySig(_cancelEjectionMessageHash(operator, recipient), apk, apkG2, sigma);
```
anupsv, line 121: We need more things in the signature that the operator would need to sign. Essentially this should not be replayable. 
1. Does an ejection have a unique entry ? We should include something like that.
2. Add Chain id. Prepping for the future cause it will be "fork" style event incase we need to add in the future.

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
```diff
@@ -0,0 +1,268 @@
...
+    function cancelEjectionByEjector(address operator) external onlyEjector(msg.sender) {
+        operator.cancelEjection();
+        operator.getEjector().addEjectorBalance(_depositAmount);
+    }
+
+    /// @inheritdoc IEigenDAEjectionManager
+    function completeEjection(address operator, bytes memory quorums) external onlyEjector(msg.sender) {
+        operator.completeEjection(quorums);
+        _tryEjectOperator(operator, quorums);
+        operator.getEjector().addEjectorBalance(_depositAmount);
```
anupsv: Is this intended ? It tries to eject and if it fails it's silent and deducts ?

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
pakim249CAL: To prevent replay attacks on a chain fork? Yes, that does make sense. I'll add this.

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
pakim249CAL: Yes, we would need to deploy a new contract. But no, this contract is not meant to be behind a proxy. Because the procedure to update the EjectionManager will be to set the RegistryCoordinator's ejector to the new contract.

I believe it is unnecessary to add a level of indirection via proxy because we already have the means to update to a new contract.

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
pakim249CAL: It is simpler to return everything they have, but it is also very painful if we ever discover that someone doesn't like this behavior. I don't think this is much added complexity.

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
pakim249CAL: Yes. There is no such requirement that a failed ejection should keep the ejector balance locked. In the ejection manager's internal storage, the ejection is considered "complete" if the ejection was attempted on the registry coordinator.

In this case, we can think of this as a sort of gnosis safe transaction. We want to verify that a function was called with parameters that were set in advance, not that it succeeded.

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
anupsv: Ok, the RegistryCoordinator seems to be behaving close enough to what i was expecting.

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
anupsv: Yea the balance locked doesn't make sense. But i guess it's upto the caller to make sure they're calling an ejection on a valid operator.

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
pakim249CAL: Addressed #2

https://github.com/Layr-Labs/eigenda/pull/1639/commits/48bfcc3c19f2d34ee72d83c8703602ff060dc7a9

An ejection message is unique by the timestamp the ejection can complete. Is that not enough uniqueness?

contracts/src/periphery/ejection/EigenDAEjectionManager.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/commit/48bfcc3c19f2d34ee72d83c8703602ff060dc7a9

PR #1614, opened 2025-05-29
feat: add logger to accountant

hopeyen:
## Why are these changes needed?

Adding more helpful logs for users 

## Checks

- [ ] I've made sure the tests are passing. Note that there might be a few flaky tests, in that case, please comment that they are not relevant.
- [ ] I've checked the new test coverage and the coverage percentage didn't drop.
- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests
   - [ ] This PR is not tested :(

PR #1641, opened 2025-06-04
fix: period index calculation

hopeyen:

[[context on circular wrapping](https://www.notion.so/eigen-labs/Accountant-Circular-Wrapping-20113c11c3e08099893bf9a5daadda4e)]

the calculation for relative period index versus absolute period index was off by a factor of `reservationWindow`. 

At the time of accountant metering implementation, `reservationPeriodIndex` was `timestamp/reservationWindow`. We received an audit issue concerning that the records can become corrupted when `reservationWindow` changes (configurable by Payment Vault owner on-chain). Our solution was to update `reservationPeriodIndex` to the nearest lower multiple of `reservationWindow` (`timestamp/reservationWindow * reservationWindow`) so that we can keep `reservationWindow` configurable without storing `reservationWindow` along each reservation period records or storing `timestamp`.

This introduced a bug into the accountant, which kept using the original calculation `reservationPeriodIndex = timestamp/reservationWindow`, `relativeBinIndex = reservationPeriodIndex % numBins`, and `overflowBinIndex = (reservationPeriodIndex + 2) % numBins`. This means that if `reservationWindow % numBins = 0`, the bin index is consistently 0 for current period, and 2 for overflow period.  The unit tests were failing because all the cases had `reservationWindow % numBins != 0`

This PR updates relative bin index calculation: `relativeBinIndex = (reservationPeriodIndex / reservationWindow) % numBins`, and `overflowBinIndex = (reservationPeriodIndex/reservationWindow + 2) % numBins`; added test to cover this edge case

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/clients/v2/accountant.go
```diff
@@ -116,8 +116,8 @@ func (a *Accountant) BlobPaymentInfo(
 		return a.cumulativePayment, nil
 	}
 	return big.NewInt(0), fmt.Errorf(
-		"no bandwidth reservation found for account %s, and current cumulativePayment balance insufficient "+
-			"to make an on-demand dispersal. Consider depositing more eth to the PaymentVault contract.", a.accountID.Hex())
+		"invalid payments: no available bandwidth reservation found for account %s, and current cumulativePayment balance insufficient "+
```
samlaf, line 117: I really feel like this function should be broken into two: a function that only checks (and updates) reservation, and IF that fails (returns nil), then its caller's job to get a cumulativePayment). This function is very hard to review because its doing too many things, and `BlobPaymentInfo` doesn't reflect the fact that it updates state, and possibly returns a cumulativePayment. Can it update reservation state when returning a nonzero cumulativePayment?

These are all questions (and many more) that are running through my brain as I'm reviewing this. We could write more documentation on the function to describe all these possible scenarios, but I feel like it would be much simpler (and self-documenting) to split into 2 functions. Can prob be done in a separate PR at a future date, but still wanted to leave this comment.

api/clients/v2/accountant.go
```diff
@@ -116,8 +116,8 @@ func (a *Accountant) BlobPaymentInfo(
...
+			"to make an on-demand dispersal. Consider increasing reservation or cumulative payment on-chain.", a.accountID.Hex())
```
samlaf: on-chain is not descriptive enough. What if a customer is using proxy for v2 for the first time and doesn't know they need to deposit or make a reservation. Where should they go? Think we should give them a proper contract name, or even possibly give them our docs page to find the address for that contract.

api/clients/v2/accountant.go
```diff
@@ -157,9 +157,24 @@ func (a *Accountant) SymbolsCharged(numSymbols uint64) uint64 {
 	return core.RoundUpDivide(numSymbols, a.minNumSymbols) * a.minNumSymbols
 }
 
+// GetRelativePeriodRecord returns the period record for the given index
+// return empty record if there is no record for the relative index
```
samlaf:
```suggestion
// GetRelativePeriodRecord returns the period record for the given index.
// Returns an empty record if there is no record for the relative index.
```

api/clients/v2/accountant.go
```diff
@@ -157,9 +157,24 @@ func (a *Accountant) SymbolsCharged(numSymbols uint64) uint64 {
...
 func (a *Accountant) GetRelativePeriodRecord(index uint64) *PeriodRecord {
-	relativeIndex := uint32(index % uint64(a.numBins))
-	if a.periodRecords[relativeIndex].Index != uint32(index) {
+	// relativeIndex := uint32((index / a.reservationWindow) % uint64(a.numBins))
+	relativeIndex := uint32((index / a.reservationWindow) % uint64(a.numBins))
```
samlaf:
```suggestion
	relativeIndex := uint32((index / a.reservationWindow) % uint64(a.numBins))
```

api/clients/v2/accountant.go
```diff
@@ -157,9 +157,24 @@ func (a *Accountant) SymbolsCharged(numSymbols uint64) uint64 {
...
+	// Return empty record if the index is greater than the number of bins (should never happen by accountant initialization)
+	if relativeIndex >= uint32(a.numBins) {
+		return &PeriodRecord{Index: 0, Usage: 0}
```
samlaf: if this should never happen, shouldnt we panic to indicate that there's a bug in the code? makes it easier to catch such bugs early in tests if someone changes the constructor code without knowing about this condition for example

api/clients/v2/accountant.go
```diff
@@ -157,9 +157,24 @@ func (a *Accountant) SymbolsCharged(numSymbols uint64) uint64 {
...
+	}
+	return &a.periodRecords[relativeIndex]
+}
+
+// GetOrRefreshRelativePeriodRecord returns the period record for the given index (which is in seconds and is the multiple of the reservation window),
+// wrapping around the circular buffer and clearing the record if the index is greater than the number of bins
+func (a *Accountant) GetOrRefreshRelativePeriodRecord(index uint64, reservationWindow uint64) *PeriodRecord {
+	relativeIndex := uint32((index / reservationWindow) % uint64(a.numBins))
+	fmt.Println("periodRecords", a.periodRecords, "relativeIndex", relativeIndex, "index", index)
```
samlaf: accountant should take in a logger in its constructor and use that everywhere, not use `fmt.Println`. This should be a debug of trace log for example, such that it can be turned off by clients if too noisy

samlaf: also what's the point of these prints? they are very undescriptive. Was it for debugging and you just left them there by mistake? If not, they should probably have a bit more description.

api/clients/v2/accountant_test.go
```diff
@@ -385,6 +385,128 @@ func TestAccountBlob_ReservationOverflowReset(t *testing.T) {
...
+// Test cases:
+// Current bin under limit -> Simple increment
+// Current bin over limit but overflow bin empty -> can use overflow for spillage
+// Current bin over limit and overflow bin used -> reject, use on-demand
+// New window - start fresh with current bin, request cannot fit into a bin -> reject, use on-demand
+// New window - start fresh with current bin, within bin limit -> Simple increment
+// New window - current bin over limit but can use overflow -> overwrite spillage in previous bin
+// New window 2 - Exact bin limit usage -> Simple increment
+// New window 2 - current bin at limit -> Simply reject and use on-demand; no spillage
+func TestAccountBlob_ReservationOverflowWithWindow(t *testing.T) {
```
samlaf: Do these tests NEED to run sequentially? They should ideally all be separate tests that are run in parallel. Easier to understand, easier to debug, and faster to run.

api/clients/v2/accountant.go
```diff
@@ -157,9 +157,24 @@ func (a *Accountant) SymbolsCharged(numSymbols uint64) uint64 {
...
+	if a.periodRecords[relativeIndex].Index < uint32(index) {
```
samlaf, line 164: not related to this line but the PeriodRecord struct should be documented, along with its fields.

samlaf: image
also your comment says it contains 3 bins but it doesn't seem to be true anymore given that `numBins` is passed as an argument to `NewAccountant`

api/clients/v2/accountant_test.go
```diff
@@ -385,6 +385,128 @@ func TestAccountBlob_ReservationOverflowReset(t *testing.T) {
...
+	windowSize := time.Duration(reservationWindow) * time.Second
+
+	// Case 1: Current bin under limit -> Simple increment
+	now := baseTime
+	period := meterer.GetReservationPeriodByNanosecond(now, reservationWindow)
+	_, err = accountant.AccountBlob(ctx, now, 1000, quorums)
+	assert.NoError(t, err)
+	assert.Equal(t, uint64(1000), accountant.GetRelativePeriodRecord(period).Usage)
+	assert.Equal(t, uint64(0), accountant.GetRelativePeriodRecord(period+reservationWindow).Usage)
+	assert.Equal(t, uint64(0), accountant.GetRelativePeriodRecord(period+2*reservationWindow).Usage)
```
samlaf: I think these tests would be MUCH cleaner if they were written in table-driven test format: https://go.dev/wiki/TableDrivenTests

So that we could focus on all the different pre- and post- conditions. Here the meaningful data is intermingled with some sequential code, which makes the whole test very hard to follow (and probably just as hard to debug for anyone but you).

It would be much better for each test scenario to define a struct containing the entire pre-update and post-update table struct (with all the periodRecords). Then just make the update, and compare `expected` post-update struct to `gotten` post-update struct.

dmanc: +1

api/clients/v2/accountant.go
dmanc: Is it okay to have panic in client code? I guess this is violating some invariant so makes sense to panic but why not returning and logging an error instead?

api/clients/v2/accountant.go
dmanc: Seems like a good refactor, only called in one spot `func (a *Accountant) AccountBlob`.

api/clients/v2/accountant.go
dmanc: Is this comment about the edge case where `relativeIndex >= uint32(a.numBins)`?

api/clients/v2/accountant.go
```diff
@@ -157,9 +157,24 @@ func (a *Accountant) SymbolsCharged(numSymbols uint64) uint64 {
...
 
+// GetRelativePeriodRecord returns the period record for the given index
+// return empty record if there is no record for the relative index
 func (a *Accountant) GetRelativePeriodRecord(index uint64) *PeriodRecord {
-	relativeIndex := uint32(index % uint64(a.numBins))
-	if a.periodRecords[relativeIndex].Index != uint32(index) {
+	// relativeIndex := uint32((index / a.reservationWindow) % uint64(a.numBins))
+	relativeIndex := uint32((index / a.reservationWindow) % uint64(a.numBins))
+	// Return empty record if the index is greater than the number of bins (should never happen by accountant initialization)
+	if relativeIndex >= uint32(a.numBins) {
```
dmanc: Is it ever possible that a.numBins can change at runtime?

api/clients/v2/accountant.go
hopeyen: Agreed with the reasoning. This is something I have planned in this [PR](https://github.com/Layr-Labs/eigenda/pull/1619/files#diff-82a9ef86c9ab9755cf59bbc305d58939a1d1d7369ddb6b1cadecbd6b62d3e35a).

api/clients/v2/accountant.go
hopeyen: I linked the [doc](https://docs.eigenda.xyz/core-concepts/payments#disperser-client-requirements) and I think we can add more descriptions there for what exactly a customer needs to do.

api/clients/v2/accountant.go
hopeyen: Hmm this should really never happen. I think panic is okay because it will fail the unit tests

api/clients/v2/accountant.go
hopeyen: they were for me and I forgot:( removed the logs. 
I was going to add logger to accountant but I closed it because I agreed with your comments in this [PR](https://github.com/Layr-Labs/eigenda/pull/1614)

api/clients/v2/accountant_test.go
hopeyen: I spent a few hours on making them parallel just now, but because of the way we have to manage initial state of accountant and setting timestamp for each test, I ended up thinking having them sequential is easier to reason through and is stronger. Maybe it is harder to debug, but I'm hoping the comments would help people other than myself

api/clients/v2/accountant.go
hopeyen: Added in [commit](https://github.com/Layr-Labs/eigenda/pull/1641/commits/978621cdf6a95030496160d7f7839b963738d424)

api/clients/v2/accountant_test.go
hopeyen: Makes sense, updated in [commit](https://github.com/Layr-Labs/eigenda/pull/1641/commits/ffb047aa2614f5624f6f50a12dc2a0d6a11bff03)

api/clients/v2/accountant.go
hopeyen: Yes. this should never happen, but just in case, it will return an empty one

api/clients/v2/accountant.go
hopeyen: no we don't have mut on it, it will be fixed to the configured value

api/clients/v2/accountant.go
samlaf: Given that this is just a refactor, could you make it a separate in-between PR, to not bloat the PR with actual new features?
In integration team, we try (but tbf its hard) to follow the RII pattern: new feature introductions should be broken into 3 separate PRs:
1. refactor (any code refactors that are needed to prepare for the feature addition)
2. interface
3. implementation

You already broke your PR into interface and implementation, so that's awesome.

api/clients/v2/accountant.go
samlaf: @dmanc I'm becoming a huge huge fan of panics to state invariants. I have a few resources to share if you're interested, but main idea is that invariants are invariants, and need to be upheld. Having panics is the best way to "push left" bugs that need to be caught early (during tests). Also in many cases a client that has such bugs should not even be running (who knows if it'll start doing some crazy stuff and spending all your money).

I've also recently developed a pet-peeve for blindly returning errors. Golang's "errors as values" philosophy as upsides, but I think it also leads to laziness from the part of programmers. errors and exceptions are not the same thing (think java checked exceptions which NEED to be caught at compile time otherwise code doesnt compile). A LOT of our functions in core and proxy return errors when they really shouldn't (for example they are just doing a type conversion and blindly calling a function that in the general case can error, but in our specific case we as programmers KNOW that it can't error. That's a perfect case for panic'ing, instead of bloating all your functions with errors.

For example this function is clean right now because it doesn't return an error. That's super nice semantic to read, and it tells a LOT about the function (that its pure-ish most likely). We want this kind of information. Otherwise you just blindly turn your codebase into every function returning an error defensively, and all of them taking a ctx "in case". That's bad, and code pollution.

/rant-over

samlaf: "this message will self-destruct 5 seconds after being read", aka feel free to resolve once read, I'm happy with the change :)

api/clients/v2/accountant.go
samlaf: I still think there are good usecases for emitting debug logs. But fine as is.

api/clients/v2/accountant.go
samlaf: This is no longer true with the panic; did you remove the comment?

api/clients/v2/accountant.go
hopeyen: Yes, I can break it up. Given this is meant to be a fix for the current accounting, I would do refactor/interface/implementation for the upcoming quorum specific accounting

api/clients/v2/accountant.go
hopeyen: removed in [commit](https://github.com/Layr-Labs/eigenda/pull/1641/commits/20cd495f8c944b0c4ae073fe068191cbd465c7da)

api/clients/v2/accountant.go
samlaf: wdym by "have mut"? A new developer could certainly add some code that changes it (by mistake). Would that be a problem? If yes, then it should be documented in the accountant struct.

api/clients/v2/accountant.go
```diff
@@ -34,8 +34,11 @@ type Accountant struct {
 	numBins uint32
```
samlaf: Why store numBins? There is an implicit dependence between `periodRecords` and `numBins`, which is not good.
Think best to not store numBins and use use `len(periodRecords)` everywhere you currently use numBins? This way no chance of introducing bugs.

samlaf: For example you have a bug right now where your comment above `periodRecords` says "contains numBins of period records", but in your constructor you set `periodRecords := make([]PeriodRecord, numBins)` and then ONLY AFTER assign `numBins: max(numBins, uint32(meterer.MinNumBins)),`, which means its possible that `len(periodRecords) != numBins`

api/clients/v2/accountant.go
```diff
@@ -158,13 +161,12 @@ func (a *Accountant) SymbolsCharged(numSymbols uint64) uint64 {
 }
 
 // GetRelativePeriodRecord returns the period record for the given index
-// return empty record if there is no record for the relative index
+// Returns an empty record if there is no record for the relative index
 func (a *Accountant) GetRelativePeriodRecord(index uint64) *PeriodRecord {
```
samlaf: this function is only used in tests right now. Is it still needed? Also why is it public?

api/clients/v2/accountant.go
```diff
@@ -173,7 +175,6 @@ func (a *Accountant) GetRelativePeriodRecord(index uint64) *PeriodRecord {
 // wrapping around the circular buffer and clearing the record if the index is greater than the number of bins
 func (a *Accountant) GetOrRefreshRelativePeriodRecord(index uint64, reservationWindow uint64) *PeriodRecord {
```
samlaf: Why is this function public? It's only ever used by internal functions.

api/clients/v2/accountant.go
```diff
@@ -173,7 +175,6 @@ func (a *Accountant) GetRelativePeriodRecord(index uint64) *PeriodRecord {
...
 	relativeIndex := uint32((index / reservationWindow) % uint64(a.numBins))
-	fmt.Println("periodRecords", a.periodRecords, "relativeIndex", relativeIndex, "index", index)
 	if a.periodRecords[relativeIndex].Index < uint32(index) {
```
samlaf, line 164: In GetRelativePeriodRecord you panic when this case arises now. Should this here also panic, or is this case different?

api/clients/v2/accountant_test.go
samlaf: That's because you didn't write them in a proper table driven tests way.
In a table driven test, all test cases should be in a single golang test, and then you run a for loop over all test cases and reuse the same initialization code once (instead of having to boilerplate rewrite it 10 times like you did here). See https://github.com/Layr-Labs/eigenda-proxy/blob/2d022c1c986fbd0f64d2891673390918787c83bd/test/e2e/op_contract_test.go#L25 for an example

Typically, as a general rule, writing tests is the best way to test your API. If it's tedious to write your tests, it probably means your API is that tedious to write/use for external users.

api/clients/v2/accountant.go
samlaf: Speaking of which.. how should one choose numBins? There is no comment as to this effect, neither on the struct nor on the constructor.
1. can it be anything?
2. does it have to align with the disperser's meterer? (I see on our docs that "The disperser will track at least 4 periods per reservation, starting from the previous period to the period after next period. "). are these period per reservation the same thing?
3. what is a good choice for this number? why? does it even matter?

api/clients/v2/accountant.go
hopeyen: Didn't think it would hurt to have it defined in the struct but moved to be a helper within the test file.

api/clients/v2/accountant.go
hopeyen: added an indices check that would panic ([commit](https://github.com/Layr-Labs/eigenda/pull/1641/commits/90617e08304afbb402181998431ff55b394be8b4))

api/clients/v2/accountant.go
hopeyen: changed functions used internal to accountant to be private ([commit](https://github.com/Layr-Labs/eigenda/pull/1641/commits/90617e08304afbb402181998431ff55b394be8b4))

api/clients/v2/accountant.go
hopeyen: good catch! removed numBins and using `len(periodRecords)` everywhere [commit](https://github.com/Layr-Labs/eigenda/pull/1641/commits/90617e08304afbb402181998431ff55b394be8b4)

api/clients/v2/accountant.go
hopeyen: added some documentation. 
1. Anything. the accountant would use that configured number or 3, whichever is bigger
2. the disperser tracks the same thing
3. 3 is the minimum and the default to account for overflow usage. In reality client could even just have one bin and doesn't provide overflow functionality within the accountant. Using a bigger number doesn't provide functional difference to the accountant, but the dispersal client could track more reservation history

api/clients/v2/accountant_test.go
hopeyen: made a [ticket](https://linear.app/eigenlabs/issue/EGDA-1566/tabledriven-accountant-unit-tests) on linear and leaving this as a TODO

api/clients/v2/accountant.go
samlaf: 2. how do you make sure the client is using the same numBins as the disperser? EDIT: I see you overwrite PeriodRecords in SetPaymentState... could we do this in constructor instead? SetPaymentState is mandatory to be called to function properly right? This should prob be documented, and also AccountBlob should just panic if SetPaymentState has not been called I think?
3. whats the point of allowing >3 then if there's no functional difference? whats the point of tracking reservation history?

samlaf (approved):
Big improvements already, thanks for the changes. Mostly LGTM at this point, still have 2 questions to be addressed https://github.com/Layr-Labs/eigenda/pull/1641#discussion_r2136427496

api/clients/v2/accountant.go
dmanc: Raised a similar question here https://github.com/Layr-Labs/eigenda/pull/1599#discussion_r2133073738. Let's move the discussion to that PR?

hopeyen: 2. numBins doesn't have to use the same `numBins` as the disperser. `SetPaymentState` are only called by the first dispersal request if the accountant isn't filled in. The idea was to allow clients to implement accountants that manage state differently such as persisting them instead of relying on the disperser to provide the payment state, or call eth client directly for the onchain states. These are thing we won't implement, but allowing the possibility for modifications. Perhaps we should make it more of a requirement. Can keep this thought for later

3. I figured some clients might want to track more history for analyzing and adjusting their usage. The circular wrapping invariants stay the same with >3, and not a requirement to configure, so I figured it won't hurt. We can iterate on whether clients should have this flexibility

hopeyen:
> Big improvements already, thanks for the changes. Mostly LGTM at this point, still have 2 questions to be addressed [#1641 (comment)](https://github.com/Layr-Labs/eigenda/pull/1641#discussion_r2136427496)

thanks! I answered the questions, but can keep the discussions open for later changes. will go ahead and merge this one 🤞

PR #1642, opened 2025-06-05
chore: Update integrations spec reflect `EigenDACertVeriferRouter` conception and tightened glossary term usage

ethenotethan:
- update polling logic notes
- update cert verifier interface references
- add router notes

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


docs/spec/src/integration/spec/2-rollup-payload-lifecycle.md
```diff
@@ -57,13 +57,14 @@ As an example, an op-stack Ethereum rollup’s `payload` is a compressed batch o
...
-- `Batch`: Batch of blobs whose blob certs are grouped into a merkle tree and dispersed together for better network efficiency.
+- `Batch`: Batch of blobs whose blob certs are aggregated into a merkle tree and dispersed together for better network efficiency.
 - `DA Certificate` (or `DACert`): contains the information necessary to retrieve and verify a blob from the EigenDA network, along with a proof of availability.
 - `AltDACommitment`: RLP serialized `DACert` prepended with rollup-specific header bytes. This commitment is what gets sent to the rollup’s batcher inbox.
 
 [**Contracts**](./4-contracts.md)
 
-- `EigenDACertVerifier`: contains one main important function verifyCertV2 which is used to verify `DACert`s.
+- `EigenDACertVerifier`: contains one main important function checkDACert which is used to verify `DACert`s.
+- `EigenDACertVerifierRouter`: contains router mapping of activation block number to `EigenDACertVerifier`.
```
samlaf:
```suggestion
- `EigenDACertVerifierRouter`: contains router mapping of activation block number to `EigenDACertVerifier` and allows for securely and deterministically upgrading CertVerification constants (security thresholds and custom quorums) over time.
```

docs/spec/src/integration/spec/3-datastructs.md
```diff
@@ -69,13 +69,13 @@ A `DA Certicate` (or short `DACert`) contains all the information needed to retr
...
+A `DACert` contains the three data structs needed to call [checkDACert](https://github.com/Layr-Labs/eigenda/blob/3e670ff3dbd3a0a3f63b51e40544f528ac923b78/contracts/src/periphery/cert/EigenDACertVerifier.sol#L46-L56) on the EigenDACertVerifier.sol contract. Please refer to the eigenda core spec for more details, but in short, the `BlobCertificate` is included as a leaf inside the merkle tree identified by the `batch_root` in the `BatchHeader`. The `BlobInclusionInfo` contains the information needed to prove this merkle tree inclusion. The `NonSignerStakesAndSignature` contains the aggregated BLS signature `sigma` of the EigenDA validators. `sigma` is a signature over the `BatchHeader`.
```
samlaf: > contains the three data structs

It's 4 structs now that we added the signedQuorumNumbers in there: https://github.com/Layr-Labs/eigenda/blob/b26987a868cb542f9b5bf92afca54c7146be96ac/contracts/src/periphery/cert/EigenDACertTypes.sol#L15

samlaf: Also you're updating to point to certVerifier v3, but we still have people using certVerifier v2. Should we add a note about v2? Since for example bowen's hokulea also currently only supports the legacy certVerifier

docs/spec/src/integration/spec/3-datastructs.md
```diff
@@ -69,13 +69,13 @@ A `DA Certicate` (or short `DACert`) contains all the information needed to retr
...
 
 ![image.png](../../assets/integration/v2-batch-hashing-structure.png)
 
 ### AltDACommitment
 
-In order to be understood by each rollup stack’s derivation pipeline, the cert must be prepended with header bytes, to turn it into an [`altda-commitment`](https://github.com/Layr-Labs/eigenda-proxy?tab=readme-ov-file#commitment-schemas) respective to each stack:
+In order to be understood by each rollup stack’s derivation pipeline, the encoded `DACert` must be prepended with header bytes, to turn it into an [`altda-commitment`](https://github.com/Layr-Labs/eigenda-proxy?tab=readme-ov-file#commitment-schemas) respective to each stack:
```
samlaf, line 78: serialized `DACert`? Do we have a section related to this serialization (and maybe say that we plan on allowing flexible serialization later even though right now its always RLP encoding)?

docs/spec/src/integration/spec/4-contracts.md
```diff
@@ -43,13 +44,13 @@ defaultSecurityThresholdsV2 = {
 }
 ```
 
-A new BlobParam version is very infrequently introduced by the EigenDA Foundation Governance, and rollups can choose which version they wish to use when dispersing a blob. Currently there is only version 0 defined, with parameters:
+A new BlobParam version is very infrequently introduced by the EigenDA Foundation Governance, and rollups can choose which version they wish to use when dispersing a blob. Currently there is only version `0` defined, with parameters:
 
 ```solidity
 versionedBlobParams[0] = {
-    maxNumOperators = ??,
+    maxNumOperators =  3537,
```
samlaf, line 50: thanks for adding this! Guessing you got those from etherscan? Can we add that link here? So people can check in the future if there are new versions that we updated.

Also are those values the same on all networks? Do we enforce that? Or could we use different values for testnet?

docs/spec/src/integration/spec/4-contracts.md
```diff
@@ -63,7 +64,7 @@ where $\gamma = confirmationThreshold - adversaryThreshold$
 
 ### EigenDARelayRegistry
 
-Contains EigenDA network registered Relays’ Ethereum address and DNS hostname or IP address. `BlobCertificate`s contain `relayKey`(s), which can be transformed into that relay’s URL by calling [relayKeyToUrl](https://github.com/Layr-Labs/eigenda/blob/77d4442aa1b37bdc275173a6b27d917cc161474c/contracts/src/core/EigenDARelayRegistry.sol#L35).
+Contains EigenDA network registered Relays’ Ethereum address and DNS hostname or IP address. `BlobCertificates` contain `relayKeys`, which can be transformed into that relay’s URL by calling [relayKeyToUrl](https://github.com/Layr-Labs/eigenda/blob/77d4442aa1b37bdc275173a6b27d917cc161474c/contracts/src/core/EigenDARelayRegistry.sol#L35).
```
samlaf, line 66: I personally think the previous version was more accurate

docs/spec/src/integration/spec/4-contracts.md
```diff
@@ -75,4 +76,23 @@ Contains EigenDA network registered Dispersers’ Ethereum address. The EigenDA
 
 ### EigenDACertVerifier
 
-This contract's main use case is exposing a function verifyDACertV2 which is used to verify `DACert`s. This function’s logic is described in the [Cert Validation](./6-secure-integration.md#cert-validation) section.
+This contract's main use case is exposing a function checkDACert which is used to verify `DACerts`. This function’s logic is described in the [Cert Validation](./6-secure-integration.md#cert-validation) section. 
+
+The contract also exposes a `certVersion` method which is called by the payload disperser client to understand which certificate version is needed for backend encoding.
```
samlaf:
```suggestion
The contract also exposes a `certVersion` method which is called by the payload disperser client to know which cert version to build in order to be verifiable by that contract. 
```

samlaf: "backend encoding" was confusing to me

docs/spec/src/integration/spec/4-contracts.md
```diff
@@ -75,4 +76,23 @@ Contains EigenDA network registered Dispersers’ Ethereum address. The EigenDA
...
+    /// @notice A mapping from an activation block number (ABN) to a cert verifier address.
+    mapping(uint32 => address) public certVerifiers;
+
+    /// @notice The list of Activation Block Numbers (ABNs) for the cert verifiers.
+    /// @dev The list is guaranteed to be in ascending order
+    ///      and corresponds to the keys of the certVerifiers mapping.
+    uint32[] public certVerifierABNs;
+```
+
+where each key refers to an `activation_block_number` (ABN). When calling `checkDACert`, the reference block number is decoded from the `DACert` bytes and is used to find the closest sibling ABN via a reverse linear search over the `certVerifierABNs`. Once found, `EigenDACertVerifier` at the particular ABN is used for calling `checkDACert` to verify the DA Cert.
```
samlaf:
```suggestion
where each key refers to an `activation_block_number` (ABN). When calling `checkDACert`, the reference block number is decoded from the `DACert` bytes and is used to find the unique CertVerifier active at that RBN (a reverse linear search over the `certVerifierABNs` is performed). Once found, `EigenDACertVerifier` at the particular ABN is used for calling `checkDACert` to verify the DA Cert.
```

docs/spec/src/integration/spec/4-contracts.md
```diff
@@ -75,4 +76,23 @@ Contains EigenDA network registered Dispersers’ Ethereum address. The EigenDA
...
 ### EigenDACertVerifier
 
-This contract's main use case is exposing a function verifyDACertV2 which is used to verify `DACert`s. This function’s logic is described in the [Cert Validation](./6-secure-integration.md#cert-validation) section.
+This contract's main use case is exposing a function checkDACert which is used to verify `DACerts`. This function’s logic is described in the [Cert Validation](./6-secure-integration.md#cert-validation) section. 
+
+The contract also exposes a `certVersion` method which is called by the payload disperser client to understand which certificate version is needed for backend encoding. 
+
+### EigenDACertVerifierRouter
+
+This contract's main use case is allowing for secure upgrades of EigenDACertVerifier contracts in a cross version compatible format. This is done through maintaining a stateful mapping:
```
samlaf: we should prob mention that its not only for logic upgrades, but most importantly for upgrading custom quorums and security thresholds

docs/spec/src/integration/spec/4-contracts.md
samlaf: 'closest sibling' was unclear to me

docs/spec/src/integration/spec/5-lifecycle-phases.md
```diff
@@ -90,77 +90,95 @@ message BlobStatusReply {
...
+// Terminal states: UNKNOWN, COMPLETE, FAILED
 enum BlobStatus {
   UNKNOWN = 0; // functionally equivalent to FAILED but for unknown unknown bugs
   QUEUED = 1; // Initial state after a DisperseBlob call returns
   ENCODED = 2; // Reed-Solomon encoded into chunks ready to be dispersed to DA Nodes
-  CERTIFIED = 3; // blob has been dispersed and attested by NA nodes
-  FAILED = 4; // permanent failure (for reasons other than insufficient signatures)
-  INSUFFICIENT_SIGNATURES = 5;
+  GATHERING_SIGNATURES = 3; // blob chunks are actively being transmitted to validators
+  COMPLETE = 4; // blob has been dispersed and attested by NA nodes
```
samlaf:
```suggestion
  COMPLETE = 4; // blob has been dispersed and attested by DA nodes
```

docs/spec/src/integration/spec/5-lifecycle-phases.md
```diff
@@ -90,77 +90,95 @@ message BlobStatusReply {
...
+  FAILED = 5;
 }
 ```
 
-After a successful DisperseBlob rpc call, `BlobStatus.QUEUED` is returned. To retrieve a `cert`, the `GetBlobStatus` rpc shall be polled until a terminal status is reached. If `BlobStatus.CERTIFIED` is received, the `signed_batch` and `blob_verification_info` fields of the `BlobStatusReply` will be returned and can be used to create the `cert` . Any other terminal status indicates failure, and a new blob dispersal will need to be made.
+After a successful DisperseBlob RPC call, the disperser returns `BlobStatus.QUEUED`. To retrieve a cert, the GetBlobStatus RPC should be polled until a terminal status is reached.
 
-**Failover to EthDA**
+If `BlobStatus.GATHERING_SIGNATURES` is returned, the `signed_batch` and `blob_verification_info` fields will be present in the `BlobStatusReply`. These can be used to construct an `DACert`, which may be verified immediately against the configured threshold parameters stored in the `EigenDACertVerifier` contract. If the verification passes, the certificate can be accepted early. If verification fails, polling should continue.
```
samlaf:
```suggestion
If `BlobStatus.GATHERING_SIGNATURES` is returned, the `signed_batch` and `blob_verification_info` fields will be present in the `BlobStatusReply`. These can be used to construct a `DACert`, which may be verified immediately against the configured threshold parameters stored in the `EigenDACertVerifier` contract. If the verification passes, the certificate can be accepted early. If verification fails, polling should continue.
```

samlaf: beautiful

docs/spec/src/integration/spec/5-lifecycle-phases.md
```diff
@@ -90,77 +90,95 @@ message BlobStatusReply {
...
 
-The proxy can be configured to retry `FAILED` dispersal n times, after which it returns to the rollup a `503` HTTP status code which rollup batchers can use to failover to EthDA. See [here](https://github.com/ethereum-optimism/specs/issues/434) for more info.
+Once `BlobStatus.COMPLETE` is returned, the `signed_batch` and `blob_verification_info` fields are guaranteed to be included and should be used to construct the final `DACert`.
```
samlaf: I don't think we should say this. Think we should put more emphasis on the fact that COMPLETE just means that the disperser has stopped gathering more signatures. If we ever reach this state its most likely that there is an issue with eigenda, otherwise we would have exit early.

So prob good to mention that from a client's perspective, both COMPLETE and FAILED are likely failures. @litt3 will prob have an opinion on a proper way to phrase this as well given he wrote the backend implementation for this.

docs/spec/src/integration/spec/5-lifecycle-phases.md
```diff
@@ -90,77 +90,95 @@ message BlobStatusReply {
...
+
+ Any other terminal status indicates failure, and a new blob dispersal will need to be made.
```
samlaf:
```suggestion
Any other terminal status indicates failure, and a new blob dispersal will need to be made.
```

docs/spec/src/integration/spec/5-lifecycle-phases.md
```diff
@@ -90,77 +90,95 @@ message BlobStatusReply {
...
+
+**Failover to Native Rollup DA**
+
+The proxy can be configured to retry `FAILED` dispersal `n` times, after which it returns to the rollup a `503` HTTP status code which rollup batchers can use to failover to EthDA or native rollup DA offerings (e.g, arbitrum anytrust). See [here](https://github.com/ethereum-optimism/specs/issues/434) for more info on the OP implementation and [here](https://hackmd.io/@epociask/SJUyIZlZkx) for Arbitrum.
```
samlaf: prob want to retry not only on `FAILED` but also on `UNKNOWN` and even `COMPLETE` (see my  msg above) right? Not sure how to best translate that to words but think it would be more accurate

docs/spec/src/integration/spec/5-lifecycle-phases.md
```diff
@@ -90,77 +90,95 @@ message BlobStatusReply {
...
-```python
+In the updated implementation, a `CertBuilder` constructs the DA Cert through direct communication with the `OperatorStateRetriever` contract, which provides the necessary information about operator stake states. This approach ensures accurate on-chain data for certificate verification. The following pseudocode demonstrates this process:
 
+```python
 class CertV2:
-    batch_header: any  # You can specify proper types here
+    batch_header: any
     blob_verification_proof: any
     nonsigner_stake_sigs: any
+    cert_version: uint8  # Updated to use uint8 instead of uint256
```
samlaf: think we only have the uint8 cert_version on v3 right? 
Also this CertV2 is confusing right given that we have CertV3 already and this document actually describes CertV3. Should we just say Cert?

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -5,11 +5,13 @@ This page is meant to be read by eigenda and rollup developers who are writing a
...
+When deriving a rollup chain by running its derivation pipeline, only EigenDA `DACerts` that satisfy the three validity conditions are considered valid and used:
+1. RBN Recency Validation - ensure that the DA Cert's reference block number (RBN) is not too old with respect to the L1 block at which the cert was included in the rollup's batcher-inbox. This ensures that the blob on EigenDA has sufficient availability time left (out of the 14 day period) in order to be downloadable if needed during a rollup fault proof window.
 2. Cert Validation - ensures sufficient operator stake has signed to make the blob available, for all specified quorums. The stake is obtained onchain at a given reference block number (RBN) specified inside the cert.
 3. Blob Validation - ensures that the blob used is consistent with the KZG commitment inside the Cert.
 
+If a certificate fails validation then it is disregarded from the rollup's derivation pipeline and treated as an empty entry.
```
samlaf:
```suggestion
If a certificate fails validation then it MUST be discarded from the rollup's derivation pipeline.
```

samlaf: actually blob validation failing shouldnt lead to discarding of cert right...? cc @bxue-l2 
Because that just means the relay or validators are being malicious, but the cert itself is correct. it just points to another blob which we need to wait for (this should stall the derivation pipeline I think?)

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -36,14 +38,16 @@ This has a second security implication. A malicious EigenDA disperser could have
...
+Cert validation is done inside the EigenDACertVerifier contract, which EigenDA deploys as-is, but is also available for rollups to modify and deploy on their own. Specifically, [checkDACert](https://github.com/Layr-Labs/eigenda/blob/2414ed6f11bd28bc631eab4da3d6b576645801b0/contracts/src/periphery/cert/EigenDACertVerifier.sol#L46-L56) is the entry point for validation. This could either be called during a normal eth transaction (either for pessimistic “bridging” like EigenDA V1 used to do, or when uploading a Blob Field Element to a one-step-proof’s [preimage contract](https://specs.optimism.io/fault-proof/index.html#pre-image-oracle)), or be zk proven using a library like [Steel](https://github.com/risc0/risc0-ethereum/blob/main/crates/steel/docs/what-is-steel.md).
+
+`checkDACert` takes in an unstructured `[]byte` input. This is done to ensure seamless upgrades where the underlying representation of the `DACert` can change over time with new version introductions.
```
samlaf: prob need to mention that this is actually only to support the router interface not changing, otherwise this is confusing because CertVerifier in itself is immutable and can't change over time

samlaf: also please mention abi-encoded somewhere, not just "unstructured []byte"

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -76,6 +80,6 @@ Note: The verification steps in point 1. for dispersal are not currently impleme
 
 |                     | Nitro V1       | OP V1 (insecure) | Nitro V2       | OP V2                                                                                |
 | ------------------- | -------------- | ---------------- | -------------- | ------------------------------------------------------------------------------------ |
-| Cert Verification   | SequencerInbox | x                | one-step proof | one-step proof: done in preimage oracle contract when uploading a blob field element |
+| Cert Verification   | SequencerInbox | x                | one-step proof | one-step proof: done in one step proof HostIO contract |
```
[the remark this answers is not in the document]
samlaf: you changed OP column instead of arb column I think? OP doesn't have a HostIO contract

docs/spec/src/integration/spec/1-apis.md
```diff
@@ -43,18 +43,30 @@ relay.Relay.GetChunks
...
+    /// @return status An enum value. Success is always mapped to 1, and other values are errors specific to each CertVerifier.
+    function checkDACert(bytes calldata abiEncodedCert) external view returns (uint8 status);
+
+    /// @notice Returns the EigenDA certificate version. Used off-chain to identify how to encode a certificate for this CertVerifier.
+    /// @return The EigenDA certificate version.
+    function certVersion() external view returns (uint8);
+```
+
+#### Upgradable Router
+`EigenDACertVerifierRouter` acts as an intermediary contract that maintains an internal mapping of `activation_block_number -> EigenDACertVerifier`. This contract can be used to enable seamless upgrades for new `EigenDACertVerifier` and provides a way for a rollup to introduce custom quorums and thresholds.
```
samlaf:
```suggestion
`EigenDACertVerifierRouter` acts as an intermediary contract that maintains an internal mapping of `activation_block_number -> EigenDACertVerifier`. This contract can be used to enable seamless upgrades for new `EigenDACertVerifier` and provides a way for a rollup to securely introduce custom quorums and/or modify the security thresholds.
```

docs/spec/src/integration/spec/4-contracts.md
ethenotethan: oof good question - if they're gonna differ then we may wanna also document them in our networks [section](https://docs.eigenda.xyz/networks/mainnet) of eigenda docs

docs/spec/src/integration/spec/4-contracts.md
ethenotethan: true - was just trying fix the janky visual formatting (e.g, `entity`(s))

docs/spec/src/integration/spec/5-lifecycle-phases.md
ethenotethan: true but thats probably out-of-scope of this PR

docs/spec/src/integration/spec/1-apis.md
ethenotethan: [f9397e1](https://github.com/Layr-Labs/eigenda/pull/1642/commits/f9397e1f4fb1cc905e3c6d6ac75d302414600647)

docs/spec/src/integration/spec/6-secure-integration.md
ethenotethan: [f9397e1](https://github.com/Layr-Labs/eigenda/pull/1642/commits/f9397e1f4fb1cc905e3c6d6ac75d302414600647)

docs/spec/src/integration/spec/6-secure-integration.md
ethenotethan: [f9397e1](https://github.com/Layr-Labs/eigenda/pull/1642/commits/f9397e1f4fb1cc905e3c6d6ac75d302414600647)

docs/spec/src/integration/spec/5-lifecycle-phases.md
ethenotethan: [f9397e1](https://github.com/Layr-Labs/eigenda/pull/1642/commits/f9397e1f4fb1cc905e3c6d6ac75d302414600647)

docs/spec/src/integration/spec/5-lifecycle-phases.md
ethenotethan: [f9397e1](https://github.com/Layr-Labs/eigenda/pull/1642/commits/f9397e1f4fb1cc905e3c6d6ac75d302414600647)

docs/spec/src/integration/spec/5-lifecycle-phases.md
ethenotethan: [f9397e1](https://github.com/Layr-Labs/eigenda/pull/1642/commits/f9397e1f4fb1cc905e3c6d6ac75d302414600647)

docs/spec/src/integration/spec/5-lifecycle-phases.md
ethenotethan: [f9397e1](https://github.com/Layr-Labs/eigenda/pull/1642/commits/f9397e1f4fb1cc905e3c6d6ac75d302414600647)

docs/spec/src/integration/spec/4-contracts.md
ethenotethan: [f9397e1](https://github.com/Layr-Labs/eigenda/pull/1642/commits/f9397e1f4fb1cc905e3c6d6ac75d302414600647)

docs/spec/src/integration/spec/4-contracts.md
ethenotethan: [f9397e1](https://github.com/Layr-Labs/eigenda/pull/1642/commits/f9397e1f4fb1cc905e3c6d6ac75d302414600647)

docs/spec/src/integration/spec/4-contracts.md
ethenotethan: [f9397e1](https://github.com/Layr-Labs/eigenda/pull/1642/commits/f9397e1f4fb1cc905e3c6d6ac75d302414600647)

docs/spec/src/integration/spec/3-datastructs.md
ethenotethan: [f9397e1](https://github.com/Layr-Labs/eigenda/pull/1642/commits/f9397e1f4fb1cc905e3c6d6ac75d302414600647)

docs/spec/src/integration/spec/2-rollup-payload-lifecycle.md
ethenotethan: [f9397e1](https://github.com/Layr-Labs/eigenda/pull/1642/commits/f9397e1f4fb1cc905e3c6d6ac75d302414600647)

docs/spec/src/integration/spec/6-secure-integration.md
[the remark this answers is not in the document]
ethenotethan: [f9397e1](https://github.com/Layr-Labs/eigenda/pull/1642/commits/f9397e1f4fb1cc905e3c6d6ac75d302414600647)

docs/spec/src/integration/spec/3-datastructs.md
ethenotethan: hmm would we just disclose a warning informing users of its deprecation?

docs/spec/src/integration/spec/5-lifecycle-phases.md
ethenotethan: we should probably instead say `DACert` given its the glossary term used through the md book

ethenotethan: [3a4077a](https://github.com/Layr-Labs/eigenda/pull/1642/commits/3a4077afb0ba4489162877d085473b947df611e6)

docs/spec/src/integration/spec/3-datastructs.md
samlaf: I think that would be sufficient yeah

docs/spec/src/integration/spec/4-contracts.md
samlaf: is it janky? I was thinking the word contained inside the `brackets` should always be something defined in the glossary, and hence any plurality should be outside. Either works tbh...

docs/spec/src/integration/spec/6-secure-integration.md
samlaf: You haven't addressed my written concern above (that blob verification failing should NOT result in cert being discarded). Let's wait for @bxue-l2 to chime in and confirm this before making the change though.

docs/spec/src/integration/spec/5-lifecycle-phases.md
samlaf: Didn't mean to change the proxy, just mean to change the phrasing here in the spec (think spec should come first. if proxy doesn't implement spec then we should change that later).

docs/spec/src/integration/spec/4-contracts.md
ethenotethan: nvm they're consistent across envs

docs/spec/src/integration/spec/5-lifecycle-phases.md
litt3: > can be used to construct a `DACert`, the `DACert` could still be invalid if an insufficient amount of signatures were collected in-regards to the threshold parameters.

While technically true that you _could_ produce a cert with this, that's not how the `PayloadDisperser` behaves: it will only construct the cert if the required thresholds are met

> So prob good to mention that from a client's perspective, both COMPLETE and FAILED are likely failures.

@samlaf In practice it's correct that the `PayloadDisperser` seeing `COMPLETE` is *likely* indicative of a failure to gather a threshold of signatures. But I think trying to express this here might muddy the waters. Even if the `PayloadDisperser` moves on before `COMPLETE` is seen, the blob will still _eventually_ reach it, from the perspective of the disperser.

docs/spec/src/integration/spec/4-contracts.md
samlaf: Indeed! Realized we're lacking good documentation about this all across the board. Wrote https://github.com/Layr-Labs/eigenda/pull/1644 to improve the proto docs... the actual docs site should also be improved regarding this. but one step at a time :D

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: Blob validation is imcomplete. It contains two parts
- kzg and blob commitment check
- decoding payload from an eigenda blob
Especially in the following paragraph

For the first item, the cert shouldn't be discarded, because for every given valid kzg commitment there is an eigenda blob, the proxy(host) only fails to deliver that. The expected behavior should either be stalling with retry or panic.

For the second item, the cert along with the blob should be discarded, because it is the batcher's fault not using the correct encoding format. 

The second part touches more on the derivation logic, I think it is my fault to glob them together at the start. But since it is about the cert validity. Perhaps it is good to separate the blob decoding part out. And have a section for deriving a blob. @samlaf @ethenotethan Happy to take this task, if you think it is good

litt3 (approved):

docs/spec/src/integration/spec/5-lifecycle-phases.md
ethenotethan: makes sense: a1d8b4ea5c6b197e3084968779801dbb1c3a5efc

docs/spec/src/integration/spec/3-datastructs.md
ethenotethan: hmm perhaps we should include this when we have a formalized proxy release to point customers to?

docs/spec/src/integration/spec/6-secure-integration.md
ethenotethan: feel free to own @bxue-l2

docs/spec/src/integration/spec/3-datastructs.md
samlaf: Well given that this spec is already pointing to V3, I think we need to write *something* about the legacy CertVerifier that is in the core repo. Don't think we need to be blocked on proxy having integrated the router/clients with a proper release.

docs/spec/src/integration/spec/6-secure-integration.md
samlaf: @bxue-l2 created and assigned https://linear.app/eigenlabs/issue/DAINT-547/mdbook-breakdown-and-explain-blob-validation-error-cases to you

docs/spec/src/integration/spec/3-datastructs.md
samlaf: Created https://linear.app/eigenlabs/issue/DAINT-548/docs-deprecate-legacycertverifier-everywhere-and-point-to-v3-addresses

samlaf (approved):
Created linear tickets to track missing work (added links in comments) so LGTM to merge in this state.

PR #1644, opened 2025-06-05
docs: better proto BlobHeader.version documentation

samlaf:

Writing docs for custom quorums and the CertVerifier's security model, and realized we haven't documented properly the assumptions behind the BlobParams version.

Also slightly improved other fields of the BlobHeader.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/proto/common/v2/common_v2.proto
+++ b/api/proto/common/v2/common_v2.proto
@@ -7,22 +7,38 @@ option go_package = "github.com/Layr-Labs/eigenda/api/grpc/common/v2";
 
 // BlobHeader contains the information describing a blob and the way it is to be dispersed.
 message BlobHeader {
-  // The blob version. Blob versions are pushed onchain by EigenDA governance in an append only fashion and store the
-  // maximum number of operators, number of chunks, and coding rate for a blob. On blob verification, these values
-  // are checked against supplied or default security thresholds to validate the security assumptions of the
-  // blob's availability.
-  uint32 version = 1;
-  // quorum_numbers is the list of quorum numbers that the blob is part of.
-  // Each quorum will store the data, hence adding quorum numbers adds redundancy, making the blob more likely to be retrievable. Each quorum requires separate payment.
+  // The BlobParams version to use when encoding the blob into chunks to be dispersed to operators.
   //
-  // On-demand dispersal is currently limited to using a subset of the following quorums:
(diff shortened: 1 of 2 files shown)


api/proto/common/v2/common_v2.proto
```diff
@@ -7,22 +7,38 @@ option go_package = "github.com/Layr-Labs/eigenda/api/grpc/common/v2";
...
+  //   EigenDAServiceManager address (implements IEigenDAThresholdRegistry): https://docs.eigenda.xyz/networks/mainnet#contract-addresses
+  // Rollup maintained:
+  //   SecurityThresholds interface: https://github.com/Layr-Labs/eigenda/blob/3e670ff3dbd3a0a3f63b51e40544f528ac923b78/contracts/src/periphery/cert/interfaces/IEigenDACertVerifier.sol#L23
+  //   checkDACert interface: https://github.com/Layr-Labs/eigenda/blob/3e670ff3dbd3a0a3f63b51e40544f528ac923b78/contracts/src/periphery/cert/interfaces/IEigenDACertVerifierBase.sol#L8
   uint32 version = 1;
-  // quorum_numbers is the list of quorum numbers that the blob is part of.
-  // Each quorum will store the data, hence adding quorum numbers adds redundancy, making the blob more likely to be retrievable. Each quorum requires separate payment.
+  // quorum_numbers is the list of quorum numbers that the blob shall be dispersed to.
+  // Each quorum will store the data independently, meaning that additional quorum numbers increase redundancy, making the blob more likely to be retrievable.
```
litt3, line 30: > Each quorum will store the data independently

Is this still true, given the recent chunk assignment changes by @mooselumph? I didn't fully grasp those changes, but I recall something about not duplicating storage when a given node is in multiple quorums.

mooselumph: It is still true. The assignment is just done in a way that overlap in chunks stored across quorums is maximized. But notwithstanding this, each quorum will satisfy the security properties for availability.

samlaf: Ah. But you could in theory have single quorum's worth of replication if all quorums align. Which I guess is what we want thinking about it...
ok so this new update removes the useless replication (same operator storing same data 3 times) but still keeps the meaningful replication (quorums having different operators will cause an actual replication of data)

PR #1645, opened 2025-06-05
perf: combine api server and relay

yujiezhu0:

Combine API Server and Relay binaries

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Integration tests


Dockerfile
```diff
@@ -150,4 +160,8 @@ ENTRYPOINT ["generator"]
 
 FROM alpine:3.22 AS generator2
 COPY --from=generator2-builder /app/test/v2/bin/load /usr/local/bin
-ENTRYPOINT ["load", "-", "-"]
\ No newline at end of file
+ENTRYPOINT ["load", "-", "-"]
+
+FROM alpine:3.18 AS apiserver-and-relay
```
dmanc: lets switch it to 3.22

Dockerfile
```diff
@@ -103,6 +103,16 @@ RUN --mount=type=cache,target=/go/pkg/mod \
     --mount=type=cache,target=/root/.cache/go-build \
     make build
 
+# Combined API Server and Relay build stage
+FROM common-builder AS apiserver-and-relay-builder
```
dmanc: I don't like the name `apiserver-and-relay`, let's think about some alternatives

pschork (commented):
Can you include a build target in the [Makefile](https://github.com/Layr-Labs/eigenda/blob/master/disperser/Makefile#L9)

docker-bake.hcl
```diff
@@ -267,6 +270,22 @@ target "controller-internal" {
...
+  context    = "."
+  dockerfile = "./Dockerfile"
+  target     = "apiserver-and-relay"
+  tags       = ["${REGISTRY}/${REPO}/controller:${BUILD_TAG}"]
+}
+
+target "apiserver-and-relay-internal" {
+  inherits = ["apiserver-and-relay"]
+  tags     = [
+    "${REGISTRY}/eigenda-controller:${BUILD_TAG}",
```
dmanc: Should not be eigenda-controller

Dockerfile
yujiezhu0: What do you have in mind?
How about "gateway" or "frontend"?

disperser/Makefile
```diff
@@ -63,4 +66,32 @@ run_encoder: build_encoder
...
+run_blobapi: build_blobapi
+	./bin/blobapi \
+	--disperser-server.s3-bucket-name myDisperserBucket \
+	--disperser-server.dynamodb-table-name myDisperserTable \
+	--disperser-server.grpc-port 51002 \
+	--disperser-server.bls-operator-state-retriever 0x9d4454B023096f34B160D6B654540c56A1F81688 \
+	--disperser-server.eigenda-service-manager 0x67d269191c92Caf3cD7723F116c85e6E9bf55933 \
+	--disperser-server.enable-metrics=false \
+	--chain.rpc http://localhost:8545 \
+	--chain.private-key ac0974bec39a17e36ba4a6b4d238ff944bacb478cbed5efcae784d7bf4f2ff80 \
```
pschork: is this a throw away key?
i think you can also use blank key value like `--chain.private-key ''\`

disperser/Makefile
```diff
@@ -63,4 +66,32 @@ run_encoder: build_encoder
   --disperser-encoder.log.level-file debug
 
 run_dataapi_tests:
-	go test -v disperser/dataapi
\ No newline at end of file
+	go test -v disperser/dataapi
+
+run_blobapi: build_blobapi
```
pschork, line 72: I would recommend removing most of these args related to buckets, contracts and anything else that would be env specific and force the user to export those values via `.env` or other script that can be sourced.

PR #1647, opened 2025-06-06
feat: LittDB benchmark

cody-littley:

Create a benchmark utility for LittDB.


litt/benchmark/config/benchmark_config.go
```diff
@@ -0,0 +1,148 @@
...
+	// There is currently a rare bug somewhere, I suspect in metadata tracking. The bug can cause
+	// the benchmark to read a key that is no longer present in the database. Until that bug is fixed,
+	// do not halt the benchmark on read failures by default.
+	PanicOnReadFailure bool
+
+	// If true, fsync cohort files to ensure atomicity. Can be set to false for unit tests that need to be fast.
+	Fsync bool
+}
+
+// DefaultBenchmarkConfig returns a default BenchmarkConfig with the given data paths.
```
litt3: > with the given data paths.

What given data paths?

litt/benchmark/benchmark_engine.go
```diff
@@ -0,0 +1,285 @@
...
+	dataTracker, err := NewDataTracker(ctx, cfg, errorMonitor)
+	if err != nil {
+		cancel()
+		return nil, fmt.Errorf("failed to create data tracker: %w", err)
+	}
+
+	writeBytesPerSecond := uint64(cfg.MaximumWriteThroughputMB * float64(units.MiB))
+	writeBytesPerSecondPerThread := writeBytesPerSecond / uint64(cfg.WriterParallelism)
+
+	// If we set the write burst size smaller than an individual value, then the rate limiter will never
```
litt3, line 102: This comment seems to indicate that you should be comparing the `writeBurstSize` to the size of an individual value (same comment goes for `readBurstSize` below)

litt/benchmark/benchmark_engine.go
```diff
@@ -0,0 +1,285 @@
...
+// Logger returns the logger used by the benchmark engine.
+func (b *BenchmarkEngine) Logger() logging.Logger {
+	return b.logger
+}
+
+// Run executes the benchmark. This method blocks forever, or until the benchmark is stopped via control-C or
+// encounters an error.
+func (b *BenchmarkEngine) Run() error {
+
+	// multiply 2 to make configured value the average
```
litt3: multiply *by* 2

litt/benchmark/benchmark_engine.go
```diff
@@ -0,0 +1,285 @@
...
+			}
+
+			b.metrics.reportRead(time.Since(start), uint64(len(readInfo.Value)))
+
+			if !exists {
+				if b.config.PanicOnReadFailure {
+					b.errorMonitor.Panic(fmt.Errorf("key %s not found in database", readInfo.Key))
+					return
+				} else {
+					b.logger.Warnf("key %s not found in database", readInfo.Key)
```
litt3: should this be an error, instead of warning?

litt/benchmark/cohort.go
```diff
@@ -0,0 +1,362 @@
...
+    | new | --> | exhausted | --> | complete |
+    +-----+     +-----------+     +----------+
+       |              |
+       v              |
+    +-----------+     |
+    | abandoned | <---|
+    +-----------+
+
+- new: the cohort was just created and is currently being used to supply keys for writing.
+- exhausted: all keys in the cohort have taken to be written, but the DB may not have ingested them all yet.
```
litt3: > have taken to be written

odd phrasing

litt/benchmark/cohort.go
litt3, line 1: It might be good to add a paragraph explaining why the concept of a cohort is necessary in the first place.

litt/benchmark/cohort.go
```diff
@@ -0,0 +1,362 @@
...
+
+	err := cohort.Write()
+	if err != nil {
+		return nil, fmt.Errorf("failed to write cohort file: %w", err)
+	}
+
+	return cohort, nil
+}
+
+func LoadCohort(path string) (*Cohort, error) {
```
litt3, line 123: exported function missing doc

litt/benchmark/cohort.go
```diff
@@ -0,0 +1,362 @@
...
+
+	parentDirectory := filepath.Dir(path)
+	indexString := strings.Replace(filepath.Base(path), CohortFileExtension, "", 1)
```
litt3, line 128: Could you write a brief doc explaining what this string parsing is doing?

litt/benchmark/cohort.go
```diff
@@ -0,0 +1,362 @@
...
+
+// GetKeyIndexForWriting gets the next key to be written to the database.
+func (c *Cohort) GetKeyIndexForWriting() (uint64, error) {
+	if c.loadedFromDisk {
+		return 0, fmt.Errorf("cannot allocate key for writing: cohort has been loaded from disk")
+	}
+	if c.allValuesWritten {
+		return 0, fmt.Errorf("cannot allocate key for writing: cohort is already complete")
+	}
+	if c.nextKeyIndex > c.highKeyIndex {
```
litt3: could use `IsExhausted`

litt/benchmark/cohort.go
```diff
@@ -0,0 +1,362 @@
...
+	// sanity check
+	if choice < c.lowKeyIndex || choice > c.highKeyIndex {
+		return 0, fmt.Errorf("invalid choice: %d not in range [%d, %d]", choice, c.lowKeyIndex, c.highKeyIndex)
+	}
+
+	return choice, nil
+}
+
+// MarkComplete marks that all key-value pairs in the cohort have been written to the database. Once done,
+// all key-value pairs in the cohort become safe to read, so long as the cohort has not yet expired. A cohort
```
litt3, line 270: >  A cohort is said to have expired when it is possible that at least one key in the cohort may be deleted from the DB due to the TTL.

Would it make sense to add `expired` to your state machine above?

litt/benchmark/cohort.go
```diff
@@ -0,0 +1,362 @@
...
+	}
+	return nil
+}
+
+// Path returns the file path of the cohort file.
+func (c *Cohort) Path() string {
+	return path.Join(c.parentDirectory, fmt.Sprintf("%d%s", c.cohortIndex, CohortFileExtension))
+}
+
+func (c *Cohort) Write() error {
```
litt3, line 299: Also missing a doc

litt/benchmark/cohort.go
```diff
@@ -0,0 +1,362 @@
...
+	c.firstValueTimestamp = time.Unix(int64(binary.BigEndian.Uint64(data[32:40])), 0)
+	c.allValuesWritten = data[40] == 1
+
+	return nil
+}
+
+// IsExpired returns true if the cohort has expired (i.e. it is no longer safe to read).
+func (c *Cohort) IsExpired(now time.Time, maxAge time.Duration) bool {
+	if !c.IsComplete() {
+		if c.loadedFromDisk && !c.IsComplete() {
```
litt3: redundant check of `c.IsComplete()`

litt/benchmark/cohort_test.go
```diff
@@ -0,0 +1,312 @@
...
+	require.NoError(t, err)
+	require.True(t, exists)
+
+	// Initialize a copy cohort from the file
+	loadedCohort, err := LoadCohort(cohort.Path())
+	require.NoError(t, err)
+	require.Equal(t, cohortIndex, loadedCohort.CohortIndex())
+	require.Equal(t, lowIndex, loadedCohort.LowKeyIndex())
+	require.Equal(t, highIndex, loadedCohort.HighKeyIndex())
+	require.Equal(t, valueSize, cohort.ValueSize())
```
litt3, line 46: Should this be `require.Equal(t, valueSize, loadedCohort.ValueSize())`?

litt/benchmark/cohort_test.go
```diff
@@ -0,0 +1,312 @@
...
+	require.Equal(t, false, cohort.IsComplete())
+
+	// Extract all keys from the cohort.
+	for i := lowIndex; i <= (lowIndex+highIndex)/2; i++ {
+		key, err := cohort.GetKeyIndexForWriting()
+		require.NoError(t, err)
+		require.Equal(t, i, key)
+
+		shouldBeExhausted := i == highIndex
```
litt3: Don't think that `shouldBeExhausted` will ever be `true`, since you stop looping at `(lowIndex+highIndex)/2`

litt/benchmark/data_generator.go
```diff
@@ -0,0 +1,74 @@
...
+	rng.Read(dataPool)
+	randPool.Put(rng)
+
+	return &DataGenerator{
+		randPool: randPool,
+		dataPool: dataPool,
+	}
+}
+
+// Key generates a new key. The value id deterministic for the same index and seed.
```
litt3: value ~~id~~ *is* deterministic

litt/benchmark/data_tracker.go
```diff
@@ -0,0 +1,538 @@
...
+	"context"
+	"fmt"
+	"math"
+	"math/rand"
+	"os"
+	"path"
+	"strings"
+	"time"
+
+	config2 "github.com/Layr-Labs/eigenda/litt/benchmark/config"
```
litt3: Why are you importing this as `config2`?

litt/benchmark/data_tracker.go
```diff
@@ -0,0 +1,538 @@
...
+	writeInfoChan := make(chan *WriteInfo, config.WriteInfoChanelSize)
+	readInfoChan := make(chan *ReadInfo, config.ReadInfoChanelSize)
+	writtenKeyIndicesChan := make(chan uint64, 64)
+
+	ttl := time.Duration(config.TTLHours * float64(time.Hour))
+	safetyMargin := time.Duration(config.ReadSafetyMarginMinutes * float64(time.Minute))
+	safeTTL := ttl - safetyMargin
+
+	closedChan := make(chan struct{}, 1)
+	closedChan <- struct{}{} // Initially closed, will be drained when the DataTracker is closed.
```
litt3: "Initially closed"? Not sure what that means

litt/benchmark/data_tracker.go
```diff
@@ -0,0 +1,538 @@
...
+					0,
+					nil,
+					fmt.Errorf("failed to load cohort: %w", err)
+			}
+			cohorts[cohort.CohortIndex()] = cohort
+
+			if cohort.CohortIndex() < lowestCohortIndex {
+				lowestCohortIndex = cohort.CohortIndex()
+			}
+			if cohort.cohortIndex > highestCohortIndex {
```
litt3, line 249: Inconsistent access of `cohortIndex`: `cohort.CohortIndex()` vs `cohort.cohortIndex`

litt/benchmark/data_tracker.go
```diff
@@ -0,0 +1,538 @@
...
+					0,
+					nil,
+					fmt.Errorf("failed to delete swap file: %w", err)
+			}
+		}
+	}
+
+	if len(cohorts) == 0 {
+		// Special case, no cohorts found.
+		return 0, 0, cohorts, nil
```
litt3, line 266: The indices returned from this function are assumed to be valid later, right? So returning `0` start and end index seems like it would be a problem in this special case

litt/benchmark/data_tracker.go
```diff
@@ -0,0 +1,538 @@
...
+		} else {
+			// Stop once we find the first cohort that is not eligible for deletion.
+			break
+		}
+	}
+
+	if len(t.cohorts) == 0 {
+		// Edge case: we've been writing data slow enough that the active cohort has expired.
+		// Create a new active cohort.
+		activeCohort, err := t.activeCohort.NextCohort(t.config.CohortSize, t.valueSize)
```
litt3, line 530: I suppose the assumption here is that the active cohort will have already been deleted?

litt/disktable/segment/key_file.go
```diff
@@ -162,9 +162,9 @@ func (k *keyFile) atomicSwap() error {
 	k.swap = false
 	newPath := k.path()
 
-	err := os.Rename(swapPath, newPath)
+	err := util.AtomicRename(swapPath, newPath, sync)
 	if err != nil {
-		return fmt.Errorf("failed to rename key file: %v", err)
+		return fmt.Errorf("failed to atomically swap key file %s with %s: %v", swapPath, newPath, err)
```
litt3: It was pre-existing, but `%v`should be `%w`.

litt/disktable/disk_table.go
```diff
@@ -35,11 +35,11 @@ type DiskTable struct {
 	// The logger for the disk table.
 	logger logging.Logger
 
-	// fatalErrorHandler is a struct that permits the DB to "panic". There are many goroutines that function under the
+	// errorMonitor is a struct that permits the DB to "panic". There are many goroutines that function under the
 	// hood, and many of these threads could, in theory, encounter errors which are unrecoverable. In such situations,
 	// the desirable outcome is for the DB to report the error and then refuse to do additional work. If the DB is in a
 	// broken state, it is much better to refuse to do work than to continue to do work and potentially corrupt data.
-	fatalErrorHandler *util.FatalErrorHandler
+	errorMonitor *util.ErrorMonitor
```
litt3, line 42: In the future, probably would be good to do large renames like `fatalErrorHandler` -> `errorMonitor` in a separate PR, to avoid cluttering the diff of the feature being developed

litt/disktable/table_metadata.go
```diff
@@ -32,18 +32,24 @@ type tableMetadata struct {
 
 	// the table's sharding factor, accessed/modified by concurrent goroutines
 	shardingFactor atomic.Uint32
+
+	// If true, metadata writes will be atomic. Should be set to true in production, but can be set to false
```
litt3, line 36: > Should be set to true in production

With this in mind, would it make sense to log errors in all the critical components when Fsync is set to `false`, to hopefully help catch any misconfigurations?

litt/util/fatal_error_handler.go
litt3, line 1: Struct was renamed, but file still has old name.

litt/util/fatal_error_handler.go
```diff
@@ -9,102 +9,101 @@ import (
...
+// When there are goroutines that function under the hood that are unable to return errors using the standard pattern,
+// this utility provides an elegant way to handle those errors. In such situations, the desirable outcome is for the
+// process to report the error and to elegantly spin itself down.
 //
-// Even though this utility can "panic", it is not the same as the panic that is built into Go. Once the DB "panics",
-// all public methods will return a meaningful error, and the DB will refuse to do additional work.
-type FatalErrorHandler struct {
+// Even though this utility can "panic", it is not the same as the panic that is built into Go. The Panic() method
+// should be called in situations where recovery is not possible, i.e. the same situations where one would otherwise use
+// as a golang panic. The big difference is that calling Panic() will not result in the process immediately being torn
```
litt3: > where one would otherwise use as a golang panic

confusing phrasing

common/logger_config.go
```diff
@@ -54,8 +54,8 @@ func LoggerCLIFlags(envPrefix string, flagPrefix string) []cli.Flag {
 
 // DefaultLoggerConfig returns a LoggerConfig with the default settings for a JSON logger.
 // In general, this should be the baseline config for most services running in production.
-func DefaultLoggerConfig() LoggerConfig {
-	return LoggerConfig{
+func DefaultLoggerConfig() *LoggerConfig {
```
litt3, line 57: This change might also have been better split out into a separate PR, to help decrease the size of this very large PR, and make reviewing easier.

litt/util/files.go
```diff
@@ -9,6 +9,107 @@ import (
 	"time"
 )
 
+// SwapFileExtension is the file extension used for temporary swap files created during atomic writes.
+const SwapFileExtension = ".swap"
+
+// SanitizePath returns a sanitized version of the given path, doing things like expanding
+// "~" to the user's home directory, converting to absolute path, normalizing slashes, etc.
+func SanitizePath(path string) (string, error) {
```
litt3, line 17: Utility function could probably benefit from some test coverage

litt/util/files.go
```diff
@@ -9,6 +9,107 @@ import (
...
+}
+
+// AtomicWrite writes data to a file atomically. The parent directory must exist and be writable.
+// If the destination file already exists, it will be overwritten.
+//
+// This method creates a temporary swap file in the same directory as the destination, but with SwapFileExtension
+// appended to the filename. If there is a crash during this method's execution, it may leave this swap file behind.
+func AtomicWrite(destination string, data []byte, fsync bool) error {
+
+	swapPath := destination + ".swap"
```
litt3: Hardcoded use of `".swap"`, rather than defined `SwapFileExtension` constant

litt/util/units.go
```diff
@@ -0,0 +1,88 @@
...
+	if quantity < steps[1].multiple {
+		// Edge case, print without a decimal point if we have the smallest unit.
+		return fmt.Sprintf("%d %s", quantity, steps[0].name)
+	}
+
+	unit := steps[0].name
+	floatQuantity := float64(quantity)
+
+	for i := 1; i < len(steps); i++ {
+		if floatQuantity > float64(steps[i].multiple) {
```
litt3: should this be `>=`?

litt/benchmark/config/benchmark_config.go
cody-littley: Removed the part about file paths.

litt/benchmark/benchmark_engine.go
cody-littley: I'm not sure what you mean by "comparing". The rate limiter is analogous to a bucket full of water. In order to do an operation, you remove an amount of water equal to the size of the operation. If there isn't enough water in the bucket, you are not allowed to do the operation. The burst size determines the size of the bucket, and the rate determines how fast new water is added to the bucket. 

If there is an operation that is larger than the bucket, then the operation can never be executed. So it's important to make the bucket at least as large as the largest possible operation. For writes, that is equal to the size of an individual value. For reads, that is equal to the largest possible value that can be written, or that has been written so far.

litt/benchmark/benchmark_engine.go
cody-littley: fixed

litt/benchmark/benchmark_engine.go
cody-littley: fixed

litt/benchmark/cohort.go
cody-littley: Updated to `all keys in the cohort have been scheduled for writing`

litt/benchmark/cohort.go
cody-littley:
```
// A Cohort is a grouping of key-value pairs used for benchmarking.
//
// If a benchmark wants to read values, it must somehow figure out which keys have been written to the database.
// If it wants to verify the validity of the data it reads, it must also be able to determine the correct value
// that should be associated with any particular key, and it must also be able to determine when keys are
// expected to be removed from the database due to TTL expiration.
//
// Tracking the sort of metadata required to do reads in a benchmark is not a trivial thing, especially when
// the scale of the benchmark is large (i.e. tens or hundreds of millions of keys over weeks or months of time). 
// Storing this information in memory is simply not plausible, and storing it on disk requires database scale similar 
// to what LittDB is handling, unless we are clever about it. A "cohort" is that clever mechanism. Each cohort tracks a
// large collection of key-value pairs in the database, and it does it in a way that uses very little disk space.
//
// Key-value pairs each have unique indices, and knowing the index of a key-value pair allows the data to be
// regenerated deterministically. All key-value pairs in a cohort have sequential indices. A single cohort can
// track multiple gigabytes worth of key-value pairs, but on disk it only requires a few dozen bytes of data.
```

litt/benchmark/cohort.go
cody-littley: added

litt/benchmark/cohort.go
cody-littley:
```
	// Cohort file names are in the format "X.cohort", where X is the cohort index.
	// Replacing ".cohort" with an empty string gives us the cohort index in string form.
	indexString := strings.Replace(filepath.Base(path), CohortFileExtension, "", 1)
```

litt/benchmark/cohort.go
cody-littley: updated

litt/benchmark/cohort.go
cody-littley:
```
/* The lifecycle of a cohort:

    +-----+     +-----------+     +----------+     +---------+
    | new | --> | exhausted | --> | complete | --> | expired |
    +-----+     +-----------+     +----------+     +---------+
       |              |
       v              |
    +-----------+     |
    | abandoned | <---|
    +-----------+

- new: the cohort was just created and is currently being used to supply keys for writing.
- exhausted: all keys in the cohort have been scheduled for writing, but the DB may not have ingested them all yet.
- complete: all keys in the cohort have been written to the DB and are safe to read.
- abandoned: before becoming complete, the benchmark was restarted. It will never be thread safe to read or write
              any keys in this cohort.
- expired: the cohort has been marked as complete, but it can no longer be read because the TTL has expired
            (or is about to expire).
*/
```

litt/benchmark/cohort.go
cody-littley:
```
// Write the data in this cohort to its file on disk. When this method returns, the cohort file is guaranteed to be
// crash durable.
```

litt/benchmark/cohort.go
cody-littley: removed

litt/benchmark/cohort_test.go
cody-littley: Yes, fixed

litt/benchmark/cohort_test.go
cody-littley: You are correct. I've hard coded this check as `require.Equal(t, false, cohort.IsExhausted())`. The previous pattern was copy-pasted from a unit test that used a cohort that actually could become exhausted.

litt/benchmark/data_generator.go
cody-littley: fixed

litt/benchmark/data_tracker.go
cody-littley: frickin goland. Still haven't tried out cursor yet, but it's on my TODO list.

litt/benchmark/data_tracker.go
cody-littley: Fixed

```
closedChan <- struct{}{} // Will be drained when the DataTracker is closed.
```

litt/benchmark/data_tracker.go
cody-littley: Fixed.

Another thing that I like about rust over golang: private struct variables are truly private. I hate that it's possible to accidentally use unexported struct variables like this.

litt/benchmark/data_tracker.go
cody-littley: Added the following to the docs:

```
// gatherCohorts loads cohorts from files on disk. The lowest/highest cohort indices are valid if and only if the
// cohorts map is not empty. If no cohorts are found, the lowest and highest cohort indices will be 0.
func gatherCohorts(cohortDirPath string) (
```

I have the exact same issue with segment files in LittDB, since the high index is inclusive not exclusive. Every time I call `GatherSegments()` I first have to check if the segment map is empty or not. I should probably switch the pattern so that `0,0` signals empty data, but the bang for the buck argument is tricky.

litt/benchmark/data_tracker.go
cody-littley: Correct. There is an invariant the code maintains that there is always an active non-exhausted cohort. If `IsExpired()` returns true for the active cohort, the loop above deletes it. This last statement restores the invariant by creating a new cohort.

litt/disktable/segment/key_file.go
cody-littley: fixed

litt/disktable/disk_table.go
cody-littley: will do

litt/disktable/table_metadata.go
cody-littley: Would you be ok with a log message in `db_impl.go`'s `NewDB()` method?

```
	if config.Fsync == false {
		config.Logger.Warnf(
			"Fsync is disabled. Ok for unit tests that need to run fast, NOT OK FOR PRODUCTION USE.")
	}
```

Or would you prefer a log in each method? It might get a bit spammy in unit tests.

litt/util/fatal_error_handler.go
cody-littley: fixed

litt/util/fatal_error_handler.go
cody-littley:
```
// Even though this utility can "panic", it is not the same as the panic that is built into Go. The Panic() method
// should be called in situations where recovery is not possible, i.e. the same situations where one would otherwise 
// call golang's panic(). The big difference is that calling Panic() will not result in the process immediately being 
// torn down.
```

litt/util/files.go
cody-littley: added

litt/util/files.go
cody-littley: fixed

litt/util/units.go
cody-littley: Makes sense, updated

litt/benchmark/benchmark_engine.go
litt3: Gotcha, I was interpreting the comment incorrectly. 

So you're setting the burst size to exactly the minimum value that it can be?

PR #1648, opened 2025-06-06
fix: Verify v2 cert instead of reflexively passing

litt3:
Since v2 cert and v3 cert are identical, except for field ordering, we can easily (though verbosely) convert from v2->v3. This allows us to provide correct verification results for v2 certs, rather than reflexively treating them as valid.

Closes DAINT-550 ([link](https://linear.app/eigenlabs/issue/DAINT-550/golang-client-fix-that-v2-certs-are-verified-using-v3certverifier))

### DIFF
--- a/api/clients/v2/coretypes/eigenda_cert.go
+++ b/api/clients/v2/coretypes/eigenda_cert.go
@@ -388,64 +388,60 @@ func (c *EigenDACertV2) Version() CertificateVersion {
 	return VersionTwoCert
 }
 
-// ConvertV2CertToV3 converts an EigenDACertV2 to an EigenDACertV3
-func ConvertV2CertToV3(certV2 *EigenDACertV2) (*EigenDACertV3, error) {
-	if certV2 == nil {
-		return nil, fmt.Errorf("input V2 certificate cannot be nil")
-	}
-
+// ToV3 converts an EigenDACertV2 to an EigenDACertV3
+func (c *EigenDACertV2) ToV3() (*EigenDACertV3, error) {
 	// Convert BlobInclusionInfo from V2 to V3 format
 	v3BlobInclusionInfo := certTypesBinding.EigenDATypesV2BlobInclusionInfo{
 		BlobCertificate: certTypesBinding.EigenDATypesV2BlobCertificate{
 			BlobHeader: certTypesBinding.EigenDATypesV2BlobHeaderV2{
-				Version:       certV2.BlobInclusionInfo.BlobCertificate.BlobHeader.Version,
-				QuorumNumbers: certV2.BlobInclusionInfo.BlobCertificate.BlobHeader.QuorumNumbers,
+				Version:       c.BlobInclusionInfo.BlobCertificate.BlobHeader.Version,
+				QuorumNumbers: c.BlobInclusionInfo.BlobCertificate.BlobHeader.QuorumNumbers,
 				Commitment: certTypesBinding.EigenDATypesV2BlobCommitment{
 					Commitment: certTypesBinding.BN254G1Point{
-						X: certV2.BlobInclusionInfo.BlobCertificate.BlobHeader.Commitment.Commitment.X,
-						Y: certV2.BlobInclusionInfo.BlobCertificate.BlobHeader.Commitment.Commitment.Y,
+						X: c.BlobInclusionInfo.BlobCertificate.BlobHeader.Commitment.Commitment.X,
+						Y: c.BlobInclusionInfo.BlobCertificate.BlobHeader.Commitment.Commitment.Y,
 					},
 					LengthCommitment: certTypesBinding.BN254G2Point{
-						X: certV2.BlobInclusionInfo.BlobCertificate.BlobHeader.Commitment.LengthCommitment.X,
-						Y: certV2.BlobInclusionInfo.BlobCertificate.BlobHeader.Commitment.LengthCommitment.Y,
+						X: c.BlobInclusionInfo.BlobCertificate.BlobHeader.Commitment.LengthCommitment.X,
+						Y: c.BlobInclusionInfo.BlobCertificate.BlobHeader.Commitment.LengthCommitment.Y,
 					},
 					LengthProof: certTypesBinding.BN254G2Point{
-						X: certV2.BlobInclusionInfo.BlobCertificate.BlobHeader.Commitment.LengthProof.X,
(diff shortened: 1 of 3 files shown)


ethenotethan (approved):
LGTM

api/clients/v2/coretypes/eigenda_cert.go
```diff
@@ -387,3 +387,95 @@ func (c *EigenDACertV2) ComputeBlobKey() (*coreV2.BlobKey, error) {
...
 }
+
+// ConvertV2CertToV3 converts an EigenDACertV2 to an EigenDACertV3
+func ConvertV2CertToV3(certV2 *EigenDACertV2) (*EigenDACertV3, error) {
+	if certV2 == nil {
+		return nil, fmt.Errorf("input V2 certificate cannot be nil")
+	}
+
+	// Convert BlobInclusionInfo from V2 to V3 format
+	v3BlobInclusionInfo := certTypesBinding.EigenDATypesV2BlobInclusionInfo{
```
bxue-l2, line 394: why there are two binding?
```
contractEigenDACertVerifierV2 "github.com/Layr-Labs/eigenda/contracts/bindings/EigenDACertVerifierV2"

certTypesBinding "github.com/Layr-Labs/eigenda/contracts/bindings/IEigenDACertTypeBindings"
```
If they are the same, then I suppose we can do assignment.

bxue-l2 (approved):

api/clients/v2/coretypes/eigenda_cert.go
samlaf: First is the actual contract, the other just contains the router cert types (>= 3) afaik

api/clients/v2/coretypes/eigenda_cert.go
```diff
@@ -387,3 +387,95 @@ func (c *EigenDACertV2) ComputeBlobKey() (*coreV2.BlobKey, error) {
...
+	v3NonSignerStakesAndSignature := certTypesBinding.EigenDATypesV1NonSignerStakesAndSignature{
+		NonSignerQuorumBitmapIndices: certV2.NonSignerStakesAndSignature.NonSignerQuorumBitmapIndices,
+		NonSignerPubkeys:             convertV2PubkeysToV3(certV2.NonSignerStakesAndSignature.NonSignerPubkeys),
+		QuorumApks:                   convertV2PubkeysToV3(certV2.NonSignerStakesAndSignature.QuorumApks),
+		ApkG2: certTypesBinding.BN254G2Point{
+			X: certV2.NonSignerStakesAndSignature.ApkG2.X,
+			Y: certV2.NonSignerStakesAndSignature.ApkG2.Y,
+		},
+		Sigma: certTypesBinding.BN254G1Point{
+			X: certV2.NonSignerStakesAndSignature.Sigma.X,
```
samlaf: I keep getting scared that we'll forget a field when doing these verbose manual conversions. Doesn't seem like go has any other good way to do this though (outside of json conversion which has runtime costs and unsafe pointer assignment whcih is.. unsafe).

Will our current set of linters catch if we forget a field? I don't think so right since we turned that off to allow for default values in a bunch of places... otherwise Claude proposed this one
image

api/clients/v2/verification/cert_verifier.go
```diff
@@ -55,39 +55,44 @@ func (cv *CertVerifier) CheckDACert(
...
-		}
 	case coretypes.VersionTwoCert:
-		_, ok := cert.(*coretypes.EigenDACertV2)
+		certV2, ok := cert.(*coretypes.EigenDACertV2)
 		if !ok {
 			return &CertVerifierInputError{Msg: fmt.Sprintf("expected cert to be of type EigenDACertV2, got %T", cert)}
 		}
-		return nil // noop verification. only used on V2 testnets and no mainnets
+
+		certV3, err = coretypes.ConvertV2CertToV3(certV2)
```
samlaf: curious why you seem to prefer functions over methods. Isn't `certV2.ToV3()` cleaner?

api/clients/v2/coretypes/eigenda_cert.go
```diff
@@ -387,3 +387,95 @@ func (c *EigenDACertV2) ComputeBlobKey() (*coreV2.BlobKey, error) {
...
+	return certV3, nil
+}
+
+// convertUint32SliceToRelayKeys converts []uint32 to []coreV2.RelayKey for V3 format
+func convertUint32SliceToRelayKeys(relayKeys []uint32) []coreV2.RelayKey {
+	result := make([]coreV2.RelayKey, len(relayKeys))
+	for i, key := range relayKeys {
+		result[i] = coreV2.RelayKey(key)
+	}
+	return result
```
samlaf, line 464: sad that golang forces us to make an allocation for these conversions.. https://go.dev/doc/faq#convert_slice_with_same_underlying_type

It is what it is I guess

mise.toml
```diff
@@ -3,7 +3,7 @@
 go = "1.21"
 
 # Tooling Dependencies
-"go:github.com/golangci/golangci-lint/cmd/golangci-lint" = "1.60.3"
+golangci-lint = "1.60.3"
```
samlaf, line 6: good catch for this! Looking at https://mise.jdx.dev/registry.html?filter=golangci#tools, it seems like this will use one of the binary installer backends (aqua, ubi, asdf) and download the binary directly. Whereas the go tool will download the source code and then build the binary.

Honestly I don't understand why you had issues with the go build... but this way should be faster anyways.

samlaf (approved):
LGTM. Added some philosophical comments but don't need to change anything

api/clients/v2/coretypes/eigenda_cert.go
litt3: @bxue-l2 Basic assignment doesn't work, I think the only way is manual here.

api/clients/v2/coretypes/eigenda_cert.go
litt3: I'm not really sure this would provide any additional safety.

If we were constructing this v3 cert in multiple locations, and we wanted to just create a single implementation that are very confident in that could be reused, that would make sense.

But we are only doing this in one place. If we were to create constructors to do this work, then those functions have equivalent bug potential to the existing code, right?

api/clients/v2/verification/cert_verifier.go
litt3: 154c482ff7f2185e8f628cd6c6011ef678ccf7d2

ethenotethan (approved):

PR #1649, opened 2025-06-06
feat: Create initial version of release management process

litt3:
Closes EGDA-1571


docs/release-process.md
```diff
@@ -0,0 +1,109 @@
...
+   - [Creating the Release](#creating-the-release)
+   - [Release Notes (TODO)](#release-notes-todo)
+6. [Post-Release Updates](#6-post-release-updates)
+
+---
+
+### 1. **Feature Freeze & Release Branch Creation**
+
+#### Feature Freeze
+
```
cody-littley: Suggested paragraph:

> Enacting a feature freeze helps to ensure that the code we publish to production environments is well tested and mature. As a general rule of thumb, we should put in place a feature freeze X weeks prior to a release on testnet (with exceptions for emergencies).

cody-littley: We should have a discussion about the proper value for X. My vote would be `X=2`.

docs/release-process.md
```diff
@@ -0,0 +1,109 @@
...
+- **Trigger**: Either a date scheduled in advance, or tied to the completion of a key feature
+- **Action**: Announce feature freeze date and time to the team
```
cody-littley:
```suggestion
- **Action**: Announce feature freeze date and time to the team and [create a release branch](create-release-branch).
```

docs/release-process.md
```diff
@@ -0,0 +1,109 @@
...
+---
+
+### 2. **Changes to a Release Branch**
+
+#### Change Policy
+
+- **High bar for inclusion**: Only critical bugfixes or business-critical features
+    - Even bugfixes should not be reflexively included: only high-severity issues
+- **Team consensus required**: Single engineer cannot make the decision
+- **Public visibility**: Must have team discussion (e.g., Slack thread) before proceeding
```
cody-littley:
```suggestion
- **Public visibility**: Must have team discussion (e.g., Slack thread) before proceeding, or somebody in management must sign off
```

docs/release-process.md
```diff
@@ -0,0 +1,109 @@
...
+
+- **Trigger**: Either a date scheduled in advance, or tied to the completion of a key feature
+- **Action**: Announce feature freeze date and time to the team
+- **Result**: Development against `master` continues uninterrupted while release is prepared
+
+#### Create Release Branch
+
+- From latest `master` commit:
+  - `git checkout master && git pull`
+  - `git checkout -b release-<MAJOR>.<MINOR>`
```
cody-littley: Our current branch format is `release/<MAJOR>.<MINOR>`. Any particular reason for changing it?

litt3: No reason, will update

docs/release-process.md
```diff
@@ -0,0 +1,109 @@
+# GitHub Release Management Process
```
ethenotethan: should we add a section specific to contract releases as well?

docs/release-process.md
```diff
@@ -0,0 +1,109 @@
...
+#### Create Release Branch
+
+- From latest `master` commit:
+  - `git checkout master && git pull`
+  - `git checkout -b release-<MAJOR>.<MINOR>`
+    - **Note:** there is no patch number in the branch name. The same branch is used across multiple patch versions.
+  - Example: `release-1.1`
+- Push and protect the branch:
+  - `git push origin release-1.1`
+  - From the GitHub UI, set branch protections to prevent direct pushes and branch deletion
```
dmanc: Do we need extra permissions to do this?

docs/release-process.md
samlaf: think this should be either `release-<MAJOR>` or `release-0.<MINOR>` (while we are pre 1.0.0)?

docs/release-process.md
```diff
@@ -0,0 +1,109 @@
...
+- **Action**: Announce feature freeze date and time to the team
+- **Result**: Development against `master` continues uninterrupted while release is prepared
+
+#### Create Release Branch
+
+- From latest `master` commit:
+  - `git checkout master && git pull`
+  - `git checkout -b release-<MAJOR>.<MINOR>`
+    - **Note:** there is no patch number in the branch name. The same branch is used across multiple patch versions.
+  - Example: `release-1.1`
```
samlaf: if we're on 1.1, why not just use release-1?

docs/release-process.md
samlaf: wdym? someone with extra permissions (e.g. anup) can set this up for us no?

samlaf: or are you asking if there's a need for elevated permissions to bypass those rules?

docs/release-process.md
samlaf: Really? Isn't it better to keep reviewers to people who understand the code?

docs/release-process.md
```diff
@@ -0,0 +1,109 @@
...
+  2. Cherry-pick the squashed commit into the release branch
+- **If change is release-only:**
+  - Submit PR directly against the release branch
+- **⚠️ NEVER push directly to the release branch**
+
+---
+
+### 3. **Tagging a Release**
+
+- **When ready**, tag from HEAD of release branch:
```
samlaf: I feel like the condition for readiness should be described in this document. What is a release? what is the purpose for creating one? how are they tested? 

Otherwise why code freeze at all? The point of code freeze is to test the code right? So is a precondition to releasing a tag that it's been released and running on preprod for 1 week? what about testnet? Do we make separate releases for testnet and mainnet? How does that work? etc etc etc

samlaf: Also should we recommend (prescribe?) the use of semver checking tools to figure out the correct semver to release?
rust has https://crates.io/crates/cargo-semver-checks which I've heard good things about.
Golang seems to have https://pkg.go.dev/golang.org/x/exp/apidiff

docs/release-process.md
```diff
@@ -0,0 +1,109 @@
...
+  - NEVER force-push a tag to a different commit
+  - If a mistake is made, create a new tag with incremented version
+
+---
+
+### 4. **Updates After Release Tag Has Been Cut**
+
+- **Additional fixes** after initial tag may be required
+- **Follow same change policy** as described in [Section 2](#2-changes-to-a-release-branch)
+- **Do not tag reflexively** after every merge:
```
samlaf: maybe this is a good use of `-rc.x` suffixes?

docs/release-process.md
```diff
@@ -0,0 +1,109 @@
...
+  - Accumulate changes until a meaningful patch set is ready
+  - Create new release tag with incremented patch version (e.g., `v1.1.1`, `v1.1.2`)
```
samlaf: I feel like this is wrong. shouldnt we test our stuff properly for 2 weeks with `-rc` suffixes and then when we release we know its a proper release?

samlaf: I guess it depends if we use `semver` releases purely robotically (we don't care about WHAT we release and whether it has a useful semantic to users, as long as it follows the correct semver versioning rules), or if we try to make our releases holistic and fix/feature/breaking related. Idk a release that says "oops please don't use `va.b.n-1` because that contains a bug that we just fixed feels wrong. Although maybe its fine because thats literally the purpose of patch updates lol

samlaf: Oh reading the next section I see why this feels weird to me. In this scheme, in public github repo releases, you'd have gaps between patch versions? Like you could see v1.1.1 and then v1.1.4 with gaps between them if those were all fixing bugs? That seems like what `rc`s are for?

docs/release-process.md
```diff
@@ -0,0 +1,109 @@
...
+- **Continue iteratively** until all critical issues are resolved
+
+---
+
+### 5. **Official Release & Notes**
+
+#### Creating the Release
+
+- **When ready to make the release public:**
```
samlaf: I think we need to be careful with this, because for go tools, a tag is already public and is the real thing. Github releases is just human fluff.

docs/release-process.md
```diff
@@ -0,0 +1,109 @@
...
+  - If necessary, tag final patch version from release branch HEAD
+  - Create GitHub release via UI, targeting the most recent tag
+  - **Note**: Release will likely have non-zero patch version
+
+#### Release Notes (TODO)
+
+- Define a consistent format and content structure
```
samlaf: https://keepachangelog.com/en/1.1.0/ is nice I feel

docs/release-process.md
dmanc: Yeah if it's a one-time setup that all release/* branches can be protected then we can do it now.

docs/release-process.md
```diff
@@ -0,0 +1,109 @@
...
+- **If change is release-only:**
+  - Submit PR directly against the release branch
+- **⚠️ NEVER push directly to the release branch**
+
+---
+
+### 3. **Tagging a Release**
+
+- **When ready**, tag from HEAD of release branch:
+  - Tag format: `v<MAJOR>.<MINOR>.<PATCH>`
```
bxue-l2: what about rc release, should we talk about it here

docs/release-process.md
litt3: cc @anupsv Do you know how this works?

If everyone _does_ have this permission be default, they probably shouldn't, right?

docs/release-process.md
litt3: I suppose it depends on the behavior of `apidiff`.

Skimming the docs, it seems like it might say something is a "breaking change" if *any* exported struct or method changes. Since we have a TON of exported things, I would say every release we create is almost guaranteed to change something that is exported.

docs/release-process.md
litt3: According to @dmanc, our current release pipeline doesn't allow multiple tags to point to the same commit

This limitation presents a problem for `rc` tags: what do you do if you want to cut a public release, and there aren't any required changes, so there is already an `rc` tag at the desired commit?

I don't understand where this limitation comes from, but maybe the right answer is to fix that problem, and then continue using `rc` releases to address the various concerns you've raised @samlaf

docs/release-process.md
litt3: Let's consolidate `rc` discussion to [this thread](https://github.com/Layr-Labs/eigenda/pull/1649#discussion_r2138604281)

docs/release-process.md
litt3: I'd argue that, even with `rc` suffixes, you wouldn't want to necessarily tag every single commit.

docs/release-process.md
litt3: Adding this statement about management signoff wouldn't impact the standard PR review process: the PR to merge into the release branch would still need normal reviews.

This is only regarding the decision of whether to move forward with including a given feature/fix in a release. I think it would be reasonable for management to decide that something belongs in a given release

docs/release-process.md
anupsv: Are we restricting who can edit or delete a release tag. Ideally no one should be able to as part of the normal process but only in rare cases.

docs/release-process.md
pakim249CAL: Is it not a thing to title a release based on the component being released? e.g. contracts/1.0.0

docs/release-process.md
litt3: > Ideally no one should be able to as part of the normal process but only in rare cases

💯 Agree. IMO your average developer should not have permission to edit or delete (e.g. I wouldn't *want* to have this permission)

docs/release-process.md
litt3: It's still an open question, how we are going to plan to split up releases. Currently working on an additional section for that

> Is it not a thing to title a release based on the component being released

This makes sense to me

docs/release-process.md
litt3: a299cd1022b1e33f9d1851dc51945fb2c152402c

litt3: I reworked this section a bit, including the proposed rule of thumb

docs/release-process.md
litt3: I reworked this section

docs/release-process.md
litt3: 7e512f0eee410c101eb1e30e5f16b996e23cef10

Made this more explicit.

docs/release-process.md
litt3: 70e23fed28fa28f77d47d6a79a79262612493e8c

litt3: 171b0336b12db8c243ec1f7deb2319cd1cfae2a9

docs/release-process.md
litt3: 171b0336b12db8c243ec1f7deb2319cd1cfae2a9

docs/release-process.md
samlaf: oh. yeah it is, just create a branch-protection rule with this pattern.

docs/release-process.md
samlaf: I really disagree as I think releases and semver should be PURELY tech focused/driven in most cases. Management/BD should not be driving release versioning. An interesting case in point is the typescript compiler, which [does function this way](https://github.com/microsoft/TypeScript/issues/14116#issuecomment-280592571) (see community answer right below this comment).

docs/release-process.md
samlaf: Well that by definition IS what a breaking change is. And we should adhere to semver imo. We are still pre 1.0.0 though so technically we don't have a stable API and we are free to do whatever the heck we want.

So maybe we can/should revisit this after we release 1.0.0?

docs/release-process.md
samlaf: I think at this point it really depends on the exact case, and should be left to the discretion of the release manager.

docs/release-process.md
samlaf: @dmanc what's up with that limitation? I guess its forcing us to make a "noop" commit just to be able to create a proper (non-rc) tag, which isn't that bad but kind of ugly.

dmanc: Not sure, we would need to debug the pipeline but I'm sure it's possible to have 2 tags point to the same commit.

docs/release-process.md
litt3: > So maybe we can/should revisit this after we release 1.0.0?

I think this makes sense

docs/release-process.md
litt3: That thread is an interesting read. A lot of the comments are along the line of: "few projects actually follow semver rules to the letter", which pretty much sums up my gut reaction.

> releases and semver should be PURELY tech focused/driven in most cases

This sounds good, though I'm skeptical about it being practically possible.

I'd be on board with pushing to institute a proper semver compliant release schema, but it's kind of a moot point until we finally go from major version `v0` -> `v1`. Until then, our release version numbers don't really have anything to do with semver at all.

@samlaf Is there a specific part of the current statement that you take issue with?

> Alternatively, management may sign-off that a feature should be included after a feature freeze has been enacted. Note that even with managment sign-off, a PR targetting a release branch must still go through the standard peer-review process.

docs/release-process.md
litt3: The rule already exists AFAIK. I discovered this first hand when I create a branch that started with the word `release`, and it was auto protected

docs/release-process.md
litt3: Since the way we are going to split up releases is a very deep rabbit hole, I'm just not going to specify it in this PR, and we can work on coming to consensus later

docs/release-process.md
litt3: 654895a0cd2aecbb52e6b1f89579ead17f548c0b

docs/release-process.md
litt3: I've updated the document to prescribe use of `rc` tags.

docs/release-process.md
litt3: Ack. LMK if my updates that use `rc` tags address your concern here, or if you feel more changes are needed

cody-littley (approved):

docs/release/release-process.md
```diff
@@ -0,0 +1,116 @@
...
+### 2. **Changes to a Release Branch**
+
+#### Change Policy
+
+- **High bar for inclusion**: Only critical bugfixes or business-critical features
+  - Even bugfixes should not be reflexively included: only high-severity issues
+- **Team consensus required**: Single engineer cannot make the decision
+- **Public visibility**: Must have team discussion (e.g. Slack thread) before proceeding. Alternatively, management
+may sign-off that a feature should be included after a feature freeze has been enacted. Note that even with management
+sign-off, a PR targetting a release branch must still go through the standard peer-review process.
```
dmanc: nit: 
```suggestion
sign-off, a PR targeting a release branch must still go through the standard peer-review process.
```

docs/release/release-example.md
```diff
@@ -0,0 +1,60 @@
...
+
+8. Create tag `v0.10.0-rc.2`
+
+    <img src="images/08-rc-tag.svg" alt="create tag v0.10.0-rc.2" />
+
+9. Create production tag `v0.10.0`
+
+    <img src="images/09-production-tag.svg" alt="create production tag v0.10.0" />
+
+10. Commit `feature 5` to `master`
```
dmanc: Do you think it would be worth describing hotfix vs feature? I think there would be some minor differences in terms of urgency and process.

pschork (approved):

docs/release/release-example.md
litt3: Good point, I'll modify one of the cherry pick features to be a hotfix.

litt3: cee6e3e8815f09762b2606ef13479a6d3f53f9be

docs/release/release-process.md
litt3: TIL cursor/VSCode don't come with a spell checker by default

e382f412f2b4be76ced504247710f6b914720cee

dmanc (dismissed):

dmanc:
When proxy gets merged into this repo are we going to use the same process for the proxy release?

docs/release-process.md
samlaf: > I'd be on board with pushing to institute a proper semver compliant release schema, but it's kind of a moot point until we finally go from major version v0 -> v1. Until then, our release version numbers don't really have anything to do with semver at all.

Agree. We should talk about releasing a `2.0.0` as we've started discussing.

> @samlaf Is there a specific part of the current statement that you take issue with?

I guess my main point is that releases should be made on a fixed cadence (I think this is the main/most important point that CICD philosophy dictates), with whatever the heck is available on master branch at that point. And the semver should be mechanically calculated.
If management/BD needs specific features to be included for a marketting stunt or whatever, then they can request those and wait until a release that contains them is ready.

I'm not fervently adament about following this approach, as I don't have experience having actually done this. But it just feels like its what we should be doing. @MadelineAu is also of this opinion I believe (she mentioned they did release every 2 days at consensys I believe).

docs/release/release-example.md
```diff
@@ -0,0 +1,56 @@
...
+
+2. Cut release branch `release/0.10`
+
+    <img src="images/02-release-branch.svg" alt="cut release branch release/0.10" />
+
+3. Commit `feature 3` to `master`
+
+    <img src="images/03-feature.svg" alt="commit feature 3 to master" />
+
+4. Cherry pick `feature 3` to `release/0.10`
```
samlaf: This doesn't look like what would happen in practice right?
It's more like `release/0.9` exists and is let's say 9 commits behind master. We want to make a new `release/0.10` from the current master HEAD. So we just do that, and then freeze + test for 2 weeks (meanwhile master keeps adding stuff, but we don't touch those).

Or are you trying to show a cherry-pick that is needed because the cut release had bugs/issues that were found during testing?

docs/release/release-process.md
```diff
@@ -0,0 +1,116 @@
...
+
+### 1. **Feature Freeze & Release Branch Creation**
+
+Enacting a feature freeze helps to ensure that the code we publish to production environments is well tested and
+mature. The start of a feature freeze is marked by the creation of a release branch, which allows development
+against `master` to continue uninterrupted while the release is prepared.
+
+#### Plan Feature Freeze
+
+- A feature freeze may be tied either to a date scheduled in advance, or to the completion of a key feature.
```
samlaf, line 24: What do people think about releasing on a fixed cadence? Say every 2 weeks to match the code freeze of the previous release. So a new code-freeze/testing period starts as soon as a new release is made. Or maybe once per month if we find this cadence too fast?

litt3: I'm in favor of a release cadence. I think two weeks might be a little bit frequent, but maybe one month. Let's define this as a follow-up to this PR.

samlaf (approved):
Very happy with where this is going. We chatted about changing https://github.com/Layr-Labs/eigenda/pull/1649#discussion_r2167011293 to use "bugfix" instead of "feature". But there's still some process around this whole runbook that needs to be tried in practice, so we'll prob need to iterate on the document a bit.

Very happy to have this as a guideline that we can point people to and start enforcing though. Next step would be to force releasing on a fixed cadence... but that's another battle for another day.

docs/release/release-example.md
litt3: 24636047e1a0652ec19bceba183c4ef621dd5f3e

> Or are you trying to show a cherry-pick that is needed because the cut release had bugs/issues that were found during testing?

This was the intention. I changed this to say bug fix instead of feature, to hopefully make the intention clearer.

docs/release-process.md
litt3: I agree, there's still some more work that needs to be done to get this process where we want it. Let's keep working on it as we move forward.

samlaf (approved):

dmanc (approved):

PR #1650, opened 2025-06-06
chore: Add type accessor to payload retriever interface

ethenotethan:
Adds a type accessor to payload retriever interface. Useful in eigenda-proxy for telemetry and debug logging. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/disperser_client.go
+++ b/api/clients/v2/disperser_client.go
@@ -14,6 +14,7 @@ import (
 	dispv2 "github.com/Layr-Labs/eigenda/disperser/common/v2"
 	"github.com/Layr-Labs/eigenda/encoding"
 	"github.com/Layr-Labs/eigenda/encoding/rs"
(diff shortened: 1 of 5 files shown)


ethenotethan:
@samlaf wouldn't that require some use of reflection or type assertion?

samlaf:
> @samlaf wouldn't that require some use of reflection or type assertion?

type switch is effectively doing type assertions yes: https://go.dev/tour/methods/16
But that's way more effective (and safe/clean) than hardcoding the type manually as a method in interface.

An interfae is basically a tuple of 2 pointers: one to the data, one to the method table:
image
type assertions/switching just to to follow the method table pointer, and then read the type. Whereas calling a type() method needs to follow the first pointer, find `type()` function in the method table, call it, and read the value.

Claude estimates a 2-5x improvement for using type assertions over what you have in this PR
image

PR #1652, opened 2025-06-09
feat(docs): Secure dispersal/retrieval system flows

ethenotethan:

Closes EGDA-1191

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


docs/spec/src/integration/spec/3-data-structs.md
```diff
@@ -55,13 +55,70 @@ A `blob` is a bn254 field elements array that has a power of 2. It is interprete
...
+The `BlobHeader` contains four main sections that must be constructed. It is passed into the `DisperseBlobRequest` and is signed over for payment authorization.
 
-Refer to the eigenda [protobufs](../../protobufs/generated/disperser_v2.md) documentation for full details of this struct.
+#### Version
+
+The `blobHeader` version refers to one of the `versionedBlobParams` structs defined in the [`EigenDAThresholdRegistry`](./4-contracts.md#eigendathreshold-registry) contract.
+
+#### QuorumNumbers
+
+`QuorumNumbers` represents a list of quorums required to sign and make the blob available. Quorum 0 represents the ETH quorum, quorum 1 represents the EIGEN quorum — both are always required. Custom quorums can also be added to this list.
```
litt3, line 71: > both are always required

I don't think that this will be true anymore, with permissionless dispersal. At least, not under the current plan 

https://linear.app/eigenlabs/issue/DAINT-551/certverifier-and-paymentvault-custom-quorums-disagreement#comment-b187f464

docs/spec/src/integration/spec/3-data-structs.md
```diff
@@ -55,13 +55,70 @@ A `blob` is a bn254 field elements array that has a power of 2. It is interprete
...
+
+#### BlobCommitment
+
+The `BlobCommitment` is a binding commitment to an EigenDA Blob. Due to the length field, a `BlobCommitment` uniquely represents a single `Blob`. It is used by the disperser to prove to EigenDA validators that the chunks they received belong to the original blob (or its Reed-Solomon extension). This commitment can either be computed locally from the blob or generated by the disperser using the `GetBlobCommitment` endpoint.
```
litt3: > Due to the length field,

It seems strange to say that it's due to the length field that a commitment represents a single blob. The length field is necessary, but the statement here makes it seem like it's the star of the show

docs/spec/src/integration/spec/3-data-structs.md
```diff
@@ -55,13 +55,70 @@ A `blob` is a bn254 field elements array that has a power of 2. It is interprete
...
+    
+
+> Note that the length here is the length of the blob (power of 2), which is different from the payload_length encoded as part of the `PayloadHeader` in the `blob` itself (see the [encoding section](#encoding)).
+> 
+
+**PaymentHeader**
+
+The paymentHeader specifies how the blob dispersal to the network will be paid for. There are 2 modes of payment, the permissionless pay-per-blob model and the permissioned reserved-bandwidth approach. See the [Payments](https://docs.eigenda.xyz/releases/payments) release doc for full details; we will only describe how to set these 4 fields here.
```
litt3: link is broken

docs/spec/src/integration/spec/5-lifecycle-phases.md
```diff
@@ -1,94 +1,73 @@
...
+1. *Proxy* takes a raw [payload](./3-data-structs.md#payload) bytes on `/put` endpoint and [converts](#payload-to-blob-encoding) it into a [blob](./3-data-structs.md#blob).
 
-**BlobCommitment**
+1. Using `latest_block_number` (lbn) number fetched from ETH RPC node, *Proxy* calls the router to get the `EigenDACertVerifier` [contract](./4-contracts.md#eigendacertverifier) address *most likely* (if using [`EigenDACertVerifierRouter`](./4-contracts.md#eigendacertverifierrouter)) to be committed to by the `reference_block_number` (rbn) returned by the EigenDA disperser.
 
-The BlobCommitment is a binding commitment for an EigenDA Blob. Because of the length field, a BlobCommitment can only represent a single unique `Blob`. It is also used by the disperser to convince EigenDA validators that the chunks that they have received are indeed part of the blob (or its reed-solomon extension). It can either be computed locally from the blob, or one can ask the disperser to generate it via the `GetBlobCommitment` endpoint.
+2. Using the `verifier`, *Proxy* fetches the `required_quorums` an embeds them into the [`BlobHeader`](./3-data-structs.md#blobheader) as part of the disperser request.
```
litt3:
```suggestion
2. Using the `verifier`, *Proxy* fetches the `required_quorums` and embeds them into the [`BlobHeader`](./3-data-structs.md#blobheader) as part of the disperser request.
```

docs/spec/src/integration/spec/5-lifecycle-phases.md
```diff
@@ -1,94 +1,73 @@
...
 
-The paymentHeader specifies how the blob dispersal to the network will be paid for. There are 2 modes of payment, the permissionless pay-per-blob model and the permissioned reserved-bandwidth approach. See the [Payments](https://docs.eigenda.xyz/releases/payments) release doc for full details; we will only describe how to set these 4 fields here.
+9. If valid, *Proxy* converts the `DACert` into an [`altda-commitment`](./3-data-structs.md#altdacommitment), which is ready for direct submission to the batcher’s inbox without requiring any further processing or transformation by the rollup stack.
+
+
+### Payload to Blob Encoding
+
+This phase occurs inside the eigenda-proxy, because the proxy acts as the “bridge” between the Rollup Domain and Data Availability Domain (see [lifecycle](./2-rollup-payload-lifecycle.md) diagram).
+
+A `payload` consists of an arbitrary byte array. The DisperseBlob endpoint accepts an `encodedPayload`, which needs to be a bn254 field element array.
```
litt3: `DisperseBlob` accepts a `blob`, not an `encodedPayload`

docs/spec/src/integration/spec/5-lifecycle-phases.md
```diff
@@ -102,23 +81,23 @@ enum BlobStatus {
...
 
 Any other terminal status indicates failure, and a new blob dispersal will need to be made.
 
-**Failover to Native Rollup DA**
+#### Failover to Native Rollup DA
 
-The proxy can be configured to retry `UNKNOWN`, `FAILED`, & `COMPLETE` dispersal `n` times, after which it returns to the rollup a `503` HTTP status code which rollup batchers can use to failover to EthDA or native rollup DA offerings (e.g, arbitrum anytrust). See [here](https://github.com/ethereum-optimism/specs/issues/434) for more info on the OP implementation and [here](https://hackmd.io/@epociask/SJUyIZlZkx) for Arbitrum. 
+*Proxy* can be configured to retry `UNKNOWN`, `FAILED`, & `COMPLETE` dispersal `n` times, after which it returns to the rollup a `503` HTTP status code which rollup batchers can use to failover to EthDA or native rollup DA offerings (e.g, arbitrum anytrust). See [here](https://github.com/ethereum-optimism/specs/issues/434) for more info on the OP implementation and [here](https://hackmd.io/@epociask/SJUyIZlZkx) for Arbitrum.
```
litt3: It will only retry on `COMPLETE` if the required thresholds weren't met.

This could be simplified to say "Proxy can be configured to retry unsuccessful dispersal `n` times..."

litt3 (approved):

docs/spec/src/integration/spec/5-lifecycle-phases.md
ethenotethan: @samlaf had suggested we list all status states which prompt retry. can update to reflect the conditional retry assumption on `COMPLETE`

ethenotethan: 37dbb75f470d006e858e7923dca7a242d4117926

docs/spec/src/integration/spec/5-lifecycle-phases.md
ethenotethan: 37dbb75f470d006e858e7923dca7a242d4117926

docs/spec/src/integration/spec/5-lifecycle-phases.md
ethenotethan: 37dbb75f470d006e858e7923dca7a242d4117926

docs/spec/src/integration/spec/3-data-structs.md
ethenotethan: 37dbb75f470d006e858e7923dca7a242d4117926

docs/spec/src/assets/integration/router-in-fraud-proof.png
samlaf, line 1: does the "dispute game" orange line represent anything other than those 2 things being dispute games? It feels like they are related somehow.. bit unclear I find. Maybe just say "dispute games" and point arrows to the 2 boxes?

docs/spec/src/assets/integration/secure-blob-dispersal.png
samlaf, line 1: nit: `eth_blockNum` -> `eth_blockNumber`

samlaf: Also I'm guessing you're just being terse and not precise, but it looks weird that the `DisperseBlobRequest` has a `BlobStatusResponse`. Maybe add a "polling GetBlobStatus" arrow?

samlaf: also do you mind changing `checkDACert(bytes)` -> `checkDACert(abi_encoded_cert)`?

samlaf: Also is it `getQuorums()` or `getRequiredQuorums()`?

docs/spec/src/assets/integration/secure-blob-retrieval.png
samlaf, line 1: I thought we were doing the 1 step approach where we directly call `checkDACert(abi_encoded_cert)` on the Router?

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -76,6 +76,19 @@ Dispersal:
...
 
+## Upgradable Security Params for Optimistic Verification
+![image.png](../../assets/integration/router-in-fraud-proof.png)
+
+The [`EigenDACertVerifierRouter`](./4-contracts.md#eigendacertverifierrouter) contract enables secure upgrades to a rollup’s [EigenDA security parameters](./3-data-structs.md#), such as quorum sets and thresholds, without compromising the integrity of previously submitted fraud proofs. It achieves this by routing certificate verification to the appropriate `EigenDACertVerifier` instance based on the `activation_block_number` associated with each cert's `reference_block_number`. This ensures backward compatibility, allowing older `DACert`s to be validated against the verifier version that was active at the time of their creation.
```
samlaf: bad link for security parameters?

samlaf: I think core uses `securityParams` to include all the chunkSize, maxNumOperators, etc stuff, which router does not enable changing... so we might need another term?

samlaf:
```suggestion
The [`EigenDACertVerifierRouter`](./4-contracts.md#eigendacertverifierrouter) contract enables secure upgrades to a rollup’s [EigenDA security parameters](./3-data-structs.md#), such as required quorums and thresholds, without compromising the integrity of previously submitted fraud proofs. It achieves this by routing certificate verification to the appropriate `EigenDACertVerifier` instance based on the `activation_block_number` associated with each cert's `reference_block_number`. This ensures backward compatibility, allowing older `DACert`s to be validated against the verifier version that was active at the time of their creation.
```

samlaf: "without compromising the integrity of previously submitted fraud proofs"

Don't think its about fraud proofs right? If a zk fraud proof has been successfully subitted for example, it can't be compromised. Think you just mean without affecting on going fraud proof games?

samlaf: "based on the `activation_block_number` associated with each cert's `reference_block_number`"

sounds weird, and not really what is happening. There is no ABN "associated" to a cert's RBN. might confuse people

samlaf: "that was active at the time of their creation"
not true; its not about creation time, its about RBN

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -76,6 +76,19 @@ Dispersal:
...
+
+The router is typically deployed behind an upgradable admin proxy and should use the same `ProxyAdmin` multisig as the rollup for consistent and secure access control.
+
+
+### Adding New Verifiers — Synchronization Risk
+
+There is a synchronization risk that can temporarily cause dispersals to fail when adding a new `verifier'` to the `EigenDACertVerifierRouter` at a future activation block number (`abn'`). If `latest_block < abn'` **and** `rbn >= abn'`, dispersals may fail if the `required_quorums` set differs between `verifier` and `verifier'`. In this case, the quorums included in the client's `BlobHeader` (based on the old verifier) would not match those expected by `checkDACert` (using the new verifier). This mismatch results in **at most** a few failed dispersals, which will resolve once `latest_block >= abn'` and `reference_block_number >= abn'`, ensuring verifier consistency. The EigenDA integrations team will explore mitigations in the future.
```
samlaf, line 89:
```suggestion
There is a synchronization risk that can temporarily cause dispersals to fail when adding a new `verifier'` to the `EigenDACertVerifierRouter` at a future activation block number (`abn'`). If `latest_block < abn' <= rbn`, dispersals may fail if the `required_quorums` set differs between `verifier` and `verifier'`. In this case, the quorums included in the client's `BlobHeader` (based on the old verifier) would not match those expected by `checkDACert` (using the new verifier). This mismatch results in **at most** a few failed dispersals, which will resolve once `latest_block >= abn'` and `reference_block_number >= abn'`, ensuring verifier consistency. The EigenDA integrations team will explore mitigations in the future.
```

samlaf: latest_block is the latest blockchain block? Then I don't think the eqn you have makes sense right? RBN has to be in the past, so how can it be >= latest_block?

docs/spec/src/integration/spec/3-data-structs.md
```diff
@@ -55,13 +55,70 @@ A `blob` is a bn254 field elements array that has a power of 2. It is interprete
 
 An `encodedPayload` can thus be transformed into a `blob` by being padded with 0s to a power of 2, with size currently limited to 16MiB. There is no minimum size, but any blob smaller than 128KiB will be charged for 128KiB.
 
-### BlobHeader And BlobCertificate
+### BlobHeader
 
-The `blobHeader` is submitted alongside the `blob` as part of the DisperseBlob request, and the hash of its rlp serialization (`blobKey` aka `blobHeaderHash`) is a unique identifier for a blob dispersal. This unique identifier is used to retrieve the blob.
+The `blobHeader` is submitted alongside the `blob` as part of the `DisperseBlob` request, and the hash of its RLP serialization (`blobKey`, also known as `blobHeaderHash`) serves as a unique identifier for a blob dispersal. This identifier is used to retrieve the blob.
```
samlaf:
```suggestion
The `blobHeader` is submitted alongside the `blob` as part of the `DisperseBlob` request, and the hash of its ABI encoding (`blobKey`, also known as `blobHeaderHash`) serves as a unique identifier for a blob dispersal. This identifier is used to retrieve the blob.
```

samlaf: think this was on my TODO list for a looong while now. Forgot it was still wrong.

samlaf: maybe we should link to the serialization code?

docs/spec/src/integration/spec/3-data-structs.md
samlaf: Actually I don't even think its true right now. I think the user gets charged for both but they can still specify only a single one afaiu.

docs/spec/src/integration/spec/3-data-structs.md
samlaf: Agree and also I think a cryptographic commitment by definition has this property, which is called the "binding property" (from wiki: "The value chosen during the commit phase must be the only one that the sender can compute and that validates during the reveal phase.")

Maybe we can say something like "The length field is added to the kzgCommitment to respect the binding property"? cc @bxue-l2 for ideas

samlaf: "computed locally" is unclear because the sentence before talks about the disperser. unclear that locally refers to the client's local pov, not the disperser.

docs/spec/src/integration/spec/3-data-structs.md
```diff
@@ -55,13 +55,70 @@ A `blob` is a bn254 field elements array that has a power of 2. It is interprete
...
+    
+
+> Note that the length here is the length of the blob (power of 2), which is different from the payload_length encoded as part of the `PayloadHeader` in the `blob` itself (see the [encoding section](#encoding)).
+> 
+
+**PaymentHeader**
+
+The paymentHeader specifies how the blob dispersal to the network will be paid for. There are 2 modes of payment, the permissionless pay-per-blob model and the permissioned reserved-bandwidth approach. See the [Payments](https://docs.eigenda.xyz/core-concepts/payments#high-level-design) release doc for full details; we will only describe how to set these 4 fields here.
```
samlaf:
```suggestion
The paymentHeader specifies how the blob dispersal to the network will be paid for. There are 2 modes of payment, the permissionless pay-per-blob model and the permissioned reserved-bandwidth approach. See the [Payments](https://docs.eigenda.xyz/core-concepts/payments#high-level-design) release doc for full details; we will only describe how to set these 3 fields here.
```

docs/spec/src/integration/spec/3-data-structs.md
```diff
@@ -55,13 +55,70 @@ A `blob` is a bn254 field elements array that has a power of 2. It is interprete
...
+  // UNIX timestamp in nanoseconds at the time of the dispersal request.
+  // Used to determine the reservation period, for the reserved-bandwidth payment model.
+  int64 timestamp = 2;
+  // Total amount of tokens paid by the requesting account, including the current request.
+  // Used for the pay-per-blob payment model.
+  bytes cumulative_payment = 3;
+}
+```
+
+Users who want to pay-per-blob need to set the cumulative_payment. `timestamp` is used by users who have paid for reserved-bandwidth. If both are set, reserved-bandwidth will be used first, and cumulative_payment only used if the entire bandwidth for the current reservation period has been used up.
```
samlaf, line 120: there will be a lot of subtleties added to this logic with the new separate-payment-per-quorum that is being added. Maybe add a TODO comment to update this?

docs/spec/src/integration/spec/3-data-structs.md
```diff
@@ -55,13 +55,70 @@ A `blob` is a bn254 field elements array that has a power of 2. It is interprete
...
+
+An rpc call to the Disperser’s `GetPaymentState` method can be made to query the current state of an `account_id`. A client can query for this information on startup, cache it, and then update it manually when making pay-per-blob payments. In this way, it can keep track of the cumulative_payment and set it correctly for subsequent dispersals.
```
samlaf:
```suggestion
An RPC call to the Disperser’s `GetPaymentState` method can be made to query the current state of an `account_id`. A client can query for this information on startup, cache it, and then update it manually when making dispersals. In this way, it can keep track of its reserved bandwidth usage and current cumulative_payment and set them correctly for subsequent dispersals.
```

docs/spec/src/integration/spec/5-lifecycle-phases.md
```diff
@@ -1,94 +1,73 @@
 
 # Lifecycle Phases
 
-## Encoding
+Secure interaction between a rollup and EigenDA is composed of three distinct system flows:
 
-This phase occurs inside the eigenda-proxy, because the proxy acts as the “bridge” between the Rollup Domain and Data Availability Domain (see [lifecycle](./2-rollup-payload-lifecycle.md) diagram).
+1. [**Dispersal**](#secure-dispersal): Submitting raw batch data to the DA network  
+2. [**Retrieval**](#secure-retrieval): Fetching batch data from the DA network
```
samlaf: why are you using the term `raw batch`? Because of op batch? Think its better to stick to `payload` which we've defined in this spec?

docs/spec/src/integration/spec/5-lifecycle-phases.md
```diff
@@ -1,94 +1,73 @@
...
 
-## Encoding
+Secure interaction between a rollup and EigenDA is composed of three distinct system flows:
 
-This phase occurs inside the eigenda-proxy, because the proxy acts as the “bridge” between the Rollup Domain and Data Availability Domain (see [lifecycle](./2-rollup-payload-lifecycle.md) diagram).
+1. [**Dispersal**](#secure-dispersal): Submitting raw batch data to the DA network  
+2. [**Retrieval**](#secure-retrieval): Fetching batch data from the DA network  
+3. **Verification**: Ensuring the integrity and quorum-based certification of retrieved data. Where and how verification is performed is often contingent on how an integration is implemented; e.g:
+- *Pessimistic Verification* where a `DACert` is checked as pre-inclusion check for a sequencer inbox
+- *Optimistic Verification* where a `DACert` is only verified in the worst-case challenge
```
samlaf: what does "worst-case challenge" mean?

docs/spec/src/integration/spec/5-lifecycle-phases.md
```diff
@@ -1,94 +1,73 @@
 
 # Lifecycle Phases
 
-## Encoding
+Secure interaction between a rollup and EigenDA is composed of three distinct system flows:
 
-This phase occurs inside the eigenda-proxy, because the proxy acts as the “bridge” between the Rollup Domain and Data Availability Domain (see [lifecycle](./2-rollup-payload-lifecycle.md) diagram).
+1. [**Dispersal**](#secure-dispersal): Submitting raw batch data to the DA network  
+2. [**Retrieval**](#secure-retrieval): Fetching batch data from the DA network  
+3. **Verification**: Ensuring the integrity and quorum-based certification of retrieved data. Where and how verification is performed is often contingent on how an integration is implemented; e.g:
```
samlaf:
```suggestion
3. **Verification**: Ensuring the integrity and quorum-based certification of data availability. Where and how verification is performed is often contingent on how an integration is implemented; e.g:
```

samlaf: if cert verification fails you never retrieve the data (because its not available)

samlaf: This is why I really like @bxue-l2 's idea of calling the `CertVerifier` a `DataAvailabilityVerifier`

docs/spec/src/integration/spec/5-lifecycle-phases.md
```diff
@@ -1,94 +1,73 @@
...
+- *Pessimistic Verification* where a `DACert` is checked as pre-inclusion check for a sequencer inbox
+- *Optimistic Verification* where a `DACert` is only verified in the worst-case challenge
 
-A `payload` consists of an arbitrary byte array. The DisperseBlob endpoint accepts an `encodedPayload`, which needs to be a bn254 field element array.
+EigenDA *Proxy* is used as the main entrypoint for secure dispersal and retrieval.
```
samlaf:
```suggestion
```

samlaf: I would remove this because its not true for zksync and other rollups already, who use the rust client. And we shouldn't prescribe proxy; at most recommend it.

docs/spec/src/integration/spec/5-lifecycle-phases.md
```diff
@@ -1,94 +1,73 @@
...
 
-## BlobHeader Construction
+## Glossary
+- *Rollup Batcher*: Sequencer rollup node component responsible for constructing and submitting user transaction batches
+- *Rollup Nodes*: Refers to any rollup node (e,g, validator, verifier) which syncs current chain state through an onchain sequencer inbox
+- *Proxy*: Side car server as a part of rollup and used for secure and trustless communication with EigenDA
```
samlaf: Fine for now but I feel like we should move the Glossary to its own separate page and grow it. Or maybe just point to https://docs.eigenda.xyz/core-concepts/glossary?

docs/spec/src/integration/spec/5-lifecycle-phases.md
```diff
@@ -1,94 +1,73 @@
...
+3. **Verification**: Ensuring the integrity and quorum-based certification of retrieved data. Where and how verification is performed is often contingent on how an integration is implemented; e.g:
+- *Pessimistic Verification* where a `DACert` is checked as pre-inclusion check for a sequencer inbox
+- *Optimistic Verification* where a `DACert` is only verified in the worst-case challenge
 
-A `payload` consists of an arbitrary byte array. The DisperseBlob endpoint accepts an `encodedPayload`, which needs to be a bn254 field element array.
+EigenDA *Proxy* is used as the main entrypoint for secure dispersal and retrieval.
 
-## BlobHeader Construction
+## Glossary
+- *Rollup Batcher*: Sequencer rollup node component responsible for constructing and submitting user transaction batches
```
samlaf:
```suggestion
- *Rollup Batcher*: Sequencer rollup node component responsible for constructing and submitting to the settlement chain user transaction batches
```

docs/spec/src/assets/integration/secure-blob-retrieval.png
ethenotethan: nope not currently: https://github.com/Layr-Labs/eigenda/blob/master/api/clients/v2/verification/cert_verifier.go#L82-L89

docs/spec/src/integration/spec/5-lifecycle-phases.md
ethenotethan: > if cert verification fails you never retrieve the data (because its not available) 

hmm I don't think thats true - what about a cert with an insufficient quorum threshold?

docs/spec/src/integration/spec/5-lifecycle-phases.md
ethenotethan: would rather not port it to our docs site - can create a separate page!

docs/spec/src/integration/spec/6-secure-integration.md
ethenotethan: rbn is selected by the disperser and latest_block is selected eigenda client rpc. For this edge case to happen the disperser has to select rbn >= latest_block selection. maybe a different variable name would help smooth this

docs/spec/src/integration/spec/5-lifecycle-phases.md
ethenotethan: [50c5203](https://github.com/Layr-Labs/eigenda/pull/1652/commits/50c52032d02541b047a847c2a5d60238c4b7368a)

docs/spec/src/integration/spec/5-lifecycle-phases.md
ethenotethan: [50c5203](https://github.com/Layr-Labs/eigenda/pull/1652/commits/50c52032d02541b047a847c2a5d60238c4b7368a)

docs/spec/src/integration/spec/5-lifecycle-phases.md
ethenotethan: [50c5203](https://github.com/Layr-Labs/eigenda/pull/1652/commits/50c52032d02541b047a847c2a5d60238c4b7368a)

docs/spec/src/integration/spec/5-lifecycle-phases.md
ethenotethan: [50c5203](https://github.com/Layr-Labs/eigenda/pull/1652/commits/50c52032d02541b047a847c2a5d60238c4b7368a)

docs/spec/src/integration/spec/5-lifecycle-phases.md
ethenotethan: [50c5203](https://github.com/Layr-Labs/eigenda/pull/1652/commits/50c52032d02541b047a847c2a5d60238c4b7368a)

docs/spec/src/integration/spec/3-data-structs.md
ethenotethan: [50c5203](https://github.com/Layr-Labs/eigenda/pull/1652/commits/50c52032d02541b047a847c2a5d60238c4b7368a)

docs/spec/src/integration/spec/3-data-structs.md
ethenotethan: [50c5203](https://github.com/Layr-Labs/eigenda/pull/1652/commits/50c52032d02541b047a847c2a5d60238c4b7368a)

docs/spec/src/integration/spec/3-data-structs.md
ethenotethan: [50c5203](https://github.com/Layr-Labs/eigenda/pull/1652/commits/50c52032d02541b047a847c2a5d60238c4b7368a)

docs/spec/src/integration/spec/3-data-structs.md
ethenotethan: [50c5203](https://github.com/Layr-Labs/eigenda/pull/1652/commits/50c52032d02541b047a847c2a5d60238c4b7368a)

docs/spec/src/integration/spec/3-data-structs.md
ethenotethan: [50c5203](https://github.com/Layr-Labs/eigenda/pull/1652/commits/50c52032d02541b047a847c2a5d60238c4b7368a)

docs/spec/src/assets/integration/secure-blob-dispersal.png
ethenotethan: [50c5203](https://github.com/Layr-Labs/eigenda/pull/1652/commits/50c52032d02541b047a847c2a5d60238c4b7368a)

docs/spec/src/assets/integration/router-in-fraud-proof.png
ethenotethan: [50c5203](https://github.com/Layr-Labs/eigenda/pull/1652/commits/50c52032d02541b047a847c2a5d60238c4b7368a)

docs/spec/src/integration/spec/3-data-structs.md
ethenotethan: [50c5203](https://github.com/Layr-Labs/eigenda/pull/1652/commits/50c52032d02541b047a847c2a5d60238c4b7368a)

docs/spec/src/assets/integration/router-in-fraud-proof.png
ethenotethan: [7b90fb8](https://github.com/Layr-Labs/eigenda/pull/1652/commits/7b90fb82aa7c9580780cf79ef524ddc367fc388e)

docs/spec/src/assets/integration/secure-blob-retrieval.png
samlaf: Why are we doing it this way? Seems like extra unneeded logic? The Router's checkDACert() already does [getVerifierAt()](https://github.com/Layr-Labs/eigenda/blob/f545b16ffccfb6b9ac1b603bdb0770fb2f09c140/contracts/src/periphery/cert/router/EigenDACertVerifierRouter.sol#L29).
So basically router and CertVerifier have same interface, so we should just call that interface directly no?

docs/spec/src/integration/spec/6-secure-integration.md
samlaf: `latest_block` is definitely confusing as its a relative term. Whose latest_block are we talking about, and at what time? disperser? rollup node during verification? etc

But despite this I still don't understand your explanation. The disperser literally cannot select RBN >= latest_block. Otherwise eigenda validators won't be able to sign until the RBN is reached. Or is that the case you are talking about?

docs/spec/src/integration/spec/5-lifecycle-phases.md
samlaf: thanks! still need to chang the retrieval `batch data` to `payload` as well though

docs/spec/src/integration/spec/5-lifecycle-phases.md
samlaf: Ah, well you *could* retrieve the data in that case, but you won't because its pointless as that cert is considered invalid from the rollup's DP's POV.

docs/spec/src/integration/spec/5-lifecycle-phases.md
```diff
@@ -24,23 +17,21 @@ EigenDA *Proxy* is used as the main entrypoint for secure dispersal and retrieva
...
-
-1. Using `latest_block_number` (lbn) number fetched from ETH RPC node, *Proxy* calls the router to get the `EigenDACertVerifier` [contract](./4-contracts.md#eigendacertverifier) address *most likely* (if using [`EigenDACertVerifierRouter`](./4-contracts.md#eigendacertverifierrouter)) to be committed to by the `reference_block_number` (rbn) returned by the EigenDA disperser.
+1. *EigenDA Client* takes a raw [payload](./3-data-structs.md#payload) bytes and [converts](#payload-to-blob-encoding) it into a [blob](./3-data-structs.md#blob).
 
-2. Using the `verifier`, *Proxy* fetches the `required_quorums` and embeds them into the [`BlobHeader`](./3-data-structs.md#blobheader) as part of the disperser request.
+1. Using `latest_block_number` (lbn) number fetched from ETH RPC node, *EigenDA Client* calls the router to get the `EigenDACertVerifier` [contract](./4-contracts.md#eigendacertverifier) address *most likely* (if using [`EigenDACertVerifierRouter`](./4-contracts.md#eigendacertverifierrouter)) to be committed to by the `reference_block_number` (rbn) returned by the EigenDA disperser.
```
samlaf: Instead of using bold `*EigenDA Client*` everywhere, should we instead use [EigenDA Client] and add a markdown link at the bottom of the page:

```
[EigenDA Client]: ../../glossary.md
```

samlaf: op does this in their docs, which is nice to always send people to the glossary

samlaf (commented):
Thanks for all the changes! Looking very clean. Should be mergeable soon.

Still a few comments you haven't addressed. Do we need to sync/discuss them in person? Or have you just not gotten to them yet?

docs/spec/src/integration/spec/5-lifecycle-phases.md
```diff
@@ -24,30 +17,28 @@ EigenDA *Proxy* is used as the main entrypoint for secure dispersal and retrieva
 
 
 ### System Flow
-1. *Proxy* takes a raw [payload](./3-data-structs.md#payload) bytes on `/put` endpoint and [converts](#payload-to-blob-encoding) it into a [blob](./3-data-structs.md#blob).
-
-1. Using `latest_block_number` (lbn) number fetched from ETH RPC node, *Proxy* calls the router to get the `EigenDACertVerifier` [contract](./4-contracts.md#eigendacertverifier) address *most likely* (if using [`EigenDACertVerifierRouter`](./4-contracts.md#eigendacertverifierrouter)) to be committed to by the `reference_block_number` (rbn) returned by the EigenDA disperser.
+1. *EigenDA Client* takes a raw [payload](./3-data-structs.md#payload) bytes and [converts](#payload-to-blob-encoding) it into a [blob](./3-data-structs.md#blob).
```
litt3: nit: You have `1.` used twice in this list

docs/spec/src/integration/spec/3-data-structs.md
bxue-l2: KZG commitment is binding to polynomial, not the binary data which blob is consisted of. 

so a length field is somehow adding a binding condition between the length of blob and the  polynomial

docs/spec/src/integration/spec/5-lifecycle-phases.md
```diff
@@ -24,30 +17,28 @@ EigenDA *Proxy* is used as the main entrypoint for secure dispersal and retrieva
...
-9. If valid, *Proxy* converts the `DACert` into an [`altda-commitment`](./3-data-structs.md#altdacommitment), which is ready for direct submission to the batcher’s inbox without requiring any further processing or transformation by the rollup stack.
+8. Using the `verification_status_code`, *EigenDA Client* determines whether to return the certificate (`CertV2Lib.StatusCode.SUCCESS`) to the *Rollup Batcher* or retry a subsequent dispersal attempt
 
 
 ### Payload to Blob Encoding
 
 This phase occurs inside the eigenda-proxy, because the proxy acts as the “bridge” between the Rollup Domain and Data Availability Domain (see [lifecycle](./2-rollup-payload-lifecycle.md) diagram).
 
-A `payload` consists of an arbitrary byte array. The DisperseBlob endpoint accepts an `encodedPayload`, which needs to be a bn254 field element array.
+A `payload` consists of an arbitrary byte array. The DisperseBlob endpoint accepts an `blob`, which needs to be an encoded bn254 field element array.
```
litt3: nit: ~~an~~ *a* `blob`

litt3 (approved):

docs/spec/src/integration/spec/6-secure-integration.md
ethenotethan: your deduction assumes `latest_block` as a global value btwn client and disperser

docs/spec/src/integration/spec/5-lifecycle-phases.md
ethenotethan: 9d35a2438b9d4e928bbab9d95f52d55815294cc7

docs/spec/src/integration/spec/5-lifecycle-phases.md
ethenotethan: 9d35a2438b9d4e928bbab9d95f52d55815294cc7

docs/spec/src/integration/spec/5-lifecycle-phases.md
ethenotethan: 9d35a2438b9d4e928bbab9d95f52d55815294cc7

docs/spec/src/integration/spec/5-lifecycle-phases.md
ethenotethan: 9d35a2438b9d4e928bbab9d95f52d55815294cc7

docs/spec/src/assets/integration/secure-blob-retrieval.png
ethenotethan: not opposed to making this change assuming the onchain verification is isomorphic in its implementation. Ie onchain calls the router directly - but if we're doing two step onchain then we might wanna leave as is

cc @litt3

litt3 (dismissed):

docs/spec/src/integration/spec/6-secure-integration.md
ethenotethan: 4252201a27e3005021e1c489ef5aa643096c0ef9

docs/spec/src/integration/spec/3-data-structs.md
samlaf: Can we say then that kzg commitment binds to the mathematical polynomial, but that in the specific coefficient representation, many vectors of coefficients map to that polynomial (pad with zeros), so the length commitment is needed to map to an exact coefficient vector?

docs/spec/src/assets/integration/secure-blob-retrieval.png
samlaf: > if we're doing two step onchain then we might wanna leave as is

There is no way to do a "one step" onchain. Someone has to lookup the cert verifier at the given RBN number. The fact that its already done onchain in the router allows the offchain to be one-step though (just call that function which will do it for you onchain).

docs/spec/src/integration/spec/6-secure-integration.md
samlaf: Are you saying `RBN >= some_random_validator_whose_eth_node_is_out_of_sync.latest_block`?

samlaf (approved):
There's a few very niche and ultra detail-specific conversations left open that might be worth getting to the bottom of (maybe during TFFT?), but overall happy with merging this PR as is.

docs/spec/src/assets/integration/secure-blob-retrieval.png
ethenotethan: - no performance nor security benefits to either solution

will leave as is

docs/spec/src/integration/spec/6-secure-integration.md
ethenotethan: rename `latest_block` to `quorum_selection_block`

PR #1653, opened 2025-06-09
refactor: payment common functions

hopeyen:

- Refactoring some common functions for payments that would be shared across disperser meterer, client accountant, and in a future PR validator node batch meterer
- Minor refactoring of the disperser client accountant
- More specific accountant errors

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/accountant.go
+++ b/api/clients/v2/accountant.go
@@ -74,9 +74,9 @@ func (a *Accountant) blobPaymentInfo(
 	symbolUsage := meterer.SymbolsCharged(numSymbols, a.minNumSymbols)
 
 	// Always try to use reservation first
-	payment, err := a.reservationUsage(symbolUsage, quorumNumbers, timestamp)
+	err := a.reservationUsage(symbolUsage, quorumNumbers, timestamp)
 	if err == nil {
-		return payment, nil
+		return big.NewInt(0), nil
 	}
 
 	// Fall back to on-demand payment if reservation fails
@@ -88,7 +88,7 @@ func (a *Accountant) blobPaymentInfo(
 func (a *Accountant) reservationUsage(
 	symbolUsage uint64,
 	quorumNumbers []uint8,
-	timestamp int64) (*big.Int, error) {
+	timestamp int64) error {
 
 	reservationWindow := a.reservationWindow
 	currentReservationPeriod := meterer.GetReservationPeriodByNanosecond(timestamp, reservationWindow)
@@ -102,32 +102,29 @@ func (a *Accountant) reservationUsage(
 	binLimit := a.reservation.SymbolsPerSecond * uint64(a.reservationWindow)
 	if relativePeriodRecord.Usage <= binLimit {
 		if err := meterer.ValidateQuorum(quorumNumbers, a.reservation.QuorumNumbers); err != nil {
-			return nil, err
+			return err
 		}
-		return big.NewInt(0), nil
+		return nil
 	}
 
 	overflowPeriodRecord := a.getOrRefreshRelativePeriodRecord(currentReservationPeriod+2*reservationWindow, reservationWindow)
 	// Allow one overflow when the overflow bin is empty, the current usage and new length are both less than the limit
 	if overflowPeriodRecord.Usage == 0 && relativePeriodRecord.Usage-symbolUsage < binLimit && symbolUsage <= binLimit {
 		if err := meterer.ValidateQuorum(quorumNumbers, a.reservation.QuorumNumbers); err != nil {
-			return nil, err
+			return err
 		}
 		overflowPeriodRecord.Usage += relativePeriodRecord.Usage - binLimit
-		return big.NewInt(0), nil
+		return nil
 	}
 
 	// Reservation not sufficient for the request, rollback the usage
 	relativePeriodRecord.Usage -= symbolUsage
-	return nil, fmt.Errorf("insufficient reservation")
+	return fmt.Errorf("insufficient reservation")
 }
 
 // onDemandUsage attempts to use on-demand payment for the given request.
 // Returns the cumulative payment if successful, or an error if on-demand cannot be used.
-func (a *Accountant) onDemandUsage(
-	symbolUsage uint64,
-	quorumNumbers []uint8) (*big.Int, error) {
-
+func (a *Accountant) onDemandUsage(symbolUsage uint64, quorumNumbers []uint8) (*big.Int, error) {
 	incrementRequired := meterer.PaymentCharged(symbolUsage, a.pricePerSymbol)
 	resultingPayment := big.NewInt(0)
 	resultingPayment.Add(a.cumulativePayment, incrementRequired)
--- a/api/clients/v2/accountant.go
+++ b/api/clients/v2/accountant.go
@@ -74,48 +74,75 @@ func (a *Accountant) BlobPaymentInfo(
 	quorumNumbers []uint8,
 	timestamp int64) (*big.Int, error) {
 
+	// Always try to use reservation first
+	payment, err := a.ReservationUsage(numSymbols, quorumNumbers, timestamp)
+	if err == nil {
+		return payment, nil
+	}
+
+	// Fall back to on-demand payment if reservation fails
+	return a.OnDemandUsage(numSymbols, quorumNumbers)
+}
+
+// ReservationUsage attempts to use the reservation for the given request.
+// Returns (0, nil) if successful, or (nil, error) if reservation cannot be used.
+func (a *Accountant) ReservationUsage(
+	numSymbols uint64,
+	quorumNumbers []uint8,
+	timestamp int64) (*big.Int, error) {
+
 	currentReservationPeriod := meterer.GetReservationPeriodByNanosecond(timestamp, a.reservationWindow)
 	symbolUsage := a.SymbolsCharged(numSymbols)
 
 	a.usageLock.Lock()
 	defer a.usageLock.Unlock()
+
 	relativePeriodRecord := a.GetRelativePeriodRecord(currentReservationPeriod)
 	relativePeriodRecord.Usage += symbolUsage
 
-	// first attempt to use the active reservation
+	// Check if we can use the reservation within the bin limit
 	binLimit := a.reservation.SymbolsPerSecond * uint64(a.reservationWindow)
 	if relativePeriodRecord.Usage <= binLimit {
 		if err := QuorumCheck(quorumNumbers, a.reservation.QuorumNumbers); err != nil {
-			return big.NewInt(0), err
+			return nil, err
 		}
 		return big.NewInt(0), nil
 	}
 
+	// Try to use overflow bin if available
 	overflowPeriodRecord := a.GetRelativePeriodRecord(currentReservationPeriod + 2)
-	// Allow one overflow when the overflow bin is empty, the current usage and new length are both less than the limit
 	if overflowPeriodRecord.Usage == 0 && relativePeriodRecord.Usage-symbolUsage < binLimit && symbolUsage <= binLimit {
 		if err := QuorumCheck(quorumNumbers, a.reservation.QuorumNumbers); err != nil {
-			return big.NewInt(0), err
+			return nil, err
 		}
 		overflowPeriodRecord.Usage += relativePeriodRecord.Usage - binLimit
 		return big.NewInt(0), nil
 	}
 
-	// reservation not available, rollback reservation records, attempt on-demand
-	//todo: rollback on-demand if disperser respond with some type of rejection?
+	// Reservation not available, rollback the usage
 	relativePeriodRecord.Usage -= symbolUsage
-	incrementRequired := big.NewInt(int64(a.PaymentCharged(numSymbols)))
+	return nil, fmt.Errorf("reservation not available")
+}
 
+// OnDemandUsage attempts to use on-demand payment for the given request.
+// Returns the cumulative payment if successful, or an error if on-demand cannot be used.
+func (a *Accountant) OnDemandUsage(
+	numSymbols uint64,
+	quorumNumbers []uint8) (*big.Int, error) {
+
+	incrementRequired := big.NewInt(int64(a.PaymentCharged(numSymbols)))
 	resultingPayment := big.NewInt(0)
 	resultingPayment.Add(a.cumulativePayment, incrementRequired)
+
 	if resultingPayment.Cmp(a.onDemand.CumulativePayment) <= 0 {
 		if err := QuorumCheck(quorumNumbers, requiredQuorums); err != nil {
-			return big.NewInt(0), err
+			return nil, err
 		}
 		a.cumulativePayment.Add(a.cumulativePayment, incrementRequired)
 		return a.cumulativePayment, nil
 	}
-	return big.NewInt(0), fmt.Errorf(
+
+	return nil, fmt.Errorf(
 		"no bandwidth reservation found for account %s, and current cumulativePayment balance insufficient "+
 			"to make an on-demand dispersal. Consider depositing more eth to the PaymentVault contract.", a.accountID.Hex())
 }
--- a/api/clients/v2/accountant.go
+++ b/api/clients/v2/accountant.go
@@ -61,32 +61,6 @@ func NewAccountant(accountID gethcommon.Address, reservation *core.ReservedPayme
 	return &a
 }
 
-// blobPaymentInfo calculates and records payment information. The accountant
-// will attempt to use the active reservation first and check for quorum settings,
-// then on-demand if the reservation is not available. It takes in a timestamp at
-// the current UNIX time in nanoseconds, and returns a cumulative payment for on-
-// demand payments in units of wei. Both timestamp and cumulative payment are used
-// to create the payment header and signature, with non-zero cumulative payment
-// indicating on-demand payment.
-// These generated values are used to create the payment header and signature, as specified in
-// api/proto/common/v2/common_v2.proto
-// func (a *Accountant) blobPaymentInfo(
-// 	numSymbols uint64,
-// 	quorumNumbers []uint8,
-// 	timestamp int64) (*big.Int, error) {
-
-// 		symbolUsage := meterer.SymbolsCharged(numSymbols, a.minNumSymbols)
-
-// 		// Always try to use reservation first
-// 		err := a.reservationUsage(symbolUsage, quorumNumbers, timestamp)
-// 		if err == nil {
-// 			return big.NewInt(0), nil
-// 		}
-
-// 		// Fall back to on-demand payment if reservation fails
-// 		return a.onDemandUsage(symbolUsage, quorumNumbers)
-// }
-
 // reservationUsage attempts to use the reservation for the given request.
 func (a *Accountant) reservationUsage(
 	symbolUsage uint64,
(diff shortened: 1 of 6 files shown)


api/clients/v2/accountant.go
```diff
@@ -74,50 +73,76 @@ func (a *Accountant) BlobPaymentInfo(
...
+	// Fall back to on-demand payment if reservation fails
+	return a.OnDemandUsage(symbolUsage, quorumNumbers)
+}
+
+// ReservationUsage attempts to use the reservation for the given request.
+// Returns (0, nil) if successful, or (nil, error) if reservation cannot be used.
+func (a *Accountant) ReservationUsage(
+	symbolUsage uint64,
+	quorumNumbers []uint8,
+	timestamp int64) (*big.Int, error) {
```
dmanc: Instead of returning (*big.Int, error), we could just return error right? Then in BlobPaymentInfo we just do 
```
	payment, err := a.ReservationUsage(symbolUsage, quorumNumbers, timestamp)
	if err == nil {
		return big.NewInt(0), nil
	}
```

api/clients/v2/accountant.go
```diff
@@ -74,50 +73,76 @@ func (a *Accountant) BlobPaymentInfo(
...
-			return big.NewInt(0), err
+		if err := meterer.ValidateQuorum(quorumNumbers, a.reservation.QuorumNumbers); err != nil {
+			return nil, err
 		}
 		overflowPeriodRecord.Usage += relativePeriodRecord.Usage - binLimit
 		return big.NewInt(0), nil
 	}
 
-	// reservation not available, rollback reservation records, attempt on-demand
-	//todo: rollback on-demand if disperser respond with some type of rejection?
```
dmanc, line 103: Is this todo captured somewhere / resolved?

api/clients/v2/accountant.go
```diff
@@ -74,50 +73,76 @@ func (a *Accountant) BlobPaymentInfo(
...
+	// Reservation not sufficient for the request, rollback the usage
 	relativePeriodRecord.Usage -= symbolUsage
-	incrementRequired := big.NewInt(int64(a.PaymentCharged(numSymbols)))
+	return nil, fmt.Errorf("insufficient reservation")
+}
+
+// OnDemandUsage attempts to use on-demand payment for the given request.
+// Returns the cumulative payment if successful, or an error if on-demand cannot be used.
+func (a *Accountant) OnDemandUsage(
+	symbolUsage uint64,
```
dmanc: nit: would look cleaner on a single line

api/clients/v2/accountant.go
hopeyen: makes sense! [updated](https://github.com/Layr-Labs/eigenda/pull/1653/commits/510fe1a875d031d2a9ce022ad8751582719e0f85)

api/clients/v2/accountant.go
hopeyen: added to [commit](https://github.com/Layr-Labs/eigenda/pull/1653/commits/510fe1a875d031d2a9ce022ad8751582719e0f85); probably not going to be done anytime soon

api/clients/v2/accountant.go
hopeyen: cool, updated in [commit](https://github.com/Layr-Labs/eigenda/pull/1653/commits/510fe1a875d031d2a9ce022ad8751582719e0f85)

api/clients/v2/accountant.go
```diff
@@ -71,51 +70,74 @@ func (a *Accountant) blobPaymentInfo(
...
+	if err == nil {
+		return big.NewInt(0), nil
+	}
+
+	// Fall back to on-demand payment if reservation fails
+	return a.onDemandUsage(symbolUsage, quorumNumbers)
+}
+
+// reservationUsage attempts to use the reservation for the given request.
+// Returns (0, nil) if successful, or (nil, error) if reservation cannot be used.
```
dmanc: Comment needs to be updated

api/clients/v2/accountant.go
```diff
@@ -71,51 +70,74 @@ func (a *Accountant) blobPaymentInfo(
...
+			return nil, err
 		}
 		a.cumulativePayment.Add(a.cumulativePayment, incrementRequired)
 		return a.cumulativePayment, nil
 	}
-	return big.NewInt(0), fmt.Errorf(
-		"invalid payments: no available bandwidth reservation found for account %s, and current cumulativePayment balance insufficient "+
-			"to make an on-demand dispersal. Consider increasing reservation or cumulative payment on-chain. For more details, see https://docs.eigenda.xyz/core-concepts/payments#disperser-client-requirements", a.accountID.Hex())
+
+	return nil, fmt.Errorf("insufficient ondemand payment")
```
dmanc: nit: We should just use errors.new if we aren't doing any formatting. Same comment in other areas.

api/clients/v2/accountant.go
```diff
@@ -71,51 +70,74 @@ func (a *Accountant) blobPaymentInfo(
...
 	relativePeriodRecord := a.getOrRefreshRelativePeriodRecord(currentReservationPeriod, reservationWindow)
 	relativePeriodRecord.Usage += symbolUsage
 
-	// first attempt to use the active reservation
+	// Check if we can use the reservation within the bin limit
 	binLimit := a.reservation.SymbolsPerSecond * uint64(a.reservationWindow)
 	if relativePeriodRecord.Usage <= binLimit {
-		if err := QuorumCheck(quorumNumbers, a.reservation.QuorumNumbers); err != nil {
-			return big.NewInt(0), err
+		if err := meterer.ValidateQuorum(quorumNumbers, a.reservation.QuorumNumbers); err != nil {
```
dmanc: Do we need to rollback relativePeriodRecord.Usage if ValidateQuorum fails?

dmanc: Same question in the other error condition inside the overflow logic.

api/clients/v2/accountant.go
```diff
@@ -71,51 +70,74 @@ func (a *Accountant) blobPaymentInfo(
...
-	symbolUsage := a.symbolsCharged(numSymbols)
 
 	a.usageLock.Lock()
 	defer a.usageLock.Unlock()
 	relativePeriodRecord := a.getOrRefreshRelativePeriodRecord(currentReservationPeriod, reservationWindow)
 	relativePeriodRecord.Usage += symbolUsage
 
-	// first attempt to use the active reservation
+	// Check if we can use the reservation within the bin limit
 	binLimit := a.reservation.SymbolsPerSecond * uint64(a.reservationWindow)
```
dmanc: we can use GetReservationBinLimit here instead?

api/clients/v2/accountant.go
```diff
@@ -71,51 +70,74 @@ func (a *Accountant) blobPaymentInfo(
...
-		return big.NewInt(0), nil
+		return nil
 	}
 
 	overflowPeriodRecord := a.getOrRefreshRelativePeriodRecord(currentReservationPeriod+2*reservationWindow, reservationWindow)
 	// Allow one overflow when the overflow bin is empty, the current usage and new length are both less than the limit
 	if overflowPeriodRecord.Usage == 0 && relativePeriodRecord.Usage-symbolUsage < binLimit && symbolUsage <= binLimit {
-		if err := QuorumCheck(quorumNumbers, a.reservation.QuorumNumbers); err != nil {
-			return big.NewInt(0), err
+		if err := meterer.ValidateQuorum(quorumNumbers, a.reservation.QuorumNumbers); err != nil {
```
dmanc: Can we move this ValidateQuorum check much earlier in the function?

api/clients/v2/accountant.go
```diff
@@ -71,51 +70,74 @@ func (a *Accountant) blobPaymentInfo(
...
-	//todo: rollback on-demand if disperser respond with some type of rejection?
+	// Reservation not sufficient for the request, rollback the usage
 	relativePeriodRecord.Usage -= symbolUsage
-	incrementRequired := big.NewInt(int64(a.paymentCharged(numSymbols)))
+	return fmt.Errorf("insufficient reservation")
+}
 
+// onDemandUsage attempts to use on-demand payment for the given request.
+// Returns the cumulative payment if successful, or an error if on-demand cannot be used.
+func (a *Accountant) onDemandUsage(symbolUsage uint64, quorumNumbers []uint8) (*big.Int, error) {
```
dmanc, line 104: why does on-demand not require a lock but reservations does?

api/clients/v2/accountant.go
```diff
@@ -71,51 +70,74 @@ func (a *Accountant) blobPaymentInfo(
...
+	incrementRequired := meterer.PaymentCharged(symbolUsage, a.pricePerSymbol)
 	resultingPayment := big.NewInt(0)
 	resultingPayment.Add(a.cumulativePayment, incrementRequired)
+
 	if resultingPayment.Cmp(a.onDemand.CumulativePayment) <= 0 {
-		if err := QuorumCheck(quorumNumbers, requiredQuorums); err != nil {
-			return big.NewInt(0), err
+		if err := meterer.ValidateQuorum(quorumNumbers, requiredQuorums); err != nil {
```
dmanc: could we move this check to the beginning?

api/clients/v2/accountant.go
```diff
@@ -126,7 +148,7 @@ func (a *Accountant) AccountBlob(
 
 	cumulativePayment, err := a.blobPaymentInfo(numSymbols, quorums, timestamp)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("cannot create payment infomation for reservation or on-demand. Consider depositing more eth to the PaymentVault contract for your account. For more details, see https://docs.eigenda.xyz/core-concepts/payments#disperser-client-requirements. Account: %s, Error: %s", a.accountID.Hex(), err.Error())
```
samlaf:
```suggestion
		return nil, fmt.Errorf("cannot create payment infomation for reservation or on-demand. Consider depositing more eth to the PaymentVault contract for your account. For more details, see https://docs.eigenda.xyz/core-concepts/payments#disperser-client-requirements. Account: %s, Error: %w", a.accountID.Hex(), err)
```

samlaf: typically more idiomatic to wrap errors, unless you were really trying to hide the lower level errors?

api/clients/v2/accountant.go
```diff
@@ -71,51 +70,74 @@ func (a *Accountant) blobPaymentInfo(
...
+	symbolUsage := meterer.SymbolsCharged(numSymbols, a.minNumSymbols)
+
+	// Always try to use reservation first
+	err := a.reservationUsage(symbolUsage, quorumNumbers, timestamp)
+	if err == nil {
+		return big.NewInt(0), nil
+	}
+
+	// Fall back to on-demand payment if reservation fails
+	return a.onDemandUsage(symbolUsage, quorumNumbers)
```
samlaf: why do we need this function? I would just get rid of this and put it inside `AccountBlob` directly. This function just has weird unclear semantics. Think its better to keep reservationUsage and onDemandUsage separate.

api/clients/v2/accountant.go
```diff
@@ -71,51 +70,74 @@ func (a *Accountant) blobPaymentInfo(
...
+func (a *Accountant) reservationUsage(
+	symbolUsage uint64,
+	quorumNumbers []uint8,
+	timestamp int64) error {
+
 	reservationWindow := a.reservationWindow
 	currentReservationPeriod := meterer.GetReservationPeriodByNanosecond(timestamp, reservationWindow)
-	symbolUsage := a.symbolsCharged(numSymbols)
 
 	a.usageLock.Lock()
```
samlaf: why is this called usageLock and not reservationLock?

samlaf: also what EXACTLY is it protecting? only the periodRecord struct? If so can prob rename to `periodRecordMutex`? or at least add comment to that effect

api/clients/v2/accountant.go
samlaf: Good point. Why do you not do `relativePeriodRecord.Usage += symbolUsage` at the very bottom of the function just for returning? I'm assuming it's used somewhere but its very hard to follow which part of this function is stateful, which part is stateless, etc.

As a general programming philosophy recommendation, trying to follow an "IO sandwich" (or [impureim sandwich](https://blog.ploeh.dk/2020/03/02/impureim-sandwich/)) is often the best way to keep code clean and easy to understand/refactor.

api/clients/v2/disperser_client.go
```diff
@@ -292,6 +292,8 @@ func (c *disperserClient) DisperseBlobWithProbe(
 
 	reply, err := c.client.DisperseBlob(ctx, request)
 	if err != nil {
+		// TODO: rollback payment for the accountant if the blob fails to disperse
+		// because ondemand request hits global ratelimit.
```
samlaf, line 296: where is this rollback supposed to happen? who is responsible for doing this? the accountant or the client? Seems like you're suggesting its the clients responsibility to do this? Why not keep it all completely hidden from the client and do it in the accountant?

samlaf (commented):
You mention in your PR description: "Refactoring some common functions for payments that would be shared across disperser meterer, client accountant, and in a future PR validator node batch meterer".

My question is: why do we have a separate struct for all of these things, which iiuc are doing very similar things? Shouldn't we have a basic low-level accountant, which is used by the disperser meter and node batch meter? Instead of just sharing functions like this PR is doing.

api/clients/v2/accountant.go
hopeyen: No rollbacks intended if `ValidateQuorum` fails, this is the behavior I kept from how disperser meterer behaves. Although, thinking about this again, since this is the client-side accountant, we can fail early without penalties like the disperser does.


Moved `ValidateQuorum` to be at the top of the function before recording usage.

`relativePeriodRecord.Usage += symbolUsage` is earlier within this function because we wanted to write first before checking the conditions in case there are concurrent requests, and only rollback when reservation can't be used. Although now that we are defer the unlock instead of explicitly unlock, writing only when checks are successful seems okay

api/clients/v2/disperser_client.go
hopeyen: Accountant should be doing this. This is tricky if we start to allow concurrent on-demand requests because cumulative payments are monotonically increasing. Need to figure out that design first before we implement the rollback

api/clients/v2/accountant.go
hopeyen: yep, moved in [commit](https://github.com/Layr-Labs/eigenda/pull/1653/commits/82aa74e9eafd106e232a79fde3aa4b4f0231a499)

api/clients/v2/accountant.go
hopeyen: added a lock to on-demand in [commit](https://github.com/Layr-Labs/eigenda/pull/1653/commits/82aa74e9eafd106e232a79fde3aa4b4f0231a499)

api/clients/v2/accountant.go
hopeyen: Got rid of it ([commit](https://github.com/Layr-Labs/eigenda/pull/1653/commits/82aa74e9eafd106e232a79fde3aa4b4f0231a499))

api/clients/v2/accountant.go
hopeyen: update the name and added comment ([commit](https://github.com/Layr-Labs/eigenda/pull/1653/commits/82aa74e9eafd106e232a79fde3aa4b4f0231a499))

api/clients/v2/accountant.go
hopeyen: Yep! with the refactoring, this logic can live at the top of the functions ([commit](https://github.com/Layr-Labs/eigenda/pull/1653/commits/82aa74e9eafd106e232a79fde3aa4b4f0231a499))

api/clients/v2/accountant.go
hopeyen: yes! updated in [commit](https://github.com/Layr-Labs/eigenda/pull/1653/commits/82aa74e9eafd106e232a79fde3aa4b4f0231a499)

api/clients/v2/accountant.go
```diff
@@ -67,55 +70,77 @@ func NewAccountant(accountID gethcommon.Address, reservation *core.ReservedPayme
 // indicating on-demand payment.
 // These generated values are used to create the payment header and signature, as specified in
 // api/proto/common/v2/common_v2.proto
-func (a *Accountant) blobPaymentInfo(
-	numSymbols uint64,
+// func (a *Accountant) blobPaymentInfo(
```
dmanc: remove commented stuff if no longer needed

api/clients/v2/accountant.go
```diff
@@ -67,55 +70,77 @@ func NewAccountant(accountID gethcommon.Address, reservation *core.ReservedPayme
...
+	binLimit := meterer.GetReservationBinLimit(a.reservation, a.reservationWindow)
 	if relativePeriodRecord.Usage <= binLimit {
-		if err := QuorumCheck(quorumNumbers, a.reservation.QuorumNumbers); err != nil {
-			return big.NewInt(0), err
-		}
-		return big.NewInt(0), nil
+		return nil
 	}
 
 	overflowPeriodRecord := a.getOrRefreshRelativePeriodRecord(currentReservationPeriod+2*reservationWindow, reservationWindow)
```
dmanc: Sorry to add more scope to the PR but I feel the math here `currentReservationPeriod+2*reservationWindow` should be documented somewhere, a comment explaining the logic should be sufficient. Also this `2 * reservationWindow` should be some variable name. Maybe `currentReservationPeriod +  overflowPeriodOffset`?

api/clients/v2/accountant.go
```diff
@@ -124,9 +149,22 @@ func (a *Accountant) AccountBlob(
...
+			Timestamp:         timestamp,
+			CumulativePayment: big.NewInt(0),
+		}, nil
+	}
+
+	// Fall back to on-demand payment if reservation fails
+	cumulativePayment, err := a.onDemandUsage(symbolUsage, quorums)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("cannot create payment infomation for reservation or on-demand. Consider depositing more eth to the PaymentVault contract for your account. For more details, see https://docs.eigenda.xyz/core-concepts/payments#disperser-client-requirements. Account: %s, Error: %w", a.accountID.Hex(), err)
```
dmanc:
```suggestion
		return nil, fmt.Errorf("cannot create payment information for reservation or on-demand. Consider depositing more eth to the PaymentVault contract for your account. For more details, see https://docs.eigenda.xyz/core-concepts/payments#disperser-client-requirements. Account: %s, Error: %w", a.accountID.Hex(), err)
```

dmanc: There's also a typo in another function but I can't leave a comment. Ctrl+f `accoutant`.

api/clients/v2/accountant.go
hopeyen: nice! I found the same bug in meterer as in the https://github.com/Layr-Labs/eigenda/pull/1641 when I refactored this, now they should both be using the same logic to find the overflow period. Also added a check from the accountant to verify that reservation is active

api/clients/v2/disperser_client.go
samlaf: If the rollback should happen in the accountant then I don't think disperser_client should have this TODO right? It's a bit confusing.

samlaf (approved):
LGTM although I do think https://github.com/Layr-Labs/eigenda/pull/1653#discussion_r2140680160  should be changed

PR #1661, opened 2025-06-13
refactor: core meterer period record module

hopeyen:

Refactoring period record and circular wrapping logic from client accountant to core meterer. The same logic will be applied to [validator for multi-account metering](https://github.com/Layr-Labs/eigenda/pull/1599)

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/meterer/meterer.go
+++ b/core/meterer/meterer.go
@@ -284,7 +284,10 @@ func GetOverflowPeriod(reservationPeriod uint64, reservationWindow uint64) uint6
 
 // PaymentCharged returns the chargeable price for a given number of symbols
 func PaymentCharged(numSymbols, pricePerSymbol uint64) *big.Int {
-	return new(big.Int).Mul(big.NewInt(int64(numSymbols)), big.NewInt(int64(pricePerSymbol)))
+	// directly convert to uint64 to avoid overflow
+	numSymbolsInt := new(big.Int).SetUint64(numSymbols)
+	pricePerSymbolInt := new(big.Int).SetUint64(pricePerSymbol)
+	return new(big.Int).Mul(numSymbolsInt, pricePerSymbolInt)
 }
 
 // SymbolsCharged returns the number of symbols charged for a given data length
@@ -292,6 +295,9 @@ func PaymentCharged(numSymbols, pricePerSymbol uint64) *big.Int {
 func SymbolsCharged(numSymbols uint64, minSymbols uint64) uint64 {
 	if numSymbols <= minSymbols {
 		return minSymbols
+	}
+	if minSymbols == 0 {
+		return numSymbols
 	}
 	// Round up to the nearest multiple of MinNumSymbols
 	roundedUp := core.RoundUpDivide(numSymbols, minSymbols) * minSymbols
--- a/core/meterer/period_record.go
+++ b/core/meterer/period_record.go
@@ -1,0 +1,124 @@
+package meterer
+
+import (
+	"fmt"
+
+	disperser_rpc "github.com/Layr-Labs/eigenda/api/grpc/disperser/v2"
+	"github.com/Layr-Labs/eigenda/core"
+)
+
+type QuorumPeriodRecords map[core.QuorumID][]*PeriodRecord
+
+// PeriodRecord contains the index of the reservation period and the usage of the period
+type PeriodRecord struct {
+	// Index is start timestamp of the period in seconds; it is always a multiple of the reservation window
+	Index uint32
+	// Usage is the usage of the period in symbols
+	Usage uint64
+}
+
+// updateRecord tracks a successful update for rollback purposes
+type UpdateRecord struct {
+	QuorumNumber core.QuorumID
+	Period       uint64
+	Usage        uint64
+}
+
+// GetRelativePeriodRecord returns the period record for the given index and quorum number; if the record does not exist, it is initialized to 0
+func (pr QuorumPeriodRecords) GetRelativePeriodRecord(index uint64, quorumNumber core.QuorumID) *PeriodRecord {
+	if _, exists := pr[quorumNumber]; !exists {
+		pr[quorumNumber] = make([]*PeriodRecord, MinNumBins)
+	}
+	relativeIndex := uint32(index % uint64(MinNumBins))
+	if pr[quorumNumber][relativeIndex] == nil {
+		pr[quorumNumber][relativeIndex] = &PeriodRecord{
+			Index: uint32(index),
+			Usage: 0,
+		}
+	}
+	return pr[quorumNumber][relativeIndex]
+}
+
+// UpdateUsage attempts to update the usage for a quorum's reservation period
+// Returns error if the update would exceed the bin limit and cannot use overflow bin
+func (pr QuorumPeriodRecords) UpdateUsage(
+	quorumNumber core.QuorumID,
+	timestamp int64,
+	numSymbols uint64,
+	reservation *core.ReservedPayment,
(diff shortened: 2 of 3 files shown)


core/meterer/period_record.go
```diff
@@ -0,0 +1,124 @@
...
+	timestamp int64,
+	numSymbols uint64,
+	reservation *core.ReservedPayment,
+	protocolConfig *core.PaymentQuorumProtocolConfig,
+) error {
+	symbolUsage := SymbolsCharged(numSymbols, protocolConfig.MinNumSymbols)
+	binLimit := GetReservationBinLimit(reservation, protocolConfig.ReservationRateLimitWindow)
+
+	if symbolUsage > binLimit {
+		return fmt.Errorf("symbol usage exceeds bin limit")
```
dmanc: nit: use errors.new instead

core/meterer/period_record.go
```diff
@@ -0,0 +1,124 @@
...
+	return pr[quorumNumber][relativeIndex]
+}
+
+// UpdateUsage attempts to update the usage for a quorum's reservation period
+// Returns error if the update would exceed the bin limit and cannot use overflow bin
+func (pr QuorumPeriodRecords) UpdateUsage(
+	quorumNumber core.QuorumID,
+	timestamp int64,
+	numSymbols uint64,
+	reservation *core.ReservedPayment,
```
dmanc, line 64: should we add nill check for reservation and protocolConfig?

core/meterer/period_record.go
```diff
@@ -0,0 +1,124 @@
...
+		pr[quorumNumber][relativeIndex] = &PeriodRecord{
+			Index: uint32(index),
+			Usage: 0,
+		}
+	}
+	return pr[quorumNumber][relativeIndex]
+}
+
+// UpdateUsage attempts to update the usage for a quorum's reservation period
+// Returns error if the update would exceed the bin limit and cannot use overflow bin
```
dmanc, line 45: Would this be a good place to explain how the circular buffer thing works?

dmanc (dismissed):
lgtm, some minor comments

core/meterer/period_record.go
hopeyen: I added a paragraph and a link https://github.com/Layr-Labs/eigenda/pull/1661/commits/0843546ea033f08207ac4c8f0383addd41b476ce

core/meterer/period_record.go
```diff
@@ -0,0 +1,135 @@
...
+	"errors"
+	"fmt"
+
+	disperser_rpc "github.com/Layr-Labs/eigenda/api/grpc/disperser/v2"
+	"github.com/Layr-Labs/eigenda/core"
+)
+
+// QuorumPeriodRecords is a map of quorum number to a slice of period records
+// Detailed explanation on circular buffer: https://www.notion.so/eigen-labs/Accountant-Circular-Wrapping-20113c11c3e08099893bf9a5daadda4e
+// TL;DR: The algorithm maintains a fixed-size circular buffer of numBins slots to track usage across an unbounded sequence of time periods by mapping each “absolute” period index onto a “relative” buffer index via modular arithmetic. Incoming timestamps are first bucketed into discrete reservation periods of length reservationWindow, yielding an integer period index. When a request for period p arrives, the system computes its buffer slot i = p mod numBins; if the stored period at slot i differs from p, the slot is reset (index updated, usage cleared) before accumulating usage. Controlled overflow allows unused capacity in a full bin to spill into future bins under strict conditions, and a sliding valid-period window ensures only recent periods are accepted.
```
dmanc:
```suggestion
// TL;DR: The algorithm maintains a fixed-size circular buffer of numBins slots 
// to track usage across an unbounded sequence of time periods by mapping each 
// "absolute" period index onto a "relative" buffer index via modular arithmetic. 
//
// Incoming timestamps are first bucketed into discrete reservation periods of 
// length reservationWindow, yielding an integer period index. When a request 
// for period p arrives, the system computes its buffer slot i = p mod numBins; 
// if the stored period at slot i differs from p, the slot is reset (index 
// updated, usage cleared) before accumulating usage. 
//
// Controlled overflow allows unused capacity in a full bin to spill into future 
// bins under strict conditions, and a sliding valid-period window ensures only 
// recent periods are accepted.
```

dmanc: Also let's remove the internal docs reference

core/meterer/meterer.go
```diff
@@ -284,7 +284,10 @@ func GetOverflowPeriod(reservationPeriod uint64, reservationWindow uint64) uint6
 
 // PaymentCharged returns the chargeable price for a given number of symbols
 func PaymentCharged(numSymbols, pricePerSymbol uint64) *big.Int {
-	return new(big.Int).Mul(big.NewInt(int64(numSymbols)), big.NewInt(int64(pricePerSymbol)))
+	// directly convert to uint64 to avoid overflow
+	numSymbolsInt := new(big.Int).SetUint64(numSymbols)
+	pricePerSymbolInt := new(big.Int).SetUint64(pricePerSymbol)
+	return new(big.Int).Mul(numSymbolsInt, pricePerSymbolInt)
```
anupsv, line 290: Is there a overflow here incase either numSymbolsInt and pricePerSymbolInt are hitting the max uint64 ? Do we check if the limits are being checked somewhere before hitting this endpoint ?

core/meterer/period_record.go
```diff
@@ -0,0 +1,135 @@
...
+import (
+	"errors"
+	"fmt"
+
+	disperser_rpc "github.com/Layr-Labs/eigenda/api/grpc/disperser/v2"
+	"github.com/Layr-Labs/eigenda/core"
+)
+
+// QuorumPeriodRecords is a map of quorum number to a slice of period records
+// Detailed explanation on circular buffer: https://www.notion.so/eigen-labs/Accountant-Circular-Wrapping-20113c11c3e08099893bf9a5daadda4e
```
anupsv: We shouldn't change this to github docs and add it there. Notion doc isn't good to be linked.

core/meterer/meterer.go
hopeyen: I don't think there would be overflow. `big.Int` should be able to handle arbitrary-precision integers
We don't have limit check. these two number could be anything Uint64 in theory

core/meterer/period_record.go
```diff
@@ -0,0 +1,147 @@
...
+// for period p arrives, the system computes its buffer slot i = p mod numBins;
+// if the stored period at slot i differs from p, the slot is reset (index
+// updated, usage cleared) before accumulating usage.
+//
+// Controlled overflow allows unused capacity in a full bin to spill into future
+// bins under strict conditions, and a sliding valid-period window ensures only
+// recent periods are accepted.
+func (pr QuorumPeriodRecords) UpdateUsage(
+	quorumNumber core.QuorumID,
+	timestamp int64,
```
anupsv, line 62: question on timestamp, should we check if this provided timestamp is within some certain period of time? I don't see the input source for this.

hopeyen: Yeah, the disperser meterer checks it [here](https://github.com/Layr-Labs/eigenda/blob/2c9179265180b72ece783295ad4a471a74406833/core/meterer/meterer.go#L332), and the batch meterer would do this check, but the client accountant doesn't have this check, thinking that accountant doesn't need to do this type of validation.

hopeyen: there's no input source atm, will be used in one of the subsequent PRs

dmanc (approved):

anupsv (approved):

PR #1662, opened 2025-06-13
refactor: onchain state interface

hopeyen:

- Simplifies the `OnchainPayment` interface (16 methods -> 5 methods)
- Single call + structured access that provides atomic access to related configuration and makes error handling more explicit and consistent

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/meterer/meterer.go
+++ b/core/meterer/meterer.go
@@ -83,12 +83,13 @@ func (m *Meterer) MeterRequest(ctx context.Context, header core.PaymentMetadata,
 		return 0, fmt.Errorf("failed to get payment global params: %w", err)
 	}
 
-	protocolConfig, err := params.GetQuorumProtocolConfig(OnDemandQuorumID)
+	// TODO(hopeyen): symbols calculations are moved to be quorum-specific in the next PR
+	_, protocolConfig, err := params.GetQuorumConfigs(OnDemandQuorumID)
 	if err != nil {
 		return 0, fmt.Errorf("failed to get protocol config for on-demand quorum: %w", err)
 	}
-
 	symbolsCharged := SymbolsCharged(numSymbols, protocolConfig.MinNumSymbols)
+
 	// Validate against the payment method
 	if !IsOnDemandPayment(&header) {
 		reservations, err := m.ChainPaymentState.GetReservedPaymentByAccountAndQuorums(ctx, header.AccountID, quorumNumbers)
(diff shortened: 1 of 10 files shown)


core/meterer/meterer.go
```diff
@@ -240,13 +268,28 @@ func (m *Meterer) ServeOnDemandRequest(ctx context.Context, header core.PaymentM
...
+	paymentConfig, err := params.GetQuorumPaymentConfig(OnDemandQuorumID)
+	if err != nil {
+		return fmt.Errorf("failed to get payment config for on-demand quorum: %w", err)
+	}
+
+	protocolConfig, err := params.GetQuorumProtocolConfig(OnDemandQuorumID)
+	if err != nil {
+		return fmt.Errorf("failed to get protocol config for on-demand quorum: %w", err)
+	}
+
```
anupsv, line 266: Are these cached or something ? I'm wondering if these will cause like a good 3-5 seconds wait time when actually polling each of them. Maybe we can batch them together or something cause we fail even if one of them isn't working.

hopeyen: These are cached and refreshed periodically by payment's on-chain state: https://github.com/Layr-Labs/eigenda/blob/7fcddcc7604bf7d17cd20314008bc763a0f68932/core/meterer/onchain_state.go#L90
when the functions are reading the configs, they shouldn't be polling from the chain directly

core/meterer/meterer.go
```diff
@@ -239,14 +257,24 @@ func (m *Meterer) ServeOnDemandRequest(ctx context.Context, header core.PaymentM
...
+	globalPeriod := GetReservationPeriod(receivedAt.Unix(), protocolConfig.OnDemandRateLimitWindow)
 
 	newUsage, err := m.MeteringStore.UpdateGlobalBin(ctx, globalPeriod, symbolsCharged)
 	if err != nil {
 		return fmt.Errorf("failed to increment global bin usage: %w", err)
 	}
-	if newUsage > m.ChainPaymentState.GetOnDemandGlobalSymbolsPerSecond(OnDemandQuorumID)*uint64(m.ChainPaymentState.GetOnDemandGlobalRatePeriodInterval(OnDemandQuorumID)) {
+	if newUsage > paymentConfig.OnDemandSymbolsPerSecond*protocolConfig.OnDemandRateLimitWindow {
```
dmanc: Can we make this a function similar to `GetReservationBinLimit`?

dmanc (approved):

core/meterer/meterer.go
hopeyen: yes! updated in [commit](https://github.com/Layr-Labs/eigenda/pull/1662/commits/fab71d50edf211e5bbba0fc12520843a02dcf040)

anupsv (approved):

PR #1663, opened 2025-06-13
refactor: core meterer consolidation

hopeyen:

- Refactored `core/meterer` component to consolidate quorum-specific payment configurations
- Updated meterer tests and some other tests

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/meterer/meterer.go
+++ b/core/meterer/meterer.go
@@ -2,6 +2,7 @@ package meterer
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"math"
 	"math/big"
@@ -113,7 +114,15 @@ func (m *Meterer) MeterRequest(ctx context.Context, header core.PaymentMetadata,
 }
 
 // serveReservationRequest handles the rate limiting logic for incoming requests
-func (m *Meterer) serveReservationRequest(ctx context.Context, globalParams *PaymentVaultParams, header core.PaymentMetadata, reservations map[core.QuorumID]*core.ReservedPayment, numSymbols uint64, quorumNumbers []uint8, receivedAt time.Time) error {
+func (m *Meterer) serveReservationRequest(
+	ctx context.Context,
+	globalParams *PaymentVaultParams,
+	header core.PaymentMetadata,
+	reservations map[core.QuorumID]*core.ReservedPayment,
+	numSymbols uint64,
+	quorumNumbers []uint8,
+	receivedAt time.Time,
+) error {
 	m.logger.Debug("Recording and validating reservation usage", "header", header, "reservation", reservations)
 	if err := ValidateReservations(reservations, globalParams.QuorumProtocolConfigs, quorumNumbers, header.Timestamp, receivedAt); err != nil {
 		return fmt.Errorf("invalid reservation: %w", err)
@@ -127,7 +136,12 @@ func (m *Meterer) serveReservationRequest(ctx context.Context, globalParams *Pay
 }
 
 // incrementBinUsage increments the bin usage atomically and checks for overflow
-func (m *Meterer) incrementBinUsage(ctx context.Context, header core.PaymentMetadata, reservations map[core.QuorumID]*core.ReservedPayment, globalParams *PaymentVaultParams, numSymbols uint64) error {
+func (m *Meterer) incrementBinUsage(
+	ctx context.Context, header core.PaymentMetadata,
+	reservations map[core.QuorumID]*core.ReservedPayment,
+	globalParams *PaymentVaultParams,
+	numSymbols uint64,
+) error {
 	charges := make(map[core.QuorumID]uint64)
 	quorumNumbers := make([]core.QuorumID, 0, len(reservations))
 	reservationWindows := make(map[core.QuorumID]uint64, len(reservations))
@@ -344,12 +358,22 @@ func ValidateQuorum(headerQuorums []uint8, allowedQuorums []uint8) error {
 	return nil
 }
 
-// ValidateReservations ensures that the quorums listed in the blobHeader are present within allowedQuorums
-// timestamp is in nanoseconds
-// Note: A reservation that does not utilize all of the allowed quorums will be accepted. However, it
-// will still charge against all of the allowed quorums. A on-demand requests require and only allow
-// the ETH and EIGEN quorums.
-func ValidateReservations(reservations map[core.QuorumID]*core.ReservedPayment, quorumConfigs map[core.QuorumID]*core.PaymentQuorumProtocolConfig, quorumNumbers []uint8, timestamp int64, receivedAt time.Time) error {
+// ValidateReservations ensures that the quorums listed in the blobHeader are present within allowedQuorums.
+//
+// Parameters:
+//   - timestamp: time in nanoseconds
+//
+// Notes:
+//   - Reservations that don't use all allowed quorums are still accepted
+//   - Charges apply to ALL allowed quorums, even if not all are used
+//   - On-demand requests have special requirements: they must use ETH and EIGEN quorums only
+func ValidateReservations(
+	reservations map[core.QuorumID]*core.ReservedPayment,
+	quorumConfigs map[core.QuorumID]*core.PaymentQuorumProtocolConfig,
+	quorumNumbers []uint8,
+	timestamp int64,
+	receivedAt time.Time,
+) error {
 	reservationQuorums := make([]uint8, 0, len(reservations))
 	reservationWindows := make(map[core.QuorumID]uint64, len(reservations))
 	requestReservationPeriods := make(map[core.QuorumID]uint64, len(reservations))
@@ -371,7 +395,7 @@ func ValidateReservations(reservations map[core.QuorumID]*core.ReservedPayment, 
 	for _, quorumID := range quorumNumbers {
 		reservation := reservations[core.QuorumID(quorumID)]
 		if !reservation.IsActiveByNanosecond(timestamp) {
-			return fmt.Errorf("reservation not active")
+			return errors.New("reservation not active")
 		}
 		if !ValidateReservationPeriod(reservation, requestReservationPeriods[quorumID], reservationWindows[quorumID], receivedAt) {
 			return fmt.Errorf("invalid reservation period for reservation on quorum %d", quorumID)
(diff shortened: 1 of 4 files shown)


core/meterer/meterer.go
```diff
@@ -82,87 +82,66 @@ func (m *Meterer) MeterRequest(ctx context.Context, header core.PaymentMetadata,
...
 			return 0, fmt.Errorf("failed to get on-demand payment by account: %w", err)
 		}
-		if err := m.ServeOnDemandRequest(ctx, params, header, onDemandPayment, symbolsCharged, quorumNumbers, receivedAt); err != nil {
+		if err := m.serveOnDemandRequest(ctx, params, header, onDemandPayment, numSymbols, quorumNumbers, receivedAt); err != nil {
 			return 0, fmt.Errorf("invalid on-demand request: %w", err)
 		}
 	}
 
+	// TODO(hopeyen): each quorum can have different min num symbols; the returned symbolsCharged is only for used for metrics.
+	// for now we simply return the charge for quorum 0, as quorums are likely to share the same min num symbols
```
anupsv, line 106: Do we allow the configuration onchain to state minimum number of symbols or so ?

anupsv: Also what does "charge for quorum 0" mean here? Is it the number of symbols for quorum 0?

core/meterer/meterer.go
```diff
@@ -190,39 +169,47 @@ func (m *Meterer) IncrementBinUsage(ctx context.Context, header core.PaymentMeta
...
 		}
 	}
-
-	// Batch increment overflow bins for overflown reservation candidates
-	for quorumID := range overflowCandidates {
-		_, err := m.MeteringStore.IncrementBinUsages(ctx, header.AccountID, []core.QuorumID{quorumID}, map[core.QuorumID]uint64{quorumID: requestReservationPeriods[quorumID] + 2}, map[core.QuorumID]uint64{quorumID: overflowAmounts[quorumID]})
+	if len(overflowAmounts) != len(overflowPeriods) {
+		return fmt.Errorf("overflow amount and period mismatch")
+	}
+	// Batch increment overflow bins for all overflown reservation candidates
```
anupsv, line 199: Thinking of a case, is it possible at all that 2 batches received by the validator from multiple dispersers and say the same client might have sent the same data to these 2 dispersers, would this cause a problem ?

core/meterer/meterer.go
```diff
@@ -82,87 +82,66 @@ func (m *Meterer) MeterRequest(ctx context.Context, header core.PaymentMetadata,
...
-	// Validate the used reservations are active and is of valid periods
-	for quorumID, reservation := range reservations {
-		if !reservation.IsActiveByNanosecond(header.Timestamp) {
-			return fmt.Errorf("reservation not active")
-		}
-		if !ValidateReservationPeriod(reservation, requestReservationPeriods[quorumID], reservationWindows[quorumID], receivedAt) {
-			return fmt.Errorf("invalid reservation period for reservation on quorum %d", quorumID)
-		}
+// serveReservationRequest handles the rate limiting logic for incoming requests
+func (m *Meterer) serveReservationRequest(ctx context.Context, globalParams *PaymentVaultParams, header core.PaymentMetadata, reservations map[core.QuorumID]*core.ReservedPayment, numSymbols uint64, quorumNumbers []uint8, receivedAt time.Time) error {
```
dmanc: can we wrap to multiple lines for readability
```suggestion
func (m *Meterer) serveReservationRequest(
    ctx context.Context,
    globalParams *PaymentVaultParams,
    header core.PaymentMetadata,
    reservations map[core.QuorumID]*core.ReservedPayment,
    numSymbols uint64,
    quorumNumbers []uint8,
    receivedAt time.Time,
) error {
```

core/meterer/meterer.go
```diff
@@ -356,6 +344,43 @@ func ValidateQuorum(headerQuorums []uint8, allowedQuorums []uint8) error {
 	return nil
 }
 
+// ValidateReservations ensures that the quorums listed in the blobHeader are present within allowedQuorums
+// timestamp is in nanoseconds
+// Note: A reservation that does not utilize all of the allowed quorums will be accepted. However, it
+// will still charge against all of the allowed quorums. A on-demand requests require and only allow
```
anupsv: Can u explain a bit more ? If someone has reservations for 2 quorums and they utilize just 1 of them in one of the dispersals, then we charge them for both ?

core/meterer/meterer.go
```diff
@@ -82,87 +82,66 @@ func (m *Meterer) MeterRequest(ctx context.Context, header core.PaymentMetadata,
...
+	if err := m.incrementBinUsage(ctx, header, reservations, globalParams, numSymbols); err != nil {
 		return fmt.Errorf("failed to increment bin usages: %w", err)
 	}
 	return nil
 }
 
-// IncrementBinUsage increments the bin usage atomically and checks for overflow
-func (m *Meterer) IncrementBinUsage(ctx context.Context, header core.PaymentMetadata, reservations map[core.QuorumID]*core.ReservedPayment, symbolsCharged uint64, reservationWindows map[core.QuorumID]uint64, requestReservationPeriods map[core.QuorumID]uint64) error {
+// incrementBinUsage increments the bin usage atomically and checks for overflow
+func (m *Meterer) incrementBinUsage(ctx context.Context, header core.PaymentMetadata, reservations map[core.QuorumID]*core.ReservedPayment, globalParams *PaymentVaultParams, numSymbols uint64) error {
```
dmanc: same comment, multiple lines for readability
```suggestion
func (m *Meterer) incrementBinUsage(
    ctx context.Context, header core.PaymentMetadata, 
    reservations map[core.QuorumID]*core.ReservedPayment, 
    globalParams *PaymentVaultParams, 
    numSymbols uint64
) error {
```

core/meterer/meterer.go
hopeyen: yeah, this configuration can be different per quorum. The new usageAuthorizationRegistry support these configs per quorum
```
	MinNumSymbols              uint64
	ReservationAdvanceWindow   uint64
	ReservationRateLimitWindow uint64

	// OnDemand is initially only enabled on Quorum 0
	OnDemandRateLimitWindow uint64
	OnDemandEnabled         bool
```
and managed by the contract owner. 
these are configs are a separate struct 
```
	ReservationSymbolsPerSecond uint64

	// OnDemand is initially only enabled on Quorum 0
	OnDemandSymbolsPerSecond uint64
	OnDemandPricePerSymbol   uint64
```
and are managed by quorum owner

core/meterer/meterer.go
```diff
@@ -356,6 +344,43 @@ func ValidateQuorum(headerQuorums []uint8, allowedQuorums []uint8) error {
 	return nil
 }
 
+// ValidateReservations ensures that the quorums listed in the blobHeader are present within allowedQuorums
+// timestamp is in nanoseconds
+// Note: A reservation that does not utilize all of the allowed quorums will be accepted. However, it
+// will still charge against all of the allowed quorums. A on-demand requests require and only allow
+// the ETH and EIGEN quorums.
+func ValidateReservations(reservations map[core.QuorumID]*core.ReservedPayment, quorumConfigs map[core.QuorumID]*core.PaymentQuorumProtocolConfig, quorumNumbers []uint8, timestamp int64, receivedAt time.Time) error {
```
dmanc: Same comment:
```suggestion
func ValidateReservations(
    reservations map[core.QuorumID]*core.ReservedPayment, 
    quorumConfigs map[core.QuorumID]*core.PaymentQuorumProtocolConfig, 
    quorumNumbers []uint8, 
    timestamp int64, 
    receivedAt time.Time
) error {
```

core/meterer/meterer.go
hopeyen: `charge for quorum 0` is the charge calculated based on quorum 0's configurations, and the quorum that on-demand would be initially enabled at; on-demand users can still use both quorum 0 and 1 by depositing to quorum 0, just like they do now

core/meterer/meterer.go
hopeyen: this meterer is on the disperser level, so for the disperser, this wouldn't be a problem, and the disperser won't rollback if validator complains then. 
The validators will run a different component (batchMeterer on a different PR) and would still accept requests from the dispersers, but fail the ones that exceed the ratelimit

core/meterer/meterer.go
```diff
@@ -356,6 +344,43 @@ func ValidateQuorum(headerQuorums []uint8, allowedQuorums []uint8) error {
 	return nil
 }
 
+// ValidateReservations ensures that the quorums listed in the blobHeader are present within allowedQuorums
```
dmanc: Maybe something like this instead is more clear
```suggestion
// ValidateReservations ensures that the quorums listed in the blobHeader are present within allowedQuorums.
// 
// Parameters:
//   - timestamp: time in nanoseconds
// 
// Notes:
//   - Reservations that don't use all allowed quorums are still accepted
//   - Charges apply to ALL allowed quorums, even if not all are used
//   - On-demand requests have special requirements: they must use ETH and EIGEN quorums only
```

core/meterer/meterer.go
hopeyen: yeah, this is the current state for the meterer because there's no quorum distinction on the contract yet. With the contracts update, we will only charge the one reservation on quorums that were indicated by the request header

core/meterer/meterer.go
```diff
@@ -356,6 +344,43 @@ func ValidateQuorum(headerQuorums []uint8, allowedQuorums []uint8) error {
...
+		requestReservationPeriods[quorumID] = GetReservationPeriodByNanosecond(timestamp, quorumConfigs[quorumID].ReservationRateLimitWindow)
+	}
+	if err := ValidateQuorum(quorumNumbers, reservationQuorums); err != nil {
+		return err
+	}
+	// Validate the used reservations are active and is of valid periods
+	for _, quorumID := range quorumNumbers {
+		reservation := reservations[core.QuorumID(quorumID)]
+		if !reservation.IsActiveByNanosecond(timestamp) {
+			return fmt.Errorf("reservation not active")
```
dmanc: nit: use errors.new

hopeyen: updated for all the comments

dmanc (approved):

core/meterer/meterer.go
anupsv: discussed offline.

anupsv (approved):

PR #1664, opened 2025-06-14
feat: GetPaymentStateForAllQuorums api impl

hopeyen:

- Implemented `GetPaymentStateForAllQuorums` endpoint, updated `GetPaymentState` to be unimplemented
- refactored payment state management and accountant client interfaces
- simplified test structure and improved protobuf conversion for PaymentVaultParams

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/clients/v2/accountant.go
```diff
@@ -5,115 +5,106 @@ import (
...
 	// on-chain states
-	accountID         gethcommon.Address
-	reservation       *core.ReservedPayment
-	onDemand          *core.OnDemandPayment
-	reservationWindow uint64
-	pricePerSymbol    uint64
-	minNumSymbols     uint64
+	accountID    gethcommon.Address
+	reservations map[core.QuorumID]*core.ReservedPayment
+	onDemand     *core.OnDemandPayment
```
samlaf: why is onDemand not a per-quorum map? If its a current limitation then should be commented/documented here

api/clients/v2/accountant.go
```diff
@@ -5,115 +5,106 @@ import (
...
-	// contains an array of period records, with length of max(MinNumBins, numBins)
-	// numBins can be arbitrarily bigger than MinNumBins if the client wants to track more history in the cache
-	periodRecords     []PeriodRecord
+	// contains 3 bins; circular wrapping of indices
+	periodRecords     meterer.QuorumPeriodRecords
 	cumulativePayment *big.Int
 
 	// locks for concurrent access to period records and on-demand payment
 	periodRecordsLock sync.Mutex
 	onDemandLock      sync.Mutex
```
samlaf: is `onDemandLock` protecting `onDemand` or `periodRecords`? Or both?

api/clients/v2/accountant.go
```diff
@@ -5,115 +5,106 @@ import (
...
+	onDemand     *core.OnDemandPayment
+
+	paymentVaultParams *meterer.PaymentVaultParams
 
 	// local accounting
-	// contains an array of period records, with length of max(MinNumBins, numBins)
-	// numBins can be arbitrarily bigger than MinNumBins if the client wants to track more history in the cache
-	periodRecords     []PeriodRecord
+	// contains 3 bins; circular wrapping of indices
+	periodRecords     meterer.QuorumPeriodRecords
```
samlaf: is it really 3 bins? Nothing seems to indicate 3 anywhere

api/clients/v2/accountant.go
samlaf: also its n bins PER quorum right?

api/clients/v2/accountant.go
```diff
@@ -5,115 +5,106 @@ import (
...
-// reservationUsage attempts to use the reservation for the given request.
-func (a *Accountant) reservationUsage(
-	symbolUsage uint64,
-	quorumNumbers []uint8,
-	timestamp int64) error {
-	if err := meterer.ValidateQuorum(quorumNumbers, a.reservation.QuorumNumbers); err != nil {
+// ReservationUsage attempts to use the reservation for the requested quorums; if any quorum fails to use the reservation, the entire operation is rolled back.
+func (a *Accountant) reservationUsage(numSymbols uint64, quorumNumbers []core.QuorumID, timestamp int64) error {
+	// The two timestamps are the same for the accountant client for validating the reservation period; the second timestamp is the received at time for the server
+	if err := meterer.ValidateReservations(a.reservations, a.paymentVaultParams.QuorumProtocolConfigs, quorumNumbers, timestamp, time.Unix(0, timestamp)); err != nil {
```
samlaf: not following what this comment is saying. Also if you are using `time.Unix(0, timestamp)` then you have an assumption that the `timestamp int64` argument is in nanoseconds. So prob rename to `timestampNs`?

samlaf:
```suggestion
func (a *Accountant) reservationUsage(numSymbols uint64, quorumNumbers []core.QuorumID, timestampNs int64) error {
	// The two timestamps are the same for the accountant client for validating the reservation period; the second timestamp is the received at time for the server
	if err := meterer.ValidateReservations(a.reservations, a.paymentVaultParams.QuorumProtocolConfigs, quorumNumbers, timestampNs, time.Unix(0, timestampNs)); err != nil {
```

api/clients/v2/accountant.go
```diff
@@ -5,115 +5,106 @@ import (
...
-		overflowPeriodRecord.Usage += relativePeriodRecord.Usage - binLimit
-		return nil
+	// deep copy of periodRecords for rollback in case of errors
+	originalPeriodRecords := a.periodRecords.DeepCopy()
+
+	for _, quorumNumber := range quorumNumbers {
+		reservation, exists := a.reservations[quorumNumber]
+		if !exists {
+			// this case should never happen because ValidateReservations should have already checked this; handle it just in case
+			a.periodRecords = originalPeriodRecords
```
samlaf: Instead of working on the original periodRecords and then rolling back on every error path (super bug prone - easy to forget one while refactoring), better to instead update the copy and at the very bottom commit it (only commit in the single happy return path)
```
periodRecordsCopy := a.periodRecords.DeepCopy()
// modify periodRecordsCopy
// ....
// at the bottom of the function...
a.periodRecords = periodRecordsCopy
return nil
```

api/clients/v2/accountant.go
```diff
@@ -158,39 +153,21 @@ func (a *Accountant) AccountBlob(
...
 	if paymentState == nil {
 		return fmt.Errorf("payment state cannot be nil")
-	} else if paymentState.GetPaymentGlobalParams() == nil {
-		return fmt.Errorf("payment global params cannot be nil")
 	}
-
-	a.minNumSymbols = paymentState.GetPaymentGlobalParams().GetMinNumSymbols()
-	a.pricePerSymbol = paymentState.GetPaymentGlobalParams().GetPricePerSymbol()
-	a.reservationWindow = paymentState.GetPaymentGlobalParams().GetReservationWindow()
+	paymentVaultParams, err := meterer.PaymentVaultParamsFromProtobuf(paymentState.GetPaymentVaultParams())
```
samlaf: accountant is not exposed as a grpc service is it? If so, then it shouldn't take a protobuf struct as input. The conversion should prob be done on the outside. Accountant should work in any case, and not be tied to our rpc implementation choice (what if we want to switch to json-rpc at some point?)

disperser/apiserver/server_v2.go
```diff
@@ -279,6 +279,10 @@ func (s *DispersalServerV2) RefreshOnchainState(ctx context.Context) error {
 // GetPaymentState returns the payment state for a given account and the related on-chain parameters
 // Deprecated: Use GetPaymentStateForAllQuorums instead.
 func (s *DispersalServerV2) GetPaymentState(ctx context.Context, req *pb.GetPaymentStateRequest) (*pb.GetPaymentStateReply, error) {
+	return nil, api.NewErrorUnimplemented()
```
samlaf: why are you doing this? No point in deprecating if you're going to make it unimplemented, since then its not backwards compatible anyways. This is actually the worst of all worlds.
Why not just redirect this call to the GetPaymentStateForAllQuorums?

disperser/apiserver/server_v2.go
```diff
@@ -294,92 +298,86 @@ func (s *DispersalServerV2) GetPaymentState(ctx context.Context, req *pb.GetPaym
...
+		periodRecords[uint32(quorumId)] = &pb.PeriodRecords{
+			Records: record.Records,
 		}
 	}
 
+	// Get largest cumulative payment
 	var largestCumulativePaymentBytes []byte
 	largestCumulativePayment, err := s.meterer.MeteringStore.GetLargestCumulativePayment(ctx, accountID)
 	if err != nil {
 		s.logger.Debug("failed to get largest cumulative payment, use zero value", "err", err, "accountID", accountID)
```
samlaf: why debug? is this not an important error that should be a logger.Error level? Also is returning an empty largestCumulativePaymentBytes really what we want to do? Or do we want to return a 500 to the client?

hopeyen: the client might not be using onDemand, so we don't want to return a 500. 
update the log level in [commit](https://github.com/Layr-Labs/eigenda/pull/1664/commits/783ec1c7184a942bdfde1ef6ad4ffc611d96c776)

api/clients/v2/accountant.go
hopeyen: updated in [commit](https://github.com/Layr-Labs/eigenda/pull/1664/commits/783ec1c7184a942bdfde1ef6ad4ffc611d96c776)

api/clients/v2/accountant.go
hopeyen: updated in [commit](https://github.com/Layr-Labs/eigenda/pull/1664/commits/783ec1c7184a942bdfde1ef6ad4ffc611d96c776)

api/clients/v2/accountant.go
hopeyen: this is refactored into https://github.com/Layr-Labs/eigenda/blob/master/core/meterer/period_record.go#L32
updated comment in [commit](https://github.com/Layr-Labs/eigenda/pull/1664/commits/783ec1c7184a942bdfde1ef6ad4ffc611d96c776)

api/clients/v2/accountant.go
hopeyen: `onDemandLock` is for protecting `onDemand`. `periodRecordsLock` is for `periodRecords`. 
Per quorum is only for periodRecords (reservations). It will lock all the quorums' period records instead of one lock per `periodRecords`. updated comment in [commit](https://github.com/Layr-Labs/eigenda/pull/1664/commits/783ec1c7184a942bdfde1ef6ad4ffc611d96c776)

api/clients/v2/accountant.go
hopeyen: okay. onDemand is going to be only enabled on quorum 0. I was thinking we can update after we decide to support custom onDemand and design decentralized ratelimiting. updated comment in [commit](https://github.com/Layr-Labs/eigenda/pull/1664/commits/783ec1c7184a942bdfde1ef6ad4ffc611d96c776)

disperser/apiserver/server_v2.go
hopeyen: I wanted to chop this api off:( 
the reply types between these two APIs are really different; I've reverted the changes here instead of redirecting. If we do redirect, it will use similar logic for converting the  GetPaymentStateForAllQuorumsReply type to GetPaymentStateReply
[commit](https://github.com/Layr-Labs/eigenda/pull/1664/commits/b615cfc2af4dfbee5bf0f6419aefa7072e4cf8e8)

api/clients/v2/accountant.go
hopeyen: I see. Disperser Client is the one that gets `SetPaymentState` replay. I'll make it convert to the native types
[commit](https://github.com/Layr-Labs/eigenda/pull/1664/commits/8141eb1a192652d9396cc9622b46574caa534dc4)

api/clients/v2/accountant.go
```diff
@@ -5,115 +5,105 @@ import (
...
 	// on-chain states
-	accountID         gethcommon.Address
-	reservation       *core.ReservedPayment
-	onDemand          *core.OnDemandPayment
-	reservationWindow uint64
-	pricePerSymbol    uint64
-	minNumSymbols     uint64
+	accountID    gethcommon.Address
+	reservations map[core.QuorumID]*core.ReservedPayment
+	// OnDemand initially only enabled on quorum. Accountant must be updated to be quorum specific
```
dmanc:
```suggestion
	// OnDemand is initially only enabled on quorum 0. Accountant must be updated to be quorum specific
```
Is this what was meant?

api/clients/v2/accountant.go
```diff
@@ -5,115 +5,105 @@ import (
...
 
-// reservationUsage attempts to use the reservation for the given request.
-func (a *Accountant) reservationUsage(
-	symbolUsage uint64,
-	quorumNumbers []uint8,
-	timestamp int64) error {
-	if err := meterer.ValidateQuorum(quorumNumbers, a.reservation.QuorumNumbers); err != nil {
+// ReservationUsage attempts to use the reservation for the requested quorums; if any quorum fails to use the reservation, the entire operation is rolled back.
+func (a *Accountant) reservationUsage(numSymbols uint64, quorumNumbers []core.QuorumID, timestampNs int64) error {
+	if err := meterer.ValidateReservations(a.reservations, a.paymentVaultParams.QuorumProtocolConfigs, quorumNumbers, timestampNs, time.Now()); err != nil {
```
dmanc: Im confused about the difference between timestampNs and requestedAt = time.Not(), couldn't we use timestampNs for both?

api/clients/v2/disperser_client.go
```diff
@@ -359,7 +366,7 @@ func (c *disperserClient) GetBlobStatus(ctx context.Context, blobKey corev2.Blob
 }
 
 // GetPaymentState returns the payment state of the disperser client
-func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.GetPaymentStateReply, error) {
+func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.GetPaymentStateForAllQuorumsReply, error) {
```
dmanc, line 369: I'm not sure what the plan for deprecation of this API is but wouldn't this break existing users who are expecting a `disperser_rpc.GetPaymentStateReply` response? Is it possible to massage the response of `disperser_rpc.GetPaymentStateForAllQuorumsReply` into `disperser_rpc.GetPaymentStateReply`?

hopeyen: if the client is running an older version that expects `disperser_rpc.GetPaymentStateReply` response, then the call is made to `GetPaymentState` endpoint of the disperser, and get the correct response type and handled the same way as before. If the client is running the new version, then it makes call to the `GetPaymentStateForAllQuorums` endpoint of the disperser, and get the corresponding rely type. Provided that the disperser server would be upgraded ahead of the client upgrade, I don't expect there to be a case of reply type being different, pls correct me if i'm wrong

api/clients/v2/accountant.go
hopeyen: `timestampNs` is int64 while `time.Now()` is time, they are different types. The reason we don't directly convert timestampNs to time is because these two values are validated against each other, that `Now` must be some range within `timestampNs` which would be written into the payment header

samlaf: I'm guessing @dmanc 's question is more about "what is timestampNs?". As in where does it come from? I'm guessing its the client set timestamp in the blobHeader? Mayube call it `clientTimestampNs` or something? or `blobHeaderTimestampNs`?

> The reason we don't directly convert timestampNs to time is because these two values are validated against each other, that Now must be some range within timestampNs which would be written into the payment header

Not following what this means. I also intuitively feel like both should ideally be of the same type.

api/clients/v2/disperser_client.go
samlaf: This diff is probably a mistake then? GetPaymentState should still be returning a `GetPaymentStateReply`, and not a `GetPaymentStateForAllQuorumsReply`?

api/clients/v2/accountant.go
```diff
@@ -17,18 +16,21 @@ type Accountant struct {
 	// on-chain states
 	accountID    gethcommon.Address
 	reservations map[core.QuorumID]*core.ReservedPayment
-	onDemand     *core.OnDemandPayment
+	// OnDemand initially only enabled on quorum. Accountant must be updated to be quorum specific
```
samlaf: "on quorum" -> missing quorums 0 and 1?

disperser/apiserver/server_v2.go
```diff
@@ -277,8 +277,109 @@ func (s *DispersalServerV2) RefreshOnchainState(ctx context.Context) error {
 }
 
 // GetPaymentState returns the payment state for a given account and the related on-chain parameters
-// Deprecated: Use GetPaymentStateForAllQuorums instead.
+// Deprecating soon: use GetPaymentStateForAllQuorums instead.
```
samlaf: Keep the `Deprecated: ` syntax because IDEs pick up on that. You can add additional context after it. See https://go.dev/wiki/Deprecated

api/clients/v2/accountant.go
samlaf: Oh wait really? so onDemand not supported for EIGEN quorum?

disperser/apiserver/server_v2.go
```diff
@@ -336,7 +433,7 @@ func (s *DispersalServerV2) GetPaymentStateForAllQuorums(ctx context.Context, re
 	reservationQuorumIds := []core.QuorumID{}
 	reservationCurrentPeriods := []uint64{}
 	if err != nil {
-		s.logger.Debug("failed to get onchain reservation, use zero values", "err", err, "accountID", accountID)
+		s.logger.Error("failed to get onchain reservation, use zero values", "err", err, "accountID", accountID)
```
samlaf: Sorry I wasnt sure what the logic was doing here. These should be `Warn`ings, not errors if we just use a default value when it fails and keep going. An error is when an error is returned immediately.

disperser/apiserver/server_v2.go
samlaf: I see. So the logic is "try reservation. If it either doesn't exist or we fail to read its state from onchain, then try ondemand. If that fails then we return an error to client" right?

Given that this API is called only once at client initialization, shouldn't the client want to make darn sure that the returned values are the real ones? Aka even if they are not using onDemand, if we fail to read the value from onchain... how do you know that they aren't using onDemand??

Also I'm not super sure how this endpoint is used to populate the client and how things are kept consistent, but is it possible that if we have intermittent network connectivity issues the client might think that it doesn't have a reservation and then start draining the user's onDemand pool of money? This is probably very bad right?

Given that this API is called only once and is the only soure of truth for client-server synchronization, its error paths should be very well documented. Its hard for me to know right now when these errors happen because I don't have a full mental model of the implementation (and I shouldn't need to). Are we trying to read onchain everytime? Or are we reading from a cache? when can we actually fail to get the largest cumulative payment? etc

disperser/apiserver/server_v2.go
samlaf: Why did you want to chop this API off? This is a very very bad habit. We need to be careful not to break out clients. Imagine a rollup using an older version of proxy, at some random point when we deploy the new disperser with this new endpoint it'll just start breaking for a random reason. That's very bad.

api/clients/v2/accountant.go
samlaf: Just read in your other comment (https://github.com/Layr-Labs/eigenda/pull/1664#discussion_r2161021242) that its only on quorum 0. I'm curious as to why. Is that documented in our docs site? cc @MadelineAu

api/clients/v2/disperser_client.go
hopeyen: In this function, the call made to disperser server has been updated to `GetPaymentStateForAllQuorums` so this version of the client is no longer using `GetPaymentState` at all.

api/clients/v2/accountant.go
hopeyen: the existing payments only have one on-demand balance per account, and that balance is used for either quorum 0 or 1 offchain. 
in the new version of payments i.e. UsageAuthorizationRegistry, we are going use a new interface that allows contract owner to enable quorum specific depositing/balances for custom quorums, but initially we will still have just one balance per account to keep UI and offchain accounting the same, and we picked quorum 0 for this purpose. 
I think the docs should update for the new updates but not now

disperser/apiserver/server_v2.go
hopeyen: understood. I've put it back in the ^ commit

api/clients/v2/accountant.go
hopeyen: not quite, onDemand is supported in the way that EIGEN quorum can share the ETH quorum onDemand  balance. This logic is completely offchain. My comment above is the onchain configuration

api/clients/v2/accountant.go
hopeyen: Right, `timestampNs` is set by the client, while `time.Now()` is when the validation occurs and I can update the name
I will update both to int64 so they are consistent
[b8588e4](https://github.com/Layr-Labs/eigenda/pull/1664/commits/b8588e4d22784f2cec75ba9e65d5db2bacebb5f1)

disperser/apiserver/server_v2.go
hopeyen: I can add more docs for the error paths. 

In short, the server will return the full payment state to the client excluding other client's information, doesn't matter which payment method the client uses later on. The only time server returns error is when they cannot provide protocol wide configurations like the reservation ratelimit interval. If the server cannot retrieve client account's information, the server will return zero onDemand/reservations, so the client won't and cannot send blobs to the server. The values provided by the server could be false but it will be consistent between the server and the client

The client doesn't make payment decision based on whether a dispersal is successful or not, but based on their view of usages. So if there's connectivity issues, the client will not be switching to onDemand when they think reservation should still be sufficient. 

Another approach is server returns error if any chain read fails, instead of using zero value. Another thing I thought we could add is the client can periodically check in with the server for consistency.

api/clients/v2/disperser_client.go
dmanc: Alright convinced myself it's good

core/meterer/meterer.go
```diff
@@ -371,8 +371,8 @@ func ValidateReservations(
 	reservations map[core.QuorumID]*core.ReservedPayment,
 	quorumConfigs map[core.QuorumID]*core.PaymentQuorumProtocolConfig,
 	quorumNumbers []uint8,
-	timestamp int64,
-	receivedAt time.Time,
+	paymentHeaderTimestampNs int64,
```
dmanc, line 371: We could remove the above parameters section in the comment, doesn't seem that useful now that we mark Ns in the name

dmanc (approved):

core/meterer/meterer.go
hopeyen: remove, thank you ([250eb78](https://github.com/Layr-Labs/eigenda/pull/1664/commits/250eb78fe934117b364da4f35d710b325d55f801))

disperser/apiserver/server_v2.go
hopeyen: update [47c0368](https://github.com/Layr-Labs/eigenda/pull/1664/commits/47c0368968b84839994e92e3f753c8ccd14f90e2)

disperser/apiserver/server_v2.go
hopeyen: updated [250eb78](https://github.com/Layr-Labs/eigenda/pull/1664/commits/250eb78fe934117b364da4f35d710b325d55f801)

dmanc (dismissed):

disperser/apiserver/server_v2.go
hopeyen: changed the behavior to if chain read fails, the server will return an error 
[commit](https://github.com/Layr-Labs/eigenda/pull/1664/commits/4f0684f5fa0c64845783072493357cd04b049a95)

api/clients/v2/accountant.go
samlaf: I see but comment still feels wrong? It should say "enabled on quorum 0" no?

samlaf (approved):

api/clients/v2/disperser_client.go
samlaf: Chased a bug for a few hours on my [proxy PR](https://github.com/Layr-Labs/eigenda/actions/runs/15885533614/job/44797671219?pr=1611). Because proxy is now in monorepo, its using the client in master, which contains this new code. By only calling `GetPaymentStateForAllQuorums` (without falling back to `GetPaymentState` on Unimplemented errors) you have made the client not backwards compatible. So the client now only works against the preprod disperser and simply refuses to disperse anything on testnet/mainnet. So now we cannot cut a release of proxy until we've updated the disperser with the `GetPaymentStateForAllQuorums` on testnet/mainnet, which feels less than ideal (?).

@hopeyen just saw your comment above
> Provided that the disperser server would be upgraded ahead of the client upgrade, I don't expect there to be a case of reply type being different, pls correct me if i'm wrong

I think going forward we have to get into the habit of not merging anything in master which is not backward compatible. Because this kind of breaking change to the client would only be acceptable if we released it as a new major version, with release note indicating "this new client is currently only available with disperser of the same version". I've asked for disperser to be augmented with a semver() endpoint, but we currently don't even have this so its impossible to for clients to know whether a given client is compatible with testnet/mainnet. I think we need to be way more careful with these things going forward. This also all falls within the purview of @litt3's current push for better release processes.

hopeyen: okay, I understood. What would be the course of action here, perhaps revert this PR for the meantime, add backward compatibility using semvar when disperser server has it? 

Noticed your https://github.com/Layr-Labs/eigenda/pull/1686 and will take a look

PR #1668, opened 2025-06-18
refactor: quorum specific metrics for metered bytes

hopeyen:

make quorum number as a label, and record metered bytes charged to each quorum for every request
also added a unit test on having different quorum configurations

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/apiserver/metrics_v2.go
+++ b/disperser/apiserver/metrics_v2.go
@@ -4,9 +4,11 @@ import (
 	"context"
 	"fmt"
 	"net/http"
+	"strconv"
 	"time"
 
 	"github.com/Layr-Labs/eigenda/common"
+	"github.com/Layr-Labs/eigenda/core"
 	"github.com/Layr-Labs/eigenda/disperser"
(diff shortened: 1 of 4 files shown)


disperser/apiserver/metrics_v2.go
```diff
@@ -176,8 +178,12 @@ func (m *metricsV2) reportDisperseBlobSize(size int) {
 	m.disperseBlobSize.WithLabelValues().Add(float64(size))
 }
 
-func (m *metricsV2) reportDisperseMeteredBytes(usageInBytes int) {
-	m.disperseBlobMeteredBytes.WithLabelValues().Add(float64(usageInBytes))
+func (m *metricsV2) reportDisperseMeteredBytes(quorumNumbers []uint8, symbolsCharges map[core.QuorumID]uint64) {
```
anupsv, line 176: Is this best suited in some helper file style thing ?

anupsv (approved):

PR #1673, opened 2025-06-18
feat: validator blacklist with mutex

anupsv:

Adding read and write mutex for validator blacklisting

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


node/blacklist_store.go
```diff
@@ -103,18 +107,32 @@ func (s *blacklistStore) hasKey(key []byte) bool {
...
 func (s *blacklistStore) DeleteByDisperserID(ctx context.Context, disperserId uint32) error {
+	s.mu.Lock()
+	defer s.mu.Unlock()
+
 	disperserIdHash := sha256.Sum256(fmt.Appendf(nil, "%d", disperserId))
 	return s.db.Delete(disperserIdHash[:])
 }
 
 // Get retrieves a blacklist by key
 func (s *blacklistStore) Get(ctx context.Context, key []byte) (*Blacklist, error) {
```
hopeyen: `ctx` not used, can it be removed?

node/blacklist_store.go
```diff
@@ -90,6 +91,9 @@ func (s *blacklistStore) BlacklistDisperserFromBlobCert(request *pb.StoreChunksR
 
 // HasDisperserID checks if a disperser ID exists in the store
 func (s *blacklistStore) HasDisperserID(ctx context.Context, disperserId uint32) bool {
+	s.mu.RLock()
+	defer s.mu.RUnlock()
+
 	// hash the disperserId and look up
 	disperserIdHash := sha256.Sum256([]byte(fmt.Sprintf("%d", disperserId)))
```
hopeyen: this uses Sprintf and L113 uses Appendf. would it be good to standardize the hashing?

node/blacklist_store.go
```diff
@@ -171,27 +202,49 @@ func (s *blacklistStore) IsBlacklisted(ctx context.Context, disperserId uint32)
...
-	// If so, we delete the disperser from the blacklist to avoid checking each time and wasting resources.
-	if s.time.Since(s.time.Unix(int64(lastUpdated), 0)) >= time.Hour*24*14 {
-		err = s.DeleteByDisperserID(ctx, disperserId)
+	// Check if we need to delete expired entry (14+ days old)
+	needsCleanup := timeSinceLastUpdate >= time.Hour*24*14
+
+	if needsCleanup {
+		// Upgrade to write lock for deletion
+		s.mu.RUnlock()
+		s.mu.Lock()
```
hopeyen, line 229: I'm wondering if there's a gap between RUnlock and Lock for another goroutine to modify the data?

node/blacklist_store.go
anupsv: yep sure.

node/blacklist_store.go
anupsv: fixed.

node/blacklist_store.go
anupsv: yea, that's handled below right ?

hopeyen: i see, okay!

hopeyen (approved):

PR #1676, opened 2025-06-19
feat: add address directory

pakim249CAL:

A POC of an address directory contract. Deploying this and using it would minimize the need for contract upgrades to deploy and use new components.

The idea is that all current and future contracts would be registered into this directory. Any future contract upgrades to our contracts will remove the respective contracts from their storage, and start referring to the directory. 

This design is meant to alleviate the circular dependencies of the eigenDA contracts, and provide a contract whos only purpose is to serve as the source of truth on what contracts exist in the eigenDA ecosystem.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


contracts/src/core/interfaces/IEigenDAAddressDirectory.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+interface IEigenDAAddressDirectory {
+    event AddressSet(string name, address indexed value);
```
samlaf: why not put an index on the name?

contracts/src/core/interfaces/IEigenDAAddressDirectory.sol
```diff
@@ -0,0 +1,15 @@
...
+
+    /// @notice Sets the address of a contract.
+    function setAddress(string memory name, address value) external;
+
+    /// @notice Gets the address by keccak256 hash of the name.
+    function getAddress(bytes32 key) external view returns (address);
```
samlaf: do we really need this? Who would use this? Should we add the key as a field in the emitted event if so?

samlaf (dismissed):
LGTM. simple enough

contracts/src/core/interfaces/IEigenDAAddressDirectory.sol
pakim249CAL: In this case, I would like the string to be readable from the event log. Otherwise re-execution would be needed to figure out the string.

contracts/src/core/interfaces/IEigenDAAddressDirectory.sol
pakim249CAL: This is the entry point that would be more efficient for contracts to use, as they can use a predefined constant instead of doing a hash. I don't think we need the key emitted, as this can be easily derived offchain. Events only exist for offchain purposes, so I don't think that the key is necessary to emit.

contracts/src/core/interfaces/IEigenDAAddressDirectory.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+interface IEigenDAAddressDirectory {
```
litt3: A description of this interface's role and purpose would be helpful

contracts/src/core/interfaces/IEigenDAAddressDirectory.sol
```diff
@@ -0,0 +1,15 @@
...
+    event AddressSet(string name, address indexed value);
+
+    /// @notice Sets the address of a contract.
+    function setAddress(string memory name, address value) external;
+
+    /// @notice Gets the address by keccak256 hash of the name.
```
litt3: Do you think it's correct to specify key format as keccak256 here? Isn't that just an implementation detail of the `AddressDirectoryStorage.sol` contract?

contracts/src/core/libraries/v3/address-directory/AddressDirectoryStorage.sol
```diff
@@ -0,0 +1,19 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+library AddressDirectoryStorage {
```
litt3, line 6: There should be a doc here explaining the storage model being used by this library

contracts/src/core/interfaces/IEigenDAAddressDirectory.sol
```diff
@@ -0,0 +1,15 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+interface IEigenDAAddressDirectory {
+    event AddressSet(string name, address indexed value);
+
+    /// @notice Sets the address of a contract.
+    function setAddress(string memory name, address value) external;
```
litt3: Is there a reason you didn't implement `removeAddress` functionality to go along with `setAddress`?

contracts/src/core/EigenDAAddressDirectory.sol
```diff
@@ -0,0 +1,33 @@
...
+import {AddressDirectoryStorage} from "src/core/libraries/v3/address-directory/AddressDirectoryStorage.sol";
+import {IEigenDAAddressDirectory} from "src/core/interfaces/IEigenDAAddressDirectory.sol";
+
+contract EigenDAAddressDirectory is OwnableUpgradeable, IEigenDAAddressDirectory {
+    function initialize(address _initialOwner) external initializer {
+        _transferOwnership(_initialOwner);
+    }
+
+    /// @inheritdoc IEigenDAAddressDirectory
+    function setAddress(string memory name, address value) external onlyOwner {
```
litt3: Perhaps it would be good to check that `name` is not an empty string?

contracts/src/core/libraries/v3/address-directory/AddressDirectoryStorage.sol
litt3, line 1: Can you help me understand the package naming scheme you're using? We've discussed this before, but I'm still not clear.

This `AddressDirectoryStorage.sol` file is nested under a directory called `v3`. If there is another change of similar degree one month from now, will there be a `v4` directory added?

contracts/src/core/interfaces/IEigenDAAddressDirectory.sol
pakim249CAL: A remove operation is simply a check and set. I can rework it to be:

* Add
* Remove
* Replace

instead of set?

contracts/src/core/EigenDAAddressDirectory.sol
pakim249CAL: I personally think it's okay. There is IMO no real reason to include this constraint, as I wanted the full understanding of the contract to simply be a hash table keyed by string. As soon as we include constraints like this, it no longer functions exactly as the primitive programmers are used to.

contracts/src/core/libraries/v3/address-directory/AddressDirectoryStorage.sol
pakim249CAL: No. I think there will need to be a naming change.
 
V3 for now is my catchall for contract designs that use unstructured storage libraries. There should be no need in the future for a V4, etc, because there is no inheritance in the V3 folder, and types/logic are separated by component.

I am open to suggestions haha. I am thinking long term that hopefully we can delete V1 and V2 from the repository entirely, eventually. Then there will be no version numbers.

contracts/src/core/interfaces/IEigenDAAddressDirectory.sol
pakim249CAL: I think it is useful considering that how to hash the string is the only thing that needs to be known about how to use this function. I am thinking from the perspective of someone who reads this comment on etherscan and is wondering how to get the key.

contracts/src/core/interfaces/IEigenDAAddressDirectory.sol
litt3: > A remove operation is simply a check and set

Is the idea that you would pass in an empty string? Wouldn't that still leave an entry in the map? 

> I can rework it

I do think `add`, `remove`, `replace` would be a better API than only `set`. The intention when making one of these calls is more explicit.

Curious if anyone else has opinions here cc @samlaf

pakim249CAL: > Is the idea that you would pass in an empty string? Wouldn't that still leave an entry in the map?

No, a remove operation would specify a key to remove the value for. And the value would be set to address(0)

> I do think add, remove, replace would be a better API than only set. The intention when making one of these calls is more explicit.

I think it is okay. My only opinion is that we should either do one or the other. Have only set, or only add/remove/replace.

pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1676/commits/3070f4af8cb5af71d3ff58d7f4091875d524ec48

contracts/src/core/libraries/v3/address-directory/AddressDirectoryStorage.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1676/commits/1c4d0539537d5fbec41f7e2b04776ea3c5affc84

contracts/src/core/interfaces/IEigenDAAddressDirectory.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1676/commits/3070f4af8cb5af71d3ff58d7f4091875d524ec48

contracts/script/deploy/eigenda-directory/config/example.config.toml
```diff
@@ -0,0 +1,3 @@
+
+contractNames = ["EXAMPLE", "EXAMPLE2"]
```
litt3, line 3: A comment here would be helpful to understand what this is an example for.

contracts/src/core/interfaces/IEigenDADirectory.sol
```diff
@@ -0,0 +1,31 @@
...
+    error NewValueIsOldValue(address value);
+
+    event AddressAdded(string name, bytes32 indexed key, address indexed value);
+    event AddressReplaced(string name, bytes32 indexed key, address indexed oldValue, address indexed newValue);
+    event AddressRemoved(string name, bytes32 indexed key);
+
+    /// @notice Adds a new address to the directory by name.
+    function addAddress(string memory name, address value) external;
+
+    /// @notice Replaces an existing address in the directory by name.
```
litt3, line 22: I think we should document the different "failure" cases on the interface functions, i.e. adding when something already exists, removing when it doesn't, etc.. What errors occur?

And for the "success" case, what events are emitted?

contracts/src/core/interfaces/IEigenDADirectory.sol
```diff
@@ -0,0 +1,31 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+/// @notice Interface for the EigenDA Directory
+///         This interface currently only includes functions for managing a directory of addresses by name.
+///         In the future, it may be extended to include access control as well.
+interface IEigenDADirectory {
+    error AddressAlreadyExists(string name);
+    error AddressNotAdded(string name);
```
litt3: `AddressNotAdded` is not unambiguous without at least comment: it could mean the address wasn't added prior to attempting some action, or it could mean the address wasn't added when attempting to add.

I think `AddressDoesNotExist` is less ambiguous (and it also matches nicely with `AddressAlreadyExists`)

pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1676/commits/48ef8dfb188f7c9a3bc6944692ac4f6ccbd6546a

contracts/src/core/interfaces/IEigenDADirectory.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1676/commits/48ef8dfb188f7c9a3bc6944692ac4f6ccbd6546a

contracts/script/deploy/eigenda-directory/config/example.config.toml
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1676/commits/e184987e3f5fbbb267fbb20f741b834f4ebe0c67

contracts/src/core/interfaces/IEigenDAAddressDirectory.sol
samlaf: I see. This is very valuable information, that should be added as documentation for the method.

pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1676/commits/aa49c22ba9f285f701af8f8ce3046b61e24f3a12

samlaf: You left out the actually valuable part of your comment: "would be more efficient FOR CONTRACTS to use". This is the only part that is useful to outsiders. WHen I look at a function in an interface, I wonder "why is this here, and who would be interested in it?"

samlaf (approved):

litt3 (approved):

PR #1681, opened 2025-06-23
feat: utility methods needed for LittDB snapshots.

cody-littley:

This code is needed to support LittDB snapshots. I split it into a stand alone PR to reduce the size of the PR.

### DIFF
--- a/litt/util/files.go
+++ b/litt/util/files.go
@@ -281,7 +281,7 @@ func recursiveMoveFile(source string, destination string, deep bool, preserveOri
 
 	// Check if source is a symlink
 	if sourceInfo.Mode()&os.ModeSymlink != 0 {
-		return handleSymlink(source, destination, deep, preserveOriginal, fsync)
+		return moveSymlink(source, destination, deep, preserveOriginal, fsync)
 	}
 
 	// Copy the file
@@ -306,7 +306,18 @@ func recursiveMoveFile(source string, destination string, deep bool, preserveOri
 	return nil
 }
 
-// recursiveMoveDirectory handles moving a directory and its contents
+// recursiveMoveDirectory handles moving a directory and its contents.
+//
+//   - If preserveOriginal is true, the original directory and its contents will be preserved after the move.
+//   - If preserveOriginal is false, the original directory and its contents will be deleted after the move.
+//     If the original contains symlinks, then both the symlinks and the files the symlinks point to will be deleted.
+//     This is intentional, and designed to support semantics needed by LittDB snapshotting operations.
+//
+// - If deep is false, then files will be hard linked where possible, and symlinks will be copied as symlinks.
+// - If deep is true, then files will always be copied, and symlinks will be followed and copied as regular files.
+//
+// There is a special interaction if preserveOriginal is false and deep is true. If the source and the destination
+// are on the same filesystem, then the file is moved instead of copied.
 func recursiveMoveDirectory(
 	source string,
 	destination string,
@@ -353,7 +364,7 @@ func recursiveMoveDirectory(
 
 		case (info.Mode() & os.ModeSymlink) != 0:
 			// Handle symlink
-			if err := handleSymlink(path, destPath, deep, preserveOriginal, fsync, source); err != nil {
+			if err := moveSymlink(path, destPath, deep, preserveOriginal, fsync, source); err != nil {
 				return fmt.Errorf("failed to handle symlink: %w", err)
 			}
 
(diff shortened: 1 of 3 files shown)


litt/util/files.go
```diff
@@ -208,15 +346,180 @@ func CopyDirectoryRecursively(source string, destination string) error {
...
+		// Sync if requested
+		if fsync {
+			if err := syncFile(destination); err != nil {
+				return fmt.Errorf("failed to sync copied symlink target: %w", err)
+			}
+		}
+
+		// Remove original target file if not preserving original
+		if !preserveOriginal {
+			if err := os.Remove(target); err != nil {
```
litt3: Shouldn't you be removing the `source` here, not the `target`? I think the current impl would delete the original file, but it seems like you should just be deleting the original symlink...

litt/util/files.go
```diff
@@ -208,15 +346,180 @@ func CopyDirectoryRecursively(source string, destination string) error {
...
+
+	// Get target file info
+	targetInfo, err := os.Stat(target)
+	if err != nil {
+		return fmt.Errorf("failed to stat symlink target %s: %w", target, err)
+	}
+
+	if targetInfo.IsDir() {
+		// Target is a directory, recursively copy it
+		return recursiveMoveDirectory(target, destination, deep, preserveOriginal, fsync)
```
litt3, line 516: Here too, it seems like you are deleting the original where you should not be. If `preserveOriginal` is false in this situation, you will do a deep copy of the symlink but delete the original directory, right? That doesn't seem like the correct behavior.

litt/util/files_test.go
```diff
@@ -1095,6 +985,1156 @@ func TestAtomicRenameAcrossFilesystems(t *testing.T) {
...
+		t.Run(tc.name, func(t *testing.T) {
+			source, dest := tc.setup()
+
+			err := RecursiveMove(source, dest, tc.deep, tc.preserveOriginal, false)
+
+			if tc.name == "error - directory with internal symlinks and deep copy" {
+				require.Error(t, err)
+				require.Contains(t, err.Error(), "cannot deep copy symlink")
+				require.Contains(t, err.Error(), "target")
+				require.Contains(t, err.Error(), "is within the source directory being moved")
```
litt3, line 2228: Why are you asserting multiple different substrings be contained in this error?

litt/util/files.go
cody-littley: It was my intention for this workflow to delete both the symlink and the file the symlink points to. 

I did find a bug there though, since I wasn't deleting the symlink like I was supposed to. I've expanded the godoc and comments to explain what is going on, and fixed it so that it deletes both the symlink and the referenced file.

litt/util/files.go
cody-littley: Added some additional documentation to the intended semantics.

// recursiveMoveDirectory handles moving a directory and its contents.
//
// - If preserveOriginal is true, the original directory and its contents will be preserved after the move.
// - If preserveOriginal is false, the original directory and its contents will be deleted after the move.
//   If the original contains symlinks, then both the symlinks and the files the symlinks point to will be deleted.
//   This is intentional, and designed to support semantics needed by LittDB snapshotting operations.
//
// - If deep is false, then files will be hard linked where possible, and symlinks will be copied as symlinks.
// - If deep is true, then files will always be copied, and symlinks will be followed and copied as regular files.
//
// There is a special interaction if preserveOriginal is false and deep is true. If the source and the destination
// are on the same filesystem, then the file is moved instead of copied.

litt/util/files_test.go
cody-littley: The string that the error contains is `cannot deep copy symlink %s: target %s is within the source directory being moved`. Claude wrote this to check the error.

Do you think it's too complex? I can delete the string checking and just make sure it returns an error... I almost never check error strings when I write unit tests by hand.

litt3: I don't think these test assertions are helpful.

I've been thinking a lot about claude generated unit tests lately. There's a philosophy I've seen where there is a commandment: `AI shall not touch tests`. The idea is that tests encode *human* intention. I think I'm kind of warming up to the idea.

Take this utility PR, for example: I assume that you had Claude help you write the utility, and then had Claude generate the unit tests. The problem is that Claude will frequently write tests that just... assert that the implementation is the way that it is. This can give the illusion of very good test coverage, but unless someone does an in depth review the 1000 lines of test code, I'm not super confident in the guarantees we are getting.

It's a difficult question because Claude is *so good* at hammering out a bunch of tests with pretty good coverage. Thoughts?

PR #1682, opened 2025-06-25
fix: payment state api doesn't fail at zero valued reservations/ondemand

hopeyen:

Added an error type of `ErrPaymentDoesNotExist`

If a reservation has zero on all the fields or ondemand account has zero balance, we treat that payment as it doesn't exist. 
Within [4f0684f](https://github.com/Layr-Labs/eigenda/pull/1664/commits/4f0684f5fa0c64845783072493357cd04b049a95), we updated the behavior from allow zero values as fallback to all fields, to returning error when there's an error from the onchain payment state.
It is too strict for users who are not using on-demand, because ondemand being zero was considered an error by the eth reader. We update to make eth reader return a specific error type, and allow this error to use zero value when constructing payment states.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/eth/utils.go
+++ b/core/eth/utils.go
@@ -1,11 +1,11 @@
 package eth
 
 import (
-	"fmt"
 	"math/big"
 	"slices"
 
 	"github.com/Layr-Labs/eigenda/core"
+	"github.com/pingcap/errors"
 
 	eigendasrvmg "github.com/Layr-Labs/eigenda/contracts/bindings/EigenDAServiceManager"
 	paymentvault "github.com/Layr-Labs/eigenda/contracts/bindings/PaymentVault"
@@ -15,6 +15,9 @@ import (
 
 var (
 	maxNumberOfQuorums = 192
+
+	// ErrPaymentDoesNotExist is returned when a payment/deposit does not exist (is zero)
+	ErrPaymentDoesNotExist = errors.New("payment does not exist")
 )
 
 type BN254G1Point struct {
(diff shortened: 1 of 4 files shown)


core/eth/utils.go
```diff
@@ -1,11 +1,11 @@
 package eth
 
 import (
-	"fmt"
 	"math/big"
 	"slices"
 
 	"github.com/Layr-Labs/eigenda/core"
+	"github.com/pingcap/errors"
```
dmanc, line 8: What's this import for? Can we use the standard library one?

core/eth/utils.go
```diff
@@ -15,6 +15,9 @@ import (
 
 var (
 	maxNumberOfQuorums = 192
+
+	// ErrPaymentDoesNotExist is returned when a reservation/deposit does not exist (is zero)
+	ErrPaymentDoesNotExist = errors.New("payment does not exist")
```
dmanc, line 20: Is this specific to reservations?

hopeyen: no, also ondemand. do you want them separate?

core/eth/utils.go
hopeyen: this package is also used in eth [writer](https://github.com/Layr-Labs/eigenda/blob/2a6845841ac66627434cb63be7c2df0f330db663/core/eth/writer.go#L25) and [reader](https://github.com/Layr-Labs/eigenda/blob/3623ddab8d84f900708ff2cfdc3c0c8854c4f6a7/core/eth/reader.go#L32). i think mainly for adding context to the error

dmanc (approved):

PR #1684, opened 2025-06-25
feat: improve file utilities

cody-littley:

Improve LittDB file utilities. Required for a following PR that adds LittDB snapshotting.


litt/util/files_utils.go
```diff
@@ -112,17 +154,16 @@ func AtomicRename(oldPath string, newPath string, fsync bool) error {
 	return nil
 }
 
-// VerifyFileProperties checks if a file has read/write permissions and is a regular file (if it exists),
-// returning an error if it does not if the file permissions or file type is not as expected.
-// Also returns a boolean indicating if the file exists and its size (to save on additional os.Stat calls).
+// ErrIfNotWritableFile verifies that a path is either a regular file that with read+write permissions,
```
litt3: > regular file that with read+write permissions,

grammar

litt/util/files_utils.go
```diff
@@ -133,15 +174,17 @@ func VerifyFileProperties(path string) (exists bool, size int64, err error) {
...
 			}
 
 			if !parentInfo.IsDir() {
 				return false, -1, fmt.Errorf("parent directory %s is not a directory", parentPath)
 			}
 
 			if parentInfo.Mode()&0700 != 0700 {
-				return false, -1, fmt.Errorf("parent directory %s has insufficent permissions", parentPath)
+				return false, -1, fmt.Errorf(
+					"parent directory %s has insufficent permissions", parentPath)
```
litt3: While you're editing the line, may as well fix the spelling of `insufficent` (also misspelled elsewhere in this file0)

litt/util/files_utils.go
```diff
@@ -245,40 +229,85 @@ func verifyDirectoryWritable(dirPath string) error {
...
+		return fmt.Errorf("failed to sync file: %w", err)
 	}
 
 	return nil
 }
 
-// copyRegularFile copies a regular file from src to dst, preserving permissions and timestamps.
-func copyRegularFile(src string, dst string, fileMode os.FileMode, modTime time.Time) error {
+// SyncDirectory syncs a directory to disk
+func SyncDirectory(path string) error {
```
litt3: This function is identical to the `SyncFile` function, apart from the error message.

Consider having only one function with a more generic message, though I don't feel strongly, so whichever you prefer is fine.

litt/util/files_utils.go
```diff
@@ -245,40 +229,85 @@ func verifyDirectoryWritable(dirPath string) error {
...
+	return nil
+}
+
+// SyncParentDirectory syncs the parent directory of the given path.
+func SyncParentDirectory(path string) error {
+	return SyncDirectory(filepath.Dir(path))
+}
+
+// CopyRegularFile copies a regular file from src to dst.
+func CopyRegularFile(src string, dst string, fsync bool) error {
```
litt3: It could be worth noting that this will delete a file if it already exists at the destination.

litt/util/files_utils.go
```diff
@@ -291,77 +320,110 @@ func copyRegularFile(src string, dst string, fileMode os.FileMode, modTime time.
...
+			return fmt.Errorf("failed to check path %s: %w", currentPath, err)
+		}
+
+		// Path doesn't exist, add to list of paths to create
+		pathsToCreate = append(pathsToCreate, currentPath)
+
+		// Move to parent directory
+		parentPath := filepath.Dir(currentPath)
+		if parentPath == currentPath {
+			// Reached filesystem root
```
litt3: Not requesting changes, just trying to understand the implications of this logic: if we arrive here, wouldn't that mean that there was no root directory and it has to be created?

litt/util/files_utils_test.go
```diff
@@ -1056,43 +837,211 @@ func TestAtomicRenameWithSymlink(t *testing.T) {
...
+				}
+				for _, file := range afterFiles {
+					if filepath.Ext(file) != SwapFileExtension {
+						afterRegularFiles = append(afterRegularFiles, file)
+					}
+				}
+				require.ElementsMatch(t, beforeRegularFiles, afterRegularFiles, "Regular files should be unchanged")
+
+				// Verify subdirectories are not affected
+				if tc.name == "delete swap files in directory with mixed files" {
```
litt3: This is a pretty fragile mechanism, e.g. what if someone changes the name of the test case, without understanding this mechanism?

Consider creating a constant for this name, or choosing an alternate mechanism to trigger this logic.

litt/util/files_utils_test.go
```diff
@@ -1247,3 +1196,147 @@ func TestSanitizePath(t *testing.T) {
...
+
+	symlinkToTestDirPath := filepath.Join(testDir, "link-to-dir")
+	err = os.Symlink(testDir, symlinkToTestDirPath)
+	require.NoError(t, err)
+	isSymlink, err = IsSymlink(symlinkToTestDirPath)
+	require.NoError(t, err)
+	require.True(t, isSymlink, "Symlink to directory should be detected as symlink")
+}
+
+func TestErrIfSymlink(t *testing.T) {
```
litt3: I suppose it's a question of testing philosophy, but I would personally just test `ErrIfSymlink` in `TestIsSymlink` above, for every file that you created.

On the one hand, it *does* test two things in one unit test. But on the other hand, it avoids duplicating a lot of test logic. Up to you which you prefer

litt/util/files_utils_test.go
```diff
@@ -1247,3 +1196,147 @@ func TestSanitizePath(t *testing.T) {
...
+	require.NoError(t, err)
+
+	err = ErrIfExists(fooPath)
+	require.Error(t, err, "Expected error for existing directory")
+
+	err = ErrIfExists(barPath)
+	require.Error(t, err, "Expected error for existing file")
+}
+
+func TestErrIfNotExists(t *testing.T) {
```
litt3: Also in this case, I think I would personally test `ErrIfExists` and `ErrIfNotExists` in the same unit test, since the setup is the same, and part of what you should be asserting is that they always yield opposite values.

litt/util/recursive_move.go
```diff
@@ -0,0 +1,189 @@
...
+}
+
+// moveFile handles moving a single file
+func moveFile(source string, destination string, preserveOriginal bool, fsync bool) error {
+	// Ensure parent directory exists
+	if err := EnsureParentDirectoryExists(destination, fsync); err != nil {
+		return fmt.Errorf("failed to ensure parent directory exists: %w", err)
+	}
+
+	// If not preserving original, try to move the file first (regardless of deep mode)
```
litt3, line 60: Stale comment, deep mode doesn't exist anymore.

litt/util/recursive_move.go
```diff
@@ -0,0 +1,189 @@
...
+	if !preserveOriginal {
+		// Try simple rename first (works if on same filesystem)
+		if err := os.Rename(source, destination); err == nil {
+			if fsync {
+				if err := SyncDirectory(filepath.Dir(destination)); err != nil {
+					return fmt.Errorf("failed to sync parent directory: %w", err)
+				}
+				if err := SyncDirectory(filepath.Dir(source)); err != nil {
+					return fmt.Errorf("failed to sync parent directory: %w", err)
```
litt3: You have the same error message for failing to sync the source and destination parent directories. It's probably worth having different error messages so that these can be distinguished.

litt/util/recursive_move.go
```diff
@@ -0,0 +1,189 @@
...
+			}
+
+			return nil
+		}
+		// Rename failed (likely different filesystem), fall back to copy+delete
+	}
+
+	err := ErrIfSymlink(source)
+	if err != nil {
+		return fmt.Errorf("symlinks not supported: %w", err)
```
litt3, line 80: I think your documentation about symlinks in the function doc is stale, since you mention copying them, but here you explicitly don't support symlinks

litt/util/recursive_move.go
```diff
@@ -0,0 +1,189 @@
...
+	if !preserveOriginal {
+		if err := os.Remove(source); err != nil {
+			return fmt.Errorf("failed to remove source file: %w", err)
+		}
+	}
+
+	return nil
+}
+
+// recursiveMoveDirectory handles moving a directory and its contents
```
litt3, line 109: Consider noting that symlinks are not supported.

litt/util/files_utils.go
cody-littley: fixed

litt/util/files_utils.go
cody-littley: I found two instances, fixed both.

litt/util/files_utils.go
cody-littley: Unified and renamed to `SyncPath()`.

litt/util/files_utils.go
cody-littley: Added note in godoc.

litt/util/files_utils.go
cody-littley: Updated comment:

```
if parentPath == currentPath {
	// Reached filesystem root. filepath.Dir("/") returns "/", so we stop here.
	break
}
```

This loop will always abort before it checks to see if `/` exists.

bedrock

litt/util/files_utils_test.go
cody-littley: Made a constant for the name.

This is an AI generated test, but it was added in a prior commit before our self enforced moratorium on AI generated tests.

litt/util/files_utils_test.go
cody-littley: I've combined these tests

litt/util/files_utils_test.go
cody-littley: tests combined

litt/util/recursive_move.go
cody-littley: removed

litt/util/recursive_move.go
cody-littley: Fixed. It makes me wish there was some macro for adding file name and line number to an error message.

litt/util/recursive_move.go
cody-littley: updated doc

litt/util/recursive_move.go
cody-littley: done

PR #1685, opened 2025-06-25
feat: use weak pointers for LittDB cache.

cody-littley:

Use weak pointers for LittDB cache. Feature can be toggled with a flag if we detect performance issues down the road.

I have not yet benchmarked this code. I will not merge this branch until I have done that benchmarking.

### DIFF
--- a/common/cache/fifo_cache.go
+++ b/common/cache/fifo_cache.go
@@ -28,6 +28,9 @@ type insertionRecord struct {
 	key any
 	// The time at which the key was added to the cache.
 	timestamp time.Time
+	// The weight of the key-value pair. Important to keep this around instead of recomputing in case the base
+	// type uses weak pointers or equivalent.
+	weight uint64
 }
 
 // NewFIFOCache creates a new FIFOCache. If the calculator is nil, the weight of each key-value pair will be 1.
@@ -71,6 +74,7 @@ func (f *FIFOCache[K, V]) Put(key K, value V) {
 		f.evictionQueue.Enqueue(&insertionRecord{
 			key:       key,
 			timestamp: time.Now(),
+			weight:    weight,
 		})
 	}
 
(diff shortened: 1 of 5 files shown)


samlaf:
Looks cool! Having skimmed the PR for 5 mins, I'm still unsure how this all works (not super familiar with weak pointers in golang).

What's the TLDR of how this works? your cache now stores weak pointers to heap allocated values?
And the weak pointers are allowed to be freed by the golang runtime at any point? Is the policy for cleaning the weak pointers related to the "weights" you use?

cody-littley:
@samlaf 

# Weak Pointers

> What's the TLDR of how this works? your cache now stores weak pointers to heap allocated values?
And the weak pointers are allowed to be freed by the golang runtime at any point? Is the policy for cleaning the weak pointers related to the "weights" you use?

You can think of a weak pointer kind of like an atomic reference, i.e. it's an object that has an internal pointer to "a thing". 

```
x := "this is some data"
var xPtr weak.Pointer[string] = weak.Make(&x)

// ...

y := xPtr.Value()
```

The primary difference between a weak pointer and an atomic reference (other than thread concurrency stuff) is that 
when you get the value from a weak pointer, it may decide to return `nil`, even if you put something into it previously. Every time you get a value out of a weak pointer, you must first check to see if the returned value is nil or not.

```
y := xPtr.Value()

if y == nil {
    fmt.Println("x is nil")
} else {
    fmt.printf("the value of x is '%s'", y)
}
```

The mechanics of when the garbage collector will free data referenced by weak pointers is entirely up to the garbage collector, there is no strong contract. If there is lots of available memory, I'd expect the garbage collector not to touch data that is weakly referenced. If there is memory pressure, the garbage collector is free reclaim weakly referenced memory any time it wants.

# LittDB Cache

LittDB's cache is a FIFO cache. It maintains a queue of values in the order they are inserted into the cache, and deletes old values to ensure that the sum of the total size of the cache stays below some threshold.

The logic for deciding when to delete things from the FIFO cache is unaware that it is holding weak pointers. It makes its decisions about when to remove data from the cache based on the potential memory utilized by its values, not the actual size of the data held by the weak pointers.

In the background, the garbage collector may decide to free the data referenced by the weak pointers. If it does so, those weak pointers will remain in the cache. If somebody attempts to fetch data referenced by these pointers, the cache will detect that the weak pointer has been released, and the return value of the `Get()` method will be be equivalent to what happens if you request a key that is not in the cache at all.

cody-littley:
ugh... why does golang make so many horrible design decisions? It turns out that weak pointers don't work like other languages (e.g. Java), where a weak pointer will prevent an object from being GC'd unless there is memory pressure.

Closing this... weak pointers as implemented are essentially useless for a cache.

PR #1686, opened 2025-06-26
fix: make disperser client backwards compatible

samlaf:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/clients/v2/disperser_client.go
```diff
@@ -389,7 +392,124 @@ func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.G
...
+		return nil, err
+	}
+
+	// Convert old response to new response format
+	return convertLegacyPaymentStateToNew(oldResult), nil
+}
+
+// convertLegacyPaymentStateToNew converts the old GetPaymentStateReply to the new GetPaymentStateForAllQuorumsReply format
+// TODO: this was generated by Claude. I haven't checked that it actually makes sense.... SO DO NOT MERGE BEFORE CHECKING THIS.
+func convertLegacyPaymentStateToNew(legacyReply *disperser_rpc.GetPaymentStateReply) *disperser_rpc.GetPaymentStateForAllQuorumsReply {
```
samlaf: @hopeyen can you take a look at this function please?

api/clients/v2/disperser_client.go
```diff
@@ -389,7 +392,124 @@ func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.G
...
+			paymentVaultParams.QuorumPaymentConfigs[quorumID] = &disperser_rpc.PaymentQuorumConfig{
+				ReservationSymbolsPerSecond: 0, // Not available in legacy format
+				OnDemandSymbolsPerSecond:    legacyReply.PaymentGlobalParams.GlobalSymbolsPerSecond,
+				OnDemandPricePerSymbol:      legacyReply.PaymentGlobalParams.PricePerSymbol,
+			}
+
+			paymentVaultParams.QuorumProtocolConfigs[quorumID] = &disperser_rpc.PaymentQuorumProtocolConfig{
+				MinNumSymbols:              legacyReply.PaymentGlobalParams.MinNumSymbols,
+				ReservationAdvanceWindow:   legacyReply.PaymentGlobalParams.ReservationWindow,
+				ReservationRateLimitWindow: 0, // Not available in legacy format
```
hopeyen: `legacyReply.PaymentGlobalParams.ReservationWindow,`

api/clients/v2/disperser_client.go
```diff
@@ -389,7 +392,124 @@ func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.G
...
+			}
+		}
+	}
+
+	// Convert aggregated reservation to per-quorum format
+	reservations := make(map[uint32]*disperser_rpc.QuorumReservation)
+	if legacyReply.Reservation != nil {
+		// Apply the reservation to all quorums mentioned in the reservation
+		quorums := legacyReply.Reservation.QuorumNumbers
+		if len(quorums) == 0 {
```
hopeyen: this should be invalid and error out

disperser/apiserver/server_v2.go
```diff
@@ -290,6 +290,11 @@ func (s *DispersalServerV2) GetPaymentState(ctx context.Context, req *pb.GetPaym
 		return nil, err
 	}
 
+	return convertAllQuorumsReplyToLegacy(allQuorumsReply), nil
+}
+
+// TODO: describe the conversion logic and what this means for users.
```
hopeyen:
```
// convertAllQuorumsReplyToLegacy converts the new per-quorum payment state format to the legacy aggregated format.
// This enables backwards compatibility by flattening multi-quorum data into a single legacy response.
//
// Conversion logic:
// - PaymentVaultParams: Uses quorum 0 configuration as the global parameters
// - Reservations: Finds the most restrictive reservation across all quorums (minimum symbols/sec, latest start, earliest end)
// - PeriodRecords: Selects the highest usage for each period index across all quorums
// - OnDemand cumulative payment amounts: Passed through unchanged as they represent account-level totals
//
// This conversion may result in information loss for clients that need per-quorum details,
// but preserves the most restrictive constraints to ensure client behavior remains within reservation or ondemand.
```

api/clients/v2/disperser_client.go
```diff
@@ -389,7 +392,124 @@ func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.G
...
+			QuorumPaymentConfigs:  make(map[uint32]*disperser_rpc.PaymentQuorumConfig),
+			QuorumProtocolConfigs: make(map[uint32]*disperser_rpc.PaymentQuorumProtocolConfig),
+			OnDemandQuorumNumbers: legacyReply.PaymentGlobalParams.OnDemandQuorumNumbers,
+		}
+
+		// Apply the global params to all quorums mentioned in on-demand quorum numbers
+		// If no on-demand quorums specified, apply to quorum 0 as default
+		quorums := legacyReply.PaymentGlobalParams.OnDemandQuorumNumbers
+		if len(quorums) == 0 {
+			quorums = []uint32{0} // Default to quorum 0
```
hopeyen: should error out here

api/clients/v2/disperser_client.go
samlaf:
```suggestion
				ReservationRateLimitWindow: legacyReply.PaymentGlobalParams.ReservationWindow,
```

samlaf: Do you mean this change? What about ReservationAdvanceWindow? Is that also ok to be set to ReservationWindow?

hopeyen: yep to the suggested change. 
`ReservationAdvanceWindow` is not used offchain at the moment so it's okay to set to any value. I wanted this field for consistency with onchain data structure but maybe we will end up removing it entirely

disperser/apiserver/server_v2.go
samlaf: [5eceac6](https://github.com/Layr-Labs/eigenda/pull/1686/commits/5eceac6dee2314460ac5dae3a95bb0970e8a25f1)

api/clients/v2/disperser_client.go
samlaf: [066fcb4](https://github.com/Layr-Labs/eigenda/pull/1686/commits/066fcb4d1f1b87f98dac98ea64bd345af61b2b35)

api/clients/v2/disperser_client.go
samlaf: [066fcb4](https://github.com/Layr-Labs/eigenda/pull/1686/commits/066fcb4d1f1b87f98dac98ea64bd345af61b2b35)

api/clients/v2/disperser_client.go
samlaf: [249d831](https://github.com/Layr-Labs/eigenda/pull/1686/commits/249d8311d80ba349c13bfad5363f169728b3aca0)

api/clients/v2/disperser_client.go
```diff
@@ -390,7 +393,126 @@ func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.G
...
+			// Fall back to old method
+			return c.getPaymentStateFromLegacyAPI(ctx, accountID, signature, timestamp)
+		}
+		return nil, err
+	}
+
+	return allQuorumsReply, nil
+}
+
+// this is true if we are targetting a disperser that hasn't upgraded to the new API yet.
```
litt3: targetting -> targeting

api/clients/v2/disperser_client.go
```diff
@@ -390,7 +393,126 @@ func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.G
...
+	if err != nil {
+		return nil, err
+	}
+
+	// Convert old response to new response format
+	return convertLegacyPaymentStateToNew(oldResult)
+}
+
+// convertLegacyPaymentStateToNew converts the old GetPaymentStateReply to the new GetPaymentStateForAllQuorumsReply format
+// TODO: this was generated by Claude. I haven't checked that it actually makes sense.... SO DO NOT MERGE BEFORE CHECKING THIS.
```
litt3: reminder to remove comment before merging

api/clients/v2/disperser_client.go
```diff
@@ -390,7 +393,126 @@ func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.G
...
+
+// this is true if we are targetting a disperser that hasn't upgraded to the new API yet.
+func isMethodNotFoundError(err error) bool {
+	if st, ok := status.FromError(err); ok {
+		return st.Code() == codes.Unimplemented
+	}
+	return false
+}
+
+func (c *disperserClient) getPaymentStateFromLegacyAPI(
```
litt3, line 422: Consider adding a godoc here explaining why this method has to exist, and when it can be removed

api/clients/v2/disperser_client.go
```diff
@@ -390,7 +393,126 @@ func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.G
...
+	ctx context.Context, accountID gethcommon.Address, signature []byte, timestamp uint64,
+) (*disperser_rpc.GetPaymentStateForAllQuorumsReply, error) {
+	// Convert old request/response format to new format
+	oldRequest := &disperser_rpc.GetPaymentStateRequest{
```
litt3, line 425: Is the idea that `GetPaymentStateRequest` is entirely deprecated now? If that's the case, it would be really good to have a document there

Actually, now that I look at it more, why is there a new struct at all? `GetPaymentStateRequest` and `GetPaymentStateForAllQuorums` seem to have all the same fields. cc @hopeyen

api/clients/v2/disperser_client.go
```diff
@@ -390,7 +393,126 @@ func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.G
...
+	return convertLegacyPaymentStateToNew(oldResult)
+}
+
+// convertLegacyPaymentStateToNew converts the old GetPaymentStateReply to the new GetPaymentStateForAllQuorumsReply format
+// TODO: this was generated by Claude. I haven't checked that it actually makes sense.... SO DO NOT MERGE BEFORE CHECKING THIS.
+func convertLegacyPaymentStateToNew(legacyReply *disperser_rpc.GetPaymentStateReply) (*disperser_rpc.GetPaymentStateForAllQuorumsReply, error) {
+
+	// Convert PaymentGlobalParams to PaymentVaultParams
+	var paymentVaultParams *disperser_rpc.PaymentVaultParams
+	if legacyReply.PaymentGlobalParams != nil {
```
litt3: This is checking whether the payment global params are nil or not. But, is it even feasible for this to be nil? Would that still be a valid reply?

api/clients/v2/disperser_client.go
hopeyen: It does have the same fields. I wanted to keep the names consistent and later remove `GetPaymentStateRequest`

samlaf: @litt3 the buf linter rules that we enforce actually have a specific rule for this, which enforces every rpc to use a separate request/reply struct: https://buf.build/docs/lint/rules/#rpc_pascal_case
image

Basically for allowing evolvability of the scheme afaiu.

api/clients/v2/disperser_client.go
samlaf: [3347eb0](https://github.com/Layr-Labs/eigenda/pull/1686/commits/3347eb0c3cb10a8bd5d918d3890ce18c28c500ba)

api/clients/v2/disperser_client.go
samlaf: [c499ebe](https://github.com/Layr-Labs/eigenda/pull/1686/commits/c499ebe443d88d861096c433b27d5f82f6e70070)

api/clients/v2/disperser_client.go
samlaf: [78f2f3c](https://github.com/Layr-Labs/eigenda/pull/1686/commits/78f2f3c59f7c970b6c44d18e72d4305160a0b37b)

api/clients/v2/disperser_client.go
samlaf: I believe you are right, I don't think this is possible, and wouldn't make sense afaiu.
Returning an error now: [f417dbd](https://github.com/Layr-Labs/eigenda/pull/1686/commits/f417dbd768732c9e8e11a7664dc0484679168518)
@hopeyen can you double check this?

samlaf:
@hopeyen @litt3 added documentation to the proto files. PTAL to make sure it makes sense: [e2a0493](https://github.com/Layr-Labs/eigenda/pull/1686/commits/e2a0493c35ceab50c72eeed90ea0ef49f3ba8f89)

api/clients/v2/disperser_client.go
litt3: Makes sense. This would be really good to document on the deprecated request.

api/clients/v2/disperser_client.go
```diff
@@ -390,7 +393,137 @@ func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.G
...
+	var paymentVaultParams *disperser_rpc.PaymentVaultParams
+	{
+		paymentVaultParams = &disperser_rpc.PaymentVaultParams{
+			QuorumPaymentConfigs:  make(map[uint32]*disperser_rpc.PaymentQuorumConfig),
+			QuorumProtocolConfigs: make(map[uint32]*disperser_rpc.PaymentQuorumProtocolConfig),
+			OnDemandQuorumNumbers: legacyReply.PaymentGlobalParams.OnDemandQuorumNumbers,
+		}
+
+		// Apply the global params to all quorums mentioned in on-demand quorum numbers
+		// If no on-demand quorums specified, apply to quorum 0 as default
```
litt3: > // If no on-demand quorums specified, apply to quorum 0 as default

This seems to be outdated. You are throwing an error if there are no quorums specified.

api/clients/v2/disperser_client.go
```diff
@@ -390,7 +393,137 @@ func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.G
...
+			}
+
+			paymentVaultParams.QuorumProtocolConfigs[quorumID] = &disperser_rpc.PaymentQuorumProtocolConfig{
+				MinNumSymbols: legacyReply.PaymentGlobalParams.MinNumSymbols,
+				// ReservationAdvanceWindow is not used offchain at the moment so it's okay to set to any value.
+				// It was added for consistency with the onchain data structure but get removed in the future.
+				ReservationAdvanceWindow:   legacyReply.PaymentGlobalParams.ReservationWindow,
+				ReservationRateLimitWindow: legacyReply.PaymentGlobalParams.ReservationWindow,
+				OnDemandRateLimitWindow:    0, // Not available in legacy format
+				OnDemandEnabled:            len(legacyReply.PaymentGlobalParams.OnDemandQuorumNumbers) > 0,
```
litt3: I don't think it's possible for the on-demand quorum numbers to *not* be greater than zero.

api/clients/v2/disperser_client.go
```diff
@@ -390,7 +393,137 @@ func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.G
...
+	}
+
+	// Convert aggregated period records to per-quorum format
+	periodRecords := make(map[uint32]*disperser_rpc.PeriodRecords)
+	// If no period records are available, return empty map
+	// This is possible if no reservations exist for the account
+	if len(legacyReply.PeriodRecords) > 0 {
+		// Apply the same period records to all relevant quorums
+		quorums := []uint32{0} // Default to quorum 0
+		if legacyReply.Reservation != nil && len(legacyReply.Reservation.QuorumNumbers) > 0 {
```
litt3: Wouldn't it constitute an error if the `legacyReply.Reservation` was nil but there are elements in `legacyReply.PeriodRecords`? 

Similarly, under what conditions would there *not* be any defined quorum numbers when there *is* a reservation defined?

api/clients/v2/disperser_client.go
```diff
@@ -390,7 +393,137 @@ func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.G
...
+				Records: legacyReply.PeriodRecords,
+			}
+		}
+	}
+
+	// Convert aggregated reservation to per-quorum format
+	reservations := make(map[uint32]*disperser_rpc.QuorumReservation)
+	// If no reservation is available, return empty map
+	// This is possible if no reservations exist for the account
+	if legacyReply.Reservation != nil {
```
litt3: I think the logic of this method could be simplified with the following strategy:

1. Before handling anything to do with a reservation, check `legacyReply.Reservation == nil`. If it is `nil`, you can immediately return your `GetPaymentStateForAllQuorumsReply`, with some empty fields.
2. Assert that all expected fields within the reservation object are non-nil.
3. Then, you are free to do all your reservation logic without having to wonder about whether there are actually any reservations, and if any field might be unexpectedly nil

api/clients/v2/disperser_client.go
hopeyen: yeah, it would return an error if the global param is nil ([here](https://github.com/Layr-Labs/eigenda/blob/2031c1d79f7703f9ffe9b698531570e85e596a5b/core/meterer/onchain_state.go#L326)). 
agreed that checking for nil then return error is cleaner

api/clients/v2/disperser_client.go
hopeyen: `OnDemandQuorumNumbers` is a list of quorum ids with on-demand enabled. Would be more accurate to set the boolean to if the current quorum id is contained within this list

api/clients/v2/disperser_client.go
hopeyen: > under what conditions would there not be any defined quorum numbers when there is a reservation defined?

shouldn't be possible by the contract checks, but could be a bad disperser

hopeyen: > under what conditions would there not be any defined quorum numbers when there is a reservation defined?

contract enforces that there must be at least one quorum number, so this would be if the disperser server didn't return the correct values

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -160,12 +160,22 @@ message GetPaymentStateReply {
...
+  // The bytes are parsed to a big.Int value.
+  // This value should always be <= cumulative_payment, as the disperser cumulative_payment kept offchain is only periodically updated onchain.
+  // See [common.v2.PaymentHeader.cumulative_payment] for more details.
+  //
+  // This value should only be nonzero for the EigenLabs disperser, as it is the only disperser that supports on-demand payments currently.
+  // Future work will support decentralized on-demand dispersals.
   bytes cumulative_payment = 4;
   // on-chain on-demand payment deposited
+  // The bytes are parsed to a big.Int value.
+  // This value should always be <= cumulative_payment, as the disperser cumulative_payment kept offchain is only periodically updated onchain.
```
hopeyen: we currently don't have payment updates on-chain at all, still something to add in the future. It would look something like disperser sees a request containing payment header, account id, and signature. disperser submits the message onchain, contract verifies, update cumulative_payment_used which would always be `onchain_cumulative_payment_used <=  onchain_cumulative_payment`. then the field above `cumulative_payment` must be within these onchain two ranges

api/clients/v2/disperser_client.go
samlaf: [44e13c0](https://github.com/Layr-Labs/eigenda/pull/1686/commits/44e13c054fdcdf5b1842a96d0b32072428c15338)

api/clients/v2/disperser_client.go
samlaf: [76faa3b](https://github.com/Layr-Labs/eigenda/pull/1686/commits/76faa3b2e5fc3ebb4bdfbe3f0e8dbee3cf4ac13c)

api/clients/v2/disperser_client.go
samlaf: [d24eacb](https://github.com/Layr-Labs/eigenda/pull/1686/commits/d24eacb7915f6b0930480958f5f8d9ed743d11ff)

api/clients/v2/disperser_client.go
samlaf: agree that this is much nicer! [d225f94](https://github.com/Layr-Labs/eigenda/pull/1686/commits/d225f946b2d1afcd258080b7b6679c0b950bdafa)

api/clients/v2/disperser_client.go
samlaf: See https://github.com/Layr-Labs/eigenda/pull/1686#discussion_r2174695047

api/proto/disperser/v2/disperser_v2.proto
samlaf: [5609dae](https://github.com/Layr-Labs/eigenda/pull/1686/commits/5609dae107a104bab33b1ac20d32c6b160360eff)

api/clients/v2/disperser_client.go
```diff
@@ -390,7 +393,132 @@ func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.G
...
+				ReservationSymbolsPerSecond: 0, // Not available in legacy format
+				OnDemandSymbolsPerSecond:    legacyReply.PaymentGlobalParams.GlobalSymbolsPerSecond,
+				OnDemandPricePerSymbol:      legacyReply.PaymentGlobalParams.PricePerSymbol,
+			}
+
+			paymentVaultParams.QuorumProtocolConfigs[quorumID] = &disperser_rpc.PaymentQuorumProtocolConfig{
+				MinNumSymbols: legacyReply.PaymentGlobalParams.MinNumSymbols,
+				// ReservationAdvanceWindow is not used offchain at the moment so it's okay to set to any value.
+				// It was added for consistency with the onchain data structure but get removed in the future.
+				ReservationAdvanceWindow:   legacyReply.PaymentGlobalParams.ReservationWindow,
```
hopeyen: nit: perhaps we set this to 0 similar to all the other "Not available in legacy format" fields, for consistency.

api/clients/v2/disperser_client.go
```diff
@@ -390,7 +393,132 @@ func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.G
...
+	var paymentVaultParams *disperser_rpc.PaymentVaultParams
+	{
+		paymentVaultParams = &disperser_rpc.PaymentVaultParams{
+			QuorumPaymentConfigs:  make(map[uint32]*disperser_rpc.PaymentQuorumConfig),
+			QuorumProtocolConfigs: make(map[uint32]*disperser_rpc.PaymentQuorumProtocolConfig),
+			OnDemandQuorumNumbers: legacyReply.PaymentGlobalParams.OnDemandQuorumNumbers,
+		}
+
+		// Apply the global params to all quorums mentioned in on-demand quorum numbers
+		quorums := legacyReply.PaymentGlobalParams.OnDemandQuorumNumbers
```
hopeyen: `quorums` won't be able to cover custom quorums, but reservations would rely on quorum configurations and the client should apply global params to all quorums it might use. I would suggest constructing `quorums` by taking an union of `legacyReply.PaymentGlobalParams.OnDemandQuorumNumbers`  and `legacyReply.Reservation.QuorumNumbers`

api/clients/v2/disperser_client.go
```diff
@@ -390,7 +393,132 @@ func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.G
...
+			}
+
+			paymentVaultParams.QuorumProtocolConfigs[quorumID] = &disperser_rpc.PaymentQuorumProtocolConfig{
+				MinNumSymbols: legacyReply.PaymentGlobalParams.MinNumSymbols,
+				// ReservationAdvanceWindow is not used offchain at the moment so it's okay to set to any value.
+				// It was added for consistency with the onchain data structure but get removed in the future.
+				ReservationAdvanceWindow:   legacyReply.PaymentGlobalParams.ReservationWindow,
+				ReservationRateLimitWindow: legacyReply.PaymentGlobalParams.ReservationWindow,
+				OnDemandRateLimitWindow:    0,    // Not available in legacy format
+				OnDemandEnabled:            true, // we are iterating over on-demand quorums, so this is always true
```
hopeyen: if we update the content of `quorums`, we might need a separate loop to set field to true for ondemand quorum numbers whereas the default is `false`

api/proto/disperser/v2/disperser_v2.proto
```diff
@@ -178,14 +190,23 @@ message GetPaymentStateForAllQuorumsReply {
...
   map<uint32, PeriodRecords> period_records = 2;
   // reservations maps quorum IDs to the on-chain account reservation record
+  // Should contain the same number of entries as the `period_records` field.
   map<uint32, QuorumReservation> reservations = 3;
-  // off-chain on-demand payment usage. This field is currently only tracked by EigenLabs disperser because on-demand requests are only
-  // supported by EigenLabs. Future work will support decentralized on-demand dispersals and this field later be tracked and shared by
-  // dispersers unlimited to EigenLabs.
+  // off-chain on-demand payment usage.
+  // The bytes are parsed to a big.Int value.
+  // This value should always be <= cumulative_payment, as the disperser cumulative_payment kept offchain is only periodically updated onchain.
```
hopeyen: this description isn't accurate atm. updating the onchain balance is something we have been putting off for the future. There's some proposal on how it updates, verifies, and handling refunds... but nothing has been decided

hopeyen (approved):

api/clients/v2/disperser_client.go
samlaf: [e00a2ba](https://github.com/Layr-Labs/eigenda/pull/1686/commits/e00a2bae835521928c5aa71a158f5ebee4a64774)

api/clients/v2/disperser_client.go
samlaf: good catch. fixed in [689f53d](https://github.com/Layr-Labs/eigenda/pull/1686/commits/689f53d88ab9426c37727d90d12fb43cc966429c)

api/clients/v2/disperser_client.go
samlaf: [689f53d](https://github.com/Layr-Labs/eigenda/pull/1686/commits/689f53d88ab9426c37727d90d12fb43cc966429c)

api/proto/disperser/v2/disperser_v2.proto
samlaf: [f10d69a](https://github.com/Layr-Labs/eigenda/pull/1686/commits/f10d69a806d7c37109a87ca79aa5e1b28cfcf18b)

api/clients/v2/disperser_client.go
```diff
@@ -390,7 +393,137 @@ func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.G
...
+
+// getPaymentStateFromLegacyAPI retrieves the payment state from the legacy GetPaymentState grpc method.
+// It is needed until we have upgraded all dispersers (testnet and mainnet) to the new API.
+// Check those endpoints for GetPaymentStateForAllQuorums using:
+// `grpcurl disperser-testnet-holesky.eigenda.xyz:443 list disperser.v2.Disperser`
+// `grpcurl disperser.eigenda.xyz:443 list disperser.v2.Disperser`
+func (c *disperserClient) getPaymentStateFromLegacyAPI(
+	ctx context.Context, accountID gethcommon.Address, signature []byte, timestamp uint64,
+) (*disperser_rpc.GetPaymentStateForAllQuorumsReply, error) {
+	// Convert old request/response format to new format
```
litt3: this doc is a bit confusing here, I'm thinking it's just something left over from claude that isn't super helpful

api/clients/v2/disperser_client.go
```diff
@@ -390,7 +393,137 @@ func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.G
...
+			OnDemandQuorumNumbers: legacyReply.PaymentGlobalParams.OnDemandQuorumNumbers,
+		}
+
+		// Apply the global params to all quorums, both on-demand and reservation.
+		onDemandQuorums := legacyReply.PaymentGlobalParams.OnDemandQuorumNumbers
+		if len(onDemandQuorums) == 0 {
+			return nil, fmt.Errorf("no on-demand quorums specified in legacy PaymentGlobalParams received from disperser")
+		}
+		reservationQuorums := legacyReply.Reservation.QuorumNumbers
+		allQuorums := append(reservationQuorums, onDemandQuorums...)
```
litt3, line 461: is appending actually the right operation? Is it guaranteed that there is no overlap between the reservation and the on-demand quorums?

api/clients/v2/disperser_client.go
```diff
@@ -390,7 +393,137 @@ func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.G
...
+			paymentVaultParams.QuorumPaymentConfigs[quorumID] = &disperser_rpc.PaymentQuorumConfig{
+				ReservationSymbolsPerSecond: 0, // Not available in legacy format
+				OnDemandSymbolsPerSecond:    legacyReply.PaymentGlobalParams.GlobalSymbolsPerSecond,
+				OnDemandPricePerSymbol:      legacyReply.PaymentGlobalParams.PricePerSymbol,
+			}
+
+			paymentVaultParams.QuorumProtocolConfigs[quorumID] = &disperser_rpc.PaymentQuorumProtocolConfig{
+				MinNumSymbols: legacyReply.PaymentGlobalParams.MinNumSymbols,
+				// ReservationAdvanceWindow is not used offchain at the moment so it's okay to set to any value.
+				// It was added for consistency with the onchain data structure but get removed in the future.
```
litt3: > but get removed in the future.

This is grammatically unclear. Is this intending to say it *will* get removed in the future? *should*?

litt3 (approved):

api/clients/v2/disperser_client.go
samlaf: [0dafcd6](https://github.com/Layr-Labs/eigenda/pull/1686/commits/0dafcd65cf96da00e097b5b6c6448f2e6d18dd85)

api/clients/v2/disperser_client.go
samlaf: [fbbb8ef](https://github.com/Layr-Labs/eigenda/pull/1686/commits/fbbb8ef84f54478b240b8013059877e68fe816b0)

api/clients/v2/disperser_client.go
samlaf: [c1f2fbc](https://github.com/Layr-Labs/eigenda/pull/1686/commits/c1f2fbcf4920a56f5e2bce5200fac0708a7c0f65)

litt3 (approved):

hopeyen (approved):

PR #1691, opened 2025-06-26
feat: eth client address directory

hopeyen:

Adds address directory support while maintaining backwards compatibility with legacy individual address configuration.

- Address directory in all core services (node, churner, batcher, etc.)
- Updates deployment scripts to include EigenDADirectory address output
- Adds address directory support to test configurations and client examples
- Implements proper validation requiring either address directory OR legacy addresses

Services that uses an Eth Client will now be taking eigendaDirectory address in configurations, instead of blsOperatorStateRetriever and eigenDAServiceManager. 

Disperser API server, batcher, controller, dataapi, node cmd, node plugin, operator churner, relay, retriever, ejection tool, quorumscan, semvarscan, proxy client

`eigenda-directory` with each service's flag prefix

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


disperser/cmd/batcher/flags/flags.go
```diff
@@ -36,16 +36,22 @@ var (
...
+		EnvVar:   common.PrefixEnvVar(envVarPrefix, "ADDRESS_DIRECTORY"),
+	}
 	BlsOperatorStateRetrieverFlag = cli.StringFlag{
 		Name:     common.PrefixFlag(FlagPrefix, "bls-operator-state-retriever"),
-		Usage:    "Address of the BLS Operator State Retriever",
-		Required: true,
-		EnvVar:   common.PrefixEnvVar(envVarPrefix, "BLS_OPERATOR_STATE_RETRIVER"),
+		Usage:    "Address of the BLS Operator State Retriever (legacy - use address-directory instead)",
+		Required: false,
+		EnvVar:   common.PrefixEnvVar(envVarPrefix, "BLS_OPERATOR_STATE_RETRIEVER"),
```
hopeyen: I changed all the `BLS_OPERATOR_STATE_RETRIVER` to `BLS_OPERATOR_STATE_RETRIEVER` bc the former has a typo, and might affect cluster env var. for vis: @dmanc

dmanc: Can we keep the old name since I assume we would deprecate this config at some point?

operators/churner/config.go
```diff
@@ -16,8 +17,9 @@ type Config struct {
 	MetricsConfig    MetricsConfig
 	ChainStateConfig thegraph.Config
 
-	BLSOperatorStateRetrieverAddr string
-	EigenDAServiceManagerAddr     string
+	AddressDirectoryAddr          string
+	BLSOperatorStateRetrieverAddr string // Legacy field, use AddressDirectoryAddr instead
```
litt3: Is there any reason to leave the legacy fields in all of these components? It seems like it would a cleaner shift if we just require anyone who updates to the new version to configure via the address directory, instead of the old way.

It's possible there's something I'm missing here, though.

hopeyen: I'm keeping these because we kept running into problems with versions running on different networks where the contracts may or may not be deployed. I thought we could wait til address registry is deployed on mainnet to remove these. This might be something to standardlized in the version release process.

if we can make sure the contract gets deployed before deploying software on different networks, then I think it's okay to go ahead and remove

api/clients/v2/examples/client_construction.go
```diff
@@ -28,10 +28,12 @@ import (
...
 const (
-	ethRPCURL                        = "https://ethereum-holesky-rpc.publicnode.com"
-	disperserHostname                = "disperser-testnet-holesky.eigenda.xyz"
-	certVerifierRouterAddress        = "0x7F40A8e1B62aa1c8Afed23f6E8bAe0D340A4BC4e"
-	registryCoordinatorAddress       = "0x53012C69A189cfA2D9d29eb6F19B32e0A2EA3490"
+	ethRPCURL                  = "https://ethereum-holesky-rpc.publicnode.com"
+	disperserHostname          = "disperser-testnet-holesky.eigenda.xyz"
+	certVerifierRouterAddress  = "0x7F40A8e1B62aa1c8Afed23f6E8bAe0D340A4BC4e"
+	registryCoordinatorAddress = "0x53012C69A189cfA2D9d29eb6F19B32e0A2EA3490"
+	addressDirectoryAddress    = "0x0000000000000000000000000000000000000000" // TODO: Replace with actual EigenDADirectory address when deployed
```
samlaf: is the directory not deployed yet?

contracts/bindings/EigenDADirectory/binding.go
samlaf, line 1: do we have a IEigenDADirectory interface instead? Better to create bindings to interfaces in general, unless there's a reason we need the contract in this case?

contracts/compile.sh
```diff
@@ -49,7 +49,8 @@ contracts="PaymentVault \
   EigenDAThresholdRegistry \
   EigenDARelayRegistry \
   IEigenDARelayRegistry \
-  EigenDADisperserRegistry"
+  EigenDADisperserRegistry \
+  EigenDADirectory"
```
samlaf: We should try to clean this up at some point and only have interfaces in here. Do we have an IEigendADirectory interface?

core/eth/reader.go
```diff
@@ -80,6 +83,44 @@ func NewReader(
...
+	}
+
+	addressDirectoryAddr := gethcommon.HexToAddress(addressDirectoryHexAddr)
+	addressDirectory, err := eigendadirectory.NewContractEigenDADirectory(addressDirectoryAddr, client)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch EigenDADirectory contract: %w", err)
+	}
+
+	// Convert contract names to keccak256 hashes as expected by the contract
+	operatorStateRetrieverKey := crypto.Keccak256Hash([]byte(ContractNames.OperatorStateRetriever))
```
samlaf: the contract actually has a mapping from names (strings) to addresses for offchain purposes.

core/eth/reader.go
```diff
@@ -80,6 +83,44 @@ func NewReader(
...
+	}
+	err = e.updateContractBindings(blsOperatorStateRetrieverAddr, eigenDAServiceManagerAddr)
+	if err != nil {
+		return nil, fmt.Errorf("failed to update contract bindings: %w", err)
+	}
+	return e, nil
+}
+
+// updateContractBindings updates the contract bindings for the reader
+// TODO: move to use address directory contract once all contracts are written into the directory
```
samlaf:
```suggestion
// TODO: update to use address directory contract once all contracts are written into the directory
```

core/eth/utils.go
```diff
@@ -176,3 +178,56 @@ func GetAllQuorumIDs(quorumCount uint8) []core.QuorumID {
 	}
 	return quorumIDs
 }
+
+// ContractNames defines the standard contract names used in the address directory
+// TODO: consider auto-generating this from the address directory contract
```
samlaf, line 181: can we do this now? prob not that hard to generate bindings for this lib. We already do something similar for cert version bindings (because its just a lib, so create a fake contract that inherits it just so we can create bindings for the consts from it)

core/eth/utils.go
```diff
@@ -176,3 +178,56 @@ func GetAllQuorumIDs(quorumCount uint8) []core.QuorumID {
...
+	ThresholdRegistry:      "THRESHOLD_REGISTRY",
+	DisperserRegistry:      "DISPERSER_REGISTRY",
+}
+
+// ValidateAddressConfig validates that either address directory is provided OR both individual addresses are provided
+// and that all provided addresses are valid hex addresses.
+func ValidateAddressConfig(addressDirectory, blsOperatorStateRetriever, eigenDAServiceManager string) error {
+	// Validate that either address directory is provided OR both individual addresses are provided
+	if addressDirectory == "" && (blsOperatorStateRetriever == "" || eigenDAServiceManager == "") {
+		return fmt.Errorf("either address-directory must be provided, or both bls-operator-state-retriever and eigenda-service-manager addresses must be provided")
```
samlaf: these kind of ambiguous error messages should be avoided. Its easy to have a more precise error in this case (check first if addressDirectory is non nil. if so, use it. If not, then check that the other 2 addresses are provided. If not, say that - instead of "either abc OR xyz", where the user needs to figure out which error he is actually hitting).

operators/churner/config.go
samlaf: Curious what kind of version issues you were running into? Overall I think (?) this approach is purely an in-place replacement for the older approach (the contract address should always be the same between both approaches if we keep both onchain for example). But are you just saying we are running into issues because of backwards compatibility until the addressRegistry gets deployed on testnet/mainnet? And after that we can sunset the other method?

I still have some questions/small-concerns regarding the approach in general, see https://github.com/Layr-Labs/eigenda/pull/1691#pullrequestreview-2971492761 (not serious concerns, just unclear as to the details of how this all works).

pakim249CAL:
> Have we thought through the consequences of using a dynamic lookup system like the AddressRegistry (as opposed to hardcoded addresses)? Does it have the exact same behavior as the current strategy of reading from serviceManager pointers?

What behaviors are you looking for here? When will it not be enough to just replace a service manager pointer with this one?

> I guess the thing I'm unsure is... would a contractAddress in the addressRegistry ever change in place? If so, what are the conditions for this/when would we want to do this? If this can happen, then can't you run into a situation where the address changes, the binary crashes, gets restarted and loads from a new contract address, but this new contract then has a breaking api change, and the (hardcoded) bindings then don't work. How do we prevent this from happening?

Address changes should happen with communication, just like any contract upgrade. It will be owned by the DA ops multisig.

api/clients/v2/examples/client_construction.go
hopeyen: no afaik. @pakim249CAL are you comfortable deploying and populating address directory on testnet holesky?

contracts/compile.sh
hopeyen: yes, updated in [commit](https://github.com/Layr-Labs/eigenda/pull/1691/commits/23bbb803d8e0b3b66cdb76e9f86bbea25e549876)
Agreed with cleaning up the compile list here. I added an linear [ticket](https://linear.app/eigenlabs/issue/EGDA-1621/use-interface-in-contract-bindings) in backlog

core/eth/reader.go
hopeyen: oh i see. I used this way for "cheaper in gas" and forgot that we are not using gas. 
updated in [commit](https://github.com/Layr-Labs/eigenda/pull/1691/commits/23bbb803d8e0b3b66cdb76e9f86bbea25e549876)

core/eth/utils.go
hopeyen: updated in [commit](https://github.com/Layr-Labs/eigenda/pull/1691/commits/23bbb803d8e0b3b66cdb76e9f86bbea25e549876)

api/clients/v2/examples/client_construction.go
pakim249CAL: Yes, I can do that.

core/eth/utils.go
samlaf: Not sure what's the best way to surface the constants to an ABI. Claude proposes a few ways:
image
image

@pakim249CAL might have a preference here. I know he's been talking about creating "periphery contracts with view functions" for a while, so maybe approach 1 would kill 2 birds with 1 stone?

pakim249CAL: A generated binding will  not give you the constant string values you are looking for. You would actually need to deploy this contract for it to be usable, which I consider to be somewhat of a nonstarter. I think the simplest solution is to just copy the strings for now. 

https://github.com/Layr-Labs/eigenda/pull/1697

This PR adds enumerability and reverse lookups, which can be used to validate against when copying the strings.

core/eth/reader.go
```diff
@@ -80,6 +83,42 @@ func NewReader(
...
+	}
+
+	addressDirectoryAddr := gethcommon.HexToAddress(addressDirectoryHexAddr)
+	addressDirectory, err := eigendadirectory.NewContractIEigenDADirectory(addressDirectoryAddr, client)
+	if err != nil {
+		return nil, fmt.Errorf("failed to fetch EigenDADirectory contract: %w", err)
+	}
+
+	// Convert contract names to keccak256 hashes as expected by the contract
+	blsOperatorStateRetrieverAddr, err := addressDirectory.GetAddress0(&bind.CallOpts{}, ContractNames.OperatorStateRetriever)
```
pakim249CAL: I did not include an error in the address directory for address(0). I think checking for 0 should be added as a validation step at all places where we retrieve an address, so that errors can be handled more gracefully than reverting when making a call to a 0 address.

core/eth/utils.go
samlaf: I'm really sorry @hopeyen I think I completely mislead you. Reading the new changes, I realized I was wrong here and the error message was fine... I think I parsed it incorrectly yesterday and thought it was confusing, but actually the new error messages are more confusing. :\
Do you agree with me? If so do you mind reverting back?

core/eth/reader.go
samlaf: Ideal approach here would prob be to create a struct with getters for each address. Those getters would get and check that the returned value is not the 0 address. Then we'd have a clean abstraction. Eventually this could be replaced by calls to the "periphery getter" contracts that @pakim249CAL has been talking about, but I think even then the same abstraction would work best.

core/eth/utils.go
samlaf: That's too bad. If there's really no way to get abigen to generate golang consts from contract consts (seems quite stupid...), then @hopeyen feel free to resolve this thread. Think we can create a nice abstraction for getters (see https://github.com/Layr-Labs/eigenda/pull/1691#discussion_r2177471578) which will eventually be tested against contract state by using @pakim249CAL's enumerability functions.

hopeyen: makes sense. I will resolve this comment for now, add a new struct for getting addresses, and rebase onto the new enumerable names contract once it is merged

contracts/bindings/EigenDADirectory/binding.go
hopeyen: replaced with IEigenDADirectory [abca944](https://github.com/Layr-Labs/eigenda/pull/1691/commits/abca944331c7d3caed57ce3d8d8bf5513e450520)

core/eth/utils.go
hopeyen: okay I reverted [abca944](https://github.com/Layr-Labs/eigenda/pull/1691/commits/abca944331c7d3caed57ce3d8d8bf5513e450520)

core/eth/reader.go
hopeyen: Added a new struct [6cc92c7](https://github.com/Layr-Labs/eigenda/pull/1691/commits/6cc92c770bd6776fb4741e297c2e140531210ea0) that comes with validation ; then a little bit abstraction [47420e9](https://github.com/Layr-Labs/eigenda/pull/1691/commits/47420e9f8c2c1ef2f32cc40a54cff52b50eecd2a)
I only added 2 contract address getters since those are what we need currently, but later on we can add full coverage for all the contracts

hopeyen:
hmmm what happened to the git history of this pr? why did commits from master got put in here?

I think I'll rebase drop these irrelevant once and force-with-lease. apologies to the reviewers

operators/churner/config.go
hopeyen: Sharing that we have address directory deployed on all the networks. I think it is safe to delete the old way now.

hopeyen:
> > hmmm what happened to the git history of this pr? why did commits from master got put in here?
> > I think I'll rebase drop these irrelevant once and force-with-lease. apologies to the reviewers
> 
> I'm not sure what happened with the git history, but IMO it's no big problem if the github UI decides to list commits from `master`. Everything gets squashed when merged

It included all the changes from other commits as well, like the littDB audit commit and subgraph changes. I didn't want them here since the review would get confusing

operators/churner/config.go
hopeyen: removed the legacy method in favor of address directory [4979872](https://github.com/Layr-Labs/eigenda/pull/1691/commits/497987270a67861bc99693023df059d6640f5ef3) (and a few little commits after that for lints and such

tools/semverscan/flags/flags.go
```diff
@@ -16,17 +16,11 @@ const (
...
-		EnvVar:   common.PrefixEnvVar(envPrefix, "BLS_OPERATOR_STATE_RETRIVER"),
-	}
-	EigenDAServiceManagerFlag = cli.StringFlag{
-		Name:     common.PrefixFlag(FlagPrefix, "eigenda-service-manager"),
-		Usage:    "Address of the EigenDA Service Manager",
-		Required: true,
-		EnvVar:   common.PrefixEnvVar(envPrefix, "EIGENDA_SERVICE_MANAGER"),
+	AddressDirectoryFlag = cli.StringFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "address-directory"),
+		Usage:    "Address of the EigenDA Directory contract (preferred over individual contract addresses)",
```
litt3: > (preferred over individual contract addresses)

Stale doc, since individual addresses isn't an option anymore

Comment also applies to the other `flags.go` files

retriever/flags/flags.go
```diff
@@ -32,12 +32,13 @@ var (
...
-		Required: true,
-		EnvVar:   common.PrefixEnvVar(envPrefix, "BLS_OPERATOR_STATE_RETRIVER"),
+	AddressDirectoryFlag = cli.StringFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "address-directory"),
+		Usage:    "Address of the EigenDA Directory contract (preferred over individual contract addresses)",
+		Required: false,
+		EnvVar:   common.PrefixEnvVar(envPrefix, "ADDRESS_DIRECTORY"),
 	}
+	// This flag is kept for retriever's fetchBatchHeader; can later be removed by utilizing AddressDirectoryFlag
 	EigenDAServiceManagerFlag = cli.StringFlag{
```
litt3, line 48: Does `EigenDAServiceManagerFlag` actually need to be kept?

You removed it from the `RetrieverFlags` method below, so it appears to not have any usages right now

api/clients/v2/examples/client_construction.go
```diff
@@ -28,12 +28,11 @@ import (
...
-	addressDirectoryAddress    = "0x0000000000000000000000000000000000000000" // TODO: Replace with actual EigenDADirectory address when deployed
-	// Legacy addresses (use addressDirectoryAddress instead when available)
+	ethRPCURL                        = "https://ethereum-holesky-rpc.publicnode.com"
+	disperserHostname                = "disperser-testnet-holesky.eigenda.xyz"
+	certVerifierRouterAddress        = "0x7F40A8e1B62aa1c8Afed23f6E8bAe0D340A4BC4e"
+	registryCoordinatorAddress       = "0x53012C69A189cfA2D9d29eb6F19B32e0A2EA3490"
+	addressDirectoryAddress          = "0x90776Ea0E99E4c38aA1Efe575a61B3E40160A2FE"
 	blsOperatorStateRetrieverAddress = "0x003497Dd77E5B73C40e8aCbB562C8bb0410320E7"
 	eigenDAServiceManagerAddress     = "0xD4A7E1Bd8015057293f0D0A557088c286942e84b"
 )
```
samlaf: all the other addresses should be removed right? From now all addresses should be fetched from the addressDirectory afaiu. Also contract is actually called EigenDADirectory and not addressDirectory right?
And we should probably have its hardcoded addresses for each env in a common/consts.go file or something, not replicated all over the codebase.

tools/ejections/flags/flags.go
```diff
@@ -47,24 +47,17 @@ var (
...
-		Required: true,
-		EnvVar:   common.PrefixEnvVar(envPrefix, "BLS_OPERATOR_STATE_RETRIVER"),
-	}
-	EigenDAServiceManagerFlag = cli.StringFlag{
-		Name:     common.PrefixFlag(FlagPrefix, "eigenda-service-manager"),
-		Usage:    "Address of the EigenDA Service Manager",
-		Required: true,
-		EnvVar:   common.PrefixEnvVar(envPrefix, "EIGENDA_SERVICE_MANAGER"),
+	AddressDirectoryFlag = cli.StringFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "address-directory"),
```
litt3: There are some discrepancies in flag naming: sometimes flag name is `address-directory`, sometimes `address-directory-addr`

contracts/script/SetUpEigenDA.s.sol
```diff
@@ -189,6 +189,7 @@ contract SetupEigenDA is EigenDADeployer, EigenLayerUtils {
         }
 
         string memory output = "eigenDA deployment output";
+        vm.serializeAddress(output, "eigenDADirectory", address(eigenDADirectory));
```
litt3, line 192: In some places, this is referred to as the `EigenDADirectory`, in other places it's called the `AddressDirectory`, or simply `Directory`

It would probably be good to decide on the terminology, and stick to the same name everywhere

core/eth/utils.go
```diff
@@ -212,22 +212,10 @@ var ContractNames = struct {
...
-func ValidateAddressConfig(addressDirectory, blsOperatorStateRetriever, eigenDAServiceManager string) error {
-	// Check that either address directory is provided OR both individual addresses are provided
-	if addressDirectory == "" && (blsOperatorStateRetriever == "" || eigenDAServiceManager == "") {
-		return fmt.Errorf("either address-directory must be provided, or both bls-operator-state-retriever and eigenda-service-manager addresses must be provided")
-	}
-
-	// Validate address formats
+func ValidateAddressConfig(addressDirectory string) error {
 	if addressDirectory != "" && !gethcommon.IsHexAddress(addressDirectory) {
 		return fmt.Errorf("address-directory must be a valid hex address")
```
samlaf: what about the case when addressDirectory == ""? I think we also want to protect against that. Prob should look like
```
if addressDirectory == "" || !gethcommon.IsHexAddress(addressDirectory) {
   return error
}
```
?

core/eth/utils.go
```diff
@@ -176,3 +178,43 @@ func GetAllQuorumIDs(quorumCount uint8) []core.QuorumID {
...
+	PaymentVault:           "PAYMENT_VAULT",
+	EjectionManager:        "EJECTION_MANAGER",
+	RelayRegistry:          "RELAY_REGISTRY",
+	ThresholdRegistry:      "THRESHOLD_REGISTRY",
+	DisperserRegistry:      "DISPERSER_REGISTRY",
+}
+
+// ValidateAddressConfig validates that an valid address directory is provided
+func ValidateAddressConfig(address string) error {
+	if address != "" && !gethcommon.IsHexAddress(address) {
```
litt3: Do we actually want to permit an empty address?

`ValidateAddressConfig` seems to be a bit misleading- this isn't validating anything about the directory, or the config. It just checks that the input is a hex address. Could this just be checked when the address is actually used, so that we don't need to call `ValidateAddressConfig` in so many places?

core/eth/address_directory_reader.go
```diff
@@ -0,0 +1,76 @@
...
+}
+
+// getAddressWithValidation reads the directory to get an address by the contract name
+// and validates it's not zero
+func (r *AddressDirectoryReader) getAddressWithValidation(contractName string) (gethcommon.Address, error) {
+	if !slices.Contains(r.contractNames, contractName) {
+		return gethcommon.Address{}, fmt.Errorf("contract %s not found in address directory", contractName)
+	}
+
+	addr, err := r.contract.GetAddress0(&bind.CallOpts{}, contractName)
```
litt3: I think all of these reads should have a `ctx` parameter passed in, e.g. `getAddressWithValidation(ctx context.Context, contractName string)`, and then `ctx` gets passed into the call in the `CallOpts`

core/eth/address_directory_reader.go
```diff
@@ -0,0 +1,76 @@
...
+	return addr, nil
+}
+
+// GetAllContractNames returns the names of all contracts in the address directory
+func (r *AddressDirectoryReader) GetAllContractNames() ([]string, error) {
+	names, err := r.contract.GetAllNames(&bind.CallOpts{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to get all contract names: %w", err)
+	}
+	r.contractNames = names
```
litt3: Updating this member variable is a side effect that I would not generally expect from a getter function.

Is this actually necessary? If it is, it should be clearly stated, and the function should probably be renamed.

core/eth/utils.go
hopeyen: updated [7a9aee4](https://github.com/Layr-Labs/eigenda/pull/1691/commits/7a9aee4d121b92296af1b424c62314831b27ad78)

contracts/script/SetUpEigenDA.s.sol
hopeyen: updated [7a9aee4](https://github.com/Layr-Labs/eigenda/pull/1691/commits/7a9aee4d121b92296af1b424c62314831b27ad78)

tools/ejections/flags/flags.go
hopeyen: updated [7a9aee4](https://github.com/Layr-Labs/eigenda/pull/1691/commits/7a9aee4d121b92296af1b424c62314831b27ad78)

tools/semverscan/flags/flags.go
hopeyen: updated [7a9aee4](https://github.com/Layr-Labs/eigenda/pull/1691/commits/7a9aee4d121b92296af1b424c62314831b27ad78)

core/eth/utils.go
hopeyen: makes sense, updated [7a9aee4](https://github.com/Layr-Labs/eigenda/pull/1691/commits/7a9aee4d121b92296af1b424c62314831b27ad78)

api/clients/v2/examples/client_construction.go
hopeyen: they should, but some other components like CertBuilder are still taking them. I think we can clean it all up later
[linear ticket](https://linear.app/eigenlabs/issue/EGDA-1644/remove-all-address-configs-and-only-keep-eigendadirectory)

retriever/flags/flags.go
hopeyen: good catch! we still need the flag, and I'll add the flag back [be3da55](https://github.com/Layr-Labs/eigenda/pull/1691/commits/be3da55f2189a8cfccb738f4309de3099906606d)

core/eth/address_directory_reader.go
hopeyen: this means we will have to add `ctx` to eth.NewReader or create a context there. I think I will leave this improvement for another PR
[ticket](https://linear.app/eigenlabs/issue/EGDA-1645/add-context-to-contract-reads)

core/eth/address_directory_reader.go
hopeyen: removed this in [a01d3cf](https://github.com/Layr-Labs/eigenda/pull/1691/commits/a01d3cfc80d2231c7fd4a4175bfa3050fb5db15a)

core/eth/address_directory_reader.go
litt3: The context would just be a parameter passed into the `getAddressWithValidation` function, so that the caller can properly timeout. Context wouldn't be passed into `NewReader`

api/proxy/config/v2/eigendaflags/cli.go
```diff
@@ -29,6 +29,7 @@ var (
 	)
 	ServiceManagerAddrFlagName        = withFlagPrefix("service-manager-addr")
 	BLSOperatorStateRetrieverFlagName = withFlagPrefix("bls-operator-state-retriever-addr")
+	EigenDADirectoryFlagName          = withFlagPrefix("eigen-da-directory")
```
litt3: you used `eigen-da-directory` here, but the other flags are named `eigenda-directory`,

api/clients/v2/examples/client_construction.go
```diff
@@ -28,12 +28,18 @@ import (
...
 // network, you will need to set these constants to the correct values, based on the chosen network.
 const (
-	ethRPCURL                        = "https://ethereum-holesky-rpc.publicnode.com"
-	disperserHostname                = "disperser-testnet-holesky.eigenda.xyz"
-	certVerifierRouterAddress        = "0x7F40A8e1B62aa1c8Afed23f6E8bAe0D340A4BC4e"
+	ethRPCURL         = "https://ethereum-holesky-rpc.publicnode.com"
+	disperserHostname = "disperser-testnet-holesky.eigenda.xyz"
+
+	// Eth client
+	addressDirectoryAddress = "0x90776Ea0E99E4c38aA1Efe575a61B3E40160A2FE"
```
litt3: here is another case where the term "address directory" is used, instead of "EigenDA Directory"

core/eth/address_directory_reader.go
```diff
@@ -0,0 +1,67 @@
...
+
+	"github.com/Layr-Labs/eigenda/common"
+	eigendadirectory "github.com/Layr-Labs/eigenda/contracts/bindings/IEigenDADirectory"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	gethcommon "github.com/ethereum/go-ethereum/common"
+)
+
+// AddressDirectoryReader wraps the address directory contract and provides
+// safe getters for contract addresses with zero address validation
+type AddressDirectoryReader struct {
```
litt3: This struct and file name use "address directory" terminology, instead of "EigenDA Directory"

core/eth/reader.go
```diff
@@ -65,21 +68,37 @@ var _ core.Reader = (*Reader)(nil)
 func NewReader(
 	logger logging.Logger,
 	client common.EthClient,
-	blsOperatorStateRetrieverHexAddr string,
-	eigenDAServiceManagerHexAddr string) (*Reader, error) {
+	addressDirectoryHexAddr string) (*Reader, error) {
```
litt3: here is another case where the term "address directory" is used, instead of "EigenDA Directory"

core/eth/writer.go
```diff
@@ -37,12 +37,11 @@ var _ core.Writer = (*Writer)(nil)
 func NewWriter(
 	logger logging.Logger,
 	client common.EthClient,
-	blsOperatorStateRetrieverHexAddr string,
-	eigenDAServiceManagerHexAddr string) (*Writer, error) {
+	addressDirectoryHexAddr string) (*Writer, error) {
```
litt3: here is another case where the term "address directory" is used, instead of "EigenDA Directory"

operators/churner/tests/churner_test.go
```diff
@@ -204,7 +204,7 @@ func TestChurner(t *testing.T) {
 	assert.NoError(t, err)
 }
 
-func createTransactorFromScratch(privateKey, operatorStateRetriever, serviceManager string, logger logging.Logger) (*eth.Writer, error) {
+func createTransactorFromScratch(privateKey, addressDirectory string, logger logging.Logger) (*eth.Writer, error) {
```
litt3: here is another case where the term "address directory" is used, instead of "EigenDA Directory"

relay/cmd/flags/flags.go
```diff
@@ -202,17 +202,11 @@ var (
...
-		Usage:    "Address of the BLS operator state retriever",
-		Required: true,
-		EnvVar:   common.PrefixEnvVar(envVarPrefix, "BLS_OPERATOR_STATE_RETRIEVER_ADDR"),
-	}
-	EigenDAServiceManagerAddrFlag = cli.StringFlag{
-		Name:     common.PrefixFlag(FlagPrefix, "eigen-da-service-manager-addr"),
-		Usage:    "Address of the Eigen DA service manager",
-		Required: true,
-		EnvVar:   common.PrefixEnvVar(envVarPrefix, "EIGEN_DA_SERVICE_MANAGER_ADDR"),
+	AddressDirectoryFlag = cli.StringFlag{
```
litt3: here is another case where the term "address directory" is used, instead of "EigenDA Directory"

litt3 (approved):

api/proxy/config/v2/eigendaflags/cli.go
hopeyen: updated [1ea1ff4](https://github.com/Layr-Labs/eigenda/pull/1691/commits/1ea1ff409d84de9069e90c620c635bed7443be0f)

operators/churner/tests/churner_test.go
hopeyen: [1ea1ff4](https://github.com/Layr-Labs/eigenda/pull/1691/commits/1ea1ff409d84de9069e90c620c635bed7443be0f)

relay/cmd/flags/flags.go
hopeyen: [1ea1ff4](https://github.com/Layr-Labs/eigenda/pull/1691/commits/1ea1ff409d84de9069e90c620c635bed7443be0f)

core/eth/writer.go
hopeyen: [4b8e39f](https://github.com/Layr-Labs/eigenda/pull/1691/commits/4b8e39f8df95d1d48303df6ac5e0ccf1fe226fc7)

core/eth/reader.go
hopeyen: [4b8e39f](https://github.com/Layr-Labs/eigenda/pull/1691/commits/4b8e39f8df95d1d48303df6ac5e0ccf1fe226fc7)

core/eth/address_directory_reader.go
hopeyen: [4b8e39f](https://github.com/Layr-Labs/eigenda/pull/1691/commits/4b8e39f8df95d1d48303df6ac5e0ccf1fe226fc7)

api/clients/v2/examples/client_construction.go
hopeyen: [4b8e39f](https://github.com/Layr-Labs/eigenda/pull/1691/commits/4b8e39f8df95d1d48303df6ac5e0ccf1fe226fc7)

pakim249CAL (approved):

api/proxy/config/v2/eigendaflags/cli.go
```diff
@@ -29,6 +29,7 @@ var (
 	)
 	ServiceManagerAddrFlagName        = withFlagPrefix("service-manager-addr")
 	BLSOperatorStateRetrieverFlagName = withFlagPrefix("bls-operator-state-retriever-addr")
```
samlaf, line 31: think we should mark these as deprecated and hint people to start moving to the eigenda-directory? Prob want to do it across all of our services (apiserver, nodes, etc). Everything in this PR. And eventually we'll remove all other address flags and only keep the eigenda-directory.

not sure if urfave has a way to mark flags as deprecated. If not then we might need to manually log a warn. cc @dmanc who will be updating our helm charts to use these new flags. Can you also add this as a note to this PR description? This kind of external config changes are very important to highlight.

hopeyen: i made a linear [ticket](https://linear.app/eigenlabs/issue/EGDA-1647/proxy-eigenda-config-flag-deprecate-legacy-addresses) for proxy, and for all services [ticket](https://linear.app/eigenlabs/issue/EGDA-1644/remove-all-address-configs-and-only-keep-eigendadirectory). I will add notes to the PR description

core/eth/address_directory_reader.go
samlaf: agree here, context should be passed in to function calls, not be stored in structs. https://go.dev/blog/context-and-structs

retriever/flags/flags.go
```diff
@@ -32,12 +32,13 @@ var (
 		Required: true,
 		EnvVar:   common.PrefixEnvVar(envPrefix, "TIMEOUT"),
 	}
-	BlsOperatorStateRetrieverFlag = cli.StringFlag{
-		Name:     common.PrefixFlag(FlagPrefix, "bls-operator-state-retriever"),
-		Usage:    "Address of the BLS Operator State Retriever",
-		Required: true,
-		EnvVar:   common.PrefixEnvVar(envPrefix, "BLS_OPERATOR_STATE_RETRIVER"),
```
samlaf: don't remove this flag, since thats a breaking change. instead mark it as deprecated and log a warning when its used, prompting the user to use the new flags.

retriever/flags/flags.go
samlaf: you didn't add the flag back, just the config. let's not delete flags until we're ready for the breaking change (which should be announced in advance).

relay/cmd/flags/flags.go
```diff
@@ -202,17 +202,11 @@ var (
...
-		Name:     common.PrefixFlag(FlagPrefix, "bls-operator-state-retriever-addr"),
-		Usage:    "Address of the BLS operator state retriever",
-		Required: true,
-		EnvVar:   common.PrefixEnvVar(envVarPrefix, "BLS_OPERATOR_STATE_RETRIEVER_ADDR"),
-	}
-	EigenDAServiceManagerAddrFlag = cli.StringFlag{
-		Name:     common.PrefixFlag(FlagPrefix, "eigen-da-service-manager-addr"),
-		Usage:    "Address of the Eigen DA service manager",
-		Required: true,
-		EnvVar:   common.PrefixEnvVar(envVarPrefix, "EIGEN_DA_SERVICE_MANAGER_ADDR"),
```
samlaf: add these flags back, we don't want to create breaking changes. Every binary should be backwards compatible with these old flags. you can mark them as deprecated but not delete.

operators/churner/flags/flags.go
```diff
@@ -32,17 +32,11 @@ var (
...
-		Name:     common.PrefixFlag(FlagPrefix, "eigenda-service-manager"),
-		Usage:    "Address of the EigenDA Service Manager",
-		Required: true,
-		EnvVar:   common.PrefixEnvVar(envPrefix, "EIGENDA_SERVICE_MANAGER"),
+	EigenDADirectoryFlag = cli.StringFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "eigenda-directory"),
+		Usage:    "Address of the EigenDA Address Directory",
+		Required: false,
+		EnvVar:   common.PrefixEnvVar(envPrefix, "EIGENDA_DIRECTORY"),
 	}
```
samlaf, line 52: dont delete these flags. see my other comments.

node/flags/flags.go
```diff
@@ -156,17 +156,11 @@ var (
...
-		Name:     common.PrefixFlag(FlagPrefix, "bls-operator-state-retriever"),
-		Usage:    "Address of the BLS Operator State Retriever",
-		Required: true,
-		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "BLS_OPERATOR_STATE_RETRIVER"),
-	}
-	EigenDAServiceManagerFlag = cli.StringFlag{
-		Name:     common.PrefixFlag(FlagPrefix, "eigenda-service-manager"),
-		Usage:    "Address of the EigenDA Service Manager",
-		Required: true,
-		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "EIGENDA_SERVICE_MANAGER"),
```
samlaf: please do not ever do this. this will break users and they will scream at us. flag breaking changes should be announced way ahead of time, and should have been deprecated for a while, and any PRs removing flags (breaking API change!) should have its title with an `!` like `feat!: eth client address directory` to mark the breaking change (to be clear we should not do this right now, but the ! is useful to flag to the release captain that a MAJOR version bump is needed)

samlaf (changes requested):
Please make the binaries backwards compatible again by keeping the old ServiceManager flag around, and allowing binaries to be started with those flags. We don't want to break people.

api/clients/v2/examples/client_construction.go
```diff
@@ -28,18 +28,13 @@ import (
...
-	// Router address provider
-	certVerifierRouterAddress = "0x7F40A8e1B62aa1c8Afed23f6E8bAe0D340A4BC4e"
-
-	// Cert builder
+	ethRPCURL                        = "https://ethereum-holesky-rpc.publicnode.com"
+	disperserHostname                = "disperser-testnet-holesky.eigenda.xyz"
+	certVerifierRouterAddress        = "0x7F40A8e1B62aa1c8Afed23f6E8bAe0D340A4BC4e"
 	registryCoordinatorAddress       = "0x53012C69A189cfA2D9d29eb6F19B32e0A2EA3490"
 	blsOperatorStateRetrieverAddress = "0x003497Dd77E5B73C40e8aCbB562C8bb0410320E7"
+	eigenDAServiceManagerAddress     = "0xD4A7E1Bd8015057293f0D0A557088c286942e84b"
```
samlaf: nit: examples should use SOTA stuff, so should only use the eigendaDirectoryAddress.

api/proxy/config/v2/eigendaflags/cli.go
```diff
@@ -119,18 +119,25 @@ func CLIFlags(envPrefix, category string) []cli.Flag {
 		},
 		&cli.StringFlag{
 			Name:     ServiceManagerAddrFlagName,
-			Usage:    "Address of the EigenDA service manager contract.",
+			Usage:    "[Deprecating soon in favor of EigenDADirectory] Address of the EigenDA Service Manager contract.",
```
samlaf:
```suggestion
			Usage:    "[Deprecated: use EigenDADirectory instead] Address of the EigenDA Service Manager contract.",
```

api/proxy/config/v2/eigendaflags/cli.go
```diff
@@ -119,18 +119,25 @@ func CLIFlags(envPrefix, category string) []cli.Flag {
...
 			Name:     BLSOperatorStateRetrieverFlagName,
-			Usage:    "Address of the BLS operator state retriever contract.",
+			Usage:    "[Deprecating soon in favor of EigenDADirectory] Address of the BLS operator state retriever contract.",
 			EnvVars:  []string{withEnvPrefix(envPrefix, "BLS_OPERATOR_STATE_RETRIEVER_ADDR")},
 			Category: category,
 			Required: false,
 		},
+		&cli.StringFlag{
+			Name:     EigenDADirectoryFlagName,
+			Usage:    "Address of the EigenDA directory contract.",
```
samlaf:
```suggestion
			Usage:    "Address of the EigenDA directory contract, which points to all other EigenDA contract addresses. This is the only contract entrypoint needed offchain.",
```

disperser/cmd/apiserver/flags/flags.go
```diff
@@ -48,14 +48,14 @@ var (
 	}
 	BlsOperatorStateRetrieverFlag = cli.StringFlag{
 		Name:     common.PrefixFlag(FlagPrefix, "bls-operator-state-retriever"),
-		Usage:    "Address of the BLS Operator State Retriever",
-		Required: true,
+		Usage:    "[Deprecating soon in favor of EigenDADirectory] Address of the BLS operator state Retriever",
```
samlaf:
```suggestion
		Usage:    "[Deprecated: use EigenDADirectory instead] Address of the BLS operator state Retriever",
```

samlaf: It is already deprecated. Deprecated means there's a better way to do things and we're giving you some time to switch over to the new method before we delete this old method.

It's already deprecated, and will eventually get deleted. But better not to use relative language like "soon" or "eventually" because thats fuzzy and unuseful.

samlaf: Plesae change everywhere

samlaf (dismissed):
LGTM except for the wording of the deprecated notice. Please change everywhere and then we can merge.

api/clients/v2/examples/client_construction.go
hopeyen: `blsOperatorStateRetrieverAddress` is still used for CertBuilder, that's not changed within this PR. 
I will update `eigenDAServiceManagerAddress` to an empty string because new EthClient take all three arguments

samlaf (approved):

PR #1694, opened 2025-06-30
feat: AccountantLedger interface for next step unifications

hopeyen:

- Create AccountLedger interface for payment state management
- Add LocalAccountLedger implementation for client-side tracking
- Refactor Accountant to use unified AccountLedger interface

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


core/meterer/accountledger.go
```diff
@@ -0,0 +1,364 @@
...
+	//   params: Payment vault parameters containing quorum configurations
+	//
+	// Returns:
+	//   nil if the usage was successfully recorded
+	//   error if the usage cannot be accommodated:
+	//     - "reservation not found for quorum X": No reservation exists for quorum
+	//     - "reservation limit exceeded": Usage would exceed rate limits
+	//     - "reservation expired": Current time is outside reservation window
+	//     - "quorum config not found for X": Missing protocol configuration
+	RecordReservationUsage(
```
litt3: Is there a reason that `RecordReservationUsage` and `RecordOnDemandUsage` should be separate methods on the interface?

It seems that a single method accepting details about a payment would be sufficient- I don't see why you would ever handle one and not the other

core/meterer/accountledger.go
```diff
@@ -0,0 +1,364 @@
...
+	//   - Synchronization: Extracting state for storage or transmission
+	//   - Testing: Verifying state after operations
+	//   - Debugging: Analyzing account state for troubleshooting
+	//
+	// Returns:
+	//   AccountState: A deep copy of the current account state including:
+	//     - All reservation settings and their current usage
+	//     - On-demand payment balance and cumulative usage
+	//     - Complete period records history
+	GetAccountState() AccountState
```
litt3: The reason I can think of to export the account state would be to send it over the wire, like when the disperser helps the client initialize. 

In that case, we would need to create a protobuf. Then we could have a `GetAccountStateProtobuf` method on this interface, factory methods which accept the protobuf to create a new account ledger object, and we could avoid having to create an `AccountState` golang object.

core/meterer/accountledger.go
```diff
@@ -0,0 +1,364 @@
...
+	//   - Testing: Setting up known state for test scenarios
+	//
+	// Parameters:
+	//   state: The complete account state to set. Nil fields will be initialized
+	//          with appropriate zero values (empty maps, zero big.Int, etc.)
+	//
+	// Thread Safety:
+	// This method is safe for concurrent use. It acquires an exclusive write lock
+	// for the entire operation to ensure atomic state replacement.
+	SetAccountState(state AccountState)
```
litt3: Why would we need to set the account state on an already created account ledger? Couldn't we just discard the old account ledger and create a new one with the new data?

core/meterer/accountledger.go
```diff
@@ -0,0 +1,364 @@
...
+// including both on-chain settings and local usage tracking.
+//
+// This struct serves as the canonical representation of all payment-related
+// data for a single account, enabling atomic state transitions and consistent
+// snapshots for synchronization between different system components.
+//
+// State Categories:
+//  1. On-chain Settings: Immutable data from blockchain (reservations, on-demand)
+//  2. Local Tracking: Mutable usage counters maintained locally (period records, cumulative payment)
+type AccountState struct {
```
litt3: IMO it helps keep file sizes more digestible if structs and interfaces are given their own file

api/clients/v2/accountant.go
```diff
@@ -1,113 +1,71 @@
...
 )
 
 type Accountant struct {
-	// on-chain states
-	accountID    gethcommon.Address
-	reservations map[core.QuorumID]*core.ReservedPayment
-	// OnDemand is initially only enabled on quorum 0. Accountant must be updated to be quorum specific
-	// after the protocol decides to support onDemand on custom quorums and decentralized ratelimiting.
-	onDemand *core.OnDemandPayment
+	// account identification
```
litt3: I'm curious why you are wrapping the new `AccountLedger` inside of the accountant. 

The original idea was for the `AccountLedger` to replace the accountant. If we just leave the current `Accountant` as-is for now, then we are able develop the new payment classes without having any impact on the current system. And when we're ready, we can just switch over.

core/meterer/accountledger.go
litt3, line 1: file name should use snake case `account_ledger`

core/meterer/accountledger.go
hopeyen: I see, okay. do you mind reviewing the debit slip PR first, and I'll address these comments in my tmr morning?

core/meterer/accountledger.go
hopeyen: updated [a5581f8](https://github.com/Layr-Labs/eigenda/pull/1694/commits/a5581f871327b47cc67b1a5d670962a9dcc4fe3f)

api/clients/v2/accountant.go
hopeyen: was eager to test out the functionality of LocalAccountLedger; I've now reverted leave `Accountant` as-is for now [a5581f8](https://github.com/Layr-Labs/eigenda/pull/1694/commits/a5581f871327b47cc67b1a5d670962a9dcc4fe3f)

core/meterer/accountledger.go
hopeyen: put in separate files [a5581f8](https://github.com/Layr-Labs/eigenda/pull/1694/commits/a5581f871327b47cc67b1a5d670962a9dcc4fe3f)

core/meterer/accountledger.go
hopeyen: good point, removed [a5581f8](https://github.com/Layr-Labs/eigenda/pull/1694/commits/a5581f871327b47cc67b1a5d670962a9dcc4fe3f)

core/meterer/accountledger.go
hopeyen: updated [a5581f8](https://github.com/Layr-Labs/eigenda/pull/1694/commits/a5581f871327b47cc67b1a5d670962a9dcc4fe3f)

core/meterer/accountledger.go
hopeyen: updated to combine both functions into `debit` [a5581f8](https://github.com/Layr-Labs/eigenda/pull/1694/commits/a5581f871327b47cc67b1a5d670962a9dcc4fe3f)

core/meterer/accountledger.go
hopeyen: I decided to remove the LocalAccountLedger from this PR so the only thing it contains now is the `AccountLedger` interface definition. Hopefully the review is easier this way. 
The LocalAccountLedger is split to https://github.com/Layr-Labs/eigenda/pull/1712

litt3 (approved):
LGTM

PR #1695, opened 2025-06-30
feat: unify accountant debit slip

hopeyen:

- New DebitSlip struct encapsulates PaymentMetadata, NumSymbols, QuorumNumbers, and metadata like ReceivedAt and RequestID
- Built-in validation methods and helper functions for type conversions
- Meterer: MeterRequest now accepts DebitSlip instead of 5 separate parameters
- Accountant: reservationUsage and onDemandUsage methods streamlined to use DebitSlip

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


core/meterer/meterer.go
```diff
@@ -74,30 +74,35 @@ func (m *Meterer) Start(ctx context.Context) {
 	}()
 }
 
-// MeterRequest validates a blob header and adds it to the meterer's state
+// MeterRequest validates a debit slip and adds it to the meterer's state
 // TODO: return error if there's a rejection (with reasoning) or internal error (should be very rare)
-func (m *Meterer) MeterRequest(ctx context.Context, header core.PaymentMetadata, numSymbols uint64, quorumNumbers []uint8, receivedAt time.Time) (uint64, error) {
-	m.logger.Info("Validating incoming request's payment metadata", "paymentMetadata", header, "numSymbols", numSymbols, "quorumNumbers", quorumNumbers)
+func (m *Meterer) MeterRequest(ctx context.Context, debitSlip *DebitSlip) (uint64, error) {
```
litt3: This is another case where I think it makes most sense to leave the current implementation alone for now, so that the new system can be developed without risk of breaking the old one

core/meterer/debit_slip.go
```diff
@@ -0,0 +1,100 @@
...
+	PaymentMetadata core.PaymentMetadata
+
+	// Blob characteristics - number of symbols in the blob
+	NumSymbols uint64
+
+	// Target quorums for blob dispersal
+	QuorumNumbers []core.QuorumID
+
+	// Timing information (for server-side processing)
+	ReceivedAt time.Time // Optional, set by server when request is received
```
litt3: Since `ReceivedAt` isn't relevant for actually debiting an account, I think it might be better for it live elsewhere. 

What does the server use this for?

core/meterer/debit_slip.go
```diff
@@ -0,0 +1,100 @@
...
+	return &DebitSlip{
+		PaymentMetadata: paymentMetadata,
+		NumSymbols:      numSymbols,
+		QuorumNumbers:   quorumNumbers,
+		ReceivedAt:      time.Now(),
+	}
+}
+
+// Validate performs basic validation on the request parameters
+func (ds *DebitSlip) Validate() error {
```
litt3: Could we just do validation inside `NewDebitSlip`, and avoid needing a separate `Validate` method?

core/meterer/debit_slip.go
```diff
@@ -0,0 +1,100 @@
...
+	return ds.PaymentMetadata.AccountID
+}
+
+// GetTimestamp returns the timestamp from the payment metadata
+func (ds *DebitSlip) GetTimestamp() int64 {
+	return ds.PaymentMetadata.Timestamp
+}
+
+// GetQuorumNumbersAsUint8 returns quorum numbers as uint8 slice for backward compatibility
+func (ds *DebitSlip) GetQuorumNumbersAsUint8() []uint8 {
```
litt3: > for backward compatibility

Curious what backward compatibility you're referring to. Do you have something in mind for this?

hopeyen: some places take quorum ID as uint8, some places use the alias QuorumID type, this is just a convenience helper

core/meterer/debit_slip.go
hopeyen: server needs to make sure timestamp included by the client in the `PaymentHeader` is within some time range of when the server received it at. It is kind of like, the client issued the header at time $t$, so server must debit the account within $t+c$ where $c$ is a constant

core/meterer/meterer.go
hopeyen: i removed the changes [ac17b64](https://github.com/Layr-Labs/eigenda/pull/1695/commits/ac17b64114afedf17d1e052f40ffb2b112826860)

core/meterer/debit_slip.go
hopeyen: yes, updated [ac17b64](https://github.com/Layr-Labs/eigenda/pull/1695/commits/ac17b64114afedf17d1e052f40ffb2b112826860)

core/meterer/debit_slip.go
litt3: It sounds like this could just be checked by the server when the request is ingested, without needing to be a struct field.

(an optional field which is only relevant for some use-cases is a sign that the field belongs elsewhere, IMO)

core/meterer/debit_slip_test.go
```diff
@@ -0,0 +1,147 @@
...
+	"testing"
+	"time"
+
+	"github.com/Layr-Labs/eigenda/core"
+	gethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDebitSlip_NewDebitSlip(t *testing.T) {
```
litt3: Putting the `DebitSlip` struct in the `meterer` package means that the `TestMain` function defined in `meterer_test` tries to connect to localstack, even when just trying to run these pure unit tests.

I think we should but these new payment structs in a different package, though I haven't spent the time to come up with an opinion on where the best place would be. I recall @samlaf commented on this on a different PR, but I don't recall the specifics of that conversation?

core/meterer/debit_slip_test.go
```diff
@@ -0,0 +1,147 @@
...
+	tests := []struct {
+		name            string
+		paymentMetadata core.PaymentMetadata
+		numSymbols      uint64
+		quorumNumbers   []core.QuorumID
+		expectedError   string
```
litt3: Matching error strings is a fragile way to write tests, and doesn't really improve the correctness guarantees IMO

core/meterer/debit_slip_test.go
hopeyen: based on sam's comment, I made this [refactor](https://github.com/Layr-Labs/eigenda/pull/1705) PR. I don't think it is particularly essential for the decentralized metering project, so please review as a secondary concern

core/meterer/debit_slip_test.go
hopeyen: added error types in [3a33ec7](https://github.com/Layr-Labs/eigenda/pull/1695/commits/3a33ec7cd3efef1e4e505380dc82952a04b953ce) if this type of test is more acceptable

core/meterer/debit_slip.go
hopeyen: that can work too, I removed the field in [3a33ec7](https://github.com/Layr-Labs/eigenda/pull/1695/commits/3a33ec7cd3efef1e4e505380dc82952a04b953ce)

core/meterer/debit_slip_test.go
litt3: Why not put new files into the "correct" place, instead of continuing to put them in `meterer`, and waiting for the refactor PR?

hopeyen: just trying to separate concerns in parallel PRs, but I've moved the new files to `core/payment`. I think we can move other files into `payment` in a separate PR. Thinking at least `payment_logic`, `period_record`, and other `account_ledger` impls can live in `payment` instead of `meterer`. `meterer` will probably become specific to disperser server 
[d49e1eb](https://github.com/Layr-Labs/eigenda/pull/1695/commits/d49e1eb3998a5310b715f4d7eaff840b8a2fe1b7)

litt3 (approved):

PR #1697, opened 2025-06-30
feat: enumerable names and script configs for address registry

pakim249CAL:

Adds enumerability and a reverse string mapping for names. Also adds configurations for the script that deploys the address registry

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


contracts/src/core/interfaces/IEigenDADirectory.sol
```diff
@@ -35,4 +35,10 @@ interface IEigenDADirectory {
 
     /// @notice Gets the address by name.
     function getAddress(string memory name) external view returns (address);
+
+    /// @notice Gets the name by keccak256 hash of the name.
+    function getName(bytes32 key) external view returns (string memory);
```
samlaf, line 40: what is this function useful for? Who would want to call this?

contracts/script/deploy/eigenda-directory/config/holesky.preprod.config.toml
```diff
@@ -0,0 +1,42 @@
...
+    "REGISTRY_COORDINATOR",
+    "EJECTION_MANAGER", 
+    "SERVICE_MANAGER",
+    "OPERATOR_STATE_RETRIEVER",
+    "THRESHOLD_REGISTRY",
+    "RELAY_REGISTRY",
+    "DISPERSER_REGISTRY",
+    "PAYMENT_VAULT",
+    "CERT_VERIFIER"
+]
```
samlaf, line 18: don't we want the same contractNames for all envs? Feel like these should be defined in a single `shared.config.toml` file, instead of being duplicated everywhere (which is bug-prone and will likely lead to inconsistencies)

contracts/script/deploy/eigenda-directory/config/holesky.preprod.config.toml
```diff
@@ -0,0 +1,42 @@
...
+# This field should be populated with the addresses of the contracts in the same order as in contractNames.
+contractAddresses = [
+    "0x9Ab2FEAf0465f0eD51Fc2b663eF228B418c9Dad1", # PAUSER_REGISTRY
+    "0xAd7f9e558170a149Ca8E90f41Ab2444A5d3bd6aD", # BLS_APK_REGISTRY
+    "0x8cE5F2a53cBd29710eb94A04e40C07A4DdF15d10", # INDEX_REGISTRY
+    "0x53668EBf2e28180e38B122c641BC51Ca81088871", # STAKE_REGISTRY
+    "0x1747ef24dbbb52cB06382d323f455D48dE1AC7fd", # SOCKET_REGISTRY
+    "0x2c61EA360D6500b58E7f481541A36B443Bc858c6", # REGISTRY_COORDINATOR
+    "0xEEed164605833263B7690bB521983b777D128698", # EJECTION_MANAGER
+    "0x54A03db2784E3D0aCC08344D05385d0b62d4F432", # SERVICE_MANAGER
```
samlaf, line 29: I'm always worried when I see lists like this with implicit ordering/alignment assumptions. Feel like it would be better to have a more resilient scheme, such as encoding a map (maybe list of tuples name-address). Even if we're duplicating the name, that's just another way to protect outselves from misconfigurations. The script will make sure that the name is part of the `shared.config.toml` list of predefined and accepted contract names, and if not, will error out.

contracts/src/core/interfaces/IEigenDADirectory.sol
pakim249CAL: contracts to save gas by not needing to hash. the hash can be provided as a constant value.

contracts/script/deploy/eigenda-directory/config/holesky.preprod.config.toml
pakim249CAL: Yes, I agree. I am just not really sure of the most sensible way to do that yet, as I was just trying to get this out the door so that it could be integrated with. Considering that I have deployed the directories with this commit, I would like to tag this commit with a contract release, merge it, and then address this particular issue in another PR. Would that be acceptable?

https://github.com/Layr-Labs/eigenda/issues/1704

contracts/script/deploy/eigenda-directory/config/holesky.preprod.config.toml
pakim249CAL: I wanted to structify it, but I was told that this was the better way because the named fields with prefixes are too confusing 😅

contracts/src/core/interfaces/IEigenDADirectory.sol
samlaf: why/when would a contract need to get the name of a contract from key?

samlaf (approved):

contracts/script/deploy/eigenda-directory/config/holesky.preprod.config.toml
litt3: > but I was told that this was the better way

Where was this discussion had? I agree that the this multi-list strategy is brittle

contracts/src/core/interfaces/IEigenDADirectory.sol
pakim249CAL: Ah, sorry I misunderstood. The reason is just to create a complete API that can fetch everything you would need to know about the contract efficiently. This way, if the key is ever encountered in the wild (like in another contract) we can always look up on etherscan what the name of that contract is.

PR #1698, opened 2025-06-30
feat: Create CLAUDE.md

litt3:
closes DAINT-578

Takes inspiration from:
- https://github.com/julep-ai/julep/blob/main/AGENTS.md
- https://diwank.space/field-notes-from-shipping-real-code-with-claude

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -21,7 +21,9 @@ this list.
 
 ---
 
-## 2. File imports
+## 2. Project Structure
+
+### 2.1 File Imports
 
 NOTE: Be aware that whatever you add to this list is automatically loaded into context (due to `@` annotation). It's
 helpful to provide project context, but only within reasonable limits.
@@ -30,12 +32,27 @@ helpful to provide project context, but only within reasonable limits.
 2. @go.mod describes golang dependencies
 3. @mise.toml describes external tool dependencies
 4. @.golangci.yml contains linting configuration
-5. @docs/CLAUDE.md causes doc files to be automatically imported
-6. @.claude/commands/CLAUDE.md defines what project slash commands are available to use
+5. @.claude/commands/CLAUDE.md defines what project slash commands are available to use
 
 If there are imports that are relevant only to a particular part of the project, then they should be added to a
-CLAUDE.md file *in the relevant subdirectory*. Then the imports will only be processed when files within that
-directory are read.
+CLAUDE.md file *in the relevant subdirectory*.
+
+### 2.2 Project Subdirectories
+
+1. **Always check for `CLAUDE.md` files in specific directories** before working on code within them. These files
+   contain targeted context.
+2. If a directory's `CLAUDE.md` is outdated or incorrect, **update it**.
+3. If you make significant changes to a directory's structure, patterns, or critical implementation details,
+   **document these in its `CLAUDE.md`**.
+4. If a directory lacks a `CLAUDE.md` but contains complex logic or patterns worth documenting for AI/humans,
+   **suggest creating one**.
+5. Use `@` annotation within CLAUDE.md files to automatically load in helpful context, e.g. `@docs/submoduleDocs`.
+   These imports will be automatically processed whenever the `CLAUDE.md` file is read.
+6. If there is domain-specific terminology relevant to a directory, consider adding a small glossary of terms.
+
+| Subdirectory | Description                                         |
+|--------------|-----------------------------------------------------|
+| ./docs       | Documentation files describing the EigenDA system.  |
 
 ---
 
@@ -102,22 +119,7 @@ directory are read.
 
 ---
 
-## 5. Directory-Specific CLAUDE.md Files
-
-1. **Always check for `CLAUDE.md` files in specific directories** before working on code within them. These files
-   contain targeted context.
-2. If a directory's `CLAUDE.md` is outdated or incorrect, **update it**.
-3. If you make significant changes to a directory's structure, patterns, or critical implementation details,
-   **document these in its `CLAUDE.md`**.
-4. If a directory lacks a `CLAUDE.md` but contains complex logic or patterns worth documenting for AI/humans,
-   **suggest creating one**.
-5. Use `@` annotation within CLAUDE.md files to automatically load in helpful context, e.g. `@docs/submoduleDocs`.
-   These imports will be automatically processed whenever the `CLAUDE.md` file is read.
-6. If there is domain-specific terminology relevant to a directory, consider adding a small glossary of terms.
-
----
-
-## 6. Common pitfalls
+## 5. Common pitfalls
 
 1. Forgetting to run `go mod tidy` after adding new dependencies.
 2. Not linting before committing code.
@@ -127,7 +129,7 @@ directory are read.
 
 ---
 
-## 7. Files to NOT modify
+## 6. Files to NOT modify
 
 These files and directories should generally not be modified without explicit permission:
 
@@ -144,40 +146,38 @@ These files and directories should generally not be modified without explicit pe
 
 ---
 
-## 8. AI Assistant Workflow: Step-by-Step Methodology
+## 7. AI Assistant Workflow: Step-by-Step Methodology
 
-When responding to user instructions, the AI assistant (Claude, Cursor, GPT, etc.) should follow this process to
-ensure clarity, correctness, and maintainability:
+When responding to user instructions, the AI assistant should follow this process to ensure clarity, correctness, and
+maintainability:
 
 1. **Only take action with sufficient context**: Do not make changes or use tools if unsure about something
-   project-specific, or without having context for a particular feature/decision. 
-2. **Consult Relevant Guidance**: When the user gives an instruction, consult the relevant instructions from
-   `CLAUDE.md` files (both root and directory-specific) for the request.
-3. **Clarify Ambiguities**: Based on what you could gather, see if there's any need for clarifications. If so, ask
-   the user targeted questions before proceeding.
-4. **Break Down & Plan**: Break down the task at hand and chalk out a rough plan for carrying it out, referencing
+   project-specific, or without having context for a particular feature/decision.
+2. **Clarify Ambiguities**: If there's any need for clarifications. If so, ask the user targeted questions before
+   proceeding.
+3. **Break Down & Plan**: Break down the task at hand and chalk out a rough plan for carrying it out, referencing
    project conventions and best practices.
-5. **Trivial Tasks**: If the plan/request is trivial, go ahead and get started immediately.
-6. **Non-Trivial Tasks**: Otherwise, present the plan to the user for review and iterate based on their feedback.
-7. **Track Progress**: Use a to-do list (internally, or optionally in a `TODOS.md` file) to keep track of your
+4. **Trivial Tasks**: If the plan/request is trivial, go ahead and get started immediately.
+5. **Non-Trivial Tasks**: Otherwise, present the plan to the user for review and iterate based on their feedback.
+6. **Track Progress**: Use a to-do list (internally, or optionally in a `TODOS.md` file) to keep track of your
    progress on multi-step or complex tasks.
-8. **If Stuck, Re-plan**: If you get stuck or blocked, return to step 3 to re-evaluate and adjust your plan.
-9. **Check for related updates**: Once the user's request is fulfilled, look for any complementary changes that
+7. **If Stuck, Re-plan**: If you get stuck or blocked, return to step 3 to re-evaluate and adjust your plan.
+8. **Check for related updates**: Once the user's request is fulfilled, look for any complementary changes that
    need to be made:
    - Code documentation / doc files that reference details that have been modified
    - Variable names that need to be updated
    - Error messages that use old terminology
    - Related functions / structures that should be renamed to match new changes
    - Links contained in documentation that were broken by the changes
-10. **Lint**: Make sure changes pass linting, and that they adhere to style and coding standards
-11. **Test**: Run tests related to the changes that have been made. Short tests should always be run, but ask
+9. **Lint**: Make sure changes pass linting, and that they adhere to style and coding standards
+10. **Test**: Run tests related to the changes that have been made. Short tests should always be run, but ask
    permission before trying to run long tests.
-12. **User Review**: After completing the task, ask the user to review what you've done, and repeat the process as
+11. **User Review**: After completing the task, ask the user to review what you've done, and repeat the process as
    needed.
-13. **Session Boundaries**: If the user's request isn't directly related to the current context and can be safely
+12. **Session Boundaries**: If the user's request isn't directly related to the current context and can be safely
    started in a fresh session, suggest starting from scratch to avoid context confusion.
 
-## 9. AI Assistant User Interactions
+## 8. AI Assistant User Interactions
 
 1. Prioritize **frankness** and **accuracy** over simply attempting the please a human. In the end, humans are most
    pleased when they receive **honest** and **direct** answers to their prompts. Being a "yes man" negatively impacts
--- a/docs/CLAUDE.md
+++ b/docs/CLAUDE.md
@@ -1,15 +1,7 @@
 # CLAUDE.md - EigenDA Documentation
 
-The `docs` directory contains documentation files describing the EigenDA system. This CLAUDE.md file causes all doc files to be read into 
-context immediately. This isn't a strategy that should be followed for every CLAUDE.md file, but it makes sense for docs specifically,
-since they are comparatively small and greatly improve Claude's ability to understand the repository.
-
-## File Imports
-
-1. @spec/src/glossary.md contains a glossary with basic EigenDA terms
-2. @spec/src/introduction.md contains a basic introduction to the EigenDA system
-3. @spec/src/integration/spec.md describes basic integration data flows
-4. @spec/src/integration/spec/1-apis.md describes important APIs relevant to integrations
-5. @spec/src/integration/spec/2-rollup-payload-lifecycle.md describes how user data moves through the EigenDA system
-6. @spec/src/integration/spec/6-secure-integration.md contains considerations relevant to secure integrations
-7. @release/release-process.md describes the EigenDA team's release processes
+| Subdirectory  | Description                                                                   |
+|---------------|-------------------------------------------------------------------------------|
+| ./spec        | Spec mdBook, containing detailed descriptions of architecture and algorithms  |
+| ./release     | Descriptions of EigenDA release process                                       |
+| ./audits      | Audit reports                                                                 |


CLAUDE.md
```diff
@@ -0,0 +1,149 @@
...
+
+---
+
+## 1. Non-negotiable golden rules
+
+| #:  | AI *must NOT* do                                                                                                                                   |
+|-----|----------------------------------------------------------------------------------------------------------------------------------------------------|
+| G-0 | Write changes or use tools when you are not sure about something project-specific, or if you don't have context for a particular feature/decision. |
+| G-1 | Touch test files, specs, or generated files without EXPLICIT permission. **Humans own tests and specs.**                                           |
+| G-2 | Refactor large modules without human guidance. For changes >300 LOC or >3 files, **ask for confirmation**.                                         |
```
ethenotethan: Hmm I'd argue Claude should by default autoprompt for any individual file changes >50 lines

CLAUDE.md
```diff
@@ -0,0 +1,149 @@
...
+| @.claude/commands/*              | Project claude commands              |
+
+If there are imports that are relevant only to a particular part of the project, then they should be added to a CLAUDE.md
+   file *in the relevant subdirectory*. Then the imports will only be processed when files within that directory are read.
+
+---
+
+## 3. Coding standards
+
+1. **Line Length**: 120 characters. This should also be observed for Markdown files, where manual newlines should be inserted.
```
ethenotethan: Couldn't we provide the existing linting yaml as context to Claude or have him run at as a post processing hook when making changes?

CLAUDE.md
```diff
@@ -0,0 +1,149 @@
...
+   there is to know about a function/type by its *name*, you should write a doc.
+3. Function/type docs should NOT simply be a rephrasing of the function/type name.
+   - E.g. the doc for `computeData` should NOT be "Computes the data".
+4. Function docs should consider the following helpful information, if relevant:
+   - What are the inputs?
+   - Are there any restrictions on what the input values are permitted to be?
+   - What is returned in the standard case?
+   - What is returned in the error case(s)?
+   - What side effects does calling the function have?
+   - Are there any performance implications that users should be aware of?
```
ethenotethan, line 96: Would it make sense to add a bullet point - what performance optimisations can be applied?

CLAUDE.md
```diff
@@ -0,0 +1,149 @@
...
+1. **Hierarchical organization**: Hierarchical numbering for sections makes referencing easier.
+2. **Tabular format for key facts**: Tables are helpful for understanding data at a glance: use them where appropriate.
+3. **Use Links**: Links are very helpful to assist a human navigating through the codebase.
+   - IMPORTANT: double check that links aren't broken after making changes to doc files. Similarly, if documentation
+   contains links directly to code, make sure that code changes are paired with the corresponding doc updates.
+
+---
+
+## 4. Directory-Specific CLAUDE.md Files
+
```
ethenotethan: Could we provide a lookup table here to the different subdirectory Claude mds once added?

CLAUDE.md
```diff
@@ -0,0 +1,149 @@
...
+3. Wrong working directory when running commands (ensure you're in project root).
+4. Large AI refactors in a single commit.
+5. Delegating test/spec writing entirely to AI (can lead to false confidence).
+
+---
+
+## 6. Files to NOT modify
+
+These files and directories should generally not be modified without explicit permission:
+
```
ethenotethan, line 135: Should we explicitly state bindings?

CLAUDE.md
```diff
@@ -0,0 +1,149 @@
...
+When responding to user instructions, the AI assistant (Claude, Cursor, GPT, etc.) should follow this process
+   to ensure clarity, correctness, and maintainability:
+
+1. **Consult Relevant Guidance**: When the user gives an instruction, consult the relevant instructions from
+   `CLAUDE.md` files (both root and directory-specific) for the request.
+2. **Clarify Ambiguities**: Based on what you could gather, see if there's any need for clarifications. If so,
+   ask the user targeted questions before proceeding.
+3. **Break Down & Plan**: Break down the task at hand and chalk out a rough plan for carrying it out,
+   referencing project conventions and best practices.
+4. **Trivial Tasks**: If the plan/request is trivial, go ahead and get started immediately.
```
ethenotethan, line 160: Could we specify a token usage threshold which deems "trivial"? Otherwise multiple instantiations of the same task could have differing side effects when prompting?

CLAUDE.md
```diff
@@ -0,0 +1,149 @@
...
+> human who edits this repository. It encodes our coding standards, guard rails, and workflow tricks.
+
+---
+
+## 1. Non-negotiable golden rules
+
+| #:  | AI *must NOT* do                                                                                                                                   |
+|-----|----------------------------------------------------------------------------------------------------------------------------------------------------|
+| G-0 | Write changes or use tools when you are not sure about something project-specific, or if you don't have context for a particular feature/decision. |
+| G-1 | Touch test files, specs, or generated files without EXPLICIT permission. **Humans own tests and specs.**                                           |
```
samlaf: what does "touch" mean here? Should we not want the AI to run `make protoc` or `make compile-contracts` to update the generated files?

CLAUDE.md
```diff
@@ -0,0 +1,149 @@
...
+| G-2 | Refactor large modules without human guidance. For changes >300 LOC or >3 files, **ask for confirmation**.                                         |
+
+---
+
+## 2. File imports
+
+NOTE: Be aware that whatever you add to this list is automatically loaded into context (due to `@` annotation). It's helpful
```
samlaf: link to https://docs.anthropic.com/en/docs/claude-code/common-workflows#reference-files-and-directories to show the @ notation semantics?

CLAUDE.md
```diff
@@ -0,0 +1,149 @@
...
+   to provide project context, but only within reasonable limits.
+
+| File Import                      | Description                          |
+|----------------------------------|--------------------------------------|
+| @Makefile                        | Build, test, and formatting commands |
+| @mise.toml                       | Dependency management                |
+| @README.md                       | Main project README file             |
+| @docs/spec/src/**/*.md           | All project spec markdown files      |
+| @.claude/commands/*              | Project claude commands              |
```
samlaf: are you sure that @ works with **/*? Don't see mentions of it in https://docs.anthropic.com/en/docs/claude-code/common-workflows#reference-files-and-directories

samlaf: what about adding go.mod? Wouldnt that be useful to give an overview of the project to the AI (roughly how many dependencies, and of what kind)?

Also this is prob out of scope for now but I feel like we should have some auto-generated file in the repo that includes a bunch of stats like code completion, yadi yada, and we can load that into claude's context?

CLAUDE.md
litt3: c113eaf8702d5496280874730aa1ff80491e2e65

CLAUDE.md
samlaf: +1 to this. Feel like we shouldn't repeat in fuzzy natural language rules that are already hardcoded in a machine-readable language somewhere else.
Unless somehow the dumb AI understands these fuzzy things better?

CLAUDE.md
litt3: a1ec752e34e6091d78c1562e0255cb495a1a1954

CLAUDE.md
```diff
@@ -0,0 +1,150 @@
...
+   file *in the relevant subdirectory*. Then the imports will only be processed when files within that directory are read.
+
+---
+
+## 3. Coding standards
+
+1. **Line Length**: 120 characters. This should also be observed for Markdown files, where manual newlines should be inserted.
+   - The line length limit may be ignored in cases where a line break hinders readability, e.g. URLs.
+2. **Imports**: Use `goimports` for sorted imports. Group standard library, third-party, and local imports.
+3. **Testing**: `*_test.go` files in same package; use `testify` for assertions.
```
samlaf: How about we point it to https://github.com/uber-go/guide/blob/master/style.md? Until we have our own golang guidelines

CLAUDE.md
litt3: 03467b6dd00bda708434547ce249a0a4fdabc93a

CLAUDE.md
```diff
@@ -0,0 +1,150 @@
...
+   - Are there any performance implications that users should be aware of?
+   - Are there any performance optimizations that should/could be undertaken in the future?
+
+### 3.3. Doc Files
+
+1. **Hierarchical organization**: Hierarchical numbering for sections makes referencing easier.
+2. **Tabular format for key facts**: Tables are helpful for understanding data at a glance: use them where appropriate.
+3. **Use Links**: Links are very helpful to assist a human navigating through the codebase.
+   - IMPORTANT: double check that links aren't broken after making changes to doc files. Similarly, if documentation
+   contains links directly to code, make sure that code changes are paired with the corresponding doc updates.
```
samlaf: why do we want the AI to write doc files? Arent those similar to tests/spec files and should be human generated, or at least guided? Kind of scared of ending up with huge pile of crap doc files

CLAUDE.md
```diff
@@ -0,0 +1,151 @@
...
+   processed whenever the `CLAUDE.md` file is read.
+6. If there is domain-specific terminology relevant to a directory, consider adding a small glossary of terms.
+
+---
+
+## 5. Common pitfalls
+
+1. Forgetting to run `go mod tidy` after adding new dependencies.
+2. Not linting before committing code.
+3. Wrong working directory when running commands (ensure you're in project root).
```
samlaf: not necessarily we have makefile commands in subdirs

CLAUDE.md
```diff
@@ -0,0 +1,151 @@
...
+5. Delegating test/spec writing entirely to AI (can lead to false confidence).
+
+---
+
+## 6. Files to NOT modify
+
+These files and directories should generally not be modified without explicit permission:
+
+1. **Generated files**: Any files that are automatically generated during build processes.
+   - Smart contract bindings are an important example of autogenerated files that shouldn't be directly modified under any circumstances
```
samlaf: can be re-generated using make commands though

CLAUDE.md
samlaf: what are tokens exactly? do they have the same semantic across different AI agents (claude, gpt, etc)?

samlaf (commented):
Thanks for adding this. Learned a ton. The two links you took inspiration from were specifically informative.

CLAUDE.md
litt3: This is a good point. 

I'm going to make a different PR to address line length linting because I think it goes beyond the scope of this PR. I'll post a link here when the PR is up.

CLAUDE.md
litt3: Hmm, interesting question. Do you have a token threshold in mind that would make sense? I don't really know what number of tokens correspond to a "small" task.

> Otherwise multiple instantiations of the same task could have differing side effects when prompting?

I'm not sure this is necessarily a problem. It seems that as AI-assisted development progresses, the threshold here is going to change, and prompts we would use today will behave differently.

CLAUDE.md
litt3: I meant "directly modify". Will fix

CLAUDE.md
litt3: I asked claude to evaluate, and he seemed happy with it. I will do some experimentation to make extra sure

litt3: > what about adding go.mod?

👍 

> Also this is prob out of scope for now but I feel like we should have some auto-generated file in the repo that includes a bunch of stats like code completion, yadi yada, and we can load that into claude's context?

This sounds interesting, and I'd like to better understand what you're proposing.

CLAUDE.md
litt3: Are you proposing that we have claude curl down that style doc every time it starts up?

It's an option, I'm just not 100% sure this is how we want to approach the problem. It seems almost a guarantee that we will encounter similar questions moving forward, so it makes sense to carefully consider procedure in this case.

If we just put a link and don't explicitly cause claude to pull the contents, I don't think it will do much.

CLAUDE.md
litt3: I agree that we don't want AI to be the main force behind generating doc files. These specifications were intended for when AI is *helping* a human write doc files. 

For example, in writing this `CLAUDE.md` file, I used the following process:

1. Had Claude generate a proposal based on the example file
2. Went through line-by-line and made a bunch of manual edits. 
    - By the time I was done, I had pretty much rewritten the whole thing, but the initial skeleton was still helpful as a starting point. 
3. Had Claude help me polish up formatting, check for internal consistency, propose improvements, etc.

These entries are to help Claude generate a helpful skeleton, and also denote preferences when improving upon the human-written doc.

If I more specifically call out the intention here, do you think that would help? I'm already planning on adding a similar section about tests, and how claude should and shouldn't behave when assisting a human.

litt3: Another thing to consider here, this file is not purely for AI assistants, but also for humans:

> This file is the onboarding manual for every AI assistant (Claude, Cursor, GPT, etc.) and every human who edits this repository.

CLAUDE.md
litt3: > directly modified

To me, modification via command doesn't fall under "direct modification". But I agree, probably worth specifying that it's ok to cause generated files to be regenerated, will add.

CLAUDE.md
litt3: Actually the relevant link is https://docs.anthropic.com/en/docs/claude-code/memory#claude-md-imports

c85b51f5654e29df16c4daf9a9be0dbc60181d82

CLAUDE.md
litt3: 84cc0c22193d70045a460fc23d03c0e2534e942f

CLAUDE.md
litt3: acf0786fe369fa35d96fb48d7b998eda8a9ff3e1 + c87d671339b17c3c1d566a6da512dbd33168e6b0

- The wildcards do _not_ work correctly. I had to create a `docs` CLAUDE.md to get this working (which I think probably makes more sense, anyway)
- Claude imports seem to not work when included in a table. Not sure if this is a bug, but I just reformatted to use a list instead of a table for now

You can use this prompt in a fresh `claude` context to verify that this is working as expected:

```
CLAUDE.md Import Test

  Test whether files are correctly imported via @ annotations. You are not permitted
  to run commands, search the repository, or read any files beyond what's already
  imported.

  Answer these questions using only pre-imported content:

  1. Go Version: What version of Go does this project use?
  2. Linting: Which staticcheck rules are excluded and why?

  If you can't answer without additional file access, state you lack that information.
```

CLAUDE.md
litt3: Reverted the addition of the link, in the interest of keeping this file limited to things that are useful to claude (as we discussed offline)

CLAUDE.md
litt3: 168eff493d65eb32142156639889e037e8351665

CLAUDE.md
litt3: b618df2a14d255ca88f72b17c90c5769e12ab286

CLAUDE.md
litt3: 71011df804a75a36d7f8f0e98db8ace131ec463c

I removed the fuzzy rules, and will work later to define these things in a better way.

I also moved the notes on testing to a dedicated `Testing` section of the doc

CLAUDE.md
litt3: 12443444e24944262282b5240d3d3e50968bafb1

CLAUDE.md
litt3: 00b5325c7abe7677a357db9f25e5aed2b93e75de

CLAUDE.md
litt3: I do think we should list out the important directories that exist in the root of the repository. e.g. 

```
./disperser <short description of content>
./contracts <short description of content>
./api <short description of content>
...
```

This is something I was planning to do as a follow-up. Or other people could also help by creating a sub-directory `CLAUDE.md` , and adding an entry to the list in the main `CLAUDE.md` 

It shouldn't be necessary to point at `CLAUDE.md` files specifically. As long as they exist in the subdirectories, Claude will parse them automatically:

https://docs.anthropic.com/en/docs/claude-code/memory#claude-md-imports

> Claude will also discover CLAUDE.md nested in subtrees under your current working directory. Instead of loading them at launch, they are only included when Claude reads files in those subtrees.

CLAUDE.md
litt3: We discussed this offline.

We will work on addressing the problem of a proper style guide in the near future.

CLAUDE.md
dmanc: Does it make sense to start utilizing precommit hooks or similar techniques for linting.

litt3: I've been thinking about this.

Definitely sounds like it could be helpful, I will do some exploration and maybe submit a proposal for how we could do this (alternatively, feel free to do a proposal of your own if so inclined)

CLAUDE.md
```diff
@@ -0,0 +1,188 @@
...
+## 2. File imports
+
+NOTE: Be aware that whatever you add to this list is automatically loaded into context (due to `@` annotation). It's
+helpful to provide project context, but only within reasonable limits.
+
+1. @Makefile contains commands for building, testing, and formatting
+2. @go.mod describes golang dependencies
+3. @mise.toml describes external tool dependencies
+4. @.golangci.yml contains linting configuration
+5. @docs/CLAUDE.md causes doc files to be automatically imported
```
dmanc: Does this mean all docs files are imported? Would that cause any issues with the context size?

litt3: Yes, this imports all docs.

I wrestled with this one, and I'm still not totally sure this is right answer.

Claude claimed that the current doc files fill up ~10% of claude code's context (200k tokens). ~10% of context seems like a pretty reasonable price for claude to have an *automatic* understanding of the entire project.

But I don't know- maybe a better solution would be just to auto import names and brief descriptions of the docs that exist, so that claude can go look stuff up if necessary. Or maybe be more selective with the ones that get auto loaded?

litt3: feb369189cb17fcca576aea47d73b981274d5e2e

I decided to remove the auto parsing for now. I noticed that I had actually missed reading in a bunch of the docs, but adding all the ones I missed would be *a lot* more than before.

We can potentially consider the possibility of auto-parsing docs in the future

CLAUDE.md
litt3: feb369189cb17fcca576aea47d73b981274d5e2e

I started the table in this commit. I only added `docs` for now, but as we make more `CLAUDE.md`s, we can add to this table

CLAUDE.md
samlaf: This should be simple, just add a precommit hook file that calls `make lint` and probably `make unit-tests` or whatever. We can add to it slowly

CLAUDE.md
samlaf: Nicely crafted reformat, like it!

samlaf (approved):
LGTM. Let's get this merged and iterate on it as possibly play around with it, add more Claude.mds in subdirs, etc.

PR #1707, opened 2025-07-01
fix: structify address config

pakim249CAL:

Makes address configs structs to reduce error vectors

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/contracts/script/deploy/eigenda-directory/DeployEigenDADirectory.sol
+++ b/contracts/script/deploy/eigenda-directory/DeployEigenDADirectory.sol
@@ -56,10 +56,11 @@ contract DeployEigenDADirectory is Script {
         string name;
         address value;
     }
+
     function _populateDirectory() internal virtual {
         // Dynamically read all contract names from the [contracts] table
         string[] memory contractNames = vm.parseTomlKeys(cfg, ".contracts");
-        
+
         for (uint256 i; i < contractNames.length; i++) {
             string memory name = contractNames[i];
             address contractAddress = cfg.readAddress(string.concat(".contracts.", name));


contracts/script/deploy/eigenda-directory/DeployEigenDADirectory.sol
```diff
@@ -52,14 +52,17 @@ contract DeployEigenDADirectory is Script {
         );
     }
 
-    function _populateDirectory() internal virtual {
-        string[] memory names = cfg.readStringArray(".contractNames");
-        address[] memory values = cfg.readAddressArray(".contractAddresses");
+    struct AddressConfig {
```
litt3, line 55: Claude suggested that you could do:

```
[contracts]
PAUSER_REGISTRY = "0x9Ab2FEAf0465f0eD51Fc2b663eF228B418c9Dad1"
BLS_APK_REGISTRY = "0xAd7f9e558170a149Ca8E90f41Ab2444A5d3bd6aD"
INDEX_REGISTRY = "0x8cE5F2a53cBd29710eb94A04e40C07A4DdF15d10"
STAKE_REGISTRY = "0x53668EBf2e28180e38B122c641BC51Ca81088871"
SOCKET_REGISTRY = "0x1747ef24dbbb52cB06382d323f455D48dE1AC7fd"
REGISTRY_COORDINATOR = "0x2c61EA360D6500b58E7f481541A36B443Bc858c6"
EJECTION_MANAGER = "0xEEed164605833263B7690bB521983b777D128698"
SERVICE_MANAGER = "0x54A03db2784E3D0aCC08344D05385d0b62d4F432"
OPERATOR_STATE_RETRIEVER = "0x003497Dd77E5B73C40e8aCbB562C8bb0410320E7"
THRESHOLD_REGISTRY = "0x41AEE4A23770045e9977CC9f964d3380D6Ff9e4E"
RELAY_REGISTRY = "0xca1ca181fCb3c4192D320569c6eB4b5161B80328"
DISPERSER_REGISTRY = "0xD45323B6d168bd6749294f66F27Dfa41b7dF2102"
PAYMENT_VAULT = "0x46E024ca6e5E1172100930c28DCCcF49BE5462C9"
CERT_VERIFIER = "0xCCFE3d87fB7D369f1eeE65221a29A83f1323043C"
```

and then parsing would look like this:

```
    function _populateDirectory() internal virtual {
        // Dynamically read all contract names from the [contracts] table
        string[] memory contractNames = vm.parseTomlKeys(cfg, ".contracts");
        
        for (uint256 i; i < contractNames.length; i++) {
            string memory name = contractNames[i];
            address contractAddress = cfg.readAddress(string.concat(".contracts.", name));
            directory.addAddress(name, contractAddress);
        }
    }
```

I don't know much about toml, so maybe this wouldn't actually work. But have you ever seen this strategy before?

pakim249CAL: I've never seen the strategy, but it makes sense to me! I like it. I will try it.

pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1707/commits/1d7da33d299a348d00708282990cbffc90c0b3c0

contracts/script/deploy/eigenda-directory/config/example.config.toml
```diff
@@ -1,12 +1,8 @@
...
-proxyAdmin = "0x9Fd7E279f5bD692Dc04792151E14Ad814FC60eC1"
+proxyAdmin = "0xF36fDAf681E3D88Ee412879aa4907B60D6ff0707"
 
 # The privileged account that can manage the contract directory.
-owner = "0xDA29BB71669f46F2a779b4b62f03644A84eE3479"
\ No newline at end of file
+owner = "0xF36fDAf681E3D88Ee412879aa4907B60D6ff0707"
+
+[contracts]
+EXAMPLE = "0xF36fDAf681E3D88Ee412879aa4907B60D6ff0707"
```
samlaf: Guessing this is just a random example address? 
Why not use 0x0, 0x1, 0x2 or something? I was puzzled looking at this for a minute, wondering why proxyAdmin and owner were the same address, until I realized EXAMPLE is the same.

samlaf (approved):

contracts/script/deploy/eigenda-directory/config/example.config.toml
pakim249CAL: Yes, I like to just use newly generated keys as example keys since I can be reasonably sure they aren't being used.

samlaf: I personally find it confusing because it looks like those could be our holesky addresses or something.

pakim249CAL: It's not really the case here, but there are scenarios in which using a known address can be very problematic especially if those addresses have any funds. If an example script were to use a known address and be used in tests, the fact that an account has funds may be missed. 

So I like to avoid using any commonly known addresses. The quickest way I know to get a clean address that nobody has access to is to generate a new one in my terminal.

PR #1709, opened 2025-07-02
feat: Payments subgraph

dmanc:

Adds a subgraph for the PaymentVault. Mainly used to get observability into reservation expirations. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


subgraphs/eigenda-payments/abis/PaymentVault.json
samlaf, line 1: Did you copy-paste this in here or is this auto-generated? Should prob be auto-generated from our contracts and have a ci to check that its still up-to-date?

subgraphs/eigenda-payments/docker-compose.yml
samlaf, line 1: are we actually using this given that we upload on satsuma? If not how about we just delete it?
Otherwise there doesn't seem to be anything specific to the eigenda-payments subgraph in here afaict? Should prob move this under `subgraphs/` directly and share it between all the subgraphs?

subgraphs/eigenda-payments/schema.graphql
samlaf, line 1: assuming this is also auto-generated?
Would be nice to know what is auto-generated and what is hand-written.
Is it possible to point the subgraph cli to auto-generate stuff under a `eigenda-payments/generated/` subdir?

subgraphs/eigenda-payments/schema.graphql
```diff
@@ -0,0 +1,92 @@
...
+  reservation_startTimestamp: BigInt! # uint64
+  reservation_endTimestamp: BigInt! # uint64
+  reservation_quorumNumbers: Bytes! # bytes
+  reservation_quorumSplits: Bytes! # bytes
+  blockNumber: BigInt!
+  blockTimestamp: BigInt!
+  transactionHash: Bytes!
+}
+
+type CurrentReservation @entity(immutable: false) {
```
samlaf: you mentioned on slack that this is the only entity that you added by hand? Maybe add a big `============= HAND-WRITTEN BELOW =========` demarcation comment to know what is auto-generated and what is hand-written?

subgraphs/eigenda-payments/src/payment-vault.ts
```diff
@@ -0,0 +1,170 @@
...
+  entity.blockNumber = event.block.number
+  entity.blockTimestamp = event.block.timestamp
+  entity.transactionHash = event.transaction.hash
+
+  entity.save()
+
+  // Create or update the ActiveReservation entity for this account
+  let activeReservation = CurrentReservation.load(event.params.account)
+  if (activeReservation == null) {
+    activeReservation = new CurrentReservation(event.params.account)
```
samlaf: do these currentReservations ever get deleted? How does the subgraph know that its still current? DO we have to filter manually for endTimestamp?

hopeyen: reservations never get deleted on-chain, so we need to filter. I would recommend adding a `status` enum type for `active`, `expired` or `inactive` since there could be reservation that gets activated in the future

subgraphs/eigenda-payments/docker-compose.yml
dmanc: It's autogenerated by the graph init command. I assume it's useful for locally debugging a subgraph but I haven't tried to use it.

subgraphs/eigenda-payments/schema.graphql
dmanc: We have to update some of these files by hand though. I do like the idea of adding a comment distinguishing which part is generated or not.

subgraphs/eigenda-payments/abis/PaymentVault.json
dmanc: I compiled the contracts and copy pasted it here, see README.md at the root of the directory. Think it's a good idea for checking if the abi is up to date and failing some check if it isn't, will add it if it's straightforward.

subgraphs/eigenda-payments/src/payment-vault.ts
dmanc: Will use the following logic
```
enum ReservationStatus {
  PENDING    # startTimestamp > current time
  ACTIVE     # startTimestamp <= current time < endTimestamp
  EXPIRED    # endTimestamp <= current time
}
```

dmanc: Actually after trying to add this status enum I realized it doesn't really make sense to do on the subgraph side since right now we only update the `CurrentReservation` entity every time there's a `ReservationUpdated` event. One potential solution is to add a block handler that loops over all existing reservations but I feel that it's starting to make the subgraph more complicated.

Currently, the semantics of `CurrentReservation` is to give the most up to date reservation for a given account. I would prefer to leave the filtering (based on the logic in the previous comment) to the clients of the subgraph. 

e.g.
```
# Active reservations
query ActiveReservations($now: BigInt!) {
  currentReservations(
    where: {
      startTimestamp_lte: $now,
      endTimestamp_gt: $now
    }
  ) {
    account
    symbolsPerSecond
  }
}

# Pending reservations
query PendingReservations($now: BigInt!) {
  currentReservations(
    where: { startTimestamp_gt: $now }
  ) {
    account
    startTimestamp
  }
}

# Expired reservations
query ExpiredReservations($now: BigInt!) {
  currentReservations(
    where: { endTimestamp_lte: $now }
  ) {
    account
    endTimestamp
    symbolsPerSecond
  }
}
```

hopeyen: makes sense, that's okay with me

dmanc: added QUERY_EXAMPLES.md for some documentation https://github.com/Layr-Labs/eigenda/pull/1709/commits/5e1c0349f5db16575741a5d81caff1d7e14cdc7e#diff-78b335265d03906c5748667a51e96c145f714a1b699ea4f980a4900a175a6044

subgraphs/eigenda-payments/docker-compose.yml
dmanc: removed bfbbaf7fd451f4d1b69673e4f28b9d4187d75581

subgraphs/eigenda-payments/abis/PaymentVault.json
dmanc: Added the checks here 2fe68a0addd38e057d7d6477ca4f1e641eab9a45. Seems like the ABIs have indeed changed but shouldn't affect the existing subgraphs at all.

I do think the check is useful so that anytime any functionality is added to the contract we can check and see if anything useful can be added to the subgraph.

subgraphs/eigenda-payments/schema.graphql
dmanc: Decided it's not worth trying to distinguish between what was generated and not, I think it's going to be fine to assume it's all hand written since after running `graph init` it's up to us to maintain the code.

subgraphs/eigenda-payments/abis/PaymentVault.json
samlaf: You should be using IPaymentVault instead of PaymentVault. Changes less often, and we only need the interface anyways. I think I'm fine with this approach (although it does look like a lot of hard to maintain ai generated code.. so I hope it won't be flakey in CI), but @cody-littley has always been against checking in compiled artifacts into the repo, so wondering if this complexity could all have been alleviated by just having a makefile command to go compile the contracts and use the generated artifacts from there?

subgraphs/eigenda-payments/src/payment-vault.ts
samlaf: Yep that makes a ton of sense, nice md! Can we just change `currentReservations` to simply `reservations` everywhere then, since expired reservations are not deleted, so they aren't "current"? Or maybe `registeredReservations` or something?

subgraphs/eigenda-payments/schema.graphql
samlaf: You mentioned in the other comment not wanting to distinguish what is auto-generated vs manually generated. I think I agree, though I still would like to be able to easily distinguish which parts of the schema are purely mapping 1:1 to onchain events, and which ones are derived state (aka only this CurrentReservation). So I think I would still want something like
```
# Everything above here maps 1:1 to onchain events
================== EVENT-DERIVED STATE BELOW ==============
```

dmanc: sg, will add that

subgraphs/eigenda-payments/src/payment-vault.ts
dmanc: Went with `Reservation` 6de385111b428c33b72851236d7ff72c0b52c08f

subgraphs/eigenda-payments/schema.graphql
dmanc: 6de385111b428c33b72851236d7ff72c0b52c08f

subgraphs/eigenda-payments/abis/PaymentVault.json
dmanc: talked online and we decided to remove the ABIs since they're really only used for bootstraping the subgraph 451a0bf50f57681523ea40c817a313157a217ef3

dmanc: oops it actually does need it when running the code generation

PR #1711, opened 2025-07-02
fix: correctness tests

cody-littley:

Gets correctness tests functional again.


test/v2/correctness/correctness_test.go
```diff
@@ -56,7 +56,7 @@ func checkAndSetCertVerifierAddress(t *testing.T, c *client.TestClient, certVeri
 // - read the blob from the relays
 // - read the blob from the validators
 func testBasicDispersal(c *client.TestClient, payload []byte) error {
-	ctx, cancel := context.WithTimeout(context.Background(), 2*time.Minute)
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
```
samlaf: nit: typically better to pass in the `*testing.Testing` instance and use `t.Context()` whose timeout/cancellation is tied to the actual test instance, as opposed to having arbitrary timeouts like this 5mins.

test/v2/correctness/correctness_test.go
```diff
@@ -82,7 +82,6 @@ func emptyBlobDispersalTest(t *testing.T, environment string) {
 	// This should fail with "data is empty" error
 	_, _, err := c.GetDisperserClient().DisperseBlob(ctx, blobBytes, 0, quorums)
 	require.Error(t, err)
-	require.ErrorContains(t, err, "blob size must be greater than 0")
```
samlaf, line 85: why are you removing this? this could pass for some other random error like this no?

test/v2/correctness/correctness_test.go
```diff
@@ -286,20 +285,27 @@ func smallBlobDispersalAllQuorumsSetsTest(t *testing.T, environment string) {
...
+		require.NoError(t, err)
+	})
 
-	checkAndSetCertVerifierAddress(t, c, config.EigenDACertVerifierAddressQuorums2)
-	err = testBasicDispersal(c, payload)
-	require.NoError(t, err)
+	t.Run("2", func(t *testing.T) {
+		checkAndSetCertVerifierAddress(t, c, config.EigenDACertVerifierAddressQuorums2)
+		err = testBasicDispersal(c, payload)
+		require.NoError(t, err)
```
samlaf: can these be run in parallel? by adding `t.Parallel()` at the beginning of each

samlaf (dismissed):
LGTM other than small nits/questions

litt3 (dismissed):

test/v2/correctness/correctness_test.go
cody-littley: Yes, but I've found that this test suite isn't very stable when run with parallelism. I need to iterate specifically on improving parallelism.

test/v2/correctness/correctness_test.go
cody-littley: The error string changed, which was causing this test to fail. IMO, tests on magic error strings are fragile, so I generally don't prefer to have that type of assertion. But I've added an assertion back here since that's your preference.

test/v2/correctness/correctness_test.go
cody-littley: Switched to using the background context. Is that ok, or is it importing to use t's context?

litt3 (approved):

test/v2/correctness/correctness_test.go
samlaf: I agree with you. Maybe we can define a sentinel error?

test/v2/correctness/correctness_test.go
samlaf: Don't think its of utmost importance, but I do think its preferred.

cody-littley: switched everything in this file to use `t.Context()`

test/v2/correctness/correctness_test.go
cody-littley: done

litt3 (approved):

PR #1712, opened 2025-07-02
feat: client-side accountant local account ledger 

hopeyen:

- Aiming for thread-safe local account ledger that mimics Accountant behavior for client-side usage tracking. 
- Atomic all-or-nothing multi-quorum transactions
- Reservation overflow bins
- Protobuf serialization compatibility with disperser RPC
- Reservation → overflow → on-demand fallback logic

Should be compatible with existing PaymentVault and Accountant interfaces

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


core/meterer/local_account_ledger.go
```diff
@@ -0,0 +1,379 @@
...
+	reservations map[uint32]*disperser_v2.QuorumReservation,
+	periodRecords map[uint32]*disperser_v2.PeriodRecords,
+	onchainCumulativePayment []byte,
+	cumulativePayment []byte,
+) (*LocalAccountLedger, error) {
+	// Convert protobuf reservations to Go structures
+	goReservations := make(map[core.QuorumID]*core.ReservedPayment)
+	for quorumID, protoReservation := range reservations {
+		if protoReservation != nil {
+			goReservations[core.QuorumID(quorumID)] = &core.ReservedPayment{
```
litt3: Might be good to create a proper constructor, instead of manually building the struct with a subset of the fields defined. This type of instantiation is difficult to maintain

core/meterer/local_account_ledger.go
```diff
@@ -0,0 +1,379 @@
...
+	accountID gethcommon.Address,
+	timestampNs int64,
+	numSymbols uint64,
+	quorumNumbers []core.QuorumID,
+	params *PaymentVaultParams,
+) (*big.Int, error) {
+	lal.mutex.Lock()
+	defer lal.mutex.Unlock()
+
+	// Store reservation error for potential fallback error message (matches Accountant behavior)
```
litt3: `(matches Accountant behavior)`

I'm assuming that these are comments left by an AI agent while copying the old code.

core/meterer/local_account_ledger.go
```diff
@@ -0,0 +1,379 @@
...
+	var reservationError error
+
+	// First, try reservation usage for ALL quorums (matches Accountant behavior)
+	// Validate all reservations exist and are valid for this timestamp
+	reservationValidationErr := ValidateReservations(lal.reservations, params.QuorumProtocolConfigs, quorumNumbers, timestampNs, time.Now().UnixNano())
```
litt3: a time source should be passed in (probably to the constructor) instead of using `time.Now()`

core/meterer/local_account_ledger.go
```diff
@@ -0,0 +1,379 @@
...
+// Debit implements AccountLedger.Debit with atomic multi-quorum transaction semantics.
+// Acquires exclusive write lock, validates all quorums, applies deep-copy-commit pattern.
+// Uses all-or-nothing logic: either ALL quorums use reservations or ALL use on-demand.
+func (lal *LocalAccountLedger) Debit(
+	ctx context.Context,
+	accountID gethcommon.Address,
+	timestampNs int64,
+	numSymbols uint64,
+	quorumNumbers []core.QuorumID,
+	params *PaymentVaultParams,
```
litt3: You're passing the `PaymentVaultParams` to each `Debit` call, but does the system actually support dynamically changing `PaymentVaultParams`? How is this deterministic across the whole network?

hopeyen: disperser server and validator node will be able to dynamically change `PaymentVaultParams`. When the debit is in progress, `PaymentVaultParams` should be fixed so that the calculations are consistent. For the client, this variable is fixed upon initialization atm

core/meterer/local_account_ledger.go
hopeyen: updated the constructor [9097dbf](https://github.com/Layr-Labs/eigenda/pull/1712/commits/9097dbf68953fa7f80170075b679198e2b634b00)

core/meterer/local_account_ledger.go
hopeyen: yep! removed

core/meterer/local_account_ledger.go
hopeyen: made a constructor [0d64da0](https://github.com/Layr-Labs/eigenda/pull/1712/commits/0d64da0922455690c4c23e260e8f461df757422f)

core/meterer/local_account_ledger.go
litt3: > disperser server and validator node will be able to dynamically change PaymentVaultParams

How will this be synchronized across the network? It seems like `PaymentValueParams` would need to be defined in terms of `RBN`, like we are doing for the `CertVerifierAddress`. Otherwise, the cost of a dispersal will be non-deterministic around parameter change boundaries.

hopeyen: Valid concerns. `PaymentVaultParams` will be fairly rare, we could even add a constraint to the contract that the price can only decrease to protect the users. Let's discuss it separately from this PR

anupsv: Reading through.

PR #1713, opened 2025-07-02
test: set up periodic tests against live environments

cody-littley:

Introduces the `live test` panel, which runs periodically against live environments.


test/v2/client/test_client.go
```diff
@@ -356,26 +357,36 @@ func NewTestClient(
...
+	var privateKey string
 	if keyPath != "" {
-		privateKeyFile, err := ResolveTildeInPath(keyPath)
+		privateKeyFile, err := util.SanitizePath(keyPath)
 		if err != nil {
 			return "", fmt.Errorf("failed to resolve tilde in path: %w", err)
 		}
-		privateKey, err := os.ReadFile(privateKeyFile)
+
+		exists, err := util.Exists(privateKeyFile)
```
litt3, line 367: Probably would be good at some point to move these utilities to a more common location, now that they're being used outside of LittDB. 

Not recommending changes, just mentioning

test/v2/client/test_client_config.go
```diff
@@ -73,7 +75,7 @@ type TestClientConfig struct {
 
 // ResolveSRSPath returns a path relative to the SRSPath root directory.
 func (c *TestClientConfig) ResolveSRSPath(srsFile string) (string, error) {
-	root, err := ResolveTildeInPath(c.SRSPath)
+	root, err := util.SanitizePath(c.SRSPath)
 	if err != nil {
 		return "", fmt.Errorf("resolve tilde in path: %w", err)
```
litt3: nit: outdated error message

test/v2/client/test_client_setup.go
```diff
@@ -110,7 +132,7 @@ func GetClient(configPath string) (*TestClient, error) {
 }
 
 func skipInCI(t *testing.T) {
-	if os.Getenv("CI") != "" {
+	if os.Getenv("CI") != "" && os.Getenv("LIVE_TESTS") == "" {
```
litt3, line 147: could you comment in english when this is intended to run, or modify the `Skip` message?

test/v2/client/test_client_setup.go
```diff
@@ -21,10 +23,30 @@ var (
...
 
-const (
-	PreprodEnv = "../config/environment/preprod.json"
-	TestnetEnv = "../config/environment/testnet.json"
-)
+// GetEnvironmentConfigPaths returns a list of paths to the environment config files.
+func GetEnvironmentConfigPaths() ([]string, error) {
+	configDir, err := util.SanitizePath("../config/environment")
+	if err != nil {
+		return nil, fmt.Errorf("failed to resolve tilde in path: %w", err)
```
litt3: tilde doesn't appear to be involved

test/v2/client/test_client_setup.go
```diff
@@ -35,7 +57,7 @@ func GetConfig(configPath string) (*TestClientConfig, error) {
 		return config, nil
 	}
 
-	configFile, err := ResolveTildeInPath(configPath)
+	configFile, err := util.SanitizePath(configPath)
 	if err != nil {
 		return nil, fmt.Errorf("failed to resolve tilde in path: %w", err)
```
litt3: old error message

test/v2/load/load_generator.go
```diff
@@ -51,7 +52,7 @@ type LoadGenerator struct {
 
 // ReadConfigFile loads a LoadGeneratorConfig from a file.
 func ReadConfigFile(filePath string) (*LoadGeneratorConfig, error) {
-	configFile, err := client.ResolveTildeInPath(filePath)
+	configFile, err := util.SanitizePath(filePath)
 	if err != nil {
 		return nil, fmt.Errorf("failed to resolve tilde in path: %w", err)
```
litt3: outdated error message

litt3 (dismissed):

.github/workflows/live-network-tests.yaml
```diff
@@ -0,0 +1,64 @@
...
+      LIVE_TEST_PREPROD_HOLESKY_KEY: ${{ secrets.LIVE_TEST_PREPROD_HOLESKY_KEY }}
+      LIVE_TEST_TESTNET_HOLESKY_KEY: ${{ secrets.LIVE_TEST_TESTNET_HOLESKY_KEY }}
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: '1.24'
```
samlaf: please use mise. Look at the other workflows for how its done

.github/workflows/live-network-tests.yaml
```diff
@@ -0,0 +1,64 @@
...
+
+      - name: Install dependencies
+        run: go mod download
+
+      - name: Run Live Network Tests
+        run: go test ./test/v2/live -v -timeout 60m
```
samlaf: can we add a makefile command for this? Reduce the diff between offchain and onchain. I should be able to rerun the same CI locally without having to go read, parse and understand the workflow yaml file imo

.github/workflows/live-network-tests.yaml
```diff
@@ -0,0 +1,64 @@
...
+              attachments: [
+                {
+                  color: $color,
+                  title: $title,
+                  title_link: $title_link
+                }
+              ]
+            }')
+
+          curl -X POST -H "Authorization: Bearer ${{ secrets.DA_TEST_REPORTER_SLACK_OATH_TOKEN }}" \
```
samlaf, line 65: should this secret be named `LIVE_TEST_...` to have same prefix as the secret keys?

test/v2/load/load_generator.go
samlaf: @litt3 this makes me wonder whether your strategy of having the call-site add context for the function name makes sense. If this was done in the callee, then it would just need to be changed in one single place and everything would remain consistent.

test/v2/config/environment/testnet.json
```diff
@@ -1,6 +1,7 @@
 {
   "SRSPath": "../../../resources/srs",
   "KeyPath": "~/.test-v2/testnet-private-key.txt",
+  "KeyVar": "LIVE_TEST_TESTNET_HOLESKY_KEY",
```
samlaf, line 4: would be very useful to have a comment here as to which one has priority over the other one. Ive asked that we stop using json for contract configs and use toml everywhere. Think we should adopt this across the board. No need to do this as part of this PR but suggestion going forward.

yaml or toml both allow for comments, and both have parsers readily available in most languages.

samlaf (approved):

test/v2/load/load_generator.go
litt3: True, this would be something to go in the "cons" column of my strategy

Once we get claude spun up to check internal consistency stuff, though, I expect pain points like this to evaporate.

test/v2/client/test_client.go
cody-littley: I'll keep this in mind. I agree that this PR is not the proper PR to move these files.

test/v2/client/test_client_config.go
cody-littley: fixed

test/v2/client/test_client_setup.go
cody-littley:
```
	// The environment variable "CI" will be set when running inside a github action.
	// The environment variable "LIVE_TESTS" will be set when running live tests, which is a specific github action.
	//
	// There are three situations we want to consider:
	//
	// 1. When running a tests locally, we want to run live tests if requested. "CI" will not be set, and so 
	//    we will not skip the test.
	// 2. When we are running general unit tests as a github action, we specifically don't want to run live tests.
	//    "CI" will be set, and "LIVE_TESTS" will not be set, so we skip the test.
	// 3. When we are running live tests as a github action, we want to run the test. Both "CI" and "LIVE_TESTS" will 
	//    be set, so we do not skip the test.
	if os.Getenv("CI") != "" && os.Getenv("LIVE_TESTS") == "" {
		t.Skip("Skipping test in CI environment")
	}
```

test/v2/client/test_client_setup.go
cody-littley: fixed

test/v2/client/test_client_setup.go
cody-littley: fixed

test/v2/load/load_generator.go
cody-littley: Fixed.

I'd kill to have a macro that would capture the file name + line number in a string at compile time. The whole error wrapping thing could effectively be automated with zero overhead in a way that results in stack traces for errors. 

Have I ever mentioned how much I despise golang's design philosophy?

.github/workflows/live-network-tests.yaml
cody-littley: Can't change it easily now, since I've lost my elevated github privileges. Does this bother you enough to loop in IT?

test/v2/config/environment/testnet.json
cody-littley: I'm in favor of toml. YAML is nice in theory, but is actually quite horrid when you look closer. Who thought it would be a good idea to make YAML capable of holding executable code? Seriously...

With that in mind, I'd prefer to safe that kind of change for a different PR. Let's start a discussion.

.github/workflows/live-network-tests.yaml
cody-littley: done

.github/workflows/live-network-tests.yaml
cody-littley: added

.github/workflows/live-network-tests.yaml
```diff
@@ -4,6 +4,12 @@ on:
   schedule:
     - cron: '0 6,18 * * *'   # Runs daily at 6 AM and 6 PM UTC
   workflow_dispatch: {}         # Allow manual triggering
+  push: # TODO: temporary for testing, do not merge
```
litt3: don't forget TODO

cody-littley: Yup, removed

litt3 (approved):

PR #1714, opened 2025-07-02
fix: Use protobuf getters instead of direct access

litt3:
- Fixes all direct protobuf field access. Uses getters instead
- Turns on the `protogetter` linting rule
- inspired by https://github.com/Layr-Labs/eigenda/pull/1710


cody-littley (dismissed):

samlaf (approved):
LGTM. Any idea why TestInterleavedVersions is failing for both sepolia and holesky? We recently increased batch interval 1->5min on sepolia, but not sure why holesky is failing (not sure what its batch interval is either..)
Should we increase their timeout? Or actually we prob want to move them to @cody-littley 's live test framework?

samlaf (approved):

PR #1715, opened 2025-07-02
feat: add goreleaser

pschork:
- Replaces `docker-publish-release` action with goreleaser action.
- Speeds up release builds by 20min
- Runs snapshot release build (no publish) on PRs and merge to master
- Builds and publishes a (draft) release on any tag push that starts with `v`
- Goreleaser also builds binary tar balls in addition to docker images
- Additional binaries can be included in the tar balls (ie littdb snapshot tool)

Example releases in my testing fork https://github.com/pschork/eigenda/releases


.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
...
+        if: ${{ !startsWith(github.ref, 'refs/tags/v') }}
+
+      # It's possible that goreleaser resolves the wrong previous tag (ie tag:contracts/mainnet/v2.0.0)
+      # so we allow workflow_dispatch to override the previous tag if needed
+      - name: Override previous tag
+        run: echo "GORELEASER_PREVIOUS_TAG=${{ inputs.previous_tag }}" >> $GITHUB_ENV
+        if: ${{ inputs.previous_tag != '' }}
+
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@v6
```
anupsv: Let's use the actual hash.

.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
...
+
+      - uses: jdx/mise-action@v2
+        with:
+          version: v2025.7.1 # [default: latest] mise version to install
+          log_level: info
+          github_token: ${{ secrets.GITHUB_TOKEN }}
+          experimental: true # required for go-ethereum/cmd/abigen
+
+      - name: Login to ghcr
+        uses: docker/login-action@v1
```
anupsv: Same here. Let's use hash instead of V1

.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
...
+  push:
+    tags:
+      - v*
+    branches:
+      - master
+  pull_request:
+  workflow_dispatch:
+    inputs:
+      previous_tag:
+        description: "Override GORELEASER_PREVIOUS_TAG"
```
dmanc, line 13: When would we use this?

.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
...
+
+      - name: Login to ghcr
+        uses: docker/login-action@v1
+        with:
+          registry: ghcr.io
+          username: ${{ github.repository_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+
+      # If the github.ref is not a release tag, set the snapshot flag to run a release build check
+      # When running on master, it will fail if it resolves a nonstandard tag (ie tag:contracts/mainnet/v2.0.0)
```
dmanc, line 47: Does this mean goreleaser only works for tags that are vx.y.z? I believe that we are thinking about separating the releases at some point so we would probably want tags for node/v.x.y.z and proxy/v.x.y.z, etc. Not sure if that throws a wrench in this setup though.

.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
...
+  contents: write
+  packages: write
+  # issues: write
+
+jobs:
+  release:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
```
anupsv: let's use the hash like in other actions?

.github/workflows/release.yml
```diff
@@ -0,0 +1,73 @@
...
+        with:
+          fetch-depth: 0
+
+      - uses: jdx/mise-action@v2
```
anupsv: same here

.github/workflows/release.yml
anupsv: So make it deploy or fail? And delete out of band ?

.github/workflows/release.yml
samlaf: +1 we would need this to work with per-service releases which use tags with prefixes

.goreleaser.yaml
```diff
@@ -0,0 +1,124 @@
...
+
+release:
+  github:
+    owner: layr-labs
+    name: eigenda
+
+  draft: true
+
+archives:
+  - formats: [ 'tar.gz' ]
```
samlaf: I tried building node with `goreleaser build --id node --single-target --snapshot` and it failed with:
image

samlaf:
```suggestion
  - format: tar.gz
```

samlaf: fixed by changing to this. I'm on version
```
GitVersion:    2.4.8
GitCommit:     377981ebd76e1bbb0dbe07d5428239ec8c5381a8
GitTreeState:  clean
BuildDate:     2024-11-22T12:56:52
```
perhaps we need to pin goreleaser version in mise.toml?

.github/workflows/release.yml
```diff
@@ -0,0 +1,72 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 #4.2.2
+        with:
+          fetch-depth: 0
+
+      - uses: jdx/mise-action@5cb1df66ed5e1fb3c670ea0b62fd17a76979826a #2.3.1
+        with:
+          version: v2025.7.1 # [default: latest] mise version to install
```
samlaf: can we stick with the same mise version that we use in other workflows?

.github/workflows/release.yml
samlaf: I'm also confused as to why/when this is needed and why. Can we add a comment explanation

.github/workflows/release.yml
```diff
@@ -0,0 +1,72 @@
...
+      - name: Run GoReleaser
+        uses: goreleaser/goreleaser-action@9c156ee8a17a598857849441385a2041ef570552 #6.3.0
+        with:
+          distribution: goreleaser
+          # 'latest', 'nightly', or a semver
+          version: "~> v2"
+          args: release --clean ${{ env.SNAPSHOT_FLAG }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GORELEASER_CURRENT_TAG: ${{ env.GORELEASER_CURRENT_TAG }}
```
samlaf, line 71: I feel like it would be better to use goreleaser for everything (including local dev builds). Meaning we add goreleaser to mise and then use makefile commands here instead of github actions. Is that possible? Is there some important caching or something else that we would miss out on if we did that?

.goreleaser.yaml
```diff
@@ -0,0 +1,124 @@
...
+
+builds:
+  - id: node
+    main: ./node/cmd/main.go
+    binary: node
+    ldflags:
+      - -X 'github.com/Layr-Labs/eigenda/node.SemVer={{ .Version }}'
+      - -X 'github.com/Layr-Labs/eigenda/node.GitCommit={{ .ShortCommit }}'
+      - -X 'github.com/Layr-Labs/eigenda/node.Arch={{ .Arch }}'
+      - -X 'github.com/Layr-Labs/eigenda/node.Os={{ .Os }}'
```
samlaf, line 19: Out of scope for this PR but I feel like we should move away from these hand-written version injection and instead use golang's vcs facility: https://pkg.go.dev/runtime/debug#ReadBuildInfo

Since go1.18 this vcs info is already automatically injected into binaries
image

Dockerfile
```diff
@@ -142,10 +142,18 @@ FROM alpine:3.22 AS node
 COPY --from=node-builder /app/node/bin/node /usr/local/bin
 ENTRYPOINT ["node"]
 
+FROM alpine:3.22 AS node-goreleaser
+COPY node /usr/local/bin
+ENTRYPOINT ["node"]
```
samlaf, line 170: How is this different from the node target? Why are we keeping the node target around (and the bake file)? Shouldn't we have a single way to build binaries/dockerfiles?

Makefile
```diff
@@ -111,6 +111,9 @@ docker-release-build:
 semver:
 	echo "${SEMVER}"
 
+release-build:
+	goreleaser release --skip publish --clean --snapshot
```
samlaf: Should we add a non-release build command? Something like `goreleaser build`?
Also why are we keeping `docker-release-build` around if we don't plan on using it? If its just incase let's mark it as deprecated and plan to get rid of it asap?

.github/workflows/release.yml
pschork: This is an escape hatch in the event that goreleaser gets confused when it tries to determine the previous release tag to construct release notes. It will never get confused (when making an official release) in identifying the current release tag because it requires the HEAD of the release branch to be tagged. But in building the release notes, it's possible for it to resolve the wrong previous release tag ie it might think `contracts/mainnet/v2.0.0` is the previous tag of `v2.1.0`. If this happened, someone could trigger the workflow on the `v2.1.0` tag, and then specify the previous tag as `v0.9.2`.

.github/workflows/release.yml
pschork: This is a problem. Not just with goreleaser, but also potentially with our existing gitversion implementation. Goreleaser totally chokes on prefixed tags like `node/vx.y.z` as invalid semver. So if we are dead set on using prefixed version tags, we probably need to some new way to parse these tags and always specify current/previous tags sans prefix.

.github/workflows/release.yml
pschork: The implication here is that we release everything. Not against it, but a potential problem to using goreleaser for local/dev builds is that it isn't going to like our prefixed version tagging format.

Makefile
pschork: that cmd is essentially a non-release build. A `goreleaser build` would only build the binaries whereas `goreleaser release` packages the dockers.
I'll remove `docker-release-build`

Dockerfile
pschork: > How is this different from the node target?

Goreleaser builds the binaries and places them in a docker context. The difference between this and the node target is the source copy location is different.

> Why are we keeping the node target around (and the bake file)?

eigenda-devops relies on it to build internal docker images

> Shouldn't we have a single way to build binaries/dockerfiles?

I'd like to unify and consolidate. I'm limiting this round to just focusing on node/nodeplugin as those are part of our public releases and want to test the waters with goreleaser. In the context of this new proposed prefix version tagging we may need to reevaluate our options.

.goreleaser.yaml
pschork: I'm using
```
GitVersion:    2.11.0
GitCommit:     bbd96a824a97ae622b4036c415285625b26e1868
BuildDate:     2025-07-09T00:08:55Z
```

Makefile
pschork: @samlaf added proxy to goreleaser since (now removed) `docker-release-build` was updated to build proxy.

.github/workflows/release.yml
pschork: Looks like we need Goreleaser Pro to support monorepo tag prefixes https://goreleaser.com/customization/monorepo/

.github/workflows/release.yml
samlaf: Thanks for the explanation. IMO such explanations should go in the code itself, otherwise we will lose this information over time, and it will only live in your head.

.github/workflows/release.yml
samlaf: Let's discuss this at our release management meeting this week!
image

.goreleaser.yaml
samlaf: I said goreleaser above but meant github. Can you add github-cli to mise? just have to `git use` and then type github-cli
image

.github/workflows/release.yml
samlaf: Right let's figure that out at the meeting this week.

Dockerfile
samlaf: > Goreleaser builds the binaries and places them in a docker context. The difference between this and the node target is the source copy location is different.

Why do we need these 2 things then? Is there some hardcoded path to the binary somewhere? Can't we just update it to use the other path? 

> I'd like to unify and consolidate. I'm limiting this round to just focusing on node/nodeplugin as those are part of our public releases and want to test the waters with goreleaser. 

Makes sense let's figure this out this week. One thing I'd like to note which is important to keep in mind is that going forward the proxy is also part of our public releases!

PR #1718, opened 2025-07-02
docs: intg spec add blob derivation and preimage for integration spec

bxue-l2:

This PR adds the high level description to eigenda derivation pipeline and use it as the main entrypoint for the secure intg discussion

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


docs/spec/src/assets/integration/eigenda-blob-derivation.png
samlaf, line 1: We prob don't want to talk about op specific stuff here right?
Instead of "check cert validity via preimage oracle" just say "check cert validity via CertVerifier eth-call"?

samlaf: same for "fetch field element via preimage oracle"

docs/spec/src/assets/integration/preimage-derivation-layers.png
samlaf, line 1: This also feels op specific? Should maybe put it under `integration/op/`?
The diagram is pretty awesome though

samlaf: Also I'm not sure that I'm sold on the "eigenda blob derivation" nomenclature. Maybe we can talk about it during standup a bit. Derivation to me is a function from L1 calldata -> L2 state. Whereas for eigenda the "derivation" is for a single cert->blob, so why not just call it "eigenda secure blob retrieval" or something?

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
-When deriving a rollup chain by running its derivation pipeline, only EigenDA `DACerts` that satisfy the three validity conditions are considered valid and used:
-1. RBN Recency Validation - ensure that the DA Cert's reference block number (RBN) is not too old with respect to the L1 block at which the cert was included in the rollup's batcher-inbox. This ensures that the blob on EigenDA has sufficient availability time left (out of the 14 day period) in order to be downloadable if needed during a rollup fault proof window.
-2. Cert Validation - ensures sufficient operator stake has signed to make the blob available, for all specified quorums. The stake is obtained onchain at a given reference block number (RBN) specified inside the cert.
-3. Blob Validation - ensures that the blob used is consistent with the KZG commitment inside the Cert.
+Unlike a trusted integration, a secure integration must handle the malicious data PUT on L1 Ethereum. A malicious batcher can post an invalid or malformed DA certificate (DA Cert) to stall the roll-up derivation pipeline. A malicious proposer can publish wrong L2 state roots. To overcome incorrect data posted on L1,
```
samlaf:
```suggestion
Unlike a trusted integration, a secure integration must be able to handle any malicious data posted on Ethereum L1:
- A malicious batcher can post an invalid or malformed DA certificate (DA Cert) to stall the roll-up derivation pipeline. 
- A malicious proposer can publish wrong L2 state roots. 
To overcome incorrect data posted on L1,
```

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
+we rely on the GET path of the rollup to reliably filter out incorrect data posted by batcher, and challenge the incorrect state root by malicious proposer.
```
samlaf: actually this isn't totally true. You can have a secure integration where the data is filtered by the batcher inbox, like we have for arbitrum

samlaf:
```suggestion
we rely on the GET path of the rollup to reliably filter out incorrect data posted by malicious batcher(s), and challenge incorrect state roots posted by malicious proposer(s).
```

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
 
-If #1 or #2 fails, then the `DA Cert` is treated as invalid and MUST be discarded from the rollup's derivation pipeline.
+We will first introduce the derivation procedures and then discuss two implementations - **eigenda-proxy** and **Hokulea**.
```
samlaf: Again I think this is too op specific for this doc. @ethenotethan 's arb work should also be included in whichever framework we write in this main entry page.

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
+
+## EigenDA blob derivation
+
+This section introduces a canonical procedure for deriving a rollup payload from a DA Cert in the GET path. The diagram below illustrates a step-by-step transition from a raw DA Cert byte string to the final roll-up payload:
+
+- Calldata Input: The pipeline starts with only the DA Cert as byte strings.
```
samlaf:
```suggestion
- Calldata Input: The pipeline starts only containing a serialized DA Cert.
```

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
+- Blob derivation: As the input transitions downward to the rollup payload at the bottom, the DA cert can be routed to one of the terminal states depending on if the DA certs is recent enough, if it is valid or if can be decoded correctly.
```
samlaf:
```suggestion
- Blob derivation: As the input transitions downward to the rollup payload at the bottom, the DA cert can be routed to one of the terminal states depending on if the DA cert is recent enough, if it is valid or if can be decoded correctly.
```

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
+- Preimage oracle: Some procedures during the derivation requires fetching EigenDA blobs and other cert metadata. Access to each piece of data is abstracted behind a pre-image oracle. More see this [section](#eigenda-blob-derivation-and-preimage-oracle).
```
samlaf:
```suggestion
- Preimage oracle: Some procedures during the derivation require fetching EigenDA blobs and other cert metadata. Access to each piece of data is abstracted behind a pre-image oracle. More see this [section](#eigenda-blob-derivation-and-preimage-oracle).
```

samlaf: what is "cert metadata"? Let's be precise

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
+- Host: the name refers to an entity hostinh the preimage server behind the preimage oracle.
```
samlaf:
```suggestion
- Host: entity hosting a preimage server which can reply to preimage oracle requests.
```

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
+
+![](../../assets/integration/eigenda-blob-derivation.png)
+
+### Terminal states
+All inputs to EigenDA derivation pipeline ends in exactly one of four terminal states:
```
samlaf:
```suggestion
All inputs to EigenDA derivation pipeline end in exactly one of four terminal states:
```

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
+| State               | Meaning                                                             |
+| ------------------- | ------------------------------------------------------------------- |
+| **dropped**         | The input is rejected and therefore ignored by the rollup execution |
+| **stalled**         | Required data is temporarily unavailable.                           |
+| **panic**           | A fundamental invariant (e.g., field-element validity) is violated. |
```
samlaf: isn't FE invalidity the only invariant that can be broken?

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
+| **roll-up payload** | Success case when the desired payload bytes are produced.           |
+
+### Failure cases
+
+If an arrow in the diagram is labelled failed, the supplied data did not satisfy EigenDA’s safeguards. The DA Cert is discarded and nothing is forwarded downstream.
+
+A path marked stall means the host cannot (yet) provide the requested data.
+A path marked panic means the blob’s coefficients are not valid BN254 field elements. Then in theory, a KZG commitment cannot be generated, therefore derivation panic.
+
+- parse failed – The batcher submitted an improperly-serialized DA Cert.
```
samlaf: if we want to be super precise its a serialized DA Cert with header bytes (which are different for OP and standard mode)

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
+- recency check failed – The DA Cert reached the roll-up inbox too late.
```
samlaf: or the eigenda disperser chose an RBN that is too old

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
+- cert validity check failed – Either the certificate does not satisfy the [quorum-attestation constraint](#2-cert-validation) or the host lies about its validity via the preimage oracle.
```
samlaf: "host lies about its validity via the preimage oracle"? What does this mean? Is this specific to your current op implementation?

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
+- decode blob failed – The host supplies an EigenDA blob that cannot be decoded into back to rollup payload per the [spec](./3-datastructs.md#data-structs). This can occur when
+  - the batcher intentionally corrupts the encoding, or
+  - the host transmits incorrect data.
+
+But if none of these errors arise, the derivation pipeline outputs the expected payload.
```
samlaf:
```suggestion
If none of these error cases arise, the derivation pipeline outputs the expected payload.
```

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
+
+If an arrow in the diagram is labelled failed, the supplied data did not satisfy EigenDA’s safeguards. The DA Cert is discarded and nothing is forwarded downstream.
+
+A path marked stall means the host cannot (yet) provide the requested data.
+A path marked panic means the blob’s coefficients are not valid BN254 field elements. Then in theory, a KZG commitment cannot be generated, therefore derivation panic.
+
+- parse failed – The batcher submitted an improperly-serialized DA Cert.
+- recency check failed – The DA Cert reached the roll-up inbox too late.
+- cert validity check failed – Either the certificate does not satisfy the [quorum-attestation constraint](#2-cert-validation) or the host lies about its validity via the preimage oracle.
+- decode blob failed – The host supplies an EigenDA blob that cannot be decoded into back to rollup payload per the [spec](./3-datastructs.md#data-structs). This can occur when
```
samlaf:
```suggestion
- decode blob failed – The host supplies an EigenDA blob that cannot be decoded back into a rollup payload as per the [spec](./3-datastructs.md#data-structs). This can occur when
```

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
+  - the batcher intentionally corrupts the encoding, or
+  - the host transmits incorrect data.
+
+But if none of these errors arise, the derivation pipeline outputs the expected payload.
+
+## EigenDA Blob derivation and Preimage Oracle
+
+We have defined the canonical derivation process, and now we described two concrete implementations and their preimage oracle:
```
samlaf:
```suggestion
We have defined the canonical derivation process, and now we describe two concrete implementations and their preimage oracle:
```

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
+
+### EigenDA-proxy (Go)
+
+Eigenda proxy is the first implementation of the eigenda blob derivation described above, it is also the first implementation  of the  preimage oracle
+server that uses the external RPC to implement the preimage oracle. The proxy also contains services like http server, secondary storage and more features on [proxy github page](https://github.com/Layr-Labs/eigenda-proxy?tab=readme-ov-file#features-and-configuration-options-flagsenv-vars). Below we highlight some parts around derivation pipeline
```
samlaf: wdym by "the external RPC"?

samlaf:
```suggestion
server that uses the external RPC to implement the preimage oracle. Below we highlight some parts around derivation pipeline
```

samlaf: doesn't belong here imo. Also "http server" is not "a service", its "the only service" that proxy provides

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
+- the calldata input in proxy is received from http listening port (typically the requester is a roll-up consensus node).
+- proxy has both ETH RPC and EigenDA network RPC to get the preimage like the diagram below
```
samlaf: "get the preimage". You never mentioned anywhere what the "preimage" means. Like its a preimage to what? the serialized cert. Typically preimage is preimage of a hash, so this is a bit of a stretch. So we should define it at the very beginning imo

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
+- Queries an Ethereum RPC endpoint to validate the cert. In practice, it is handled by implemented by `verification` Go package.
```
samlaf:
```suggestion
- Queries an Ethereum RPC endpoint to validate the cert.
```

samlaf: This should only be a spec imo. We should stay high-level. You could just mention in the beginning that the proxy is just a simple REST server that wraps our eigenda clients, and that all the logic is implemented in the clients, which could be used directly instead of via the proxy.

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
+- the EigenDA blob can be retrieved from EigenDA network either directly with an EigenDA relay node RPC or via distributed retrieval by communicating directly to the endpoints exposed by EigenDA validators. In practice, it is handled by implemented by `payloadretrieval` Go package.
```
samlaf:
```suggestion
- the EigenDA blob can be retrieved from the EigenDA network either directly via an EigenDA relay node RPC or via distributed retrieval by communicating directly to the endpoints exposed by EigenDA validators.
```

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
+- if everything succeeds, the roll-up payload is returned by the proxy to the requester via the HTTP response.
+
+Eigenda-Proxy is the implemention which rollup consensus node runs in the normal mode for optimistic rollup.
```
samlaf: this should be the very first sentence of this section imo

samlaf:
```suggestion
Eigenda-Proxy is the implemention which rollup consensus nodes run in the normal mode for optimistic rollup.  
```

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
+- if everything succeeds, the roll-up payload is returned by the proxy to the requester via the HTTP response.
+
+Eigenda-Proxy is the implemention which rollup consensus node runs in the normal mode for optimistic rollup.  
+
+![](../../assets/integration/preimage-derivation-layers.png)
+
+### Hokulea (Rust)
+
+We have developed another implementation of EigenDA blob derivation for OP stack, primarily for securing OP stack rollup with optimistic fault-proof,
+ZK Fault Proof or ZK proof. The hokulea client (and crates it uses) implements the identical eigenda blob derivation logics as mentioned in the [eigenda blob derivation section](#eigenda-blob-derivation), the client is intended to be imported as a library into OP consensus rust implementation [Kona](https://github.com/op-rs/kona).
```
samlaf:
```suggestion
We have developed another implementation of EigenDA blob derivation for OP stack, which supports all of optimistic and ZK fault proof,
as well as validity ZK proofs. The hokulea client (and crates it uses) implement the eigenda blob derivation logic described in the [eigenda blob derivation section](#eigenda-blob-derivation). The client is intended to be imported as a library into OP consensus rust implementation [Kona](https://github.com/op-rs/kona).
```

samlaf: TBH I feel like all of this shouldn't be here. It probably should just point to our op spec page?

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
+
+
+The major difference between the two implementations comes to how the preimage oracle is abstracted and implemented.
+For hokulea, the interface is abstracted as key-value map to make preimage oracle verifiable on L1 Ethereum,
+whose relation is verifiable on L1 Ethereum.
```
samlaf:
```suggestion
For Hokulea, the interface is abstracted as key-value map to make preimage oracle verifiable on L1 Ethereum.
```

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,106 @@
...
+Proxy must be able to return the encoded payload alone, though the specific implementation has not been decided yet.
+As described in [data-struct page](./3-data-structs.md). During the creation of a DA cert a rollup payload, the data is first transformed into the encoded paylod
+then undergo IFFT transform into the right polynomial format before sending to EigenDA that created a DA cert.
+
+Normally the DecodeBlob functional interface on the GET path of the proxy does the FFT transformation on the received EigenDA blob to convert into the encoded
+payload which is the original polynomial format.
+To remove redundant work which the upper layer (Hokulea host) has to do (i.e. the FFT step), the proxy when used as a component for providing preimage, it returns
+the encoded payload as opposed to the raw eigenda blob, which has the IFFT transformation, which is needed for optimistic challenge for opening points onchain.
+
+In the following sections, We provide the definition for each check in the diagram above.
```
samlaf:
```suggestion
In the following sections, we provide the definition for each check in the diagram above. 
```

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: The intent is yo highlight the fact that proxy relies on those rpc as preimage, do you get that part?

docs/spec/src/assets/integration/preimage-derivation-layers.png
bxue-l2: I think it is also common for the derivation to contain a part that gets the rollup payload from the blob.

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: What is your question here?

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: Let us not.

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: I think it's implied.

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: In Nitro M2 integration where cert valitidy is checked offchain, the preimage host can lie about the validity of the cert, @ethenotethan ?

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: I want to emphasize the fact that Proxy is also reliant on some preimage oracle. I can provide a separate definition for it.

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: I wanna provide an example how preimage oracle is abstracted. Following up to your comments above, I think in the beginning I should have a section for that. I agree that it would be nice to have the host client into the OP integration part of the spec.

docs/spec/src/integration/spec/6-secure-integration.md
ethenotethan: +1 - we should also be using terms defined in the glossary to make it more rollup agnostic and be better compliant with other integration spec terminologies

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,108 @@
...
+- A malicious proposer can publish wrong L2 state roots. 
+To overcome incorrect data posted on L1,
+we rely on the GET path of the rollup to reliably filter out incorrect data posted by malicious batcher(s), and challenge incorrect state roots posted by malicious proposer(s).
 
-If #1 or #2 fails, then the `DA Cert` is treated as invalid and MUST be discarded from the rollup's derivation pipeline.
+We will first introduce the derivation procedures and then discuss two implementations - **eigenda-proxy** and **Hokulea**.
+
+## EigenDA blob derivation
+
+This section introduces a canonical procedure for deriving a rollup payload from a DA Cert in the GET path. The diagram below illustrates a step-by-step transition from a raw DA Cert byte string to the final roll-up payload:
```
ethenotethan:
```suggestion
This section introduces a canonical procedure for deriving a rollup payload from a DA Cert in the GET path. The diagram below illustrates a step-by-step transition from a raw DA Cert byte string to the final rollup payload:
```

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,108 @@
...
+![](../../assets/integration/eigenda-blob-derivation.png)
+
+### Terminal states
+All inputs to EigenDA derivation pipeline end in exactly one of four terminal states:
+| State               | Meaning                                                             |
+| ------------------- | ------------------------------------------------------------------- |
+| **dropped**         | The input is rejected and therefore ignored by the rollup execution |
+| **stalled**         | Required data is temporarily unavailable.                           |
+| **panic**           | A fundamental invariant (e.g., field-element validity) is violated. |
+| **roll-up payload** | Success case when the desired payload bytes are produced.           |
```
ethenotethan:
```suggestion
| **rollup payload** | Success case when the desired payload bytes are produced.           |
```

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,108 @@
...
+
+### Failure cases
+
+If an arrow in the diagram is labelled failed, the supplied data did not satisfy EigenDA’s safeguards. The DA Cert is discarded and nothing is forwarded downstream.
```
ethenotethan: https://www.notion.so/eigen-labs/M2-x-V2-Arbitrum-x-EigenDA-Design-1bf13c11c3e080ccb2d8d3a3fc15b780?source=copy_link#1ca13c11c3e080e68744cf577ac9ba73

docs/spec/src/integration/spec/6-secure-integration.md
ethenotethan: @bxue-l2 that is correct, with M2 we plan to do a prefix encoding of the cert's validity into the sequencer message bytes

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,108 @@
...
+
+### EigenDA-proxy (Go)
+
+Eigenda proxy is the first implementation of the eigenda blob derivation described above, it is also the first implementation  of the  preimage oracle
+server that uses the external RPC to implement the preimage oracle. Below we highlight some parts around derivation pipeline
+- the calldata input in proxy is received from http listening port (typically the requester is a roll-up consensus node).
+- proxy has both ETH RPC and EigenDA network RPC to get the preimage like the diagram below
+- Queries an Ethereum RPC endpoint to validate the cert.
+- the EigenDA blob can be retrieved from the EigenDA network either directly via an EigenDA relay node RPC or via distributed retrieval by communicating directly to the endpoints exposed by EigenDA validators.
+- if everything succeeds, the roll-up payload is returned by the proxy to the requester via the HTTP response.
```
ethenotethan: is this worth enumerating? seems like duplication with existing retrieval lifecycle spec

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: Wait, does it imply that the sequencer posted something on L1 but claimed the posted thing is invalid?

docs/spec/src/integration/spec/6-secure-integration.md
samlaf: you say that panic happens when "A fundamental invariant (e.g., field-element validity) is violated.".
Isnt FE invalidity the only invariant that can be broken? According to the diagram at least. So shouldnt say "e.g.", it should just say "panic": Field-element validity is violated.

docs/spec/src/integration/spec/6-secure-integration.md
samlaf: why not? this is a spec

docs/spec/src/integration/spec/6-secure-integration.md
samlaf: It's not.

docs/spec/src/integration/spec/6-secure-integration.md
samlaf: agree with this. I've never seen `roll-up` in two words like that

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: removed the imprecise paragraphs

docs/spec/src/assets/integration/eigenda-blob-derivation.png
bxue-l2: See the new graph. Remove the panic because I think it is too detailed. And it might as well be part of the preimage request or Blob Decoding.

docs/spec/src/assets/integration/preimage-derivation-layers.png
bxue-l2: I have moved the OP Hokulea diagram into its own section.

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: 02ec55a415be21222d8557374cf3a11f7d882e86

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: The whole paragraph was rewritten and reviewed by Claude.

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: Removed

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: WisprFlow likes -, Now everything went through a claude. This should not happen.

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: See the new version. The definition of Premier Drop Record should be clear in the beginning.

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: I would just make it precise.

docs/spec/src/assets/integration/hokulea-preimage-derivation-impl.png
samlaf, line 1: Should we add the recency_window inside the hokulea client also? Since it won't be part of the onchain preimage oracle right?

docs/spec/src/glossary.md
```diff
@@ -14,4 +14,20 @@ Side car server as a part of rollup and used for secure and trustless communicat
 
 ## EigenDA Client
 
-Refers to collection of client [modules](https://github.com/Layr-Labs/eigenda/tree/bb91b829995c28e813fce46412a77f9fa428b0af/api/clients/v2) used for securely dispersing and reading EigenDA blobs. 
\ No newline at end of file
+A collection of client [modules](https://github.com/Layr-Labs/eigenda/tree/bb91b829995c28e813fce46412a77f9fa428b0af/api/clients/v2) used for securely dispersing and reading EigenDA blobs.
```
samlaf: "modules" is wrong. Those are golang packages, but they aren't even nicely separated into packages.

samlaf:
```suggestion
A collection of [clients](https://github.com/Layr-Labs/eigenda/tree/bb91b829995c28e813fce46412a77f9fa428b0af/api/clients/v2) used for securely dispersing and reading EigenDA blobs.
```

docs/spec/src/glossary.md
```diff
@@ -14,4 +14,20 @@ Side car server as a part of rollup and used for secure and trustless communicat
...
+
+## Rollup Payload
+
+Compressed batches of transactions or state transition diffs.
```
samlaf:
```suggestion
Compressed batches of transactions or state diffs.
```

samlaf: I think its either "state diff" or "state transition"? since the transition implies its a state diff.

docs/spec/src/glossary.md
```diff
@@ -14,4 +14,20 @@ Side car server as a part of rollup and used for secure and trustless communicat
...
+
+An EigenDA Certificate (or DACert for short) contains all the information needed to retrieve a blob from the EigenDA network and validate it.
+
+## EigenDA Blob Derivation
+
+A sequence of procedures to convert a byte array representing a DA certificate to the final rollup payload.
+
+## Preimage Oracle
+
+An interface for fetching additional data during EigenDA blob derivation.
```
samlaf: I wouldn't say its just an interface. Prob best to mention that it has 2 implementations (proxy vs onchain contract)?

docs/spec/src/integration/rollup-stacks/2-op-hokulea-secure-integration.md
```diff
@@ -0,0 +1,33 @@
+### Hokulea
+
+Hokulea provides a Rust implementation of EigenDA blob derivation for the OP stack. The Hokulea client (and its associated crates) implements the EigenDA blob derivation logic described in the [EigenDA blob derivation section](#eigenda-blob-derivation). The client is designed to be imported as a library into the OP consensus Rust implementation [Kona](https://github.com/op-rs/kona).
+
+Since the OP rollup inbox is not a smart contract, the secure integration requires EigenDA blob derivation to take place entirely offchain (see the design rationale in [secure integration](../spec/6-secure-integration.md#secure-integration-framework)). Depending on the choice of VM and game type, Hokulea can support optimistic fault proofs, ZK fault proofs, and validity ZK proofs.
```
samlaf:
```suggestion
Since the OP rollup inbox is not a smart contract, the secure integration requires EigenDA blob derivation to take place entirely offchain (see the design rationale in [secure integration](../spec/6-secure-integration.md#secure-integration-framework)). Depending on the choice of VM and game type, Hokulea can support optimistic interactive fault proofs and ZK fault proofs, as well as validity ZK proofs.
```

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,122 @@
...
-3. Blob Validation - ensures that the blob used is consistent with the KZG commitment inside the Cert.
+A secure integration must handle malicious data posted on Ethereum L1, unlike trusted integrations. Potential threats include:
 
-If #1 or #2 fails, then the `DA Cert` is treated as invalid and MUST be discarded from the rollup's derivation pipeline.
+- **Malicious batcher:** Posts invalid or malformed DA certificates (DA Cert)
+- **Malicious proposer:** Publishes incorrect L2 state roots 
+
+## EigenDA Blob Derivation
+
+This section describes the canonical procedure for deriving a rollup payload from a DA Certificate. This derivation is integral to rollup consensus and must be implemented in both rollup nodes and secure integrations.
```
samlaf: Unclear what "implemented in secure itnegrations" means. Do we have another name for this? Or shall we just say "challenger" (because that is the binary that ultimately makes the integration secure I think?)

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,122 @@
...
+
+### Current Implementations
+
+- **EigenDA Proxy**
+- **OP EigenDA Secure Integration** ([Hokulea](https://github.com/Layr-Labs/hokulea/tree/master))
```
samlaf, line 19: we have to figure out whether and if so how @ethenotethan 's arbitrum implementation fits into this.

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,122 @@
...
+
+### Derivation Process
+
+The diagram below shows the step-by-step transformation from input to final rollup payload:
+
+**Key Components:**
+- **Input:** Serialized DA Cert (as calldata) + block number of DA Cert inclusion
+- **Blob Derivation:** Routes DA cert through validation to one of several terminal states
+- **Preimage Oracle:** Interface for fetching additional data during derivation
+  - Implementation varies by requirement (e.g., key-value mapping for optimistic fault proofs)
```
samlaf, line 29: isnt op fault proofs the only implementation?
Or do you mean different stacks have different preimage oracles? dont think zksync falls into this model for example. Does arbitrum?

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,122 @@
...
 
-## Validity Conditions
+## Overview
 
-EigenDA is a service that assures the availability and integrity of payloads posted to it for 14 days.
-When deriving a rollup chain by running its derivation pipeline, only EigenDA `DACerts` that satisfy the three validity conditions are considered valid and used:
-1. RBN Recency Validation - ensure that the DA Cert's reference block number (RBN) is not too old with respect to the L1 block at which the cert was included in the rollup's batcher-inbox. This ensures that the blob on EigenDA has sufficient availability time left (out of the 14 day period) in order to be downloadable if needed during a rollup fault proof window.
-2. Cert Validation - ensures sufficient operator stake has signed to make the blob available, for all specified quorums. The stake is obtained onchain at a given reference block number (RBN) specified inside the cert.
-3. Blob Validation - ensures that the blob used is consistent with the KZG commitment inside the Cert.
```
samlaf, line 11: I still feel that this is accurate, at least as a high-level overview. I think its a looot more useful to keep this here, as a first-step mental model for people. Otherwise they will just get lost in the details.

Why do you feel its not accurate anymore actually? iiuc the derivation pipeline is still only doing this, + parsing. But thats it.

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,122 @@
...
+![](../../assets/integration/eigenda-blob-derivation.png)
+
+### Terminal States
+
+All inputs to the EigenDA derivation pipeline end in exactly one of these states:
+
+| State | Description |
+|-------|-------------|
+| **Dropped** | Input rejected and ignored by rollup execution |
+| **Stalled** | Required preimage data temporarily unavailable |
```
samlaf:
```suggestion
| **Stalled** | Required preimage data unavailable |
```

samlaf: not necessarily temporary right?

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,122 @@
...
+#### Parse Failed
+- Batcher submitted improperly-serialized or unrecognized DA Cert
+
+#### Recency Check Failed
+- DA Cert reached rollup inbox after reference block number + recency window 
+- Host provides false recency window size that leads to failure
+
+#### Cert Validity Check Failed
+- Certificate doesn't satisfy [quorum-attestation constraint](#2-cert-validation)
+- Host provides false validity information via preimage oracle
```
samlaf, line 57:
```suggestion
- Host provides an incorrect validity boolean via preimage oracle
```

samlaf: Actually what is this trying to say? That the host-provided validity boolean doesn't match the one provided from the onchain preimage oracle?

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,122 @@
...
+
+#### Parse Failed
+- Batcher submitted improperly-serialized or unrecognized DA Cert
+
+#### Recency Check Failed
+- DA Cert reached rollup inbox after reference block number + recency window 
+- Host provides false recency window size that leads to failure
+
+#### Cert Validity Check Failed
+- Certificate doesn't satisfy [quorum-attestation constraint](#2-cert-validation)
```
samlaf: broken link I think?

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,122 @@
...
+- Host provides false validity information via preimage oracle
+
+#### Decode Blob Failed
+- EigenDA blob cannot be decoded back to rollup payload per [spec](./3-datastructs.md#data-structs)
+- Causes:
+  - Batcher intentionally corrupts encoding
+  - Host transmits incorrect data
```
samlaf:
```suggestion
  - Host or Batcher intentionally corrupts encoding
```

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,122 @@
...
+
+**Success:** If no failures occur, the pipeline outputs the expected payload.
+
+## Secure Integration Framework
+
+Rollup consensus must cover all aspects of EigenDA blob derivation. Designers have two degrees of freedom:
+
+1. **Derivation Split:** Choose which parts are checked on-chain (pessimistically) vs off-chain (optimistically)
```
samlaf: a validity rollup providing a zk proof that is verified onchain would count as pessimistic right?

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,122 @@
...
+2. **Proving VM Choice:** Select the on-chain proving virtual machine
+
+Each integration can be tailored to fit specifi rollup protocol constraints.
+
+### Splitting EigenDA Blob Derivation
+
+Rollups can split the derivation pipeline between on-chain and off-chain verification:
+
+**Examples:**
+- **Arbitrum with EigenDA V1:** All components through cert validity checked in rollup inbox
```
samlaf, line 79:
```suggestion
- **Arbitrum with EigenDA V1:** All steps up to the cert validity are checked onchain in the rollup inbox
```

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,122 @@
...
+- **OP Integration:** Entire EigenDA blob derivation takes place off-chain
```
samlaf: not necessarily if using op succinct with validity proofs that are checked pessimistically for every cert/blob pair.

docs/spec/src/integration/rollup-stacks/2-op-hokulea-secure-integration.md
```diff
@@ -0,0 +1,33 @@
...
+
+
+![](../../assets/integration/hokulea-preimage-derivation-impl.png)
+
+#### Preimage Oracle Architecture
+
+The major difference between implementations lies in how the preimage oracle is abstracted and implemented. In Hokulea, the interface is abstracted as a key-value map to make the preimage oracle verifiable on L1 Ethereum.
```
samlaf: do we have any other example? Otherwise no need to generalize like this

docs/spec/src/integration/rollup-stacks/2-op-hokulea-secure-integration.md
```diff
@@ -0,0 +1,33 @@
...
+
+The Hokulea preimage host for the key-value oracle interface is built on top of the EigenDA proxy (see diagram above). The proxy handles all the heavy lifting to retrieve the actual preimage data, while the Hokulea host serves as a thin layer that translates HTTP status codes into preimage data or errors.
```
samlaf:
```suggestion
The Hokulea preimage host for the key-value oracle interface communicates with the EigenDA proxy (see diagram above). The proxy handles all the heavy lifting to retrieve the actual preimage data, while the Hokulea host serves as a thin layer that translates HTTP status codes into preimage data or errors.
```

docs/spec/src/integration/rollup-stacks/2-op-hokulea-secure-integration.md
```diff
@@ -0,0 +1,33 @@
...
+
+#### Communication Between Hokulea Host and EigenDA Proxy
+
+The proxy uses an HTTP interface to serve as a base layer for abstraction. The proxy exposes the following JSON status codes in addition to HTTP status codes to convey information about the preimage:
+
+| Message             | HTTP Status Code | JSON Status Code | Indication           |
```
samlaf:
```suggestion
The proxy uses an HTTP interface to serve as a base layer for abstraction. The proxy exposes the following app-layer status codes in addition to HTTP status codes to convey information about the preimage:

| Message             | HTTP Status Code | App-Layer Status Code | Indication           |
```

samlaf: not strictly necessary but when I first saw "json status codes" I was confused thinking these were json-rpc status codes.

docs/spec/src/integration/rollup-stacks/2-op-hokulea-secure-integration.md
```diff
@@ -0,0 +1,33 @@
...
+| ------------------- | ---------------- | ---------------- | -------------------- |
+| **Decoded blob (rollup payload)** | 200 | N/A | Successful request |
+| **Certificate validity** | 418 | 1 | Certificate is invalid |
+| **Certificate recency** | 418 | 2 | Certificate is too old |
+| **Encoded payload** | 418 | 3 | Blob decoding error |
+
+#### Encoded Payload vs. Decoded Blob
+
+For developers familiar with the EigenDA proxy: on a default GET http query, the proxy returns the decoded blob (the rollup payload) as a byte string in an HTTP 200 response. However, to enable the proxy as part of the preimage oracle for other Hokulea, the preimage data must be a valid blob polynomial where every 32 bytes is a valid field element on BN254 (the [encoded payload](../spec/3-data-structs.md)).
```
samlaf: "to enable the proxy as part of the preimage oracle for other Hokulea"
Not sure what this is trying to say

docs/spec/src/integration/rollup-stacks/2-op-hokulea-secure-integration.md
```diff
@@ -0,0 +1,33 @@
...
+
+The proxy must be able to return the encoded payload independently.
+
+To eliminate redundant work that the upper layer (Hokulea host) would otherwise need to perform (the FFT step), the proxy needs to convert the EigenDA blob into encoded payload.
```
samlaf, line 33: add a TODO to clean this up once we add the new endpoint/query_params?

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,122 @@
...
+3. **Security:** Off-chain derivation secured by proof system
+4. **Complete Coverage:** Combined on-chain and off-chain logics covers entire derivation
+
+**Preimage Oracle Requirement:** Both on-chain and off-chain implementations needed. 
+
+![](../../assets/integration/secure-integration-model.png)
+
+### ZKVM for ZK Fault Proof
+
+**Key Difference:** ZKVMs don't require preimage oracles since all preimage verification can be attached to EigenDA blob derivation usable as a library.
```
samlaf: dont understand what this means

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,122 @@
...
+
+**Hybrid Approach:** Secure integrations can theoretically combine both ZKVM and optimistic fault proof VMs.
+
+## EigenDA Blob Derivation in EigenDA Proxy
+
+We have dedicated page for secure integrations, but let's review the **EigenDA Proxy** GET path implementation, which has been used in rollup consensus nodes since EigenDA integration began. Proxy also implements WRITE path, solely used by rollup batcher for rollup liveness.
```
samlaf: add link to the page

docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -1,16 +1,122 @@
...
+- **OP Integration:** Entire EigenDA blob derivation takes place off-chain 
+
+### VM for Optimistic Fault Proof
+
+In order to secure parts from the EigenDA blob derivation taking place off-chain:
+
+1. **Integration Required:** Blob derivation must be imported into L2 consensus as a library
+2. **Compilation:** The library compiles to instructions replayable by on-chain VM
+3. **Security:** Off-chain derivation secured by proof system
+4. **Complete Coverage:** Combined on-chain and off-chain logics covers entire derivation
```
samlaf: This is also true for zkvms right? Feel like this is good and should be moved up somewhere more prominent?

samlaf (commented):
Lots of small comments, but overall this is super super nice as is.

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: panic is removed

docs/spec/src/glossary.md
bxue-l2: sg

docs/spec/src/integration/rollup-stacks/2-op-hokulea-secure-integration.md
bxue-l2: I think it's generally true for anything you want to be provable on Ethereum? For example, Arbitrum also relies on key-value.

bxue-l2: But it would be good if we get some solid confirmation from Arbitrum's side.

docs/spec/src/integration/rollup-stacks/2-op-hokulea-secure-integration.md
bxue-l2: to integrate

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: This derivation is integral to rollup consensus and must be integrated in both rollup nodes and the proof system in secure integrations.

is it clearer

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: I am thinking that Arbitrum can find a way to fit the Nitro integration into the framework using eigenDA derivation pipeline. Then in rollup-stack adding more details

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: I believe Nitro has some notion of preimage oracle.
If zksync really doesn't need the preimage oracle, (just like OP ZK integration does not need it), you can ignore it. But honestly, at the lower layer, the preimage oracle shows up, because you have to figure out a way to get a blob and cert validity.

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: How about "| **Stalled** | Required preimage data unavailable at the moment, and it should be retried |"

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: fixed

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: The host provides an image that is against the truth

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: oh, that is a good point. Yes, it is everything pessimistic.

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: I guess what I really want to emphasize is what part is handled by the proving system, versus all parts are handled by the native L1 consensus by executing it.

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: Still think this is a good place because in the immediate section above I introduced the splitting the duration.

bxue-l2: But I'm considered daring to move it because it looks like it's true for all integrations.

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: Maybe another way to make a distinction is the execution is taking place of change.

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: We'll reorganize it.

docs/spec/src/integration/spec/6-secure-integration.md
bxue-l2: I feel this is an incomplete treatment on the filtering rule.
The first and second points are briefly described in the section before under **Failure Cases**.

We currently didn't enforce the block validation within the derivation, that part isn't accurate. I feel it is not in the right presentation. 

The 3 properties is more like a preimage property mixed with some derivation property, really not sure how to include this.

docs/spec/src/assets/integration/hokulea-preimage-derivation-impl.png
bxue-l2: added

docs/spec/src/glossary.md
bxue-l2: https://github.com/Layr-Labs/eigenda/pull/1718/commits/57a148ec9194f1cae9faee53eca229e71bc4b0cf#diff-ca4fa1567c17a4dfddf313e59215b4a36ecc70b7cf7c328cf02be4d2edcc2ab0R33

docs/spec/src/integration/spec/6-secure-integration.md
samlaf: yes

samlaf (approved):

PR #1719, opened 2025-07-02
fix: Update icicle encoder go version and fix build issue

dmanc:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/cmd/encoder/icicle.Dockerfile
+++ b/disperser/cmd/encoder/icicle.Dockerfile
@@ -1,11 +1,10 @@
+# This Dockerfile has been tested on Ubuntu 24.04
+# Note: Will fail on macOS with "gcc: error: unrecognized command-line option '-m64'" during cgo compilation
 FROM nvidia/cuda:12.2.2-devel-ubuntu22.04 AS builder
 
-# Install Go
-# TODO: this probably won't work given that we've updated go.mod to use go 1.24.
-# However `docker buildx bake encoder-icicle` is failing on current master (0a61560a77)
-# so I'm a bit confused about the state of this Dockerfile...
-ENV GOLANG_VERSION=1.21.13
-ENV GOLANG_SHA256=502fc16d5910562461e6a6631fb6377de2322aad7304bf2bcd23500ba9dab4a7
+# Install Go 1.24.4 to match go.mod requirements
+ENV GOLANG_VERSION=1.24.4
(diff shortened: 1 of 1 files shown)


disperser/cmd/encoder/icicle.Dockerfile
```diff
@@ -1,11 +1,10 @@
+# This Dockerfile has been tested on Ubuntu 24.04
+# Note: Will fail on macOS with "gcc: error: unrecognized command-line option '-m64'" during cgo compilation
```
samlaf:
```suggestion
# Note: Will fail on macOS with "gcc: error: unrecognized command-line option '-m64'" during cgo compilation, which is expected because cuda is not available.
```

disperser/cmd/encoder/icicle.Dockerfile
```diff
@@ -1,11 +1,10 @@
...
 FROM nvidia/cuda:12.2.2-devel-ubuntu22.04 AS builder
 
-# Install Go
-# TODO: this probably won't work given that we've updated go.mod to use go 1.24.
-# However `docker buildx bake encoder-icicle` is failing on current master (0a61560a77)
-# so I'm a bit confused about the state of this Dockerfile...
-ENV GOLANG_VERSION=1.21.13
-ENV GOLANG_SHA256=502fc16d5910562461e6a6631fb6377de2322aad7304bf2bcd23500ba9dab4a7
+# Install Go 1.24.4 to match go.mod requirements
+ENV GOLANG_VERSION=1.24.4
```
samlaf, line 6: nit: this is actually not important. you could even use 1.21 and it would work fine because go binary is forward compatible now, and will download the correct version pointed to by the go.mod file: https://go.dev/blog/toolchain

disperser/cmd/encoder/icicle.Dockerfile
```diff
@@ -19,10 +18,13 @@ WORKDIR /app
 # Copy go.mod and go.sum first to leverage Docker cache
 COPY go.mod go.sum ./
 
+# Copy api/proxy/clients for the replace directive
+COPY api/proxy/clients ./api/proxy/clients
```
samlaf, line 22: Interesting this probably shouldnt be needed, but totally fine for now. will think through whether there's better ways to deal with this dependency though.

PR #1722, opened 2025-07-03
fix: load generator improvements

samlaf:

Many small improvements + bug fixes for the load generator.
Please see the commit details for more description on each error/fix.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/cert_builder.go
+++ b/api/clients/v2/cert_builder.go
@@ -35,6 +35,16 @@ func NewCertBuilder(
 	if logger == nil {
 		return nil, fmt.Errorf("logger cannot be nil")
 	}
+	if ethClient == nil {
+		return nil, fmt.Errorf("ethClient cannot be nil")
+	}
+	if opsrAddr == (gethcommon.Address{}) {
+		return nil, fmt.Errorf("opsrAddr cannot be empty")
+	}
+	if registryCoordinatorAddr == (gethcommon.Address{}) {
+		return nil, fmt.Errorf("registryCoordinatorAddr cannot be empty")
+	}
+
 	// Create the Operator State Retriever caller
 	opsrCaller, err := opsrbinding.NewContractOperatorStateRetrieverCaller(opsrAddr, ethClient)
 	if err != nil {
--- a/api/clients/v2/cert_builder.go
+++ b/api/clients/v2/cert_builder.go
@@ -2,8 +2,10 @@ package clients
 
 import (
 	"context"
+	"encoding/hex"
 	"fmt"
 	"math/big"
+	"strings"
 
 	"github.com/Layr-Labs/eigenda/api/clients/v2/coretypes"
 	coreEth "github.com/Layr-Labs/eigenda/core/eth"
@@ -21,6 +23,7 @@ import (
 type CertBuilder struct {
 	logger                  logging.Logger
 	opsrCaller              *opsrbinding.ContractOperatorStateRetrieverCaller
+	opsrAddr                gethcommon.Address
 	registryCoordinatorAddr gethcommon.Address
 }
 
@@ -54,6 +57,7 @@ func NewCertBuilder(
 	return &CertBuilder{
 		logger:                  logger,
 		opsrCaller:              opsrCaller,
+		opsrAddr:                opsrAddr,
 		registryCoordinatorAddr: registryCoordinatorAddr,
 	}, nil
 }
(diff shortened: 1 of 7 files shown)


test/v2/config/environment/preprod.json
```diff
@@ -1,5 +1,5 @@
 {
-  "SRSPath": "../../../resources/srs",
```
samlaf: Relative paths are evil. Not sure how this was working though... I'm running this from the Makefile in `test/v2`, so should be `../..` to reach root dir. We should prob use relative paths wrt root though?

api/clients/v2/cert_builder.go
```diff
@@ -113,6 +127,20 @@ func (cb *CertBuilder) getNonSignerStakesAndSignature(
 	rbn := signedBatch.GetHeader().GetReferenceBlockNumber()
 
 	// 3 - call operator state retriever to fetch signature indices
+	nonSignerOperatorIDsHex := make([]string, len(nonSignerOperatorIDs))
+	for i, id := range nonSignerOperatorIDs {
+		nonSignerOperatorIDsHex[i] = "0x" + hex.EncodeToString(id[:])
+	}
+	// We log the call parameters for debugging purposes. If execution reverts, we can debug using tenderly.
+	cb.logger.Info("eth-call",
```
dmanc: How would debugging work? Do I have to manually convert to calldata or something like that?

test/v2/load/load_generator_metrics.go
```diff
@@ -48,7 +54,7 @@ func newLoadGeneratorMetrics(registry *prometheus.Registry) *loadGeneratorMetric
 		[]string{},
 	)
 
-	relayReadSuccesses := promauto.With(registry).NewCounterVec(
+	relayReadSuccesses := promauto.With(registerer).NewCounterVec(
```
dmanc, line 57: nit: internet says registrar is more common

dmanc (approved):

api/clients/v2/cert_builder.go
samlaf: Idea right now it to use tenderly, which loads the ABI from etherscan and allows to input each field individually.
I would like to also log the calldata directly such that we could also use tenderly or whatever other tool even without the ABI, but that currently doesn't work with the abigen generated code which hides the conversion logic to calldata.

I do plan eventually to switch to abigen v2 which would make this possible though.

test/v2/load/load_generator_metrics.go
samlaf: I'm just following the prometheus defined interface:
image

api/clients/v2/cert_builder.go
```diff
@@ -113,6 +127,20 @@ func (cb *CertBuilder) getNonSignerStakesAndSignature(
...
+		"contract", "OperatorStateRetriever",
+		"contractAddr", cb.opsrAddr.Hex(),
+		"method", "GetCheckSignaturesIndices",
+		"registryCoordinatorAddr", cb.registryCoordinatorAddr.Hex(),
+		"referenceBlockNumber", rbn,
+		"quorumNumbers", "0x"+hex.EncodeToString(quorumNumbers),
+		"nonSignerOperatorIDs", "["+strings.Join(nonSignerOperatorIDsHex, ",")+"]",
+	)
```
litt3: Would it make sense to only log this if there is actually an error?

test/v2/config/environment/preprod.json
litt3: The relative path was correct for running the tests from the IDE (confirmed by trying to run the tests, and they fail).

litt3 (approved):

test/v2/config/environment/preprod.json
cody-littley: I agree this is ugly. I'd have documented this better, but json doesn't allow comments (😠).

I'd prefer to leave the paths so that running the test from an IDE continues to work. It's often quite helpful to run a failing test with a debugger, and it's way more convenient to do that directly from the IDE. The way to get this to also work from the makefile is to run it from the same directory the IDE runs it from.

test/v2/Makefile
```diff
@@ -1,11 +1,17 @@
...
+# Makefile doesn't allow forwarding of arguments, so we use ARGS. Call this as:
+# make generate-load ARGS="config/environment/preprod.json config/load/100kb_s-1mb-3x.json"
 generate-load: build
 	./bin/load $(ARGS)
+
+clean:
+	rm -rf bin 2>/dev/null || true
+
+test:
+	cd correctness && go test
```
cody-littley: Just a heads up, I'm rebranding this test suite from `correctness tests` to `live tests`. Once my branch merges, this will need to change to `cd live`. I'll keep an eye on this. If your branch merges before mine, I'll fix it before my branch merges.

api/clients/v2/cert_builder.go
samlaf: hmmm this is a good question. Are you afraid that this will pollute stdout during tests or local runs?
I think its use case is probably indeed only to debug when an error occurs. But somehow I feel like logging more than not enough is good, as we can anyways filter out log lines that we don't care about in grafana.
Could potentially make it debug though? wdyt?

test/v2/config/environment/preprod.json
samlaf: Ok think I found the issue and a solution. Turns out golang tests automatically change cwd to be the directory of the test that is currently working. Which is why you guys' `../../../` relative paths were working. However this didn't work with the load-generator because that relative path was interpreted from where the load generator was started.

I changed the paths to be interpreted wrt the config file location, and not the test location. This I argue makes reading the config easier, and also makes using this config file from anywhere work: [59cda27](https://github.com/Layr-Labs/eigenda/pull/1722/commits/59cda27f79cd3fdf1a29aaf5922c60cd7fb2dcbb)

api/clients/v2/cert_builder.go
litt3: > Are you afraid that this will pollute stdout during tests or local runs

Yes

> But somehow I feel like logging more than not enough is good

"Too much" is better than "not enough," but even better is "just right". Can you think of any reason we'd want this info for calls that *don't* fail?

True, we can filter in grafana, but sometimes all you have is a log dump, and bloat complicates debugging.

samlaf: You convinced me. Probably helpful for your claude log parsing command. [0d189b5](https://github.com/Layr-Labs/eigenda/pull/1722/commits/0d189b5452a1114afd90ffc98adf185256fd54ad)

test/v2/Makefile
samlaf: Fixed in [11ec5a4](https://github.com/Layr-Labs/eigenda/pull/1722/commits/11ec5a4c7171b1d7bbebc7d446ba28c4e09cbf2d)

litt3 (dismissed):

cody-littley (approved):

litt3 (approved):

PR #1726, opened 2025-07-07
refactor: pointsIO functions and docs

samlaf:

My initial goal was to see whether we could bring the parsing times lower (takes ~10sec to parse 32MiB G2 file), which makes for long startup times of some binaries like load-generator, proxy, etc.

Wrote this comment in the test file:
> We currently store the points in compressed form for smaller file sizes.
We could store in uncompressed form (double size) for faster binary startup time.
See https://docs.gnark.consensys.io/HowTo/serialize#compression
The tests/benchmarks below can be used to compare the performance of reading compressed vs uncompressed points files.
Results when I ran them on my M1 MacBook Pro were 2x faster parsing at the cost of 2x larger file sizes:
> - G2 points: 32 MiB Compressed (9.5s parsing) vs 64 MiB Uncompressed (4.9s parsing)

TLDR is that we could store our SRS files in uncompressed format, which takes 2x the space but allows for 2x faster parsing. Not sure this is worth the introduced complexity (we'd need to document better that customers can use whichever srs points they want, etc) right now so opted to leave the tests/benchmarks related to uncompressed files, but to not commit the uncompressed SRS files.

Otherwise just refactored the entire pointsIO file to now consist of a single main generic function, and a bunch of wrappers that just call it (still feel like we should delete most of them though...)

## Future Works

Another issue I find annoying is that our parsing functions don't make use of the passed in context, so it's impossible to ctrl-c the proxy while its parsing, so we're forced to wait 10sec before shutdown. But I'm tired already of refactoring this, so will likely do this in the future PR.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests

### DIFF
--- a/encoding/kzg/pointsIO.go
+++ b/encoding/kzg/pointsIO.go
@@ -45,14 +45,14 @@ func ReadG1PointSection(filepath string, from, to uint64, numWorker uint64) ([]b
 }
 
 // Convenience wrapper for reading all G1 points from the start of the file.
-// See [readPointSection] for details on the parameters.
+// n is the number of points to read, numWorker is the number of goroutines to use for parallel parsing.
 func ReadG1Points(filepath string, n uint64, numWorker uint64) ([]bn254.G1Affine, error) {
 	// ReadG1Points is just ReadG1PointSection starting from 0
 	return ReadG1PointSection(filepath, 0, n, numWorker)
 }
 
 // Convenience wrapper for reading all G1 points in uncompressed format.
-// This is just a wrapper around [readPointSection] with the uncompressed size.
+// n is the number of points to read, numWorker is the number of goroutines to use for parallel parsing.
 // We don't currently use uncompressed file formats; see [BenchmarkReadG2PointsCompressedVsUncompressed] for performance comparison.
 func ReadG1PointsUncompressed(filepath string, n uint64, numWorker uint64) ([]bn254.G1Affine, error) {
 	// ReadG1PointsUncompressed is just ReadG1PointSection starting from 0
@@ -77,7 +77,8 @@ func ReadG2Point(n uint64, srsOrder uint64, g2Path string) (bn254.G2Affine, erro
 	return g2point[0], nil
 }
 
-// Convenience wrapper around [readPointSection] for reading a section of G2 points.
+// Convenience wrapper around [readPointSection] for reading G2 points from the start of the file.
+// n is the number of points to read, numWorker is the number of goroutines to use for parallel parsing.
 func ReadG2Points(filepath string, n uint64, numWorker uint64) ([]bn254.G2Affine, error) {
 	result, err := ReadG2PointSection(filepath, 0, n, numWorker)
 	if err != nil {
@@ -88,7 +89,7 @@ func ReadG2Points(filepath string, n uint64, numWorker uint64) ([]bn254.G2Affine
 }
 
 // Convenience wrapper for reading all G2 points in uncompressed format.
-// This is just a wrapper around [readPointSection] with the uncompressed size.
+// n is the number of points to read, numWorker is the number of goroutines to use for parallel parsing.
 // We don't currently use uncompressed file formats; see [BenchmarkReadG2PointsCompressedVsUncompressed] for performance comparison.
 func ReadG2PointsUncompressed(filepath string, n uint64, numWorker uint64) ([]bn254.G2Affine, error) {
 	// ReadG2PointsUncompressed is just ReadG2PointSection starting from 0
@@ -101,7 +102,8 @@ func ReadG2PointsUncompressed(filepath string, n uint64, numWorker uint64) ([]bn
 }
 
 // Convenience wrapper for reading a section of G2 points.
-// See [readPointSection] for details on the parameters.
+// from and to specify the range of point indices to read (inclusive from, exclusive to).
+// numWorker specifies the number of goroutines to use for parallel parsing.
 func ReadG2PointSection(filepath string, from, to uint64, numWorker uint64) ([]bn254.G2Affine, error) {
 	return readPointSection[bn254.G2Affine](filepath, from, to, G2PointBytes, numWorker)
 }
--- a/encoding/kzg/pointsIO.go
+++ b/encoding/kzg/pointsIO.go
@@ -182,9 +182,9 @@ func readPointSection[T bn254.G1Affine | bn254.G2Affine](
 		numWorker = n
 	}
 
-	buf, err := readDesiredBytes(reader, n*pointSizeBytes)
+	buf, err := readBytes(reader, n*pointSizeBytes)
 	if err != nil {
-		return nil, fmt.Errorf("readDesiredBytes: %w", err)
+		return nil, fmt.Errorf("readBytes: %w", err)
 	}
 
 	points := make([]T, n)
@@ -231,9 +231,9 @@ func readPointSection[T bn254.G1Affine | bn254.G2Affine](
 	return points, nil
 }
 
-// readDesiredBytes reads exactly numBytesToRead bytes from the reader and returns
+// readBytes reads exactly numBytesToRead bytes from the reader and returns
 // the result.
-func readDesiredBytes(reader *bufio.Reader, numBytesToRead uint64) ([]byte, error) {
+func readBytes(reader *bufio.Reader, numBytesToRead uint64) ([]byte, error) {
 	buf := make([]byte, numBytesToRead)
 	_, err := io.ReadFull(reader, buf)
 	// Note that ReadFull() guarantees the bytes read is len(buf) IFF err is nil.
--- a/encoding/kzg/pointsIO.go
+++ b/encoding/kzg/pointsIO.go
@@ -8,84 +8,25 @@ import (
 	"log"
 	"math"
 	"os"
-	"time"
 
 	"github.com/consensys/gnark-crypto/ecc/bn254"
 )
 
 const (
-	// Num of bytes per G1 point in serialized format in file.
-	G1PointBytes = 32
-	// Num of bytes per G2 point in serialized format in file.
-	G2PointBytes = 64
+	// We store the points in compressed form for smaller file sizes.
+	// We could store in uncompressed form (double size) for faster binary startup time.
+	// See https://docs.gnark.consensys.io/HowTo/serialize#compression
+	// and [BenchmarkReadG2PointsCompressedVsUncompressed] for performance comparison.
+
+	// Num of bytes per G1 point in (compressed) serialized format in file.
+	G1PointBytes = bn254.SizeOfG1AffineCompressed
+	// Num of bytes per G2 point in (compressed) serialized format in file.
+	G2PointBytes = bn254.SizeOfG2AffineCompressed
 )
-
-// from is inclusive, to is exclusive
-func ReadG1PointSection(filepath string, from, to uint64, numWorker uint64) ([]bn254.G1Affine, error) {
-	if to <= from {
-		return nil, fmt.Errorf("the range to read is invalid, from: %v, to: %v", from, to)
-	}
-	g1f, err := os.Open(filepath)
-	if err != nil {
-		return nil, fmt.Errorf("error cannot open g1 points file %w", err)
-	}
-
-	defer func() {
-		if err := g1f.Close(); err != nil {
-			log.Printf("g1 close error %v\n", err)
-		}
-	}()
-
-	n := to - from
-
-	g1r := bufio.NewReaderSize(g1f, int(n*G1PointBytes))
-
-	_, err = g1f.Seek(int64(from)*G1PointBytes, 0)
-	if err != nil {
-		return nil, err
-	}
-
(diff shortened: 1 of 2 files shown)


dmanc:
I remember that in addition to loading the points there is some subgroup check that takes some time. @bxue-l2 suggested we could potentially disable that but from the public API it seems it's always forcing the subgroup check.

https://github.com/Consensys/gnark-crypto/blob/da0317fd013308db6ce847bc9c3d506a2a3ae0ff/ecc/bn254/marshal.go#L1059

encoding/kzg/pointsIO.go
```diff
@@ -8,38 +8,25 @@ import (
...
-	// See https://pkg.go.dev/io#ReadFull.
-	if err != nil {
-		return nil, fmt.Errorf("cannot read file %w", err)
-	}
-	return buf, nil
-}
-
 // Read the n-th G1 point from SRS.
 func ReadG1Point(n uint64, srsOrder uint64, g1Path string) (bn254.G1Affine, error) {
+	// Do we really need to check srsOrder here? Or can we just read the file and let the error propagate if n is out of bounds?
```
litt3: nit: prefix comment with TODO

encoding/kzg/pointsIO.go
```diff
@@ -52,6 +39,31 @@ func ReadG1Point(n uint64, srsOrder uint64, g1Path string) (bn254.G1Affine, erro
...
+// See [readPointSection] for details on the parameters.
+func ReadG1Points(filepath string, n uint64, numWorker uint64) ([]bn254.G1Affine, error) {
+	// ReadG1Points is just ReadG1PointSection starting from 0
+	return ReadG1PointSection(filepath, 0, n, numWorker)
+}
+
+// Convenience wrapper for reading all G1 points in uncompressed format.
+// This is just a wrapper around [readPointSection] with the uncompressed size.
+// We don't currently use uncompressed file formats; see [BenchmarkReadG2PointsCompressedVsUncompressed] for performance comparison.
+func ReadG1PointsUncompressed(filepath string, n uint64, numWorker uint64) ([]bn254.G1Affine, error) {
```
litt3, line 57: You are pointing to `readPointSection` in lots of places for param docs, but `n` isn't a parameter there. 

I think it would make sense to explicitly state what `n` is in all these function godocs, or choose a parameter name that is unambiguous

encoding/kzg/pointsIO.go
```diff
@@ -97,92 +138,40 @@ func ReadG2PointOnPowerOf2(exponent uint64, srsOrder uint64, g2PowerOf2Path stri
...
 
 	defer func() {
-		if err := g1f.Close(); err != nil {
-			log.Printf("g1 close error %v\n", err)
+		if err := file.Close(); err != nil {
+			log.Printf("close error %v\n", err)
 		}
 	}()
 
 	n := to - from
```
litt3, line 173: It was preexisting, so no action required, but I'm not a fan of `n` as a variable name

encoding/kzg/pointsIO.go
```diff
@@ -191,207 +180,63 @@ func ReadG1PointSection(filepath string, from, to uint64, numWorker uint64) ([]b
...
-
-	startTimer = time.Now()
-
-	s2Outs := make([]bn254.G2Affine, n)
-
-	results := make(chan error, numWorker)
-
-	start := uint64(0)
-	end := uint64(0)
 	size := n / numWorker
```
litt3: `size` isn't a very good name for this, consider `pointsPerWorker`

encoding/kzg/pointsIO.go
```diff
@@ -191,207 +180,63 @@ func ReadG1PointSection(filepath string, from, to uint64, numWorker uint64) ([]b
...
-	for i := uint64(0); i < numWorker; i++ {
-		start = i * size
 
+	for i := uint64(0); i < numWorker; i++ {
+		start := i * size
+		end := start + size
 		if i == numWorker-1 {
 			end = n
```
litt3: this logic of this would be good to document, e.g. `the final worker reads the rest of the points`

encoding/kzg/pointsIO.go
```diff
@@ -191,207 +180,63 @@ func ReadG1PointSection(filepath string, from, to uint64, numWorker uint64) ([]b
...
-		} else {
-			end = (i + 1) * size
 		}
 
-		go readG2Worker(buf, s2Outs, start, end, G2PointBytes, results)
+		go func(start, end uint64) {
```
litt3: consider moving this function definition out of `readPointSection`

encoding/kzg/pointsIO.go
```diff
@@ -191,207 +180,63 @@ func ReadG1PointSection(filepath string, from, to uint64, numWorker uint64) ([]b
...
+			for j := start; j < end; j++ {
```
litt3: `j` here is a `pointIndex`

encoding/kzg/pointsIO.go
```diff
@@ -191,207 +180,63 @@ func ReadG1PointSection(filepath string, from, to uint64, numWorker uint64) ([]b
...
-
-	results := make(chan error, numWorker)
-
-	start := uint64(0)
-	end := uint64(0)
 	size := n / numWorker
-	for i := uint64(0); i < numWorker; i++ {
-		start = i * size
 
+	for i := uint64(0); i < numWorker; i++ {
```
litt3: `i` here is a `workerIndex`

encoding/kzg/pointsIO.go
```diff
@@ -191,207 +180,63 @@ func ReadG1PointSection(filepath string, from, to uint64, numWorker uint64) ([]b
...
+		start := i * size
```
litt3: `start` and `end` are ambiguous when we're dealing closely with both `points` and `bytes`. Consider `startPoint` and `endPoint`

encoding/kzg/pointsIO.go
```diff
@@ -191,207 +180,63 @@ func ReadG1PointSection(filepath string, from, to uint64, numWorker uint64) ([]b
...
-	}
-
-	if n < numWorker {
-		numWorker = n
-	}
-
-	buf, err := ReadDesiredBytes(g2r, n*G2PointBytes)
+// readDesiredBytes reads exactly numBytesToRead bytes from the reader and returns
+// the result.
+func readDesiredBytes(reader *bufio.Reader, numBytesToRead uint64) ([]byte, error) {
```
litt3: Why `readDesiredBytes` instead of just `readBytes`? I think `desired` is implied

litt3 (approved):

encoding/kzg/pointsIO.go
bxue-l2: Yes, you can.

encoding/kzg/pointsIO.go
```diff
@@ -8,38 +8,25 @@ import (
...
 	"github.com/consensys/gnark-crypto/ecc/bn254"
 )
 
 const (
-	// Num of bytes per G1 point in serialized format in file.
-	G1PointBytes = 32
-	// Num of bytes per G2 point in serialized format in file.
-	G2PointBytes = 64
+	// We store the points in compressed form for smaller file sizes.
+	// We could store in uncompressed form (double size) for faster binary startup time.
```
bxue-l2, line 17: If you are really optimizing for speed, a possible solution is to disable the subgroup check, I think it will speed up things significantly. https://github.com/consensys/gnark-crypto/blob/v0.18.0/ecc/bn254/marshal.go#L862

encoding/kzg/pointsIO.go
```diff
@@ -191,207 +180,63 @@ func ReadG1PointSection(filepath string, from, to uint64, numWorker uint64) ([]b
...
-
-	start := uint64(0)
-	end := uint64(0)
 	size := n / numWorker
-	for i := uint64(0); i < numWorker; i++ {
-		start = i * size
 
+	for i := uint64(0); i < numWorker; i++ {
+		start := i * size
+		end := start + size
```
bxue-l2, line 194: What if n / numWorker isn't cleanly divided.

encoding/kzg/pointsIO.go
```diff
@@ -97,92 +138,40 @@ func ReadG2PointOnPowerOf2(exponent uint64, srsOrder uint64, g2PowerOf2Path stri
...
 		}
 	}()
 
 	n := to - from
+	reader := bufio.NewReaderSize(file, int(n*pointSizeBytes))
 
-	g1r := bufio.NewReaderSize(g1f, int(n*G1PointBytes))
-
-	_, err = g1f.Seek(int64(from)*G1PointBytes, 0)
+	_, err = file.Seek(int64(from)*int64(pointSizeBytes), 0)
```
bxue-l2: use https://pkg.go.dev/io#pkg-constants SeekStart to replace magic 0

encoding/kzg/pointsIO.go
```diff
@@ -191,207 +180,63 @@ func ReadG1PointSection(filepath string, from, to uint64, numWorker uint64) ([]b
...
+			for j := start; j < end; j++ {
+				pointData := buf[j*pointSizeBytes : (j+1)*pointSizeBytes]
+				switch p := any(&points[j]).(type) {
+				case *bn254.G1Affine:
+					if _, err := p.SetBytes(pointData); err != nil {
+						results <- fmt.Errorf("error setting G1 point bytes: %w", err)
+						return
+					}
+				case *bn254.G2Affine:
+					// G2 points are stored as uncompressed points, so we can directly set bytes
```
bxue-l2: https://github.com/Consensys/gnark-crypto/blob/f8ab23ad72836de95a017e7df1930992f291a509/ecc/bn254/marshal.go#L1114
even wityh Compressed point setByte can work for G2.

dmanc (commented):
Should we upgrade the gnark-crypto library since we're here?

encoding/kzg/pointsIO.go
samlaf: foo

samlaf: Fixed in commit 39b85d8 - Added TODO prefix as requested

encoding/kzg/pointsIO.go
samlaf: Fixed in commit dff2aac - Updated parameter documentation to explicitly state what 'n' means in each function instead of referencing readPointSection

encoding/kzg/pointsIO.go
samlaf: Fixed - renamed size to pointsPerWorker for clarity

encoding/kzg/pointsIO.go
samlaf: Fixed in commit f2ec10a - renamed start/end to startPoint/endPoint for clarity

encoding/kzg/pointsIO.go
samlaf: Fixed in commit c3d0462 - renamed readDesiredBytes to readBytes

encoding/kzg/pointsIO.go
samlaf: Fixed in commit 02ef344 - renamed i to workerIndex for clarity (better than inline comment)

encoding/kzg/pointsIO.go
samlaf: Fixed in commit 02ef344 - renamed j to pointIndex for clarity (better than inline comment)

encoding/kzg/pointsIO.go
samlaf: Fixed in commit 0d30d68 - added comment explaining that the final worker reads the rest of the points

encoding/kzg/pointsIO.go
samlaf: Fixed in commit e73e835 - replaced magic number 0 with io.SeekStart

samlaf:
Holy shit with a lot of handholding I got claude-code to address most comments by itself, with a single commit per comment, and update each discussion with a commit link like I would do (figured out you can use `gh api graphql` api for that).

encoding/kzg/pointsIO.go
samlaf: [c07643d](https://github.com/Layr-Labs/eigenda/pull/1726/commits/c07643d4f0775af1343c7e278a9d4321f317f2e1)

encoding/kzg/pointsIO.go
samlaf: [551e897](https://github.com/Layr-Labs/eigenda/pull/1726/commits/551e897abeb43d491f82179c79936cbcade2ae42)

encoding/kzg/pointsIO.go
samlaf: Bit scared of just removing it. But gnark-crypto 0.18 has improved subgroup check speed on bn254 by 50% it looks like: https://github.com/Consensys/gnark-crypto/pull/635, so will update to it in a separate PR.

samlaf: https://github.com/Layr-Labs/eigenda/pull/1755

PR #1728, opened 2025-07-07
feat(payment): validator node batch ledger impl

hopeyen:

- `BatchLedger` for validating and tracking payment usage at the node level; atomic batch processing with a validate-then-commit approach to ensure data consistency across multi-account, multi-quorum operations.
- `UsageRequest` aggregated blob requests in a batch to a single usage request for an account/quorum/period
- `AccountUsage` tracks per-account period records in memory

Processing Flow:
1. Parse batch into usage requests, aggregating by account/quorum/period
2. Fetch all required reservations in batch
3. Validate all requests without modifying state (read-only)
4. Commit all changes after validation passes (write)

Validation Logic:
- Reservation activity and period validity
- Usage limit enforcement with overflow bin support
- Rate limit window
- Atomic failure handling (all-or-nothing)

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/meterer/batch_ledger.go
+++ b/core/meterer/batch_ledger.go
@@ -1,0 +1,299 @@
+package meterer
+
+import (
+	"context"
+	"fmt"
+	"slices"
+	"sync"
+	"time"
+
+	pb "github.com/Layr-Labs/eigenda/api/grpc/disperser/v2"
+	"github.com/Layr-Labs/eigenda/core"
+	"github.com/Layr-Labs/eigenda/core/meterer/payment_logic"
+	corev2 "github.com/Layr-Labs/eigenda/core/v2"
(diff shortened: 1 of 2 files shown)


core/meterer/batch_ledger.go
```diff
@@ -0,0 +1,299 @@
...
+	Period    uint64
+	Usage     uint64
+}
+
+type AccountUsage struct {
+	PeriodRecords map[core.QuorumID][]*pb.PeriodRecord
+	Lock          sync.RWMutex
+}
+
+type BatchLedger struct {
```
litt3, line 30: It seems like this `BatchLedger` should contain individual `AccountLedgers`, which would actually be doing the payment logic

pakim249CAL (approved):

PR #1730, opened 2025-07-07
refactor: add detail to accountant reservation failure

hopeyen:

More detailed error message

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/meterer/period_record.go
+++ b/core/meterer/period_record.go
@@ -90,7 +90,7 @@ func (pr QuorumPeriodRecords) UpdateUsage(
 	}
 
 	if oldUsage >= binLimit {
-		return fmt.Errorf("reservation limit exceeded for quorum %d", quorumNumber)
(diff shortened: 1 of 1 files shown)


core/meterer/period_record.go
```diff
@@ -90,7 +90,7 @@ func (pr QuorumPeriodRecords) UpdateUsage(
 	}
 
 	if oldUsage >= binLimit {
-		return fmt.Errorf("reservation limit exceeded for quorum %d", quorumNumber)
+		return fmt.Errorf("reservation limit exceeded for quorum %d; binLimit: %d, currentPeriod: %d, oldUsage: %d, relativePeriodRecord: %d", quorumNumber, binLimit, currentPeriod, oldUsage, relativePeriodRecord.Usage)
```
samlaf, line 93: Don't think we should be returning internal stack-trace like information like this in errors returned to customers.
We should instead be logging this info, and then we can correlate the timing of errors with our internal grafana logs.

PR #1732, opened 2025-07-07
refactor: account ledger interface incorporates debit slip

hopeyen:

Account ledger interfaces uses Debit Slip

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/payment/account_ledger.go
+++ b/core/payment/account_ledger.go
@@ -1,39 +1,32 @@
-package meterer
+package payment
 
 import (
 	"context"
 	"math/big"
 
 	disperser_v2 "github.com/Layr-Labs/eigenda/api/grpc/disperser/v2"
-	"github.com/Layr-Labs/eigenda/core"
-	gethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/Layr-Labs/eigenda/core/meterer"
 )
 
 // AccountLedger defines the interface for tracking account payment state.
 // Handles both reservation-based and on-demand payment modes with automatic fallback.
 type AccountLedger interface {
-	// Debit records symbol usage against the account.
+	// Debit records symbol usage against the account using a DebitSlip.
 	//
 	// Returns cumulative payment for on-demand usage, nil for reservation usage.
 	// Returns error if neither payment method can handle the request.
 	Debit(
 		ctx context.Context,
-		accountID gethcommon.Address,
-		timestampNs int64,
-		numSymbols uint64,
-		quorumNumbers []core.QuorumID,
-		params *PaymentVaultParams,
+		slip *DebitSlip,
+		params *meterer.PaymentVaultParams,
 	) (*big.Int, error)
 
-	// RevertDebit undoes a previous debit operation, restoring the account state.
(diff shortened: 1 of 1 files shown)


core/payment/account_ledger.go
```diff
@@ -1,39 +1,32 @@
...
 		ctx context.Context,
-		accountID gethcommon.Address,
-		timestampNs int64,
-		numSymbols uint64,
-		quorumNumbers []core.QuorumID,
-		params *PaymentVaultParams,
-		payment *big.Int,
+		slip *DebitSlip,
+		params *meterer.PaymentVaultParams,
+		previousCumulativePayment *big.Int,
```
litt3, line 29: Why is `previousCumulativePayment` necessary here?

Shouldn't the `DebitSlip` itself contain all information necessary to revert the debit?

hopeyen: `DebitSlip` contains the new cumulative payment to update to and it doesn't contain the cumulative payment that gets overwritten. 
Revert needs to know what value cumulative payment was previously so it can write it back

litt3: If you know the payment params, and the size of the blob, can't you compute what the previous cumulative payment was?

hopeyen: Nope, because on-demand payment allows any cumulative_payment greater than the required amount, rather than exactly equal to the required amount.

litt3: I'm just wondering if it would actually be a problem to *not* revert to the previous `cumulative_payment`, and instead just subtract the cost of the blob. i.e., the contract would be:

"if you submit a payment that skips over some account credit, that credit is lost immediately upon submission, regardless of whether the payment goes through"

If a client is submitting with a `cumulative_payment` greater than necessary, they would lose the extra value *anyway* if the payment succeeds.

hopeyen: I think you can even argue that there shouldn't be revert debit at all. The whole point of having revert is for protecting the users, and reverting to the previous `cumulative_payment` is more protective than to calculate and subtract the cost of the blob. 

I don't think there's a problem, more of what we want the user experience to be.

hopeyen: another thing is, every `cumulative_payment` tracked by the server/node should come with a corresponding blob dispersal signature. While the signature could technically represent the entire range from 0 to the signed over `cumulative_payment`, it isn't as clean as an injective function. 

and... probably a future consideration, is escalating fees. There would be a difference between the payment of a blob, and the minimum cost of the blob. Payment is `cumulative_payment - previousCumulativePayment`, while the cost is `pricePerSymbol * blobSize`. If some client is bidding for on-demand payments but got reverted, they shouldn't be charged for the difference between the two.

litt3: > I think you can even argue that there shouldn't be revert debit at all

I don't understand what you mean by this, but I agree now that it makes sense to keep the `previousCumulativePayment`

litt3 (approved):

PR #1733, opened 2025-07-07
feat: Add reservations expiration tracking to DataAPI 

dmanc:

Will change base branch to master after https://github.com/Layr-Labs/eigenda/pull/1709 is merged.

Reservation expiration tracking queries a subgraph to get the latests reservations for all accounts. The metrics are updated on every scrape to the `/metrics` endpoint.

Sample output on Sepolia
```
# HELP eigenda_reservation_time_until_expiry_seconds Time until reservation expiration in seconds
# TYPE eigenda_reservation_time_until_expiry_seconds gauge
eigenda_reservation_time_until_expiry_seconds{account="0x41fa832fad553c3b92976344d44b3548517679ac"} 9.88159285152387e+08
eigenda_reservation_time_until_expiry_seconds{account="0x7ccfa1e81b8d1d85552e860a3f3f63c94bd58259"} 2.3241885152387e+07
eigenda_reservation_time_until_expiry_seconds{account="0xae2c15d8354d14c05fe38733bd33ddeb00ab71ed"} 9.222690152387e+06
eigenda_reservation_time_until_expiry_seconds{account="0xc212b67c76586979cd736ea3e9738bb7bbf01b04"} 9.242137152387e+06
eigenda_reservation_time_until_expiry_seconds{account="0xcd1161b78f01da838ce0d42ec750891ec8708f1d"} 9.242137152387e+06
eigenda_reservation_time_until_expiry_seconds{account="0xd9309b3cf1b7dbf59f53461c2a66e2783dd1766f"} 5.012028152387e+06
eigenda_reservation_time_until_expiry_seconds{account="0xe25c3dec16f989793d230dc48c17388d53b4ac6e"} 2.3241885152387e+07
# HELP eigenda_reservations_active Number of active reservations
# TYPE eigenda_reservations_active gauge
eigenda_reservations_active 7
```

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


disperser/dataapi/subgraph_client.go
```diff
@@ -97,6 +98,10 @@ type (
 		// The operatorIds of nonsigners for the batch.
 		NonSigners []string
 	}
+	Reservation struct {
```
pschork, line 101: Should we also throw reservation symbols in here? Would allow us to monitor total allocated capacity.

pschork (dismissed):

disperser/dataapi/subgraph_client.go
dmanc: Yup I can add that

disperser/cmd/dataapi/flags/flags.go
```diff
@@ -71,6 +71,13 @@ var (
 		EnvVar:   common.PrefixEnvVar(envVarPrefix, "SUBGRAPH_OPERATOR_STATE_API_SOCKET_ADDR"),
 		Required: true,
 	}
+	SubgraphApiPaymentsAddrFlag = cli.StringFlag{
+		Name: common.PrefixFlag(FlagPrefix, "sub-payments-socket-addr"),
+		//We need the socket address of the subgraph payments api to pull the subgraph data from.
+		Usage:    "the socket address of the subgraph payments api",
```
samlaf: is this really a socket address? aka ip:port? Or is it a full URL with http endpoint mapping to specific satsuma resource?

disperser/cmd/dataapi/config.go
```diff
@@ -28,6 +28,7 @@ type Config struct {
 	PrometheusApiAddr            string
 	SubgraphApiBatchMetadataAddr string
 	SubgraphApiOperatorStateAddr string
+	SubgraphApiPaymentsAddr      string
```
samlaf, line 31: I'm assuming all of these addresses are the same? aka pointing to satsuma ip:port? Or are these full URLs? "socket address" is a bit confusing

disperser/dataapi/metrics.go
```diff
@@ -206,6 +206,11 @@ func (g *Metrics) UpdateSemverCounts(semverData map[string]*semver.SemverMetrics
 	}
 }
 
+// RegisterCollector registers a prometheus collector with the metrics registry
+func (g *Metrics) RegisterCollector(collector prometheus.Collector) {
+	g.registry.MustRegister(collector)
+}
```
samlaf: this is just a nit (no need to change) as I haven't looked at the whole metrics code and am not 100% sure, but typically I feel like a Metrics struct should only deal with metrics, not with a registry. Aka it should get the registry injected at construction time, and not expose a method like this.

disperser/dataapi/subgraph_client.go
samlaf: +1. How are the metrics meant to be used? I'm assuming this is purely for us internally to add grafana alerts when a reservation is about to expire?

Would we ever want to expose reservations info on the blob explorer? If so then we might as well read all the info exposed per reservation by the contract/subgraph?

disperser/dataapi/subgraph_client.go
```diff
@@ -358,6 +363,26 @@ func (sc *subgraphClient) QueryIndexedRegisteredOperatorsForTimeWindow(ctx conte
 
 }
 
+func (sc *subgraphClient) QueryReservations(ctx context.Context, currentTimestamp uint64, limit, skip int) ([]*Reservation, error) {
+	reservationsGql, err := sc.api.QueryReservations(ctx, currentTimestamp, limit, skip)
+	if err != nil {
+		return nil, err
```
samlaf: wrap errors

tools/ejections/cmd/main.go
```diff
@@ -77,7 +77,7 @@ func RunScan(ctx *cli.Context) error {
 	if chainState == nil {
 		return errors.New("failed to create chain state")
 	}
-	subgraphApi := subgraph.NewApi(config.SubgraphEndpoint, config.SubgraphEndpoint)
+	subgraphApi := subgraph.NewApi(config.SubgraphEndpoint, config.SubgraphEndpoint, config.SubgraphEndpoint)
```
samlaf, line 87: This kind of shows that the NewApi config/constructor is generic for no reasons. Are there ever cases where the different subgraphs don't use the same endpoint? This looks weird that we're just passing the same endpoint 3 times. If we don't ever need the different subgraphs to have different endpoints, then prob best to consolidate them into a single endpoint field like in this config

disperser/dataapi/subgraph/queries.go
```diff
@@ -118,4 +121,7 @@ type (
 	queryOperatorEjectedsByOperatorID struct {
 		OperatorEjections []*OperatorEjection `graphql:"operatorEjecteds(orderBy: blockTimestamp, where: {and: [{blockTimestamp_gte: $blockTimestamp_gte}, {operatorId: $operatorId}]}, first: $first, skip: $skip)"`
 	}
+	queryReservations struct {
+		Reservations []*Reservation `graphql:"currentReservations(where: {startTimestamp_lte: $currentTimestamp, endTimestamp_gte: $currentTimestamp}, orderBy: startTimestamp, orderDirection: asc, first: $first, skip: $skip)"`
```
samlaf:
```suggestion
	queryActiveReservations struct {
		Reservations []*Reservation `graphql:"currentReservations(where: {startTimestamp_lte: $currentTimestamp, endTimestamp_gte: $currentTimestamp}, orderBy: startTimestamp, orderDirection: asc, first: $first, skip: $skip)"`
```

samlaf: For consistency think we should use "queryActiveReservations" everywhere. I also commented on your other PR that "currentReservations" is unclear as it sounds like it means "activeReservations" but it actually just means "historicalReservations".

samlaf: TLDR: think we should change
- in other PR: currentReservations -> reservations (or allReservations)
- in this PR: queryReservations -> queryActiveReservations (because it filters for only those currently active)

disperser/dataapi/v2/reservation_collector.go
```diff
@@ -0,0 +1,109 @@
...
+
+	c.updateCounts(ctx)
+
+	// Collect metrics
+	c.reservationsActive.Collect(ch)
+	c.reservationTimeUntilExpiry.Collect(ch)
+}
+
+// updateCounts queries the GraphQL endpoint and updates the metrics
+func (c *ReservationExpirationCollector) updateCounts(ctx context.Context) {
```
samlaf: why `updateCounts` and not `updateMetrics`? it updates timeUntilExpiry which is not a "count"

disperser/dataapi/subgraph_client.go
dmanc: First use case is for reservation expiration tracking internally but we can extend the scope to any blob explorer requirements in future PRs.

disperser/cmd/dataapi/flags/flags.go
dmanc: It's just a full URL of an http endpoint

disperser/cmd/dataapi/config.go
dmanc: These are all full URLs and they're all different depending on which subgraph you are querying

tools/ejections/cmd/main.go
dmanc: This was weird when I saw it but seems like this specific ejection tool thing only needs the middle one but this object needs all three. Open to suggestions on refactoring but feel like it's gonna be a big change.

disperser/dataapi/metrics.go
dmanc: 6aae9802fb8446907ac3b7d5cedc26e6e08e8842

disperser/dataapi/subgraph_client.go
dmanc: e023781be3161eab54a3caee6db7871c730e6cd2

disperser/dataapi/subgraph_client.go
dmanc: 68084d22e72744a0e3e76d62ac3a92bbed965887

disperser/dataapi/subgraph/queries.go
dmanc: This was done, not sure what commit though

disperser/dataapi/v2/reservation_collector.go
dmanc: 65c828af1d633f531075c9bea7e13dd3b24c1e42

disperser/cmd/dataapi/flags/flags.go
samlaf: no need to do in this PR but austin has raised this in the past and I agree, "socket address" typically mans "ip:port" pair. At least in offchain code/comments we should mention its a full URL
image

tools/ejections/cmd/main.go
samlaf: oh I see. Fine with just adding a comment then.

samlaf (approved):

disperser/cmd/dataapi/flags/flags.go
pschork: I think all we need to do here is change the usage text for these subgraph flags to
`Usage: "the URL of the subgraph payments api",` and we're golden.
image

tools/ejections/cmd/main.go
pschork: yea the ejection tool is my work. Passing duplicate urls is a hack since `NewApi` is an abstraction over multiple subgraph clients (ejection only needs one) and I didn't want to justify yak shaving a refactor of 2yo code assumptions built for dependency injection because some esoteric ejection reporting tool needed to call a subgraph 💀

pschork (approved):

tools/ejections/cmd/main.go
samlaf: I'm totally fine with these decisions, just think they should be documented in the code itself.

dmanc: fdedee733bf19d4fdd40792d75565a37c93d5b84

disperser/cmd/dataapi/flags/flags.go
dmanc: fdedee733bf19d4fdd40792d75565a37c93d5b84

samlaf (approved):

PR #1740, opened 2025-07-09
docs: Create repo style guide

litt3:
- drafts a style guide. It's by no means fully complete, but I just put some thoughts down as a starting point. We can work on this as a team, during review of this PR, and with subsequent PRs
- I'm trying to get something in place here, so that I can make claude automatically do a style / nitpick review pass, to take load off of human reviewers

### DIFF
--- a/.github/workflows/buf-proto.yaml
+++ b/.github/workflows/buf-proto.yaml
@@ -12,14 +12,10 @@
 name: Buf Proto
 on:
   push:
-    paths:
-      - "api/proto/**"
-      - "api/buf.yaml"
+    branches:
+      - master
   pull_request:
-    types: [opened, synchronize, reopened, labeled, unlabeled]
-    paths:
-      - "api/proto/**"
-      - "api/buf.yaml"
+  merge_group:
 
 permissions:
   contents: read
--- a/.github/workflows/codeql-scanning.yaml
+++ b/.github/workflows/codeql-scanning.yaml
@@ -25,6 +25,7 @@ on:
       - "relay/**"
       - ".github/codeql/**"
       - ".github/workflows/codeql-scanning.yaml"
+  merge_group:
   schedule:
     - cron: "0 9 * * *"
 
--- a/.github/workflows/compile-protobufs.yaml
+++ b/.github/workflows/compile-protobufs.yaml
@@ -4,8 +4,7 @@ on:
     branches:
       - master
   pull_request:
-    branches:
-      - master
+  merge_group:
 
 jobs:
   golangci:
--- a/.github/workflows/docker-publish.yaml
+++ b/.github/workflows/docker-publish.yaml
@@ -4,6 +4,7 @@ on:
     branches:
       - master
   pull_request:
+  merge_group:
   workflow_dispatch:
 
 env:
--- a/.github/workflows/golangci-lint.yml
+++ b/.github/workflows/golangci-lint.yml
@@ -4,6 +4,7 @@ on:
     branches:
       - master
   pull_request:
+  merge_group:
 
 env:
   MISE_VERSION: 2024.12.14
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -4,8 +4,7 @@ on:
     branches:
       - master
   pull_request:
-    branches:
-      - master
+  merge_group:
 
 env:
   MISE_VERSION: 2024.12.14
--- a/.github/workflows/pr-title.yaml
+++ b/.github/workflows/pr-title.yaml
@@ -3,6 +3,10 @@ name: PR Title Linting
 on:
   pull_request:
     types: [opened, edited, synchronize]
+  # This workflow is currently not required on github because it doesn't work
+  # for merge_group events, because of its use of '.pull_request.title' below.
+  # TODO: update this workflow to work with merge_group events.
+  # merge_group:
 
 jobs:
   lint-pr-title:
@@ -13,7 +17,7 @@ jobs:
         run: |
           PR_TITLE=$(jq -r '.pull_request.title' "$GITHUB_EVENT_PATH")
           echo "PR title: $PR_TITLE"
-          
+
           # Define the valid pattern (supports conventional commit format with breaking changes)
           if [[ ! "$PR_TITLE" =~ ^(feat|fix|chore|docs|refactor|test|style|ci|perf)(\([a-z0-9-]+\))?(!)?:\ .* ]]; then
             echo "❌ Invalid PR title: '$PR_TITLE'"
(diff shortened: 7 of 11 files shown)


docs/style-guide.md
```diff
@@ -0,0 +1,118 @@
+## Style Guide
+
+This style guide contains coding style guidelines for the EigenDA project. This guide is not exhaustive, but rather
+builds on top of the guidelines expressed in [Effective Go](https://go.dev/doc/effective_go). It is intended as a guide
+for human engineers, and to provide AI agents with a checklist for code review.
+
+### 1. Error handling
+
+1. Return errors explicitly; don't panic except for unrecoverable errors.
+   - Exceptions may be made for test code, where returning an error adds more complexity than benefit.
```
cody-littley:
```suggestion
1. Return errors explicitly; don't panic except for unrecoverable errors. And even for unrecoverable errors, prefer errors over panicking if returning an error is plausible.
   - Exceptions may be made for test code, where returning an error adds more complexity than benefit.
```

docs/style-guide.md
```diff
@@ -0,0 +1,118 @@
...
+   - TODO comments that must be addressed prior to merging a PR should clearly be marked, e.g.
+   `// TODO: MUST BE ADDRESSED PRIOR TO MERGE`
+
+### 3. Spelling and Grammar
+
+Proper spelling and grammar are important, because they help keep code and documentation unambiguous, easy to read, 
+and professional. They should be checked and carefully maintained.
+
+1. Overly strict adherence to arbitrary grammar "rules" that don't impact readability is not beneficial. Some 
+   examples of "rules" that shouldn't be enforced or commented on are:
```
cody-littley, line 61: Add the following sentence:

```
As a general rule of thumb, colloquial language that is acceptable in a professional setting is acceptable.
```

CLAUDE.md
```diff
@@ -33,6 +33,7 @@ helpful to provide project context, but only within reasonable limits.
 3. @mise.toml describes external tool dependencies
 4. @.golangci.yml contains linting configuration
 5. @.claude/commands/CLAUDE.md defines what project slash commands are available to use
+6. @docs/style-guide.md defines team coding style, and should be considered whenever making code changes
```
dmanc, line 36: The style guide I'm guessing would only apply to new code added? How do we deal with all the code that currently doesn't follow the style guide? Would we try to update it whenever we can?

docs/style-guide.md
```diff
@@ -0,0 +1,118 @@
...
+   - E.g. the doc for `computeData` should NOT be "Computes the data".
+4. Function docs should consider the following helpful information, if relevant:
+   - What are the inputs?
+   - Are there any restrictions on what the input values are permitted to be?
+   - What is returned in the standard case?
+   - What is returned in the error case(s)?
+   - What side effects does calling the function have?
+   - Are there any performance implications that users should be aware of?
+   - Are there any performance optimizations that should/could be undertaken in the future?
+5. TODO comments should be added to denote future work.
```
pschork, line 47: When should TODOs include attribution ie `// TODO(pschork): ...` or not?

litt3: Personally I prefer always adding the attribution. Do you think we should explicitly require it?

docs/style-guide.md
litt3: a68bb3ff7e40989c643549a600c6be5a076d1229

docs/style-guide.md
litt3: 3aab00407427b10dd39d63b13dd951a721cf83b7

CLAUDE.md
litt3: 79e1f5c17286fa7f85a13558ed64fdb06b3f7cbf

litt3: Don't know why that sha isn't being turned into a link 🤷 

I added a section to try to clarify, LMK what you think

docs/style-guide.md
litt3: 805f23ccf74011d4b0428fe581fd24444d5717d9

Specified that TODOs should include attribution

CLAUDE.md
dmanc: 💯

dmanc (approved):

docs/style-guide.md
```diff
@@ -0,0 +1,152 @@
+## Style Guide
```
pakim249CAL, line 1: This file seems to consume ~2k tokens, out of a budget of 200k token context window for claude. Isn't this a bit high?

https://support.anthropic.com/en/articles/7996856-what-is-the-maximum-prompt-length

litt3: > Isn't this a bit high?

Doesn't seem high to me. I think the context of how to behave in a repo is some of the most important stuff claude can have, 1% seems pretty cheap

How did you determine the 2k tokens?

pakim249CAL: https://console.anthropic.com/

If you put in a prompt here it will count tokens for you

I don't necessarily think it is so high that it is a blocker, but I just wanted to bring to your attention that we need to be mindful of filling up the context window, as this leaves less room for anything else. In combination with having to keep context of multiple files and prompts, I think this stuff can add up very quickly. IMO, all preprompts in totality should probably not consume more than ~5-10% of the total context window if possible.

pakim249CAL: Oh never mind, I now realize this is not meant to be used in our workflows, but instead as claude reviews PRs.

docs/style-guide.md
```diff
@@ -0,0 +1,152 @@
...
+   - Changes may be made in surrounding code for the sake of readability, but there's a fine line between
+   "improving readability", and "aggressively imposing personal preference".
+   - If there is a disagreement between engineers about style, the team should come to consensus and enshrine the
+   result as an entry in this style guide.
+
+### 2. Error Handling
+
+1. Return errors explicitly; don't panic except for unrecoverable errors, where returning an error is not plausible.
+   - Exceptions may be made for test code, where returning an error adds more complexity than benefit.
+2. Use error wrapping with `fmt.Errorf("context: %w", err)` for additional context.
```
dmanc, line 28: Have a question about this. Should we always wrap errors? Like should we never see:
```
if err != nil {
	return nil, err
}
```

litt3: I think in *most* cases errors should be wrapped, but there are arguably some exceptions. For example, in the past in some cases I've chosen not to wrap, with a comment: `// wrapping not necessary, since utility method gives all necessary context`

Since exceptions exist, I stopped short of making it a hard requirement in this doc. Though if other people were on board, I would personally be fine enforcing wrapping ALL errors....

@samlaf brought my attention to the `wrapcheck` linter rule, which enforces wrapping errors returned from external packages. I think it would be good to turn on this linting rule (it's on my list of TODOs to try out), and then make the decision on a case-by-case basis for wrapping of internal errors.

PR #1742, opened 2025-07-10
fix: traffic generator now able to send high traffic

cody-littley:

There was a bug in the traffic generator setup where the maximum submission rate was capped at 1hz. This PR fixes that, as well as adding some debug logs that will assist with understanding ramp up time.

```
2025-07-10 10:56:27.133	
Jul 10 15:56:27.133 DBG common/variable_ticker.go:181 Current ticker frequency: 0.000 Hz, target frequency: 4 Hz. Acceleration is 0.0025 Hz/s, it will take 26.67 minutes to reach the target frequency.
```

### DIFF
--- a/common/variable_ticker.go
+++ b/common/variable_ticker.go
@@ -5,19 +5,25 @@ import (
 	"fmt"
 	"math"
 	"time"
+
+	"github.com/Layr-Labs/eigensdk-go/logging"
 )
 
 // Any frequency at or below this value will be interpreted as a frequency of 0 Hz. Needed to avoid overflow.
 // The factor of 2 is to take care of floating point precision issues.
-const MinimumFrequency = float64(time.Nanosecond/math.MaxInt64) * 2.0
+const MinimumFrequency = float64(time.Second/math.MaxInt64) * 2.0
 
 // Any frequency above this value will be interpreted as a frequency of MaximumFrequency Hz. Needed to avoid overflow.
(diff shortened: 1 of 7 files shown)


common/variable_ticker.go
```diff
@@ -5,19 +5,25 @@ import (
...
 )
 
 // Any frequency at or below this value will be interpreted as a frequency of 0 Hz. Needed to avoid overflow.
 // The factor of 2 is to take care of floating point precision issues.
-const MinimumFrequency = float64(time.Nanosecond/math.MaxInt64) * 2.0
+const MinimumFrequency = float64(time.Second/math.MaxInt64) * 2.0
 
 // Any frequency above this value will be interpreted as a frequency of MaximumFrequency Hz. Needed to avoid overflow.
-const MaximumFrequency = float64(time.Nanosecond)
+const MaximumFrequency = float64(time.Second)
```
litt3, line 19: This is kind of a confusing way to define the maximum frequency, any reason not to just do 1e9 directly? I suppose you're tying the max frequency to the time system's precision, but it would be useful to state what this is doing: at first non-critically-thinking glance, this looks like "max frequency is 1 Hz"

litt3 (approved):

common/variable_ticker.go
cody-littley: updated comment

dmanc (approved):

PR #1748, opened 2025-07-10
fix: update eigenda-releaser.yaml

anupsv:

wrong action item fixed.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/.github/workflows/eigenda-releaser.yaml
+++ b/.github/workflows/eigenda-releaser.yaml
@@ -19,7 +19,7 @@ jobs:
 
     steps:
       - name: Generate a token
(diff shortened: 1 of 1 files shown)


pschork (commented):
Can we rename this to something like `eigenda-release-branch` as is conflicts with `eigenda-goreleaser` which actually creates a GH release + artifacts.
See https://github.com/Layr-Labs/eigenda/pull/1715

anupsv:
> Can we rename this to something like `eigenda-release-branch` as is conflicts with `eigenda-goreleaser` which actually creates a GH release + artifacts. See #1715

yea i just want to test it to see.

PR #1757, opened 2025-07-11
feat: initializable

pakim249CAL:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/contracts/src/core/libraries/v3/initializable/InitializableLib.sol
+++ b/contracts/src/core/libraries/v3/initializable/InitializableLib.sol
@@ -1,0 +1,45 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.12;
+
+/// @notice Defines the storage layout for an address directory based on ERC-7201
+///         https://eips.ethereum.org/EIPS/eip-7201
+library InitializableStorage {
+    /// @custom: storage-location erc7201:address.directory.storage
+    struct Layout {
+        uint8 initialized;
+    }
+
+    string internal constant STORAGE_ID = "initializable.storage";
+    bytes32 internal constant STORAGE_POSITION =
+        keccak256(abi.encode(uint256(keccak256(abi.encodePacked(STORAGE_ID))) - 1)) & ~bytes32(uint256(0xff));
+
+    function layout() internal pure returns (Layout storage s) {
+        bytes32 position = STORAGE_POSITION;
+        assembly {
+            s.slot := position
+        }
+    }
+}
+
+library InitializableLib {
+    event Initialized(uint8 version);
+
+    error AlreadyInitialized();
+
+    function s() private pure returns (InitializableStorage.Layout storage) {
+        return InitializableStorage.layout();
+    }
+
+    function setInitializedVersion(uint8 version) internal {
+        if (s().initialized >= version) {
+            revert AlreadyInitialized();
+        }
+
+        s().initialized = version;
+        emit Initialized(version);
+    }
+
+    function getInitializedVersion() internal view returns (uint8 version) {
+        version = s().initialized;
+    }
+}


contracts/src/core/libraries/v3/initializable/InitializableLib.sol
```diff
@@ -0,0 +1,45 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.12;
+
+/// @notice Defines the storage layout for an address directory based on ERC-7201
+///         https://eips.ethereum.org/EIPS/eip-7201
+library InitializableStorage {
+    /// @custom: storage-location erc7201:address.directory.storage
```
samlaf: what does address.directory.storage mean here? how is it related to STORAGE_ID below for example

contracts/src/core/libraries/v3/initializable/InitializableLib.sol
```diff
@@ -0,0 +1,45 @@
...
+        if (s().initialized >= version) {
+            revert AlreadyInitialized();
+        }
+
+        s().initialized = version;
+        emit Initialized(version);
+    }
+
+    function getInitializedVersion() internal view returns (uint8 version) {
+        version = s().initialized;
```
samlaf, line 33: what exactly is this version? Given that we've been talking about adding semver, should version not be a triple of uint8 and not just a uint8? Or is this for some other purpose?

samlaf (dismissed):

contracts/src/core/libraries/v3/initializable/InitializableLib.sol
pakim249CAL: It is a mistake. Will fix.

contracts/src/core/libraries/v3/initializable/InitializableLib.sol
pakim249CAL: This is NOT semver and should not be treated as such. This is because this initialized version is only used on the contract level to ensure that any initializations for any contract upgrades only ever happen once. It is a variant of the simple solution of just setting a boolean on whether the contract was initialized. But expanded to allow new implementation contracts to also define their own initialize functions that can only ever be run once, and initializations are never out of order.

It cannot be semver because not all contract upgrades require re-initializations. Because of this, there is no strong assurance that the initialized version would match the intended semver. Also, this version lives in contract storage, whereas I believe you want semver to live in the actual contract code.

contracts/src/core/libraries/v3/initializable/InitializableLib.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1757/commits/e6f41ce9d5e1ed1b04bbdf8cc81aa0b71868286f

contracts/src/core/libraries/v3/initializable/InitializableLib.sol
samlaf: That makes sense!

samlaf (dismissed):

contracts/src/core/libraries/v3/initializable/InitializableLib.sol
```diff
@@ -0,0 +1,45 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.12;
+
+/// @notice Defines a storage layout based on ERC-7201
+///         https://eips.ethereum.org/EIPS/eip-7201
+library InitializableStorage {
```
litt3: `Initialize` seems to imply a boolean relationship. I think `UpgradableStorage`, and generally moving from `initialize` -> `upgrade` terminology might make more sense

contracts/src/core/libraries/v3/initializable/InitializableLib.sol
```diff
@@ -0,0 +1,45 @@
...
+    }
+
+    string internal constant STORAGE_ID = "initializable.storage";
+    bytes32 internal constant STORAGE_POSITION =
+        keccak256(abi.encode(uint256(keccak256(abi.encodePacked(STORAGE_ID))) - 1)) & ~bytes32(uint256(0xff));
+
+    function layout() internal pure returns (Layout storage s) {
+        bytes32 position = STORAGE_POSITION;
+        assembly {
+            s.slot := position
```
litt3: IMO one character parameters (`s`) and functions (`s()`, below) make it more difficult to understand at a glance what is happening

contracts/src/core/libraries/v3/initializable/InitializableLib.sol
pakim249CAL: I don't disagree, but also this naming is already used by most common libraries that already implement it exactly this way, so I would find it problematic to diverge from that. [Here](https://github.com/solidstate-network/solidstate-solidity/blob/master/contracts/access/initializable/_Initializable.sol) is an example. Maybe what I can do is follow their pattern more closely by splitting up initialize and reinitialize like they do.

contracts/src/core/libraries/v3/initializable/InitializableLib.sol
pakim249CAL: Maybe I'll just put the storage library in a separate file, as I personally find the conciseness of referring to this library's intended storage quite convenient and neat.

pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1757/commits/59315a8c91b4c6b1dc1108ecfdf4c77ec12bc638

contracts/src/core/libraries/v3/initializable/InitializableLib.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1757/commits/d9368f01bf6ad6e380531ca30a64cbbdf91dbbff

litt3 (approved):

PR #1758, opened 2025-07-11
feat: access control contract

pakim249CAL:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/contracts/src/core/EigenDAAccessControl.sol
+++ b/contracts/src/core/EigenDAAccessControl.sol
@@ -1,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+import {AccessControl} from "@openzeppelin/contracts/access/AccessControl.sol";
+import {AccessControlConstants} from "src/core/libraries/v3/access-control/AccessControlConstants.sol";
+
+contract EigenDAAccessControl is AccessControl {
+    constructor(address owner) {
+        _setupRole(AccessControlConstants.OWNER_ROLE, owner);
+    }
+
+    function setupRole(bytes32 role, address account) external {
+        require(hasRole(AccessControlConstants.OWNER_ROLE, msg.sender), "Caller is not the owner");
+        _setupRole(role, account);
+    }
+}
--- a/contracts/src/core/libraries/v3/access-control/AccessControlConstants.sol
+++ b/contracts/src/core/libraries/v3/access-control/AccessControlConstants.sol
@@ -1,0 +1,11 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+library AccessControlConstants {
+    bytes32 internal constant OWNER_ROLE = keccak256("OWNER");
+    bytes32 internal constant QUORUM_OWNER_SEED = keccak256("QUORUM_OWNER");
+
+    function QUORUM_OWNER_ROLE(uint64 quorumId) internal pure returns (bytes32) {
+        return bytes32(uint256(QUORUM_OWNER_SEED) + quorumId);
+    }
+}
--- a/contracts/script/EigenDADeployer.s.sol
+++ b/contracts/script/EigenDADeployer.s.sol
@@ -33,6 +33,7 @@ import {IEigenDADisperserRegistry} from "src/core/interfaces/IEigenDADisperserRe
 import {EigenDARelayRegistry} from "src/core/EigenDARelayRegistry.sol";
 import {ISocketRegistry, SocketRegistry} from "../lib/eigenlayer-middleware/src/SocketRegistry.sol";
 import {IEigenDADirectory, EigenDADirectory} from "src/core/EigenDADirectory.sol";
+import {EigenDAAccessControl} from "src/core/EigenDAAccessControl.sol";
 import {
     DeployOpenEigenLayer,
     ProxyAdmin,
@@ -70,8 +71,10 @@ contract EigenDADeployer is DeployOpenEigenLayer {
     IPaymentVault public paymentVault;
     EigenDARelayRegistry public eigenDARelayRegistry;
     IEigenDADisperserRegistry public eigenDADisperserRegistry;
+    EigenDAAccessControl public eigenDAAccessControl;
 
     EigenDADirectory public eigenDADirectoryImplementation;
+
     BLSApkRegistry public apkRegistryImplementation;
     EigenDAServiceManager public eigenDAServiceManagerImplementation;
     EigenDACertVerifierRouter public eigenDACertVerifierRouterImplementation;
@@ -155,6 +158,9 @@ contract EigenDADeployer is DeployOpenEigenLayer {
                 )
             )
         );
+
+        eigenDAAccessControl = new EigenDAAccessControl(addressConfig.eigenLayerCommunityMultisig);
+        eigenDADirectory.addAddress(AddressDirectoryConstants.ACCESS_CONTROL_NAME, address(eigenDAAccessControl));
 
         /**
          * First, deploy upgradeable proxy contracts that **will point** to the implementations. Since the implementation contracts are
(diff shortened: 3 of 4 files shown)


contracts/src/core/EigenDAAccessControl.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+import {AccessControl} from "@openzeppelin/contracts/access/AccessControl.sol";
+import {AccessControlConstants} from "src/core/libraries/v3/access-control/AccessControlConstants.sol";
+
+contract EigenDAAccessControl is AccessControl {
+    constructor(address owner) {
+        _setupRole(AccessControlConstants.OWNER_ROLE, owner);
```
samlaf: which version of OZ are we on? seems like _setupRole is deprecated in favor of _grantRole in v4.9.6: https://github.com/OpenZeppelin/openzeppelin-contracts/blob/dc44c9f1a4c3b10af99492eed84f83ed244203f6/contracts/access/AccessControl.sol#L204

contracts/src/core/libraries/v3/access-control/AccessControlConstants.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+library AccessControlConstants {
+    bytes32 internal constant OWNER_ROLE = keccak256("OWNER");
+    bytes32 internal constant QUORUM_OWNER_SEED = keccak256("QUORUM_OWNER");
+
+    function QUORUM_OWNER_ROLE(uint64 quorumId) internal pure returns (bytes32) {
```
samlaf, line 13: how should this be used? Is the idea that the caller would call
```
EigenDAAccessControl.setupRole(AccessControlConstants.QUORUM_OWNER_ROLE(quorumId))
```
? Or how else is this supposed to be used? should we add a quorum_setter in EigenDAAccessControl already?

contracts/src/core/EigenDAAccessControl.sol
```diff
@@ -0,0 +1,16 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+import {AccessControl} from "@openzeppelin/contracts/access/AccessControl.sol";
+import {AccessControlConstants} from "src/core/libraries/v3/access-control/AccessControlConstants.sol";
+
+contract EigenDAAccessControl is AccessControl {
```
samlaf, line 9: Document that this is meant to be deployed as the single centralized source of truth ACL for all EigenDA contracts (which will need to call into this contract), as opposed to the traditional OZ way (iiuc) of having each contract inherit AccessControl directly and own its own ACL.

contracts/src/core/libraries/v3/access-control/AccessControlConstants.sol
```diff
@@ -0,0 +1,11 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+library AccessControlConstants {
+    bytes32 internal constant OWNER_ROLE = keccak256("OWNER");
```
samlaf, line 7: iiuc this is only the OWNER of the EigenDAAccessControl itself right? How would you set owner roles for other contracts?

contracts/src/core/libraries/v3/access-control/AccessControlConstants.sol
pakim249CAL: This function defines a protocol in which the role identifier for a quorum owner for a particular quorum ID is generated. It can be used in contracts, tests, and production as a library function. Your example is one use case, but it would also be necessary to have this kind of function for contracts to efficiently query for quorum owners.

contracts/src/core/libraries/v3/access-control/AccessControlConstants.sol
pakim249CAL: No, the idea is for this role to represent what we currently understand as an owner, which is the DA Ops multisig on mainnet. This access control contract would eventually define roles for other contracts, but we do not have those roles currently. We should define additional roles if we find a need for a separate security framework other than ownership, such as payment vault reservations.

contracts/src/core/EigenDAAccessControl.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1758/commits/fcf8dadb1063baed05334720fe7370cdc964c9f9

contracts/src/core/EigenDAAccessControl.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1758/commits/233f929b3e0084abe2dfdccaadddbea57cc806c5#diff-7b41b24b87d5d57407ba4c159c745fd8d8a9c3786f31b3a231486a14b70bdeb7

contracts/src/core/libraries/v3/access-control/AccessControlConstants.sol
samlaf: This intended use case should probably be documented in the contract then.

contracts/src/core/libraries/v3/access-control/AccessControlConstants.sol
samlaf: I see. This should be documented above OWNER_ROLE then

contracts/src/core/EigenDAAccessControl.sol
```diff
@@ -0,0 +1,18 @@
...
+import {AccessControlConstants} from "src/core/libraries/v3/access-control/AccessControlConstants.sol";
+
+/// @title EigenDAAccessControl
+/// @notice This contract is to serve as the centralized source of truth for access control in all EigenDA contracts.
+contract EigenDAAccessControl is AccessControl {
+    constructor(address owner) {
+        _grantRole(AccessControlConstants.OWNER_ROLE, owner);
+    }
+
+    function setupRole(bytes32 role, address account) external {
```
samlaf:
```suggestion
    function grantRole(bytes32 role, address account) external {
```

samlaf: think renaming to match the OZ function makes more sense?

contracts/src/core/libraries/v3/access-control/AccessControlConstants.sol
samlaf: Also does it make sense for the OWNER to be a role? This way right now for example the contract allows giving the OWNER role to multiple people. Seems like his could lead to bugs? Are we fine with this?

samlaf (dismissed):

contracts/src/core/EigenDAAccessControl.sol
pakim249CAL: In this case, no because then we would be overloading the existing grantRole function inherited from AccessControl. This function is meant to use the DEFAULT_ADMIN_ROLE defined in the AccessControl contract which has ownership over all roles by default, but I leave this role unused.

contracts/src/core/libraries/v3/access-control/AccessControlConstants.sol
pakim249CAL: IMO, there is no problem with allowing multiple owners. It is to the discretion of the contract owner how many of any role there are. What the contract provides is simply a flexible, transparent, and easily understood mechanism for managing roles.

If we want only a single address for a given key, we might just leverage the address directory instead. But IDK if this is really necessary.

contracts/src/core/EigenDAAccessControl.sol
samlaf: I'm already confused. We can't overwrite that function because... its used? or because of some other reason AND we should make sure that nobody ever uses the grantRole function which is unfortunately exposed because of inheritance?
All this is the kind of stuff that should be documented...

samlaf (approved):

contracts/src/core/EigenDAAccessControl.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1758/commits/a44a23b1f5720acd583cac11cde53a17e1bb2760

I decided to simply just use the DEFAULT_ADMIN_ROLE to use the paradigm that openzeppelin intends, which is for their defined DEFAULT_ADMIN_ROLE to be able to grant roles. And so now the contract is basically just OZ's AccessControl with a constructor.

PR #1759, opened 2025-07-11
chore: revert per-quorum reservation payments prs

samlaf:
Got Claude to go through the reverts one by one and fix the merge conflicts (still not 100% why there were conflicts). The diff with 0.9.2 is available in https://www.notion.so/eigen-labs/pr-1759-diffs-with-v0-9-2-22d13c11c3e080fea264d56a1820427e, and it consists of only a few good additions coming from PRs other than the ones reverted in here. Then fixed a few things by hand.

Trying to start Payments afresh. See https://linear.app/eigenlabs/project/payments-recalibration-27b0d6698b61/issues and https://www.notion.so/eigen-labs/Release-2-1-0-22413c11c3e0809ea585eafa44a3fc23?source=copy_link#22413c11c3e080728739e7e2ce126c62 for more info.

This is needed to make our 2.1.0 release, since the per-quorum reservation PRs that are being reverted here contain some racey logic on the disperser which causes tx errors. We want to rethink the architecture and properly serialize payments processing. We will likely un-revert the protobuf PR and just fix the implementation, but for now we consider cleaner to just revert everything.

## Protobuf breaking changes

All the protobuf breaking changes are from per-quorum payments methods that we are removing. This is fine because those only landed on preprod which is internal to us, so these changes won't break any clients.


litt3 (changes requested):
As discussed offline, the following PR should be reapplied:

https://github.com/Layr-Labs/eigenda/pull/1641

PR #1760, opened 2025-07-12
chore: revert ntp clock

samlaf:
DO NOT MERGE until we've discussed this and came to consensus on whether to revert this or not.

I personally believe, and I think many others in the team also do, that adding ntp concerns at the application level was a mistake. This should be taken care of by the environment/devops/deployer/infra person. Furthermore, even if we did decide to want to manage clock time synchrony at the app level, the way the ntp clock was implemented was very subpar: it doesn't use dependency injection, so its impossible to mock during tests, which leads to proxy tests often having flakes by not being to reach the DNS server in time when running as a GHA workflow.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/clients/v2/disperser_client.go
```diff
@@ -216,7 +197,7 @@ func (c *disperserClient) DisperseBlobWithProbe(
 	}
 
 	symbolLength := encoding.GetBlobLengthPowerOf2(uint(len(data)))
-	payment, err := c.accountant.AccountBlob(c.ntpClock.Now().UnixNano(), uint64(symbolLength), quorums)
+	payment, err := c.accountant.AccountBlob(time.Now().UnixNano(), uint64(symbolLength), quorums)
```
litt3, line 200: Probably should pass in a time instance at construct time, rather than doing `time.Now()`

litt3 (approved):

api/clients/v2/disperser_client.go
samlaf: at construct time?? Or do you mean as an argument to DisperseBlobWithProbe?

Actually wdym by "time instance". Like the result of a `time.Now()`, but passed in as argument? What's wrong with doing time.Now() here?

litt3: I mean you should pass an instance of `time.Time` into the disperser client constructor, keep that as a member, and use it as your time source.

IMO this is how we should handle time in pretty much every case, since it lets us write tests where we pass in a mocked time source

samlaf: We talked offline. I'm onboard with this. Would allow us to test different timing scenarios (for eg clock drift on client). However we agreed this should be done in a future PR given that this is just a revert PR.

PR #1762, opened 2025-07-13
fix(proxy)!: `MaxBlobSizeBytes` processing

ethenotethan:
- Fixes debug log to present # of bytes post payload -> blob encoding
- Reduce config ingestion for digital information unit to only support a maximum unit of mib with only base2

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proxy/common/common.go
```diff
@@ -63,14 +63,6 @@ func ParseBytesAmount(s string) (uint64, error) {
 		return uint64(num * 1024 * 1024), nil
 	case "mb":
 		return uint64(num * 1000 * 1000), nil // Decimal megabyte
```
litt3: I think we should just remove the base 10 versions of these units... they just don't have a place in our system.

Practically, it would be too easy for someone to use `MB` and mean binary, but get decimal.

litt3 (approved):

api/proxy/common/common.go
samlaf: +1

ethenotethan: [ddfe6b6](https://github.com/Layr-Labs/eigenda/pull/1762/commits/ddfe6b6126243660c799ece2c95fce677d83380d)

litt3 (approved):

PR #1765, opened 2025-07-15
ci: move v1 client holesky test to live test

samlaf:

Our integration test ci was constantly needing to be restarted after 10 mins because of flakey rpc (see [this run](https://github.com/Layr-Labs/eigenda/actions/runs/16294027237/job/46010739178?pr=1763) for example). Changed the rpc-url to use same as that from the v2 live tests because the 1rpc one was super flaky.

Also moved to only run that CI as part of the twice per day live tests, since that's what we should be doing with tests that run against our live prod systems.

Also moved the test/integration_test under test/v1 to match the test/v2 structure. Otherwise I find it confusing.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/test/v1/integration_test.go
+++ b/test/v1/integration_test.go
@@ -18,6 +18,7 @@ import (
 	"time"
 
 	"github.com/docker/go-units"
+	"github.com/gammazero/workerpool"
 
 	"github.com/stretchr/testify/require"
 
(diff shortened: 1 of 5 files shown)


test/v1/integration_test.go
```diff
@@ -273,7 +273,7 @@ func mustMakeDisperser(t *testing.T, cst core.IndexedChainState, store disperser
 		dockertestPool, dockertestResource, err = deploy.StartDockertestWithLocalstackContainer(localStackPort)
 		if err != nil {
 			teardown()
-			panic("failed to start localstack container")
+			panic("failed to start localstack container" + err.Error())
```
litt3: nit: panic string should have a trailing space, to be formatted correctly with the appended error

litt3 (approved):

test/v1/integration_test.go
samlaf: fixed everywhere: [2d5c409](https://github.com/Layr-Labs/eigenda/pull/1765/commits/2d5c40959759812f714cf54ad93dd093fe49f66b)

PR #1768, opened 2025-07-15
refactor: Add RelayUseSecureGRPC flag to node

dmanc:

Noticed that we use the churner flag in the relay client, this PR simply adds another flag called `RELAY_USE_SECURE_GRPC`.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


node/config.go
```diff
@@ -78,6 +78,7 @@ type Config struct {
 	EnableGnarkBundleEncoding       bool
 	ClientIPHeader                  string
 	UseSecureGrpc                   bool
+	RelayUseSecureGrpc              bool
```
samlaf, line 81: Add comment that the other flag is only for churner now. The flag is properly labeled but not the config field

node/flags/flags.go
```diff
@@ -186,6 +186,12 @@ var (
 		Required: false,
 		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "CHURNER_USE_SECURE_GRPC"),
 	}
+	RelayUseSecureGRPC = cli.BoolTFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "relay-use-secure-grpc"),
+		Usage:    "Whether to use secure GRPC connection to Relay (defaults to true)",
+		Required: false,
```
samlaf, line 192: So this is not required but true by default right?
Can validators even set this to false? Or is this only for inabox/preprod? If so should we hide this flag?

dmanc: Not sure actually, I don't think a validator should ever set this to false in production since it wouldn't be a good security practice and only makes sense to use in a testing environment. Not sure if we should hide the flag or not.

cc: @anupsv

node/config.go
dmanc: Renamed ac03e6196b8b5b147c852a0b46358c44fa9e923f

inabox/deploy/config.go
```diff
@@ -471,14 +472,13 @@ func (env *Config) generateOperatorVars(ind int, name, key, churnerUrl, logPath,
 // Generates retriever .env
 func (env *Config) generateRetrieverVars(ind int, key string, graphUrl, logPath, grpcPort string) RetrieverVars {
 	v := RetrieverVars{
-		RETRIEVER_LOG_FORMAT:                   "text",
-		RETRIEVER_HOSTNAME:                     "",
-		RETRIEVER_GRPC_PORT:                    grpcPort,
-		RETRIEVER_TIMEOUT:                      "10s",
-		RETRIEVER_EIGENDA_DIRECTORY:            env.EigenDA.EigenDADirectory,
-		RETRIEVER_BLS_OPERATOR_STATE_RETRIEVER: env.EigenDA.OperatorStateRetriever,
```
samlaf, line 479: think this still exists? Its just called RETRIEVER_BLS_OPERATOR_STATE_RETRIVER (with the RETRIVER typo).

I remember you asked hope not to fix this typo in one of her PRs. Why was it not typoe'd here though?? I'm so confused.

dmanc: tbh not sure what the retriever is, it's not something we run

dmanc: Oh I see what happened, it's related to the typo. I'm fine leaving a breaking change to this component.

cody-littley (approved):

node/flags/flags.go
anupsv: Validator shouldn't. I think this was part of replay attack somewhere but there's a distinction. If they front it with cloudflare or something similar, which they should, they can potentially terminate SSL at that level and then have the node not need SSL. This is fine but if we can force them to terminate ssl on the node directly using even self signed cert, it's a stronger model. I'm ok hiding the flag as long as we generate the self signed cert and start up the node or error out properly.

samlaf (approved):

node/flags/flags.go
samlaf: oh ssl termination is a good point. let's keep it then I agree, since we don't know how operators are setting this up I'm assuming.

PR #1769, opened 2025-07-16
fix: Initialize connection with validator in background goroutine

cody-littley:
Do not initialize grpc connections on the main batch creation goroutine. In extreme cases, grpc connections can take several seconds to set up.

### DIFF
--- a/disperser/controller/dispatcher.go
+++ b/disperser/controller/dispatcher.go
@@ -216,6 +216,8 @@ func (d *Dispatcher) HandleBatch(
 		validatorProbe.SetStage("pool_submission")
 
 		d.pool.Submit(func() {
+			defer validatorProbe.End()
+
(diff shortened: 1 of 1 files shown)


disperser/controller/dispatcher.go
```diff
@@ -209,48 +209,48 @@ func (d *Dispatcher) HandleBatch(
...
+					Operator:        opID,
+					BatchHeaderHash: batchData.BatchHeaderHash,
+					TimeReceived:    time.Now(),
+					Err:             err,
+				}
+				return
+			}
 
-		d.pool.Submit(func() {
 			defer validatorProbe.End()
```
litt3: this defer should be moved to the top of the func

cody-littley: Good catch, done

litt3 (approved):

PR #1774, opened 2025-07-17
fix: node shouldnt require eigenda directory flag

samlaf:

Preprod Validator was seeing 2025/07/16 22:21:34 application failed: failed to create address directory reader: address directory must be a valid hex address:
When using the old config. EigenDADirectory was added, but should be backward compatible and just seamlessly fallback to using the provided EigenDAServiceManager passed in as argument when Directory is not present.

We will later deprecate and remove the other flags but for now we want the flags to be backwards compatible.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/proxy/.env.exampleV1.holesky
+++ b/api/proxy/.env.exampleV1.holesky
@@ -3,7 +3,7 @@
 # Hex-encoded signer private key. In V1, this key should not be associated with an Ethereum address holding any funds.
 EIGENDA_PROXY_EIGENDA_SIGNER_PRIVATE_KEY_HEX=
 
-# JSON RPC node endpoint for the Ethereum network used for finalizing DA blobs. See available list here: https://docs.eigenlayer.xyz/eigenda/networks/
+# JSON RPC node endpoint for the Ethereum network used for finalizing DA blobs. See available list here: https://docs.eigencloud.xyz/products/eigenda/networks/mainnet
 EIGENDA_PROXY_EIGENDA_ETH_RPC=https://ethereum-holesky-rpc.publicnode.com
 
 # RPC URL of the EigenDA disperser service. Mainnet is `disperser.eigenda.xyz:443`).
--- a/node/node.go
+++ b/node/node.go
@@ -210,14 +210,21 @@ func NewNode(
 		return nil, fmt.Errorf("failed to create new blacklist store: %w", err)
 	}
 
-	// Get service manager address from address directory
-	addressReader, err := eth.NewEigenDADirectoryReader(config.EigenDADirectory, client)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create address directory reader: %w", err)
-	}
-	eigenDAServiceManagerAddr, err := addressReader.GetServiceManagerAddress()
-	if err != nil {
-		return nil, fmt.Errorf("failed to get service manager address: %w", err)
+	// If EigenDADirectory is provided, use it to get service manager addresses
+	// Otherwise, use the provided address (legacy support; will be removed as a breaking change)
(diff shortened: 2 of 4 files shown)


node/node.go
```diff
@@ -210,14 +204,21 @@ func NewNode(
...
+		addressReader, err := eth.NewEigenDADirectoryReader(config.EigenDADirectory, client)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create address directory reader: %w", err)
+		}
+		eigenDAServiceManagerAddr, err = addressReader.GetServiceManagerAddress()
+		if err != nil {
+			return nil, fmt.Errorf("failed to get service manager address from EigenDADirectory: %w", err)
+		}
+	} else {
+		logger.Warn("EigenDADirectory is not set or is not a valid address, using provided EigenDAServiceManagerAddr. " +
```
litt3: Maybe worth logging the value of `EigenDADirectory` if not empty string. I.e. the user tried to configure it, but it wasn't a valid address.

node/node.go
```diff
@@ -210,14 +204,21 @@ func NewNode(
...
-		return nil, fmt.Errorf("failed to get service manager address: %w", err)
+	// If EigenDADirectory is provided, use it to get service manager addresses
+	// Otherwise, use the provided address (legacy support; will be removed as a breaking change)
+	eigenDAServiceManagerAddr := gethcommon.HexToAddress(config.EigenDAServiceManagerAddr)
+	if config.EigenDADirectory != "" && gethcommon.IsHexAddress(config.EigenDADirectory) {
+		addressReader, err := eth.NewEigenDADirectoryReader(config.EigenDADirectory, client)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create address directory reader: %w", err)
+		}
+		eigenDAServiceManagerAddr, err = addressReader.GetServiceManagerAddress()
```
litt3, line 215: Consider checking whether an `eigenDAServiceManagerAddr` is explicitly defined which differs from the address found in the directory. I think it would make sense to error out in this case, since it's a clear misconfiguration

litt3 (approved):

node/node.go
samlaf: [8eef05c](https://github.com/Layr-Labs/eigenda/pull/1774/commits/8eef05ce847a5120a861884ac8c47b1d83750820)

node/node.go
samlaf: [a701d57](https://github.com/Layr-Labs/eigenda/pull/1774/commits/a701d57b2a8a5661958d7728b0cb0e94b645db0e)

litt3 (approved):

api/proxy/.env.exampleV1.holesky
```diff
@@ -3,7 +3,7 @@
 # Hex-encoded signer private key. In V1, this key should not be associated with an Ethereum address holding any funds.
 EIGENDA_PROXY_EIGENDA_SIGNER_PRIVATE_KEY_HEX=
 
-# JSON RPC node endpoint for the Ethereum network used for finalizing DA blobs. See available list here: https://docs.eigenlayer.xyz/eigenda/networks/
+# JSON RPC node endpoint for the Ethereum network used for finalizing DA blobs. See available list here: https://docs.eigencloud.xyz/products/eigenda/networks/mainnet
```
dmanc, line 6: nit: I don't think we should include changes not relevant to the PR since if we wanted to revert we will revert these docs.

dmanc: Feel free to keep it though since it's not a huge change

samlaf: I agree. I got lazy. Don't see why we would revert this but I agree with you..

PR #1777, opened 2025-07-18
ci(proxy): improve flag test flakiness

samlaf:
test was just sleeping 10 seconds and trying health endpoint once. Had many flakes (eg https://github.com/Layr-Labs/eigenda/actions/runs/16373235673/job/46266622666?pr=1736) because proxy sometimes takes longer to startup in CI for random reasons.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/proxy/scripts/test-proxy-startup-with-env-vars.sh
+++ b/api/proxy/scripts/test-proxy-startup-with-env-vars.sh
@@ -38,12 +38,17 @@ PID=$!
 # Ensure we kill the process on script exit
 trap "kill $PID" EXIT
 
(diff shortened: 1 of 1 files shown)


samlaf:
> I thought we had said we were considering deleting the example flag files? If we did that, then this test would go away, right?
> 
> I'm fine merging this, though, since we _do_ have these tests for now, and this removes flakiness

I'm not sure, we should discuss this over standup.
My undrstanding is that we wanted to remove the duplicate files (we have like 6 right now), and only keep a minimal one with the non-default values, which clearly show to customers which values MUST be populated by them with network-specific values.

I will merge this for now, but agree we still need to improve on this and find a better solution.

PR #1778, opened 2025-07-18
docs(router): owner should be a timelock

samlaf:

Closes DAINT-635

I personally don't like so much that the ABN has to be set far enough in the future with a buffer to make sure that if the timelock doesn't get executed too fast then the txn will revert because ABN will be in the past.

Feel like it would be simpler to have the logic inside the router itself, although @anupsv did mention that one benefit of having the timelock logic in an external contract is that that contract can have a bypass functionality to allow immediately activating a new cert verifier in case of emergency.

So overall I think it makes sense to just document this behavior, and roll with this. @pakim249CAL thoughts?

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


pakim249CAL:
Maybe a crazy idea, but another way we could resolve this is by removing the check for whether the ABN is in the future when adding a cert verifier. 

This would allow for past ABN cert verifier results that previously failed to possibly succeed.

Is this actually a problem if we have a timelock?

pakim249CAL (approved):

PR #1782, opened 2025-07-21
chore(proxy-clients): new sub go module

samlaf:
Closes DAINT-644.

We were still using the eigenda-proxy/clients package. Now moving this package over to monorepo.

This is also a good chance to cleanup the weird godoc behavior that won't listen to our retract directive in go.mod to remove the 1.0.1 version.

So with this new package our client docs will at least point to the most up-to-date examples.

TODO: publish this new module as v0.2.1 so that it lands on pkg.dev (using same version as the current one for the eigenda-proxy/clients pkg)

TODO: will need to update our eigencloud docs to point to this new module.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -9,9 +9,6 @@ go 1.24
 // See https://tip.golang.org/doc/devel/release#go1.24.0
 toolchain go1.24.4
 
-// We need to keep eigenda-proxy clients separate because they are imported from modules that are not on go1.24 yet.
-replace github.com/Layr-Labs/eigenda-proxy/clients => ./api/proxy/clients
-
 // Pointing to latest eigenda-develop commit that contains https://github.com/Layr-Labs/optimism/pull/49
 // TODO: update to a proper version once we make the next release.
 replace github.com/ethereum-optimism/optimism => github.com/Layr-Labs/optimism v1.13.1-0.20250616152614-59132dd56ae5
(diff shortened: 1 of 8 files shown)


go.mod
```diff
@@ -23,7 +20,9 @@ replace github.com/ethereum-optimism/optimism => github.com/Layr-Labs/optimism v
 replace github.com/ethereum/go-ethereum => github.com/ethereum-optimism/op-geth v1.101503.2-rc.5
 
 require (
-	github.com/Layr-Labs/eigenda-proxy/clients v1.0.1
+	// TODO: update this to correct version after we've tested that it works
+	// should we use 0.2.1 which is the current eigenda-proxy/clients module version?
+	github.com/Layr-Labs/eigenda/api/proxy/clients v0.1.0
```
samlaf: I've tested this and it works. Just need to decide which version we want to publish:
- restart @ 0.1.0 ?
- 0.2.1 to keep same version as previous module?
- 2.1.0 to match the new monorepo release version?

cc @ethenotethan

ethenotethan: in stronger favor of matching monorepo release to avoid having to do separate version mgmt for this - fail to see any immediate drawbacks with this approach

samlaf: Hmm not sure that's possible though. Or do you mean everytime we create a `v2.x.y` tag for main monorepo go.mod, we also create a `api/proxy/clients/v2.x.y` tag for this?

litt3: It sounds like @ethenotethans suggestion might be to bundle the release of the `clients` in with all the other releases entirely. @samlaf, you don't think that would work?

litt3 (dismissed):

go.mod
samlaf: hmm, good question. I guess it could work. I was assuming cannot work because its not part of the same go module, but that doesn't mean we can't still bundle them into a single release with matching version numbers..

samlaf: Getting back to this. Don't think it makes sense, because most releases won't have any proxy client changes, so doesn't make sense to tag a release randomly imo. A release for a golang module is simply pushing a tag, so it's very easy to do and we can do it whenever needed. I think I'm in favor of keeping this module separate and just bumping it whenever we need it.

litt3 (approved):

PR #1783, opened 2025-07-21
feat: SSH utilities for LittDB CLI

cody-littley:

Adds new utilities that will eventually be used by the LittDB CLI.


litt/util/ssh.go
```diff
@@ -0,0 +1,202 @@
...
+	logger logging.Logger,
+	user string,
+	host string,
+	port uint64,
+	keyPath string,
+	verbose bool,
+) (*SSHSession, error) {
+	config := &ssh.ClientConfig{
+		User:            user,
+		HostKeyCallback: ssh.InsecureIgnoreHostKey(),
```
anupsv: Probably don't want this ? We want to verify the host key right ?

litt/util/ssh.go
```diff
@@ -0,0 +1,202 @@
...
+			}
+		}
+	}
+
+	return filteredFiles, nil
+}
+
+// Mkdirs creates the specified directory on the remote machine, including any necessary parent directories.
+func (s *SSHSession) Mkdirs(path string) error {
+	_, stderr, err := s.Exec(fmt.Sprintf("mkdir -p %s", path))
```
anupsv: We need to quote and escape the path to prevent injection.

litt/util/ssh.go
```diff
@@ -0,0 +1,202 @@
...
+	if err != nil {
+		return fmt.Errorf("failed to close SSH client: %w", err)
+	}
+
+	return nil
+}
+
+// Search for all files matching a regex inside a file tree at the specified root path.
+func (s *SSHSession) FindFiles(root string, extensions []string) ([]string, error) {
+	command := fmt.Sprintf("find \"%s\" -type f", root)
```
anupsv, line 106: Escape is still possible with manipulating quotes. I believe `q` does it correctly.

litt/util/ssh.go
cody-littley: If I understand this correctly, this is just a callback to check if we like server's key. On the command line, this is managed the `~/.ssh/known_hosts` file (i.e. where it prompts you to confirm if it's the first time you are talking to a host).

Is it a problem to ignore this check? If we want to permit this check, what policy should we use?

litt/util/ssh.go
cody-littley: Added quotes to support paths that contain spaces. IMO, we shouldn't care about the user of this utility injecting commands, as the user of this utility will already have access to both the server and the client machines. 

- this utility is used by a CLI program, meaning the "attacker" would already have access to run bash commands on the client machine
- in order for this utility to work, there must be valid and registered SSH keys for the server on the client machine. An attacker could simply ssh to the server and do whatever they wanted.

litt/util/ssh.go
cody-littley: This doesn't need to handle adversarial input. This tool will be used by an administrator of both the client and server machines, and such an individual will have many other ways to attack either machine.

litt/util/ssh.go
cody-littley: I made it so that this now checks the known hosts file.

litt/util/ssh.go
```diff
@@ -0,0 +1,222 @@
...
+			}
+		}
+	}
+
+	return filteredFiles, nil
+}
+
+// Mkdirs creates the specified directory on the remote machine, including any necessary parent directories.
+func (s *SSHSession) Mkdirs(path string) error {
+	_, stderr, err := s.Exec(fmt.Sprintf("mkdir -p '%s'", path))
```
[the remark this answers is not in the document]
cody-littley: Not a real vulnerability. Any user capable of using this tool already has administrative control over both the client and server machines.

litt/util/ssh.go
```diff
@@ -0,0 +1,222 @@
...
+		return fmt.Errorf("failed to create directory '%s': %w, stderr: %s", path, err, stderr)
+	}
+
+	return nil
+}
+
+// Rsync transfers files from the local machine to the remote machine using rsync. The throttle is ignored
+// if less than or equal to 0.
+func (s *SSHSession) Rsync(sourceFile string, destFile string, throttleMB float64) error {
+	sshCmd := fmt.Sprintf("ssh -i %s -p %d -o StrictHostKeyChecking=no", s.keyPath, s.port)
```
litt3: > StrictHostKeyChecking=no

Just double-checking that this is actually the option you want to use here.

litt3 (dismissed):

litt/util/ssh.go
cody-littley: I've removed this for now. I can't recall why I added it, but if we run into difficulty when we start testing the full backup workflow, I'll circle back on this.

litt3 (approved):

PR #1785, opened 2025-07-21
chore: copy registry coordinator

pakim249CAL:

EigenDA requires some custom logic in the Registry Coordinator, but for now there is no good option to add this custom logic. This PR copies the registry coordinator into our repository from the middleware submodule so that it can be modified.

Here are a list of questions that I thought of that need to be answered about this PR. I'd like to organize the commits accordingly:

* Is the registry coordinator copied correctly?
* Is the fixes and renaming kosher? (Must be a separate commit from copying)
* Are all references changed properly to the copied contract?
* Do we need any additional documentation on this copied contract?

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,6 +16,7 @@ lightnode/docker/args.sh
 .idea
 .env
 .vscode
+.serena
 
 icicle/*
 
--- a/api/clients/v2/metrics/accountant.go
+++ b/api/clients/v2/metrics/accountant.go
@@ -17,12 +17,14 @@ var (
 
 type AccountantMetricer interface {
 	RecordCumulativePayment(accountID string, wei *big.Int)
+	RecordReservationPayment(remainingCapacity float64)
 
 	Document() []metrics.DocumentedMetric
 }
 
 type AccountantMetrics struct {
-	CumulativePayment *prometheus.GaugeVec
+	CumulativePayment            *prometheus.GaugeVec
+	ReservationRemainingCapacity prometheus.Gauge
 
 	factory *metrics.Documentor
 }
@@ -43,6 +45,12 @@ func NewAccountantMetrics(registry *prometheus.Registry) AccountantMetricer {
 		}, []string{
 			"account_id",
 		}),
+		ReservationRemainingCapacity: factory.NewGauge(prometheus.GaugeOpts{
+			Name:      "reservation_remaining_capacity",
+			Namespace: namespace,
+			Subsystem: accountantSubsystem,
+			Help:      "Remaining capacity in reservation bucket (symbols)",
+		}),
 		factory: factory,
 	}
 }
@@ -57,6 +65,10 @@ func (m *AccountantMetrics) RecordCumulativePayment(accountID string, wei *big.I
 	m.CumulativePayment.WithLabelValues(accountID).Set(gweiFloat64)
 }
 
+func (m *AccountantMetrics) RecordReservationPayment(remainingCapacity float64) {
+	m.ReservationRemainingCapacity.Set(remainingCapacity)
+}
+
 func (m *AccountantMetrics) Document() []metrics.DocumentedMetric {
 	return m.factory.Document()
 }
@@ -67,6 +79,9 @@ type noopAccountantMetricer struct {
 var NoopAccountantMetrics AccountantMetricer = new(noopAccountantMetricer)
 
 func (n *noopAccountantMetricer) RecordCumulativePayment(_ string, _ *big.Int) {
+}
+
+func (n *noopAccountantMetricer) RecordReservationPayment(_ float64) {
 }
 
 func (n *noopAccountantMetricer) Document() []metrics.DocumentedMetric {
--- a/api/proxy/docs/metrics_out.txt
+++ b/api/proxy/docs/metrics_out.txt
@@ -8,5 +8,6 @@
 | eigenda_proxy_secondary_requests_total              | Total requests to the secondary storage            | backend_type,method,status                 | counter   |
 | eigenda_proxy_secondary_request_duration_seconds    | Histogram of secondary storage request durations   | backend_type                               | histogram |
 | eigenda_accountant_cumulative_payment               | Current cumulative payment balance (gwei)          | account_id                                 | gauge     |
+| eigenda_accountant_reservation_remaining_capacity   | Remaining capacity in reservation bucket (symbols) |                                            | gauge     |
 | eigenda_dispersal_blob_size_bytes                   | Size of blobs created from payloads in bytes       |                                            | histogram |
 | eigenda_retrieval_payload_size_bytes                | Size of decoded payloads in bytes                  |                                            | histogram |
--- a/core/payments/clientledger/client_ledger.go
+++ b/core/payments/clientledger/client_ledger.go
@@ -142,7 +142,7 @@ func (cl *ClientLedger) debitReservationOnly(
 ) (*core.PaymentMetadata, error) {
 	// As the client, "now" and the dispersal time are the same. The client is responsible for populating the
 	// dispersal time when constructing the payment header, and it does so with its conception of "now"
-	success, err := cl.reservationLedger.Debit(now, now, blobLengthSymbols, quorums)
+	success, remainingCapacity, err := cl.reservationLedger.Debit(now, now, blobLengthSymbols, quorums)
 	if err != nil {
 		var timeMovedBackwardErr *reservation.TimeMovedBackwardError
 		if errors.As(err, &timeMovedBackwardErr) {
@@ -153,6 +153,8 @@ func (cl *ClientLedger) debitReservationOnly(
 		// all other modes of failure are fatal
 		panic(fmt.Sprintf("reservation debit failed: %v", err))
 	}
+
+	cl.accountantMetricer.RecordReservationPayment(remainingCapacity)
 
 	if !success {
 		return nil, fmt.Errorf(
@@ -197,7 +199,7 @@ func (cl *ClientLedger) debitReservationOrOnDemand(
 ) (*core.PaymentMetadata, error) {
 	// As the client, "now" and the dispersal time are the same. The client is responsible for populating the
 	// dispersal time when constructing the payment header, and it does so with its conception of "now"
-	success, err := cl.reservationLedger.Debit(now, now, blobLengthSymbols, quorums)
+	success, remainingCapacity, err := cl.reservationLedger.Debit(now, now, blobLengthSymbols, quorums)
 	if err != nil {
 		var timeMovedBackwardErr *reservation.TimeMovedBackwardError
 		if errors.As(err, &timeMovedBackwardErr) {
@@ -208,6 +210,8 @@ func (cl *ClientLedger) debitReservationOrOnDemand(
 		// all other modes of failure are fatal
 		panic(fmt.Sprintf("reservation debit failed: %v", err))
 	}
+
+	cl.accountantMetricer.RecordReservationPayment(remainingCapacity)
 
 	if success {
 		paymentMetadata, err := core.NewPaymentMetadata(cl.accountID, now, nil)
@@ -260,10 +264,12 @@ func (cl *ClientLedger) RevertDebit(
 		enforce.NotNil(cl.reservationLedger,
 			"payment metadata is for a reservation payment, but ReservationLedger is nil")
 
-		err := cl.reservationLedger.RevertDebit(cl.getNow(), blobSymbolCount)
+		remainingCapacity, err := cl.reservationLedger.RevertDebit(cl.getNow(), blobSymbolCount)
 		if err != nil {
 			return fmt.Errorf("revert reservation debit: %w", err)
 		}
+
+		cl.accountantMetricer.RecordReservationPayment(remainingCapacity)
 	}
 
 	return nil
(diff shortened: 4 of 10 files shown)


contracts/src/core/EigenDARegistryCoordinator.sol
samlaf, line 1: Just to be sure and not do a needless review, this contract and the storage one are at this point unmodified copies of the code in middleware? Your PR description stopped midsentence:

contracts/src/core/EigenDARegistryCoordinatorStorage.sol
```diff
@@ -0,0 +1,96 @@
...
+    mapping(bytes32 => bool) public isChurnApproverSaltUsed;
+    /// @notice mapping from quorum number to the latest block that all quorums were updated all at once
+    mapping(uint8 => uint256) public quorumUpdateBlockNumber;
+
+    /// @notice the dynamic-length array of the registries this coordinator is coordinating
+    address[] public registries;
+    /// @notice the address of the entity allowed to sign off on operators getting kicked out of the AVS during registration
+    address public churnApprover;
+    /// @notice the address of the entity allowed to eject operators from the AVS
+    address public ejector;
```
samlaf, line 72: out of curiosity, ae these things that you eventually want to add to either the EigenDADirectory or the EigenDAAccessControl?

Also how are you thinking of those contracts? Would Directory contain contract + actor addresses, and the AccessControl their roles?

samlaf (commented):
Is this ready for review? Would like to know whats copied and whats new, not sure right now. Is it just that the CoordinatorStorage is new to use your storage patterns, but all the logic in the coordinator remains the same (for now?)?

Also I see that this is built on top of `feat/access-control`. Let's please merge access-control first to keep them separate!

contracts/src/core/EigenDARegistryCoordinator.sol
pakim249CAL: Sorry, it was "done" at one point but I discovered more changes to be made afterwards 😅 I've returned this to a draft state for now.

contracts/src/core/EigenDARegistryCoordinatorStorage.sol
pakim249CAL: Yes, these are things I want to add to access control.

Rather than dictating the purpose of the directory and access control, I think it is more useful to consider their structure. The address directory maps a key to a single address, while access control maps a key to a SET of addresses, where it is efficient to check an address's membership in a set.

All variables that fit into these categories would be a good fit for access control/directory.

After discussions, I decided that In this PR and for the immediate future, I do not plan on implementing these changes in this contract to limit the scope for auditing. The EigenDADirectory and AccessControl contracts can be audited at a later point. And so this particular point is likely not going to be relevant to this PR for now.

samlaf: > "it is more useful to consider their structure"

You have mentioned things like this a few times in the past, and I'm curious where you got this from/what is your mental model for saying this.

My initial guess is that you're trying to dopt the https://en.wikipedia.org/wiki/Separation_of_mechanism_and_policy principle?

But saying that "just care about its structure" is taking this a bit too far, and actually not applying the principle how it should. Just because 2 systems use the same underlying data structure doesn't mean they should be the same! That's why we have separation of concerns / classes / modules / etc. Policies should imo dictate how systems are architected.

Unless there's something absolutely funky about the smart contract case with constrained resources that makes modularity/encapsulation not achievable? I kind of doubt it though... but would be curious if this is your reasoning.

pakim249CAL: I don't think it's going too far, at least when it comes to addresses. There is really no use case for addresses other than as an identifier, and it just so happens that the only real use cases for an identifier is to use it for validation and identification, hence access control and directory. I would agree with you if the type was more general.

pakim249CAL: I don't think this is too relevant for this case, because addresses are only used for identification, and are basically already type safe. I don't think there is much use for addresses beyond access control and pointers. TBH I completely fail to think of a use case beyond this, and so I don't know if we need to entertain the possibility of using these contracts for something else. We can revisit the discussion when it comes to more general types like configs.

contracts/src/core/EigenDARegistryCoordinatorStorage.sol
samlaf, line 1: Looks good. Checked diff against the RegistryCoordinator we are pointing at in middleware and its only stylistic changes:
``` $ diff contracts/src/core/EigenDARegistryCoordinatorStorage.sol contracts/lib/eigenlayer-middleware/src/RegistryCoordinatorStorage.sol    
4,9c4,9
< import {IBLSApkRegistry} from "lib/eigenlayer-middleware/src/interfaces/IBLSApkRegistry.sol";
< import {IStakeRegistry} from "lib/eigenlayer-middleware/src/interfaces/IStakeRegistry.sol";
< import {IIndexRegistry} from "lib/eigenlayer-middleware/src/interfaces/IIndexRegistry.sol";
< import {IServiceManager} from "lib/eigenlayer-middleware/src/interfaces/IServiceManager.sol";
< import {IRegistryCoordinator} from "lib/eigenlayer-middleware/src/interfaces/IRegistryCoordinator.sol";
< import {ISocketRegistry} from "lib/eigenlayer-middleware/src/interfaces/ISocketRegistry.sol";
---
> import {IBLSApkRegistry} from "./interfaces/IBLSApkRegistry.sol";
> import {IStakeRegistry} from "./interfaces/IStakeRegistry.sol";
> import {IIndexRegistry} from "./interfaces/IIndexRegistry.sol";
> import {IServiceManager} from "./interfaces/IServiceManager.sol";
> import {IRegistryCoordinator} from "./interfaces/IRegistryCoordinator.sol";
> import {ISocketRegistry} from "./interfaces/ISocketRegistry.sol";
11,16c11
< abstract contract EigenDARegistryCoordinatorStorage is IRegistryCoordinator {
<     /**
<      *
<      *                            CONSTANTS AND IMMUTABLES
<      *
<      */
---
> abstract contract RegistryCoordinatorStorage is IRegistryCoordinator {
17a13,16
>     /*******************************************************************************
>                                CONSTANTS AND IMMUTABLES 
>     *******************************************************************************/
> 
19,21c18,19
<     bytes32 public constant OPERATOR_CHURN_APPROVAL_TYPEHASH = keccak256(
<         "OperatorChurnApproval(address registeringOperator,bytes32 registeringOperatorId,OperatorKickParam[] operatorKickParams,bytes32 salt,uint256 expiry)OperatorKickParam(uint8 quorumNumber,address operator)"
<     );
---
>     bytes32 public constant OPERATOR_CHURN_APPROVAL_TYPEHASH =
>         keccak256("OperatorChurnApproval(address registeringOperator,bytes32 registeringOperatorId,OperatorKickParam[] operatorKickParams,bytes32 salt,uint256 expiry)OperatorKickParam(uint8 quorumNumber,address operator)");
48,52c46,48
<     /**
<      *
<      *                                    STATE
<      *
<      */
---
>     /*******************************************************************************
>                                        STATE 
>     *******************************************************************************/
```

contracts/src/core/EigenDARegistryCoordinator.sol
```diff
@@ -0,0 +1,983 @@
...
+    EigenDARegistryCoordinatorStorage,
+    ISignatureUtils
+{
+    using BitmapUtils for *;
+    using BN254 for BN254.G1Point;
+
+    modifier onlyEjector() {
+        _checkEjector();
+        _;
+    }
```
samlaf, line 46: do the () matter? see diff with middleware RegistryCoordinator
image

contracts/src/core/EigenDARegistryCoordinator.sol
samlaf, line 1: diff with middleware ReigstryCoordinator is pretty large (526 lines)... looks like all stylistic changes but might be nice to format it using the same `forge fmt` version so that we can focus on the actual meaningful changes.

```
$ diff contracts/src/core/EigenDARegistryCoordinator.sol contracts/lib/eigenlayer-middleware/src/RegistryCoordinator.sol | wc -l
     526
```

samlaf: example stylistic differences 

image

contracts/src/core/EigenDARegistryCoordinator.sol
```diff
@@ -0,0 +1,983 @@
...
+     */
+    function updateOperatorsForQuorum(address[][] calldata operatorsPerQuorum, bytes calldata quorumNumbers)
+        external
+        onlyWhenNotPaused(PAUSED_UPDATE_OPERATOR)
+    {
+        // Input validation
+        // - all quorums should exist (checked against `quorumCount` in orderedBytesArrayToBitmap)
+        // - there should be no duplicates in `quorumNumbers`
+        // - there should be one list of operators per quorum
+        require(
```
samlaf, line 290: You got rid of the (unused) quorumBitmap that was defined above this line. Are we sure that wasn't used to check that the bytes are ordered?

image

pakim249CAL: Yes it seems this is what this does. My mistake. I reintroduce it here.

https://github.com/Layr-Labs/eigenda/pull/1785/commits/349f2c73cc6e99c897e61ed30c7d6f0c3f334308

contracts/src/core/EigenDARegistryCoordinator.sol
pakim249CAL: The diff is actually the result of forge fmt, because the submodule is not formatted with forge fmt. I cannot import the changes into our repo without breaking CI without doing forge fmt.

Maybe this PR would be worth rebasing into separate commits for easier reviewing? not sure.

contracts/src/core/EigenDARegistryCoordinator.sol
pakim249CAL: It does not matter. It seems this was added during the formatting of the code. There are no parameters, so the syntax is identical.

contracts/src/core/EigenDARegistryCoordinator.sol
samlaf: or how about we create a branch in middleware which just formats the code using forge fmt, and then we can diff against that?

contracts/src/core/EigenDARegistryCoordinator.sol
samlaf: i remember seeing it in 2 places. might have misremembered though...

pakim249CAL: This should be easier to verify now with the smaller diff.

contracts/src/core/EigenDARegistryCoordinator.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1785/commits/e1dfb6359d731d2fffc61903b4474219fabfdb25

contracts/src/core/EigenDARegistryCoordinator.sol
samlaf: yep looks good seems like there's only that one.
image
nit: no need to assign to an (unused) variable

contracts/src/core/EigenDARegistryCoordinator.sol
samlaf: Looks good now, 40 line diff. checked all its just imports and other normal stuff.

contracts/src/core/EigenDARegistryCoordinatorStorage.sol
samlaf: Even smaller now:
```
$ diff contracts/src/core/EigenDARegistryCoordinatorStorage.sol contracts/lib/eigenlayer-middleware/src/RegistryCoordinatorStorage.sol
4,9c4,9
< import {IBLSApkRegistry} from "lib/eigenlayer-middleware/src/interfaces/IBLSApkRegistry.sol";
< import {IStakeRegistry} from "lib/eigenlayer-middleware/src/interfaces/IStakeRegistry.sol";
< import {IIndexRegistry} from "lib/eigenlayer-middleware/src/interfaces/IIndexRegistry.sol";
< import {IServiceManager} from "lib/eigenlayer-middleware/src/interfaces/IServiceManager.sol";
< import {IRegistryCoordinator} from "lib/eigenlayer-middleware/src/interfaces/IRegistryCoordinator.sol";
< import {ISocketRegistry} from "lib/eigenlayer-middleware/src/interfaces/ISocketRegistry.sol";
---
> import {IBLSApkRegistry} from "./interfaces/IBLSApkRegistry.sol";
> import {IStakeRegistry} from "./interfaces/IStakeRegistry.sol";
> import {IIndexRegistry} from "./interfaces/IIndexRegistry.sol";
> import {IServiceManager} from "./interfaces/IServiceManager.sol";
> import {IRegistryCoordinator} from "./interfaces/IRegistryCoordinator.sol";
> import {ISocketRegistry} from "./interfaces/ISocketRegistry.sol";
11c11
< abstract contract EigenDARegistryCoordinatorStorage is IRegistryCoordinator {
---
> abstract contract RegistryCoordinatorStorage is IRegistryCoordinator {
27c27
<     uint16 internal constant BIPS_DENOMINATOR = 10000;
---
>     uint16 internal constant BIPS_DENOMINATOR = 10_000;
```

PR #1786, opened 2025-07-21
feat: Improve claude log preprocessing command

litt3:
Makes some small improvements to the batch preprocessing claude command:

- additional search patterns
- usage help if user doesn't specify log directory
- formatting + cleanup


.claude/commands/preprocess-logs.md
```diff
@@ -65,44 +80,71 @@ at this point: we are simply generating an index of lines that might potentially
...
-Use a **two-stage search strategy** to balance precision and recall:
+### Search Profiles
+
+Use targeted search profiles based on the type of failures you're looking for. If the user didn't specify what you are
+searching for, you should iteratively search using each profile.
+
+#### Profile 1: Test Failures
+For standard test output failures:
+```bash
+rg --line-number --ignore-case --json -C 5 -- "^[-]{3} FAIL:|\\s+FAIL\$|\\s+FAIL\\t|\\[FAILED\\]|panic: test timed out" "<original_log_directory>/analysis/shards/" > "<original_log_directory>/analysis/search_results/test_failures_search.jsonl"
```
samlaf, line 91: one question I have is how do we make these files more readable for humans. I would want to request comments that explain what this is doing so that a human can improve on it in the future, but I'm assuming if we put comments here they will be ingested by the LLM, and potentially take some context needlessly, or worse confuse it?

litt3: Yeah, I intentionally didn't leave comments on what these commands are specifically doing. It ends up looking like a regurgitation of the `rg` docs, which generally isn't very helpful to claude, and it introduces content duplication.

Is it just these explicit `rg` commands that you find unreadable, or do you think the document *in general* suffers from poor readability? IMO the document itself ought to be digestible (and if it isn't, then this needs more work), but it's not the end of the world if a specific command like this is tailored for an agent. While not as easily accessible as a comment, it's also not too hard to just ask claude to summarize what such a command is doing, if an engineer wants to understand specifics.

samlaf: that's a good point actually. Makes it even more explicit now that comments should almost exclusively be targetted for expressing the author's intentions. "explanation" docs can be generated automatically by llms at read time.

PR #1788, opened 2025-07-22
fix: remove blacklist wiring from nodes

anupsv:

We are removing the wiring of the blacklist feature. I.E we are removing the place where it is added to the list and the place where it was being checked. We are also removing the integration test which was testing these features.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/node/grpc/server_v2.go
+++ b/node/grpc/server_v2.go
@@ -134,13 +134,6 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
 		return nil, api.NewErrorInvalidArg(fmt.Sprintf("failed to serialize batch header hash: %v", err))
 	}
 
-	// If the disperser is blacklisted and the blob authenticator is not nil, return an error
-	// we don't want to blacklist the disperser if the blob authenticator is nil since that indicated v1
-	if s.node.BlacklistStore.IsBlacklisted(ctx, in.GetDisperserID()) && s.config.EnableV2 {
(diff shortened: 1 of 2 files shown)


node/grpc/server_v2.go
```diff
@@ -134,13 +134,6 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
 		return nil, api.NewErrorInvalidArg(fmt.Sprintf("failed to serialize batch header hash: %v", err))
 	}
 
-	// If the disperser is blacklisted and the blob authenticator is not nil, return an error
-	// we don't want to blacklist the disperser if the blob authenticator is nil since that indicated v1
-	if s.node.BlacklistStore.IsBlacklisted(ctx, in.GetDisperserID()) && s.config.EnableV2 {
```
litt3, line 139: Given it's not longer used, consider also deleting the instantiation of the `BlacklistStore` inside `node.go`

samlaf:
I think it LGTM. Super tired though and need some sleep so I'll let @litt3 and @dmanc review this one.

node/grpc/server_v2.go
anupsv: Will make the PR bigger, but sure.

anupsv: done

dmanc (approved):
Deployed the latest commit and everything looks good.

PR #1791, opened 2025-07-22
feat: adds account index to blobmetadatastore-v2

pschork:
Blob explorer is building an Accounts page and needs an endpoint that provides a list of accounts that have recently dispersed blobs over a given look back interval (ie 24h, 7d, 14d). This change adds an Account partition to blobmetadatastore and AccountUpdatedAtIndex GSI that orders the partition by timestamp of last dispersal.

Flow
- Accounts are upserted with (address, updated_at) after blob has been stored by disperser.
- The account update is async so as to not add latency to dispersal response.
- Dataapi returns a list of `[](address, dispersed_at)`

### DIFF
--- a/disperser/apiserver/disperse_blob_v2.go
+++ b/disperser/apiserver/disperse_blob_v2.go
@@ -34,6 +34,13 @@ func (s *DispersalServerV2) DisperseBlob(ctx context.Context, req *pb.DisperseBl
 		return nil, api.NewErrorInvalidArg(fmt.Sprintf("failed to validate the request: %v", err))
 	}
 
+	// Update AccountIndex after successful validation
+	accountID := blobHeader.PaymentMetadata.AccountID
+	timestamp := uint64(time.Now().Unix())
+	if err := s.blobMetadataStore.UpdateAccountIndex(ctx, accountID, timestamp); err != nil {
+		s.logger.Warn("failed to update account index", "accountID", accountID.Hex(), "error", err)
+	}
+
 	if err := s.checkBlobExistence(ctx, blobHeader); err != nil {
 		return nil, err
 	}
--- a/disperser/apiserver/disperse_blob_v2.go
+++ b/disperser/apiserver/disperse_blob_v2.go
@@ -60,7 +60,12 @@ func (s *DispersalServerV2) DisperseBlob(ctx context.Context, req *pb.DisperseBl
 	go func() {
 		accountID := blobHeader.PaymentMetadata.AccountID
 		timestamp := uint64(time.Now().Unix())
-		if err := s.blobMetadataStore.UpdateAccount(context.Background(), accountID, timestamp); err != nil {
+
+		// Use a timeout context for the async database operation
+		ctx, cancel := context.WithTimeout(context.Background(), 5*time.Second)
+		defer cancel()
+
+		if err := s.blobMetadataStore.UpdateAccount(ctx, accountID, timestamp); err != nil {
 			s.logger.Warn("failed to update account", "accountID", accountID.Hex(), "error", err)
 		}
 	}()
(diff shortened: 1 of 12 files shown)


disperser/apiserver/disperse_blob_v2.go
```diff
@@ -56,6 +56,15 @@ func (s *DispersalServerV2) DisperseBlob(ctx context.Context, req *pb.DisperseBl
 	}
 	s.logger.Debug("stored blob", "blobKey", blobKey.Hex())
 
+	// Update Account asynchronously after successful blob storage
+	go func() {
+		accountID := blobHeader.PaymentMetadata.AccountID
+		timestamp := uint64(time.Now().Unix())
+		if err := s.blobMetadataStore.UpdateAccount(context.Background(), accountID, timestamp); err != nil {
```
dmanc: Should we set some ctx with a timeout?

disperser/common/v2/blobstore/dynamo_metadata_store.go
```diff
@@ -476,6 +480,67 @@ func (s *BlobMetadataStore) GetBlobMetadataByAccountID(
...
+// UpdateAccount updates the Account partition to track account activity.
+// This method performs an upsert operation, creating or updating an entry for the given account
+// with the current timestamp.
+func (s *BlobMetadataStore) UpdateAccount(ctx context.Context, accountID gethcommon.Address, timestamp uint64) error {
+	s.logger.Debug("updating account", "accountID", accountID.Hex(), "timestamp", timestamp)
+
+	item := commondynamodb.Item{
+		"PK":          &types.AttributeValueMemberS{Value: accountPK},
+		"SK":          &types.AttributeValueMemberS{Value: accountID.Hex()},
+		"Address":     &types.AttributeValueMemberS{Value: accountID.Hex()},
```
dmanc: Do we need both SK and Address?

disperser/common/v2/blobstore/dynamo_metadata_store.go
```diff
@@ -476,6 +480,67 @@ func (s *BlobMetadataStore) GetBlobMetadataByAccountID(
...
+// GetAccounts returns accounts within the specified lookback period (newest first)
+func (s *BlobMetadataStore) GetAccounts(ctx context.Context, lookbackSeconds uint64) ([]*v2.Account, error) {
+	s.logger.Debug("querying accounts", "lookbackSeconds", lookbackSeconds)
+
+	// Calculate the cutoff timestamp
+	now := uint64(time.Now().Unix())
+	cutoffTime := now - lookbackSeconds
+
+	// Query the AccountUpdatedAtIndex GSI with time filter
+	// All account records have AccountType = "AccountIndex" which allows us to query
```
dmanc: What's the difference between AccountType and PK? Like couldn't we also just check that PK = "Account"?

dmanc (commented):
lgtm, couple questions

disperser/common/v2/blobstore/dynamo_metadata_store.go
pschork: Yea ~we do.~ we can dump the address field

disperser/common/v2/blobstore/dynamo_metadata_store.go
pschork: Each item needs an AccountType field (set to some string constant) to enable GSI on `(PK:AccountType, SK:UpdatedAt)`. This ensures that the GSI keeps the AccountUpdatedAtIndex in 1 partition.

disperser/common/v2/blobstore/dynamo_metadata_store.go
dmanc: Can we do something like SK = Address#0xAddress maybe?

disperser/common/v2/blobstore/dynamo_metadata_store.go
pschork: Renamed `AccountType` to `AccountIndex` for hopefully better context.

> Like couldn't we also just check that PK = "Account"?

Yea we could do a table scan, but we'd have to sort in post.

dmanc (dismissed):

disperser/apiserver/disperse_blob_v2.go
pschork: Done

dmanc (approved):

dmanc (approved):

PR #1792, opened 2025-07-22
chore: cherry pick 378663d92ad02cf2c56bbf11f32c7fc2d8bf15af onto release 2.1.0

dmanc:
* removing blacklist wiring and check

* removing blacklist test

* fixing unit tests

* removing node initialization

---------

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/node/node.go
+++ b/node/node.go
@@ -190,20 +190,6 @@ func NewNode(
 		return nil, fmt.Errorf("failed to create new store: %w", err)
 	}
 
-	// Create new blacklist store
-	// We disable seeks compaction and enable sync writes to ensure that the blacklist is always up to date.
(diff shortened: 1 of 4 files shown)


node/node.go
```diff
@@ -268,7 +254,7 @@ func NewNode(
 		Metrics:                 metrics,
 		NodeApi:                 nodeApi,
 		Store:                   store,
-		BlacklistStore:          blacklistStore,
+		BlacklistStore:          nil,
```
pschork, line 257: Why not delete?

pschork (approved):

node/node.go
dmanc: We're still discussing if we want to fully remove the feature but for now we're moving forward just disabling the code path.

PR #1793, opened 2025-07-22
fix(proxy-release): make proxy release containing arm build

bxue-l2:

Eigenda-proxy is relied by Kurtosis DevNet, which is often run by developers on their local machine. 

but Kurtosis Config cannot specify the platform. So we have to prepare the ARM image.

If someone uses a MacBook with ARM chips, they will very likely be unable to pull the Docker image and break their testing. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -294,7 +294,7 @@ target "proxy" {
   target     = "proxy"
   # We push to layr-labs/ directly instead of layr-labs/eigenda/ for historical reasons,
   # since proxy was previously in its own repo: https://github.com/Layr-Labs/eigenda-proxy
-  tags       = ["${REGISTRY}/layr-labs/eigenda-proxy:${BUILD_TAG}"]
+  tags       = ["${REGISTRY}/layr-labs/eigenda-proxy-init:${BUILD_TAG}"]
 }
 
 target "proxy-internal" {
@@ -348,4 +348,9 @@ target "node-release" {
 target "nodeplugin-release" {
   inherits = ["nodeplugin", "_release"]
   tags     = ["${REGISTRY}/${REPO}/opr-nodeplugin:${BUILD_TAG}"]
+}
+
+target "proxy-release" {
+  inherits = ["proxy", "_release"]
+  tags     = ["${REGISTRY}/eigenda-proxy:${BUILD_TAG}"]
 }
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -352,5 +352,7 @@ target "nodeplugin-release" {
 
 target "proxy-release" {
   inherits = ["proxy", "_release"]
+  # We push to layr-labs/ directly instead of layr-labs/eigenda/ for historical reasons,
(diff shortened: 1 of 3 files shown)


docker-bake.hcl
```diff
@@ -294,7 +294,7 @@ target "proxy" {
   target     = "proxy"
   # We push to layr-labs/ directly instead of layr-labs/eigenda/ for historical reasons,
   # since proxy was previously in its own repo: https://github.com/Layr-Labs/eigenda-proxy
-  tags       = ["${REGISTRY}/layr-labs/eigenda-proxy:${BUILD_TAG}"]
+  tags       = ["${REGISTRY}/layr-labs/eigenda-proxy-init:${BUILD_TAG}"]
```
pschork: I don't think that registry path exists

bxue-l2: The reason I add "init" is to avoid collision of tags here, https://github.com/Layr-Labs/eigenda/pull/1793/files#diff-870f6fe23fc034f008f5203ee1e628d7bfa65a49095d5a4688db498328449ed2R353-R358

Should I just remove "-init"?

bxue-l2:
> You need to update Makefile `docker-release-build:` to call the new proxy-release target
fixed here
https://github.com/Layr-Labs/eigenda/pull/1793/commits/90a595ed927d2f85a7c1d8fdf8396d2e54e269d6

docker-bake.hcl
pschork: That target will never get pushed to GCR, so it doesn't matter. If you do a local docker build it will show up there.

docker-bake.hcl
```diff
@@ -73,7 +73,7 @@ group "ci-release" {
     "controller",
     "relay",
     "blobapi",
-    "proxy",
+    "proxy-release",
```
pschork: I think CI should call proxy

pschork (approved):

docker-bake.hcl
bxue-l2: I actually don't quite understand how ci-release used. 
- https://github.com/Layr-Labs/eigenda/blob/157240f079e86483676e84af8e1694b65d5b37cf/.github/workflows/docker-publish.yaml#L53
- https://github.com/Layr-Labs/eigenda/pkgs/container/eigenda%2Fdisperser

but will change back

bxue-l2: ff97d1721350d73e5e55e6cc13d663aab0f83d7a

docker-bake.hcl
```diff
@@ -294,7 +294,7 @@ target "proxy" {
   target     = "proxy"
   # We push to layr-labs/ directly instead of layr-labs/eigenda/ for historical reasons,
   # since proxy was previously in its own repo: https://github.com/Layr-Labs/eigenda-proxy
-  tags       = ["${REGISTRY}/layr-labs/eigenda-proxy:${BUILD_TAG}"]
+  tags       = ["${REGISTRY}/layr-labs/proxy:${BUILD_TAG}"]
```
samlaf, line 297: Why did you change this?

PR #1796, opened 2025-07-23
chore: cleanup docker targets in bake make ci

samlaf:

Also fixed some proxy CI issue: previously it was testing the latest tag which is already pushed to ghcr, not the one built locally in ci.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -7,8 +7,10 @@ variable "REPO" {
   default = "layr-labs/eigenda"
 }
 
+# We use the `dev` tag for local development builds.
+# CI builds will overwrite this with the `master` or `v*` tag.
 variable "BUILD_TAG" {
-  default = "latest"
+  default = "dev"
 }
 
 variable "SEMVER" {
@@ -47,7 +49,8 @@ group "all" {
     "traffic-generator-v2",
     "controller",
     "relay",
-    "blobapi"
+    "blobapi",
+    "proxy",
   ]
 }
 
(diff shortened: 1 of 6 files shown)


docker-bake.hcl
```diff
@@ -47,37 +49,18 @@ group "all" {
...
-    "dataapi",
-    "controller",
-    "relay",
-    "blobapi",
-    "proxy",
-  ]
-}
-
-# Internal devops builds
+# Internal devops builds. These targets are used by the eigenda-devops CI pipeline.
```
bxue-l2, line 61: Similarly, I don't think internal-release is used anywhere.

docker-bake.hcl
```diff
@@ -7,8 +7,10 @@ variable "REPO" {
   default = "layr-labs/eigenda"
 }
 
+# We use the `dev` tag for local development builds.
+# CI builds will overwrite this with the `master` or `v*` tag.
 variable "BUILD_TAG" {
-  default = "latest"
+  default = "dev"
```
bxue-l2, line 13: I tried to track down the latest tag, but I'm not familiar with it enough to be sure that this does not have bad downstream consequences.

Our devops repo uses the master tag, so I think that's fine.
Anyway, people shouldn't be used for anything serious. So I'll just let it be your judgment. But I wouldn't change it.

bxue-l2 (approved):

docker-bake.hcl
samlaf: Hmm indeed it seems like this group target is never used since our eigenda-devops ci uses matrix strategy to build each individually? https://github.com/Layr-Labs/eigenda-devops/blob/master/.github/workflows/build-docker.yml

Afaiu this should have worse performance since building a group together will share the buildx DAG components (say the shared builder docker step).

But not sure.. maybe there was a weird reason why we decided to go this way? cc @pschork @dmanc

docker-bake.hcl
samlaf: My reasoning for wanting to change it is twofold:
1. we shouldn't be afraid to change things. That's how you end up with tech debt and people get scared to touch anything and your release pipeline gets super long and scary. Of course theory needs to meet practice, so if there are good reasons why in this specific situation we shouldnt change it (for example we know itll have really bad downstream consequences that take a long tiem to fix), then I'm happy to revert.
2. `latest` is dangerous to use. If someone builds locally and by mistake pushes it'll push a latest tag. default tag imo should be something innocuous like `dev`, so that any mistakes won't affect important image tags.

@pschork @dmanc thoughts?

docker-bake.hcl
dmanc: That pipeline was built before the docker bake stuff was introduced so it can likely be refactored to build everything at once.

dmanc (approved):

docker-bake.hcl
samlaf: Makes sense, then I'll keep it here, and we can eventually refactor the devops pipeline.

PR #1800, opened 2025-07-24
chore(proxy): consolidate dotenv examples files into single one

samlaf:

Closes DAINT-649.

Having multiple example files is a source of bug and confusion. its also hard to maintain. Consolidating into a single .env.example file. People should figure out how to populate it for their own network anyways.

Also removed all the config values that have sane defaults, and only left the ones that are mandatory to set. This removes a lot of cognitive load on our users, and will be a helpful simplification I believe.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


.github/workflows/test-proxy.yml
```diff
@@ -50,7 +37,7 @@ jobs:
           experimental: true
           working_directory: api/proxy
       - name: Run flag test
-        run: ${{ github.workspace }}/api/proxy/scripts/test-proxy-startup-with-env-vars.sh ${{ matrix.env-file }}
+        run: ${{ github.workspace }}/api/proxy/scripts/test-proxy-startup-with-env-vars.sh .env.example
```
bxue-l2, line 40: Are you going to take the suggestion to not use the.env file?

api/proxy/.env.example
```diff
@@ -0,0 +1,90 @@
...
+# sent, decreasing this value is a crude sort of optimization.
+EIGENDA_PROXY_EIGENDA_V2_MAX_BLOB_LENGTH=1MiB
+
+# Address of the EigenDACertVerifierRouter contract, or of an immutable EigenDACertVerifier contract of version >= 3.
+# This is required for performing eth_calls to verify EigenDA certificates, as well as fetching
+# required_quorums and signature_thresholds needed when creating new EigenDA certificates during dispersals (POST routes).
+#
+# See https://layr-labs.github.io/eigenda/integration/spec/6-secure-integration.html#upgradable-quorums-and-thresholds-for-optimistic-verification
+# for more details.
+EIGENDA_PROXY_EIGENDA_V2_CERT_VERIFIER_ROUTER_OR_IMMUTABLE_VERIFIER_ADDR=0x61692e93b6B045c444e942A91EcD1527F23A3FB7
```
bxue-l2: think it is better to remove both mainnet and this address, but feel free to ignore

api/proxy/.env.example
```diff
@@ -0,0 +1,90 @@
...
+# This populates default values for other flags (disperser URL and contract addresses).
+# It is optional (those other flags can be set manually), but recommended.
+EIGENDA_PROXY_EIGENDA_V2_NETWORK=mainnet
+
+# Maximum blob length for V2. This value is permitted to be up to 16MiB
+# This value is configured to be lower than the permitted maximum, since this configuration determines the amount of
+# SRS data that will be read into memory. Reading in a large amount of SRS data can cause long startup times, and since
+# you only actually need to read the amount of SRS data that corresponds to the size of the largest blob that will be
+# sent, decreasing this value is a crude sort of optimization.
+EIGENDA_PROXY_EIGENDA_V2_MAX_BLOB_LENGTH=1MiB
```
bxue-l2, line 44: Since we have spent so much word here, maybe it's also good to add a caution to readers that this is different from the payload size because of padding and headers.

api/proxy/.env.example
```diff
@@ -0,0 +1,90 @@
...
+# required_quorums and signature_thresholds needed when creating new EigenDA certificates during dispersals (POST routes).
+#
+# See https://layr-labs.github.io/eigenda/integration/spec/6-secure-integration.html#upgradable-quorums-and-thresholds-for-optimistic-verification
+# for more details.
+EIGENDA_PROXY_EIGENDA_V2_CERT_VERIFIER_ROUTER_OR_IMMUTABLE_VERIFIER_ADDR=0x61692e93b6B045c444e942A91EcD1527F23A3FB7
+
+# Allowed distance (in L1 blocks) between the eigenDA cert's reference block number (RBN)
+# and the L1 block number at which the cert was included in the rollup's batch inbox.
+# A cert is considered valid if certL1InclusionBlock <= cert.RBN + rbnRecencyWindowSize,
+# otherwise an HTTP 418 (TEAPOT) error is returned to indicate that is should be dropped.
```
bxue-l2, line 57: For secure integration, for insecure integration, I don't think we have made a decision on which one is better. 
- stall or 
- skip
when encountering a derivation error. What is your vote on this, though this is a broader discussion.

api/proxy/.env.example
bxue-l2: And if you want to leave a default one, perhaps it is better to leave the one for Sepolia. 

Because for me now it's good to have a mental load.

bxue-l2 (dismissed):

api/proxy/.env.example
```diff
@@ -0,0 +1,90 @@
...
+EIGENDA_PROXY_STORAGE_DISPERSAL_BACKEND=V2
+
+# Caching stores to enable. Options are [redis, s3].
+# Make sure to populate their respective environment variables if you enable them.
+EIGENDA_PROXY_STORAGE_CACHE_TARGETS=""
+
+# === V2 Configuration ===
+
+# Hex-encoded signer private key for payments with EigenDA V2 disperser.
+EIGENDA_PROXY_EIGENDA_V2_SIGNER_PRIVATE_KEY_HEX="0000000000000000000100000000000000000000000000000000000000000000"
```
bxue-l2, line 27: Pretty surprised that this private key is allowed.

api/proxy/config/v2/eigendaflags/cli.go
```diff
@@ -190,14 +191,13 @@ loaded into memory for KZG commitments. Example units: '15MiB', '4Kib'.`,
...
+			Usage: fmt.Sprintf(`The EigenDA network that is being used. This is an optional flag, 
+to configure default values for different EigenDA contracts and disperser URL. 
+See https://github.com/Layr-Labs/eigenda/blob/master/api/proxy/common/eigenda_network.go
+for the exact values getting set by this flag. All of those values can also be manually
+set via their respective flags, and take precedence over the default values set by the network flag.
+If all of those other flags are manually configured, the network flag may be omitted. 
+Permitted EigenDANetwork values include %s, %s, %s, & %s.`,
 				common.MainnetEigenDANetwork,
 				common.HoleskyTestnetEigenDANetwork,
 				common.HoleskyPreprodEigenDANetwork,
```
litt3, line 203: Here's a thought:

right now, the `eigenda_network` is a bit hard to maintain. If you want to add a network type, you have to define the name constant, and then make sure you add the correct values to each switch statement in that file.

It would be a lot easier if we just defined an `EigenDANetwork` struct type, which contains all the necessary fields. Then all values for a given network would be defined together.

Additionally, if we maintained a list of all existing `EigenDANetwork` types in `eigenda_network.go`, we could use it for logs like this (join all the names with `, `), and then adding a new network type wouldn't entail modifying multiple files, it would all be contained in `eigenda_network.go`

This isn't necessarily something that needs to be done right now, or at all.

api/proxy/store/cli.go
```diff
@@ -96,12 +96,33 @@ func ReadConfig(ctx *cli.Context) (Config, error) {
 		return Config{}, fmt.Errorf("string to eigenDA backend: %w", err)
 	}
 
+	// We need to filter the cache targets and fallback targets to remove empty strings,
+	// since our code downstream doesn't work well with empty strings.
+	// Specifically, if the env var is simply set to nothing like `EIGENDA_PROXY_STORAGE_CACHE_TARGETS=`,
+	// it will result in an empty string being added to the slice
+	// for some reason... seems like a bug in urfave/cli?
```
litt3, line 103: If not a bug, at least an unpleasant user experience.

api/proxy/.env.example
litt3: Hey, how'd you get access to my cold storage key?!?!?

api/proxy/.env.example
```diff
@@ -0,0 +1,90 @@
+# This .env.example file only describes the mandatory flags that must be set when running the EigenDA proxy.
+# All flags that are optional and have sane default values are omitted for brevity.
+# For a full list of flags, please refer to the output of `eigenda-proxy --help`,
+# which is also available at https://github.com/Layr-Labs/eigenda/blob/master/api/proxy/docs/help_out.txt.
+
+# === Storage Configuration ===
+
+# The storage backends to enable. Options are [V1, V2]
+EIGENDA_PROXY_STORAGE_BACKENDS_TO_ENABLE=V1,V2
```
litt3, line 14: Consider adding additional guidance here:

The contents of this field dictate whether the structures to support the given backend are instantiated or not.

Reasons to include both `V1` *and* `V2` include:
1. You're dispersing to one, but still want to support reads from the other
2. You want to change which backend you are dispersing to on the fly, i.e. you're migrating from V1 to V2. So everything would be going through V1 at first, but the necessary structures would exist to switch to V2 dispersals, without requiring a restart

api/proxy/.env.example
```diff
@@ -0,0 +1,90 @@
...
+
+# Target EigenDA backend version (V1 or V2) for blob dispersals (POST routes).
+EIGENDA_PROXY_STORAGE_DISPERSAL_BACKEND=V2
```
litt3, line 18: consider noting that whichever you choose, it must also be part of the `EIGENDA_PROXY_STORAGE_BACKENDS_TO_ENABLE` list

api/proxy/.env.example
```diff
@@ -0,0 +1,90 @@
...
+# This value is configured to be lower than the permitted maximum, since this configuration determines the amount of
+# SRS data that will be read into memory. Reading in a large amount of SRS data can cause long startup times, and since
+# you only actually need to read the amount of SRS data that corresponds to the size of the largest blob that will be
+# sent, decreasing this value is a crude sort of optimization.
+EIGENDA_PROXY_EIGENDA_MAX_BLOB_LENGTH=1MiB
+
+# === Metrics Configuration ===
+
+# Enable the metrics server, which is disabled by default.
+EIGENDA_PROXY_METRICS_ENABLED=true
```
litt3: Depending on which merges first, you'll need to make sure to remove this, considering your PR to enable metrics by default

api/proxy/README.md
```diff
@@ -215,24 +215,24 @@ If you prefer a more controlled migration with explicit service updates, follow
...
 
 2. **Prepare V2 Configuration**
    - Prepare a configuration file that includes settings for both V1 and V2 backends
-      - See `.env.exampleV1AndV2.holesky` for an example configuration
+      - See `.env.example` for an example configuration
    - Set `EIGENDA_PROXY_STORAGE_DISPERSAL_BACKEND=V2`, so that the proxy started with this config will immediately
 enable V2 dispersal
 
-3. **Scheduled Migration**
+1. **Scheduled Migration**
```
litt3: improper numbering

litt3 (dismissed):

api/proxy/.env.example
samlaf: That's a good point. I actually don't like that we have so many words here, and feel like all the words should be in the flag documentation itself. But still not 100% sure, so was waiting for you guys' take.

api/proxy/.env.example
samlaf: what are you saying here? 
Are you saying that for insecure integration you would want to stall teh op-node's derivation pipeline if a cert fails the recency check?

api/proxy/.env.example
samlaf: I think the code is just parsing a hex value. So it doesn't allow having nothing, but any random data will do. It'll just fail at dispersal time since there's no money in the account (prob the accountant will catch this...?)

api/proxy/config/v2/eigendaflags/cli.go
samlaf: I 200% agree with you. Code that isn't meant to be dynamically dispatched should not be written with that flexibility, because it loses the static typing guarantees.
I do want to change this, but will do in a followup PR: https://linear.app/eigenlabs/issue/DAINT-662/make-eigenda-networks-statically-typed

api/proxy/.env.example
samlaf: merged that proxy in here by mistake.. so will resolve this.

.github/workflows/test-proxy.yml
samlaf: For now I would opt to keep it. Just because I think its simpler than pointing people to a random section of the README and have them copy-paste stuff. Also we can have it in CI like this.

api/proxy/.env.example
samlaf: [b91807f](https://github.com/Layr-Labs/eigenda/pull/1800/commits/b91807f2e8375cff57fcf8317d3284985e9ddfed)

api/proxy/.env.example
samlaf: [986a616](https://github.com/Layr-Labs/eigenda/pull/1800/commits/986a616ade5bf5d2fd80c02174ece37238e83fe8)

api/proxy/.env.example
samlaf: [49ef9e8](https://github.com/Layr-Labs/eigenda/pull/1800/commits/49ef9e86a99c12678662892e15f9e7b2c6219c04)

api/proxy/.env.example
samlaf: switched to sepolia values in [e97bf95](https://github.com/Layr-Labs/eigenda/pull/1800/commits/e97bf9513890fe14ba6e225206e47e2d48b32917).

I also think its slightly clearner without the sepolia values. But the thing is that I think its even better to make sure the file is up-to-date, and we can only do that by running it in CI, and removing those flags leads to 
image

api/proxy/README.md
samlaf: [b87b930](https://github.com/Layr-Labs/eigenda/pull/1800/commits/b87b930c8b246a8fb0b1a988526c76ff9fbcec84)

PR #1802, opened 2025-07-24
chore(proxy): crash on startup if certverifier != v3

samlaf:

Closes [DAINT-648](https://linear.app/eigenlabs/issue/DAINT-648/crash-early-if-cert-verifier-doesnt-contain-a-version-function).

Release 2.1.0 introduced CertVerifierV3+Router support, but didn't crash on startup when a CertVerifierV2 was passed in; it only resulted in a crash later on, or some very hard to debug eth-call errors.

This introduces defensive programming by only allowing proxy to start up if passed a certVerifierRouter or certVerifier =v3.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proxy/store/builder/storage_manager_builder.go
```diff
@@ -282,6 +282,21 @@ func buildEigenDAV2Backend(
...
 		return nil, fmt.Errorf("new cert verifier: %w", err)
 	}
+	if !isRouter {
+		// We call GetCertVersion to ensure that the cert verifier is at least version 3. See
+		// https://github.com/Layr-Labs/eigenda/blob/d0a14fa44/contracts/src/integrations/cert/interfaces/IVersionedEigenDACertVerifier.sol#L12
+		// https://github.com/Layr-Labs/eigenda/blob/d0a14fa44/contracts/src/integrations/cert/EigenDACertVerifier.sol#L79
+		// We pass in block 0 because a static certVerifierAddress provider is used when not using a router,
+		// so the block number is not relevant.
+		// TODO: we might want to only enforce this if the proxy is started in dispersal mode, to allow for
+		// proxy to be started in verification-only mode using a CertVerifierV2?
```
samlaf: actually I think this is safe because even the read path, when it gets a v2 cert, transforms it into a v3 and assumes that its only working with certV3 or above.

samlaf: verification-only mode also needed but orthogonal to do, so should be done in separate PR that shouldnt block this from merging.

bxue-l2: If it is not a router you want to check if you can get the cert version correctly. And if you can not get the cert version then you know it is a v2 cert verifier. 

but that basically just means you can not start a proxy with a way to cert verifier. Is that right?

litt3: Yeah, I think that's right.

If you are using this version of the proxy, you must have an up to date cert verifier

api/proxy/store/builder/storage_manager_builder.go
```diff
@@ -282,6 +282,21 @@ func buildEigenDAV2Backend(
...
+		_, err = certVerifier.GetCertVersion(ctx, 0)
+		if err != nil {
+			return nil, fmt.Errorf(
+				"%s doesn't appear to be a CertVerifier version >= V3: failed to call certVersion(): %w",
```
litt3: "doesn't appear to be" is pretty fluffy. Consider "is not"

litt3 (dismissed):

api/proxy/store/builder/storage_manager_builder.go
bxue-l2: But if you merge this PR, it basically can't read from v2 cert verifier, because this logic was run at the setup stage, we wouldn't get to the translation part.

@samlaf is my understanding correct that you still want people to use v2 cert verifier in the Read mode?

api/proxy/store/builder/storage_manager_builder.go
samlaf: issue is that it could also just be a normal network error. I updated the message to clarify this, let me know what you think: [7bd08e9](https://github.com/Layr-Labs/eigenda/pull/1802/commits/7bd08e9b1a6f0f1cd3a5c0de7f6ccbe7fff592e2)

api/proxy/store/builder/storage_manager_builder.go
samlaf: So it turns out I was wrong here. The release 2.1.0 of proxy already doesn't support CertVerifierV2 because of the way certs are serialized! It supports certV2s, by converting them to v3 and submitting them to a CertVerifierV3.

1. Updated the comment and added a check so that we only support V3 (not future versions, as we discussed last week when tallking about semver): [ef0527b](https://github.com/Layr-Labs/eigenda/pull/1802/commits/ef0527bea420e705eeffa8c6e5afa795e7d71205)
2. Created linear issue to support read-only mode: https://linear.app/eigenlabs/issue/DAINT-660/support-read-only-mode
3. updated the PR description
4. Updated release 2.1.0 msg to explain all of these subtleties and requirements better (PTAL and let me know if its clear enough)

image

bxue-l2: To be honest, after reading the text in the screenshot, I'm super confused. Why on earth we use "v2" for EigenDA upgrade (do we hate ourselves, and by this much)? does .env.exampleV2.mainnet for eigenda V2 or cert V2.

bxue-l2: Cool, so you intend to crash if v2 cert verifier is passed?

api/proxy/store/builder/storage_manager_builder.go
```diff
@@ -282,6 +283,28 @@ func buildEigenDAV2Backend(
...
+			return nil, fmt.Errorf(
+				"failed to eth-call certVersion(), meaning that you either have network problems with your eth node, or "+
+					"%s is not a CertVerifier version >= V3, which is required by this version of proxy: %w",
+				routerOrImmutableVerifierAddr.Hex(), err)
+		}
+		// Note that we also support certV2s, just not V2 CertVerifiers.
+		// This is because we transform certV2s into certV3s and verified using the CertVerifierV3 contract.
+		// However, the serialization logic, as well as some functions needed during the dispersal path (eg. requiredQuorums),
+		// are only compatible/available with CertVerifier V3, hence the requirement here.
+		if certVersion != 3 {
```
ethenotethan, line 303: quick glance but why not?
```
certVersion >= 3
```

samlaf: Because certVersion should be treated as a major semver version (patrick is working on changing to an actual semver value). Then version 4+ means the interface change, so the proxy code/bindings is VERY likely not going to be compatible with the new cert version interface.

semver basically gives us protection from future incompatible changes (as long as our future selves are good enough to notice when the breaking change happens and to bump the major version). If we do that then old versions of proxy will start crashing on startup and force their users to update to more recent version that supports v4 for example. This is best of all words imo.

PR #1803, opened 2025-07-24
fix: allow liveness checks on private IPs (ie preprod)

pschork:
Preprod operators are failing liveness checks because they are registered with private IPs which we currently block from scans.

When we originally added liveness checks, we were only doing a port scan and had concerns that liveness could be used to enumerate internal resources. Current liveness now does a full grpc reflection to validate ports map to the correct dispersal/retrieval service which will prevent any enumeration attacks.


dmanc (approved):

anupsv (approved):
Have alerted about potential issue with internal IP scanning.

PR #1804, opened 2025-07-25
feat(tools): new eigenda contract addresses reading cli

samlaf:

My endgoal target is to create a "discovery" CLI (hence the name) that given a network would discover all addresses (contracts, disperser/relays URLs, operator IPs, etc). This means it should probably incorporate the quorumscan tool, and others. I think ideally we would merge all of them into a single `eigenda` CLI with subcommands, which would be easier to use for everyone.

Right now it only discovers contract address by reading them from the EigenDADirectory address provided (or using the default one for the chain-id or network provided).
Given an eth-rpc it prints a table, json, or csv output of all contract names and their address.
image

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proxy/common/eigenda_network.go
```diff
@@ -107,13 +115,36 @@ func EigenDANetworksFromChainID(chainID string) ([]EigenDANetwork, error) {
...
+// In this case, we can use this function to automatically choose a default network.
+func DefaultEigenDANetworkFromChainID(chainID string) (EigenDANetwork, error) {
+	networks, err := EigenDANetworksFromChainID(chainID)
+	if err != nil {
+		return "", err
+	}
+	if len(networks) == 0 {
+		return "", fmt.Errorf("no EigenDA network found for chain ID: %s", chainID)
+	}
+	return networks[0], nil
```
litt3: I think it make more sense to only do this default selection iff `len(networks) == 1`.

In cases where there are multiple options, seems safer to make the user explicitly choose, instead of potentially choosing the wrong one

core/eth/eigenda_directory_reader.go
```diff
@@ -64,4 +63,23 @@ func (r *EigenDADirectoryReader) GetServiceManagerAddress() (gethcommon.Address,
 	return r.getAddressWithValidation(ContractNames.ServiceManager)
 }
 
+// GetAllAddresses returns all contract addresses from the directory in a map keyed by contract name
+func (r *EigenDADirectoryReader) GetAllAddresses() (map[string]gethcommon.Address, error) {
+	names, err := r.contract.GetAllNames(&bind.CallOpts{})
```
litt3: no timeout?

core/eth/eigenda_directory_reader.go
```diff
@@ -64,4 +63,23 @@ func (r *EigenDADirectoryReader) GetServiceManagerAddress() (gethcommon.Address,
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to get all contract names: %w", err)
```
litt3: Your error messages in this method include "failed to"

core/eth/eigenda_directory_reader.go
```diff
@@ -64,4 +63,23 @@ func (r *EigenDADirectoryReader) GetServiceManagerAddress() (gethcommon.Address,
...
+	}
+
+	addresses := make(map[string]gethcommon.Address)
+	for _, name := range names {
+		addr, err := r.contract.GetAddress0(&bind.CallOpts{}, name)
```
litt3: no timeout?

core/eth/eigenda_directory_reader.go
```diff
@@ -64,4 +63,23 @@ func (r *EigenDADirectoryReader) GetServiceManagerAddress() (gethcommon.Address,
...
+		if err != nil {
+			return nil, fmt.Errorf("failed to get %s address: %w", name, err)
+		}
+		addresses[name] = addr
+	}
+
+	return addresses, nil
+}
+
 // TODO: add other getters for other contracts; they are not needed for the current usage
```
litt3, line 85: I notice you don't attribute TODOs to yourself, e.g. `TODO(samlaf)`. Any reason? Curious your thoughts on the attribution- when I wrote the style guide I included it as a soft requirement, but we could roll that back if you disagree

tools/discovery/main.go
```diff
@@ -0,0 +1,211 @@
...
+
+const (
+	ethRpcUrlFlagName        = "eth-rpc-url"
+	NetworkFlagName          = "network"
+	directoryAddressFlagName = "directory-address"
+	outputFlagName           = "output"
+)
+
+// ContractInfo represents a contract name and its address
+type ContractInfo struct {
```
litt3: Unused struct?

tools/discovery/main.go
```diff
@@ -0,0 +1,211 @@
...
+	Name    string
+	Address string
+}
+
+func main() {
+	app := cli.NewApp()
+	if bi, ok := debug.ReadBuildInfo(); ok {
```
litt3: nit: `buildInfo` > `bi`, IMO.

tools/discovery/main.go
```diff
@@ -0,0 +1,211 @@
...
+	app.Flags = []cli.Flag{
+		&cli.StringFlag{
+			Name:     ethRpcUrlFlagName,
+			Usage:    "Ethereum RPC URL",
+			EnvVars:  []string{"ETH_RPC_URL"},
+			Required: true,
+		},
+		&cli.StringFlag{
+			Name: NetworkFlagName,
+			Usage: fmt.Sprintf(`The EigenDA network to discover (%s, %s, %s, %s).
```
litt3: Two things about this message:

1. Elsewhere, you used join when logging all the viable networks. I think that's better, since it removes the risk of getting out of sync
2. The way you have the parenthetical here doesn't make it explicit that the input value must match one of the listed networks. Consider stating explicitly, e.g. `must be one of ...`

tools/discovery/main.go
```diff
@@ -0,0 +1,211 @@
...
+			},
+		},
+		&cli.StringFlag{
+			Name:    directoryAddressFlagName,
+			Usage:   "EigenDADirectory contract address (overrides the default network address)",
+			EnvVars: []string{"EIGENDA_DIRECTORY_ADDRESS"},
+		},
+		&cli.StringFlag{
+			Name:    outputFlagName,
+			Usage:   "Output format: table, csv, json (default: table)",
```
litt3: You're duplicating `table, csv, json` here. Consider using join on the `validFormats` list instead

Though, maybe not possible with golang randomizing set order, since help string might not match the version we have committed?

tools/discovery/main.go
```diff
@@ -0,0 +1,211 @@
...
+		logger.Printf("No directory address provided, attempting to auto-detect.")
+		if networkName != "" {
+			logger.Printf("network %s provided, attempting to use default directory address.", networkName)
+			network, err := proxycmn.EigenDANetworkFromString(ctx.String(NetworkFlagName))
+			if err != nil {
+				return err
+			}
+			directoryAddr, err = network.GetEigenDADirectory()
+			if err != nil {
+				return fmt.Errorf("error getting EigenDADirectory address for network %s: %w", network, err)
```
litt3: Here too you're prefixing errors with words like "error getting". Intentional, or LLM not following your preferences?

tools/discovery/main.go
```diff
@@ -0,0 +1,211 @@
...
+	// 3. no directory address and no network: use the default network for the chainID
+	if directoryAddr == "" {
+		logger.Printf("No directory address provided, attempting to auto-detect.")
+		if networkName != "" {
+			logger.Printf("network %s provided, attempting to use default directory address.", networkName)
+			network, err := proxycmn.EigenDANetworkFromString(ctx.String(NetworkFlagName))
+			if err != nil {
+				return err
+			}
+			directoryAddr, err = network.GetEigenDADirectory()
```
litt3: I know the logger output is slightly different, but theoretically you could pull this code out of the `if else`, since both branches are doing essentially the same thing

```
			directoryAddr, err = network.GetEigenDADirectory()
			if err != nil {
				return fmt.Errorf("error getting EigenDADirectory address for network %s: %w", network, err)
			}
			logger.Printf("Auto-detected EigenDADirectory address %s for network %s", directoryAddr, networkName)
```

tools/discovery/main.go
```diff
@@ -0,0 +1,211 @@
...
+		return fmt.Errorf("invalid EigenDADirectory address: %s", directoryAddr)
+	}
+
+	// Use the directory reader from core/eth package
+	directoryReader, err := eth.NewEigenDADirectoryReader(directoryAddr, client)
+	if err != nil {
+		return fmt.Errorf("failed to create EigenDADirectory reader: %w", err)
+	}
+
+	// Get all addresses
```
litt3: not exactly a helpful comment on a method called `GetAllAddresses`

tools/discovery/main.go
```diff
@@ -0,0 +1,211 @@
...
+		printJSON(addressMap)
+	}
+
+	return nil
+}
+
+func printTable(addressMap map[string]gethcommon.Address) {
+	t := table.NewWriter()
+	t.SetOutputMirror(os.Stdout)
+	t.AppendHeader(table.Row{"Contract Name", "Address"})
```
litt3, line 155: `Contract Name` and `Address` are duplicated strings in this file

tools/discovery/main.go
```diff
@@ -0,0 +1,211 @@
...
+
+func printJSON(addressMap map[string]gethcommon.Address) {
+	fmt.Println("[")
+	i := 0
+	for name, addr := range addressMap {
+		comma := ","
+		if i == len(addressMap)-1 {
+			comma = ""
+		}
+		fmt.Printf("  {\"name\": \"%s\", \"address\": \"%s\"}%s\n", name, addr.Hex(), comma)
```
litt3: any reason to just use `name` here, instead of `contract_name`, as with the other print methods?

tools/discovery/main.go
```diff
@@ -0,0 +1,211 @@
...
+	proxycmn "github.com/Layr-Labs/eigenda/api/proxy/common"
+	"github.com/Layr-Labs/eigenda/core/eth"
+	gethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/ethereum/go-ethereum/ethclient"
+	"github.com/jedib0t/go-pretty/v6/table"
+	"github.com/urfave/cli/v2"
+)
+
+const (
+	ethRpcUrlFlagName        = "eth-rpc-url"
```
litt3: I'm not a fan of how we handle string literals when doing flag+env configurations.

All the flag names are specified as const strings at the top, but then the env var name string is defined far away, where the flag struct is initialized. Modifying a name is tedious at best, and error prone at worst (e.g., the flag name gets changed, but the env name is forgotten)

If we had a simple struct, named `FlagAndEnvField` (or something better...) which contains fields for `FlagName`, and `EnvVarName`, then the relevant names for a given field would be right next to each other, and it would be easier to keep them properly in sync.

May or may not make sense to start such a system in this PR. It's just a painpoint I've noticed before, that probably would be good to address at some point.

api/proxy/common/eigenda_network.go
```diff
@@ -107,13 +115,36 @@ func EigenDANetworksFromChainID(chainID string) ([]EigenDANetwork, error) {
...
 func EigenDANetworkFromString(inputString string) (EigenDANetwork, error) {
 	network := EigenDANetwork(inputString)
 
 	switch network {
 	case HoleskyTestnetEigenDANetwork, HoleskyPreprodEigenDANetwork, SepoliaTestnetEigenDANetwork, MainnetEigenDANetwork:
 		return network, nil
 	default:
-		return "", fmt.Errorf("unknown network type: %s", inputString)
+		allowedNetworks := []string{
+			MainnetEigenDANetwork.String(),
```
pschork, line 129: MainnetBeta missing

core/eth/eigenda_directory_reader.go
samlaf: What does that attribution imply? That I created it or that I should be the one to fix it? If the former, then git blame serves that purpose imo. If the latter, I think anyone should be allowed/encouraged to fix it.

tools/discovery/main.go
samlaf: [7e1b727](https://github.com/Layr-Labs/eigenda/pull/1804/commits/7e1b727fa40a268ec4f4503e92bb93ee5d2b0438)

tools/discovery/main.go
samlaf: [05a0704](https://github.com/Layr-Labs/eigenda/pull/1804/commits/05a0704799a4d430477cac815b80eee0d9b4d16a)

tools/discovery/main.go
samlaf: for 1 I don't think that's possible right now. I will fix it in another PR when I fix the network.go file with your suggestions which I totally agree to.
For 2: [e3bfc73](https://github.com/Layr-Labs/eigenda/pull/1804/commits/e3bfc732d3a4659be178297e8a70985143c9d1ef)

tools/discovery/main.go
samlaf: [2947baa](https://github.com/Layr-Labs/eigenda/pull/1804/commits/2947baac322bec140932df8f5848d501c65ac19b)

tools/discovery/main.go
samlaf: LLM. not sure I got all of them but fixed a bunch of them in [caa60fb](https://github.com/Layr-Labs/eigenda/pull/1804/commits/caa60fb7743f31cd15567c629348e9ce785056b0)

tools/discovery/main.go
samlaf: [2679ab9](https://github.com/Layr-Labs/eigenda/pull/1804/commits/2679ab9ea0964b992d10960d1b14ef9526b5d5f1)

tools/discovery/main.go
samlaf: since the json is different, it's only duplicated twice. personally I think that's totally acceptable and DRY should not be followed at all costs.

tools/discovery/main.go
samlaf: [095a0a5](https://github.com/Layr-Labs/eigenda/pull/1804/commits/095a0a5cff59cfb1e06b60ef6b7a5209e853f68f)

tools/discovery/main.go
samlaf: [8b998f3](https://github.com/Layr-Labs/eigenda/pull/1804/commits/8b998f335bf2d3dae6a13489a67afe5da4e34d1d)

api/proxy/common/eigenda_network.go
samlaf: Good point. Will add in a future PR if you don't mind, since I already need to refactor how networks are used as per austin's comment. Created https://linear.app/eigenlabs/issue/DAINT-662/make-eigenda-networks-statically-typed

core/eth/eigenda_directory_reader.go
samlaf: [176305a](https://github.com/Layr-Labs/eigenda/pull/1804/commits/176305ada4555e06ffa87810901b984bc65b9345)

core/eth/eigenda_directory_reader.go
samlaf: oh man you uncovered the beast. Was just following hope's code for the other functions in here... :(
Changed in [16fb1ea](https://github.com/Layr-Labs/eigenda/pull/1804/commits/16fb1ea1449cbda4f5b624b511419aae2daa2fc9) but our whole codebase is just not setup to inject context properly... left 2 TODOs elsewhere in the code. Not planning on fixing our entire codebase in this tools PR unfortunately lol :'(

core/eth/eigenda_directory_reader.go
samlaf: https://github.com/Layr-Labs/eigenda/pull/1804#discussion_r2237782365

tools/discovery/main.go
samlaf: That's interesting. I was actually about to go and implement this because I agree with the idea, but then realized only the flag name is ever actually needed elsewhere (when parsing the flag into a var).

I don't think it makes a whole lot of sense anymore to pick the env var out of its normal place of definition. Because if we do that, then might as well ask: why not also move the usage string into the field `FlagAndEnvAndUsageFields`, and then what about the other fields?

I could be convinced otherwise but it doesn't feel like a clear win to me atm.

api/proxy/common/eigenda_network.go
samlaf: Totally agree. That function only belongs in the tool, if anywhere. I deleted it.
[9504291](https://github.com/Layr-Labs/eigenda/pull/1804/commits/950429178e2f66d9b37f12d07c154db69320c94f)

core/eth/eigenda_directory_reader.go
litt3: I personally use the attribution to indicate two things:

1. who created the TODO
2. this TODO is actually expected to merge into the repo, and isn't a temporary "fix before merging" TODO

I suppose git blame helps with # 1 to a certain extent. But frequently lines/files get moved, or slight changes are made to the line (e.g. updating a name used in the comment) which change the git blame to someone else, making it difficult to discern the true source of the comment

Regarding # 2: whether or not we decide that the *attribution* should be what performs this function, it seems like an important distinction to have. When I'm reviewing a PR, I need to know whether the author simply forgot to finish something, or if they want to merge the PR with TODO comment intact. Taking this PR as an example: how should I know whether you want to implement these getters now, or if it's something you want to be done in the future? With nothing to indicate, I'm stuck needing to comment, *just in case* this is something you meant to fix and forgot about.

tools/discovery/main.go
litt3: Good point. What do you think of this idea:

```
var (
	EthereumRPCFlag  = cli.StringFlag{
		Name:     "eth-rpc-url",
		Usage:    "Ethereum RPC URL",
		EnvVars:  []string{"ETH_RPC_URL"},
		Required: true,
	}
)
```

Define the all the flag structs in a `var` block at the top of the file, and then the `main` function would simply add all of the defined flag vars to the `app.Flags` list.

Then, anyone who needs access to the flag name could just use `EthereumRPCFlag.Name`

tools/discovery/main.go
```diff
@@ -108,43 +107,45 @@ func discoverAddresses(ctx *cli.Context) error {
...
-			network, err := proxycmn.DefaultEigenDANetworkFromChainID(chainID.String())
+			networks, err := proxycmn.EigenDANetworksFromChainID(chainID.String())
 			if err != nil {
-				return fmt.Errorf("error determining EigenDA network from chain ID %s: %w", chainID, err)
+				return fmt.Errorf("DefaultEigenDANetworkFromChainID (chainid: %s): %w", chainID, err)
 			}
-			directoryAddr, err = network.GetEigenDADirectory()
-			if err != nil {
-				return fmt.Errorf("error getting EigenDADirectory address for network %s: %w", network, err)
+			network = networks[0]
```
litt3: Probably should make sure len is >0 before indexing in like this (regardless of whether `EigenDANetworksFromChainID` would *actually* return an empty list with a nil error)

tools/discovery/main.go
```diff
@@ -0,0 +1,211 @@
...
+			chainID, err := client.ChainID(ctx.Context)
+			if err != nil {
+				return fmt.Errorf("chainID call: %w", err)
+			}
+			if chainID == nil {
+				return fmt.Errorf("received nil chainID from Ethereum client")
+			}
+			networks, err := proxycmn.EigenDANetworksFromChainID(chainID.String())
+			if err != nil {
+				return fmt.Errorf("DefaultEigenDANetworkFromChainID (chainid: %s): %w", chainID, err)
```
litt3: error no longer matches method name

tools/discovery/main.go
```diff
@@ -0,0 +1,211 @@
...
+			}
+			network = networks[0]
+			if len(networks) > 1 {
+				logger.Printf("Multiple EigenDA networks found for chain ID %s: %v. Using the first one: %s", chainID, networks, network)
```
litt3: You claimed to agree with my comment in [this thread](https://github.com/Layr-Labs/eigenda/pull/1804#discussion_r2237800256)

But you're still simply choosing the first network in cases where there are multiple possible networks. I think we should just force users to explicitly choose the correct network in cases where there are multiple options, and throw an error here if they don't.

"Taking a guess" at the network a user wants doesn't seem helpful... what if we guess wrong? Maybe they'll see this log, but maybe not

litt3 (dismissed):

tools/discovery/main.go
samlaf: completely genius idea... we should do this everywhere! [dca5fac](https://github.com/Layr-Labs/eigenda/pull/1804/commits/dca5facd63e06fcb5681e074a89d22a2df04fa2f)

tools/discovery/main.go
samlaf: This I think is not necessary for 2 reasons:
1. extra code that imo is not needed becasue,
2. if function contract is broken and some refactor introduces a bug where it returns an empty list, then the program will crash with a stack trace, whic is exactly what we need to debug this. User can take the stack trace, and copy-paste it to us. IMO this is more useful than giving a little concise error (wrapped golang errors are basically hand-constructed stack traces)
3. its just a cli tool so fine if a bug crashes without a structured error anyways

tools/discovery/main.go
samlaf: updated in [541422f](https://github.com/Layr-Labs/eigenda/pull/1804/commits/541422f3102a8f5612436c2bb73240ebd243f1e2), but this code actually doesn't even exist anymore because of the other refactor in https://github.com/Layr-Labs/eigenda/pull/1804#discussion_r2240918936. Resolving thread.

tools/discovery/main.go
samlaf: I was opting for having a simpler CLI, without forcing user to pass too many required flags before getting an answer. But I agree that choosing network is pretty fundamental. Refactored in [b882d9c](https://github.com/Layr-Labs/eigenda/pull/1804/commits/b882d9c443627de794235e20d354c545ba51ba9f) by making network flag required. This also greatly simplified the code. Good change!

litt3 (approved):

PR #1805, opened 2025-07-25
test: proxy wrapper

cody-littley:

Augments traffic generator and live test framework to be able to use the proxy as a mediator of the EigenDA clients.


test/v2/client/test_client.go
```diff
@@ -334,6 +341,61 @@ func NewTestClient(
...
+				SignerPaymentKey: privateKey,
+				EthRPCURL:        ethRPCUrls[0],
+			},
+			ServerConfig: proxyserver.Config{
+				Host:        "localhost",
+				Port:        config.ProxyPort,
+				EnabledAPIs: []string{"admin"},
+			},
+			MetricsServerConfig: proxymetrics.Config{
+				Enabled: false, // TODO enable this
```
litt3: TODO

test/v2/client/test_client.go
```diff
@@ -570,20 +636,77 @@ func (c *TestClient) DisperseAndVerify(ctx context.Context, payload []byte) erro
...
+	_, err = c.ReadBlobWithProxy(ctx, cert, payload, 0)
+	if err != nil {
+		return fmt.Errorf("failed to read blob with proxy: %w", err)
+	}
+
+	return nil
+}
+
 // DispersePayload sends a payload to the disperser. Returns the blob key.
 func (c *TestClient) DispersePayload(ctx context.Context, payloadBytes []byte) (coretypes.EigenDACert, error) {
```
litt3: for `DispersePayloadWithProxy` you use a named return value `err`, consider using the same pattern in both places

test/v2/client/test_client.go
```diff
@@ -570,20 +636,77 @@ func (c *TestClient) DisperseAndVerify(ctx context.Context, payload []byte) erro
...
 	start := time.Now()
+	c.metrics.startOperation("dispersal")
+
+	// Important: don't redefine err. It's used by the deferred function to report success or failure.
+	var err error
+	defer func() {
+		c.metrics.endOperation("dispersal")
+		if err == nil {
+			c.metrics.reportDispersalSuccess()
+			c.metrics.reportDispersalTime(time.Since(start))
```
litt3, line 666: in `DispersePayloadWithProxy`, you report the time even in the failure case. Any reason for different strategy here?

test/v2/client/test_client.go
```diff
@@ -716,3 +868,56 @@ func (c *TestClient) ReadBlobFromValidators(
 
 	return nil
 }
+
+// ReadBlobWithProxy reads a blob from the proxy wrapper and compares it to the expected payload bytes.
+// The timeout is ignored if zero. If the proxy wrapper is not enabled, this method returns an error.
+func (c *TestClient) ReadBlobWithProxy(ctx context.Context,
+	cert []byte,
+	expectedPayloadBytes []byte,
+	timeout time.Duration,
```
litt3, line 876: There is currently a `RetrievalTimeout` defined in `ValidatorPayloadRetrieverConfig`

We probably should move that field into `PayloadClientConfig`, since it's a valid concepts for all types of retrieval (in a future PR, not in this PR)

cc @samlaf any thoughts here?

test/v2/client/test_client.go
```diff
@@ -716,3 +868,56 @@ func (c *TestClient) ReadBlobFromValidators(
...
+			c.metrics.reportProxyReadTime(time.Since(start))
+		} else {
+			c.metrics.reportProxyReadFailure()
+		}
+	}()
+
+	var data []byte
+	data, err = c.proxyWrapper.GetPayload(ctx, cert)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read blob from proxy: %w", err)
```
litt3: The method that failed here deals in `payloads`, not `blobs`

test/v2/client/test_client.go
```diff
@@ -716,3 +868,56 @@ func (c *TestClient) ReadBlobFromValidators(
...
+	}
+
+	if !bytes.Equal(data, expectedPayloadBytes) {
+		return nil, fmt.Errorf("read payload does not match expected payload")
+	}
+
+	c.metrics.reportProxyReadTime(time.Since(start))
```
litt3: You're double reporting this metric.

test/v2/client/test_client_config.go
```diff
@@ -69,6 +70,25 @@ type TestClientConfig struct {
 	ValidatorReadConnectionPoolSize int
 	// The size of the thread pool for CPU heavy operations.
 	ValidatorReadComputePoolSize int
+	// The port to use for the proxy.
+	ProxyPort int
+}
+
+// TODO create default config
```
litt3: todo

test/v2/config/environment/broken/preprod.json
```diff
@@ -13,10 +12,5 @@
   "EigenDACertVerifierAddressQuorums0_1_2": "0x2ccB62f3349DD04FEeff189C7f0568193eFd4430",
   "EigenDACertVerifierAddressQuorums2":     "0xBC23d765540b77803910E65F62BF01a9F8CaB30A",
   "SubgraphURL": "https://subgraph.satsuma-prod.com/51caed8fa9cb/eigenlabs/eigenda-operator-state-preprod-holesky/version/v0.7.0/api",
-  "SRSOrder": 268435456,
-  "MaxBlobSize": 16777216,
-  "MetricsPort": 9101,
-  "DisableMetrics": true,
-  "ValidatorReadConnectionPoolSize": 20,
-  "ValidatorReadComputePoolSize": 20
+  "DisableMetrics": true
```
litt3, line 15: Do you actually still want this disabled here?

test/v2/load/load_generator.go
```diff
@@ -236,39 +236,105 @@ func (l *LoadGenerator) disperseBlob(rand *random.TestRandom) (
...
+	err = l.doReadsWithProxy(rand, cert, payload)
+	<-l.relayReadLimiter
+	if err != nil {
+		l.client.GetLogger().Errorf("failed to read blob from proxy: %w", err)
+	}
+}
+
+// Disperses a blob using the proxy (as opposed to using the GRPC clients directly). Returns the blob cert in byte
+// form since this is how the proxy forces the user to interact with it.
+func (l *LoadGenerator) disperseBlobWithProxy(rand *random.TestRandom) (
```
litt3: Given this method accepts a `payload`, not a `blob`, it should probably be called `dispersePayloadWithProxy`

test/v2/live/proxy_test.go
```diff
@@ -0,0 +1,118 @@
...
+	"github.com/Layr-Labs/eigenda/common/testutils/random"
+	"github.com/Layr-Labs/eigenda/encoding"
+	"github.com/Layr-Labs/eigenda/encoding/utils/codec"
+	"github.com/Layr-Labs/eigenda/test/v2/client"
+	"github.com/docker/go-units"
+	"github.com/stretchr/testify/require"
+)
+
+// Disperse an empty payload. Blob will not be empty, since payload encoding entails adding bytes
+func emptyPayloadProxyDispersalTest(t *testing.T, environment string) {
```
litt3, line 15: In the `test_client`, there were previously two separate pathways you were testing: "underlying clients", and payload clients.

With the addition of these proxy tests, you're essentially just adding a third pathway. It seems a bit odd that the first two pathways are combined, but this new pathway is in entirely separate methods. 

Is there any reason you decided to implement this separately as opposed to integrating the proxy methods into the pre-existing code?

test/v2/client/test_client.go
cody-littley: Converted to `// TODO (cody.littley) enable proxy metrics`. This is non-trivial, since the metrics servers will clash with each other. It's not urgent to have these metrics right now, so let's leave it to a future PR.

test/v2/client/test_client.go
cody-littley: done

test/v2/client/test_client.go
cody-littley: Only reporting time on success is the correct behavior, I've made `DispersePayloadWithProxy()` do the same.

test/v2/client/test_client.go
cody-littley: fixed

test/v2/client/test_client.go
cody-littley: fixed

test/v2/client/test_client_config.go
cody-littley: task was already completed, removed comment

test/v2/config/environment/broken/preprod.json
cody-littley: Yes. These json files are not used in environments where we collect metrics. The config files we use for the traffic generator are defined in the devops repo.

test/v2/load/load_generator.go
cody-littley: fixed

test/v2/live/proxy_test.go
cody-littley: The other tests validate things that are tricky to validate with the proxy. For example, ensuring that the blob is available from each relay (if there is more than one relay ID), forcing a read to go to validators instead of going to the relays, validating that the data in the certificate is what we expect to see, etc. Since the proxy hides a lot of those details, we can't do the same types of validation.

PR #1806, opened 2025-07-25
chore(proxy): deprecate fallback reads in favor of cache reads

samlaf:
Closes DAINT-245.

Fallback read behavior (read from cache target only after trying to read from EigenDA) was afaiu not used by anyone. Opting to clean up the code and keep the configuration simpler.

I don't currently see a reason why a user wouldn't want to read from the cache before EigenDA, since if the data is there, it will clearly be read faster and reduce latency and global network costs.

Reading from EigenDA validators should be the last bastion since it puts a lot of stress on the network. It thus makes more sense to have a layer of caches (including reading from the eigenda relays) in front of the validators, not after them.

> One of my most productive days was throwing away 207 lines of code
> -- Ken Thompson

### DIFF
--- a/api/proxy/store/secondary/secondary.go
+++ b/api/proxy/store/secondary/secondary.go
@@ -91,15 +91,12 @@ func (sm *SecondaryManager) WriteOnCacheMissEnabled() bool {
 	return sm.CachingEnabled() && sm.writeOnCacheMiss
 }
 
-// HandleRedundantWrites ... writes to both enabled caches
-// and returns an error if NONE of them succeed
+// Writes to all enabled caches and returns an error if NONE of them succeed
 func (sm *SecondaryManager) HandleRedundantWrites(ctx context.Context, commitment []byte, value []byte) error {
-	sources := sm.caches
-
 	key := crypto.Keccak256(commitment)
 	successes := 0
 
-	for _, src := range sources {
+	for _, src := range sm.caches {
 		sm.log.Debug("Attempting to write to secondary storage", "backend", src.BackendType())
 		cb := sm.m.RecordSecondaryRequest(src.BackendType().String(), http.MethodPut)
 


api/proxy/store/secondary/secondary.go
```diff
@@ -87,27 +83,18 @@ func (sm *SecondaryManager) Topic() chan<- PutNotify {
...
-func (sm *SecondaryManager) FallbackEnabled() bool {
-	return len(sm.fallbacks) > 0
-}
-
 func (sm *SecondaryManager) WriteOnCacheMissEnabled() bool {
 	return sm.CachingEnabled() && sm.writeOnCacheMiss
 }
 
-// HandleRedundantWrites ... writes to both sets of backends (i.e, fallback, cache)
+// HandleRedundantWrites ... writes to both enabled caches
```
litt3: I think you need to remove the word "both" from this comment, this method operates on a range of sources, not two in particular

api/proxy/store/secondary/secondary.go
```diff
@@ -87,27 +83,18 @@ func (sm *SecondaryManager) Topic() chan<- PutNotify {
...
 // and returns an error if NONE of them succeed
 func (sm *SecondaryManager) HandleRedundantWrites(ctx context.Context, commitment []byte, value []byte) error {
 	sources := sm.caches
```
litt3: You can just use `sm.caches` directly, no need to create `sources`

litt3 (dismissed):

api/proxy/store/secondary/secondary.go
samlaf: [4f9354c](https://github.com/Layr-Labs/eigenda/pull/1806/commits/4f9354cc2241300afcf28d9ac9b063f5821cf07f)

api/proxy/store/secondary/secondary.go
samlaf: [4f9354c](https://github.com/Layr-Labs/eigenda/pull/1806/commits/4f9354cc2241300afcf28d9ac9b063f5821cf07f)

PR #1808, opened 2025-07-28
feat: adds an option to start pprof when running LittDB CLI.

cody-littley:

Adds the ability to start the golang profiler when running with the LittDB CLI.

### DIFF
--- a/litt/cli/litt_cli.go
+++ b/litt/cli/litt_cli.go
@@ -2,8 +2,10 @@ package main
 
 import (
 	"bufio"
+	"fmt"
 	"os"
 
+	"github.com/Layr-Labs/eigenda/common/pprof"
 	"github.com/Layr-Labs/eigensdk-go/logging"
 	"github.com/urfave/cli/v2"
 )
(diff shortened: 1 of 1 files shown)


litt/cli/litt_cli.go
```diff
@@ -259,25 +272,50 @@ func buildCLIParser(logger logging.Logger) *cli.App {
...
+	logger.Infof("Waiting for debugger to attach (pid: %d).\n", pid)
+
+	logger.Infof("Press Enter to continue...")
+	reader := bufio.NewReader(os.Stdin)
+	_, _ = reader.ReadString('\n') // block until newline is read
+}
+
+// If pprof is enabled, this function starts the pprof server.
+func handlePProfMode(ctx *cli.Context, logger logging.Logger) error {
+	pprofEnabled := ctx.Bool("pprof")
```
litt3, line 306: I don't love that we have duplicate string literals like this.

I proposed an idea on [this PR](https://github.com/Layr-Labs/eigenda/pull/1804#discussion_r2240882895), which would have us define these cli flags in a var block, which would also solve the duplication problem here.

Curious your thoughts on this.

litt/cli/litt_cli.go
```diff
@@ -259,25 +272,50 @@ func buildCLIParser(logger logging.Logger) *cli.App {
...
+	if !pprofEnabled {
 		return nil
 	}
+
+	pprofPort := ctx.Int("pprof-port")
+	if pprofPort <= 0 {
```
litt3: Any reason to check the lower bound here but not the upper?

litt3 (dismissed):

litt/cli/litt_cli.go
cody-littley: As discussed offline, I will make this change in a follow up PR

litt/cli/litt_cli.go
cody-littley: added upper bound check

PR #1809, opened 2025-07-28
feat: add ls command to LittDB CLI

cody-littley:

Adds the `litt ls` command.

To manually test:

1. generate a test DB via `litt benchmark ./path/to/config.json`. (There is a good example config at `litt/benchmark/config/basic-config.json`.), let it run for a few seconds
2. run `litt ls --src /path/to/benchmark/volume1`, expected output is `benchmark` (i.e. the table where the benchmark stores its data).


litt/cli/ls.go
```diff
@@ -0,0 +1,120 @@
...
+	"github.com/urfave/cli/v2"
+)
+
+func lsCommand(ctx *cli.Context) error {
+	logger, err := common.NewLogger(common.DefaultConsoleLoggerConfig())
+	if err != nil {
+		return fmt.Errorf("failed to create logger: %v", err)
+	}
+
+	sources := ctx.StringSlice("src")
```
litt3, line 24: I left a similar comment on a different PR:

It's unfortunate how there are meaningful string literals scattered about that are expected to match.

I'm wondering if there's some standard strategy we could put in place, which gets rid of this sort of thing. My proposal is to define the CLI commands / flags in a `var` block inside the `cli.go` file, and if you need specific strings like this, you'd get them by directly accessing the member fields of the cli structs.

litt/cli/ls.go
```diff
@@ -0,0 +1,120 @@
...
+		sb.WriteString("\n")
+	}
+
+	logger.Infof("Tables found:\n%s", sb.String())
+
+	return nil
+}
+
+// Similar to ls, but searches for tables in multiple paths.
+func lsPaths(logger logging.Logger, rootPaths []string, lock bool, fsync bool) ([]string, error) {
```
litt3, line 53: is there a reason you need the `lock` and `fsync` params? The only real usage of this function hardcodes these to `true`

If these params are removed, you would also be able to remove them from `ls`

litt/cli/ls.go
```diff
@@ -0,0 +1,120 @@
...
+	tableSet := make(map[string]struct{})
+
+	for _, rootPath := range rootPaths {
+		tables, err := ls(logger, rootPath, lock, fsync)
+		if err != nil {
+			return nil, fmt.Errorf("error finding tables: %v", err)
```
litt3: %v

litt/cli/ls.go
```diff
@@ -0,0 +1,120 @@
...
+// Returns a list of LittDB tables at the specified LittDB path. Tables are alphabetically sorted by their names.
+// Returns an error if the path does not exist or if no tables are found.
+func ls(logger logging.Logger, rootPath string, lock bool, fsync bool) ([]string, error) {
+
+	if lock {
+		// Forbid touching tables in active use.
+		lockPath := path.Join(rootPath, util.LockfileName)
+		fLock, err := util.NewFileLock(logger, lockPath, fsync)
+		if err != nil {
+			return nil, fmt.Errorf("failed to acquire lock on %s: %v", rootPath, err)
```
litt3: %v

litt/cli/ls.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+		defer fLock.Release()
+	}
+
+	// LittDB has one directory under the root directory per table, with the name
+	// of the table being the name of the directory.
+	possibleTables, err := os.ReadDir(rootPath)
+	if err != nil {
+		return nil, fmt.Errorf("failed to read dir %s: %v", rootPath, err)
```
litt3: %v

litt/cli/ls.go
```diff
@@ -0,0 +1,120 @@
...
+	tables := make([]string, 0, len(possibleTables))
+	for _, entry := range possibleTables {
+		if !entry.IsDir() {
+			continue
+		}
+
+		segmentPath := filepath.Join(rootPath, entry.Name(), segment.SegmentDirectory)
+		exists, err := util.Exists(segmentPath)
+		if err != nil {
+			return nil, fmt.Errorf("failed to check if segment path %s exists: %v", segmentPath, err)
```
litt3: %v

litt/cli/ls.go
```diff
@@ -0,0 +1,120 @@
...
+		}
+		if exists && entry.IsDir() {
```
litt3: why check `entry.IsDir()` again?

litt/cli/ls.go
cody-littley: That makes sense. I've created a linear ticket to track this action item, since doing it with all the code currently in flight would cause a merge nightmare. https://linear.app/eigenlabs/issue/EGDA-1794/eliminate-magic-strings-in-littdb-cli

litt/cli/ls.go
cody-littley: Although it's not visible in this PR, `lsPaths()` is used by other LittDB CLI utilities which haven't been merged yet. Since those utilities hold the lock when they call this function, they need a way to make this function not acquire a lock when it is called.

The `fsync` flag is actually quite important to have for unit tests. Doing fsync operations slows the code down significantly. In some unit tests, using fsync can be the difference between a 1 second test and a 30 second test. As a result, I've made all LittDB code have fsync operations as being capable of being skipped during unit tests. This is a continuation of that pattern.

Even though unit tests can skip fsync, I usually make it so that the option is not exposed in production configuration. There are no circumstances when a production workflow should be running with fsync mode disabled.

litt/cli/ls.go
cody-littley: fixed

litt/cli/ls.go
cody-littley: fixed

litt/cli/ls.go
cody-littley: fixed

litt/cli/ls.go
cody-littley: fixed

litt/cli/ls.go
cody-littley: This should have been checking if `segmentDir` is a directory. I've refactored this to fix the problem.

litt3 (approved):

PR #1810, opened 2025-07-28
feat(proxy-memstore): add memstore feat that allows put to inject derivation error for its get

bxue-l2:

Adds ability to configure memstore to inject specific DerivationError responses during GET operations while allowing PUT operations to succeed normally. This enables testing client handling of derivation errors without requiring complex setup.

It is driven by our need to test equivalence between (OP node+proxy) and (hokulea+kona) in the failure mode. 

Specifically, users send a PATCH request that sets the desired derivation error for all subsequent GET requests. After that, when the user sends data to the proxy, the PUT operation succeeds as normal—the error injection only affects the GET path. Behind the scenes, upon a GET request, the proxy returns either the stored data or the specified derivation error depending on its configuration.
The PATCH request is sticky, meaning it will take effect on multiple GET requests unless reset.

### update
We also added the functionality in HTTP requests to reset the derivation error. To do it, we add an additional parameter in the HTTP request and call the struct NullableDerivationError, see this [commit](https://github.com/Layr-Labs/eigenda/pull/1810/commits/ca0ebfdc9ec390b1566d83e56526b9c5be850a91). For usage, see the readMe file in this PR.

Key Changes

- Modified DerivationError.Validate() to return errors instead of panicking
- Added PutWithGetReturnsDerivationError configuration option to memstore
- Extended EphemeralDB to store and return configured errors
- Added comprehensive tests verifying the functionality

Usage

Can be enabled via the memstore config API by setting a specific status code (1-4) and disabled by setting to null.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/coretypes/derivation_errors.go
+++ b/api/clients/v2/coretypes/derivation_errors.go
@@ -31,10 +31,7 @@ func (e DerivationError) Error() string {
 // We panic if marshalling fails, since the caller won't be able to handle the derivation error
 // properly, so they'll receive a 500 and must retry.
 func (e DerivationError) MarshalToTeapotBody() string {
-	err := e.Validate()
-	if err != nil {
-		panic(err)
-	}
+	e.Validate()
 	bodyJSON, err := json.Marshal(e)
 	if err != nil {
 		panic(fmt.Errorf("failed to marshal derivation error: %w", err))
@@ -53,13 +50,12 @@ func (e DerivationError) WithMessage(msg string) DerivationError {
 
 // Validate that the DerivationError has a valid status code.
 // The only valid status codes are 1-4, as defined in the sentinel errors below, eg [ErrCertParsingFailedDerivationError].
-func (e DerivationError) Validate() error {
+func (e DerivationError) Validate() {
 	if e.StatusCode < 1 || e.StatusCode > 4 {
-		return fmt.Errorf("DerivationError: invalid status code %d, must be between 1 and 4", e.StatusCode)
+		panic(fmt.Errorf("DerivationError: invalid status code %d, must be between 1 and 4", e.StatusCode))
 	}
 	// The Msg field should ideally be a human-readable string that explains the error,
 	// but we don't enforce it.
-	return nil
 }
 
 // These errors can be used as sentinels to indicate specific derivation errors,
--- a/api/clients/v2/coretypes/derivation_errors.go
+++ b/api/clients/v2/coretypes/derivation_errors.go
@@ -31,7 +31,10 @@ func (e DerivationError) Error() string {
 // We panic if marshalling fails, since the caller won't be able to handle the derivation error
 // properly, so they'll receive a 500 and must retry.
 func (e DerivationError) MarshalToTeapotBody() string {
-	e.Validate()
+	err := e.Validate()
+	if err != nil {
+		panic(err)
+	}
 	bodyJSON, err := json.Marshal(e)
 	if err != nil {
 		panic(fmt.Errorf("failed to marshal derivation error: %w", err))
@@ -50,12 +53,13 @@ func (e DerivationError) WithMessage(msg string) DerivationError {
 
 // Validate that the DerivationError has a valid status code.
 // The only valid status codes are 1-4, as defined in the sentinel errors below, eg [ErrCertParsingFailedDerivationError].
-func (e DerivationError) Validate() {
+func (e DerivationError) Validate() error {
 	if e.StatusCode < 1 || e.StatusCode > 4 {
-		panic(fmt.Sprintf("DerivationError: invalid status code %d, must be between 1 and 4", e.StatusCode))
+		return fmt.Errorf("DerivationError: invalid status code %d, must be between 1 and 4", e.StatusCode)
 	}
 	// The Msg field should ideally be a human-readable string that explains the error,
 	// but we don't enforce it.
+	return nil
 }
 
 // These errors can be used as sentinels to indicate specific derivation errors,
--- a/api/proxy/clients/memconfig_client/client.go
+++ b/api/proxy/clients/memconfig_client/client.go
@@ -19,10 +19,10 @@ type Config struct {
 	URL string // EigenDA proxy REST API URL
 }
 
-// this is a copy derivation error to avoid cyclic deps
+// This is a copy derivation error to avoid cyclic deps
 // see implementation at
 // https://github.com/Layr-Labs/eigenda/blob/e5f489aae34a1f68eb750e0da7ded52c200d7c36/api/clients/v2/coretypes/derivation_errors.go#L20
-// for all possible status code, see
+// for all possible status codes, see
 // https://github.com/Layr-Labs/eigenda/blob/66834223356d2ed230a8ffbbba13c6bb36d04139/api/clients/v2/coretypes/derivation_errors.go#L73
 type DerivationError struct {
 	StatusCode uint8
--- a/api/proxy/clients/memconfig_client/client.go
+++ b/api/proxy/clients/memconfig_client/client.go
@@ -37,7 +37,7 @@ type MemConfig struct {
 	PutLatency                       time.Duration
 	GetLatency                       time.Duration
 	PutReturnsFailoverError          bool
-	PutWithGetReturnsDerivationError DerivationError
+	PutWithGetReturnsDerivationError *DerivationError
 }
 
 // MarshalJSON implements custom JSON marshaling for Config.
@@ -62,7 +62,7 @@ type intermediaryCfg struct {
 	PutLatency                       string
 	GetLatency                       string
 	PutReturnsFailoverError          bool
-	PutWithGetReturnsDerivationError DerivationError
+	PutWithGetReturnsDerivationError *DerivationError
 }
 
 // IntoMemConfig ... converts an intermediary config into a memconfig
@@ -148,6 +148,7 @@ func (c *Client) GetConfig(ctx context.Context) (*MemConfig, error) {
 // as a POST and modifies every associated field. This could present issues if
 // misused in a testing framework which imports it.
 func (c *Client) UpdateConfig(ctx context.Context, update *MemConfig) (*MemConfig, error) {
+	fmt.Printf("update %v\n", update)
 	body, err := update.MarshalJSON()
 	if err != nil {
 		return nil, fmt.Errorf("failed to marshal config update to json bytes: %w", err)
@@ -163,6 +164,8 @@ func (c *Client) UpdateConfig(ctx context.Context, update *MemConfig) (*MemConfi
 		return nil, fmt.Errorf("failed to do request: %w", err)
 	}
 	defer resp.Body.Close()
+
+	fmt.Printf("resp.Status %v\n", resp.Header)
 
 	if resp.StatusCode != http.StatusOK {
 		return nil, fmt.Errorf("failed to update config, status code: %d", resp.StatusCode)
--- a/api/proxy/clients/memconfig_client/client.go
+++ b/api/proxy/clients/memconfig_client/client.go
@@ -19,9 +19,11 @@ type Config struct {
 	URL string // EigenDA proxy REST API URL
 }
 
-// copy derivation error to avoid cyclic deps
-// see full implementation at
+// this is a copy derivation error to avoid cyclic deps
+// see implementation at
 // https://github.com/Layr-Labs/eigenda/blob/e5f489aae34a1f68eb750e0da7ded52c200d7c36/api/clients/v2/coretypes/derivation_errors.go#L20
+// for all possible status code, see
+// https://github.com/Layr-Labs/eigenda/blob/66834223356d2ed230a8ffbbba13c6bb36d04139/api/clients/v2/coretypes/derivation_errors.go#L73
 type DerivationError struct {
 	StatusCode uint8
 	Msg        string
--- a/api/proxy/clients/memconfig_client/client.go
+++ b/api/proxy/clients/memconfig_client/client.go
@@ -148,7 +148,6 @@ func (c *Client) GetConfig(ctx context.Context) (*MemConfig, error) {
 // as a POST and modifies every associated field. This could present issues if
 // misused in a testing framework which imports it.
 func (c *Client) UpdateConfig(ctx context.Context, update *MemConfig) (*MemConfig, error) {
-	fmt.Printf("update %v\n", update)
 	body, err := update.MarshalJSON()
 	if err != nil {
 		return nil, fmt.Errorf("failed to marshal config update to json bytes: %w", err)
@@ -164,8 +163,6 @@ func (c *Client) UpdateConfig(ctx context.Context, update *MemConfig) (*MemConfi
 		return nil, fmt.Errorf("failed to do request: %w", err)
 	}
 	defer resp.Body.Close()
-
-	fmt.Printf("resp.Status %v\n", resp.Header)
 
 	if resp.StatusCode != http.StatusOK {
 		return nil, fmt.Errorf("failed to update config, status code: %d", resp.StatusCode)
--- a/api/proxy/clients/memconfig_client/client.go
+++ b/api/proxy/clients/memconfig_client/client.go
@@ -19,16 +19,25 @@ type Config struct {
 	URL string // EigenDA proxy REST API URL
 }
 
+// copy derivation error to avoid cyclic deps
+// see full implementation at
+// https://github.com/Layr-Labs/eigenda/blob/e5f489aae34a1f68eb750e0da7ded52c200d7c36/api/clients/v2/coretypes/derivation_errors.go#L20
+type DerivationError struct {
+	StatusCode uint8
+	Msg        string
+}
+
 // MemConfig ... contains properties that are used to configure the MemStore's behavior.
 // this is copied directly from /store/generated_key/memstore/memconfig.
 // importing the struct isn't possible since it'd create cyclic dependency loop
 // with core proxy's go.mod
 type MemConfig struct {
-	MaxBlobSizeBytes        uint64
-	BlobExpiration          time.Duration
-	PutLatency              time.Duration
-	GetLatency              time.Duration
-	PutReturnsFailoverError bool
+	MaxBlobSizeBytes                 uint64
+	BlobExpiration                   time.Duration
+	PutLatency                       time.Duration
+	GetLatency                       time.Duration
+	PutReturnsFailoverError          bool
+	PutWithGetReturnsDerivationError DerivationError
 }
 
 // MarshalJSON implements custom JSON marshaling for Config.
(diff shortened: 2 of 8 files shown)


api/clients/v2/coretypes/derivation_errors.go
```diff
@@ -31,7 +31,10 @@ func (e DerivationError) Error() string {
 // We panic if marshalling fails, since the caller won't be able to handle the derivation error
 // properly, so they'll receive a 500 and must retry.
 func (e DerivationError) MarshalToTeapotBody() string {
-	e.Validate()
+	err := e.Validate()
+	if err != nil {
```
bxue-l2: refactor the old Validate function.

api/proxy/store/generated_key/memstore/README.md
```diff
@@ -48,6 +48,17 @@ $ curl http://localhost:3100/memstore/config | \
   curl -X PATCH http://localhost:3100/memstore/config -d @-
 ```
 
+#### PUT with GET returning derivation error
+The configuration allows users to configure memstore to inject specific derivation error responses during GET operations while allowing PUT operations to succeed normally. This enables testing client handling of derivation errors without requiring complex setup.
```
ethenotethan: nit: 
```suggestion
The configuration allows users to configure memstore to return specific derivation error responses during `/get` payload retrievals while still allowing `/put` request operations to succeed normally with the payload persisted to ephemeral db. This enables fast iteration testing of a rolllup client's handling of derivation errors without requiring a complex setup.
```

api/proxy/store/generated_key/memstore/README.md
```diff
@@ -48,6 +48,17 @@ $ curl http://localhost:3100/memstore/config | \
...
+Specifically, users send a PATCH request that sets the desired derivation error for all subsequent GET requests. After that, when the user sends data to the proxy, the PUT operation succeeds as normal—the error injection only affects the GET path. Behind the scenes, upon a GET request, the proxy returns either the stored data or the specified derivation error depending on its configuration.
```
ethenotethan: hmm are there any unintended side effects with marking all `/get` requests (global) as a specific derivation error vs having it as a localized mutation per ephemeral db key => value entry. e.g, we disperse a payload through proxy and pre-mark it with a specific `DerivationError` vs having a global config. in the wild this error will come up  with some inbox sequence like:
```
[valid_cert, invalid_cert, valid_cert]
```

api/proxy/store/generated_key/memstore/ephemeraldb/ephemeral_db.go
```diff
@@ -68,14 +75,17 @@ func (db *DB) InsertEntry(key []byte, value []byte) error {
 
 	strKey := string(key)
 
+	derivationError := db.config.GetPUTWithGetReturnsDerivationError()
+
+	// disallow any overwrite
```
ethenotethan, line 79: ah I see this is unique to each entry - nvm plz ignore the aforementioned comment

api/proxy/store/generated_key/memstore/README.md
ethenotethan: nvm - just saw the ephemeral db changes and realized its unique to each ephemeral entry (cert, blob). Can we make this more explicit in the docs?

api/proxy/clients/memconfig_client/client.go
```diff
@@ -19,39 +19,52 @@ type Config struct {
 	URL string // EigenDA proxy REST API URL
 }
 
+// this is a copy derivation error to avoid cyclic deps
+// see implementation at
+// https://github.com/Layr-Labs/eigenda/blob/e5f489aae34a1f68eb750e0da7ded52c200d7c36/api/clients/v2/coretypes/derivation_errors.go#L20
+// for all possible status code, see
+// https://github.com/Layr-Labs/eigenda/blob/66834223356d2ed230a8ffbbba13c6bb36d04139/api/clients/v2/coretypes/derivation_errors.go#L73
```
iquidus: nit:

```suggestion
// This is a copy derivation error to avoid cyclic deps
// see implementation at
// https://github.com/Layr-Labs/eigenda/blob/e5f489aae34a1f68eb750e0da7ded52c200d7c36/api/clients/v2/coretypes/derivation_errors.go#L20
// for all possible status codes, see
// https://github.com/Layr-Labs/eigenda/blob/66834223356d2ed230a8ffbbba13c6bb36d04139/api/clients/v2/coretypes/derivation_errors.go#L73
```

api/proxy/store/generated_key/memstore/README.md
bxue-l2: https://github.com/Layr-Labs/eigenda/pull/1810/commits/76f8a0d769f31bcc76ee54a6e7fc00c4f4616f64

bxue-l2:
## update
We also added the functionality in HTTP requests to reset the derivation error. To do it, we add an additional parameter in the HTTP request and call the struct NullableDerivationError, see this [commit](https://github.com/Layr-Labs/eigenda/pull/1810/commits/ca0ebfdc9ec390b1566d83e56526b9c5be850a91). For usage, see the readMe file in this PR.

api/clients/v2/coretypes/derivation_errors.go
```diff
@@ -50,12 +53,13 @@ func (e DerivationError) WithMessage(msg string) DerivationError {
 
 // Validate that the DerivationError has a valid status code.
 // The only valid status codes are 1-4, as defined in the sentinel errors below, eg [ErrCertParsingFailedDerivationError].
-func (e DerivationError) Validate() {
+func (e DerivationError) Validate() error {
 	if e.StatusCode < 1 || e.StatusCode > 4 {
-		panic(fmt.Sprintf("DerivationError: invalid status code %d, must be between 1 and 4", e.StatusCode))
+		return fmt.Errorf("DerivationError: invalid status code %d, must be between 1 and 4", e.StatusCode)
```
samlaf: why not keep the panic? I personally think its cleaner to not return errors for bugs like this.

Overall we should also just refactor our code such that invalid DerivationErrors are impossible to create in the first place (by only making valid constructors Public)

api/proxy/clients/memconfig_client/client.go
```diff
@@ -19,39 +19,52 @@ type Config struct {
 	URL string // EigenDA proxy REST API URL
 }
 
+// This is a copy derivation error to avoid cyclic deps
+// see implementation at
+// https://github.com/Layr-Labs/eigenda/blob/e5f489aae34a1f68eb750e0da7ded52c200d7c36/api/clients/v2/coretypes/derivation_errors.go#L20
+// for all possible status codes, see
+// https://github.com/Layr-Labs/eigenda/blob/66834223356d2ed230a8ffbbba13c6bb36d04139/api/clients/v2/coretypes/derivation_errors.go#L73
+type DerivationError struct {
```
samlaf, line 27: what's the cyclic dependency? We should try to fix it. why does coretype import memconfig_client??

Also instead of adding github links, you should add golang links, like [coretypes.DerivationError]

api/proxy/clients/memconfig_client/client.go
```diff
@@ -19,39 +19,52 @@ type Config struct {
...
-	BlobExpiration          time.Duration
-	PutLatency              time.Duration
-	GetLatency              time.Duration
-	PutReturnsFailoverError bool
+	MaxBlobSizeBytes                 uint64
+	BlobExpiration                   time.Duration
+	PutLatency                       time.Duration
+	GetLatency                       time.Duration
+	PutReturnsFailoverError          bool
+	PutWithGetReturnsDerivationError *DerivationError
```
samlaf: This name reads very weird. I had to look at the README documentation to understand what it's even trying to say.
Suggest changing to just `GetReturnsDerivationError`. Why does "PutWith" need to be included?

api/proxy/store/generated_key/memstore/README.md
```diff
@@ -48,6 +48,35 @@ $ curl http://localhost:3100/memstore/config | \
   curl -X PATCH http://localhost:3100/memstore/config -d @-
 ```
 
+#### PUT with GET returning derivation error
+The configuration allows users to configure memstore to return specific derivation error responses during `/get` payload retrievals while still allowing `/put` request operations to succeed normally with the payload persisted to ephemeral db. This enables fast iteration testing of a rollup client's handling of derivation errors without requiring a complex setup. The error is applied to individual PUT requests in the ephemeral db.
```
samlaf:
```suggestion
The configuration allows users to configure memstore to return specific derivation error responses during GET payload retrievals while still allowing POST request operations to succeed normally with the payload persisted to ephemeral db. This enables fast iteration testing of a rollup client's handling of derivation errors without requiring a complex setup. The error is applied to individual POST requests in the ephemeral db.
```

samlaf: I personally think that for consistency we should be using the REST verbs and not the random resource prefixes that op confusingly added

api/proxy/clients/memconfig_client/client.go
samlaf: Oh wait I see the semantic is that you want previously written blobs to be returned without errors, but only blobs POSTed while this was turned on to return a derivation error?

Is this complex semantic really needed? Can we instead just make all GET requests return a derivation error when this is turned on?

If not, then how about changing name to something like `PostedBlobsReturnGetDerivationError`? Also a bit awkward... otherwise please at least add a comment here to explain the complex semantics.

api/proxy/store/generated_key/memstore/README.md
```diff
@@ -48,6 +48,35 @@ $ curl http://localhost:3100/memstore/config | \
...
+
+##### Setting a derivation error
+Configure memstore to return a specific derivation error for all subsequent `/get` requests:
+
+```bash
+curl -X PATCH http://localhost:3100/memstore/config \
+  -d '{"PutWithGetReturnsDerivationError": {"StatusCode": 3, "Msg": "Invalid cert", "Reset": false}}'
+```
+
+This will cause all future `/get` requests to return an HTTP 418 error with the specified derivation error, while `/put` requests continue to succeed normally.
```
samlaf: its not all future get requests right? its just future GET requests from POSTs that are posted while this is turned on?

We could name the error `PutBlobsContainDerivationError` maybe?

api/proxy/store/generated_key/memstore/README.md
```diff
@@ -48,6 +48,35 @@ $ curl http://localhost:3100/memstore/config | \
   curl -X PATCH http://localhost:3100/memstore/config -d @-
 ```
 
+#### PUT with GET returning derivation error
+The configuration allows users to configure memstore to return specific derivation error responses during `/get` payload retrievals while still allowing `/put` request operations to succeed normally with the payload persisted to ephemeral db. This enables fast iteration testing of a rollup client's handling of derivation errors without requiring a complex setup. The error is applied to individual PUT requests in the ephemeral db.
+
+The `PutWithGetReturnsDerivationError` field supports three states:
+1. **Field omitted**: No change to current configuration
```
samlaf: this is the behavior of PATCH, not of this specific field...
If you PUT instead of PATCH then if the field is omitted it should mean to reset.

api/proxy/store/generated_key/memstore/ephemeraldb/ephemeral_db.go
```diff
@@ -18,6 +18,13 @@ const (
 	DefaultPruneInterval = 500 * time.Millisecond
 )
 
+// a wrapper around payload with status code
+// if instructed mode is not enabled, the status code is 1
+type payloadWithDerivationError struct {
+	payload         []byte
+	derivationError error // the underlying type is coretypes.DerivationError
```
samlaf:
```suggestion
	derivationError error // the underlying type is [coretypes.DerivationError]
```

api/proxy/store/generated_key/memstore/ephemeraldb/ephemeral_db.go
```diff
@@ -18,6 +18,13 @@ const (
 	DefaultPruneInterval = 500 * time.Millisecond
 )
 
+// a wrapper around payload with status code
+// if instructed mode is not enabled, the status code is 1
+type payloadWithDerivationError struct {
```
samlaf: I don't really like "instructed mode". Personally find it confusing. Also does this mean statusCode=1 is equivalent to derivationError=nil?

api/proxy/store/generated_key/memstore/memconfig/config.go
```diff
@@ -118,6 +128,39 @@ func (sc *SafeConfig) SetMaxBlobSizeBytes(maxBlobSizeBytes uint64) {
 	sc.config.MaxBlobSizeBytes = maxBlobSizeBytes
 }
 
+func (sc *SafeConfig) GetPUTWithGetReturnsDerivationError() error {
```
samlaf:
```suggestion
func (sc *SafeConfig) PutWithGetReturnsDerivationError() error {
```

samlaf: be consistent with the other GETters

api/proxy/store/generated_key/memstore/memconfig/config.go
```diff
@@ -17,6 +21,10 @@ type Config struct {
 	// after sleeping PutLatency duration.
 	// This can be used to simulate eigenda being down.
 	PutReturnsFailoverError bool
+	// if nil, any subsequent put requests will succeed without error, and all gets will succeed without error
```
samlaf: Is this second part true? What about values that were input with a non-nil DerivationError? WOuldnt they return a derivation error?

api/proxy/store/generated_key/memstore/README.md
samlaf: Very much agree with @ethenotethan here. I find the semantics of that config very hard to understand. Not even sure I fully grasp them yet (left a bunch of comments).

@ethenotethan 's suggestion is not bad actually... should we consider instead adding a query_param to POST requests which are only read from the memstore? Something like
```
POST /put/0x...?derivation_error_status=2
```

Feel like this would be a lot easier to understand, and also face a lot less race issues.

api/proxy/clients/memconfig_client/client.go
bxue-l2: > GET requests return a derivation error when this is turned on?

That is not great, because if you are not careful, it is possible for two GET getting two different answers. I think it is exactly what @ethenotethan is against

One property the current impl achieves is GETs will always return the same value

api/proxy/store/generated_key/memstore/README.md
bxue-l2: The reason I chose this approach is to avoid congesting the routing logics, just for sack of doing testing. 

One thing I don't want to see is that in the normal mode, we make it possible for users to inject some derivation error that shoot themself in the foot

api/clients/v2/coretypes/derivation_errors.go
bxue-l2: I make a panic at the call site, to me, this is some logic function that can be shared. Some might want to panic, some do not.

In my case, we want to intentionally inject an error to the proxy by setting the status code. Then there are some valid and invalid status code, so that's why I want to separate this function out.

bxue-l2: check if panic! kill process or server thread

api/proxy/store/generated_key/memstore/memconfig/config.go
bxue-l2: fix it

api/proxy/store/generated_key/memstore/README.md
bxue-l2: Talked offline. It's possible to pass the parameter in, but there are some changes to propagate the parameter into the memstore. Overall, it doesn't seem to worth it to increase the code complexity for memstore.

api/clients/v2/coretypes/derivation_errors.go
bxue-l2: will revert. panic only kills the thread.

api/proxy/clients/memconfig_client/client.go
bxue-l2: The Optimism repo needs to import this client? In that case, Optimism will import Eigen DA. But I think Eigen DA also imports Optimism for tests. And that's a cycle.

In order for [coretypes.DerivationError], I think you have to import the package.

api/proxy/store/generated_key/memstore/ephemeraldb/ephemeral_db.go
bxue-l2: This is old comment, deleted it.

api/proxy/clients/memconfig_client/client.go
bxue-l2: No! All puts should succeed. Some put is configured with its get returning derivation error. Some others don't.

The configuration is done through the HTTP patch method.

api/proxy/store/generated_key/memstore/README.md
bxue-l2: I'm going to use NullableDerivationError for http PATCH params.

Within the config, I am going to use "OverwritePutWithDerivationError". Challenge me

api/proxy/store/generated_key/memstore/README.md
bxue-l2: Yes, I fixed it

api/proxy/store/generated_key/memstore/memconfig/config.go
bxue-l2: But it is now called OverwritePutWithDerivationError

api/clients/v2/coretypes/derivation_errors.go
bxue-l2: https://github.com/Layr-Labs/eigenda/pull/1810/commits/54a7f3fbfc8eff766fa590a3d3ed12528f12bc7a#diff-f74932e44ef41a829b92fe9944907ee262fc1bff15fe242b8fbc8a62567876a2R55

api/proxy/clients/memconfig_client/client.go
bxue-l2: https://github.com/Layr-Labs/eigenda/pull/1810/commits/ec4881e1df0f7e950b2beb14e6130672d04c74f4#diff-c075f80fe1d7e1727d9a842484f023a9c6a2d7495c0d98153cb184de79c7d679R24

api/proxy/clients/memconfig_client/client.go
bxue-l2: new name https://github.com/Layr-Labs/eigenda/pull/1810/commits/54a7f3fbfc8eff766fa590a3d3ed12528f12bc7a#diff-554add61612dbffb3585e41634e6f540195eb1065f09e3193b04870104e6e3baR53

api/proxy/store/generated_key/memstore/README.md
bxue-l2: I basically rewrote and renamed, see 
https://github.com/Layr-Labs/eigenda/pull/1810/commits/54a7f3fbfc8eff766fa590a3d3ed12528f12bc7a

api/proxy/store/generated_key/memstore/README.md
bxue-l2: rewriten, https://github.com/Layr-Labs/eigenda/pull/1810/commits/54a7f3fbfc8eff766fa590a3d3ed12528f12bc7a

api/proxy/store/generated_key/memstore/memconfig/config.go
bxue-l2: https://github.com/Layr-Labs/eigenda/pull/1810/commits/54a7f3fbfc8eff766fa590a3d3ed12528f12bc7a#diff-4c4d2db49a55c172cff0eb2daa1dd6dab9d1bc2afdfe9c51b6aaf7f049c210e5R24

samlaf (approved):

PR #1811, opened 2025-07-28
feat: max connection age

cody-littley:

For both the relays and API servers, enforce a maximum connection age. The goal of this change is to ensure that clients aren't stuck talking to a small subset of relays/API servers after the number of those servers has been increased.

I will not merge this PR until it has been tested on preprod.

### DIFF
--- a/disperser/server_config.go
+++ b/disperser/server_config.go
@@ -17,7 +17,7 @@ type ServerConfig struct {
 	// connections based on age.
 	MaxConnectionAge time.Duration
 
-	// When the server closes a connection due to MaxConnectionAgeSeconds, it will wait for this grace period before
+	// When the server closes a connection due to MaxConnectionAge, it will wait for this grace period before
 	// forcibly closing the connection. This allows in-flight requests to complete.
 	MaxConnectionAgeGrace time.Duration
 


litt3:
@claude use the nitpicker subagent to review this PR. Once you have the result from the nitpicker subagent, please edit your comment to contain the human-readable output received from the subagent *verbatim*.

litt3 (approved):
LGTM

disperser/cmd/apiserver/flags/flags.go
```diff
@@ -46,11 +46,33 @@ var (
...
+		Required: false,
+		EnvVar:   common.PrefixEnvVar(envVarPrefix, "MAX_IDLE_CONNECTION_AGE_SECONDS"),
+		Value:    time.Minute,
+	}
 	BlsOperatorStateRetrieverFlag = cli.StringFlag{
 		Name:     common.PrefixFlag(FlagPrefix, "bls-operator-state-retriever"),
 		Usage:    "[Deprecated: use EigenDADirectory instead] Address of the BLS operator state Retriever",
 		Required: false,
-		EnvVar:   common.PrefixEnvVar(envVarPrefix, "BLS_OPERATOR_STATE_RETRIVER"),
+		EnvVar:   common.PrefixEnvVar(envVarPrefix, "BLS_OPERATOR_STATE_RETRIVER"), // sigh
```
dmanc, line 75: fyi this will break our API server unless we make the relevant changes for the next release. Not sure it's worth fixing the typo since we can just remove the flag in the future instead.

dmanc: nvm misread this diff

dmanc (approved):

PR #1813, opened 2025-07-29
feat(proxy): new GET query param to return EncodedPayload

samlaf:
Keeping as draft until I implement tests, but would like to get a first round of review in just to make sure the approach is sound.

Closes DAINT-569.

image

This is needed for secure integrations that need to receive the encodedPayload instead of the decoded payload, and do decoding themselves inside the fpvm, so that they can prove when a payload was wrongly encoded, and discard it from their derivation pipeline.

TODO: add tests

## Notes

This feature adds quite a bit of tech debt that I would like to eventually get rid of, but given that @bxue-l2 absolutely needs this for hokulea and the secure integration audit happening soon, I think we should just ship it. Some of the tech debt that I'd like to clean up:
1. get rid of redis (don't think anyone is using it)
2. figure out if we can store blobs in S3 (only secondary storage we should support imo); this would allow for getting rid of the cache/fallback flags and only define an ordered storage array (where s3, relays, and validators are all treated equal and can be fetched in any order).
3. use coretypes in proxy to prevent mistakes (right now our GET functions return []byte that can either be a payload or an encoded payload... very bug prone)
4. prob more stuff I can't think of right now


api/proxy/server/handlers_cert.go
```diff
@@ -133,6 +143,18 @@ func parseCommitmentInclusionL1BlockNumQueryParam(r *http.Request) (uint64, erro
...
+// Happy path:
+//   - if the return_encoded_payload query parameter is present (with any value), it returns true
+//   - this means it accepts ?return_encoded_payload, ?return_encoded_payload=true, ?return_encoded_payload=anything
+//
+// Unhappy paths:
+//   - if the return_encoded_payload query parameter is not provided, it returns false
+func parseReturnEncodedPayloadQueryParam(r *http.Request) bool {
+	_, exists := r.URL.Query()["return_encoded_payload"]
+	return exists
+}
```
samlaf: TODO: test this. I don't know if it actually behaves the way I think it does (aka can we actually just pass `?return_encoded_payload` without a value?)

api/clients/v2/payloadretrieval/relay_payload_retriever.go
```diff
@@ -65,7 +65,44 @@ func NewRelayPayloadRetriever(
...
+	encodedPayload, err := pr.GetEncodedPayload(ctx, eigenDACert)
+	if err != nil {
+		return nil, err
+	}
+
+	payload, err := encodedPayload.Decode()
+	if err != nil {
+		blobKey, keyErr := eigenDACert.ComputeBlobKey()
+		if keyErr != nil {
+			return nil, fmt.Errorf("compute blob key from eigenDACert: %w", keyErr)
```
litt3: this should also include the original error that caused us to enter this block

api/clients/v2/payloadretrieval/relay_payload_retriever.go
```diff
@@ -65,7 +65,44 @@ func NewRelayPayloadRetriever(
...
+		}
+
+		pr.log.Error(
```
litt3: I think I might have been the one to originally do it... but do we actually need this log here? Shouldn't the returned error be enough?

api/clients/v2/payloadretrieval/relay_payload_retriever.go
```diff
@@ -127,21 +161,18 @@ func (pr *RelayPayloadRetriever) GetPayload(
...
+				"blobKey", blobKey.Hex(), "relayKey", relayKey, "error", err)
+			continue
 		}
 
-		return payload, nil
+		return encodedPayload, nil
 	}
 
-	return nil, fmt.Errorf("unable to retrieve blob %v from any relay. relay count: %d", blobKey.Hex(), relayKeyCount)
+	return nil, fmt.Errorf("unable to retrieve encoded payload with blobkey %v from any relay. relay count: %d", blobKey.Hex(), relayKeyCount)
```
litt3: nit: elsewhere we use "blob key" or "blobKey", as opposed to "blobkey"

api/clients/v2/payloadretrieval/validator_payload_retriever.go
```diff
@@ -50,11 +50,49 @@ func NewValidatorPayloadRetriever(
...
+	encodedPayload, err := pr.GetEncodedPayload(ctx, eigenDACert)
+	if err != nil {
+		return nil, err
+	}
+
+	payload, err := encodedPayload.Decode()
+	if err != nil {
+		blobKey, keyErr := eigenDACert.ComputeBlobKey()
+		if keyErr != nil {
+			return nil, fmt.Errorf("compute blob key from eigenDACert: %w", keyErr)
```
litt3: this should include original `err` that caused us to enter this block

api/clients/v2/payloadretrieval/validator_payload_retriever.go
```diff
@@ -50,11 +50,49 @@ func NewValidatorPayloadRetriever(
...
+		}
+		pr.logger.Error(
```
litt3: Same question: do we actually need to log this?

api/clients/v2/payloadretrieval/validator_payload_retriever.go
```diff
@@ -99,20 +137,18 @@ func (pr *ValidatorPayloadRetriever) GetPayload(
...
+				"blobKey", blobKey.Hex(), "quorumID", quorumID, "error", err)
+			continue
 		}
 
-		return payload, nil
+		return encodedPayload, nil
 	}
 
-	return nil, fmt.Errorf("unable to retrieve payload from quorums %v", blobHeader.QuorumNumbers)
+	return nil, fmt.Errorf("unable to retrieve encoded payload with blobkey %v from quorums %v", blobKey.Hex(), blobHeader.QuorumNumbers)
```
litt3: same comment regarding "blobkey"

api/clients/v2/payload_retriever.go
```diff
@@ -11,9 +11,19 @@ import (
 //
 // This interface may be implemented to provide alternate retrieval methods, for example payload retrieval from an S3
 // bucket instead of from EigenDA relays or nodes.
+//
+// TODO(samlaf): I don't think we need this interface. We probably shouldn't have the separate relay
```
litt3, line 15: :+1:

api/proxy/common/store.go
```diff
@@ -65,7 +65,7 @@ func StringToBackendType(s string) BackendType {
 	}
 }
 
-type CertVerificationOpts struct {
+type GETOpts struct {
```
litt3, line 71: Exported type without documentation (I know it was preexisting, up to you if you want to address now or not)

api/proxy/common/store.go
```diff
@@ -78,6 +78,11 @@ type CertVerificationOpts struct {
 	// within a certain number of blocks after the RBN.
 	// validity condition is: certRBN < L1InclusionBlockNum <= RBN + RBNRecencyWindowSize
 	L1InclusionBlockNum uint64
+
+	// When true, the Get method will return the encoded_payload without decoding
+	// it. This is useful when clients need to decode the encoded_payload themselves,
+	// such as inside a fpvm to prove that a decoding fails and can thus be discarded.
```
litt3: nit: "a fpvm" -> "an fpvm"

api/proxy/server/handlers_cert.go
```diff
@@ -90,21 +90,31 @@ func (svr *Server) handleGetShared(
 	if err != nil {
 		return err // doesn't need to be wrapped; already a proxyerrors
 	}
-	input, err := svr.sm.Get(
+
+	// Check if client requested encoded payload
+	// This is currently used by secure integration (e.g. optimism hokulea), which need
```
litt3: "used by secure integration" -> "used by secure *integrations*"

api/proxy/server/handlers_cert.go
litt3: > if the return_encoded_payload query parameter is present

Wait, so `?return_encoded_payload=false` *also* would cause this to return true?

api/proxy/store/generated_key/memstore/v2/memstore_test.go
```diff
@@ -50,7 +50,12 @@ func TestGetSet(t *testing.T) {
...
 
-	actual, err := msV2.Get(t.Context(), cert)
+	actual, err := msV2.Get(t.Context(), cert, false)
 	require.NoError(t, err)
 	require.Equal(t, expected, actual)
+	
+	// Test getting the encoded payload
+	encodedPayload, err := msV2.Get(t.Context(), cert, true)
+	require.NoError(t, err)
+	require.NotEqual(t, expected, encodedPayload) // Should be different
```
litt3: nit: the `// Should be different` comment could be a string param to the `NotEqual` function, so that this context is shown to the user wherever the test is failing

api/proxy/store/manager.go
```diff
@@ -110,43 +112,56 @@ func (m *Manager) Get(ctx context.Context,
 			return nil, errors.New("expected EigenDA V2 backend for DA commitment type with CertV1")
 		}
 
+		// The eigendav2 verify function verifies the cert, but not the kzg commitment.
+		// TODO: should we be verifying those when retrieving from secondary storages like S3...?
```
litt3, line 116: Not sure it matters from a *security* perspective... but it still might be useful as a sanity check, i.e. "Am I actually getting what I expect to be getting?"

api/proxy/store/manager.go
```diff
@@ -110,43 +112,56 @@ func (m *Manager) Get(ctx context.Context,
...
-			m.log.Debug("Retrieving data from cached backends")
-			data, err := m.secondary.MultiSourceRead(ctx, versionedCert.SerializedCert, false, verifyMethod, verifyOpts)
+		// 1 - read payload from cache if enabled
+		// Secondary storages (cache and fallback) store payloads instead of blobs.
+		// For simplicity, we bypass secondary storages when requesting encoded payloads,
+		// since those requests are only for secure integrations and run by provers/challengers.
+		// TODO: would be nice to store blobs instead of payloads in secondary storages, such that we could standardize all
+		// storages and make them all implement the [clients.PayloadRetriever] interface.
+		// We could then get rid of the proxy notion of caches/fallbacks and only have storages.
+		if m.secondary.CachingEnabled() && !opts.ReturnEncodedPayload {
```
litt3, line 129: You've added `!opts.ReturnEncodedPayload` to all these conditions. 

Alternatively, you could just do 

```go
if opts.ReturnEncodedPayload {
    encodedPayload, err := m.getFromCorrectEigenDABackend(ctx, versionedCert, opts)
    if err != nil {
        return nil, err
    }
    return encodedPayload, err
}
```

And then the rest of the logic would be unchanged

api/proxy/store/manager.go
```diff
@@ -200,21 +215,25 @@ func (m *Manager) backupToSecondary(ctx context.Context, commitment []byte, valu
...
 // getVerifyMethod returns the correct verify method based on commitment type
-func (m *Manager) getVerifyMethod(commitmentType certs.VersionByte) (
-	func(context.Context, []byte, []byte, common.CertVerificationOpts) error,
+func (m *Manager) getVerifyMethod(certVersion certs.VersionByte) (
+	func(context.Context, []byte, []byte, uint64) error,
 	error,
 ) {
-	v2VerifyWrapper := func(ctx context.Context, cert []byte, payload []byte, opts common.CertVerificationOpts) error {
-		return m.eigendaV2.Verify(ctx, certs.NewVersionedCert(cert, commitmentType), opts)
+	v1VerifyWrapper := func(ctx context.Context, cert []byte, payload []byte, l1InclusionBlockNumber uint64) error {
```
litt3: any reason we can't call this `v0VerifyWrapper`, and the other one simply `VerifyWrapper`, or something like that?

The switch statement below is killing me

V0VersionByte -> v1VerifyWrapper
V1VersionByte and V2VersionByte -> v2VerifyWrapper

api/proxy/store/manager.go
```diff
@@ -200,21 +215,25 @@ func (m *Manager) backupToSecondary(ctx context.Context, commitment []byte, valu
...
-	switch commitmentType {
+	switch certVersion {
 	case certs.V0VersionByte:
-		return m.eigenda.Verify, nil
+		return v1VerifyWrapper, nil
 	case certs.V1VersionByte, certs.V2VersionByte:
 		return v2VerifyWrapper, nil
 	default:
-		return nil, fmt.Errorf("commitment version unknown: %b", commitmentType)
+		return nil, fmt.Errorf("commitment version unknown: %b", certVersion)
```
litt3: variable name has changed, error text should change too

api/proxy/store/manager.go
```diff
@@ -246,15 +265,20 @@ func (m *Manager) putToCorrectEigenDABackend(ctx context.Context, value []byte)
...
 	ctx context.Context,
 	versionedCert certs.VersionedCert,
-	verifyOpts common.CertVerificationOpts,
+	opts common.GETOpts,
 ) ([]byte, error) {
 	switch versionedCert.Version {
 	case certs.V0VersionByte:
 		m.log.Debug("Reading blob from EigenDAV1 backend")
+
+		// V1 backend does not support returning encoded payloads
```
litt3: Similar comment here:

I think referring to `V1` like this, inside a switch case labeled `V0VersionByte` is just *super* confusing.

Maybe we should do our best pretend like `EigenDAV1` and `EigenDAV2` don't exist in the proxy, and ONLY talk about cert versions. Otherwise, we have all this mixed `v` terminology that is barely comprehensible to the people who wrote this code, MUCH LESS anyone else.

Thoughts?

api/clients/v2/payloadretrieval/validator_payload_retriever.go
```diff
@@ -50,11 +50,49 @@ func NewValidatorPayloadRetriever(
...
+		blobKey, keyErr := eigenDACert.ComputeBlobKey()
+		if keyErr != nil {
+			return nil, fmt.Errorf("compute blob key from eigenDACert: %w", keyErr)
+		}
+		pr.logger.Error(
+			`Commitment verification was successful, but decoding of blob to payload failed!
+				This client only supports blobs that were encoded using the codec(s) defined in
+				https://github.com/Layr-Labs/eigenda/blob/86e27fa03/api/clients/codecs/blob_codec.go.`,
+			"blobKey", blobKey.Hex(), "eigenDACert", eigenDACert, "error", err)
+		return nil, fmt.Errorf("decode blob: %w", err)
```
bxue-l2: Why does Relay payload retriever return the derivation error, but this doesn't? It looks like it should. 

> coretypes.ErrBlobDecodingFailedDerivationError.WithMessage(
> 			fmt.Sprintf("blob %v: %v", blobKey.Hex(), err))

api/clients/v2/payloadretrieval/relay_payload_retriever.go
```diff
@@ -110,10 +147,7 @@ func (pr *RelayPayloadRetriever) GetPayload(
 		//  serialization of a blob. Commitment generation operates on field elements, which is how a blob is stored
 		//  under the hood, so it's actually duplicating work to serialize the blob here. I'm declining to make this
 		//  change now, to limit the size of the refactor PR.
-		valid, err := verification.GenerateAndCompareBlobCommitment(
-			pr.g1Srs,
-			blob.Serialize(),
-			blobCommitments.Commitment)
+		valid, err := verification.GenerateAndCompareBlobCommitment(pr.g1Srs, blob.Serialize(), blobCommitments.Commitment)
```
bxue-l2, line 144: There are a lot of common codes between the different retrievers. Is it why you mentioned above that it shouldn't be the case that both retrievers implement one interface.

I guess instead they should implement some common lower interfaces like get blob?

api/clients/v2/payload_retriever.go
```diff
@@ -11,9 +11,19 @@ import (
...
+// and validator retrieval clients that implement this interface. Instead,
+// we should have a single PayloadRetriever that knows how to retrieve blobs from either
+// relays or validators, and then decodes them to (encoded) payloads.
```
bxue-l2, line 18: I think encoded payload is better because if we separate the pure processing logic, like we did in the derivation spec, then we can reuse the package for GO secure integration, like linea, I think?

api/clients/v2/payloadretrieval/validator_payload_retriever.go
```diff
@@ -99,20 +137,18 @@ func (pr *ValidatorPayloadRetriever) GetPayload(
...
 		if err != nil {
 			pr.logger.Error(
-				`Commitment verification was successful, but decoding of blob to payload failed!
-					This client only supports blobs that were encoded using the codec(s) defined in
-					https://github.com/Layr-Labs/eigenda/blob/86e27fa03/api/clients/codecs/blob_codec.go.`,
-				"blobKey", blobKey.Hex(), "quorumID", quorumID, "eigenDACert", eigenDACert, "error", err)
-			return nil, fmt.Errorf("decode blob: %w", err)
+				"convert blob to encoded payload failed",
+				"blobKey", blobKey.Hex(), "quorumID", quorumID, "error", err)
+			continue
```
bxue-l2: In theory, this procedure should not fail, and to be honest, if given the blob is valid, then it has identical data and identical length for all Quorum. I found the following calls for the error:

- taking FFT
- unknown polyForm (problem when starting the proxy)
- blobLengthSymbols from the Cert is 0
- blobLengthSymbols isn't power of 2
- payload header size is not 32 (in the code it wrote BlobHeaderSize btw)

None of them are random. So it might make sense to return early, or even panic, since there is no way above can be wrong. 

The same applies to "ToEncodedPayload" for the other retriever

api/proxy/common/store.go
```diff
@@ -92,8 +97,8 @@ type EigenDAV1Store interface {
 	Put(ctx context.Context, payload []byte) (serializedCert []byte, err error)
 	// Get retrieves the given key if it's present in the key-value (serializedCert-payload) data store.
 	Get(ctx context.Context, serializedCert []byte) (payload []byte, err error)
-	// Verify verifies the given key-value pair. opts is only used for EigenDA V2.
-	Verify(ctx context.Context, serializedCert []byte, payload []byte, opts CertVerificationOpts) error
+	// Verify verifies the cert and that the payload (after encoding) matches the kzg commitment in the cert.
+	Verify(ctx context.Context, serializedCert []byte, payload []byte) error
 }
 
 // EigenDAV2Store is the interface for an EigenDA V2 data store as well as V2 memstore.
```
bxue-l2, line 107: Should we indicate this V2 is Blazar?

api/proxy/server/handlers_cert.go
```diff
@@ -133,6 +143,18 @@ func parseCommitmentInclusionL1BlockNumQueryParam(r *http.Request) (uint64, erro
...
 
+// Parses the return_encoded_payload query parameter from the request.
+// Happy path:
+//   - if the return_encoded_payload query parameter is present (with any value), it returns true
+//   - this means it accepts ?return_encoded_payload, ?return_encoded_payload=true, ?return_encoded_payload=anything
+//
+// Unhappy paths:
+//   - if the return_encoded_payload query parameter is not provided, it returns false
+func parseReturnEncodedPayloadQueryParam(r *http.Request) bool {
+	_, exists := r.URL.Query()["return_encoded_payload"]
```
bxue-l2: Is it correct to say that if I query with a parameter `return encoded payload`, it will still go through the parsing, cert verification - basically a sequence of derivation according to our spec.

api/proxy/server/handlers_cert.go
bxue-l2: This is a good point.

api/proxy/store/generated_key/memstore/v2/memstore.go
```diff
@@ -176,12 +176,19 @@ func (e *MemStore) generateRandomCert(blobContents []byte) (coretypes.EigenDACer
 }
 
 // Get fetches a value from the store.
-func (e *MemStore) Get(_ context.Context, versionedCert certs.VersionedCert) ([]byte, error) {
+// If returnEncodedPayload is true, it returns the encoded blob without decoding.
+func (e *MemStore) Get(
```
bxue-l2, line 180: Why memstore does not observe the IManager interface, which I believe is 

> 	Get(ctx context.Context, versionedCert certs.VersionedCert,
> 		cm commitments.CommitmentMode, opts common.GETOpts) ([]byte, error)

api/proxy/store/generated_key/memstore/v2/memstore.go
```diff
@@ -176,12 +176,19 @@ func (e *MemStore) generateRandomCert(blobContents []byte) (coretypes.EigenDACer
...
+	_ context.Context, versionedCert certs.VersionedCert, returnEncodedPayload bool,
+) ([]byte, error) {
 	encodedBlob, err := e.FetchEntry(crypto.Keccak256Hash(versionedCert.SerializedCert).Bytes())
```
bxue-l2, line 183: I thought memstore will just store the payload. We just have to deterministically encode it.

api/proxy/store/generated_key/memstore/v2/memstore.go
bxue-l2: I see,  EigenDAV2store has its own interface of Get

api/proxy/common/store.go
```diff
@@ -102,9 +107,13 @@ type EigenDAV2Store interface {
...
 	Put(ctx context.Context, payload []byte) (serializedCert []byte, err error)
 	// Get retrieves the given key if it's present in the key-value (serializedCert-payload) data store.
-	Get(ctx context.Context, versionedCert certs.VersionedCert) (payload []byte, err error)
-	// Verify verifies the given key-value pair.
-	Verify(ctx context.Context, versionedCert certs.VersionedCert, opts CertVerificationOpts) error
+	// If returnEncodedPayload is true, the payload is returned without decoding.
+	Get(ctx context.Context,
+		versionedCert certs.VersionedCert,
+		returnEncodedPayload bool,
+	) (maybeEncodedPayload []byte, err error)
```
bxue-l2: PayloadOrEncodedPayload? maybeEncodedPayload is pretty vague

api/proxy/store/manager.go
```diff
@@ -110,43 +112,56 @@ func (m *Manager) Get(ctx context.Context,
...
-		// 2 - read blob from EigenDA
-		data, err := m.getFromCorrectEigenDABackend(ctx, versionedCert, verifyOpts)
+		// 2 - read maybeEncodedPayload from EigenDA
+		maybeEncodedPayload, err := m.getFromCorrectEigenDABackend(ctx, versionedCert, opts)
 		if err == nil {
-			if m.secondary.WriteOnCacheMissEnabled() {
-				m.backupToSecondary(ctx, versionedCert.SerializedCert, data)
+			// Only backup to secondary storage if we're returning the decoded payload
+			// since the secondary stores are currently hardcoded to store payloads only.
+			// TODO: we could consider also storing encoded payloads under separate keys?
```
bxue-l2, line 144: It might make sense just to require secondary storage to be payload-only because we can always re-encode it later deterministically. Saving as a payload has the benefit of saving space.

api/proxy/store/manager.go
```diff
@@ -110,43 +112,56 @@ func (m *Manager) Get(ctx context.Context,
...
+			if m.secondary.WriteOnCacheMissEnabled() && !opts.ReturnEncodedPayload {
+				m.backupToSecondary(ctx, versionedCert.SerializedCert, maybeEncodedPayload)
 			}
 
-			return data, nil
+			return maybeEncodedPayload, nil
```
bxue-l2: Maybe put a comment to indicate that this is indeed the encoded payload, but I feel there might be a better way to reflect this in the code

api/proxy/store/manager.go
```diff
@@ -110,43 +112,56 @@ func (m *Manager) Get(ctx context.Context,
...
 		}
 
 		// 3 - read blob from fallbacks if enabled and data is non-retrievable from EigenDA
-		if m.secondary.FallbackEnabled() {
-			data, err = m.secondary.MultiSourceRead(ctx, versionedCert.SerializedCert, true, verifyMethod, verifyOpts)
+		// Only use fallbacks if we're not requesting encoded payload
+		if m.secondary.FallbackEnabled() && !opts.ReturnEncodedPayload {
```
bxue-l2, line 154: Yeah, I agree that having another if condition is convoluted and is a tech debt.

api/proxy/store/manager.go
bxue-l2: Same here. https://github.com/Layr-Labs/eigenda/pull/1802#discussion_r2238582797

Suggest to always append names such as. BlazarV2, CoreV1, CertV2 etc

bxue-l2: Btw, why V1 does not? I think it is because we don't intend to support it as opposed to we cannot.

api/proxy/store/manager.go
```diff
@@ -264,21 +288,21 @@ func (m *Manager) getFromCorrectEigenDABackend(
...
-		data, err := m.eigendaV2.Get(ctx, versionedCert)
+		m.log.Debug("Reading blob from EigenDAV2 backend", "returnEncodedPayload", opts.ReturnEncodedPayload)
+		maybeEncodedPayload, err := m.eigendaV2.Get(ctx, versionedCert, opts.ReturnEncodedPayload)
 		if err != nil {
-			return nil, fmt.Errorf("get data from V2 backend: %w", err)
+			return nil, fmt.Errorf("get payload from EigenDA V2 backend: %w", err)
 		}
 
-		return data, nil
+		return maybeEncodedPayload, nil
```
bxue-l2: I guess there's no chance the old pts is ever changed while this function is called. Feels like you can return a struct that contains the data bytes and a boolean variable indicating if it is encoded.

api/clients/v2/payloadretrieval/validator_payload_retriever.go
samlaf: agree. will look at ToEncodedPayload to see where/how it can return an error. My intuition is that function itself should panic instead of returning an error

api/proxy/common/store.go
samlaf: nitpicker!

api/proxy/common/store.go
samlaf: o'oh good catch.. parsing bug:
(maybeEncoded)Payload
vs
(maybe)EncodedPayload

api/proxy/server/handlers_cert.go
samlaf: Good catch, cleaned up in [dd69627](https://github.com/Layr-Labs/eigenda/pull/1813/commits/dd69627b284989fbd88d4e6dfd38bcee4c2e73b3)

api/clients/v2/payloadretrieval/relay_payload_retriever.go
samlaf: yes. They already do, we have 2 lower level clients. These 2 higher level ones should be combined. Will do in a future PR though

api/clients/v2/payloadretrieval/relay_payload_retriever.go
samlaf: done

api/clients/v2/payloadretrieval/relay_payload_retriever.go
samlaf: [c72beb1](https://github.com/Layr-Labs/eigenda/pull/1813/commits/c72beb149be696d5c02ad48c75769a48deec2fa5)

api/clients/v2/payloadretrieval/relay_payload_retriever.go
samlaf: gosh good catch... this was totally absolutely wrong. Fixed: [c72beb1](https://github.com/Layr-Labs/eigenda/pull/1813/commits/c72beb149be696d5c02ad48c75769a48deec2fa5)

api/clients/v2/payloadretrieval/validator_payload_retriever.go
samlaf: [c72beb1](https://github.com/Layr-Labs/eigenda/pull/1813/commits/c72beb149be696d5c02ad48c75769a48deec2fa5)

api/clients/v2/payloadretrieval/validator_payload_retriever.go
samlaf: [c72beb1](https://github.com/Layr-Labs/eigenda/pull/1813/commits/c72beb149be696d5c02ad48c75769a48deec2fa5)

api/proxy/common/store.go
samlaf: [102bd98](https://github.com/Layr-Labs/eigenda/pull/1813/commits/102bd98d5acfad25ece39160df89c3a290b8a6a7)

api/proxy/store/generated_key/memstore/v2/memstore_test.go
samlaf: [102bd98](https://github.com/Layr-Labs/eigenda/pull/1813/commits/102bd98d5acfad25ece39160df89c3a290b8a6a7)

api/proxy/store/manager.go
samlaf: that's not true. We run getFromCorrectEigenDABackend in both cases, but the others only when ReturnEncodedPayload is false.

api/proxy/store/manager.go
samlaf: The point is that for blobs that can't be decoded to payload, then we can't store them.

api/proxy/store/manager.go
samlaf: [ce98992](https://github.com/Layr-Labs/eigenda/pull/1813/commits/ce98992873e90f5d6ad3f7d0c650357dd25c3fd6)

api/proxy/store/manager.go
samlaf: [ce98992](https://github.com/Layr-Labs/eigenda/pull/1813/commits/ce98992873e90f5d6ad3f7d0c650357dd25c3fd6)

api/proxy/store/manager.go
samlaf: I can get behind the BlazarV2, CoreV1, CertV2 nomenclature. Only changed here but can lazily change other places.
And yes agree with your last comment @bxue-l2 , I updated comment: [ce98992](https://github.com/Layr-Labs/eigenda/pull/1813/commits/ce98992873e90f5d6ad3f7d0c650357dd25c3fd6)

api/clients/v2/payloadretrieval/validator_payload_retriever.go
samlaf: yep fixed

api/proxy/store/manager.go
samlaf: Agree, but think we should use the proper coretypes type system instead of introducing another ad-hoc encoding here. Plan to do this in a future PR.

api/proxy/store/manager.go
samlaf: it could be either... we don't know. it depends on the value in opts

api/proxy/common/store.go
samlaf: [7b6c201](https://github.com/Layr-Labs/eigenda/pull/1813/commits/7b6c201a641aab236b1620f7499ab2793fd1995a)

api/proxy/store/generated_key/memstore/v2/memstore.go
samlaf: I'm confused by this comment. It seems like it stores blobs. What's your point? I didn't change this logic.

api/proxy/server/routing.go
```diff
@@ -131,3 +134,43 @@ func notCommitmentModeStandard(r *http.Request, _ *mux.RouteMatch) bool {
 	commitmentMode := r.URL.Query().Get("commitment_mode")
 	return commitmentMode == "" || commitmentMode != "standard"
 }
+
+// ================== QUERY PARAMS PARSING FUNCTION ==================================================
+// These query params don't affect routing, but we keep them here so that everything related to query URLs is in one place,
+// and its easy to deduct what kind of queries are supported by the proxy server by just looking at this file.
```
litt3:
```suggestion
// and its easy to deduce what kind of queries are supported by the proxy server by just looking at this file.
```

api/proxy/store/manager.go
litt3: I suppose that's my point. If you use the `if` block in my comment above, then you get the `opts.ReturnEncodedPayload == true` case out of the way, and then all remaining logic could remain unchanged. The cost is that you call `getFromCorrectEigenDABackend` in two places instead of 1.

Bikeshedding for sure... resolving

api/proxy/store/generated_key/memstore/v2/memstore.go
bxue-l2: I wasn't aware that memstore was always storing the EncodedPayload

api/proxy/server/routing.go
samlaf: [9f33e17](https://github.com/Layr-Labs/eigenda/pull/1813/commits/9f33e1742ed46c63d05339bd4828063f2a0fb6a2)

api/clients/v2/payloadretrieval/validator_payload_retriever.go
bxue-l2: Are you going to change it here or in the future PR?

api/proxy/store/manager.go
bxue-l2: talked yesterday. Now I get what you're trying to do.

api/clients/v2/payloadretrieval/validator_payload_retriever.go
samlaf: This is a very very very good catch. We just chatted offline. This surfaced that our EncodedPayload as defined in golang coretypes is wrong, or at least the blob->decodedPayload path is doing too many checks that it shouldn't be doing. We need to be more lenient with what we call a decodedPayload, and any checks related to payload header or payload size limits etc should ONLY be validated in the encodedPayload -> payload step.

Basically we ned `encodedPayload` to be closer to a `blob` than it is a `payload`. Any payload related decoding logic should only be performed in the rollup's derivation pipeline, such that if it fails the cert/blob can be discarded.
Will change to a panic here and then fix this in a followup PR... will require some thinking about how to do this properly... cc @litt3

litt3 (approved):

api/clients/v2/payloadretrieval/validator_payload_retriever.go
samlaf: [c2b1341](https://github.com/Layr-Labs/eigenda/pull/1813/commits/c2b13414005823dd5b6505a9ebba487a275aa7a8)

bxue-l2: I looked at this commit, and it looks good to me. https://github.com/Layr-Labs/eigenda/pull/1813/commits/c2b13414005823dd5b6505a9ebba487a275aa7a8

litt3 (approved):

PR #1814, opened 2025-07-29
feat(tooling): Create a `/nitpick` command

litt3:
Closes EGDA-1765

This PR:
- Defines a `/nitpick` slash command
- Makes some tweaks to the style guide, to improve on weaknesses discovered while testing the command
- Remove the `CLAUDE.md` that was in the `.claude/commands` directory. We don't need to tell claude how to be claude: commands are a core part of the system.
- Removes `.claude` from gitignore. I think this being added was a mistake during a reversion
- Tweaks `CLAUDE.md`, since the nitpicker now assumes some of the responsibilities that were being detailed there

NOTE: This was originally going to be a subagent, but I found during experimentation that the subagent really wasn't performing at the level I'd hoped. A slash command has much better behavior for now.

### DIFF
--- a/.claude/agents/nitpicker.md
+++ b/.claude/agents/nitpicker.md
@@ -103,13 +103,3 @@ This is an example of how to format the output nitpick report:
 > -If the server finds a message from a source to be invalid, then it's blacklisted.
 > +If the server finds a message from a source to be invalid, then the source is blacklisted.
 > ```
-
-## 5. Helpful Commands
-
-These commands are helpful for enforcing the style guide. They are intended to *augment* manual style checking, not
-to replace careful consideration of input: many rules included in the style guide have not been or cannot be formalized.
-
-1. Undocumented exported stuff:
-  `rg --pcre2 -n "^(?!\s*//|^\s*/\*|^\s*$)(?=\s*(?:func\s+[A-Z]\w*|type\s+[A-Z]\w*\s+(?:struct|interface)|type\s+[A-Z]\w*\s+=|const\s+[A-Z]\w*|var\s+[A-Z]\w*)).*$" <FILES>`
-2. Error wrapping verb:
-  `rg --pcre2 -n "fmt\.Errorf\([^)]*%v[^)]*\b(err|error|e)\b[^)]*\)" <FILES>`
--- a/.claude/agents/nitpicker.md
+++ b/.claude/agents/nitpicker.md
@@ -9,7 +9,7 @@ changes to identify issues that do not comply with the EigenDA style guide at do
 ## 1. Rules
 
 1. Never provide praise: only include actionable output
-2. Do not deviate from the prescribed output format: the users of this sub agent expect and require the precise format,
+2. Do not deviate from the prescribed output format: the users of this subagent expect and require the precise format,
 and any deviation, whether additive or subtractive is strictly detrimental.
   - JSON AND human readable formats must be returned! ALWAYS return both formats. This is "duplicate" data, but that's
   ok: the caller of the subagent can decide how to use the output, and what to display to the user.
--- a/.claude/agents/nitpicker.md
+++ b/.claude/agents/nitpicker.md
@@ -11,8 +11,6 @@ changes to identify issues that do not comply with the EigenDA style guide at do
 1. Never provide praise: only include actionable output
 2. Do not deviate from the prescribed output format: the users of this subagent expect and require the precise format,
 and any deviation, whether additive or subtractive is strictly detrimental.
-  - JSON AND human readable formats must be returned! ALWAYS return both formats. This is "duplicate" data, but that's
-  ok: the caller of the subagent can decide how to use the output, and what to display to the user.
 3. When making a suggestion, double check that the original and suggested text actually differ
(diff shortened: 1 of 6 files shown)


litt3:
@claude use the nitpicker subagent to review this PR.

.claude/agents/nitpicker.md
```diff
@@ -0,0 +1,339 @@
+---
```
cody-littley: Looking at the result it posted to github, several comments.

1. What's the purpose of the json output? Is it needed for something? If not, perhaps we can remove it to reduce visual clutter.
2. It suggests `ok -> okay`. Perhaps we should tell claude to allow commonly accepted language, even if that language isn't strictly "correct" as defined by some stodgy manuscript from a century ago

cody-littley (approved):

.claude/agents/nitpicker.md
```diff
@@ -0,0 +1,339 @@
...
+> Suggested fix:
+> ```go
+> // Fixed code
+> ```
+>
+> ### 2. another/file.go:78
+>
+> Brief description of what *specifically* is wrong with the original.
+>
+> Original:
```
dmanc: What about using diff blocks like

```diff
- Old
+ New
```

.claude/agents/nitpicker.md
```diff
@@ -0,0 +1,339 @@
...
+> Suggested fix:
+> ```
+> // Fixed doc
+> ```
+>
+> ...
+
+Example:
+
+> # Nitpick Report
```
dmanc: It would be good if the report was a collapsible comment, not sure how to do that maybe the bot can leave reviews rather than comment?

litt3: We definitely need to work on how this is displayed.

The main purpose of this PR is just to get the subagent checked in, and I intend to work on how to present this output in the best way in PR contexts in a subsequent PR.

Note that I manually triggered this review with a comment above: this will *not* be done automatically on all PRs until further action is taken.

> maybe the bot can leave reviews rather than comment?

I think it might be possible. We'll have to see if that's something we want: if there are too many false positives, it might be annoying to have a bunch of review comments left that are not helpful

.claude/agents/nitpicker.md
litt3: 1. JSON output is provided by the subagent because it's easier for the prime agent to use. When we have this review happening automatically on PRs, we definitely need to figure out how to prevent this from being exposed: it isn't intended for human consumption
2. The style guide already mentions not being overly strict for stuff like this. I will add this one specifically to the list of things not to check... I'm not sure yet how much of a nuisance this kind of thing is going to be

litt3: a223e7c34adfa9a092f41cc17d13999815841a1b

.claude/agents/nitpicker.md
litt3: Hmm. I like this in general, but there is some awkwardness. Consider the following example:

```diff
- func ProcessBatch(items []Item) error {
+ // ProcessBatch processes a batch of blobs before sending to the client.
+ func ProcessBatch(items []Item) error {
```

I think this is how I would express a report of "missing documentation" . But is this intuitive? It seems like we're deleting the function definition, but really we just want to add a comment.

I suppose the "standard" way would be to *just* show the addition, but then the reader would lack context about the original, making the output more difficult to read

Curious your thoughts on how you'd approach this.

.claude/agents/nitpicker.md
litt3: Thinking about it more, I think if I use the standard diff format proposed by Daniel below, that's "standard" enough to be useful for the prime agent, without needing the json. I think I'll probably remove the json format entirely

.claude/agents/nitpicker.md
litt3: Ok, I made this change and also a bunch of simplifications. TLDR; diff view is really nice, and I think it makes the json format unnecessary. So I removed the json

litt3:
@claude Re-run the nitpicker subagent for the changes in this PR. Update your previous comment to include the new findings verbatim, as provided by the subagent

.claude/agents/nitpicker.md
litt3: Done

cody-littley (dismissed):

dmanc (approved):

anupsv (approved):

anupsv:
@litt3 i wonder, can we actually get it to make the changes instead of just reviewing it?

litt3:
> @litt3 i wonder, can we actually get it to make the changes instead of just reviewing it?

@anupsv Potentially. I definitely want to do some experimentation with how this agent can be applied, after we have the definition checked in. Some potential end-states:

1. claude leaves suggestions on the actual code, where the engineer can click "accept"
2. claude leaves a single comment with nitpick output, and the engineer pastes that into a local claude instance to make the fixes
3. claude (with permission) actually makes the fixes

With # 3, I'm not sure it will be a good option right now. It depends on how many false positives claude is flagging. I've been making some tweaks to try to tune out overly pedantic suggestions, but it's still not perfect

PR #1815, opened 2025-07-29
Grpc connection pool

cody-littley:

Open multiple parallel gRPC connections with the relays and API servers.

Do not merge until after https://github.com/Layr-Labs/eigenda/pull/1811


common/grpc_client_pool.go
```diff
@@ -0,0 +1,93 @@
...
+type GRPCClientBuilder[T any] func(grpc.ClientConnInterface) T
+
+// GRPCClientPool manages a pool of one or more gRPC clients.
+type GRPCClientPool[T any] struct {
+	// clients is a slice of gRPC clients of type T.
+	clients []T
+
+	// connections is a slice of gRPC client connections. We need to track this in order to be able to close the
+	// connections when the pool is no longer needed.
+	connections []*grpc.ClientConn
```
samlaf, line 21: can't you get these from clients? Feels a bit hacky because each client can hold an arbitrary of underlying connections. How are you sure you're tracking all of them? Or are you just assuming that each client only holds 1 underlying grpc.ClientConn and `len(clients)==len(connections)`?

common/grpc_client_pool.go
```diff
@@ -0,0 +1,93 @@
...
+	"google.golang.org/grpc"
+)
+
+// TODO unit test this
+
+// A function that builds a gRPC client of type T.
+type GRPCClientBuilder[T any] func(grpc.ClientConnInterface) T
+
+// GRPCClientPool manages a pool of one or more gRPC clients.
+type GRPCClientPool[T any] struct {
```
samlaf, line 15: Won't ask for a complete rewrite but I think the "proper" way to do this would probably be to use the native grpc load balancing features: https://grpc.io/docs/guides/custom-load-balancing/

Basically a `grpc.ClientConn` is a virtual connection (a channel), that itself can hold and load balance over multiple underlying tcp connections. So the fact that we are writing our own ClientPool and load balancing those manually feels like we're not making the best use of the underlying grpc infra (for example if we want to implement the same in rust we could just use the same underlying grpc library features, but your custom implementation here would have to be rewritten in rust).

The only problem with following the grpc load-balancing way is that grpc client-side load-balancing is meant for server-side trusted clients, whereas in our case we have cloudflare and a sticky nlb in front of our servers. It's still doable but I think it would require some perhaps non-trivial custom resolver/picker logic.

So prob not worth wasting a lot of time into this, but I think its good to keep in mind, especially as we talk about potential backend rearchitectures etc (how we use dns, how we use cloudflare, etc)

common/grpc_client_pool.go
```diff
@@ -0,0 +1,93 @@
...
+		poolSize = 1
+	}
+
+	// Create the clients up front.
+	connections := make([]*grpc.ClientConn, 0, poolSize)
+	clients := make([]T, 0, poolSize)
+	for i := 0; i < poolSize; i++ {
+		conn, err := grpc.NewClient(url, dialOptions...)
+		if err != nil {
+			return nil, fmt.Errorf("failed to create gRPC client connection to %s: %w", url, err)
```
samlaf, line 46: Note: I really hate that the golang library decided to use "connection" instead of "channel", because I was confused up until today and thought that this NewClient was actually creating an underlying physical tcp connection. But it's actually not! Those are lazily created when sending the first grpc....

I think all that this does here is do name resolution, so this could fail if the dns name doesn't resolve to an ip for example.

disperser/controller/BatchMetadataCache.go
```diff
@@ -0,0 +1,27 @@
+package controller
+
+import "github.com/Layr-Labs/eigenda/core"
+
+// When creating a new batch, the controller needs two important pieces of metadata:
+// 1. a reference block number (or RBN)
+// 2. for each quorum number associated with the batch, the operator state for that quorum
+//
+// When it comes time to create a new batch, the primary goal of this cache is to avoid having to perform a
+// blocking RPC call in order to retrieve the required metadata.
```
samlaf: I believe our current policy is to choose `RBN=current_block - 75`. How do you plan to achieve such a policy if we cache like this?

We were talking at some point about using the latest finalized RBN block number as well. That would def require a blocking call.

In the context of CAFE/multichain though we have talked about potentially setting RBN to some up-to-24h-stale block (block at which state is bridged to other chains, once per day). In that case this cache would make a lot of sense.

node/config.go
```diff
@@ -80,10 +80,12 @@ type Config struct {
 	ChurnerUseSecureGrpc            bool
 	RelayUseSecureGrpc              bool
 	RelayMaxMessageSize             uint
-	ReachabilityPollIntervalSec     uint64
-	DisableNodeInfoResources        bool
-	StoreChunksRequestMaxPastAge    time.Duration
-	StoreChunksRequestMaxFutureAge  time.Duration
+	// The number of connections to establish with each relay node.
+	RelayConnectionPoolSize        int
```
samlaf: why do you use ints and not uints for poolSizes?

api/clients/v2/disperser_client.go
```diff
@@ -14,14 +14,16 @@ import (
...
+	"github.com/Layr-Labs/eigensdk-go/logging"
 	"github.com/docker/go-units"
-	"google.golang.org/grpc"
 )
 
 type DisperserClientConfig struct {
 	Hostname          string
 	Port              string
 	UseSecureGrpcFlag bool
+	// The number of grpc connections to the disperser server.
```
samlaf:
```suggestion
	// The number of grpc connections to the disperser server. Values <= 0 are mapped to 1.
```

samlaf: prob good to use uint instead?

api/clients/v2/disperser_client.go
```diff
@@ -388,13 +398,12 @@ func (c *disperserClient) initOnceGrpcConnection() error {
...
 		addr := fmt.Sprintf("%v:%v", c.config.Hostname, c.config.Port)
 		dialOptions := GetGrpcDialOptions(c.config.UseSecureGrpcFlag, 4*units.MiB)
-		conn, err := grpc.NewClient(addr, dialOptions...)
-		if err != nil {
-			initErr = err
-			return
-		}
-		c.conn = conn
-		c.client = disperser_rpc.NewDisperserClient(conn)
+		c.clientPool, initErr = common.NewGRPClientPool(
```
samlaf: note: As I said in another comment, I think this function `initOnceGrpcConnection` (that I initially added back then) is misguided, because the underlying tcp connections are only lazily created when an rpc is first made. So we could just create these once inside the constructor.

maybe or may not be worth including as part of this PR, but I think we should eventually do this refactor.

api/clients/v2/relay/relay_client.go
```diff
@@ -60,15 +62,14 @@ type relayClient struct {
 	config *RelayClientConfig
 	// relayLockProvider provides locks that correspond to individual relay keys
 	relayLockProvider *KeyLock[corev2.RelayKey]
+	// connectionPoolSize is the number of parallel connections open to each relay.
+	connectionPoolSize int
```
samlaf: nit: don't think this needs to be stored here if we initialize the connectionPool inside the constructor like I mentioned in other comments

node/flags/flags.go
```diff
@@ -453,9 +460,9 @@ var (
 	}
 	DownloadPoolSizeFlag = cli.IntFlag{
 		Name:     common.PrefixFlag(FlagPrefix, "download-pool-size"),
-		Usage:    "The size of the download pool. The default value is 16.",
+		Usage:    "The size of the download pool.",
```
samlaf, line 471: for my own knowledge, what is a download pool?

api/clients/v2/relay/relay_client.go
```diff
@@ -91,13 +92,19 @@ func NewRelayClient(
 		return nil, errors.New("max gRPC message size must be greater than 0")
 	}
 
+	connectionPoolSize := config.ConnectionPoolSize
+	if connectionPoolSize <= 0 {
```
anupsv, line 99: Is there a max above which we should not allow ?

api/clients/v2/disperser_client.go
```diff
@@ -388,13 +398,12 @@ func (c *disperserClient) initOnceGrpcConnection() error {
...
+			c.logger,
+			disperser_rpc.NewDisperserClient,
+			c.config.DisperserConnectionCount,
```
anupsv: we should limit ?

common/grpc_client_pool.go
anupsv: Had to claude the golang implmentation to understand but is true.

common/grpc_client_pool.go
cody-littley: > can't you get these from clients?

The clients are a generic type `T`. The actual values of `T` are the compiled gRPC clients, e.g. `DisperserClient`. Unfortunately, these compiled gRPC clients don't have any method that allows you to get the internal `grpc.ClientConn` object, and they don't have any parent interface that allows you to fetch it. So if we want to be able to access the `grpc.ClientConn` object, we have to save it at construction time.

cody-littley: > Feels a bit hacky because each client can hold an arbitrary of underlying connections.

I'm under the impression that the golang gRPC implementation has exactly one TCP connection per client object, and that one TCP connection is multiplexed into multiple logical channels.

In the `NewGRPCClientPool()` constructor, note that I construct exactly one `grpc.ClientConn` connection object per client, and the client constructor (auto-generated) only provides the option to use one connection object.

common/grpc_client_pool.go
cody-littley: Looking through the linked example, the way they describe doing load balancing seems way more complex than the way I do it in this branch. https://github.com/grpc/grpc-go/blob/master/examples/features/customloadbalancer/client/customroundrobin/customroundrobin.go Even if this is the intended way of using the gRPC library, the API seems half baked and hard to use.

disperser/controller/BatchMetadataCache.go
cody-littley: Oops, this file doesn't belong in this commit. Deleted.

node/config.go
cody-littley: changed to uint

api/clients/v2/disperser_client.go
cody-littley: done

api/clients/v2/disperser_client.go
cody-littley: Yeah, I think I agree with you on both counts. Will refactor in a future PR.

api/clients/v2/relay/relay_client.go
cody-littley: The current relay sharding algorithm allows relay shards to be added dynamically. If that were to ever happen, it would be necessary to create a new connection to that new relay shard, and so I'm not sure we can always initialize all of your relay clients in the constructor.

node/flags/flags.go
cody-littley: There is a thread pool where the chunk download happens. The pool is called the "download pool". Each thread can be downloading a chunk simultaneously. It is intended to act as a sanity limit on the number of concurrent downloads.

api/clients/v2/relay/relay_client.go
cody-littley: Added an upper limit. Note that we can't actually enforce this on the client side if the client is malicious. But it will prevent it from going above a certain limit by an unmodified client.

api/clients/v2/disperser_client.go
cody-littley: added a hard coded limit

common/grpc_client_pool.go
samlaf: > one TCP connection per client object, and that one TCP connection is multiplexed into multiple logical channels

Afaiu its the other way around (see [this](https://github.com/grpc/grpc-go/issues/682#issuecomment-360531305) for example). Chatting with claude it looks like (?) the relationship might be
```
Client 1---n Channel 1---n HTTP/2 Connection 1---n RPC (HTTP/2 Stream) 1---n Messages (HTTP/2 Data Frames)
```

> client constructor (auto-generated) only provides the option to use one connection object

I'd have to look up the details again, but you have to use the [load balancing](https://github.com/grpc/grpc/blob/master/doc/load-balancing.md) features. I believe this happens under the hood via the dns resolution (one of the techniques to get this going). If you give the client a dns address which resolves to multiple ip addresses, it'll open a connection to all of them. There are other simpler ways to get this setup that would work for our case afaiu (like writing a custom resolver).

common/grpc_client_pool.go
samlaf: Most normal policies work out of the box though. Custom load balancing policies are very advanced use cases. But tbh last time I tried to set this up it was also clonky (afair you need to pass your whole config as a large string ala sql).

node/flags/flags.go
samlaf: Ah gotcha. Suggest changing the usage string to explicitly mention chunk then: "The size of the chunk download pool."
Assuming that new operators will be just as confused as me.

common/grpc_client_pool.go
cody-littley: I did some further digging, and it appears as if the only way to use the built in load balancer is if there are multiple public endpoints. The golang library collapses duplicate URLs down, so we can't use the trick of passing the load balancer multiples of the same URL.

Since adding multiple URLs is going to require a broader conversation and potentially higher load for users to configure, I think I'd prefer to stick with my hacky solution in the short term.

samlaf: Yep makes sense. I was thinking there's prob a way to implement our own dns resolver to not collapse the URLs, but no need to do this now.

api/clients/v2/disperser_client.go
samlaf: Wrote https://github.com/Layr-Labs/eigenda/pull/1876

PR #1817, opened 2025-07-30
feat: Enable additional linter rules just for new changes

litt3:
- This PR modfies the github linter action to only lint the diff between a PR branch and the base branch
- This PR enables revive and some other linting rules
- This PR removes the separate proxy linting rules and CI action.
    - Now that proxy is part of mono repo, it makes sense to just define this stuff once


dmanc (approved):

litt3:
@anupsv This PR removes the separate proxy linting CI action.

I think it needs to be removed from the list of `required` actions, but AFAICT I don't have the permissions do to this. Would you mind doing that for me?

.github/workflows/golangci-lint.yml
```diff
@@ -17,10 +17,21 @@ jobs:
       - name: Checkout EigenDA
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 #4.2.2
 
+      # Fetch git history of the base branch so golangci-lint can analyze only the diff
+      # between the feature branch and base branch (via --new-from-rev flag)
+      - name: Fetch base branch history
+        if: github.event_name == 'pull_request'
+        run: bash .github/scripts/fetch-base-incrementally.sh "${{ github.event.pull_request.base.ref }}"
```
samlaf: do you really need this complex `fetch-base-incrementally.sh` script? Does the github actions/checkout not already provide a way to checkout the target branch?

Looking at https://stackoverflow.com/questions/62331829/how-to-get-the-target-branch-of-the-github-pull-request-from-an-actions it seems like it does?

.golangci.yml
```diff
@@ -6,21 +6,50 @@ run:
...
-  # Right now revive raises 50+ issues, whereas golint didn't have any.
-  # despite revive being recommended as a replacement for golint.
-  # TODO: should we turn on revive and fix the issues?
-  #     - revive # drop-in replacement for golint
-  exclusions:
+    - protogetter # reports direct reads from proto message fields when getters should be used
+    - revive      # drop-in replacement for golint
+    - lll         # enforces line length limits
+    - errorlint   # makes sure errors are wrapped correctly
+    - misspell    # checks for common misspellings
```
samlaf, line 13: oh wut :O can't wait to see if this is good or not

.golangci.yml
```diff
@@ -6,21 +6,50 @@ run:
...
+    - funlen      # limits function length
+    - nestif      # limits nesting depth
+    - exhaustive  # makes sure enum switch statements are exhaustive
```
samlaf, line 16: this gud

.golangci.yml
```diff
@@ -6,21 +6,50 @@ run:
...
+    - errcheck    # enforces that all errors are checked
+    - unused      # checks for unused constants, variables, functions and types
+    - unconvert   # removes unnecessary type conversions
+    - wrapcheck   # checks that errors returned from external packages are wrapped
+    - exhaustruct # checks if all structure fields are initialized
```
samlaf, line 21: this I think can be annoying, and also goes against the golang philosophy of not initializing default values?

.golangci.yml
```diff
@@ -6,21 +6,50 @@ run:
...
+    - govet       # reports suspicious constructs
```
samlaf, line 22: oh they have this here?? Then can remove from the makefile `lint` target (I had added manually `go vet`)

Let's make sure its the same thing first I guess

.github/workflows/golangci-lint.yml
```diff
@@ -17,10 +17,21 @@ jobs:
...
         with:
           version: ${{ env.MISE_VERSION }}
           experimental: true
       - run: go version
-      - run: make lint
+      
+      - name: Run linter
+        run: |
+          # Use PR base branch if available, otherwise the command line flag is ignored and master is used
+          golangci-lint run --new-from-rev=origin/${{ github.event.pull_request.base.ref || 'master' }}
```
samlaf, line 31: hmm so by default itll lint against master branch? But doing this will lint against the PR target branch instead?

For ultra simplicity, I would suggest KISS and only run when target branch is master? So that we don't need this complex fetch-base-incrementally script that I dont understand (didnt take the time to read it tbh)

samlaf: Otherwise I would add this comment above the Makefile `lint` target, something like:
"this will lint by comparing against master. To comapre against another branch, run `golangci-lint run --new-from-rev=...`"

.github/scripts/fetch-base-incrementally.sh
```diff
@@ -0,0 +1,58 @@
+#!/bin/bash
+set -euo pipefail
+
+# Incrementally fetch git history until we find the merge base with the target branch
+# Usage: ./fetch-base-incrementally.sh <base-ref>
```
samlaf: hmmm.. can't we just `git checkout target-branch` and revert back to current branch instead and it'll download everything we need?

samlaf (approved):
LGTM. Left some nits but happy merging this as is

.golangci.yml
litt3: > golang philosophy of not initializing default values

I feel pretty strongly that this philosophy is a foot gun. We've seen first hand multiple critical bugs that came about because of this!

Essentially, this linter forces you to use raw struct initialization like you'd use a constructor, where each field is a parameter. For cases where you want some default values, you should just create a constructor that accepts any arbitrary parameters, and explicitly configures each struct field internally, IMO

.github/workflows/golangci-lint.yml
litt3: I struggled with this one. It _seems_ like it should work, but the github actions checkout did not work for this out of the box. 

It seems likely that there is some simpler solution out there, but both claude and chatgpt seemed convinced that this was a reasonable solution 🤷 

I'll take a look at your link, and reconsider if this can be simplified. I agree the current script feels overcomplicated

.golangci.yml
litt3: I would definitely be against it if it was a "normal" spellcheck. But it seems written to just check for common, unambiguous misspellings. I've high hopes for this one

.github/workflows/golangci-lint.yml
litt3: I'll spend a bit more time trying to figure out how to avoid the complex script.

If I can't figure out how to ditch the script for something simpler, I'll follow KISS and only run against master.

.github/workflows/golangci-lint.yml
samlaf: don't waste too much time..! let's run with this

litt3: Lol. I removed all my fancy logic and just made it so we fetch _everything_, and it increased the lint time by... negative 8 seconds 🫠 

This was a waaay premature optimization in hindsight. Removed and good riddance

.golangci.yml
litt3: Claude claims that these are the exact same thing. I removed the manual `go vet` call.

c7a7988ecd98afbb999c94706a137af97e2753ce

.github/workflows/golangci-lint.yml
litt3: see https://github.com/Layr-Labs/eigenda/pull/1817#discussion_r2245581092

TLDR; keeping the bit that runs lint against the target branch, but ugly script is gone

.github/scripts/fetch-base-incrementally.sh
litt3: see https://github.com/Layr-Labs/eigenda/pull/1817#discussion_r2245581092. Script is gone

anupsv (dismissed):

.github/workflows/golangci-lint.yml
```diff
@@ -16,11 +16,18 @@ jobs:
     steps:
       - name: Checkout EigenDA
         uses: actions/checkout@11bd71901bbe5b1630ceea73d27597364c9af683 #4.2.2
+        with:
+          fetch-depth: 0  # Fetch all history for all branches so golangci-lint can analyze the diff
 
       - uses: jdx/mise-action@v2
```
anupsv, line 22: can we use the hash here as well please.

litt3: You mean for mise?

I can open a separate PR for that if you'd like, since it's not related to this linter change.

bxue-l2 (dismissed):
lgtm

Makefile
```diff
@@ -37,7 +37,6 @@ protoc-local:
 
 lint:
 	golangci-lint run
-	go vet ./...
 	# Uncomment this once we update to go1.23 which makes the -diff flag available.
 	# See https://tip.golang.org/doc/go1.23#go-command
 	# go mod tidy -diff
```
samlaf, line 43: how about we uncomment this line that I forgot to uncomment and also add it to CI?

Makefile
```diff
@@ -37,7 +37,6 @@ protoc-local:
 
 lint:
```
samlaf, line 39:
```suggestion
# Only lints the diff between current branch and master because of settings in .golangci.yml
lint:
```

samlaf (dismissed):

Makefile
litt3: Ok with me doing this in a followup PR, to avoid doubling up on PR purpose?

Makefile
litt3: 032016330b85f0ed325f4e7dfec08b5d7ef89471

samlaf (approved):

.golangci.yml
samlaf: I think I can agree to this. Even stronger claim is that having "useful zero values" is a mistake (see https://news.ycombinator.com/item?id=26637510) and is one of the reasons why golang will never be able to add an Option type.

One possible counter-argument is if a config struct has a lot of useful default zero values (for example in test), and I want to only change one or two for a specific test. Then if I'm force to add all the other zero values because of this linter, it makes it harder to see which values are actually important/valuable to my test?

litt3: > One possible counter-argument is if a config struct has a lot of useful default zero values

Fair point. I think in this case, I still prefer adding test overhead to gain the added safety. There's always the option of making a "default config", which in practice sidesteps this linting rule, and lets you rely on default values.

PR #1820, opened 2025-07-30
docs: Simplify and improve payment header docs

litt3:
closes EGDA-1671

- Docs were written to explain current usage, *and* be compatible with proposed leaky bucket reservation algorithm
- More doc changes will follow, I'm just starting here
- I took the opportunity to add `protoc` tools to `mise`, since it's necessary whenever updating protobuf docs


api/proto/common/v2/common_v2.proto
```diff
@@ -78,149 +78,72 @@ message Batch {
   repeated BlobCertificate blob_certificates = 2;
 }
 
-// PaymentHeader contains payment information for a blob, which is crucial for validating and processing dispersal requests.
-// The PaymentHeader is designed to support two distinct payment methods within the EigenDA protocol:
+// PaymentHeader contains payment information for a blob. Payments are handled on-chain by the PaymentVault contract:
```
samlaf: Shouldn't we say that reservations and on-demand *deposits* are made on-chain, but "payments" themselves are made/checked at the disperser?

Otherwise what would you call "setting the cumulative_payment in the PaymentHeader to pay for a blob dispersal"?

api/proto/common/v2/common_v2.proto
```diff
@@ -78,149 +78,72 @@ message Batch {
...
-// 1. Reservation-based payment system:
-//    This system allows users to reserve bandwidth in advance for a specified time period. It's designed for
-//    users who need predictable throughput with a fixed ratelimit bin in required or custom quorums.
-//    Under this method, the user pre-arranges a reservation with specific parameters on the desired quorums:
-//    - symbolsPerSecond: The rate at which they can disperse data
-//    - startTimestamp and endTimestamp: The timeframe during which the reservation is active
+// Two payment methods are supported:
+// 1. Reservation:
+//    - Users reserve bandwidth in advance for a specified time period.
+//    - Reservations are procured out-of-band, and are tracked in the PaymentVault.
```
samlaf:
```suggestion
//    - Reservations are procured out-of-band, and are set in the PaymentVault by the EigenFoundation.
```

api/proto/common/v2/common_v2.proto
```diff
@@ -78,149 +78,72 @@ message Batch {
...
+//    - On-demand payments are limited to quorums 0 and 1.
+//    - On-demand payments can only be used when dispersing through the EigenDA disperser. Currently, the EigenDA
+//      disperser is the *only* disperser, but this restriction will remain in place even with decentralized dispersal.
 //
-// 2. On-demand payment system:
-//    This is a pay-as-you-go model where users deposit funds into the PaymentVault contract and
-//    payments are deducted as they make dispersal requests. This system is more flexible but has
-//    more restrictions on which quorums can be used (currently limited to quorums 0 and 1).
+// The "size" of a dispersal, with respect to payments, is defined as the number of symbols being dispersed, rounded up
+// to the nearest multiple of the minNumSymbols defined in the PaymentVault contract. This size is relevant for both
```
samlaf: oh really? So that means we can only pay for 128KiB, 256KiB, 384KiB, etc?
Thought blobs had to be a multiple of 2? Aka 384KiB is not possible

api/proto/common/v2/common_v2.proto
```diff
@@ -78,149 +78,72 @@ message Batch {
...
+//    cumulative payment.
+//    - On-demand payments are limited to quorums 0 and 1.
+//    - On-demand payments can only be used when dispersing through the EigenDA disperser. Currently, the EigenDA
+//      disperser is the *only* disperser, but this restriction will remain in place even with decentralized dispersal.
 //
-// 2. On-demand payment system:
-//    This is a pay-as-you-go model where users deposit funds into the PaymentVault contract and
-//    payments are deducted as they make dispersal requests. This system is more flexible but has
-//    more restrictions on which quorums can be used (currently limited to quorums 0 and 1).
+// The "size" of a dispersal, with respect to payments, is defined as the number of symbols being dispersed, rounded up
```
samlaf:
```suggestion
// The "size" of a dispersal, with respect to payments, is defined as the number of symbols (aka bn254 field elements) being dispersed, rounded up
```

api/proto/common/v2/common_v2.proto
```diff
@@ -78,149 +78,72 @@ message Batch {
...
-// If no valid reservation exists or if the reservation doesn't have enough remaining bandwidth,
-// the client will fall back to on-demand payment, provided the user has deposited sufficient funds
-// in the PaymentVault contract.
+// The cost of an on-demand dispersal is calculated by multiplying this size by the pricePerSymbol defined in the
+// PaymentVault contract.
 //
-// The distinction between these two payment methods is made by examining:
-// - For reservation-based: The timestamp must be within an active reservation period, and cumulative_payment is zero or empty
-// - For on-demand: The cumulative_payment field contains a non-zero value representing the total payment for all dispersals
+// Note: the quorum set being dispersed to has no impact on payment accounting with the current implementation.
```
samlaf, line 102: so dispsersing to a custom quorum doesnt cost any more than dispersing to only eth/eigen quorums?

api/proto/common/v2/common_v2.proto
```diff
@@ -78,149 +78,72 @@ message Batch {
...
+// TODO(litt3): once accounting logic has been properly abstracted, put a link here to provide specific documentation of
+// how payments are processed.
 message PaymentHeader {
-  // The account ID of the disperser client, represented as an Ethereum wallet address in hex format
-  // (e.g., "0x1234...abcd"). This field is critical for both payment methods as it:
-  //
-  // 1. Identifies whose reservation to check for reservation-based payments
-  // 2. Identifies whose on-chain deposit balance to check for on-demand payments
-  // 3. Provides the address against which the BlobHeader signature is verified
+  // The account ID of the dispersing user, represented as an Ethereum wallet address in hex format.
```
samlaf: do we allow 0x prefix? is it mandatory?

api/proto/common/v2/common_v2.proto
```diff
@@ -78,149 +78,72 @@ message Batch {
...
-  //    - It must be < the reservation's endTimestamp (in seconds)
+  // The timestamp represents the nanosecond UNIX timestamp at the time the dispersal request is created.
   //
-  // 3. Period Window Check:
-  //    The server validates that the request's reservation period is either:
-  //    - The current period (based on server time)
-  //    - The immediately previous period
-  //    This prevents requests with future timestamps or very old timestamps.
+  // The timestamp plays the role of a nonce, optionally allowing the same blob data to be dispersed multiple times
+  // while still having a unique blob header hash.
```
samlaf:
```suggestion
  // while still having a unique blob header hash (which is used as an idempotency key).
```

api/proto/common/v2/common_v2.proto
```diff
@@ -78,149 +78,72 @@ message Batch {
...
   //
-  // 4. Rate Limiting:
-  //    The server uses the timestamp to allocate the request to the appropriate rate-limiting bucket.
-  //    Each reservation period has a fixed bandwidth limit (symbolsPerSecond * reservationPeriodInterval).
+  // When dealing with reservations, the timestamp determines which reservation bucket the dispersal falls into.
+  // TODO(litt3): there is an ongoing effort to use a leaky bucket algorithm instead of a bin algorithm to track
```
samlaf:
```suggestion
  // TODO(litt3): there is an ongoing effort to use a leaky bucket algorithm instead of a fixed window algorithm to track
```

samlaf: not sure about this one, but I tend towards using "fixed window" since it is googleable. Unless by "bin" we mean "our weird fixed window with overfills" implementation

api/proto/common/v2/common_v2.proto
```diff
@@ -78,149 +78,72 @@ message Batch {
...
-  //    Treating all on-demand requests as an user-agnostic more frequent reservation, timestamp is checked
-  //    against the OnDemandSymbolsPerSecond and OnDemandPeriodInterval.
-  //
-  // The timestamp is typically acquired by calling time.Now().UnixNano() in Go and accounted for NTP offsets
-  // by periodically syncing with a configuratble NTP server endpoint. The client's Accountant component
-  // (api/clients/v2/accountant.go) expects the caller to provide this timestamp, which it then
-  // uses to determine the correct reservation period and check bandwidth availability.
+  // The timestamp is currently unused in the context of on-demand payments, but this is subject to change without
+  // notice! Failure to populate this with a proper timestamp could result in failed dispersals and loss of associated
+  // payments.
```
samlaf, line 131: hmm I'm against this. a 400 validation error should come before debiting the account imo.

api/proto/common/v2/common_v2.proto
```diff
@@ -78,149 +78,72 @@ message Batch {
...
-  //    - If other requests from the same account are currently processing, ensures this new cumulative
-  //      value is consistent with those (preventing double-spending)
-  //
-  // 4. On-chain Implementation:
-  //    The PaymentVault contract maintains:
-  //    - A deposit balance for each account
-  //    - Global parameters including minNumSymbols, GlobalSymbolsPerSecond and pricePerSymbol
-  // Due to the use of cumulative payments, if a client loses track of their current cumulative payment value,
-  // they can query the disperser server for their current payment state using the GetPaymentState RPC.
+  // The cumulative_payment field is a serialized uint256 big integer representing the total wei paid by the account
```
samlaf:
```suggestion
  // The cumulative_payment field is a big endian serialized uint256 representing the total wei paid by the account
```

samlaf: big integer is meaningless here, uint256 contains all the information we need (I'm guessing bigInt is how its represented in golang, which is why its here, but don't think that should matter, we could switch the representation in golang). 

Not 100% sure its big endian serialized.. we should check.

api/proto/common/v2/common_v2.proto
```diff
@@ -78,149 +78,72 @@ message Batch {
...
+  // for this and all previous dispersals.
+  //
+  // If this field is *not* set, or is zero, reservation accounting will be used. If this field *is* set, and non-zero,
+  // on-demand accounting will be used EVEN IF a given account has a reservation. There is no fallback between these
+  // payment mechanisms: the dispersal will either succeed or fail on the basis of the implicitly defined payment
+  // mechanism, regardless of whether the alternate mechanism would have succeeded.
+  //
+  // Since the cumulative payment covers all historical on-demand dispersals, a client starting up must obtain the
+  // value of the latest cumulative payment for its account via the GetPaymentState disperser RPC.
```
samlaf, line 148: newcomers to this proto would be confused I think. Prob good for first paragraph to explain what cumulative is and means?

samlaf: Perhaps giving an example of dispersing 3 reservations, and cumulative payment being 100, 200, 300 or some arbitrary example?

api/proto/common/v2/common_v2.proto
```diff
@@ -78,149 +78,72 @@ message Batch {
...
+  //
+  // IMPORTANT: With the current implementation, the cumulative payment of dispersals must be strictly increasing from
+  // the perspective of the entity doing the accounting. If a given cumulative payment X is <= the cumulative payment
+  // of a previous dispersal, then X is considered to be invalid. The implication is that a user must not behave in any
+  // way that could result in payments being processed out of order, or risk dispersals failing without refund. In
+  // practice, that means waiting for confirmation from the disperser that a blob has been received before submitting
+  // the next blob.
```
samlaf, line 155: ugh thanks for finally writing this clearly! This explains the mutex we have on the client then. This is so bad...

I'm not sure how to protect from this. Even if we have the singleton controller where payment is serialized, the requests could still arrive out of order... we need to make sure causal order is maintained. Meaning either we serialize on the client, or else the server is forced to buffer somewhere (either in tcp if we force grpc to use a single tcp connection with single http/2 stream, or on app-side, as you mention below).

samlaf:
> I propose the following:
> 
> 1. I'll revert the mise changes in this PR for now
> 2. In a followup PR, I'll delete the docker based protoc entirely, and define the correct tools in `mise`.
>    
>    * Then, there will only be 1 version of the `protoc` make command
>    * The only thing I need to double check is whether the github action is already using `mise` or not to install dependencies.  If this is already done / easy to add, perfect. If it's too much work to do this, we can just say "whatever", and leave things unchanged

Big fan of this

api/proto/common/v2/common_v2.proto
litt3: 73eef451ff5062b173fac1eba16bfd67d32d4412

Attempted to clarify the various payment sources of truth

api/proto/common/v2/common_v2.proto
litt3: d75da9740b01e52b1595a8a89baaaf08faf71199

api/proto/common/v2/common_v2.proto
litt3: I think this statement is _technically_ correct. The cost _is_ rounded up to the next multiple of `128KB`... it's just that the only case where a blob size isn't _already_ a multiple of `128KB` (due to the power of 2 requirement) is for blobs that are smaller than `128KB`. 

So I think an equally correct, but simpler, way to express this would be to say that blobs smaller than `128KB` are charged for `128KB`.

Please do check my reasoning here, but I'm gonna go ahead and make this change.

litt3: 75127856378c376201ad47f4d8873bebf0c1f998

api/proto/common/v2/common_v2.proto
litt3: I reworded this sentence for a previous comment, so this comment isn't directly applicable anymore.

Still, I'm curious if you you think we should be defining "symbols" here. IMO it seems like `symbols` are fundamental enough to our system that we shouldn't feel obligated to give the "aka" everywhere we talk about them.

api/proto/common/v2/common_v2.proto
litt3: Correct. 

1. On-demand payments don't work for custom quorum
2. For reservations, the only thing that matters is if you're entitled to disperse to that quorum. It doesn't increase cost to do so, it's just a boolean check

api/proto/common/v2/common_v2.proto
litt3: 74a0c122bedae84283d0dbda5041b821d2d387ca

api/proto/common/v2/common_v2.proto
litt3: 196ceb293326930b9606dbf950e0de306ac985af

api/proto/common/v2/common_v2.proto
litt3: 64aa57e738b6a3d8644bb19501958d8bff4de8eb

No strong preference, your argument for "fixed window" makes sense to me

api/proto/common/v2/common_v2.proto
litt3: I'm not claiming we _should_ do anything in particular here. Just that there is a contract for users to populate this honestly, and the behavior if they don't is subject to change without notice.

I don't think we should box ourselves into any specific behavior here: my statement is intended as a disclaimer so if people _don't_ follow this and get broken later, they can't complain

api/proto/common/v2/common_v2.proto
litt3: Like I mention in the TODO below, I don't think it would actually be that hard to permit some level of concurrency. If you keep track of some metadata for recent payments, it's ok if they arrive a bit out of order

api/proto/common/v2/common_v2.proto
litt3: a8c9ad18efa821d77d3c777c158da4373f4f227c

api/proto/common/v2/common_v2.proto
litt3: f608886fb47dd11d23e0f65ac690f1add1598e47

Added a brief example

litt3:
@claude /nitpick this PR. 

Once obtaining the output from the nitpick command, leave individual PR suggestions directly on the code with the results. Additionally, separately comment the entire `/nitpick` command output verbatim

litt3:
@claude /nitpick this PR, using the claude slash command found at `.claude/commands/nitpick.md`

Once obtaining the output from the nitpick command, leave individual PR suggestions directly on the code with the results. Additionally, separately comment the entire `/nitpick` command output verbatim

api/proto/common/v2/common_v2.proto
samlaf: That makes sense. So are we actually rounding up to the next multiple of 128KiB in the code somewhere? Might be simpler and require less reasoning to only enforce the min constraint in the code, so that people don't have to make this reasoning jump that you made (which I agree is correct)?

api/proto/common/v2/common_v2.proto
samlaf: Asked the question on slack: https://eigenlabs.slack.com/archives/C08SR37JZ7X/p1754503676460419

I wish we only used a single term everywhere. I agree it doesn't make a lot of sense to say "aka" everywhere though.

samlaf (approved):

api/proto/common/v2/common_v2.proto
litt3: Correct. Relevant code at https://github.com/Layr-Labs/eigenda/blob/master/api/clients/v2/accountant.go#L154

```
func (a *Accountant) paymentCharged(numSymbols uint64) uint64 {
	return a.symbolsCharged(numSymbols) * a.pricePerSymbol
}

// symbolsCharged returns the number of symbols charged for a given data length
// being at least MinNumSymbols or the nearest rounded-up multiple of MinNumSymbols.
func (a *Accountant) symbolsCharged(numSymbols uint64) uint64 {
	if numSymbols <= a.minNumSymbols {
		return a.minNumSymbols
	}
	// Round up to the nearest multiple of MinNumSymbols
	return core.RoundUpDivide(numSymbols, a.minNumSymbols) * a.minNumSymbols
}
```

api/proto/common/v2/common_v2.proto
litt3: +1 would be really nice to use FE everywhere, and ditch "symbols"

api/proto/common/v2/common_v2.proto
litt3: Agreed.

PR #1821, opened 2025-07-31
fix: don't return internal error for data not found

cody-littley:

Avoid returning internal errors when data can't be found.
https://linear.app/eigenlabs/issue/EGDA-1797/dont-return-internal-error-when-data-not-found


relay/server.go
```diff
@@ -188,8 +188,12 @@ func (s *Server) GetBlob(ctx context.Context, request *pb.GetBlobRequest) (*pb.G
 	keys := []v2.BlobKey{key}
 	mMap, err := s.metadataProvider.GetMetadataForBlobs(ctx, keys)
 	if err != nil {
-		return nil, api.NewErrorInternal(fmt.Sprintf(
-			"error fetching metadata for blob, check if blob exists and is assigned to this relay: %v", err))
+		if strings.Contains(err.Error(), blobstore.ErrMetadataNotFound.Error()) {
```
dmanc, line 191: Shouldn't this also check for ErrBlobNotFound?

relay/server.go
```diff
@@ -307,8 +311,12 @@ func (s *Server) GetChunks(ctx context.Context, request *pb.GetChunksRequest) (*
 
 	mMap, err := s.metadataProvider.GetMetadataForBlobs(ctx, keys)
 	if err != nil {
-		return nil, api.NewErrorInternal(fmt.Sprintf(
-			"error fetching metadata for blob, check if blob exists and is assigned to this relay: %v", err))
+		if strings.Contains(err.Error(), blobstore.ErrMetadataNotFound.Error()) {
```
dmanc, line 316: similar question, shouldn't we check for some error like "ChunkNotFound"?

relay/server.go
cody-littley: I looked at the code, and it looks like `ErrBlobNotFound` cannot be returned here, since this is only fetching metadata. I did a search for uses of `ErrBlobNotFound`, and made those places return the correct error code though.

relay/server.go
cody-littley: Still need to do this, will push another commit

cody-littley: Hmm... I'm actually not sure what strings are returned when there is a "not found" error for chunk data. Would it be ok to handle this in a future PR?

PR #1823, opened 2025-07-31
feat: litt unlock command

cody-littley:

Adds the ability to override LittDB lock files in the advent of an emergency.
https://linear.app/eigenlabs/issue/EGDA-1795/flag-to-ignore-littdb-lock-files


node/config.go
```diff
@@ -136,6 +136,10 @@ type Config struct {
 	// Directories do not need to be on the same filesystem.
 	LittDBStoragePaths []string
 
+	// If true, then purge LittDB locks on startup. Potentially useful to get rid of zombie lock files,
```
dmanc, line 139: Didn't quite understand the scenario in which zombie lock files appear? Is it due to some sort of ungraceful termination with containers?

litt/cli/litt_cli.go
```diff
@@ -267,6 +277,15 @@ func buildCLIParser(logger logging.Logger) *cli.App {
 				},
 				Action: nil, // syncCommand, // TODO this will be added in a follow up PR
 			},
+			{
+				Name:      "unlock",
+				Usage:     "Manually delete LittDB lock files. Dangerous if used improperly, use with caution.",
```
dmanc, line 287: Should we add some sort of confirmation flow where the user has to say "yes" and if they want to skip there's the `force-unlock` or `--force` flag for skipping that flow?

litt/test/unlock_test.go
```diff
@@ -0,0 +1,163 @@
...
+	for i := 0; i < 100; i++ {
+		key := rand.PrintableBytes(32)
+		value := rand.PrintableVariableBytes(1, 100)
+
+		expectedData[string(key)] = value
+		err = table.Put(key, value)
+		require.NoError(t, err, "Failed to put data in table")
+	}
+
+	// Look for lock files. We should see one for each volume.
```
litt3, line 49: you claim `We should see one for each volume` here, but you never assert it

node/flags/flags.go
```diff
@@ -451,6 +451,12 @@ var (
 		Required: false,
 		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "LITT_DB_STORAGE_PATHS"),
 	}
+	LittUnsafePurgeLocksFlag = cli.BoolFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "litt-unsafe-purge-locks"),
+		Usage:    "Unsafe flag to purge locks in LittDB. Use with caution, as it may lead to data loss or corruption.",
+		Required: false,
+		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "LITT_UNSAFE_PURGE_LOCKS"),
```
litt3, line 458: Not something that needs to happen in this PR- but it would be nice to have a method which transforms a flag name into an env var name, so there is no possibility for mismatch

node/flags/flags.go
```diff
@@ -630,6 +636,7 @@ var optionalFlags = []cli.Flag{
 	EigenDADirectoryFlag,
 	BlsOperatorStateRetrieverFlag,
 	EigenDAServiceManagerFlag,
+	LevelDBDisableSeeksCompactionV1Flag,
```
litt3: I don't think this is what you intended to put here

litt3 (dismissed):

node/config.go
cody-littley: Expanded the documentation here:

```
	// If true, then purge LittDB locks on startup. Potentially useful to get rid of zombie lock files,
	// but also dangerous (multiple LittDB processes operating on the same files can lead to data corruption).
	//
	// When LittDB starts up, it attempts to create lock files. When a validator is forcefully shut down, lock files 
	// may be left behind. At startup time, if LittDB observes existing lock files, it first checks to see
	// if the process that created the lock files is still running. The lock files contain the creator's PID, and so 
	// LittDB checks to see if there is any process with that PID still running.
	//
	// Although it should be rare, it's possible that another process may be started with the same PID as the
	// PID used to create the lock files. When this happens, LittDB will be prevented from starting up out of
	// fear of another process trying to access the same files, even though the original process that created the 
	// lock files is no longer running. If that happens, this flag is a safe way to force LittDB to start up
	// without being blocked by those lock files. BE VERY CERTAIN THAT THE OTHER PROCESS IS ACTUALLY DEAD!
	// If two instances of LittDB are running on the same files, it WILL lead to data corruption.
	//
	// An alternate way to clear the LittDB lock files is via the LittDB CLI with the "litt unlock" command.
	// Run "litt unlock --help" for more information.
	LittUnsafePurgeLocks bool
```

litt/cli/litt_cli.go
cody-littley: Added. You either must type `I know what I am doing`, or include a `--force` tag.

litt/test/unlock_test.go
cody-littley: added `require.Equal(t, 3, lockFileCount)`

node/flags/flags.go
cody-littley: Agree.

node/flags/flags.go
cody-littley: good catch, fixed

litt/test/unlock_test.go
```diff
@@ -55,11 +57,13 @@ func TestUnlock(t *testing.T) {
 			return nil
 		}
 		if strings.HasSuffix(path, util.LockfileName) {
+			fmt.Printf("Found lockfile %s\n", path) // TODO
```
litt3: todo

litt3 (dismissed):

cody-littley:
@pschork 

> deleting zombie locks that prevent the node from starting
> allowing operators to backup littdb

This change is not needed for operators to back up LittDB. I'm currently merging a series of PRs that add this functionality, this change is tangential to that one.

> In the case of 1, is it possible for the operator to just manually delete the zombie lock file?

Yes, it's completely possible to do this manually without this tool. The reason why I automate this process is that the lock file is buried in the LittDB file structure, and there may be multiple lock files that need to be deleted. If the user configures LittDB to store its data in `N` physical volumes, then there will be one lockfile for each physical volume.

litt/test/unlock_test.go
cody-littley: removed

dmanc (dismissed):

litt3 (approved):

PR #1826, opened 2025-08-01
refactor: fft errors and fct docs

samlaf:

Main refactor here is deleting the misused sentinel errors in fft package, and adding better docs to describe its returned errors. This is needed to cleanup the EncodedPayload manipulation functions (we need to understand precisely the errors being returned).
Also included lots of small miscellaneous cleanups.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


encoding/fft/fft.go
```diff
@@ -93,6 +94,10 @@ func NewFFTSettings(maxScale uint8) *FFTSettings {
 
 // FFTSettingsFromBlobLengthSymbols accepts a blob length, and returns a new instance of FFT settings
 func FFTSettingsFromBlobLengthSymbols(blobLengthSymbols uint32) *FFTSettings {
+	isPowerOf2 := blobLengthSymbols != 0 && (blobLengthSymbols&(blobLengthSymbols-1)) == 0
```
litt3:
```
// IsPowerOfTwo returns true if the provided integer v is a power of 2.
func IsPowerOfTwo(v uint64) bool {
	return (v&(v-1) == 0) && (v != 0)
}
```

This logic already exists as a utility function

encoding/fft/fft.go
```diff
@@ -93,6 +94,10 @@ func NewFFTSettings(maxScale uint8) *FFTSettings {
...
+	if !isPowerOf2 {
+		panic(fmt.Sprintf("blobLengthSymbols %d is not a power of 2", blobLengthSymbols))
```
litt3: Is there a good reason that this function shouldn't return an error, as opposed to panicking?

It's ok if you want to leave as-is to keep scope small.

encoding/fft/zero_poly.go
```diff
@@ -95,32 +95,30 @@ func padPoly(out []fr.Element, poly []fr.Element) {
 // The scratch space must be at least 3 times the output space.
 // The output must have a power of 2 length.
 // The input polynomials must not be empty, and sum to no larger than the output.
+//
+// TODO(samlaf): most errors returned should be panicing instead. But I'm unsure of the consequences of changing this.
```
litt3: Why should they be panicking?

encoding/fft/fft.go
samlaf: New error handling philosophy I'm trying to slowly roll into our codebase.

The idea is that I don't really like how error treats all errors the same. IMO there are errors and exceptions, and exceptions are things that can be handled, retried, etc., whereas errors are bugs in the implementation that can't be solved other than to bubble up the error to stderr and let a human know.

I think for those things we should be using panics, because it does precisely what we want! bubbles up to the human with a nice stack trace, as opposed to a weirdly hand-constructed error wrapping stack-trace-wannabe.
Plus the nice thing is that it keeps the function interface much cleaner and easier to reason about. Reading our functions in the payloadEncoder/FFT stuff, there's just so many different errors that can be returned that it's impossible to know what represents a 4xx, a 5xx, what is just logically impossible because I've already checked those conditions in the wrapping function, etc.

My main take is that panics reduce the error surface to a set of issues that the caller of a function would want to handle (either via retries, or trying different argument, etc).

encoding/fft/zero_poly.go
samlaf: See my reasoning in https://github.com/Layr-Labs/eigenda/pull/1826#discussion_r2252298615

encoding/fft/fft.go
bxue-l2: Maybe add a comment to this function

encoding/fft/zero_poly.go
```diff
@@ -41,8 +42,7 @@ type ZeroPolyFn func(missingIndices []uint64, length uint64) ([]fr.Element, []fr
 
 func (fs *FFTSettings) makeZeroPolyMulLeaf(dst []fr.Element, indices []uint64, domainStride uint64) error {
 	if len(dst) < len(indices)+1 {
-		log.Printf("expected bigger destination length: %d, got: %d", len(indices)+1, len(dst))
-		return ErrInvalidDestinationLength
+		return fmt.Errorf("expected bigger destination length: %d, got: %d", len(indices)+1, len(dst))
```
bxue-l2, line 45: why removing ErrInvalidDestinationLength error? unnecessary?

encoding/fft/zero_poly.go
```diff
@@ -95,32 +95,30 @@ func padPoly(out []fr.Element, poly []fr.Element) {
...
 // The output must have a power of 2 length.
 // The input polynomials must not be empty, and sum to no larger than the output.
+//
+// TODO(samlaf): most errors returned should be panicing instead. But I'm unsure of the consequences of changing this.
 func (fs *FFTSettings) reduceLeaves(scratch []fr.Element, dst []fr.Element, ps [][]fr.Element) ([]fr.Element, error) {
 	n := uint64(len(dst))
 	if !IsPowerOfTwo(n) {
-		log.Println("destination must be a power of two")
-		return nil, ErrDestNotPowerOfTwo
+		return nil, fmt.Errorf("destination must be a power of two, got %d", n)
```
bxue-l2, line 101: This is MIT code from protolambda, why not use your new error type, NewFFTInputNotPowerOfTwoError

bxue-l2: Or is it because it is kind of some destination error?

encoding/fft/fft.go
litt3: I'm a bit skeptical of this philosophy.

Using `panic` can turn a transitory problem into a fatal one. Maybe there's a race condition that causes a bug to happen only occasionally. Or maybe a malicious party figures out how to trigger this case for someone else. In such cases, burning everything down because something unexpected happened is the wrong choice. Being able to limp along after encountering a bug can prevent a disaster. 

Right now this code is only being used by clients, so the danger of these edge cases isn't too high... but it's likely that this FFT stuff will be used on full nodes eventually. In any case, if you're trying to apply this more broadly, we need to consider the implications of this being done in more critical areas of the code.

Could you make a `CriticalError` type, which would decrease the number of error types being returned in the blob conversion chain? I don't think we need a unique error type for every failure mode. This solution wouldn't give the benefit of simplified function interfaces, but would solve the problem of not immediately knowing how handle a returned error.

I blame golang for not having proper throwable exceptions... but I'm not sold on using `panic` as a replacement

samlaf: > limp along after encountering a bug can prevent a disaster.

What kind of limp alone are you expecting? The kind of errors that are panic'ing here are breaking invariants, there's literally no one to "limp" with them.

> I blame golang for not having proper throwable exceptions... but I'm not sold on using panic as a replacement

What do you see as a difference? panics can be caught the same way as throwable exceptions. In fact this is done in the json library.

cody-littley: My main complaint with `panic()` as a general concept is that golang makes it very tricky to intercept them. I agree with the philosophy that some errors should result in an immediate shutdown. But even when we want software to immediately shut down, it's often quite useful to implement special procedures as things are being torn down. For example, logging diagnostics, flushing files out to disk, triggering alerts/metrics, etc. Use of the `panic()` keyword makes this sort of thing hard to implement.

encoding/fft/zero_poly.go
samlaf: My reasoning was that the errors are uninformative as they don't contain any value wrappings. Also imo sentinel values should only be defined if they are actually caught/used somewhere, but these aren't. Doesn't feel useful. 

I also had in mind to replace all of them with panics but after our discussion I won't be doing that.

encoding/fft/zero_poly.go
samlaf: Because it's an internal error that represents a bug, not something that should be surfaced to the user using a structured error.

encoding/fft/fft.go
samlaf: [5f77329](https://github.com/Layr-Labs/eigenda/pull/1826/commits/5f77329ee8e8df40d7466342fbea36c75037f963)

encoding/fft/fft.go
samlaf: Thanks for the discussion! Summary of the few points we discussed offline:
1. shouldn't panic in public functions because we don't know how they'll be used, and could be used dynamically and on a per-request basis and end up crashing an entire server for a single bad request (like our current encoding logic with srs tables missing on holesky)
2. fine to panic in few places to enforce very strong and important invariants
3. we could define some `CriticalError`s/`UnrecoverableError`s as @litt3 suggested. They are imo very similar to panics, in that they will be parsed very similarly. But they are more targeted and we are forced to think about them more critically (panic'ing forces the same kind of sloppiness that I don't like about golang errors actually...)
4. Agree with @cody-littley's point, although as I just mentioned, I think the concept of graceful shutdown is orthogonal to the error signaling techniques we use. And afaiu we don't have graceful shutdown anywhere (yet)?

Reverted the specific panic pointed here: [e2d5879](https://github.com/Layr-Labs/eigenda/pull/1826/commits/e2d5879e67ede314732d65be21271ee2f7f959bf)

encoding/fft/zero_poly.go
samlaf: Removed this comment after our discussion: [7c233bc](https://github.com/Layr-Labs/eigenda/pull/1826/commits/7c233bcea35d7675c373b406dd71c5ffe74da547)

encoding/fft/zero_poly.go
bxue-l2: sg

bxue-l2 (dismissed):
lgtm

PR #1834, opened 2025-08-04
feat: LittDB table info command

cody-littley:

Adds an implementation for the `litt table-info` command.

To test locally, first generate a LittDB table via `litt benchmark`. Then, run the following:

`litt table-info -s ~/benchmark/volume1 -s ~/benchmark/volume2 -s ~/benchmark/volume3 benchmark`
```
Aug  4 08:16:13.542 INF cli/table_info.go:76 Table:                       benchmark
Aug  4 08:16:13.542 INF cli/table_info.go:77 Key count:                   1,309
Aug  4 08:16:13.542 INF cli/table_info.go:78 Size:                        2.56 GiB
Aug  4 08:16:13.542 INF cli/table_info.go:79 Is snapshot:                 false
Aug  4 08:16:13.542 INF cli/table_info.go:80 Oldest segment age:          6.93 days
Aug  4 08:16:13.542 INF cli/table_info.go:81 Oldest segment seal time:    2025-07-28T09:51:56-05:00
Aug  4 08:16:13.542 INF cli/table_info.go:82 Newest segment age:          405.19 ms
Aug  4 08:16:13.542 INF cli/table_info.go:83 Newest segment seal time:    2025-08-04T08:16:13-05:00
Aug  4 08:16:13.542 INF cli/table_info.go:84 Segment span:                6.93 days
Aug  4 08:16:13.542 INF cli/table_info.go:85 Lowest segment index:        0
Aug  4 08:16:13.542 INF cli/table_info.go:86 Highest segment index:       1313
Aug  4 08:16:13.542 INF cli/table_info.go:87 Key map type:                LevelDBKeymap

```

### DIFF
--- a/litt/cli/table_info.go
+++ b/litt/cli/table_info.go
@@ -26,7 +26,7 @@ type TableInfo struct {
 	IsSnapshot bool
 	// The time when the oldest segment was sealed.
 	OldestSegmentSealTime time.Time
-	// The time when the newest segment was sealed. This is used to determine the age of the oldest segment.
+	// The time when the newest segment was sealed.
 	NewestSegmentSealTime time.Time
 	// The index of the oldest segment in the table.
 	LowestSegmentIndex uint32
@@ -124,7 +124,7 @@ func tableInfo(logger logging.Logger, tableName string, paths []string, fsync bo
 		segmentPaths,
 		false,
 		time.Now(),
-		true,
+		false,
 		fsync)
 
 	if err != nil {
--- a/litt/cli/table_info.go
+++ b/litt/cli/table_info.go
@@ -1,0 +1,199 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"path"
+	"time"
+
+	"github.com/Layr-Labs/eigenda/common"
+	"github.com/Layr-Labs/eigenda/litt"
+	"github.com/Layr-Labs/eigenda/litt/disktable"
+	"github.com/Layr-Labs/eigenda/litt/disktable/segment"
+	"github.com/Layr-Labs/eigenda/litt/littbuilder"
+	"github.com/Layr-Labs/eigenda/litt/util"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	"github.com/urfave/cli/v2"
+)
+
+// TableInfo contains high level information about a table in LittDB.
+type TableInfo struct {
+	// The number of key-value pairs in the table.
+	KeyCount uint64
+	// The size of the table in bytes.
+	Size uint64
+	// If true, the table at the specified path is a snapshot of another table.
+	IsSnapshot bool
+	// The time when the oldest segment was sealed.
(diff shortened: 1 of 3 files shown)


litt/cli/table_info.go
```diff
@@ -0,0 +1,199 @@
...
+
+	errorMonitor := util.NewErrorMonitor(context.Background(), logger, nil)
+
+	lowestSegmentIndex, highestSegmentIndex, segments, err := segment.GatherSegmentFiles(
+		logger,
+		errorMonitor,
+		segmentPaths,
+		false,
+		time.Now(),
+		true,
```
litt3: You're cleaning orphans here, but it's not documented anywhere. Is it reasonable that running a `table info` command would be making modifications to the table?

Are there any other modifications being made, inherent to the logic of the utilized methods?

litt/cli/table_info.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	isSnapshot, err := segments[lowestSegmentIndex].IsSnapshot()
+	if err != nil {
+		return nil, fmt.Errorf("failed to check if segment %d is a snapshot: %w", lowestSegmentIndex, err)
+	}
+
+	if isSnapshot {
+		if len(paths) != 1 {
+			return nil, fmt.Errorf("table %s is a snapshot, but multiple paths were provided: %v",
```
litt3, line 151: I don't follow the logic here. Why does being a snapshot mean only 1 path can be provided?

litt/cli/table_info.go
```diff
@@ -0,0 +1,199 @@
...
+type TableInfo struct {
+	// The number of key-value pairs in the table.
+	KeyCount uint64
+	// The size of the table in bytes.
+	Size uint64
+	// If true, the table at the specified path is a snapshot of another table.
+	IsSnapshot bool
+	// The time when the oldest segment was sealed.
+	OldestSegmentSealTime time.Time
+	// The time when the newest segment was sealed. This is used to determine the age of the oldest segment.
```
litt3: This doc doesn't seem to match the implementation. The logged age of the oldest segment is not influenced by the `NewestSegmentSealTime`

litt3 (dismissed):

litt/cli/table_info.go
cody-littley: Good catch, this should be false. Fixed.

litt/cli/table_info.go
cody-littley: None of the LittDB code/tooling will ever split a snapshot across multiple directories. So if the command is asked to investigate multiple directories and one contains symlinks (i.e. it's a snapshot), then either the command is being misused or there is some sort of data corruption.

litt/cli/table_info.go
cody-littley: I honestly can't remember the purpose of the second part of the comment, it doesn't make any sense. I wonder if copilot snuck without me noticing?

fixed

PR #1840, opened 2025-08-05
perf: larger memory safety buffer

cody-littley:

Increase the default size of the memory safety buffer, and allow it to be configured as a percentage of the total memory.

### DIFF
--- a/node/config.go
+++ b/node/config.go
@@ -123,6 +123,10 @@ type Config struct {
 	// this config value overrides the LittDBWriteCacheSizeFraction value.
 	LittDBWriteCacheSizeGB float64
 
+	// This is a derived value based on the configuration provided by LittDBWriteCacheSizeGB and
+	// LittDBWriteCacheSizeFraction. Stores the size of the LittDB write cache in bytes.
+	littDBWriteCacheSize uint64
+
 	// The percentage of the total memory to use for the read cache in littDB as a fraction of 1.0, where 1.0
 	// means that all available memory will be used for the read cache (don't actually use 1.0, that leaves no buffer
 	// for other stuff). Ignored if LittDBReadCacheSizeGB is set.
@@ -131,6 +135,10 @@ type Config struct {
 	// The size of the cache for storing recently read chunks in littDB, in gigabytes. Ignored if 0. If set,
 	// this config value overrides the LittDBReadCacheSizeFraction value.
 	LittDBReadCacheSizeGB float64
+
+	// This is a derived value based on the configuration provided by LittDBReadCacheSizeGB and
+	// LittDBReadCacheSizeFraction. Stores the size of the LittDB read cache in bytes.
+	littDBReadCacheSize uint64
 
 	// The list of paths to the littDB storage directories. Data is spread across these directories.
 	// Directories do not need to be on the same filesystem.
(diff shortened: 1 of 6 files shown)


node/config.go
```diff
@@ -123,6 +123,10 @@ type Config struct {
 	// this config value overrides the LittDBWriteCacheSizeFraction value.
 	LittDBWriteCacheSizeGB float64
 
+	// This is a derived value based on the configuration provided by LittDBWriteCacheSizeGB and
+	// LittDBWriteCacheSizeFraction. Stores the size of the LittDB write cache in bytes.
+	littDBWriteCacheSize uint64
```
litt3: Why do you need this value stored as `LittDBWriteCacheSizeGB` _and_ as bytes? Looking at the actual usage, it seems like this can be accomplished with a single field in this config. Unless I'm missing something

node/config.go
```diff
@@ -132,6 +136,10 @@ type Config struct {
 	// this config value overrides the LittDBReadCacheSizeFraction value.
 	LittDBReadCacheSizeGB float64
 
+	// This is a derived value based on the configuration provided by LittDBReadCacheSizeGB and
+	// LittDBReadCacheSizeFraction. Stores the size of the LittDB read cache in bytes.
+	littDBReadCacheSize uint64
```
litt3: Same question here, why bytes and gigabytes?

node/node.go
```diff
@@ -323,6 +331,84 @@ func NewNode(
...
+		err = memory.SetGCMemorySafetyBuffer(safetyBufferSize)
+		if err != nil {
+			return fmt.Errorf("failed to set memory limit: %w", err)
+		}
+	}
+	totalAllocated += safetyBufferSize
+
+	if config.EnableV2 {
+		var readCacheSize uint64
+		if config.LittDBWriteCacheSizeGB > 0 {
```
litt3: You have your fields mixed up, setting `readCacheSize` if `LittDBWriteCacheSizeGB` > 0, and mixing fields in logs.

Consider asking claude to do a deep logic review, to catch these kinds of mismatches

node/config.go
```diff
@@ -419,5 +433,6 @@ func NewConfig(ctx *cli.Context) (*Config, error) {
 		GetChunksColdCacheReadLimitMB:       ctx.GlobalFloat64(flags.GetChunksColdCacheReadLimitMBFlag.Name),
 		GetChunksColdBurstLimitMB:           ctx.GlobalFloat64(flags.GetChunksColdBurstLimitMBFlag.Name),
 		GCSafetyBufferSizeGB:                ctx.GlobalFloat64(flags.GCSafetyBufferSizeGBFlag.Name),
+		GCSafetyBufferSizeFraction:          ctx.GlobalFloat64(flags.GCSafetyBufferSizeFractionFlag.Name),
```
litt3: nit: no validation that this is between 0 and 1 (also applies to preexising fraction fields)

node/node.go
```diff
@@ -323,6 +331,84 @@ func NewNode(
...
+		//  limit here. It's important to ensure that all reserved memory buckets sum to less than the maximum
+		//  available memory.
+	}
+
+	if totalAllocated > maxMemory {
+		return fmt.Errorf("total memory allocated (%d bytes) "+
+			"exceeds maximum available memory (%d bytes)", totalAllocated, maxMemory)
+	}
+
+	bytesRemaining := totalAllocated - safetyBufferSize
```
litt3: Shouldn't this be `bytesRemaining := maxMemory - totalAllocated`?

node/flags/flags.go
```diff
@@ -418,7 +418,7 @@ var (
 	LittDBWriteCacheSizeGBFlag = cli.IntFlag{
 		Name: common.PrefixFlag(FlagPrefix, "litt-db-write-cache-size-gb"),
 		Usage: "The size of the LittDB write cache in gigabytes. Overrides " +
-			"LITT_DB_WRITE_CACHE_SIZE_FRACTION if > 0, otherwise is ignored.",
+			"NODE_LITT_DB_WRITE_CACHE_SIZE_FRACTION if > 0, otherwise is ignored.",
```
litt3, line 421: For all of these sizes, are you intentionally using int flags, even though the config has these fields as floats?

cody-littley: Changed to float64 flags.

node/node.go
cody-littley: You are correct. Fixed.

node/config.go
cody-littley: added validation

node/config.go
cody-littley: let's discuss this on a call

node/node.go
cody-littley: You are right, these fields are a mess. I refactored this code to use a utility function, since the pattern is the same in each instance. Now it's very easy to verify that I use the correct config fields in the correct places.

node/config.go
cody-littley: made change as we discussed

node/config.go
cody-littley: made change as we discussed

litt3 (approved):

PR #1841, opened 2025-08-05
refactor!: encoded payload meaning

samlaf:
Closes EGDA-1800

Ready for first round of review. Want to get some feedback on general approach, before I put in more time.
Ended up moving quite a bit of stuff around... can break up if PR is too large to review, lmk.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/clients/v2/coretypes/blob.go
```diff
@@ -7,50 +7,72 @@ import (
...
+		// If the coefficients are not a power of two, we pad them to the next power of two.
+		blobLengthSymbols := uint32(encoding.NextPowerOf2(len(coefficients)))
+		paddedCoefficients = make([]fr.Element, blobLengthSymbols)
+		copy(paddedCoefficients, coefficients)
+	}
 	return &Blob{
-		coeffPolynomial:   coeffPolynomial,
-		blobLengthSymbols: blobLengthSymbols,
-	}, nil
+		coeffPolynomial: paddedCoefficients,
```
litt3: If the passed in coefficients *are* a power of two, then the `coeffPolynomial` in the returned blob shares the underlying array of the input `coefficients`

This was already the case before your refactor, not something you introduced. But if we decide this is ok, it should probably *at least* have a comment making this explicit.

api/clients/v2/coretypes/encoded_payload.go
```diff
@@ -27,29 +29,14 @@ func (ep *EncodedPayload) Serialize() []byte {
...
-		uint32(len(payloadBytes))) // uint32 should be more than enough to store the length (approx 4gb)
-
-	// encode payload modulo bn254, and align to 32 bytes
-	encodedData := codec.PadPayload(payloadBytes)
-	encodedPayloadBytes := append(encodedPayloadHeader, encodedData...)
-
-	return &EncodedPayload{encodedPayloadBytes}
+// LenSymbols returns the number of symbols in the encoded payload
+func (ep *EncodedPayload) LenSymbols() uint32 {
+	return uint32(len(ep.bytes)+encoding.BYTES_PER_SYMBOL-1) / encoding.BYTES_PER_SYMBOL
```
litt3: What's this logic for? Doesn't the `bytes` field already have the encoded payload header in it? Seems like the `LenSymbols` would just be len(bytes) / BYTES_PER_SYMBOL?

api/clients/v2/coretypes/encoded_payload.go
```diff
@@ -77,91 +63,51 @@ func (ep *EncodedPayload) Decode() (*Payload, error) {
 			unpaddedDataLength, claimedLength)
 	}
 
-	// unpadded data length can be slightly bigger than the claimed length, since RemoveInternalPadding doesn't
-	// do anything to remove trailing zeros that may have been added when the data was initially padded.
-	// however, this extra padding shouldn't exceed 31 bytes, because that's the most that would be added
-	// when padding the data length to 32 bytes. If this error occurs, that means there must be a flaw in the logic at
-	// construction time (or someone was bad and didn't use the proper construction methods)
-	if unpaddedDataLength > claimedLength+31 {
```
litt3, line 85: Why are you removing this `unpaddedDataLength > claimedLength+31` check?

api/clients/v2/coretypes/encoded_payload_test.go
```diff
@@ -1,23 +1,23 @@
...
 
 // TestDecodeShortBytes checks that an encoded payload with a length less than claimed length fails at decode time
-func TestDecodeShortBytes(t *testing.T) {
+func TestDecodeInvalidPayloadLen(t *testing.T) {
 	testRandom := random.NewTestRandom()
 	originalData := testRandom.Bytes(testRandom.Intn(1024) + 33)
-	encodedPayload := newEncodedPayload(NewPayload(originalData))
+	encodedPayload := Payload(originalData).ToEncodedPayload()
 
 	// truncate
```
litt3: Comment doesn't match anymore- you're increasing claimed length, as opposed to truncating

api/clients/v2/coretypes/encoded_payload_test.go
```diff
@@ -38,80 +38,20 @@ func TestDecodeLongBytes(t *testing.T) {
...
+	truncatedBlob := BlobFromCoefficients(truncatedCoefficients)
 
+	reconstructedEncodedPayload := truncatedBlob.ToEncodedPayload(codecs.PolynomialFormCoeff)
 	// even though the actual length will be less than the claimed length, we shouldn't see any error
-	reconstructedEncodedPayload, err := encodedPayloadFromElements(
-		originalFieldElements,
-		uint32(len(originalFieldElements))*encoding.BYTES_PER_SYMBOL)
-	require.NoError(t, err)
-	require.NotNil(t, reconstructedEncodedPayload)
+	require.Equal(t, encodedPayload, reconstructedEncodedPayload)
```
litt3: What are you testing here? Here's how I'm understanding the test:
1. create encoded payload
2. turn it into a blob
3. truncate the blob elements
4. convert the blob with truncated elements back into an encoded payload
5. assert that the original encoded payload is equal to the `reconstructedEncodedPayload` (❓)

But I don't understand why these should be equal? Shouldn't the truncation have caused the `reconstructedEncodedPayload` to be different from the original?

api/clients/v2/coretypes/payload.go
```diff
@@ -1,78 +1,47 @@
...
-// ToBlob converts the Payload bytes into a Blob
-//
-// The payloadForm indicates how payloads are interpreted. The form of a payload dictates what conversion, if any, must
-// be performed when creating a blob from the payload.
-func (p *Payload) ToBlob(payloadForm codecs.PolynomialForm) (*Blob, error) {
-	encodedPayload := newEncodedPayload(p)
+	// Calculate the length of the EncodedPayload in symbols (including the header) which has to be a power of 2.
+	encodedDataLenSymbols := uint32(len(encodedData)+encoding.BYTES_PER_SYMBOL-1) / encoding.BYTES_PER_SYMBOL
+	encodedHeaderAndDataLenSymbols := codec.PayloadHeaderSizeSymbols + encodedDataLenSymbols
+	encodedPayloadLenSymbols := encoding.NextPowerOf2(encodedHeaderAndDataLenSymbols)
```
litt3, line 22: Given the redefinition of an encoded payload (always power of 2 length):

1. encoded payload doc needs to be updated to reflect the change
2. would it make sense to create an encoded payload constructor, which checks that the input bytes are a proper power of 2?

encoding/fft/fft.go
```diff
@@ -36,6 +36,11 @@ import (
 	"math/bits"
 )
 
+// IsPowerOfTwo returns true if the provided integer v is a power of 2.
+func IsPowerOfTwo(v uint64) bool {
```
litt3: Now you have a duplicate of the preexisting method... probably should either use the preexisting method, or move the preexisting method to a better spot if it cannot be used due to import cycles

litt3 (commented):
Broad strokes, seems like a good direction to move in.

I haven't done a full review yet, there's a lot of implications to puzzle through... I can take another pass once we talk through some of these existing things

api/clients/v2/coretypes/encoded_payload.go
samlaf: Agree I wrote it as if the bytes could not align with BYTES_PER_SYMBOL, but they always should.

encoding/fft/fft.go
samlaf: lol fml this function was already there in fft, just private. I'll make it use ours.
Personally find it weird that this function is under `encoding` though... `common/math` feels like a better fit?

samlaf:
Apologies rebased on top of https://github.com/Layr-Labs/eigenda/pull/1826, and then merged it to master, and then merged master into here. Should have only merged master into here but I was impatient.

api/clients/v2/coretypes/blob.go
```diff
@@ -7,50 +7,75 @@ import (
...
 	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
 )
 
 // Blob is data that is dispersed on eigenDA.
 //
-// A Blob is represented under the hood by an array of field elements, which represent a polynomial in coefficient form
+// A Blob is represented under the hood by an array of field elements (symbols),
+// which represent a polynomial in coefficient form.
+// A Blob must have a length (in symbols) that is a power of two. In particular, blobs of length 0 are not allowed.
+// A Blob's length must match the blobLength in the BlobHeader's [encoding.BlobCommitments.Length].
```
bxue-l2, line 17: which is also measured by number of symbol

api/clients/v2/coretypes/blob.go
```diff
@@ -62,81 +87,61 @@ func (b *Blob) Serialize() []byte {
...
-		payloadElements = b.coeffPolynomial
+		encodedPayloadElements = b.coeffPolynomial
 	case codecs.PolynomialFormEval:
 		// the payload is interpreted as evaluations of the polynomial, so the coefficient representation contained
 		// in the blob must be converted to the evaluation form
-		payloadElements, err = b.toEvalPoly()
+		encodedPayloadElements, err = b.toEvalPoly()
 		if err != nil {
-			return nil, fmt.Errorf("compute eval poly: %w", err)
+			return nil, fmt.Errorf("convert blob to eval poly: %w", err)
```
bxue-l2: can we panic on error, not in the FFT library, but here? In a secure integration, it is wrong for L2 consensus node to proceed, because that would lead to a system posting incorrect L2 state. 

It is like your database example, keeping it running will only corrupt the system.

bxue-l2: And this function should not return error. Because within this logic, we know for sure Blob has power of 2 number of symbols.

bxue-l2: The only other place where it returns error is 

> if n > fs.MaxWidth {
> 	return nil, fmt.Errorf("got %d values but only have %d roots of unity", n, fs.MaxWidth)
> }

however, this is just impossible at this level of context, so suggest panic. see
	

> fftSettings, err := fft.FFTSettingsFromBlobLengthSymbols(uint32(len(b.coeffPolynomial)))
> 	if err != nil {
> 		return nil, fmt.Errorf("create FFT settings from blob length symbols: %w", err)
> 	}
> 
> 	// the FFT method pads to the next power of 2, so we don't need to do that manually
> 	fftedElements, err := fftSettings.FFT(b.coeffPolynomial, false)
> 	if err != nil {
> 		return nil, fmt.Errorf("perform FFT on blob coefficients: %w", err)
> 	}

api/clients/v2/coretypes/blob.go
```diff
@@ -62,81 +87,61 @@ func (b *Blob) Serialize() []byte {
...
+	fftSettings, err := fft.FFTSettingsFromBlobLengthSymbols(uint32(len(b.coeffPolynomial)))
 	if err != nil {
 		return nil, fmt.Errorf("create FFT settings from blob length symbols: %w", err)
 	}
 
 	// the FFT method pads to the next power of 2, so we don't need to do that manually
 	fftedElements, err := fftSettings.FFT(b.coeffPolynomial, false)
 	if err != nil {
-		return nil, fmt.Errorf("perform FFT: %w", err)
+		return nil, fmt.Errorf("perform FFT on blob coefficients: %w", err)
```
bxue-l2: should eventually panic

api/clients/v2/coretypes/blob.go
```diff
@@ -62,81 +87,61 @@ func (b *Blob) Serialize() []byte {
...
-	return encodedPayload, nil
 }
 
 // toEvalPoly converts a blob's coeffPoly to an evalPoly, using the FFT operation
 func (b *Blob) toEvalPoly() ([]fr.Element, error) {
 	// TODO (litt3): this could conceivably be optimized, so that multiple objects share an instance of FFTSettings,
 	//  which has enough roots of unity for general use. If the following construction of FFTSettings ever proves
 	//  to present a computational burden, consider making this change.
-	fftSettings, err := fft.FFTSettingsFromBlobLengthSymbols(b.blobLengthSymbols)
+	fftSettings, err := fft.FFTSettingsFromBlobLengthSymbols(uint32(len(b.coeffPolynomial)))
```
bxue-l2: this should eventually panic

api/clients/v2/coretypes/payload.go
samlaf: Let me know what you think of [b49315b](https://github.com/Layr-Labs/eigenda/pull/1841/commits/b49315bf2cad41636a56591de4ef44ccacd6f1e7)

Don't think we need another constructor. I mentioned in the docs that people should use `Payload.ToEncodedPayload` for that.

api/clients/v2/coretypes/blob.go
samlaf: [c6ff91a](https://github.com/Layr-Labs/eigenda/pull/1841/commits/c6ff91ae62dd7f6a7b2efd96ebd0ff3d54720199)

api/clients/v2/coretypes/blob.go
samlaf: eventually?

api/clients/v2/coretypes/encoded_payload.go
samlaf: [9324c84](https://github.com/Layr-Labs/eigenda/pull/1841/commits/9324c84a682d73ffa39dfd49bd24c0137954830f)

api/clients/v2/coretypes/encoded_payload_test.go
samlaf: [ae94cb5](https://github.com/Layr-Labs/eigenda/pull/1841/commits/ae94cb5a17ea6c4290f9983dfcd11ae7b570abc2)

api/clients/v2/coretypes/encoded_payload.go
samlaf: Hmm it was failing in a test and I assumed it was now wrong because EncodedPayload has a power of 2. But I will admit I don't even understand how this check was working in the previous code.

If you have a blob (power of 2), and you unpad it, you can still end up with a bunch of 0s at the end right? How does RemoveInternalPadding know to remove them if it doesn't have access to `claimedLength`?

api/clients/v2/coretypes/blob.go
bxue-l2: I think [here](https://github.com/Layr-Labs/eigenda/pull/1841/files/ae94cb5a17ea6c4290f9983dfcd11ae7b570abc2#diff-a422850d044b4bb3896365a7cb7496555aa51a405064a5906325974755cc6cd6R120)

api/clients/v2/coretypes/encoded_payload_test.go
samlaf: I will need to get back to you on this one, tired after doing this test refactor for the EncodedPayload tests: [fef20e4](https://github.com/Layr-Labs/eigenda/pull/1841/commits/fef20e478dc26372a8cbe1ff3697aa8f6690c0d1)

But my intuition for the reason why the test above works is only because its in coefficient form: the last elements of the blob are just never read (because there's no FFT going on like in eval form), so its fine to modify them to whatever, since they are past the actual payload elements that we need to read.

api/clients/v2/coretypes/blob.go
samlaf: Not fully sure I'm following your first example, but I think [9e40fc6](https://github.com/Layr-Labs/eigenda/pull/1841/commits/9e40fc6f0516c5cda4b2afd9b2c4facbb767ea89) solves your issue. cc @litt3 for thoughts since we talked about panics yesterday

api/clients/v2/coretypes/blob.go
samlaf: https://github.com/Layr-Labs/eigenda/pull/1841#discussion_r2256631423

api/clients/v2/coretypes/blob.go
samlaf: https://github.com/Layr-Labs/eigenda/pull/1841#discussion_r2256631423

api/clients/v2/coretypes/encoded_payload.go
```diff
@@ -105,20 +102,18 @@ func (ep *EncodedPayload) ToBlob(payloadForm codecs.PolynomialForm) (*Blob, erro
...
-	if err != nil {
-		return nil, fmt.Errorf("create FFT settings from blob length symbols: %w", err)
-	}
+	fftSettings := fft.FFTSettingsFromBlobLengthSymbols(encoding.NextPowerOf2(blobLengthSymbols))
 
 	// the FFT method pads to the next power of 2, so we don't need to do that manually
 	ifftedElements, err := fftSettings.FFT(evalPoly, true)
 	if err != nil {
-		return nil, fmt.Errorf("perform IFFT: %w", err)
+		panic("bug: FFT only returns an error if we don't have enough roots of unity, " +
```
bxue-l2, line 184: This is exactly what I hope it to be. Looking into the function "FFTSettingsFromBlobLengthSymbols", It unnecessarily creates an array of expansion of [root of unity](https://github.com/Layr-Labs/eigenda/blob/9e40fc6f0516c5cda4b2afd9b2c4facbb767ea89/encoding/fft/fft.go#L82), which in theory is identical for that length. 

This is the optimization problem, but we have to do that safely. As now, it is good

api/clients/v2/coretypes/encoded_payload.go
```diff
@@ -5,18 +5,28 @@ import (
...
 	"github.com/Layr-Labs/eigenda/encoding/utils/codec"
 	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
 )
 
 // EncodedPayload represents a payload that has had an encoding applied to it,
-// meaning that it is guaranteed to be a multiple of 32 bytes in length,
-// each such 32 bytes representing a bn254 field element.
+// meaning that it must be a multiple of 32 bytes in length, each such 32 bytes
+// representing a bn254 field element. Furthermore, it must have a power of 2 number
+// of such field elements; that is, [EncodedPayload.LenSymbols] must return a power of 2.
```
bxue-l2: Maybe just add an example. 
(32, 64, 128, 256). But not 96

api/clients/v2/coretypes/encoded_payload.go
```diff
@@ -5,18 +5,28 @@ import (
...
+//
+// To uphold these invariants, always construct an EncodedPayload by using [Payload.ToEncodedPayload].
```
bxue-l2: This is not necessarily true. Also, the invariant above only states that it's a multiple of 32 bytes, There are other ways to construct it. For instance, we store the EncodedPayload in memstore, so when we retrieve bytes, those bytes need to go immediately to encodedpayload. How about the implementation below, having it in this PR will make it easier to fix the memstore later by not touching the encoded payload again to expose the interface

```
// NewEncodedPayload constructs [EncodedPayload] from bytes array. The constructor requires the input bytes
// must be a multiple of 32 bytes in length.
func NewEncodedPayload(bytes []byte) (EncodedPayload, error) {
	if len(bytes) < codec.PayloadHeaderSizeBytes {
		return EncodedPayload{}, fmt.Errorf("encoded payload must be at least 32 bytes long, but got %d bytes", len(bytes))
	}
	numfieldElements := len(bytes) / encoding.BYTES_PER_SYMBOL
	if !encoding.IsPowerOfTwo(numfieldElements) {
		return EncodedPayload{}, fmt.Errorf("encoded payload must be a multiple of 32 bytes in length and the multiple must be power of 2, but got %d bytes", len(bytes))
	}
	return EncodedPayload{bytes}, nil
}
```

api/clients/v2/coretypes/payload_test.go
```diff
@@ -0,0 +1,18 @@
...
+
+import (
+	"github.com/Layr-Labs/eigenda/encoding"
+	"testing"
+
+	"github.com/Layr-Labs/eigenda/common/testutils/random"
+	"github.com/stretchr/testify/require"
+)
+
+func TestPayloadToEncodedPayloadPowerOf2(t *testing.T) {
```
bxue-l2: Can you add a test whose payload length is zero. I.e the case blobLength is 1, and claimed length in payload header is 0. So payload is an empty byte

api/clients/v2/coretypes/blob.go
```diff
@@ -62,81 +87,51 @@ func (b *Blob) Serialize() []byte {
...
 	default:
-		return nil, fmt.Errorf("invalid polynomial form")
+		panic(fmt.Sprintf("invalid codecs.PolynomialForm enum value: %d", payloadForm))
 	}
 
-	maxPermissiblePayloadLength, err := codec.BlobSymbolsToMaxPayloadSize(b.blobLengthSymbols)
-	if err != nil {
-		return nil, fmt.Errorf("get max permissible payload length: %w", err)
-	}
+	return &EncodedPayload{rs.SerializeFieldElements(encodedPayloadElements)}
```
litt3: Perhaps worth adding to the method doc that the contents of the resulting encoded payload cannot be trusted, i.e. the claimed length might be incoherent

api/clients/v2/coretypes/encoded_payload.go
litt3: An example of what a "power of 2" means?

api/clients/v2/coretypes/encoded_payload.go
litt3: Ok, I think the difference is that before the encoded payload wasn't padded to a power of 2.

Now that it is, the old check doesn't make sense. It was only a sanity check before, anyway

api/clients/v2/coretypes/blob.go
```diff
@@ -7,50 +7,75 @@ import (
...
-// BlobFromPolynomial initializes a blob from a polynomial
-func BlobFromPolynomial(coeffPolynomial []fr.Element, blobLengthSymbols uint32) (*Blob, error) {
+// BlobFromCoefficients creates a blob from the coefficients of a polynomial.
+// A Blob must have a power of two coefficients. Thus:
+//   - If the passed coefficients are a power of 2 in length, they will be used as is.
+//   - If the coefficients are not a power of two in length, they will be copied to a new slice that is padded with zeros
+//     to the next power of two in length.
+func BlobFromCoefficients(coefficients []fr.Element) *Blob {
+	paddedCoefficients := coefficients
+	if !encoding.IsPowerOfTwo(len(coefficients)) {
```
litt3: I don't think the pad to power of two is necessary. Both callers of this function are sources that should already be correctly padded (`DeserializeBlob` above, and encoded payload)

On the one hand, this could be construed as being "extra careful". But on the other, padding to a power of two here could obfuscate a logic flaw elsewhere in the conversion pipeline

api/clients/v2/coretypes/encoded_payload.go
```diff
@@ -77,91 +71,49 @@ func (ep *EncodedPayload) Decode() (*Payload, error) {
...
-	} else if encodedPayloadLength < polynomialByteCount {
-		// we assume that the polynomialBytes might have additional trailing 0s beyond the expected size of the encoded
-		// payload. Here, we check the assumption that all trailing values are 0. If there are any non-zero trailing
-		// values, something has gone wrong in the data pipeline, and this should produce a loud failure. Either a
-		// dispersing client is playing sneaky games, or there's a bug somewhere.
-		err := checkTrailingZeros(polynomialBytes, encodedPayloadLength)
-		if err != nil {
-			return nil, fmt.Errorf("check that trailing values in polynomial are zeros: %w", err)
-		}
+	blobLengthSymbols := uint32(encoding.NextPowerOf2(len(fieldElements)))
```
litt3: Why pad to a power of two here?

You've defined the encoded payload as _always_ being a power of two, so it seems that you should return an error here if you discover the encoded payload you're working with is invalid.

api/clients/v2/coretypes/encoded_payload.go
```diff
@@ -77,91 +71,49 @@ func (ep *EncodedPayload) Decode() (*Payload, error) {
...
-	for i := uint32(len(inputBytes)) - 1; i >= nonZeroLength; i-- {
-		if inputBytes[i] != 0x0 {
-			return fmt.Errorf("byte at index %d was expected to be 0x0, but instead was %x", i, inputBytes[i])
-		}
+// blobLengthSymbols is required, to be able to choose the correct parameters when performing FFT
+func evalToCoeffPoly(evalPoly []fr.Element, blobLengthSymbols uint32) []fr.Element {
+	// TODO (litt3): this could conceivably be optimized, so that multiple objects share an instance of FFTSettings,
+	//  which has enough roots of unity for general use. If the following construction of FFTSettings ever proves
+	//  to present a computational burden, consider making this change.
+	fftSettings := fft.FFTSettingsFromBlobLengthSymbols(encoding.NextPowerOf2(blobLengthSymbols))
```
litt3: I think padding something to the power of 2, _when it should already be a power of 2_ could obfuscate logic flaws. If we manage to arrive here, with `blobLengthSymbols` *not* being a power of 2, something is seriously wrong and we should know about it

api/clients/v2/coretypes/encoded_payload.go
```diff
@@ -77,91 +71,49 @@ func (ep *EncodedPayload) Decode() (*Payload, error) {
...
+// evalToCoeffPoly converts an evalPoly to a coeffPoly, using the IFFT operation
 //
-// This function returns an error if any byte in the array after these permitted non-zero values is found to be non-zero
-func checkTrailingZeros(inputBytes []byte, nonZeroLength uint32) error {
-	for i := uint32(len(inputBytes)) - 1; i >= nonZeroLength; i-- {
-		if inputBytes[i] != 0x0 {
-			return fmt.Errorf("byte at index %d was expected to be 0x0, but instead was %x", i, inputBytes[i])
-		}
+// blobLengthSymbols is required, to be able to choose the correct parameters when performing FFT
+func evalToCoeffPoly(evalPoly []fr.Element, blobLengthSymbols uint32) []fr.Element {
```
litt3: Why do you need to pass in `blobLengthSymbols` as a parameter at all? The encoded payload is now definitionally the same length as the blob, so you can just infer the blob length by the length of the encoded payload, right?

api/clients/v2/coretypes/encoded_payload_test.go
```diff
@@ -1,117 +1,98 @@
...
-		_, err = encodedPayloadFromElements(almostTooLongFieldElements, maxPermissiblePayloadLength)
-		require.NoError(t, err)
+// TestEncodePayload tests that the encoding of a Payload to an EncodedPayload works as expected.
+func TestEncodeDecodePayload(t *testing.T) {
+
+	// map of hex-encoded payloads (inputs) and their expected EncodedPayloads (outputs)
+	testCases := map[string]string{
+		// empty payload should only have a header symbol
+		"": "0000000000000000000000000000000000000000000000000000000000000000",
+		"01": "0000000000010000000000000000000000000000000000000000000000000000" + // header with len 1 payload
```
litt3: nit: I think breaking these strings across newlines makes this hard to read. Consider putting the map value on its own line, unbroken

api/clients/v2/coretypes/encoded_payload_test.go
```diff
@@ -1,117 +1,98 @@
...
+			"0001000000000000000000000000000000000000000000000000000000000000", // first byte is always 0 due to bn254 encoding
+		"0102": "0000000000020000000000000000000000000000000000000000000000000000" +
```
litt3: Can you add comments for what you're trying to accomplish with the final two mappings? It's hard to know what's being tested by just reading the bytes

api/clients/v2/coretypes/encoded_payload_test.go
```diff
@@ -1,117 +1,98 @@
...
+			name:       "First byte must be 0x00",
+			encodedHex: "01000000000000000000000000000000000000000000000000000000",
+		},
+		{
+			name:       "Only version 0x00 is supported",
+			encodedHex: "00010000000000000000000000000000000000000000000000000000",
+		},
+		{
+			name:       "Payload length must be a multiple of 32 bytes",
+			encodedHex: "00000000000100000000000000000000000000000000000000000000" + "000100",
```
litt3: unnecessary string concatenation

api/clients/v2/coretypes/encoded_payload_test.go
```diff
@@ -1,117 +1,98 @@
...
+		},
+		{
+			name: "wrong payload length: 32 bytes of data, but header says 64",
+			encodedHex: "00000000000200000000000000000000000000000000000000000000" +
```
litt3: also here, it would be easier to read if the string wasn't broken across newlines

api/clients/v2/coretypes/encoded_payload_test.go
litt3: Sounds like this would fail if the random payload length is close to a power of 2...

In any case, the question is less "why does this work", and more "why does this test exist"? Though we should _also_ be pretty confident we understand why it works.

api/clients/v2/coretypes/payload.go
```diff
@@ -1,81 +1,47 @@
...
+	encodedData := codec.PadPayload(p)
 
-// ToBlob converts the Payload bytes into a Blob
-//
-// The payloadForm indicates how payloads are interpreted. The form of a payload dictates what conversion, if any, must
-// be performed when creating a blob from the payload.
-func (p *Payload) ToBlob(payloadForm codecs.PolynomialForm) (*Blob, error) {
-	encodedPayload := newEncodedPayload(p)
+	// Calculate the length of the EncodedPayload in symbols (including the header) which has to be a power of 2.
+	encodedDataLenSymbols := uint32(len(encodedData)+encoding.BYTES_PER_SYMBOL-1) / encoding.BYTES_PER_SYMBOL
```
litt3: `+encoding.BYTES_PER_SYMBOL-1`

Why are you adding 31 bytes? The `PadPayload` result from above is already 32 byte aligned

api/clients/v2/coretypes/payload_to_blob_test.go
```diff
@@ -21,39 +22,26 @@ func Example_payloadToBlobConversion() {
...
+	payload := Payload(payloadBytes)
+	fmt.Printf("Payload bytes (len %d):\n%s\n", len(payload), hex.EncodeToString(payload))
 
-	encodedPayload := newEncodedPayload(payload)
+	encodedPayload := payload.ToEncodedPayload()
 	fmt.Printf("Encoded Payload bytes (len %d):\n%s\n", len(encodedPayload.Serialize()), hex.EncodeToString(encodedPayload.Serialize()))
 
-	// The encoded payload field elements are interpreted as evaluations of a polynomial.
-	evalPolyFieldElements, err := encodedPayload.toFieldElements()
+	// Change to [codecs.PolynomialFormCoeff] to see that blob will have the same bytes as the encoded payload.
```
litt3: When you say "Change to", what do you mean? Is this a claude instruction which survived as a comment?

In any case, the `Output` comment below would seem to now be outdated, since the claimed bytes there are *different* from the encoded payload bytes

encoding/fft/fft.go
litt3: Yeah, definitely could be placed better

encoding/fft/fft.go
```diff
@@ -92,11 +96,9 @@ func NewFFTSettings(maxScale uint8) *FFTSettings {
...
-func FFTSettingsFromBlobLengthSymbols(blobLengthSymbols uint32) (*FFTSettings, error) {
-	if !encoding.IsPowerOfTwo(blobLengthSymbols) {
-		return nil, fmt.Errorf("blobLengthSymbols %d is not a power of 2", blobLengthSymbols)
-	}
-	maxScale := uint8(math.Log2(float64(blobLengthSymbols)))
-	return NewFFTSettings(maxScale), nil
+// FFTSettingsFromBlobLengthSymbols accepts a blob length, and returns a new instance of FFT settings.
+// blobLengthSymbols should be a power of 2, but we round it up to the next power of 2 if not.
+func FFTSettingsFromBlobLengthSymbols(blobLengthSymbols uint32) *FFTSettings {
+	maxScale := uint8(math.Log2(float64(encoding.NextPowerOf2(blobLengthSymbols))))
```
litt3: Here is another case of padding something to power of 2, when it should already be a power of 2.

It cleans up the function interface at the cost of potentially covering up a bug

encoding/fft/fft_fr.go
```diff
@@ -130,9 +130,10 @@ func (fs *FFTSettings) FFT(vals []fr.Element, inv bool) ([]fr.Element, error) {
 	valsCopy := make([]fr.Element, n)
 	for i := 0; i < len(vals); i++ {
 		valsCopy[i].Set(&vals[i])
-
 	}
 	for i := uint64(len(vals)); i < n; i++ {
+		// TODO: would probably be better to pad with evaluations instead of zeros when doing the IFFT?
```
litt3: I don't understand what this TODO is getting at

api/clients/v2/coretypes/encoded_payload.go
bxue-l2: I think it is more straightforward, also Claude suggests this [PR](https://github.com/Layr-Labs/hokulea/pull/151/files#diff-6a8f96b5008639a0692bd61b03922a0a4aa2600e53a4fa0ece85c2223d26c62eR32)

litt3: IMO "power of 2" is a fundamental concept that doesn't need an example. We don't include an example when referencing "even numbers" (2, 4, 6, but not 5)

samlaf: Agree with litt here. Resolving.

api/clients/v2/coretypes/blob.go
samlaf: True but I'm not sure this detail belongs here? Since:
1. It would be true even if it was in bytes
2. You can click on the [encoding.BlobCommitments.Length] link to see the comment above that Length and see that it is in symbols.

api/clients/v2/coretypes/encoded_payload_test.go
samlaf: I can, but there's no specific reason for adding these, just extra input/output cases that I randomly constructed.
Eventually these could all be used as seed inputs for a fuzz test.

api/clients/v2/coretypes/encoded_payload_test.go
samlaf: I purposefully break strings into 32 byte chunks, so as to make it easier to parse what's going on! The first string is the header, second string is the payload. Guess I'll need to add a comment to make that more obvious.

api/clients/v2/coretypes/encoded_payload_test.go
samlaf: See https://github.com/Layr-Labs/eigenda/pull/1841#discussion_r2263999218

api/clients/v2/coretypes/encoded_payload_test.go
samlaf: The point of breaking it up is that it should actually HELP read it, because the string is broken up into 32 byte chunks.

api/clients/v2/coretypes/encoded_payload.go
samlaf: Created DeserializedEncodedPayload function that doesn't do any checks. See [6335309](https://github.com/Layr-Labs/eigenda/pull/1841/commits/6335309adc2103b805394959be0bd6f2519543c9) and let me know what you think

encoding/fft/fft.go
samlaf: Just made it private since there's a circular import between fft and encoding and didn't feel like making a bigger refactor for this: [251b227](https://github.com/Layr-Labs/eigenda/pull/1841/commits/251b227e7ce190672ea7dbafc2f152cdd031701d)

api/clients/v2/coretypes/payload_test.go
samlaf: Removed this test and incorporated it into the encoded_payload_test.go: [0c379bb](https://github.com/Layr-Labs/eigenda/pull/1841/commits/0c379bb6621fed588c7c542820dc46a7f89db1f7)

api/clients/v2/coretypes/blob.go
samlaf: See [18a45cd](https://github.com/Layr-Labs/eigenda/pull/1841/commits/18a45cd8a4cca9e229ddeeaf10fbda74ba28f664)

api/clients/v2/coretypes/encoded_payload.go
samlaf: believe I removed this in this refactor: [6335309](https://github.com/Layr-Labs/eigenda/pull/1841/commits/6335309adc2103b805394959be0bd6f2519543c9)

api/clients/v2/coretypes/encoded_payload.go
samlaf: believe this was removed in this refactor: [6335309](https://github.com/Layr-Labs/eigenda/pull/1841/commits/6335309adc2103b805394959be0bd6f2519543c9)

api/clients/v2/coretypes/encoded_payload.go
samlaf: believe it was removed in [6335309](https://github.com/Layr-Labs/eigenda/pull/1841/commits/6335309adc2103b805394959be0bd6f2519543c9)

api/clients/v2/coretypes/blob.go
samlaf: Agree. Changed fct signature to return error and made it private since don't think we should expose our internal implementation like this: [bf6ada6](https://github.com/Layr-Labs/eigenda/pull/1841/commits/bf6ada6b9d5e15914744eb0c1e77f5a29d0ec134)

api/clients/v2/coretypes/payload.go
samlaf: [2ffda5c](https://github.com/Layr-Labs/eigenda/pull/1841/commits/2ffda5ccc1ccaec95b86553213b9cb8c6400eb43)

api/clients/v2/coretypes/payload_to_blob_test.go
samlaf: No I meant this as an instruction to the user, that they should feel free to update it to check how the created blob differs, for learning purposes. Tried to clarify the comment in [7592310](https://github.com/Layr-Labs/eigenda/pull/1841/commits/75923107c26f776ec81fb09844da125496d72445)

encoding/fft/fft.go
samlaf: Ended up moving this function back to `coretypes` package as a private function since that makes it safer to panic: [26d6cc0](https://github.com/Layr-Labs/eigenda/pull/1841/commits/26d6cc0a17fad6a03c4f6e7a80dcccf625cd3f21)

encoding/fft/fft_fr.go
samlaf: Probably not needed, and I'm not super sure. Just had this intuition that by padding 0s to evaluation form, we change the polynomial (as opposed to if its in coeff form). If we instead padded with evaluations in eval form, then we wouldn't change the polynomial. Doesn't really matter but one thing this would allow is that we could compute the kzg commitment ahead of time or in parallel as one example, whereas by padding with 0s we change the resulting polynomial and hence its kzg commitment.

cc @bxue-l2 for thoughts here

api/clients/v2/coretypes/encoded_payload_test.go
samlaf: Agree with you and didn't want to spend too much energy trying to think if there was a specific reason why we needed this test, so just went ahead and deleted it: [084e95d](https://github.com/Layr-Labs/eigenda/pull/1841/commits/084e95d2460db012c27ab23546849a8611aaee98)

It was here only because I tried to keep it when I first refactored the library. But that likely changed the test's semantic and hence usefulness. Let me know if you think we should keep it.

api/clients/v2/coretypes/encoded_payload_test.go
litt3: `"00000000000100000000000000000000000000000000000000000000"` is 28 bytes, not 32. It's why I assumed this was random string concatenation

api/clients/v2/coretypes/encoded_payload.go
samlaf: Actually changed my mind, because realized the concept is not "power of 2", its that "the bytes need to have a power of 2 number of field elements". I did add the example lengths (powers of 2 starting at 32) somewhere in some function docs.

api/clients/v2/coretypes/blob.go
```diff
@@ -62,81 +68,70 @@ func (b *Blob) Serialize() []byte {
...
+			"which is impossible because we already checked it above")
 	}
+	return fftedElements
+}
 
-	return fftedElements, nil
+// blobFromCoefficients creates a blob from the coefficients of a polynomial.
+// A Blob must have a power of two coefficients. Thus:
+//   - If the passed coefficients are a power of 2 in length, they will be used as is.
+//   - If the coefficients are not a power of 2 in length, they will be copied to a new slice that is padded with zeros
```
litt3: doc is outdated, no padding takes place in the function

api/clients/v2/coretypes/encoded_payload_test.go
samlaf: Oh good catch...! Updated in [4f9d83f](https://github.com/Layr-Labs/eigenda/pull/1841/commits/4f9d83fa2e7c10ee01ecac4f6e5b195ab6fdadc8)

api/clients/v2/coretypes/encoded_payload.go
```diff
@@ -3,165 +3,197 @@ package coretypes
...
-	return fieldElements, nil
+	if len(ep.bytes)%encoding.BYTES_PER_SYMBOL != 0 {
+		return fmt.Errorf("encoded payload must be a multiple of %d bytes (bn254 field element), "+
+			"but got %d bytes", encoding.BYTES_PER_SYMBOL, len(ep.bytes))
+	}
+	// We could equivalently check that len(ep.bytes) is a power of 2 given that we've already
+	// checked that it's a multiple of 32, but this invariant is closer to the representation of
+	// the encoded payload as a polynomial, and is also more meaningful given
+	// that the length in [encoding.BlobCommitments.Length] is in field elements.
+	numfieldElements := len(ep.bytes) / encoding.BYTES_PER_SYMBOL
```
litt3: nit: improper camel case

`numfieldElements` -> `numFieldElements`

api/clients/v2/coretypes/encoded_payload_test.go
litt3: > but there's no specific reason for adding these

I still think it's worth a comment. Reviewing this, I spent time trying to read the bytes to figure out what case was being exercised here and ended up confused

api/clients/v2/coretypes/encoded_payload_test.go
litt3: Deleting seems like the way to go 👍

api/clients/v2/coretypes/blob.go
```diff
@@ -5,52 +5,58 @@ import (
...
-	// user fetching and reconstructing this blob would determine that the blob length is 1 symbol, when it's actually 2.
-	blobLengthSymbols uint32
 }
 
-// DeserializeBlob initializes a Blob from bytes
+// DeserializeBlob initializes a Blob from bytes.
+// blobLengthSymbols is the length of the blob, which is present in the BlobHeader's [encoding.BlobCommitments.Length].
+// The bytes passed in will be appended with zeros to match the blobLengthSymbols if they are shorter than that length,
+// or an error will be returned if they are longer than that length.
 func DeserializeBlob(bytes []byte, blobLengthSymbols uint32) (*Blob, error) {
```
litt3, line 26: the previous code didn't have a check... but should we be checking that the input `blobLengthSymbols` is a power of 2 here?

api/clients/v2/coretypes/encoded_payload.go
```diff
@@ -3,165 +3,197 @@ package coretypes
...
+//
+// An EncodedPayload can be interpreted as a polynomial, with each 32 byte chunk
+// representing either a coefficient or an evaluation. Interpreting as coefficients has the advantage
+// that the EncodedPayload already represents a [Blob]. Interpreting as evaluations has the advantage that
+// point openings can be made (useful for interactive fraud proofs).
 type EncodedPayload struct {
-	// the size of these bytes is guaranteed to be a multiple of 32
+	// These bytes are kept private in order to force encapsulation, in case we decide in the future
+	// to change the EncodedPayload's representation (eg. store [fr.Element]s directly instead).
+	// Use [DeserializeEncodedPayload] to reconstruct a serialized EncodedPayload.
```
litt3: Outdated doc, `DeserializeEncodedPayload` was renamed

api/clients/v2/coretypes/encoded_payload.go
```diff
@@ -3,165 +3,197 @@ package coretypes
...
+	//
+	// The bytes should contain a power of 2 field elements, each 32 bytes long (see [EncodedPayload.checkLenInvariant]),
+	// meaning valid lengths are [32, 64, 128, 256, ...]
+	// The first 32 bytes represent the header (see [EncodedPayload.decodeHeader]),
+	// and the body (bytes after header) contain serialized bn254 field elements.
 	bytes []byte
 }
 
-// Serialize returns the raw bytes of the encoded payload
+// NewEncodedPayloadUnchecked constructs an [EncodedPayload] from bytes array.
```
litt3: > NewEncodedPayloadUnchecked

Doc has wrong function name

litt3 (dismissed):

api/clients/v2/coretypes/blob.go
samlaf: Ehhhh good question! Looks like we actually never validate that its a power of 2 even in the CertVerifier contract. @bxue-l2 does the [length](https://github.com/Layr-Labs/eigenda/blob/4f9d83fa2e7c10ee01ecac4f6e5b195ab6fdadc8/contracts/src/core/libraries/v2/EigenDATypesV2.sol#L39) proof enforce the length to be a power of 2? Or is this something we need to check in addition to the length proof?

If so we might need to update the CertVerifier contract (first semver patch bump)

api/clients/v2/coretypes/encoded_payload.go
samlaf: [b53ce75](https://github.com/Layr-Labs/eigenda/pull/1841/commits/b53ce7568b60ecab4fad3bb786e97893879977bc)

api/clients/v2/coretypes/encoded_payload.go
samlaf: [b53ce75](https://github.com/Layr-Labs/eigenda/pull/1841/commits/b53ce7568b60ecab4fad3bb786e97893879977bc)

api/clients/v2/coretypes/encoded_payload_test.go
samlaf: Totally forgot: [619b3b8](https://github.com/Layr-Labs/eigenda/pull/1841/commits/619b3b858c74ac12743eb5923913692fd6617abf)

api/clients/v2/coretypes/encoded_payload.go
samlaf: [a56af8f](https://github.com/Layr-Labs/eigenda/pull/1841/commits/a56af8f887f3a9a7b50913126459e2ae60a3ffbe)

api/clients/v2/coretypes/blob.go
samlaf: [6676b9d](https://github.com/Layr-Labs/eigenda/pull/1841/commits/6676b9d4bf6e07fb16f39c8fc9f10dca3c53ab28)

api/clients/v2/coretypes/encoded_payload_test.go
litt3: I like that comment, but the test cases here still aren't being explained. The question that I think would be good to answer is: 

"What's the significance of the `0102` and `01020304050607080910111213141516171819202122232425262728293031` map entries?

encoding/fft/fft_fr.go
bxue-l2: 0 is better. Moreover, we should improve the library to disallow this kind of automatically padding. 
But still I think 0 is better because

Padding is needed (for our context) to allow larger FFT for the Reed solomon encoding. It does not mater what constant is used. As long as it is part of spec and everyone understand so they can do RS recovery.

Suppose your data is [1 2 3 4], it becomes [1 2 3 4 0 0 0 0] after padding. You can replace 0 to other data if you want to. FFT still work, but you shouldn't inject random information

To me padding original evaluation has the benefits that, you put more original data in the polynomial. I guess there are some optimization you can do?

api/clients/v2/coretypes/blob.go
bxue-l2: Validator only checks,

https://github.com/Layr-Labs/eigenda/blob/7dfb40e3f0a3722977ce1c6537f9242a45a099d1/core/v2/validator.go#L65
https://github.com/Layr-Labs/eigenda/blob/7dfb40e3f0a3722977ce1c6537f9242a45a099d1/core/v2/assignment.go#L307

samlaf: TODO: add this error and panic inside retriever clients (meaning rollup should stall and need to slash eigenda operators)

bxue-l2 (dismissed):

api/clients/v2/coretypes/encoded_payload.go
bxue-l2: Why do you have it call "*Unchecked"
- ToEncodedPayloadUnchecked
- DeserializeEncodedPayloadUnchecked

Isn't EncodedPayload by definition not having check on them. I found it a bit confusing

samlaf: EncodedPayload have a bunch of invariants that they must respect (length is power of 2 FEs, header is correct, etc). Those are done in different places (Decode function in the blob->payload path, or in the payload.ToEncodedPayload function in the payload->blob path). The point is that the invariants should always be checked, the question is when. When constructing an EncodedPayload from a payload they are checked in the "constructor", whereas in the decoding path they are checked in the Payload "constructor".

But the `Unchecked` suffix here is a reminder that the invariant check is not being called, and hence should be called at some point.

api/clients/v2/coretypes/blob.go
samlaf: Created a separate PR for this: https://github.com/Layr-Labs/eigenda/pull/1860

encoding/fft/fft_fr.go
samlaf: decided to just remove the comment: [e68b27b](https://github.com/Layr-Labs/eigenda/pull/1841/commits/e68b27baddeaa09f7e9bb6d3c8fc330e8670fec7)

api/clients/v2/coretypes/encoded_payload_test.go
samlaf: I see that's a good point. Decided to refactor the test to follow the usual testStruct with a name. Also added an extra case. Let me know what you think: [49d12ac](https://github.com/Layr-Labs/eigenda/pull/1841/commits/49d12ac20643014b1f8c6897cc869ec0e4fa5104)

litt3 (approved):

api/clients/v2/coretypes/encoded_payload.go
samlaf: Also `DeserializeEncodedPayloadUnchecked` already says:
```
// It does not validate the bytes, to mimic the [Blob.ToEncodedPayloadUnchecked] process.
// The length, header, and body invariants are checked when calling [EncodedPayload.Decode].
```

I think as a team we could/should follow the convention that `Unchecked` means that the function performing validation/invariants-checking have not been called, either because they are costly and should be done by the user if needed only, or for other weird reasons like here (needs to be performed inside fpvm)

litt3: I'm on board with this

PR #1843, opened 2025-08-06
feat: improve contract directory wrapper

cody-littley:

Improves the implementation of a wrapper for the the EigenDA contract directory. Removes the old implementation of the contract directory and replaced it with the new one. Also removed the old implementation from several places it should not be.

### DIFF
--- a/api/clients/v2/examples/client_construction.go
+++ b/api/clients/v2/examples/client_construction.go
@@ -32,8 +32,6 @@ const (
 	disperserHostname          = "disperser-testnet-holesky.eigenda.xyz"
 	certVerifierRouterAddress  = "0x7F40A8e1B62aa1c8Afed23f6E8bAe0D340A4BC4e"
 	registryCoordinatorAddress = "0x53012C69A189cfA2D9d29eb6F19B32e0A2EA3490"
-	// To build Eth Client
-	eigendaDirectoryAddress = "0x90776Ea0E99E4c38aA1Efe575a61B3E40160A2FE"
 	// These two addresses are no longer required for the Eth Client, but parameter is still being taken until we deprecate the flags
 	eigenDAServiceManagerAddress = ""
 	// blsOperatorStateRetrieverAddress is still used for CertBuilder
@@ -335,7 +333,6 @@ func createEthReader(logger logging.Logger, ethClient common.EthClient) (*eth.Re
 	ethReader, err := eth.NewReader(
 		logger,
 		ethClient,
-		eigendaDirectoryAddress,
 		blsOperatorStateRetrieverAddress,
 		eigenDAServiceManagerAddress,
 	)
--- a/core/eth/directory/contract_directory.go
+++ b/core/eth/directory/contract_directory.go
@@ -1,0 +1,131 @@
+package directory
+
+import (
+	"context"
+	"fmt"
+	"sync"
+
+	"github.com/Layr-Labs/eigenda/common"
+	contractIEigenDADirectory "github.com/Layr-Labs/eigenda/contracts/bindings/IEigenDADirectory"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	gethcommon "github.com/ethereum/go-ethereum/common"
+)
+
+type ContractName string
+
+// A golang wrapper around the EigenDA contract directory contract. Useful for looking up contract addresses
+// from within golang code.
+type ContractDirectory struct {
+	logger logging.Logger
+
+	// Only look up each address once. Most of our code only looks this stuff up at startup, so there isn't much
+	// point in checking a particular contract address multiple times.
+	addressCache map[ContractName]gethcommon.Address
+
+	// a handle for calling the EigenDA directory contract.
+	caller *contractIEigenDADirectory.ContractIEigenDADirectoryCaller
+
+	// Used to make this utility thread safe.
+	lock sync.Mutex
+}
+
+// Create a new ContractDirectory instance.
+//
+// The requireCompleteness flag can be used to enforce that the offchain contract list matches the onchain contract
+// list. In general, this should be enabled in test code to raise the alarm when the offchain contract list drifts
+// out of sync with the onchain contract list. In production code, this should be disabled just in case the offending
+// contracts are not important for the code to function correctly.
+func NewContractDirectory(
+	ctx context.Context,
+	logger logging.Logger,
+	ethClient common.EthClient,
+	directoryAddress gethcommon.Address,
+) (*ContractDirectory, error) {
+
+	caller, err := contractIEigenDADirectory.NewContractIEigenDADirectoryCaller(directoryAddress, ethClient)
+	if err != nil {
+		return nil, fmt.Errorf("NewContractDirectory: %w", err)
+	}
+
+	d := &ContractDirectory{
+		logger:       logger,
+		addressCache: make(map[ContractName]gethcommon.Address),
+		caller:       caller,
+	}
+
+	err = d.verifyContractList(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("verifyContractList: %w", err)
+	}
+
+	return d, nil
+}
+
+// GetContractAddress returns the address of a contract by its name. Only contracts defined in the const
+// block above should be used. It is sharply discouraged to use this function with a magic string.
+func (d *ContractDirectory) GetContractAddress(
+	ctx context.Context,
+	contractName ContractName,
+) (gethcommon.Address, error) {
+	if contractName == "" {
+		return gethcommon.Address{}, fmt.Errorf("contract name cannot be empty")
+	}
+
+	// This is not very granular. But since this is uniquely to be a performance hotspot, we can do the simple thing.
+	d.lock.Lock()
+	defer d.lock.Unlock()
+
+	address, ok := d.addressCache[contractName]
+	if ok {
+		return address, nil
+	}
+
+	address, err := d.caller.GetAddress0(&bind.CallOpts{Context: ctx}, (string)(contractName))
+	if err != nil {
+		return gethcommon.Address{}, fmt.Errorf("GetAddress0: %w", err)
+	}
+
+	if address == (gethcommon.Address{}) {
+		return gethcommon.Address{}, fmt.Errorf("constract %s is not registered onchain", contractName)
+	}
+
+	d.addressCache[contractName] = address
+
+	d.logger.Debugf("fetched address for contract %s: %s", contractName, address.Hex())
+	return address, nil
+}
+
+// Checks to see if the list of contracts tracked by this ContractDirectory matches the contracts currently registered
+// in the EigenDA directory contract. Creates some noisy logs if there are any discrepancies.
+func (d *ContractDirectory) verifyContractList(ctx context.Context) error {
+	registeredContracts, err := d.caller.GetAllNames(&bind.CallOpts{Context: ctx})
+	if err != nil {
+		return fmt.Errorf("GetAllNames: %w", err)
+	}
+
+	complete := true
+
+	registeredContractSet := make(map[string]struct{}, len(registeredContracts))
+	for _, name := range registeredContracts {
+		registeredContractSet[name] = struct{}{}
+	}
+
+	for _, contractName := range knownContracts {
+		_, exists := registeredContractSet[string(contractName)]
+		if !exists {
+			d.logger.Errorf(
+				"Contract %s is known to offchain code but not registered in the "+
+					"onchain EigenDA contract directory", contractName)
+			complete = false
+		}
+	}
+
+	if complete {
+		d.logger.Infof("Onchain contract list matches offchain contract list")
+	} else {
+		d.logger.Warnf("Onchain contract list does not match offchain contract list")
+	}
+
+	return nil
+}
--- a/core/eth/directory/contract_names.go
+++ b/core/eth/directory/contract_names.go
@@ -1,0 +1,53 @@
+package directory
+
+// TODO: pare down this list just to the contracts we actually use offchain
+
+// All contracts that the EigenDA offchain code interacts with should be defined here.
+// It is ok to remove contracts from this list if the offchain code doesn't interact with them anymore.
+
+const (
+	PauserRegistry                              ContractName = "PAUSER_REGISTRY"
+	BlsApkRegistry                              ContractName = "BLS_APK_REGISTRY"
+	IndexRegistry                               ContractName = "INDEX_REGISTRY"
+	StakeRegistry                               ContractName = "STAKE_REGISTRY"
+	SocketRegistry                              ContractName = "SOCKET_REGISTRY"
+	RegistryCoordinator                         ContractName = "REGISTRY_COORDINATOR"
+	EjectionManager                             ContractName = "EJECTION_MANAGER"
+	ServiceManager                              ContractName = "SERVICE_MANAGER"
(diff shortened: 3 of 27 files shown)


core/eth/directory/contract_directory.go
```diff
@@ -0,0 +1,151 @@
...
+	// A set of all known contract addresses. Used to prevent magic strings from sneaking into the codebase.
+	legalContractSet map[ContractName]struct{}
+
+	// Used to make this utility thread safe.
+	lock sync.Mutex
+}
+
+// Create a new ContractDirectory instance.
+//
+// The requireCompleteness flag can be used to enforce that the offchain contract list matches the onchain contract
```
litt3: > requireCompleteness

outdated doc

core/eth/directory/contract_directory.go
```diff
@@ -0,0 +1,151 @@
...
+	client bind.ContractBackend,
+	contractDirectoryAddress gethcommon.Address,
+) (*ContractDirectory, error) {
+
+	caller, err := contractIEigenDADirectory.NewContractIEigenDADirectoryCaller(contractDirectoryAddress, client)
+	if err != nil {
+		return nil, fmt.Errorf("NewContractDirectory: %w", err)
+	}
+
+	legalAddressSet := make(map[ContractName]struct{})
```
litt3: better local variable name would probably be `legalContractSet`. These aren't addresses

core/eth/directory/contract_directory.go
```diff
@@ -0,0 +1,151 @@
...
+	err = d.verifyContractList(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("verifyContractList: %w", err)
+	}
+
+	return d, nil
+}
+
+// GetContractAddress returns the address of a contract by its name. Only contracts defined in the const
+// block above should be used. It is sharply discouraged to use this function with a magic string.
```
litt3: > const block above

outdated doc, there aren't any consts above

core/eth/directory/contract_directory.go
```diff
@@ -0,0 +1,151 @@
...
+		return gethcommon.Address{}, fmt.Errorf("contract %s is not a known contract", contractName)
+	}
+
+	address, err := d.caller.GetAddress0(&bind.CallOpts{Context: ctx}, (string)(contractName))
+	if err != nil {
+		return gethcommon.Address{}, fmt.Errorf("GetAddress0: %w", err)
+	}
+
+	if address == (gethcommon.Address{}) {
+		return gethcommon.Address{}, fmt.Errorf("constract %s is not registered onchain", contractName)
```
litt3: constract

core/eth/directory/contract_directory.go
```diff
@@ -0,0 +1,151 @@
...
+// block above should be used. It is sharply discouraged to use this function with a magic string.
+func (d *ContractDirectory) GetContractAddress(
+	ctx context.Context,
+	contractName ContractName,
+) (gethcommon.Address, error) {
+	if contractName == "" {
+		return gethcommon.Address{}, fmt.Errorf("contract name cannot be empty")
+	}
+
+	// This is not very granular. But since this is uniquely to be a performance hotspot, we can do the simple thing.
```
litt3: uniquely -> unlikely

core/eth/directory/contract_directory.go
```diff
@@ -0,0 +1,151 @@
...
+		return gethcommon.Address{}, fmt.Errorf("constract %s is not registered onchain", contractName)
+	}
+
+	d.addressCache[contractName] = address
+
+	d.logger.Debugf("fetched address for contract %s: %s", contractName, address.Hex())
+	return address, nil
+}
+
+// Checks to see if the list of contracts tracked by this ContractDirectory matches the contracts currently registered
```
litt3: "matches" is ambiguous here. it's checking that all offchain contract names are found onchain, but doesn't care about the reverse

The same ambiguity exists in the logger messages at the end of this method

api/clients/v2/examples/client_construction.go
```diff
@@ -32,8 +32,6 @@ const (
 	disperserHostname          = "disperser-testnet-holesky.eigenda.xyz"
 	certVerifierRouterAddress  = "0x7F40A8e1B62aa1c8Afed23f6E8bAe0D340A4BC4e"
 	registryCoordinatorAddress = "0x53012C69A189cfA2D9d29eb6F19B32e0A2EA3490"
-	// To build Eth Client
-	eigendaDirectoryAddress = "0x90776Ea0E99E4c38aA1Efe575a61B3E40160A2FE"
```
litt3, line 36: Ideally, this example would use the directory address to get the other addresses, but feel free to ignore for now to manage scope

core/eth/directory/directory_scanner.go
```diff
@@ -0,0 +1,39 @@
...
+	"fmt"
+
+	contractIEigenDADirectory "github.com/Layr-Labs/eigenda/contracts/bindings/IEigenDADirectory"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	gethcommon "github.com/ethereum/go-ethereum/common"
+)
+
+// A utility method that looks up all contracts in the EigenDA directory contract and returns a map from name
+// to address.
+func GetContractAddressMap(
```
litt3: Why is this utility method separate from the `ContractDirectory` struct you have created? Seems like this would be a reasonable method to put on that struct

(note to future Austin: I didn't review this file yet, if it is kept, review before approving)

tools/discovery/main.go
```diff
@@ -10,14 +10,15 @@ import (
...
-	"github.com/Layr-Labs/eigenda/core/eth"
-	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	"github.com/Layr-Labs/eigenda/core/eth/directory"
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	"github.com/ethereum/go-ethereum/ethclient"
 	"github.com/jedib0t/go-pretty/v6/table"
 	"github.com/urfave/cli/v2"
 )
 
+// TODO what is the purpose of this tool?
```
litt3: todo

https://github.com/Layr-Labs/eigenda/pull/1804

core/eth/directory/contract_directory.go
cody-littley: removed

core/eth/directory/contract_directory.go
cody-littley: updated

core/eth/directory/contract_directory.go
cody-littley: updated

core/eth/directory/contract_directory.go
cody-littley: fixed

core/eth/directory/contract_directory.go
cody-littley: fixed

core/eth/directory/contract_directory.go
cody-littley: Rewrote this doc

api/clients/v2/examples/client_construction.go
cody-littley: Agree. I will circle back on this. The main thing I take issue in the original implementation is that it doesn't actually use the contract directory for anything real, and it causes more problems than it solves to force it to be present.

tools/discovery/main.go
cody-littley: removed

core/eth/directory/directory_scanner.go
cody-littley: I don't think this belongs with the other stuff, the semantics of how it works are different. The `ContractDirectory` limits itself to the things defined in `contract_names.go`, but this doesn't. I've moved this function to be colocated with the one place that wants to use it.

api/clients/v2/examples/client_construction.go
samlaf: Agree that we want external examples to use the EigenDADirectory, so please do get back to this at some point

core/eth/directory/contract_names.go
```diff
@@ -0,0 +1,15 @@
+package directory
+
+// All contracts that the EigenDA offchain code interacts with should be defined here.
+// It is ok to remove contracts from this list if the offchain code doesn't interact with them anymore.
+
+const (
+	ServiceManager         ContractName = "SERVICE_MANAGER"
+	EigenDAEjectionManager ContractName = "EIGEN_DA_EJECTION_MANAGER"
+)
```
samlaf, line 9: guessing this is just a cleanup and you'll add more contracts later like the BLS_OPERATOR_STATE_RETRIEVER?

core/eth/directory/contract_directory.go
```diff
@@ -0,0 +1,146 @@
...
+	addressCache map[ContractName]gethcommon.Address
+
+	// a handle for calling the EigenDA directory contract.
+	caller *contractIEigenDADirectory.ContractIEigenDADirectoryCaller
+
+	// A set of all known contract addresses. Used to prevent magic strings from sneaking into the codebase.
+	legalContractSet map[ContractName]struct{}
+
+	// Used to make this utility thread safe.
+	lock sync.Mutex
```
samlaf:
```suggestion
	// Used to make this utility thread safe.
	cacheLock sync.Mutex
```

samlaf: In general I like to clearly mention what EXACTLY a lock protects, since it makes it a lot easier later on to figure out what exactly the author had in mind. For example some locks that protect weird things like serializability of the requests in the disperser_client and really hard to refactor without fearing of breaking things. And here I had to think and infer "is this lock trying to protect the eth-client as well? probably not... oh its only used for the cache probably"

core/eth/directory/contract_directory.go
```diff
@@ -0,0 +1,146 @@
...
+	ctx context.Context,
+	contractName ContractName,
+) (gethcommon.Address, error) {
+	if contractName == "" {
+		return gethcommon.Address{}, fmt.Errorf("contract name cannot be empty")
+	}
+
+	// This is not very granular. But since this is unlikely to be a performance hotspot, we can do the simple thing.
+	d.lock.Lock()
+	defer d.lock.Unlock()
```
samlaf: Actually pretty bad imo to hold the lock around network calls. For example I will update my discovery CLI to use this, and in that case it has to make 20+ calls to fetch every address, and the cache is not useful because its just a CLI.

core/eth/directory/contract_directory.go
```diff
@@ -0,0 +1,146 @@
...
+	}
+
+	if complete {
+		d.logger.Infof("Onchain contract list matches offchain contract list")
+	} else {
+		d.logger.Warnf("Onchain contract list does not match offchain contract list")
+	}
+
+	return nil
+}
```
samlaf: is this intended? looks buggy to me? Shouldn't you return an error when a known contract doesn't exist onchain?

Also is this complete flag really needed? You already log an Error when a contract doesn't exist. I think we can just return early with an error

core/eth/writer.go
```diff
@@ -37,11 +37,10 @@ var _ core.Writer = (*Writer)(nil)
 func NewWriter(
 	logger logging.Logger,
 	client common.EthClient,
-	eigendaDirectoryHexAddr string,
 	blsOperatorStateRetrieverHexAddr string,
```
samlaf, line 40: no need to do anything here but curious for your thoughts: I've heard people say that you in general should try to not have 2 arguments to a function that have the same type, because its error prone (for example a call site here could easily swap the eigendadirectory and blsOperatorStateRetriever which would prob result in a weird nasty to debug bug).

Not sure that there's an easy way to prevent this here, outside of creating types for each contract address (which is prob also awkward), but wondering if there's a way to use your utility to prevent call sites like this (prob will remove these since we'll only pass the EigenDADirectory everywhere going forwad right?)

node/node.go
```diff
@@ -115,13 +116,13 @@ func NewNode(
 		return nil, fmt.Errorf("could not create DB directory at %s: %w", config.DbPath, err)
 	}
 
-	chainID, err := client.ChainID(context.Background())
+	chainID, err := client.ChainID(ctx)
```
samlaf, line 126: nit: this chainID is fetched here, stored in the node struct, and then only used for debugging purposes in Start(). Think this call should be moved to Start().

In general I've read that golang philosophy favors having constructors that don't require a ctx (so that structs can be created easily and cheaply in tests), and the Start() function should take the context instead. Wonder if you have thoughts about this

core/eth/directory/contract_names.go
cody-littley: Correct, this is the place where we will eventually define all of the contracts we interact with. But this PR was mostly about creating the utility wrapper, not integrating it with things.

core/eth/directory/contract_directory.go
cody-littley: Makes sense. Regardless, this lock is now gone, and instead I use a sync map. Although potentially wasteful if called concurrently for the same contract, I think it unlikely to make a big difference given our intended use patterns.

core/eth/directory/contract_directory.go
cody-littley: This lock is now removed.

Ideally, there would be a function in the contract that would return a list of all entries in the contract directory, which would allow us to gather all data in one efficient fell swoop. But I met resistance when I requested this feature.

core/eth/directory/contract_directory.go
cody-littley: Removed the `complete` flag as well as the final log statement.

I don't think we need to return an error here. There are actually some cases where we might recover from this. Not all things we deploy will use all contracts we list in `contract_names.go`, and so there is no reason to force those things to crash.

core/eth/writer.go
cody-littley: My solution to this particular struct's problems is that I think we should deprecate `reader.go` and `writer.go`. There is way too much unrelated functionality packed into these files. As an alternative, I'd prefer to write simple and purposeful small tools that each individually do a small number of related things.

node/node.go
cody-littley: `chainID` is also used in a log statement in the constructor.

I have a bunch of issues with the way the `Node` struct is instantiated. But I'd prefer to fix these problems in future PRs in order to limit the scope of this one.

I'm not super opinionated about passing a `ctx` to a constructor. I don't really understand the argument as to why structs are more expensive to instantiate if they take a `ctx` argument... just pass in `t.Context()`.

core/eth/directory/contract_directory.go
samlaf: This is typically done by calling a "multicall" contract. There are a few different implementations for this, such as https://github.com/mds1/multicall3. We could write a util to make use of this if its ever needed.

core/eth/directory/contract_directory.go
samlaf: Wdym by "not all things we deploy? Are you saying we would deploy an EigenDADirectory which wouldn't contain all the contract names listed in `contract_names`? I don't see why that would ever be the case?

core/eth/writer.go
samlaf: absolutely 100% agree with this

samlaf (approved):

core/eth/directory/contract_directory.go
cody-littley: Yup. But until it's needed, my preference would be to keep things as simple as possible.

core/eth/directory/contract_directory.go
cody-littley: Let's say I'm deploying the proxy, and for whatever reason the ejection contract is not properly registered. The proxy never calls into the ejection contract, and so there is no reason for it to refuse to start up due to the ejection contract being missing.

PR #1848, opened 2025-08-07
perf: limit chunk memory use

cody-littley:

Create an upper limit on the amount of memory that a validator may use to store chunk data for the `GetChunks()` and `StoreChunks()` gRPCs.


core/data.go
```diff
@@ -662,6 +662,25 @@ type BlobVersionParameters struct {
...
+	if blobLengthSymbols == 0 {
+		return 0, fmt.Errorf("blob length must be greater than 0")
+	}
+
+	// Check that the blob length is a power of 2 using bit manipulation
+	if blobLengthSymbols&(blobLengthSymbols-1) != 0 {
+		return 0, fmt.Errorf("blob length %d is not a power of 2", blobLengthSymbols)
+	}
+
+	chunkLength := blobLengthSymbols * bvp.CodingRate / bvp.NumChunks
```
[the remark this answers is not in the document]
cody-littley: Won't fix. 

- These values come from onchain and blob header. 
- These values are also validated before this point.
- This is not new code, I just moved it.

node/config.go
```diff
@@ -182,6 +173,19 @@ type Config struct {
 	// to aggressively garbage collect so as to keep this amount of memory free. Useful for preventing kubernetes
 	// from OOM-killing the process. Overrides the GCSafetyBufferSizeFraction value if greater than 0.
 	GCSafetyBufferSizeBytes uint64
+
+	// The maximum amount of time to wait to acquire buffer capacity to store chunks in the StoreChunks() gRPC request.
+	StoreChunksBufferTimeout time.Duration
+
+	// StoreChunksBufferSizeFraction controls the maximum memory that can be used to store chunks in the
+	// StoreChunks() gRPC request buffer, as a fraction of the total memory available to the process.
+	// Ignored if StoreChunksBufferSizeGB is greater than 0.
```
litt3: > StoreChunksBufferSizeGB

Actual config field is named `StoreChunksBufferSizeBytes`

node/config.go
```diff
@@ -182,6 +173,19 @@ type Config struct {
...
+	StoreChunksBufferSizeFraction float64
+
+	// StoreChunksBufferSizeGB controls the maximum memory that can be used to store chunks in the
```
litt3: > StoreChunksBufferSizeGB

wrong name here too

node/node.go
```diff
@@ -275,15 +297,14 @@ func NewNode(
...
-		ctx := context.Background()
 		// 12s per block
 		ttl := time.Duration(blockStaleMeasure+storeDurationBlocks) * 12 * time.Second
 		n.ValidatorStore, err = NewValidatorStore(logger, config, time.Now, ttl, reg)
 		if err != nil {
 			return nil, fmt.Errorf("failed to create new store v2: %w", err)
 		}
 
-		blobParams, err := tx.GetAllVersionedBlobParams(ctx)
+		blobParams, err := tx.GetAllVersionedBlobParams(context.Background())
```
litt3: what's the reason for removing the `ctx` variable, and getting a new context for every call? I suspect merge conflicts

node/node.go
```diff
@@ -93,6 +99,13 @@ type Node struct {
 	// TODO: utilize meterer onchain state later to check quorum ID and minimum payments
 	// QuorumCount is the number of quorums in the network.
 	QuorumCount atomic.Uint32
+
+	// TODO future cody: this is not wired in yet
```
litt3: Are you intending to merge like this into master?

litt3 (approved):

node/config.go
cody-littley: fixed

node/config.go
cody-littley: fixed

node/node.go
cody-littley: This will happen in a later PR. For the time being I've deleted this unused field.

node/node.go
cody-littley: I had a bunch of changes in this branch that I decided were not necessary, so I reverted them. This was an unintended side effect of my reversion. I've fixed this so that it is the same as before.

litt3 (dismissed):

litt3 (approved):

PR #1850, opened 2025-08-08
feat(metrics): add v2 client metrics and expose them to the proxy

iquidus:
Closes DAINT-654

Adds additional v2 client metrics and exposes them to the proxy

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/disperser_client.go
+++ b/api/clients/v2/disperser_client.go
@@ -162,8 +162,10 @@ func (c *disperserClient) DisperseBlobWithProbe(
 	quorums []core.QuorumID,
 	probe *common.SequenceProbe,
 ) (*dispv2.BlobStatus, corev2.BlobKey, error) {
-
 	if len(quorums) == 0 {
+		if c.clientMetrics != nil {
+			c.clientMetrics.ReportValidationError("empty_quorums")
+		}
 		return nil, [32]byte{}, api.NewErrorInvalidArg("quorum numbers must be provided")
 	}
 	if c.signer == nil {
@@ -171,6 +173,9 @@ func (c *disperserClient) DisperseBlobWithProbe(
 	}
 	for _, q := range quorums {
 		if q > corev2.MaxQuorumID {
+			if c.clientMetrics != nil {
+				c.clientMetrics.ReportValidationError("invalid_quorum_id")
+			}
 			return nil, [32]byte{}, api.NewErrorInvalidArg("quorum number must be less than 256")
 		}
 	}
@@ -198,11 +203,13 @@ func (c *disperserClient) DisperseBlobWithProbe(
 		return nil, [32]byte{}, fmt.Errorf("error accounting blob: %w", err)
 	}
 
+	paymentMethod := "reservation"
 	if payment.CumulativePayment == nil || payment.CumulativePayment.Sign() == 0 {
 		// This request is using reserved bandwidth, no need to prevent parallel dispersal.
 		c.accountantLock.Unlock()
 	} else {
 		// This request is using on-demand bandwidth, current implementation requires sequential dispersal.
+		paymentMethod = "on-demand"
 		defer c.accountantLock.Unlock()
 	}
 
@@ -211,6 +218,9 @@ func (c *disperserClient) DisperseBlobWithProbe(
 	// check every 32 bytes of data are within the valid range for a bn254 field element
 	_, err = rs.ToFrArray(data)
 	if err != nil {
+		if c.clientMetrics != nil {
+			c.clientMetrics.ReportValidationError("invalid_field_element")
+		}
 		return nil, [32]byte{}, fmt.Errorf(
 			"encountered an error to convert a 32-bytes into a valid field element, "+
 				"please use the correct format where every 32bytes(big-endian) is less than "+
@@ -290,7 +300,15 @@ func (c *disperserClient) DisperseBlobWithProbe(
 	probe.SetStage("verify_blob_key")
 
 	if verifyReceivedBlobKey(blobHeader, reply) != nil {
+		if c.clientMetrics != nil {
+			c.clientMetrics.ReportBlobKeyVerificationError()
+		}
 		return nil, [32]byte{}, fmt.Errorf("verify received blob key: %w", err)
+	}
+
+	// Report successful blob dispersal
+	if c.clientMetrics != nil {
+		c.clientMetrics.ReportBlobDispersal(paymentMethod, uint64(len(data)))
 	}
 
 	return &blobStatus, corev2.BlobKey(reply.GetBlobKey()), nil
@@ -340,7 +358,12 @@ func (c *disperserClient) GetBlobStatus(ctx context.Context, blobKey corev2.Blob
 	request := &disperser_rpc.BlobStatusRequest{
 		BlobKey: blobKey[:],
 	}
-	return c.client.GetBlobStatus(ctx, request)
+
+	result, err := c.client.GetBlobStatus(ctx, request)
+	if err != nil {
+		return nil, fmt.Errorf("GetBlobStatus failed: %w", err)
+	}
+	return result, nil
 }
 
 // GetPaymentState returns the payment state of the disperser client
@@ -367,7 +390,12 @@ func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.G
 		Signature: signature,
 		Timestamp: timestamp,
 	}
-	return c.client.GetPaymentState(ctx, request)
+
+	result, err := c.client.GetPaymentState(ctx, request)
+	if err != nil {
+		return nil, fmt.Errorf("GetPaymentState failed: %w", err)
+	}
+	return result, nil
 }
 
 // GetBlobCommitment is a utility method that calculates commitment for a blob payload.
@@ -383,7 +411,12 @@ func (c *disperserClient) GetBlobCommitment(ctx context.Context, data []byte) (*
 	request := &disperser_rpc.BlobCommitmentRequest{
 		Blob: data,
 	}
-	return c.client.GetBlobCommitment(ctx, request)
+
+	result, err := c.client.GetBlobCommitment(ctx, request)
+	if err != nil {
+		return nil, fmt.Errorf("GetBlobCommitment failed: %w", err)
+	}
+	return result, nil
 }
 
 // initOnceGrpcConnection initializes the grpc connection and client if they are not already initialized.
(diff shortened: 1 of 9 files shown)


api/proxy/metrics/metrics.go
```diff
@@ -28,6 +29,7 @@ type Config struct {
 
 // Metricer ... Interface for metrics
 type Metricer interface {
+	client_metrics.ClientMetricer
 	RecordInfo(version string)
```
samlaf, line 33: Just realized RecordInfo seems like its not used anywhere? Prob can delete?

api/proxy/metrics/metrics.go
```diff
@@ -28,6 +29,7 @@ type Config struct {
 
 // Metricer ... Interface for metrics
 type Metricer interface {
+	client_metrics.ClientMetricer
```
samlaf, line 32: These metrics are never called from within proxy right? They shouldn't be here then. This interface should only contain metrics that are called by proxy instances.

api/proxy/metrics/metrics.go
samlaf: Also doesn't look like `Document()` is ever called either.

api/proxy/metrics/metrics.go
samlaf: I'm guessing you did this because the registry is created inside the `NewMetrics` constructor? I think that's a mistake. The registry should be created outside in the upper main function and passed down to all different metrics structs (imo each client should maintain its own set of structs, there shouldn't be a massive Metrics struct that contains all metrics for everyone, thats too hard to maintain)

api/clients/v2/payloadretrieval/relay_payload_retriever.go
```diff
@@ -67,9 +69,11 @@ func (pr *RelayPayloadRetriever) GetPayload(
 	ctx context.Context,
 	eigenDACert coretypes.RetrievableEigenDACert,
 ) (*coretypes.Payload, error) {
-
 	encodedPayload, err := pr.GetEncodedPayload(ctx, eigenDACert)
 	if err != nil {
+		if pr.metrics != nil {
```
samlaf, line 74: instead of having != nil checks everywhere, think it would be cleaner to use an interface and always call the function, and just pass in a *noopMetrics struct if caller doesn't want to record metrics?

api/clients/v2/payloadretrieval/relay_payload_retriever.go
```diff
@@ -146,12 +163,18 @@ func (pr *RelayPayloadRetriever) GetEncodedPayload(
 			pr.log.Warn(
 				"generate and compare blob commitment",
 				"blobKey", blobKey.Hex(), "relayKey", relayKey, "error", err)
+			if pr.metrics != nil {
+				pr.metrics.ReportCommitmentVerificationFailure(fmt.Sprintf("%d", relayKey))
```
samlaf, line 167: hmm not sure this should be incrementing same metrics as the one below, since this one might just be a local failure right? Doesn't mean the commitment was actually incorrect.

api/clients/v2/metrics/metrics.go
```diff
@@ -0,0 +1,541 @@
...
+	ReportValidatorPayloadRetrieval(status, method string, payloadSize uint64, quorumAttempts int)
+	ReportValidatorQuorumRequest(quorumId string, status string)
+	ReportValidatorCommitmentVerificationFailure(quorumId string)
+	ReportValidatorBlobDeserializationFailure(quorumId string)
+	ReportValidatorPayloadDecodeFailure(blobKey string)
+	ReportValidatorTimeout(quorumId string)
+	ReportValidatorConnectionError(quorumId string)
+}
+
+type ClientMetrics struct {
```
samlaf, line 45: Curious why you went with this giant struct instead of using a per-client smaller struct. I think it would be cleaner for each separate service/client to have its own metrics struct? They can all be embedded in a high-level ClientMetrics struct like this, but each service should only need to care about its own metrics, and more importantly, should be allowed to be built with a metrics struct that only contains the metrics it cares about.

api/clients/v2/disperser_client.go
```diff
@@ -420,3 +458,7 @@ func (c *disperserClient) initOncePopulateAccountant(ctx context.Context) error
 	}
 	return nil
 }
+
+func (c *disperserClient) SetClientMetrics(metrics client_metrics.ClientMetricer) {
+	c.clientMetrics = metrics
+}
```
samlaf, line 464: why use this instead of injecting the metrics directly in the constructor? This seems more error prone/less obvious?

api/clients/v2/disperser_client.go
```diff
@@ -362,7 +390,12 @@ func (c *disperserClient) GetPaymentState(ctx context.Context) (*disperser_rpc.G
 		Signature: signature,
 		Timestamp: timestamp,
 	}
-	return c.client.GetPaymentState(ctx, request)
+
+	result, err := c.client.GetPaymentState(ctx, request)
+	if err != nil {
+		return nil, fmt.Errorf("GetPaymentState failed: %w", err)
+	}
+	return result, nil
```
samlaf, line 398: How do you differentiate this error from the one a level above? Given that both functions (this low-level grpc call) and the wrapping function we are in have the same name?

samlaf: Also this is orthogonal to this PR, so don't think it should be included here personally, even if we decide that we want this.

api/clients/v2/metrics/metrics.go
```diff
@@ -0,0 +1,541 @@
...
+	PaymentFailures   *prometheus.CounterVec
+	// disperser_client
+	BlobDispersalRequests     *prometheus.CounterVec
+	BlobDispersalDuration     *prometheus.HistogramVec
+	BlobSize                  *prometheus.HistogramVec
+	GrpcRequests              *prometheus.CounterVec
+	GrpcRequestDuration       *prometheus.HistogramVec
+	ValidationErrors          *prometheus.CounterVec
+	BlobKeyVerificationErrors *prometheus.CounterVec
+	ConnectionErrors          *prometheus.CounterVec
```
samlaf, line 59: a lot of these are defined but never used anywhere afaict. For example ConnectionErrors and GrpcRequests.
Also you only define 3 disperser_client functions in the interface above but have 8 metrics here. Was the idea that a few of them should be incremented by those methods but you forgot to do so?

api/clients/v2/disperser_client.go
```diff
@@ -57,6 +58,7 @@ type disperserClient struct {
 	prover             encoding.Prover
 	accountant         *Accountant
 	accountantLock     sync.Mutex
+	clientMetrics      client_metrics.ClientMetricer
```
samlaf, line 61: We already have the disperse_with_probe method which contains some metrics. Are you sure nothing overlaps with it? Would be good to describe here in a comment how they are different. Is there a way they could be merged? If so I would personally prefer that by a lot.

api/clients/v2/metrics/metrics.go
```diff
@@ -0,0 +1,541 @@
...
+func (m *ClientMetrics) ReportCommitmentVerificationFailure(relayKey string) {
+	m.CommitmentVerificationFailures.WithLabelValues(relayKey).Inc()
+}
+
+func (m *ClientMetrics) ReportBlobDeserializationFailure(relayKey string) {
+	m.BlobDeserializationFailures.WithLabelValues(relayKey).Inc()
+}
+
+func (m *ClientMetrics) ReportPayloadDecodeFailure(blobKey string) {
+	m.PayloadDecodeFailures.WithLabelValues(blobKey).Inc()
```
samlaf, line 434: What's your philosophy wrt having a bunch of Failure metrics like this?
My first gut intuition is that this is too fine grained and will be hard to maintain going forward.
Also we already have error logs to catch these and alert on, and we could, if we wanted, build metrics from those logs dynamically in prometheus.
Also if we really want them I feel like we should only have a single metrics with the failure type as a label?

api/clients/v2/metrics/metrics.go
```diff
@@ -0,0 +1,541 @@
...
+func (m *ClientMetrics) ReportValidatorCommitmentVerificationFailure(quorumId string) {
+	m.ValidatorCommitmentVerificationFailures.WithLabelValues(quorumId).Inc()
+}
+
+func (m *ClientMetrics) ReportValidatorBlobDeserializationFailure(quorumId string) {
+	m.ValidatorBlobDeserializationFailures.WithLabelValues(quorumId).Inc()
+}
+
+func (m *ClientMetrics) ReportValidatorPayloadDecodeFailure(blobKey string) {
+	m.ValidatorPayloadDecodeFailures.WithLabelValues(blobKey).Inc()
```
samlaf, line 464: Not a good idea to have a label with super high cardinality like this. This won't scale, won't be very usable (can't really build a dashboard from this), and will be super expensive. See prometheus best practices: https://prometheus.io/docs/practices/naming/#labels

api/proxy/store/builder/storage_manager_builder.go
```diff
@@ -580,6 +593,10 @@ func buildPayloadDisperser(
 		return nil, fmt.Errorf("new disperser client: %w", err)
 	}
 
+	if metrics != nil {
+		disperserClient.SetClientMetrics(metrics)
+	}
```
samlaf, line 598: In order to get the probe metrics to show up in proxy, you will need to pass in a registry to the NewPayloadDisperser function below, instead of nil

samlaf (commented):
Left a bunch of comments. There's some good stuff in here but also I think you're trying to do too much in a single PR.

Think it would be better to break this up into:
1. small first PR just to expose the probe metrics in proxy (this is what I had in mind with https://linear.app/eigenlabs/issue/DAINT-654/expose-client-metrics-in-proxy, maybe should have been clearer)
2. second small PR with just accountant metrics for example (where we can debate dependency injection style and metrics conventions like high-cardinality labels)
3. Rest of client metrics once we're set on an approach which can generalize to other clients

api/proxy/metrics/metrics.go
iquidus: `Document()` is called here: https://github.com/ethereum-optimism/optimism/blob/develop/op-service/metrics/doc/cmd.go#L32C5-L32C37

Which we are using for the `eigenda-proxy doc metrics` cli command here: https://github.com/Layr-Labs/eigenda/blob/master/api/proxy/cmd/server/main.go#L41

This cli command only renders metrics created via the [factory](https://github.com/Layr-Labs/eigenda/blob/master/api/proxy/metrics/metrics.go#L55) in the output.

I'm not sure what value this cli command adds if it only shows a few metrics defined in `api/proxy/metrics/metrics.go` and ignores everything else. We should consider removing it (or handling it differently) when we remove the `op-service/metrics` dependency.

PR #1852, opened 2025-08-08
feat: tool for compacting LevelDB

cody-littley:

An offline utility for compacting LevelDB instances.

### DIFF
--- a/tools/compactotron/compactotron.go
+++ b/tools/compactotron/compactotron.go
@@ -1,0 +1,119 @@
+package main
+
+import (
+	"fmt"
+	"os"
+
+	"github.com/Layr-Labs/eigenda/common"
+	"github.com/Layr-Labs/eigenda/litt/util"
+	"github.com/docker/go-units"
+	"github.com/syndtr/goleveldb/leveldb"
+)
+
+// The maximum size of a batch to write to LevelDB.
+const maxBatchSize = 100 * units.MiB
+
+func main() {
+	if len(os.Args) != 3 {
+		fmt.Println("Usage: compactotron <source_path> <destination_path>")
+		os.Exit(1)
+	}
+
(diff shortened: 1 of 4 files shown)


tools/compactotron/compactotron.go
```diff
@@ -0,0 +1,119 @@
...
+	if err != nil {
+		fmt.Printf("Error compacting LevelDB: %v\n", err)
+		os.Exit(1)
+	}
+
+	fmt.Println("Compaction completed successfully.")
+}
+
+// Compacts LevelDB database at the given source path and writes the compacted data to the destination path.
+func CompactLevelDB(source string, destination string) error {
```
litt3, line 35: Consider asserting that the source and destination paths are different. Could accidentally trying to copy a db into itself bork it?

tools/compactotron/compactotron.go
```diff
@@ -0,0 +1,119 @@
...
+	err = util.ErrIfNotExists(source)
+	if err != nil {
+		return fmt.Errorf("source path does not exist: %w", err)
+	}
+
+	exists, err := util.Exists(destination)
+	if err != nil {
+		return fmt.Errorf("failed to check if destination exists: %w", err)
+	}
+	if !exists {
```
litt3: Maybe it would just be safer to fail if the destination already exists. I'm concerned about someone accidentally overwriting their database using this tool, and I can't think of a reason why you'd want to compact into a preexisting DB

tools/compactotron/compactotron.go
cody-littley: added check

tools/compactotron/compactotron.go
cody-littley: makes sense, done

PR #1853, opened 2025-08-08
feat(payments): Implement leaky bucket algorithm

litt3:
Implements the leaky bucket algorithm

Closes EGDA-1831

### DIFF
--- a/core/payments/bias_behavior.go
+++ b/core/payments/bias_behavior.go
@@ -1,0 +1,22 @@
+package payments
+
+// In the leaky bucket implementation, there are different points where we need to decide whether we should err on the
+// side of permitting *more* or *less* throughput.
+//
+// Consider the different users of the leaky bucket:
+//   - Validator nodes should err on the side of permitting *more* throughput. Processing a little extra data isn't
+//     a big deal, but denying usage that a user is entitled to is something to be avoided at all costs.
+//   - Clients should err on the side of utilizing *less* throughput. They should do their best to use the
+//     full capacity of the reservation they're entitled to, but should prefer slight under-use.
+type BiasBehavior string
+
+const (
+	// When in doubt, permit *more* throughput instead of less.
+	//
+	// This is what a validator node should use.
+	BiasPermitMore BiasBehavior = "permitMore"
+	// When in doubt, permit *less* throughput instead of more.
+	//
+	// This is what a client should use.
+	BiasPermitLess BiasBehavior = "permitLess"
+)
--- a/core/payments/leaky_bucket.go
+++ b/core/payments/leaky_bucket.go
@@ -1,0 +1,285 @@
+package payments
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/Layr-Labs/eigenda/common"
+)
+
+// This struct implements the [leaky bucket](https://en.wikipedia.org/wiki/Leaky_bucket) algorithm as a meter.
+//
+// Symbols "leak out" of the bucket at a constant rate, creating capacity for new symbols. The bucket can be "filled"
+// with additional symbols if there is enough available capacity.
+//
+// The standard golang golang.org/x/time/rate.Limiter is not suitable for our use-case, for the following reasons:
+//
+//  1. The Limiter doesn't support the concept of overfilling the bucket. We require the concept of overfill, for cases
+//     where a bucket size might be too small to fit the largest permissible blob size. We don't want to prevent users
+//     with a small reservation size from submitting large blobs.
+//  2. The Limiter uses floating point math. Though it would *probably* be ok to use floats, it makes the distributed
+//     system harder to reason about. What level of error accumulation would we see with frequent updates? Under
+//     what conditions would it be possible for the client and server representations of a given leaky bucket to
+//     diverge, and what impact would that have on our assumptions? These questions can be avoided entirely by using
+//     an integer based implementation.
+//
+// NOTE: This struct doesn't do any synchronization! The caller is responsible for making sure that only one goroutine
+// is using it at a time.
+type LeakyBucket struct {
+	// Defines whether we should err on the side of permitting more or less throughput
+	biasBehavior BiasBehavior
+
+	// Defines different ways that overfilling the bucket should be handled
+	overfillBehavior OverfillBehavior
+
+	// The total number of symbols that fit in the bucket
+	bucketCapacity int64
+
+	// The number of symbols that leak out of the bucket each second
+	symbolsPerSecondLeakRate int64
+
+	// The number of symbols currently in the bucket
+	currentFillLevel int64
+
+	// The time at which the previous leak calculation was made
+	previousLeakTime time.Time
+
+	// The number of symbols which leaked in the "partial second" of the previous leak calculation. A "partial" second
+	// is `epochNanoTime % 1e9`.
+	//
+	// Since the leaky bucket uses integers instead of floats, leak math isn't straight forward. It's easy to calculate
+	// the number of symbols that leak in a full second, since leak rate is defined in terms of symbols / second. But
+	// determining how many symbols leak in a number of nanoseconds requires making a rounding choice. Leak calculation
+	// N needs to take the partialSecondLeakage of calculation N-1 into account, so that the precisely correct number
+	// of symbols are leaked for each full second.
+	//
+	// It would be possible to recalculate this value for N-1 when doing the calculation for N, but storing this value
+	// as a member variable keeps things simple and avoids re-doing the math.
+	previousPartialSecondLeakage int64
+}
+
+// Creates a new instance of the leaky bucket algorithm
+func NewLeakyBucket(
+	// how fast symbols leak out of the bucket
+	symbolsPerSecondLeakRate int64,
+	// symbolsPerSecondLeakRate * bucketCapacityDuration becomes the bucket capacity
+	bucketCapacityDuration time.Duration,
+	// whether to err on the side of permitting more or less throughput
+	biasBehavior BiasBehavior,
+	// how to handle overfilling the bucket
+	overfillBehavior OverfillBehavior,
+	// the current time, when this is being constructed
+	now time.Time,
+) (*LeakyBucket, error) {
+	if symbolsPerSecondLeakRate <= 0 {
+		return nil, fmt.Errorf("symbolsPerSecondLeakRate must be > 0, got %d", symbolsPerSecondLeakRate)
+	}
+
+	product, err := common.SafeMultiplyInt64(symbolsPerSecondLeakRate, bucketCapacityDuration.Nanoseconds())
+	if err != nil {
+		return nil, fmt.Errorf("safe multiply when computing bucket capacity: %w", err)
+	}
+	bucketCapacity := product / 1e9
+
+	if bucketCapacity <= 0 {
+		return nil, fmt.Errorf("bucket capacity must be > 0, got %d (from leak rate %d symbols/sec and duration %s)",
+			bucketCapacity, symbolsPerSecondLeakRate, bucketCapacityDuration)
+	}
+
+	var currentFillLevel int64
+	switch biasBehavior {
+	case BiasPermitMore:
+		// starting with a fill level of 0 means the bucket starts out with available capacity
+		currentFillLevel = 0
+	case BiasPermitLess:
+		// starting with a full bucket means some time must elapse to allow leakage before the bucket can be used
+		currentFillLevel = bucketCapacity
+	default:
+		return nil, fmt.Errorf("unknown bias behavior %s", biasBehavior)
+	}
+
+	return &LeakyBucket{
+		biasBehavior:                 biasBehavior,
+		overfillBehavior:             overfillBehavior,
+		bucketCapacity:               bucketCapacity,
+		symbolsPerSecondLeakRate:     symbolsPerSecondLeakRate,
+		currentFillLevel:             currentFillLevel,
+		previousLeakTime:             now,
+		previousPartialSecondLeakage: 0,
+	}, nil
+}
+
+// Fill the bucket with a number of symbols.
+//
+// Returns nil if the leaky bucket has enough capacity to accept the fill. Returns an
+// InsufficientReservationCapacityError if bucket lacks capacity to permit the fill.
+//
+// If the bucket doesn't have enough capacity to accommodate the fill, symbolCount IS NOT added to the bucket, i.e. a
+// failed fill doesn't count against the meter.
+func (lb *LeakyBucket) Fill(now time.Time, symbolCount int64) error {
+	if symbolCount <= 0 {
+		return fmt.Errorf("symbolCount must be > 0, got %d", symbolCount)
+	}
+
+	err := lb.leak(now)
+	if err != nil {
+		return fmt.Errorf("leak: %w", err)
+	}
+
+	// this is how full the bucket would be, if the fill were to be accepted
+	newFillLevel, err := common.SafeAddInt64(lb.currentFillLevel, symbolCount)
+	if err != nil {
+		return fmt.Errorf("safe add to compute newFillLevel: %w", err)
+	}
+
+	// if newFillLevel is <= the total bucket capacity, no further checks are required
+	if newFillLevel <= lb.bucketCapacity {
+		lb.currentFillLevel = newFillLevel
+		return nil
+	}
+
+	// this fill would result in the bucket being overfilled, so we check the overfill behavior to decide what to do
+	switch lb.overfillBehavior {
+	case OverfillNotPermitted:
+		return &InsufficientReservationCapacityError{symbolCount}
+	case OverfillOncePermitted:
+		zeroCapacityAvailable := lb.currentFillLevel >= lb.bucketCapacity
+
+		// if there is no available capacity whatsoever, dispersal is never permitted, no matter the overfill behavior
+		if zeroCapacityAvailable {
+			return &InsufficientReservationCapacityError{symbolCount}
+		}
+
+		lb.currentFillLevel = newFillLevel
+		return nil
+	default:
+		return fmt.Errorf("unknown overfill behavior %s", lb.overfillBehavior)
+	}
+}
+
+// Reverts a previous fill, i.e. removes the number of symbols that got added to the bucket
+//
+// The input time should be the most up-to-date time, NOT the time of the original fill.
+func (lb *LeakyBucket) RevertFill(now time.Time, symbolCount int64) error {
+	if symbolCount <= 0 {
+		return fmt.Errorf("symbolCount must be > 0, got %d", symbolCount)
+	}
+
+	err := lb.leak(now)
+	if err != nil {
+		return fmt.Errorf("leak: %w", err)
+	}
+
+	newFillLevel, err := common.SafeSubtractInt64(lb.currentFillLevel, symbolCount)
+	if err != nil {
+		return fmt.Errorf("safe subtract to compute newFillLevel: %w", err)
+	}
+
+	// don't let the bucket get emptier than "totally empty"
+	if newFillLevel < 0 {
+		lb.currentFillLevel = 0
+		return nil
+	}
+
+	lb.currentFillLevel = newFillLevel
+	return nil
+}
+
+// Lets the correct number of symbols leak out of the bucket, based on when we last leaked
+//
+// Returns an error if any of the calculations fail, which should not happen during normal usage.
+func (lb *LeakyBucket) leak(now time.Time) error {
+	if now.Before(lb.previousLeakTime) {
+		return &TimeMovedBackwardError{PreviousTime: lb.previousLeakTime, CurrentTime: now}
+	}
+
+	defer func() {
+		lb.previousLeakTime = now
+	}()
+
+	fullSecondLeakage, err := lb.computeFullSecondLeakage(now.Unix())
+	if err != nil {
+		return fmt.Errorf("compute full second leakage: %w", err)
+	}
+
+	// We need to correct the full-second leakage value: the previous leak calculation already let some symbols from a
+	// partial second period leak out, and those symbols shouldn't leak twice
+	//
+	// This value can be negative if the previous leak calculation was within the same second as this calculation.
+	correctedFullSecondLeakage, err := common.SafeSubtractInt64(fullSecondLeakage, lb.previousPartialSecondLeakage)
+	if err != nil {
+		return fmt.Errorf("safe subtract to compute correctedFullSecondLeakage: %w", err)
+	}
+
+	partialSecondLeakage, err := lb.computePartialSecondLeakage(int64(now.Nanosecond()))
+	if err != nil {
+		return fmt.Errorf("compute partial second leakage: %w", err)
+	}
+	lb.previousPartialSecondLeakage = partialSecondLeakage
+
+	actualLeakage, err := common.SafeAddInt64(correctedFullSecondLeakage, partialSecondLeakage)
+	if err != nil {
+		return fmt.Errorf("safe add to compute actualLeakage: %w", err)
+	}
+
+	newFillLevel, err := common.SafeSubtractInt64(lb.currentFillLevel, actualLeakage)
+	if err != nil {
+		return fmt.Errorf("safe subtract to update currentFillLevel: %w", err)
+	}
+
+	// don't let the bucket get emptier than "totally empty"
+	if newFillLevel < 0 {
+		lb.currentFillLevel = 0
+		return nil
+	}
+
+	lb.currentFillLevel = newFillLevel
+	return nil
+}
+
+// Accepts the current number of seconds since epoch. Returns the number of symbols that should leak from the bucket,
+// based on when we last leaked.
+//
+// Since this method only takes full seconds into consideration, the returned value must be used carefully. See leak()
+// for details.
+//
+// Returns an error if the leakage calculation fails, which should not happen during normal usage.
+func (lb *LeakyBucket) computeFullSecondLeakage(epochSeconds int64) (int64, error) {
+	secondsSinceLastUpdate, err := common.SafeSubtractInt64(epochSeconds, lb.previousLeakTime.Unix())
+	if err != nil {
+		return 0, fmt.Errorf("safe subtract to compute secondsSinceLastUpdate: %w", err)
+	}
+
+	fullSecondLeakage, err := common.SafeMultiplyInt64(secondsSinceLastUpdate, lb.symbolsPerSecondLeakRate)
+	if err != nil {
+		return 0, fmt.Errorf("safe multiply to compute fullSecondLeakage: %w", err)
+	}
+	return fullSecondLeakage, nil
+}
+
+// Accepts a number of nanoseconds, which represent a fraction of a single second.
+//
+// Computes the number of symbols which leak out in the given fractional second. Since this deals with integers,
+// the configured bias determines which direction we round in.
+func (lb *LeakyBucket) computePartialSecondLeakage(nanos int64) (int64, error) {
+	product, err := common.SafeMultiplyInt64(nanos, lb.symbolsPerSecondLeakRate)
+	if err != nil {
+		return 0, fmt.Errorf("safe multiply to compute nanos * symbolsPerSecondLeakRate: %w", err)
+	}
+
+	switch lb.biasBehavior {
+	case BiasPermitMore:
+		// Round up, to permit more (more leakage = more capacity freed up)
+		// Add (1e9 - 1) before dividing to round up
+		sum, err := common.SafeAddInt64(product, 1e9-1)
+		if err != nil {
+			return 0, fmt.Errorf("safe add to compute rounding sum: %w", err)
+		}
+		return sum / 1e9, nil
+	case BiasPermitLess:
+		// Round down, to permit less (less leakage = less capacity freed up)
+		return product / 1e9, nil
+	default:
+		return 0, fmt.Errorf("unknown bias: %s", lb.biasBehavior)
+	}
+}
--- a/core/payments/payment_errors.go
+++ b/core/payments/payment_errors.go
@@ -1,0 +1,40 @@
+package payments
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/Layr-Labs/eigenda/encoding"
+)
+
+// InsufficientReservationCapacityError is returned when the leaky bucket doesn't have enough capacity to accommodate
+// a requested dispersal.
+type InsufficientReservationCapacityError struct {
+	// The number of symbols that were requested to be dispersed
+	RequestedSymbols int64
+}
+
+// Implements the error interface
+func (e *InsufficientReservationCapacityError) Error() string {
+	return fmt.Sprintf("insufficient reservation capacity to disperse %d symbols (%d bytes)",
+		e.RequestedSymbols, e.RequestedSymbols*encoding.BYTES_PER_SYMBOL)
+}
+
+// TimeMovedBackwardError is returned when a timestamp is observed that is before a previously observed timestamp.
+//
+// This should not normally happen, but with clock drift and NTP adjustments, system clocks can occasionally jump
+// backward. This error allows the system to handle such cases gracefully rather than fatally erroring.
+type TimeMovedBackwardError struct {
+	// The current time that was provided
+	CurrentTime time.Time
+	// The previously observed time that is after CurrentTime
+	PreviousTime time.Time
+}
+
+// Implements the error interface
+func (e *TimeMovedBackwardError) Error() string {
+	return fmt.Sprintf("time moved backward: current time %s is before previous time %s (delta: %v)",
+		e.CurrentTime.Format(time.RFC3339Nano),
+		e.PreviousTime.Format(time.RFC3339Nano),
+		e.PreviousTime.Sub(e.CurrentTime))
+}
--- a/common/overflow_utils.go
+++ b/common/overflow_utils.go
@@ -1,0 +1,63 @@
+package common
+
+import (
+	"fmt"
+	"math"
+)
+
+// TODO: write tests
+
+// SafeAddInt64 performs addition with overflow detection for int64.
+// Returns the result, or an error if the addition would overflow.
+func SafeAddInt64(a, b int64) (int64, error) {
+	// positive overflow
+	if b > 0 && a > math.MaxInt64-b {
+		return 0, fmt.Errorf("positive addition overflow: %d + %d", a, b)
+	}
+
+	// negative overflow
+	if b < 0 && a < math.MinInt64-b {
+		return 0, fmt.Errorf("negative addition overflow: %d + %d", a, b)
+	}
+
+	return a + b, nil
+}
+
+// SafeSubtractInt64 performs subtraction with overflow detection for int64.
+// Returns the result, or an error if the subtraction would overflow.
+func SafeSubtractInt64(a, b int64) (int64, error) {
+	// positive overflow
+	if b < 0 && a > math.MaxInt64+b {
+		return 0, fmt.Errorf("positive subtraction overflow: %d - %d", a, b)
+	}
+
+	// negative overflow
+	if b > 0 && a < math.MinInt64+b {
+		return 0, fmt.Errorf("negative subtraction overflow: %d - %d", a, b)
+	}
+
+	return a - b, nil
+}
+
+// SafeMultiplyInt64 performs multiplication with overflow detection for int64.
+// Returns the result, or an error if the multiplication would overflow.
+func SafeMultiplyInt64(a, b int64) (int64, error) {
+	// handle zero case early to avoid division by zero
+	if a == 0 || b == 0 {
+		return 0, nil
+	}
+
+	// positive overflow
+	if (a > 0 && b > 0 && a > math.MaxInt64/b) ||
+		(a < 0 && b < 0 && a < math.MaxInt64/b) {
+		return 0, fmt.Errorf("positive multiplication overflow: %d * %d", a, b)
+	}
+
+	// negative overflow
+	if (a > 0 && b < 0 && b < math.MinInt64/a) ||
+		(a < 0 && b > 0 && a < math.MinInt64/b) {
+		return 0, fmt.Errorf("negative multiplication overflow: %d * %d", a, b)
+	}
+
+	return a * b, nil
+}
--- a/core/payments/leaky_bucket_test.go
+++ b/core/payments/leaky_bucket_test.go
@@ -1,0 +1,262 @@
+package payments
+
+import (
+	"errors"
+	"testing"
+	"time"
+
+	"github.com/Layr-Labs/eigenda/common/testutils/random"
+	"github.com/stretchr/testify/require"
+)
+
+var testStartTime = time.Date(1971, 8, 15, 0, 0, 0, 0, time.UTC)
+
+// Tests the NewLeakyBucket constructor
+func TestNewLeakyBucket(t *testing.T) {
+	t.Run("create with valid parameters", func(t *testing.T) {
+		leakyBucket, err := NewLeakyBucket(10, 10*time.Second, BiasPermitLess, OverfillNotPermitted, testStartTime)
+		require.NotNil(t, leakyBucket)
+		require.NoError(t, err)
+	})
+
+	t.Run("create with invalid leak rate", func(t *testing.T) {
+		leakyBucket, err := NewLeakyBucket(-10, 10*time.Second, BiasPermitLess, OverfillNotPermitted, testStartTime)
+		require.Nil(t, leakyBucket)
+		require.Error(t, err, "negative leak rate should cause error")
+
+		leakyBucket, err = NewLeakyBucket(0, 10*time.Second, BiasPermitLess, OverfillNotPermitted, testStartTime)
+		require.Nil(t, leakyBucket)
+		require.Error(t, err, "zero leak rate should cause error")
+	})
+
+	t.Run("create with invalid bucket size duration", func(t *testing.T) {
+		leakyBucket, err := NewLeakyBucket(10, -10*time.Second, BiasPermitLess, OverfillNotPermitted, testStartTime)
+		require.Nil(t, leakyBucket)
+		require.Error(t, err, "negative bucket duration should cause error")
+
+		leakyBucket, err = NewLeakyBucket(10, 0, BiasPermitLess, OverfillNotPermitted, testStartTime)
+		require.Nil(t, leakyBucket)
+		require.Error(t, err, "zero bucket duration should cause error")
+	})
+}
+
+// Test the Fill method
+func TestFill(t *testing.T) {
+	// verify that overfill logic is working as expected
+	t.Run("test overfill", func(t *testing.T) {
+		leakyBucket, err := NewLeakyBucket(11, 10*time.Second, BiasPermitMore, OverfillOncePermitted, testStartTime)
+		require.NoError(t, err)
+		require.NotNil(t, leakyBucket)
+
+		err = leakyBucket.Fill(testStartTime, leakyBucket.bucketCapacity+10)
+		require.NoError(t, err)
+		require.Equal(t, leakyBucket.bucketCapacity+10, leakyBucket.currentFillLevel, "first overfill should succeed")
+
+		// no time elapses, so bucket is still overfilled
+		err = leakyBucket.Fill(testStartTime, 1)
+		require.Error(t, err)
+		var insufficientErr *InsufficientReservationCapacityError
+		require.True(t, errors.As(err, &insufficientErr), "overfill should fail, if bucket is overfilled")
+
+		// let some time elapse, so there is a little bit of available capacity
+		err = leakyBucket.Fill(testStartTime.Add(time.Second), leakyBucket.bucketCapacity+10)
+		require.NoError(t, err, "any available capacity should permit overfill")
+	})
+
+	// make sure filling without overfill works
+	t.Run("valid non-overfill", func(t *testing.T) {
+		leakyBucket, err := NewLeakyBucket(100, 10*time.Second, BiasPermitMore, OverfillNotPermitted, testStartTime)
+		require.NoError(t, err)
+		require.NotNil(t, leakyBucket)
+
+		err = leakyBucket.Fill(testStartTime, leakyBucket.bucketCapacity-10)
+		require.NoError(t, err)
+
+		require.Equal(t, leakyBucket.bucketCapacity-10, leakyBucket.currentFillLevel)
+	})
+
+	// test edge case of filling directly to capacity
+	t.Run("fill to exact capacity", func(t *testing.T) {
+		leakyBucket, err := NewLeakyBucket(100, 10*time.Second, BiasPermitMore, OverfillNotPermitted, testStartTime)
+		require.NoError(t, err)
+
+		err = leakyBucket.Fill(testStartTime, leakyBucket.bucketCapacity)
+		require.NoError(t, err)
+		require.Equal(t, leakyBucket.bucketCapacity, leakyBucket.currentFillLevel)
+	})
+
+	t.Run("fill with invalid symbol count", func(t *testing.T) {
+		leakyBucket, err := NewLeakyBucket(100, 10*time.Second, BiasPermitMore, OverfillNotPermitted, testStartTime)
+		require.NoError(t, err)
+		require.NotNil(t, leakyBucket)
+
+		err = leakyBucket.Fill(testStartTime, 0)
+		require.Error(t, err, "zero fill should not be permitted")
+
+		err = leakyBucket.Fill(testStartTime, -10)
+		require.Error(t, err, "negative fill should not be permitted")
+
+		require.Equal(t, int64(0), leakyBucket.currentFillLevel, "nothing should have been added to the bucket")
+	})
+
+	// tests that waiting a really long time leaks the bucket empty, and that filling after that behaves as expected
+	t.Run("large idle leakage to empty", func(t *testing.T) {
+		leakyBucket, err := NewLeakyBucket(100, 10*time.Second, BiasPermitLess, OverfillNotPermitted, testStartTime)
+		require.NoError(t, err)
+
+		// wait longer than the bucket duration
+		futureTime := testStartTime.Add(15 * time.Second)
+
+		err = leakyBucket.Fill(futureTime, 50)
+		require.NoError(t, err)
+
+		require.Equal(t, int64(50), leakyBucket.currentFillLevel, "bucket should leak empty, then be filled")
+	})
+}
+
+// Tests that revert fill works
+func TestRevertFill(t *testing.T) {
+	t.Run("valid revert fill", func(t *testing.T) {
+		leakyBucket, err := NewLeakyBucket(100, 10*time.Second, BiasPermitMore, OverfillNotPermitted, testStartTime)
+		require.NoError(t, err)
+		require.NotNil(t, leakyBucket)
+
+		err = leakyBucket.Fill(testStartTime, 500)
+		require.NoError(t, err)
+		require.Equal(t, int64(500), leakyBucket.currentFillLevel)
+
+		err = leakyBucket.RevertFill(testStartTime, 200)
+		require.NoError(t, err)
+
+		require.Equal(t, int64(300), leakyBucket.currentFillLevel)
+	})
+
+	t.Run("revert fill resulting in 0 capacity", func(t *testing.T) {
+		leakyBucket, err := NewLeakyBucket(100, 10*time.Second, BiasPermitMore, OverfillNotPermitted, testStartTime)
+		require.NoError(t, err)
+		require.NotNil(t, leakyBucket)
+
+		err = leakyBucket.Fill(testStartTime, 500)
+		require.NoError(t, err)
+		require.Equal(t, int64(500), leakyBucket.currentFillLevel)
+
+		// revert fill with greater than the amount in the bucket
+		err = leakyBucket.RevertFill(testStartTime, 600)
+		require.NoError(t, err)
+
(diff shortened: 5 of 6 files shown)


core/payments/bias_behavior.go
```diff
@@ -0,0 +1,22 @@
...
+const (
+	// When in doubt, permit *more* throughput instead of less.
+	//
+	// This is what a validator node should use.
+	BiasPermitMore BiasBehavior = "permitMore"
+	// When in doubt, permit *less* throughput instead of more.
+	//
+	// This is what a client should use.
+	BiasPermitLess BiasBehavior = "permitLess"
+)
```
cody-littley: Curious to know your reasoning for making these values strings. Not strictly opposed to strings, but I usually avoid them personally for enum equivalents.

- Do you think you will ever have more than two options? If "no", should this be represented by a boolean value?
- Do you expect these strings to be typed into a config flag? If "yes", then this is potentially an error prone way to configure something.

Same question also applies to `overfill_behavior.go`.

core/payments/bias_behavior.go
cody-littley, line 1: You put these files directly into the `payments` directory. Will the `payments` directory eventually contain a bunch of different files? If so, would it make sense for this stuff to live in `payments/bucket`?

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,283 @@
...
+	overfillBehavior OverfillBehavior
+
+	// The total number of symbols that fit in the bucket
+	bucketCapacity int64
+
+	// The number of symbols that leak out of the bucket each second
+	symbolsPerSecondLeakRate int64
+
+	// The number of symbols currently in the bucket
+	currentFillLevel int64
```
cody-littley: Can these fields go negative? If not, is there any particular reason why you use signed integers?

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,283 @@
...
+	biasBehavior BiasBehavior
+
+	// Defines different ways that overfilling the bucket should be handled
+	overfillBehavior OverfillBehavior
+
+	// The total number of symbols that fit in the bucket
+	bucketCapacity int64
+
+	// The number of symbols that leak out of the bucket each second
+	symbolsPerSecondLeakRate int64
```
cody-littley: Something to consider. This utility is intended for use by payments... but is it something that we'd ever want to use elsewhere? If so, we'd probably want to make it more generic.

I'd accept the argument that this should be strongly scoped to payments, if that's how you feel. But if you think this could be generic, then we should probably use generic terminology. A more generic term might be `tokens` instead of `symbols`.

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,283 @@
...
+	//    previousPartialSecond
+	previousPartialSecondLeakage int64
+}
+
+// Creates a new instance of the leaky bucket algorithm
+func NewLeakyBucket(
+	// how fast symbols leak out of the bucket
+	symbolsPerSecondLeakRate int64,
+	// bucketCapacityDuration * symbolsPerSecondLeakRate becomes the bucket capacity
+	bucketCapacityDuration time.Duration,
```
cody-littley: Not a problem if you decide that this should be specific to payments and not allowed for use elsewhere. But if we make this generic, it would probably make more sense to have the bucket size configured by an integer value, and move the duration calculation up a layer to where the payments logic constructs the bucket.

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,283 @@
...
+	// how to handle overfilling the bucket
+	overfillBehavior OverfillBehavior,
+	// the current time, when this is being constructed
+	now time.Time,
+) (*LeakyBucket, error) {
+	if symbolsPerSecondLeakRate <= 0 {
+		return nil, fmt.Errorf("symbolsPerSecondLeakRate must be > 0, got %d", symbolsPerSecondLeakRate)
+	}
+
+	bucketCapacity := symbolsPerSecondLeakRate * bucketCapacityDuration.Nanoseconds() / 1e9
```
cody-littley:
```suggestion
	bucketCapacity := symbolsPerSecondLeakRate * bucketCapacityDuration.Nanoseconds() / time.Second
```

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,283 @@
...
+// Fill the bucket with a number of symbols.
+//
+// - Returns nil if the leaky bucket has enough capacity to accept the fill.
+// - Returns an InsufficientReservationCapacityError if bucket lacks capacity to permit the fill.
+// - Returns a TimeMovedBackwardError if input time is before previous leak time.
+// - Returns a generic error for all other modes of failure.
+//
+// If the bucket doesn't have enough capacity to accommodate the fill, symbolCount IS NOT added to the bucket, i.e. a
+// failed fill doesn't count against the meter.
+func (lb *LeakyBucket) Fill(now time.Time, symbolCount int64) error {
```
cody-littley: Will `InsufficientReservationCapacityError` have special handling logic? If so, would you consider returning `(bool, error)`, and returning `false, nil` when there is insufficient capacity?

Feel free to push back on this if you don't agree.

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,283 @@
...
+	if symbolCount <= 0 {
+		return fmt.Errorf("symbolCount must be > 0, got %d", symbolCount)
+	}
```
cody-littley: these checks become unnecessary if we operate on unsigned types

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,283 @@
...
+		return fmt.Errorf("leak: %w", err)
+	}
+
+	newFillLevel := lb.currentFillLevel - symbolCount
+
+	// don't let the bucket get emptier than "totally empty"
+	if newFillLevel < 0 {
+		lb.currentFillLevel = 0
+		return nil
+	}
```
cody-littley: If you change to use unsigned types you'd have to do this differently

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,283 @@
...
+
+	// We need to correct the full-second leakage value: the previous leak calculation already let some symbols from a
+	// partial second period leak out, and those symbols shouldn't leak twice
+	//
+	// This value can be negative if the previous leak calculation was within the same second as this calculation,
+	// since in that case fullSecondLeakage would be 0.
+	//
+	// Previous leak (N-1)                    Current Leak (N)
+	//        ↓                                      ↓
+	//   |----*----------|----------------|----------*-----|
```
cody-littley: It took me a little while to realize that the first part of the segment `//   |----*` corresponded to `lb.previousPartialSecondLeakage`. Could you label it?

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,283 @@
...
+
+// Accepts a number of nanoseconds, which represent a fraction of a single second.
+//
+// Computes the number of symbols which leak out in the given fractional second. Since this deals with integers,
+// the configured bias determines which direction we round in.
+//
+//	Previous leak (N-1)                      Current Leak (N)
+//	       ↓                                      ↓
+//	  |----*----------|----------------|----------*-----|
+//	                                   ↑__________↑
```
cody-littley: Although I was eventually able to groc your reasoning, I think it would have been easier if the timelines from `computeFullSecondLeakage()` and `computePartialSecondLeakage()` also appeared inline in the `leak()` method.

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,283 @@
...
+//	            returns the correct number of symbols for this time span
+func (lb *LeakyBucket) computePartialSecondLeakage(nanos int64) (int64, error) {
+	switch lb.biasBehavior {
+	case BiasPermitMore:
+		// Round up, to permit more (more leakage = more capacity freed up)
+		// Add (1e9 - 1) before dividing to round up
+		return (nanos*lb.symbolsPerSecondLeakRate + 1e9 - 1) / 1e9, nil
+	case BiasPermitLess:
+		// Round down, to permit less (less leakage = less capacity freed up)
+		return nanos * lb.symbolsPerSecondLeakRate / 1e9, nil
```
cody-littley: use `time.Second` instead of `1e9`.

core/payments/leaky_bucket.go
litt3: I had originally started to name everything very generically, but ended up switching to payment-centric language.

My rationale is that it's probably pretty unlikely that we'd want to reuse this code. But if we ever _do_ decide to reuse it, it would be pretty easy to do a quick naming pass, to make stuff generic.

core/payments/leaky_bucket.go
litt3: Originally, I put everything as `int64`, to avoid doing casts all over the place.

I think you're right, though, this isn't so nice. Fixed all types to be better in 720fe53f1b87d79501a68ed06851238cf7ae72a7

core/payments/leaky_bucket.go
litt3: I do think it's ok for us to say this is payment specific for now.

But, just in case we want to change that in the future, I implemented this suggestion. It seems a better place to draw the line f88932f597570e5350d13f91c7da9609060a5922

core/payments/bias_behavior.go
litt3: ## Are strings a good mechanism for enum-equivalents?

I personally like using strings for enums, since they are self documenting. If you use a `uint8`, you're stuck either having crappy logs, or defining some enum-to-string conversion method.

When it comes to configs and cli flags, I actually think that the `string` strategy is superior to `uint` enums. If you have some random `uint` value defined in a config or as a flag, it's hard to know what that represents. There's the consideration that the user needs to exactly match the enum string, but you can make a checker `StringToEnum` method for cases where this matters. For this particular case, though, the user doesn't ever have to explicitly configure this.

## Should we prefer a `bool` over an enum, for concepts that aren't likely to get a third option?

I've been playing around with defining enums for configurations like `BiasBehavior`, even though I don't expect there ever to be more than two options. I like the readability of it, and it provides a nice location to document the implications of this field. I don't feel super strongly about this, though, so if you find it distasteful I'd be fine converting to a `bool`.

For `OverfillBehavior`, I do think it would be _possible_ that we'd want additional options in the future. So the case for using an enum for that one is a lot stronger.

core/payments/bias_behavior.go
litt3: I guess the question is, how many files are reasonable to have in a single package, before making subdirs?

image

This is what `payments` looks like with a lot of the stuff drafted, but there will be a couple more files before everything is done. Do you think this calls for subdirs?

core/payments/leaky_bucket.go
litt3: 36c802085bddfc7ffa6a630ae1295fcf9dfd6a02

core/payments/leaky_bucket.go
litt3: No longer relevant, since I changed the constructor to accept a capacity directly instead of a duration

core/payments/leaky_bucket.go
litt3: Yes, it will have special logic.

I think you're right, that returning an additional bool makes sense here. Will modify.

litt3: 8d1847302e418039b23b652684d0b143d3327950

core/payments/leaky_bucket.go
litt3: switched to unsigned types

core/payments/leaky_bucket.go
litt3: modified strategy to be compatible with unsigned types

core/payments/leaky_bucket.go
litt3: I reorganized these comment diagrams, to make the train of logic easier to follow. LMK if you think this is good

fff53abd3a957c9d0356b250e3f75cdff6c5923a

I decided to remove duplicate diagrams, to make sure they don't get out of sync. Seems reasonable that anyone really wanting to understand the nuts and bolts of the leakage calculation would go to the `leak` method, which is where these now all live.

core/payments/leaky_bucket.go
litt3: Reorganized

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,271 @@
...
+	secondsSinceLastUpdate := epochSeconds - uint64(lb.previousLeakTime.Unix())
+	fullSecondLeakage := secondsSinceLastUpdate * lb.symbolsPerSecondLeakRate
+	return fullSecondLeakage, nil
+}
+
+// Accepts a number of nanoseconds, which represent a fraction of a single second.
+//
+// Computes the number of symbols which leak out in the given fractional second. Since this deals with integers,
+// the configured bias determines which direction we round in.
+func (lb *LeakyBucket) computePartialSecondLeakage(nanos uint64) (uint64, error) {
```
mooselumph: This indicates that the difference between the bucket sizes for the validator and the client is 1 second. Shouldn't this be both larger and likely configurable?

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,271 @@
...
+// is using it at a time.
+type LeakyBucket struct {
+	// Defines whether we should err on the side of permitting more or less throughput
+	biasBehavior BiasBehavior
+
+	// Defines different ways that overfilling the bucket should be handled
+	overfillBehavior OverfillBehavior
+
+	// The total number of symbols that fit in the bucket
+	bucketCapacity uint64
```
mooselumph: Do we plan to support changing the rate on the fly? (e.g. update to reservation)? 

If so, it's probably better to measure the bucket in units of time instead of capacity. When updating the bucket, just divide the size of the item by the rate to get a time increment.

core/payments/bias_behavior.go
samlaf: yes, big time. IMO every concept should be its own concept, even if small. We already have this problem with the clients which are all in the same package. Now we're trying to add metrics to all of them. Much cleaner imo if they're all in their own package and we just add a `metrics.go` file to all of them.

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,271 @@
...
+
+import (
+	"errors"
+	"fmt"
+	"time"
+)
+
+// This struct implements the [leaky bucket](https://en.wikipedia.org/wiki/Leaky_bucket) algorithm as a meter.
+//
+// Symbols "leak out" of the bucket at a constant rate, creating capacity for new symbols. The bucket can be "filled"
```
samlaf: Symbols -> Field Elements? :D

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,271 @@
...
+// with additional symbols if there is enough available capacity.
+//
+// The standard golang golang.org/x/time/rate.Limiter is not suitable for our use-case, for the following reasons:
+//
+//  1. The Limiter doesn't support the concept of overfilling the bucket. We require the concept of overfill, for cases
+//     where a bucket size might be too small to fit the largest permissible blob size. We don't want to prevent users
+//     with a small reservation size from submitting large blobs.
```
samlaf: can't we just give everyone the same sized buckets, say 32MiB?

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,271 @@
...
+//     system harder to reason about. What level of error accumulation would we see with frequent updates? Under
+//     what conditions would it be possible for the client and server representations of a given leaky bucket to
+//     diverge, and what impact would that have on our assumptions? These questions can be avoided entirely by using
+//     an integer based implementation.
+//
+// NOTE: This struct doesn't do any synchronization! The caller is responsible for making sure that only one goroutine
+// is using it at a time.
+type LeakyBucket struct {
+	// Defines whether we should err on the side of permitting more or less throughput
+	biasBehavior BiasBehavior
```
samlaf: why not use a percentage value directly instead of an enum?

core/payments/leaky_bucket.go
samlaf: dumb question: what about using a fixed capacity for everyone (say 32MiBs)?

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,271 @@
...
+
+	// Defines different ways that overfilling the bucket should be handled
+	overfillBehavior OverfillBehavior
+
+	// The total number of symbols that fit in the bucket
+	bucketCapacity uint64
+
+	// The number of symbols that leak out of the bucket each second
+	symbolsPerSecondLeakRate uint64
```
samlaf: mention that this should be set to the reservation? Or are we trying to keep this implementation free of eigenda concerns?

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,271 @@
...
+	previousLeakTime time.Time
+
+	// The number of symbols which leaked in the "partial second" of the previous leak calculation.
+	//
+	// Since the leaky bucket uses integers instead of floats, leak math isn't straight forward. It's easy to calculate
+	// the number of symbols that leak in a full second, since leak rate is defined in terms of symbols / second. But
+	// determining how many symbols leak in a number of nanoseconds requires making a rounding choice. Leak calculation
+	// N needs to take the partialSecondLeakage of calculation N-1 into account, so that the precisely correct number
+	// of symbols are leaked for each full second.
+	previousPartialSecondLeakage uint64
```
samlaf: this comment doesn't tell me what this is

core/payments/payment_errors.go
```diff
@@ -0,0 +1,25 @@
...
+
+// TimeMovedBackwardError is returned when a timestamp is observed that is before a previously observed timestamp.
+//
+// This should not normally happen, but with clock drift and NTP adjustments, system clocks can occasionally jump
+// backward. This error allows the system to handle such cases gracefully rather than fatally erroring.
+type TimeMovedBackwardError struct {
+	// The current time that was provided
+	CurrentTime time.Time
+	// The previously observed time that is after CurrentTime
+	PreviousTime time.Time
```
samlaf: are you sure that this is needed for our use case? This is why people use monotonic clocks: https://pkg.go.dev/time#hdr-Monotonic_Clocks

Haven't looked into golang's API for it, but I assume there are ways to not be subject to clocks that go backwards

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,271 @@
...
+	// the current time, when this is being constructed
+	now time.Time,
+) (*LeakyBucket, error) {
+	if symbolsPerSecondLeakRate == 0 {
+		return nil, errors.New("symbolsPerSecondLeakRate must be > 0")
+	}
+
+	if bucketCapacity == 0 {
+		return nil, errors.New("bucketCapacity must be > 0")
+	}
```
samlaf: is this really needed? I always thought that with the OverfillOncePermitted behavior, a bucket of size 0 actually makes sense (you can disperse 1 blob, and then you need to wait for its entire size to have leaked before you can disperse another one).

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,271 @@
...
+
+	var currentFillLevel uint64
+	switch biasBehavior {
+	case BiasPermitMore:
+		// starting with a fill level of 0 means the bucket starts out with available capacity
+		currentFillLevel = 0
+	case BiasPermitLess:
+		// starting with a full bucket means some time must elapse to allow leakage before the bucket can be used
+		currentFillLevel = bucketCapacity
```
samlaf: this feels like it would play bad with the overfill not permitted behavior. Imagine a client that can't overfill and is set to biasPermitLess (what you suggest). Then they need to wait for an entire blob size of time to leak before they can submit? Feels not very ideal?

core/payments/payment_errors.go
samlaf: image

samlaf: Which timestamp are we thinking of using for the leaky bucket? the timestamp in the blobheader? If so I have a weird feeling that it won't play well with the monotonic clock component, as that is most likely only related to the current computer on which it is running.

So have a feeling different validators would compute different time differences because of this

core/payments/leaky_bucket.go
litt3: I decided to make this a capacity, so that it's a "pure" leaky bucket implementation, that could potentially be reused in the future.

Note that there will be a `ReservationLedger` which wraps this algorithm, and _that_ struct will have access to the reservation configuration details, be able to compute the correct capacity in the case of an updated reservation.

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,271 @@
...
+	}
+}
+
+// Reverts a previous fill, i.e. removes the number of symbols that got added to the bucket
+//
+// - Returns a TimeMovedBackwardError if input time is before previous leak time.
+// - Returns a generic error for all other modes of failure.
+//
+// The input time should be the most up-to-date time, NOT the time of the original fill.
+func (lb *LeakyBucket) RevertFill(now time.Time, symbolCount uint32) error {
```
samlaf: whats the intention for this function? When would this be called?

core/payments/leaky_bucket.go
litt3: > This indicates that the difference between the bucket sizes for the validator and the client is 1 second

Could you explain what leads you to this conclusion? That's not the case here: this logic is only for computing the local state of the bucket.

The bucket size delta between client and validator would be configured simply by setting the `bucketCapacity` on each to a different value, such that `validatorBucketCapacity - clientBucketCapacity == desiredDelta`

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,271 @@
...
+}
+
+// Lets the correct number of symbols leak out of the bucket, based on when we last leaked
+//
+// - Returns a TimeMovedBackwardError if input time is before previous leak time.
+// - Returns a generic error if any of the calculations fail, which should not happen during normal usage.
+func (lb *LeakyBucket) leak(now time.Time) error {
+	if now.Before(lb.previousLeakTime) {
+		return &TimeMovedBackwardError{PreviousTime: lb.previousLeakTime, CurrentTime: now}
+	}
```
samlaf: As I mentioned elsewhere, I'm not sure this error is possible: https://pkg.go.dev/time#hdr-Monotonic_Clocks

aka this should prob be a panic indicating a bug (we didn't set previousLeakTime properly)

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,271 @@
...
+//  1. The Limiter doesn't support the concept of overfilling the bucket. We require the concept of overfill, for cases
+//     where a bucket size might be too small to fit the largest permissible blob size. We don't want to prevent users
+//     with a small reservation size from submitting large blobs.
+//  2. The Limiter uses floating point math. Though it would *probably* be ok to use floats, it makes the distributed
+//     system harder to reason about. What level of error accumulation would we see with frequent updates? Under
+//     what conditions would it be possible for the client and server representations of a given leaky bucket to
+//     diverge, and what impact would that have on our assumptions? These questions can be avoided entirely by using
+//     an integer based implementation.
+//
+// NOTE: This struct doesn't do any synchronization! The caller is responsible for making sure that only one goroutine
```
samlaf: curious: why this choice? Why not add a mutex to this?

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,271 @@
...
+// Accepts a number of nanoseconds, which represent a fraction of a single second.
+//
+// Computes the number of symbols which leak out in the given fractional second. Since this deals with integers,
+// the configured bias determines which direction we round in.
+func (lb *LeakyBucket) computePartialSecondLeakage(nanos uint64) (uint64, error) {
+	switch lb.biasBehavior {
+	case BiasPermitMore:
+		// Round up, to permit more (more leakage = more capacity freed up)
+		// Add (1e9 - 1) before dividing to round up
+		return (nanos*lb.symbolsPerSecondLeakRate + uint64(time.Second) - 1) / uint64(time.Second), nil
```
samlaf: imo using time.Second here is a bit confusing, as it relies on the user knowing that golang stores times in nanos. I was very confused because it reads like you are adding a second to nanos. IMO `time.Second` should be used when we are dealing with time types, but here since we are just dealing with plain uint64 type, might as well just use 1e9 directly? or define a `nanosPerSecond = 1e9` var?

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,271 @@
...
+//
+// Since this method only takes full seconds into consideration, the returned value must be used carefully. See leak()
+// for details.
+//
+// Returns an error if the leakage calculation fails, which should not happen during normal usage.
+func (lb *LeakyBucket) computeFullSecondLeakage(epochSeconds uint64) (uint64, error) {
+	secondsSinceLastUpdate := epochSeconds - uint64(lb.previousLeakTime.Unix())
+	fullSecondLeakage := secondsSinceLastUpdate * lb.symbolsPerSecondLeakRate
+	return fullSecondLeakage, nil
+}
```
samlaf:
```suggestion
func (lb *LeakyBucket) computeFullSecondLeakage(epochSeconds uint64) uint64 {
	secondsSinceLastUpdate := epochSeconds - uint64(lb.previousLeakTime.Unix())
	fullSecondLeakage := secondsSinceLastUpdate * lb.symbolsPerSecondLeakRate
	return fullSecondLeakage
}
```

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,271 @@
...
+	case BiasPermitMore:
+		// Round up, to permit more (more leakage = more capacity freed up)
+		// Add (1e9 - 1) before dividing to round up
+		return (nanos*lb.symbolsPerSecondLeakRate + uint64(time.Second) - 1) / uint64(time.Second), nil
+	case BiasPermitLess:
+		// Round down, to permit less (less leakage = less capacity freed up)
+		return nanos * lb.symbolsPerSecondLeakRate / uint64(time.Second), nil
+	default:
+		return 0, fmt.Errorf("unknown bias: %s", lb.biasBehavior)
+	}
```
samlaf: imo this is a clear case for a panic. It puts a big mental toll on the reader of this library to see functions returning errors:
- how can this time function return an error??
- if the leak() function errors, what should I do? retry? or will it always retry because its a deterministic error (most likely?)

IMO there's no point in forcing all these questions down your users throats. Things that are bugs should just panic.

core/payments/leaky_bucket.go
```diff
@@ -0,0 +1,271 @@
...
+	}, nil
+}
+
+// Fill the bucket with a number of symbols.
+//
+// - Returns (true, nil) if the leaky bucket has enough capacity to accept the fill.
+// - Returns (false, nil) if bucket lacks capacity to permit the fill.
+// - Returns (false, error) for actual errors:
+//   - TimeMovedBackwardError if input time is before previous leak time.
+//   - Generic error for all other modes of failure.
```
samlaf: Mentioned this in a few places but I don't think any of these errors are actual errors that should be returned to users (becaues they won't know how to handle them!).

All (I think?) of them are either non-existent or should be panics.

core/payments/bias_behavior.go
cody-littley: Sounds good, I'm ok leaving this as is. It's too bad there isn't a formally supported golang enum concept. Other languages with enums allow you to programmatically convert them to/from strings.

core/payments/bias_behavior.go
cody-littley: I'll leave it up to your judgment. This is not a blocking issue for me, since it's trivial to rearrange this kind of stuff down the road.

core/payments/leaky_bucket.go
cody-littley: I think sizing the bucket proportionately to the fill rate is a good idea.

- With proportional bucket size (as currently implemented), the maximum burst traffic is the same regardless of whether there are many small reservations or a few large reservations.
- It's reasonable to assume that customers with larger reservations will need larger bursts.
- A customer with an extremely small reservation might be able to send hours or days worth of traffic with 32mb (as far as I can tell, the current algorithm doesn't really require any minimum reservation unless we arbitrarily enforce one).

core/payments/leaky_bucket.go
cody-littley: The bias behavior is used to decide if we should round up or down when doing integer arithmetic, or whether the bucket starts full or empty. While a percentage would be plausible for initial bucket fill, it isn't clear how that would map onto rounding behavior. Also, as I perceive it, I think we will always want the bucket to start completely full or completely empty (depending on whether you are a client or a validator).

core/payments/payment_errors.go
cody-littley: I think we are planning on using the wall clock time when the message is received.

Although I agree we should never encounter this error in practice, it makes sense to me to watch out for it as a sanity check.

core/payments/leaky_bucket.go
cody-littley: In the past when I've done rate limiting, sometimes an undo is nice if you encounter an error and end up needing to cancel an operation. Not sure if this method is in anticipation of a specific use case, or just a general purpose "nice to have" feature.

core/payments/leaky_bucket.go
cody-littley: I'm not personally a fan of the golang `panic()` implementation, given how difficult it is to intercept. Also, if we end up using the gRPC middleware as you suggest that catches panics, then panicking won't tear down the process and might simultaneously leave this struct in an invalid state.

core/payments/leaky_bucket.go
cody-littley: If you design your threading data flows correctly, you can go completely lock free, and the resulting code ends up being a lot easier to reason about when it comes to thread safety. It's a pattern @litt3  and I have come to embrace over the last few years.

Instead of having N threads all accessing a data structure concurrently, instead have N threads submit messages to a channel. Then, a single thread dequeues things from the channel one at a time and updates the data structure. After each modification, it sends a message back to the thread that sent the message.

For data structures like the bucket algorithm, a single thread is more than sufficient to handle millions of data structure modifications per second. And you don't have to think about whether or not your locks properly keep things safe, because there are no locks! Only a single thread reads or writes from the data structure.

cody-littley (approved):

core/payments/bias_behavior.go
litt3: Reorganized

a89350058c2c82f75af51e50cc4e4a0d7d66df1f

core/payments/leaky_bucket.go
litt3: > dumb question: what about using a fixed capacity for everyone (say 32MiBs)?

The idea is that everyone is entitled to the same amount of burst, relative to the size of their reservation, for the sake of fairness. The "overfill" logic is just to cover the edge case of people with small reservations.

core/payments/leaky_bucket.go
litt3: Well, I guess the question is whether we are going to change the terminology surrounding reservations *specifically* from `symbol` to `field element`.

Right now, reservations are defined in terms of symbols, on chain and off. If we have plans to change reservation terminology to `field elements`, then I agree that this is as good a place to start as any. But if we _aren't_ planning on making that change, e.g. "It's too much work to change the terminology in the contracts, so we're not going to do it", I think mixing terms is counterproductive

core/payments/leaky_bucket.go
litt3: > The bias behavior is used to decide if we should round up or down when doing integer arithmetic, or whether the bucket starts full or empty.

Yep, this is the reason it makes sense to have an enum. These behaviors are tied together.

> we will always want the bucket to start completely full or completely empty

👍

core/payments/leaky_bucket.go
litt3: fec1dbb8839eff34377751c66c9009cab67799c9

core/payments/leaky_bucket.go
litt3: The first sentence says what this is: 

```
	// The number of symbols which leaked in the "partial second" of the previous leak calculation.
```

Are you unclear on what this is in general? Or do you understand the concept, but find the doc lacking?

Definitely willing to improve the doc, but it's not clear to me how.

core/payments/payment_errors.go
litt3: This is *always* wallclock time, whether from the perspective of the client crafting a dispersal, or the disperser or validator checking if a dispersal is valid.

I agree that this shouldn't actually happen, given the way it's implemented with `Before`, and how golang monotonic clocks work. But I feel better keeping the check, just in case our assumptions are wrong. Better an extra assertion than one assertion too few.

The purpose of the special error is so that everyone can handle this gracefully, if it somehow does happen

core/payments/leaky_bucket.go
litt3: You can only disperse a blob if there is *at least 1 symbol* of capacity available. So if total capacity is 0, you'd never be able to disperse anything.

core/payments/leaky_bucket.go
litt3: That's by design. The point is that a client would play it safe, and wait before doing the first dispersal. Since the client doesn't persist state, it has to assume the worst case: that a different client instance was *just* running, and had used the full reservation capacity.

In practice, I expect that we will always have overfill enabled (though that's not a guarantee yet). But if we decide _not_ to have overfill enabled, than the behavior you describe is exactly how we would want it to work.

core/payments/leaky_bucket.go
litt3: Anticipated usecase is if a client tries to make a dispersal, but it fails for some reason at the disperser, prior to being sent to validator nodes. 

In such a case, the disperser would return an error, and *not* do accounting for the dispersal. In such a case, the client should "refund" the debit in the local ledger, so that it matches the state of the other ledgers being kept.

I didn't document this here, because I think the explanation will be better placed in the upcoming `ReservationLedger` struct. The doc here describes *how* the reversion works, and the `ReservationLedger` will describe *why* it wants to revert in a given scenario.

core/payments/leaky_bucket.go
litt3: I _really_ don't like panic here. Given our current set of assumptions, I don't *think* time will go backwards. But, maybe our assumptions are wrong? Maybe someone does a refactor which changes these assumptions: chooses a different time source, changes the member variable to a unix timestamp, home rolls an NTP clock, etc..

The case for using `panic` IMO is when you literally have NO IDEA what to do in a situation. E.g. "I explicitly checked an invariant, but somehow it's still untrue!!!". But in this situation, we know exactly what to do (wait for time to advance). And we have a reasonable theory for how the situation could arise, i.e. "underlying clock was slightly off, but gets synced"

Panicking in situations like this makes things brittle instead of robust.

core/payments/leaky_bucket.go
litt3: It's vastly easier to reason about and test an algorithm if you don't have to consider concurrent access. 

I actually intend to use locks to protect access to this in the initial implementation. But, we could change strategies in the future and not have to make any updates to the underlying algorithm.

core/payments/leaky_bucket.go
litt3: fcf0b570de07134be203e1c8e44e39247f6a5563

core/payments/leaky_bucket.go
samlaf: > The idea is that everyone is entitled to the same amount of burst, relative to the size of their reservation, for the sake of fairness.

In that case I agree with @mooselumph that it makes more sense to measure in terms of time. This is done all throughout networking literature afaiu, when speaking of queues lengths wrt BDPs.

Also its a nice reparameterization that makes both parameters orthogonal (rate + size(mb) are not, but rate+size(seconds) is). Imo its always better to have orthogonal parameters, because its easier to understand the degrees of freedom that way.

core/payments/leaky_bucket.go
samlaf: Totally forgot about that. Agree this can't be done now, so resolving.

core/payments/leaky_bucket.go
samlaf: Makes sense. In that case why not just have a boolean named `bucketStartsFull` or something?
Was a lot harder and had to look around the code, and still wasn't clear until I asked what this bias was doing.

core/payments/leaky_bucket.go
samlaf: Must have missed that, my bad! I'm still a bit confused as to how this works (admitedly haven't spent enough time reading the full implementation).

If I call leak many times during single second, will `previousPartialSecondLeakage` increment each time? So `previous` is not really correct?

aka if I call leak at times 0, 0.25s, 0.5s, 0.75s, and 1s, then would previousPartialSecondLeakage be: [0, 0.25, 0.5, 0.75, 1] * reservationRate?

core/payments/payment_errors.go
samlaf: if its always wallclock time at different locations, then why are we going out of our way to use this complicated int-based implementation instead of using floats? We don't have consensus on the wallclock times so do we actually gain any benefits from the preciseness of using ints?

core/payments/leaky_bucket.go
samlaf: You could also just change the semantic to be: you can disperse a blob if numSymbols <= bucketCapacity (which would work for the case capacity=0). But I'm just bike-shedding at this point.

core/payments/leaky_bucket.go
litt3: > becaues they won't know how to handle them!

The users of this struct _will_ know how to handle them. If you're a client: shut down, since your local accounting is borked, there's nothing to do. If you're a disperser or a validator, log the occurrence and move on.

At the end of the day, panicking in places like this seems like an unnecessary risk, and I'm still not sold on the philosophy. It places too much confidence in how we _think_ code will behave.

I also I worry about how adopting a `panic` philosophy will interact with preexisting code. It's best practice to release locks with `defer`, to be panic safe, but do we actually follow that everywhere? Claude identified the following deviations from that policy:

```
  1. relay/cache/cache_accessor.go:105,110
  2. relay/cache/cache_accessor.go:105,125
  3. relay/cache/cache_accessor.go:184,211
  4. relay/cache/cache_accessor.go:234,237
  5. disperser/batcher/encoding_streamer.go:217,220
  6. disperser/batcher/encoding_streamer.go:245,247
  7. disperser/batcher/encoding_streamer.go:389,391
  8. disperser/batcher/encoding_streamer.go:448,450
  9. disperser/batcher/encoding_streamer.go:454,456
  10. disperser/batcher/encoding_streamer.go:514,516
  11. core/meterer/onchain_state.go:112,115
  12. core/meterer/onchain_state.go:112,121
  13. core/meterer/onchain_state.go:171,175
  14. core/meterer/onchain_state.go:193,196
  15. disperser/apiserver/server.go:951,953
  16. node/node.go:844,853
  17. litt/disktable/control_loop.go:216,218
  18. litt/disktable/control_loop.go:339,341
  19. litt/disktable/disk_buffer.go:149,151
  20. litt/littbuilder/db_impl.go:309,314
  21. disperser/dataapi/server.go:1041,1043
  22. disperser/batcher/inmem/minibatch_store.go:259,261
  23. disperser/cmd/encoder/main.go:229,231
  24. litt/index_lock.go:19,25
  25. litt/key_lock.go:59,71
  26. litt/key_lock.go:75,87
  27. node/grpc/server.go:340,342
  28. relay/metadata_provider.go:308,310
```

Maybe none of these would interact negatively with a panic here... but my confidence is *low* that the preexisting code in our repo is panic-safe.

core/payments/leaky_bucket.go
litt3: > In that case why not just have a boolean named bucketStartsFull or something?

It seems counterintuitive to me that a boolean called `bucketStartsFull` would affect rounding math. We don't want to round one way or another *because* of the initial capacity. We want to choose an initial capacity *and* configure rounding behavior based on the configured bias, which is determined by the role of the entity running the algorithm

I also think defining it in terms of `bias` helps convey _why_ you'd want to start the bucket full or empty.

core/payments/leaky_bucket.go
litt3: It's difficult to make the purpose of the field explicitly clear without fully explaining how the integer-based leak logic works. Since the `leak()` function already has detailed documentation of how that logic functions, I opted to point the reader to the `leak()` function documentation. 

IMO, since this is an internal variable not intended for external consumption, it seems reasonable that the reader might have to dig into the actual implementation to get a full understanding- but I think the doc as it now exists should be enough to get people pointed in the right direction

5cfbaff8edd0d3d5716470b8b1a5e971ded4ea64

core/payments/leaky_bucket.go
litt3: The size of the bucket _will_ be defined in terms of time, the question is just where to do the multiplication. If we define it in terms of time at the level of the `LeakyBucket`, then we remove the option in the future to utilize this leaky bucket for a purpose that *doesn't* size the bucket based on time.

I don't think it makes sense to force the `LeakyBucket` to understand the reason behind why it has a given capacity: it only needs to know what the allowed capacity is. This `LeakyBucket` is a tool used by the `ReservationLedger`. It seems reasonable to me that we give the `ReservationLedger` leeway to configure the bucket capacity how it sees fit.

I wonder if the disagreement here might be because you haven't seen the full context of this utility yet. The `LeakyBucket` won't be a struct you see directly inside a client: it's a pure algorithm that the struct in charge of tracking a reservation can utilize.

core/payments/payment_errors.go
litt3: 1. The standard `Limiter` isn't an option, because it doesn't support overfill. So we need to implement the algorithm regardless
2. I think it's just best practice to avoid floats. Edge cases are harder to reason about, and it's not _that_ much more complicated to use ints instead of floats.

If it weren't for # 1, I'd have just used the standard `Limiter`. But since we're implementing the algorithm anyway, ints seem the better option IMO

core/payments/leaky_bucket.go
litt3: 34af64700b0705204bd6ef6b864b7a872806f8a1

I think `time.Second` is the standard way of working with durations, but I agree it's not super clear. Tried to make it less confusing, LMK what you think

core/payments/payment_errors.go
cody-littley: I think the primary benefit from using integer math is the ability to say "if your clock is accurate, you get EXACLY what you reserved, no rounding errors". Would floating point rounding errors be large enough to actually make much of a difference? Probably not? But I don't see harm in keeping the integer math since the effort has already been completed.

core/payments/leaky_bucket.go
mooselumph: @litt3, are you understanding the point about the orthogonal parameterization? 

The idea to me would be that we probably want to parameterize this in terms of: 
- DataRate (in symbols/time)
- WindowLength (in time)

If these are the natural, orthogonal parameters of the system, then using these in the algorithm allows us to change one or the other without side effects. 

I get that you could parameterize the ReservationLedger to calculate the capacity in symbols by doing DataRate*WindowLength. But when I do this dynamically, the existing state of the bucket will be broken unless I also do some kind of update to the bucket state, which might not be supported by the interface. 

The main thing though is that between these two choices, there isn't one that is more natural or unopinionated than the other, IMO. Both are just possible parameterizations of the algorithm, and we should choose the one that makes the most sense for our use case.

core/payments/leaky_bucket.go
mooselumph: Ah, okay, I misinterpreted, thanks

mooselumph: I think it's because you said this: 

// Consider the different users of the leaky bucket:
//   - Validator nodes should err on the side of permitting *more* throughput. Processing a little extra data isn't
//     a big deal, but denying usage that a user is entitled to is something to be avoided at all costs.
//   - Clients should err on the side of utilizing *less* throughput. They should do their best to use the
//     full capacity of the reservation they're entitled to, but should prefer slight under-use.

Is the idea that the validators and clients will use different bias behavior *and* have different capacities? Is this actually necessary?

core/payments/leaky_bucket.go
litt3: Yeah, that makes sense. I'll go ahead and revert to defining duration instead of capacity 👍

core/payments/leaky_bucket.go
litt3: > Is the idea that the validators and clients will use different bias behavior and have different capacities?

Yes, that's the idea.

> Is this actually necessary?

Fair question, but I think the answer is `yes`. Here are some consideration:

1. The current plan is that the clients would start with a `full` bucket, to err on the side of _underutilization_.
    - If we reverse this and start with an `empty` bucket, that means a client that has a `full` bucket prior to a restart would boot up, and be prone to over-utilization ("My bucket is empty, disperse away 🚀")
    - If a client has some sort of recurring problem, and restarts multiple times in rapid succession, it might over-utilize the reservation to such an extent that validator nodes will start rejecting dispersals! No good ❌ 
2. The current plan is that validators will start with an `empty` bucket, to err on the side of permitting _more_ throughput.
    - If we reverse this and start validators with a `full` bucket, we have a problem: even though the buckets are different sizes, a "full bucket" still means "no capacity"
    - Therefore, if a bucket is `empty` prior to a restart, when a validator comes online and initializes with a `full` bucket, it won't allow any dispersals until some capacity becomes available!
    - In this situation, an honest client could be denied its due. No good ❌ 
3. I don't think the question of rounding is all that important, it would be just fine if everyone rounds in the same direction. I only implemented separate branches for rounding since I was already implementing the concept of a `bias` (starting full or empty)

TLDR; I don't see a way that we can get around different starting fills. I could see the argument for making everyone round in the same direction, and don't have super strong opinions on that.

core/payments/leaky_bucket.go
litt3: See [this related conversation](https://github.com/Layr-Labs/eigenda/pull/1853#discussion_r2271070778)

If we remove the difference in rounding behavior, I'd be fine switching this to a bool `start full`

core/payments/leaky_bucket.go
litt3: e899236e2d5ef5bac3dfe6603cc77db606651c16

core/payments/reservation/leaky_bucket.go
```diff
@@ -0,0 +1,280 @@
...
+		lb.currentFillLevel = newFillLevel
+		return true, nil
+	}
+
+	// this fill would result in the bucket being overfilled, so we check the overfill behavior to decide what to do
+	switch lb.overfillBehavior {
+	case OverfillNotPermitted:
+		return false, nil
+	case OverfillOncePermitted:
+		zeroCapacityAvailable := lb.currentFillLevel >= lb.bucketCapacity
```
anupsv, line 148: Probably some debug logs.

core/payments/reservation/leaky_bucket.go
```diff
@@ -0,0 +1,280 @@
...
+		return false, errors.New("symbolCount must be > 0")
+	}
+
+	err := lb.leak(now)
+	if err != nil {
+		return false, fmt.Errorf("leak: %w", err)
+	}
+
+	// this is how full the bucket would be, if the fill were to be accepted
+	newFillLevel := lb.currentFillLevel + uint64(symbolCount)
```
anupsv, line 135: Should we account for overflows ?  i know the total number of symbols would be anywhere close but more on the lines of caution. There's a few overflow and one underflow if i remember.

core/payments/leaky_bucket.go
samlaf: So you only changed the constructor here, not the actual internal parameterization, which I think is what @mooselumph had in mind, re his initial comment:

> If so, it's probably better to measure the bucket in units of time instead of capacity. When updating the bucket, just divide the size of the item by the rate to get a time increment.

Haven't thought through all the consequences and it might be too much bike shedding to figure this out until we actually have a use case where we update the bucket dynamically without requiring a restart though.

core/payments/leaky_bucket.go
samlaf: I see! I think I only now realized that this bias affects 2 things. I think it would be good to explicitly list them in the comment for the bias if we stick with this nomenclature.

core/payments/payment_errors.go
samlaf: To be clear at this point I'm purely arguing for the sake of learning and because it's fun. Weekend lecture: https://perso.ens-lyon.fr/jean-michel.muller/goldberg.pdf

> "if your clock is accurate, you get EXACLY what you reserved, no rounding errors"

Is that true, from the other conversations it seems like we have rounding in this implementation as well and need a bias parameter to affect that rounding.

core/payments/leaky_bucket.go
litt3: Ah, ok. I wasn't understanding, you mean literally measure the fullness in amounts of time. It's pretty unintuitive way to think about it like that, IMO. 

Writing the logic to do dynamic updates really wouldn't be complicated, regardless of how this is parameterized...

core/payments/leaky_bucket.go
samlaf: > panicking won't tear down the process and might simultaneously leave this struct in an invalid state

Hmm this is a good point. Was talking with @litt3 about this on slack, we're basically getting into https://en.wikipedia.org/wiki/Exception_safety territory.

So the main point of the grpc middleware is that it ONLY recovers panics to prevent the whole server from crashing, and instead just let the goroutine that was dealing with that specific user close and send a 500 to the user. This defer recover behavior would only be installed on goroutines that are dealing with user requests.

I guess we need to talk about where this leaky bucket would live then. Because if we add a mutex to it and have this leaky bucket struct shared by all goroutines, then I agree with you we need to worry about it being left in an invalid state (although here it doesn't seem to be the case because it just panics before doing anything).

If we instead install it in a special goroutine that only deals with payments and receives requests from user-processing goroutines via a channel, then if it crashes we need to decide what to do... this feels like a reason to prefer the mutex implementation to me?

samlaf: > But in this situation, we know exactly what to do (wait for time to advance)

Good point.

> Given our current set of assumptions, I don't think time will go backwards. But, maybe our assumptions are wrong?

Problem with this line of thinking is that you can go darn deep with this and start questioning every single assumption about your computer. If we don't trust the golang compiler then do we need to add checks after literally everything we do? Even the hardware can lie to you: personal laptop RAMs don't have ECC and random bit flips do actually happen! Then you can start looking into hardware protections like what https://en.wikipedia.org/wiki/Tandem_Computers does for aviation to run 3 CPUs doing the same computation and take the mode response. My point is that questioning all assumptions goes veeeryyyy deep. Should we add array bound checks everywhere like modern languages do? What about adding overflow/underflow? Languages (and hardware!) make safety tradeoffs in the name of performance.

> Maybe someone does a refactor which changes these assumptions: chooses a different time source, changes the member variable to a unix timestamp, home rolls an NTP clock

The point of panic'ing is PRECISELY to catch these things quicker! I really don't think we should be scared of panics. We'll catch them in tests most likely, and if not in preprod. And the main point which Joe Duffy keeps repeating is that this will at least force you to catch them, whereas soft error handling will leave weird bugs out in the wild. He gives an example in his blog post where when they ported some microsoft app to their midori platform they immediately caught and fixed some pretty major bugs that were affecting japanese users.

core/payments/leaky_bucket.go
samlaf: That's a very good point. As I said elsewhere though, I don't think we need to think of panic the way we do `throwing exceptions` in other languages (assuming you're thinking like this because of Java influence). panics are not meant to be caught as a control-flow mechanism like in other languages. The only reason for recovering them should be to kill the goroutine and return a 500 msg to its user in a webserver. In that case iiuc the mutex would be GC'd.

Hmm I am assuming that there's a finalizer/destructor that will release the mutex on GC here... and also I do remember reading that this is problematic because we're never sure WHEN the GC will actually run (might take some time depending on memory pressure). See https://tamir.dev/posts/resource-management-02/

Ok good points good points...!

core/payments/reservation/leaky_bucket.go
litt3: I don't think I'd like to mix logging into this utility, we can log at a higher level if it makes sense

core/payments/reservation/leaky_bucket.go
litt3: @samlaf and I had a long discussion about overflows [here](https://github.com/Layr-Labs/eigenda/pull/1856)

We came to the conclusion that it's not necessary to check in the util

core/payments/leaky_bucket.go
litt3: The problem is that the failure modes can be rare, and NOT get caught before production, in which case decreasing blast radius is a good thing.

I agree, though, there's really not clear answer when it comes to questioning assumptions. You can go deep or shallow, and just gotta settle somewhere 🤷

core/payments/leaky_bucket.go
litt3: > I am assuming that there's a finalizer/destructor that will release the mutex on GC here

In a perfect world, yeah. But I almost  guarantee some/most/all of the examples above _don't_ do that.

core/payments/leaky_bucket.go
litt3: b6b462e868a072c44302de636f5a71fea2fcd014

core/payments/payment_errors.go
litt3: There's _deterministic_ rounding, i.e. you don't have to worry about float precision.

But really, I think either method would be serviceable here, and the tradeoffs are pretty minor

core/payments/leaky_bucket.go
litt3: e726813a0a878f8c4010c5eaef9ffde7bcc53e2f

core/payments/reservation/leaky_bucket.go
litt3: LMK if you disagree, or if you'd like to sync about this!

anupsv: No it makes sense. I had a similar intuition that we should be doing some of these checks elsewhere. Good to know you've gone through it.

PR #1856, opened 2025-08-11
feat: Implement overflow utils

litt3:
Implement overflow utils required for leaky bucket payments

### DIFF
--- a/common/overflow_utils.go
+++ b/common/overflow_utils.go
@@ -1,0 +1,61 @@
+package common
+
+import (
+	"fmt"
+	"math"
+)
+
+// SafeAddInt64 performs addition with overflow detection for int64.
+// Returns the result, or an error if the addition would overflow.
+func SafeAddInt64(a, b int64) (int64, error) {
+	// positive overflow
+	if b > 0 && a > math.MaxInt64-b {
+		return 0, fmt.Errorf("positive addition overflow: %d + %d", a, b)
+	}
+
+	// negative overflow
+	if b < 0 && a < math.MinInt64-b {
+		return 0, fmt.Errorf("negative addition overflow: %d + %d", a, b)
+	}
+
+	return a + b, nil
+}
+
+// SafeSubtractInt64 performs subtraction with overflow detection for int64.
+// Returns the result, or an error if the subtraction would overflow.
+func SafeSubtractInt64(a, b int64) (int64, error) {
+	// positive overflow
+	if b < 0 && a > math.MaxInt64+b {
+		return 0, fmt.Errorf("positive subtraction overflow: %d - %d", a, b)
+	}
+
+	// negative overflow
+	if b > 0 && a < math.MinInt64+b {
+		return 0, fmt.Errorf("negative subtraction overflow: %d - %d", a, b)
+	}
+
+	return a - b, nil
+}
+
+// SafeMultiplyInt64 performs multiplication with overflow detection for int64.
+// Returns the result, or an error if the multiplication would overflow.
+func SafeMultiplyInt64(a, b int64) (int64, error) {
+	// handle zero case early to avoid division by zero
+	if a == 0 || b == 0 {
+		return 0, nil
+	}
+
+	// positive overflow
+	if (a > 0 && b > 0 && a > math.MaxInt64/b) ||
+		(a < 0 && b < 0 && a < math.MaxInt64/b) {
+		return 0, fmt.Errorf("positive multiplication overflow: %d * %d", a, b)
+	}
+
+	// negative overflow
+	if (a > 0 && b < 0 && b < math.MinInt64/a) ||
+		(a < 0 && b > 0 && a < math.MinInt64/b) {
+		return 0, fmt.Errorf("negative multiplication overflow: %d * %d", a, b)
+	}
+
+	return a * b, nil
+}
--- a/common/overflow_utils_test.go
+++ b/common/overflow_utils_test.go
@@ -1,0 +1,79 @@
+package common
+
+import (
+	"math"
+	"testing"
+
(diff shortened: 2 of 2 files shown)


common/overflow_utils.go
```diff
@@ -0,0 +1,61 @@
...
+	if (a > 0 && b > 0 && a > math.MaxInt64/b) ||
+		(a < 0 && b < 0 && a < math.MaxInt64/b) {
+		return 0, fmt.Errorf("positive multiplication overflow: %d * %d", a, b)
+	}
+
+	// negative overflow
+	if (a > 0 && b < 0 && b < math.MinInt64/a) ||
+		(a < 0 && b > 0 && a < math.MinInt64/b) {
+		return 0, fmt.Errorf("negative multiplication overflow: %d * %d", a, b)
+	}
```
samlaf, line 58: Doing 4 divisions in order to perform 1 safe multiplication is crazy expensive. See http://ithare.com/infographics-operation-costs-in-cpu-clock-cycles/

samlaf: Do we absolutely need ints instead of uints? If not, claude recommends using https://pkg.go.dev/math/bits#Mul64, which are compiler optimized functions which return the carryover bit which we can simply check for overflow.

litt3: > Doing 4 divisions in order to perform 1 safe multiplication is crazy expensive.

4x the cost of an incredibly cheap operation is still incredibly cheap. At the scale these functions will be used (basic accounting math for each blob), I don't think it's really worth being concerned about a couple additional operations

litt3: > Do we absolutely need ints instead of uints?

There are negative numbers involved in the accounting logic, filling and emptying buckets. Ints are necessary here

litt3:
@samlaf 

> How are you thinking of using this?

When doing accounting logic, we're doing integer math. It adds high cognitive load to consider the possibility of over/underflow for every operation. If you just use overflow safe functions for your math, you can ignore all the various possible edge cases and just know "if some weird overflow thing happens, we will just safely fail"

> Given the discussions we've had last weeks, would these be bug examples where we should just panic in these functions instead?

Weren't you saying something about how catching panics doesn't necessary work correctly when using multiple goroutines? The logic in the PR will not always be running as part of a grpc handler, so that middleware you were talking about won't be enough.

TBH I'm just not sold on using panics for these cases. There seem to be "gotchas" involved that I don't like having to consider, and the benefit in this case is just super minimal... *slightly* simpler function signatures for some utility methods.

> Also seems like https://pkg.go.dev/math/bits is recommended way to implement a library like this, since the compiler will optimize the performance of the calls. Haven't thought through whether this works for underflows carefully though..

I had claude show me what these utils would look like using `bits`, and it's really complicated. I think it's a premature optimization to use hard-to-reason-about bitwise logic here. IMO these utils being easy to reason about is more valuable than them being perfectly CPU optimized

common/overflow_utils.go
samlaf: > At the scale these functions will be used (basic accounting math for each blob)

Won't these functions also be used also to perform validation and reject incorrect blob requests for example? Could be an easy source of DDoS for example.

In any case if you really think its not needed, then maybe add a TODO comment saying if this ever does become a bottleneck then we can switch to a more optimized implementation.

litt3: In an app level DDoS situation, my guess would be that other costs would outstrip the cost of this more expensive integer math.

But you're right, it's a good point to note. I'll add a comment with some details, explaining the selected tradeoffs with the current implementation, and we can potentially revisit this further down the road when we see the performance in action.

samlaf:
> When doing accounting logic, we're doing integer math. It adds high cognitive load to consider the possibility of over/underflow for every operation. If you just use overflow safe functions for your math, you can ignore all the various possible edge cases and just know "if some weird overflow thing happens, we will just safely fail"

My issue is that this should never overflow/underflow if the implementation is not buggy. The only reason it would overflow is if someone sends a gargantuan blob size right? But in that case we would want to check that logic somewhere higher in the call stack, where we actually have the blob semantic in mind (say a check for max blob size).

My point is that there's just nothing to do with an error returned from this safeMath library. It literally will never be caught. And instead itll just bubble up the call stack in a bunch of different places and its hard to reason about where it will end up.

> I think I recall you saying something about how catching panics doesn't necessary work correctly when using multiple goroutines? The logic in the PR will not always be running as part of a grpc handler, so that middleware you were talking about won't be enough.

It's actually the other way around. Panics work well in multiple goroutines because they walk up the goroutine call stack, and then in servers we install a recover at the top which catches those and returns a 500 to the user. In non-server cases there's no safe way to recover anyways... so just let it panic and give the user a stack trace for debugging.

> I had claude show me what these utils would look like using bits, and it's really complicated. I think it's a premature optimization to use hard-to-reason-about bitwise logic here. IMO these utils being easy to reason about is more valuable than them being perfectly CPU optimized

Hmm I'm surprised, as my understanding (for add/sub at least) is that you'll just check if the carryover bits are nonzero and panic if so. For mult I'm really not sure how to use it. Maybe the complexity you're talking about comes from the fact that you're trying to use ints?

litt3:
> The only reason it would overflow is if someone sends a gargantuan blob size right? But in that case we would want to check that logic somewhere higher in the call stack, where we actually have the blob semantic in mind (say a check for max blob size).

Yeah, I think you're right about this. I went through each case where I was using a "safe" method, and I cannot find any cases where it would _actually_ be possible to reach, given the invariants we enforce elsewhere in our system. 

I was trying to make this "general purpose", but at the end of the day, I think it's just more complex than necessary. I'm closing this PR, and tweaking my leaky bucket to just use basic math operations. Thanks for the review!

PR #1857, opened 2025-08-11
feat: add 'litt prune' CLI tool

cody-littley:

Adds a tool for doing manual garbage collection on a LittDB file system. Works against both snapshots and root DB files.

### DIFF
--- a/litt/cli/prune.go
+++ b/litt/cli/prune.go
@@ -131,7 +131,7 @@ func pruneTable(
 			return 0, fmt.Errorf("this is a symlinked snapshot directory, " +
 				"snapshot directory cannot be spread across multiple sources")
 		}
-		upperBoundFile, err := disktable.LoadBoundaryFile(false, path.Join(sources[0], tableName))
+		upperBoundFile, err := disktable.LoadBoundaryFile(disktable.UpperBound, path.Join(sources[0], tableName))
 		if err != nil {
 			return 0, fmt.Errorf("failed to load boundary file for table %s at path %s: %w",
 				tableName, sources[0], err)
@@ -197,7 +197,7 @@ func writeLowerBoundFile(snapshotRoot string, tableName string, deletedSegments 
 		// No segments were deleted, no need to write a lower bound file.
 		return nil
 	}
-	lowerBoundFile, err := disktable.LoadBoundaryFile(true, path.Join(snapshotRoot, tableName))
+	lowerBoundFile, err := disktable.LoadBoundaryFile(disktable.LowerBound, path.Join(snapshotRoot, tableName))
 	if err != nil {
 		return fmt.Errorf("failed to load boundary file for table %s at path %s: %w",
 			tableName, snapshotRoot, err)
--- a/litt/cli/table_info.go
+++ b/litt/cli/table_info.go
@@ -152,7 +152,7 @@ func tableInfo(logger logging.Logger, tableName string, paths []string, fsync bo
 				tableName, paths)
 		}
 
-		upperBoundFile, err := disktable.LoadBoundaryFile(false, path.Join(paths[0], tableName))
+		upperBoundFile, err := disktable.LoadBoundaryFile(disktable.UpperBound, path.Join(paths[0], tableName))
 		if err != nil {
 			return nil, fmt.Errorf("failed to load boundary file for table %s at path %s: %w",
 				tableName, paths[0], err)


litt/cli/prune.go
```diff
@@ -0,0 +1,230 @@
...
+	defer releaseLocks()
+
+	// Determine which tables to prune.
+	var tables []string
+	foundTables, err := lsPaths(logger, sources, false, fsync)
+	if err != nil {
+		return fmt.Errorf("failed to list tables in paths %v: %w", sources, err)
+	}
+	if len(allowedTables) == 0 {
+		tables = foundTables
```
litt3, line 67: This seems a bit risky to me, e.g. what if someone wants to prune a single table, but accidentally hits `enter` before specifying tables?

It might be safer to have a separate `--all` flag to prune everything

litt/cli/prune.go
```diff
@@ -0,0 +1,230 @@
...
+		for _, table := range foundTables {
+			if _, ok := allowedTablesSet[table]; ok {
+				tables = append(tables, table)
+			}
+		}
+	}
+
+	// Prune each table.
+	for _, table := range tables {
+		bytesDeleted, err := pruneTable(logger, sources, table, maxAgeSeconds, fsync)
```
litt3, line 78: - The logic is built around pruning multiple tables.
- In `pruneTable`, you are throwing an error if there is more than one total source for snapshots
- Couldn't it be the case that you want to prune a snapshot table, which only exists in a single source, _and_ prune other tables that exist in other sources?

litt/cli/prune.go
```diff
@@ -0,0 +1,230 @@
...
+	}
+
+	if ok, err := errorMonitor.IsOk(); !ok {
+		return 0, fmt.Errorf("error monitor reports errors: %w", err)
+	}
+
+	if isSnapshot {
+		// This is a snapshot. Write a lower bound file to tell the DB not to re-snapshot files than have been pruned.
+		if len(deletedSegments) > 1 {
+			lowerBoundFile, err := disktable.LoadBoundaryFile(true, path.Join(sources[0], tableName))
```
litt3: any reason to use `true`, instead of your defined boundary type for the input param?

litt/cli/prune.go
```diff
@@ -0,0 +1,230 @@
...
+		return 0, fmt.Errorf("failed to check if segment %d is a snapshot: %w", lowestSegmentIndex, err)
+	}
+
+	if isSnapshot {
+		// If we are dealing with a snapshot, respect the snapshot upper bound specified by LittDB.
+		if len(sources) > 1 {
+			return 0, fmt.Errorf("this is a symlinked snapshot directory, " +
+				"snapshot directory cannot be spread across multiple sources")
+		}
+		upperBoundFile, err := disktable.LoadBoundaryFile(false, path.Join(sources[0], tableName))
```
litt3: same question here, why pass in `false`, instead of the defined enum type?

litt/cli/prune.go
cody-littley: As discussed offline, the main intended use case of this command is to prune all tables, and pruning by specific tables was an afterthought. We agreed to leave this as-is for the time being.

litt/cli/prune.go
cody-littley: Pruning multiple unrelated things at the same time adds some complexity I'd rather not have to handle. Before pruning, the code gathers up the segment files, which could be spread across multiple directories. If it has to untangle which directories associated together and which are not, then the workflow suddenly becomes very complicated.

I'd prefer to require the user to issue several prune commands if they want to prune both a snapshot and a regular DB, as opposed to supporting it all in a single command.

litt/cli/prune.go
cody-littley: An enum would be better. But change this here would cause merge conflicts in other pending PRs which utilize boundary files. Much easier to fix this once the snapshotting feature fully merges. Here's a ticket to track the work https://linear.app/eigenlabs/issue/EGDA-1858/make-littdb-boundary-file-type-an-enum

litt/cli/prune.go
cody-littley: Similar to before. I think you make a valid point, but I don't want to change it in this PR due to the merge conflicts it would cause.

litt/cli/prune.go
litt3: As far as I can tell, it already is an enum which wraps around a boolean. But you're just using a raw boolean here instead of the enum you already have defined. 

Maybe I'm misunderstanding, but in any case, no big deal. I was just curious. Feel free to resolve this comment if you feel there's nothing else to do here.

cody-littley: Right! I forgot I already did this. I've fixed all cases that use the raw boolean to instead use the corresponding enum value.

litt/cli/prune.go
cody-littley: fixed

litt3 (approved):

PR #1860, opened 2025-08-11
feat(coretypes): add MaliciousOperatorsError

samlaf:
Started from this discussion: https://github.com/Layr-Labs/eigenda/pull/1841#discussion_r2266759587 We realized that we need to differentiate errors that are the result of malicious operators, since these should eventually lead to operators getting slashed.

This PR adds the MaliciousOperatorsError type, and the first error of this kind, for when the dacert's blob length is not a power of 2.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/clients/v2/coretypes/derivation_errors.go
```diff
@@ -5,10 +5,46 @@ import (
...
+	// See https://layr-labs.github.io/eigenda/integration/spec/6-secure-integration.html#2-cert-validation
+	ErrInvalidCertDerivationError = DerivationError{StatusCode: 3}
+	// Signifies that the blob is incorrectly encoded, and cannot be decoded into a valid payload.
+	// See [codecs.PayloadEncodingVersion] for the different supported encodings.
+	// See https://layr-labs.github.io/eigenda/integration/spec/6-secure-integration.html#3-blob-validation
+	ErrBlobDecodingFailedDerivationError = DerivationError{StatusCode: 4}
+)
+
+// Sentinel [MaliciousOperatorsError] errors.
+// Extend these with
```
litt3: > // Extend these with

stub?

api/clients/v2/payloadretrieval/validator_payload_retriever.go
```diff
@@ -170,6 +178,13 @@ func (pr *ValidatorPayloadRetriever) retrieveBlobWithTimeout(
 	}
 
 	blob, err := coretypes.DeserializeBlob(blobBytes, uint32(header.BlobCommitments.Length))
+	if errors.Is(err, coretypes.ErrBlobLengthSymbolsNotPowerOf2) {
```
litt3, line 176: I liked the function doc addition you made to the relay retriever, any reason not to put a similar doc on this method?

litt3 (approved):

api/clients/v2/coretypes/derivation_errors.go
```diff
@@ -5,10 +5,46 @@ import (
...
+var (
+	// [encoding.BlobCommitments.Length] needs to be a power of 2, and that is checked by the eigenda validators:
+	// https://github.com/Layr-Labs/eigenda/blob/cc392dbabef362f2e03a4b35616a407d45fad510/core/v2/assignment.go#L308
+	// Therefore, if we ever receive a cert with a blob length that is not a power of 2,
+	// it means that the eigenda validators are colluding and doing something fishy.
+	ErrCertCommitmentBlobLengthNotPowerOf2MaliciousOperatorsError = MaliciousOperatorsError{
```
litt3, line 40: Is this the naming schema we want to follow for error extension? Concatenation?

I'm ok with this big ole verbose name if other people are, just wanna make sure we all get on the same page. We should be doing more of this kind of thing, I suspect

bxue-l2: I am ok with long name, but I think it is Go's insufficient error handling. If it is in Rust, I will define a Enum of MaliciousOperatorsError, with each variant not having the "MaliciousOperatorsError" part.

api/clients/v2/payloadretrieval/relay_payload_retriever.go
```diff
@@ -189,6 +197,13 @@ func (pr *RelayPayloadRetriever) retrieveBlobWithTimeout(
 	}
 
 	blob, err := coretypes.DeserializeBlob(blobBytes, blobLengthSymbols)
+	if errors.Is(err, coretypes.ErrBlobLengthSymbolsNotPowerOf2) {
+		panic(fmt.Errorf(
+			"retrieveBlobWithTimeout: blobLengthSymbols=%d is not power of 2: "+
+				"this is a major broken invariant, that should have been checked by the validators, "+
+				"and the caller (GetEncodedPayload) should already have checked this invariant "+
+				"and returned a MaliciousOperatorsError", blobLengthSymbols))
+	}
```
ethenotethan: nit - why put panic vs log error and return 500 via proxy? I guess since blob length is preimage input to blob key calculation used for retrieval lookups then it's guaranteed to be tamper resistant by the client with the first invariant check. iiuc you've modified retrievals like:
```
1. checkPow2(blobLengthSymbols)
2. fetch blob
3. checkPow2(blobLengthSymbols) when deserializing
```

also in the worst case if this is called it would just crash the proxy's routed handler go routine and not the core daemon server process. in that circumstance what would the client see / how would it interpet? or is the idea that this code would never ever be trigged which is why we embed panic to explicitly highlight the extreme irregularity of this conditional being triggered?

api/clients/v2/coretypes/derivation_errors.go
samlaf: This is the convention that I've been following. Like Bowen said its because there is no namespacing for structured errors in golang. So I concatenate the enum type all enums actually... not only errors.

api/clients/v2/payloadretrieval/relay_payload_retriever.go
samlaf: panic is equivalent to returning 500 + logging error (actually stack trace)! See https://pkg.go.dev/net/http#Handler
image

However you did make me realize that if the invariant is ever broken.. probably still better to return the same MaliciousOperatorErr to user anyways... especially in case we ever decide to handle those specially in some middleware.

samlaf: updated: [8b249b6](https://github.com/Layr-Labs/eigenda/pull/1860/commits/8b249b6e87d7cc0d0a5805e3d715d3bd5f6dc4ea)

api/clients/v2/payloadretrieval/validator_payload_retriever.go
samlaf: [2893303](https://github.com/Layr-Labs/eigenda/pull/1860/commits/2893303743f5493d7e963566a144fc149870a13e)

api/clients/v2/coretypes/derivation_errors.go
samlaf: [97ff068](https://github.com/Layr-Labs/eigenda/pull/1860/commits/97ff068961ab432c96d71f0d5291c4cafd6b1a87)

api/clients/v2/coretypes/blob.go
```diff
@@ -28,9 +28,13 @@ type Blob struct {
 
 // DeserializeBlob initializes a Blob from bytes
 func DeserializeBlob(bytes []byte, blobLengthSymbols uint32) (*Blob, error) {
+	blobLengthBytes := blobLengthSymbols * encoding.BYTES_PER_SYMBOL
 	// we check that length of bytes is <= blob length, rather than checking for equality, because it's possible
 	// that the bytes being deserialized have had trailing 0s truncated.
-	if uint32(len(bytes)) > blobLengthSymbols*encoding.BYTES_PER_SYMBOL {
+	if !encoding.IsPowerOfTwo(blobLengthSymbols) {
+		return nil, ErrBlobLengthSymbolsNotPowerOf2
+	}
```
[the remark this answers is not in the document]
samlaf: duplicate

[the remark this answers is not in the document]
samlaf: done while merging master below

api/clients/v2/payloadretrieval/validator_payload_retriever.go
```diff
@@ -147,7 +147,11 @@ func (pr *ValidatorPayloadRetriever) GetEncodedPayload(
 	return nil, fmt.Errorf("unable to retrieve encoded payload with blobKey %v from quorums %v", blobKey.Hex(), blobHeader.QuorumNumbers)
 }
 
-// retrieveBlobWithTimeout attempts to retrieve a blob from a given quorum, and times out based on config.RetrievalTimeout
+// retrieveBlobWithTimeout attempts to retrieve a blob from a given quorum,
+// and times out based on [ValidatorPayloadRetrieverConfig.RetrievalTimeout].
+//
+// blobLengthSymbols MUST be taken from the eigenDACert for the blob being retrieved,
+// and MUST be a power of 2. If not, this function will panic.
```
litt3: > function will panic

No longer true

samlaf: [083408f](https://github.com/Layr-Labs/eigenda/pull/1860/commits/083408f45248b582b5893ff05e13b3d79cedd5fa)

litt3 (approved):

PR #1861, opened 2025-08-11
fix(proxy): memstore uses the new encoded payload data structure

bxue-l2:

Memstore was using the old (codec) methods to store the data. It is incompatible to the [spec](https://layr-labs.github.io/eigenda/integration/spec/6-secure-integration.html#derivation-process) and Hokulea blob decoding procedures cannot produce valid payload.

This PR uses the new EncodedPayload data structure that is recently [merged](https://github.com/Layr-Labs/eigenda/pull/1841). 

This PR has tested against a hokulea [branch](https://github.com/Layr-Labs/hokulea/tree/fix--hokulea-host-using-encoded-payload-from-proxy) that uses encoded payload endpoint to retrieve the data.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/proxy/store/generated_key/memstore/v2/memstore.go
+++ b/api/proxy/store/generated_key/memstore/v2/memstore.go
@@ -226,7 +226,7 @@ func (e *MemStore) Put(_ context.Context, value []byte) ([]byte, error) {
 
 	blobSerialized := blob.Serialize()
 
-	// generateRandomCert produces valid kzg commitment
+	// generateRandomV3Cert produces valid kzg commitment
 	artificialV3Cert, err := e.generateRandomV3Cert(blobSerialized)
 	if err != nil {
 		return nil, fmt.Errorf("generating random cert: %w", err)
--- a/api/proxy/store/generated_key/memstore/v2/memstore.go
+++ b/api/proxy/store/generated_key/memstore/v2/memstore.go
@@ -81,7 +81,7 @@ func New(
 }
 
 // generateRandomCert ... generates a pseudo random EigenDA V3 certificate
-func (e *MemStore) generateRandomV3Cert(blobContents []byte) (coretypes.EigenDACert, error) {
+func (e *MemStore) generateRandomV3Cert(blobContents []byte) (*coretypes.EigenDACertV3, error) {
 	// compute kzg data commitment. this is useful for testing
 	// READPREIMAGE functionality in the arbitrum x eigenda integration since
 	// preimage key is computed within the VM from hashing a recomputation of the data
@@ -226,15 +226,15 @@ func (e *MemStore) Put(_ context.Context, value []byte) ([]byte, error) {
 
 	blobSerialized := blob.Serialize()
 
-	// generateRandomV3Cert produces valid kzg commitment
+	// generateRandomV3Cert produces valid blob commitment on G1
 	artificialV3Cert, err := e.generateRandomV3Cert(blobSerialized)
 	if err != nil {
 		return nil, fmt.Errorf("generating random cert: %w", err)
 	}
 
-	certBytes, err := rlp.EncodeToBytes(artificialV3Cert)
+	certBytes, err := artificialV3Cert.Serialize(coretypes.CertSerializationRLP)
 	if err != nil {
-		return nil, fmt.Errorf("rlp decode v2 cert: %w", err)
+		return nil, fmt.Errorf("rlp decode v3 cert: %w", err)
 	}
 
 	err = e.InsertEntry(crypto.Keccak256Hash(certBytes).Bytes(), blobSerialized)
--- a/api/proxy/store/generated_key/memstore/v2/memstore.go
+++ b/api/proxy/store/generated_key/memstore/v2/memstore.go
@@ -6,12 +6,12 @@ import (
 	"fmt"
 	"math/big"
 
+	"github.com/Layr-Labs/eigenda/api/clients/codecs"
 	"github.com/Layr-Labs/eigenda/api/clients/v2/coretypes"
 	"github.com/consensys/gnark-crypto/ecc/bn254"
 	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/ethereum/go-ethereum/rlp"
 
-	"github.com/Layr-Labs/eigenda/api/clients/codecs"
 	"github.com/Layr-Labs/eigenda/api/clients/v2/verification"
 	"github.com/Layr-Labs/eigenda/api/proxy/common"
 	"github.com/Layr-Labs/eigenda/api/proxy/common/types/certs"
@@ -61,7 +61,8 @@ type MemStore struct {
 	log logging.Logger
 
 	g1SRS []bn254.G1Affine
-	codec codecs.BlobCodec
+
+	polyForm codecs.PolynomialForm
 }
 
 var _ common.EigenDAV2Store = (*MemStore)(nil)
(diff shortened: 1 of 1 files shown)


api/proxy/store/generated_key/memstore/v2/memstore.go
```diff
@@ -180,39 +181,63 @@ func (e *MemStore) generateRandomCert(blobContents []byte) (coretypes.EigenDACer
...
 	}
 
-	certBytes, err := rlp.EncodeToBytes(artificialV2Cert)
+	certBytes, err := rlp.EncodeToBytes(artificialV3Cert)
 	if err != nil {
 		return nil, fmt.Errorf("rlp decode v2 cert: %w", err)
 	}
 
-	err = e.InsertEntry(crypto.Keccak256Hash(certBytes).Bytes(), encodedVal)
+	err = e.InsertEntry(crypto.Keccak256Hash(certBytes).Bytes(), blobSerialized)
```
ethenotethan: just realized that we effectively hard-fork memstore every time we introduce a new cert version. for testability it might be beneficial to support this within memconfig going forward for when we need to test cross-compatibility between multiple cert versions in a rollup derivation pipeline. def out-of-scope of this PR though

api/proxy/store/generated_key/memstore/v2/memstore.go
```diff
@@ -180,39 +181,63 @@ func (e *MemStore) generateRandomCert(blobContents []byte) (coretypes.EigenDACer
...
+	blob, err := payload.ToBlob(e.polyForm)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("generating blob: %w", err)
 	}
 
-	artificialV2Cert, err := e.generateRandomCert(encodedVal)
+	blobSerialized := blob.Serialize()
+
+	// generateRandomV3Cert produces valid kzg commitment
```
ethenotethan: nit - valid data commitment. we use random values for the G2 length commitment and proof

api/proxy/store/generated_key/memstore/v2/memstore.go
```diff
@@ -180,39 +181,63 @@ func (e *MemStore) generateRandomCert(blobContents []byte) (coretypes.EigenDACer
 func (e *MemStore) Get(
 	_ context.Context, versionedCert certs.VersionedCert, returnEncodedPayload bool,
 ) ([]byte, error) {
-	encodedBlob, err := e.FetchEntry(crypto.Keccak256Hash(versionedCert.SerializedCert).Bytes())
+	blobSerialized, err := e.FetchEntry(crypto.Keccak256Hash(versionedCert.SerializedCert).Bytes())
+	if err != nil {
+		return nil, fmt.Errorf("fetching entry via memstore: %w", err)
```
ethenotethan, line 186: @samlaf not for this PR but we should probably define a detailed error for when a blob no longer exists

api/proxy/store/generated_key/memstore/v2/memstore.go
```diff
@@ -180,39 +181,63 @@ func (e *MemStore) generateRandomCert(blobContents []byte) (coretypes.EigenDACer
...
+	// generateRandomV3Cert produces valid kzg commitment
+	artificialV3Cert, err := e.generateRandomV3Cert(blobSerialized)
 	if err != nil {
 		return nil, fmt.Errorf("generating random cert: %w", err)
 	}
 
-	certBytes, err := rlp.EncodeToBytes(artificialV2Cert)
+	certBytes, err := rlp.EncodeToBytes(artificialV3Cert)
 	if err != nil {
 		return nil, fmt.Errorf("rlp decode v2 cert: %w", err)
```
ethenotethan:
```suggestion
		return nil, fmt.Errorf("rlp decode v3 cert: %w", err)
```

api/proxy/store/generated_key/memstore/v2/memstore.go
```diff
@@ -75,12 +76,12 @@ func New(
 		ephemeraldb.New(ctx, config, log),
 		log,
 		g1SRS,
-		codecs.NewIFFTCodec(codecs.NewDefaultBlobCodec()),
+		codecs.PolynomialFormEval,
```
samlaf, line 79: why not use coeffs? Wouldnt that just mean we save on the (I)FFTs?

api/proxy/store/generated_key/memstore/v2/memstore.go
samlaf: only for memstore or always eigenda? How do you tell the difference between a blob having reached TTL and no longer being available, and a blob just never having been stored at all?

api/proxy/store/generated_key/memstore/v2/memstore.go
```diff
@@ -180,39 +181,63 @@ func (e *MemStore) generateRandomCert(blobContents []byte) (coretypes.EigenDACer
...
+	blobSerialized := blob.Serialize()
+
+	// generateRandomV3Cert produces valid kzg commitment
+	artificialV3Cert, err := e.generateRandomV3Cert(blobSerialized)
 	if err != nil {
 		return nil, fmt.Errorf("generating random cert: %w", err)
 	}
 
-	certBytes, err := rlp.EncodeToBytes(artificialV2Cert)
+	certBytes, err := rlp.EncodeToBytes(artificialV3Cert)
```
samlaf: If you change `generateRandomV3Cert` to return a *EigenDACertV3 instead of the cert interface, you'll be able to call `artificialV3Cert.serialize(CertSerializationRLP)` instead of hardcoding the serialization format like this

samlaf (dismissed):

api/proxy/store/generated_key/memstore/v2/memstore.go
bxue-l2: good catch, not aware of it

api/proxy/store/generated_key/memstore/v2/memstore.go
bxue-l2: Would be nice to let recency window to bind TTL

api/proxy/store/generated_key/memstore/v2/memstore.go
bxue-l2: Because in hokulea, we assume IFFT's blob for generating the blob proof for consistency to the kzg commitment. 

Although, our spec is flexible

api/proxy/store/generated_key/memstore/v2/memstore.go
bxue-l2: This is better

api/proxy/store/generated_key/memstore/v2/memstore.go
bxue-l2: 64973e06bfec0e5300af48487a80c3216c68898a

api/proxy/store/generated_key/memstore/v2/memstore.go
bxue-l2: 64973e06bfec0e5300af48487a80c3216c68898a

api/proxy/store/generated_key/memstore/v2/memstore.go
bxue-l2: 64973e06bfec0e5300af48487a80c3216c68898a

api/proxy/store/generated_key/memstore/v2/memstore.go
ethenotethan: > How do you tell the difference between a blob having reached TTL and no longer being available, and a blob just never having been stored at all?

that is a question which warrants deeper analysis

ethenotethan (approved):
LGTM 🚀

PR #1863, opened 2025-08-12
refactor: expose payload disperser probe to proxy

iquidus:
Closes DAINT-680

Refactors the proxy metrics constructor:
* Removes the internal call to `prometheus.NewRegistry()`
* Adds `registry` argument

Passes the `prometheus.Registry` created in `StartProxySvr` down to `NewPayloadDisperser` in order to create a `StageTimer`

The disperser probe is not currently exposed to the proxy as the registry required to construct the `StageTimer` is set to `nil`.

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Integration tests
   - [x] Local e2e tests

### DIFF
--- a/api/proxy/metrics/metrics.go
+++ b/api/proxy/metrics/metrics.go
@@ -44,9 +44,9 @@ type Metrics struct {
 
 var _ Metricer = (*Metrics)(nil)
 
-func NewMetrics(registry *prometheus.Registry) *Metrics {
+func NewMetrics(registry *prometheus.Registry) Metricer {
 	if registry == nil {
-		return nil
+		return NoopMetrics
 	}
 
 	registry.MustRegister(collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}))
--- a/api/proxy/metrics/metrics.go
+++ b/api/proxy/metrics/metrics.go
@@ -15,6 +15,7 @@ import (
 
 const (
 	namespace           = "eigenda_proxy"
+	subsystem           = "default"
 	httpServerSubsystem = "http_server"
 	secondarySubsystem  = "secondary"
 )
@@ -57,12 +58,9 @@ type Metrics struct {
 
 var _ Metricer = (*Metrics)(nil)
 
-func NewMetrics(registry *prometheus.Registry, subsystem string) *Metrics {
+func NewMetrics(registry *prometheus.Registry) *Metrics {
 	if registry == nil {
 		return nil
-	}
-	if subsystem == "" {
-		subsystem = "default"
 	}
 
 	registry.MustRegister(collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}))
--- a/api/proxy/metrics/metrics.go
+++ b/api/proxy/metrics/metrics.go
@@ -57,12 +57,14 @@ type Metrics struct {
 
 var _ Metricer = (*Metrics)(nil)
 
-func NewMetrics(subsystem string) *Metrics {
+func NewMetrics(registry *prometheus.Registry, subsystem string) *Metrics {
+	if registry == nil {
+		return nil
+	}
 	if subsystem == "" {
 		subsystem = "default"
 	}
 
-	registry := prometheus.NewRegistry()
 	registry.MustRegister(collectors.NewProcessCollector(collectors.ProcessCollectorOpts{}))
 	registry.MustRegister(collectors.NewGoCollector())
 	factory := metrics.With(registry)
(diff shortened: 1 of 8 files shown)


api/proxy/metrics/metrics.go
```diff
@@ -57,12 +57,14 @@ type Metrics struct {
 
 var _ Metricer = (*Metrics)(nil)
 
-func NewMetrics(subsystem string) *Metrics {
+func NewMetrics(registry *prometheus.Registry, subsystem string) *Metrics {
+	if registry == nil {
+		return nil
```
ethenotethan: is returning nil here safe? iiuc that nil value will be passed via dependency injection to proxy components and when callable will cause the handler routines to fail nil reference exceptions. I think we may want to expose the `NoopMetrics` object instead since it stubs the interface's methods to do nothing

ethenotethan: oh looks like copilot also caught this too haha

api/proxy/metrics/metrics.go
```diff
@@ -57,12 +57,14 @@ type Metrics struct {
 
 var _ Metricer = (*Metrics)(nil)
 
-func NewMetrics(subsystem string) *Metrics {
+func NewMetrics(registry *prometheus.Registry, subsystem string) *Metrics {
```
samlaf: is there any reason to keep the "subsystem" around? This would be useful if we reused this struct for different subsystems (as for logs for example), but afaiu this metrics struct is only ever created once, for the proxy, and we just use "default" subsystem. No need in having it as an argument imo

samlaf (commented):
LGTM, but please address ethen's comment

api/proxy/metrics/metrics.go
iquidus: Yes, will fix shortly. `Metrics.StartServer` func is not part of the `Metricer` or `NoopMetrics` so needs a little more work than simply returning the noop.

samlaf: Might be a good time to actually make this refactor then. Imo the metrics struct that keeps track of counters/gauges/histograms, and the metrics server which exposes them on an http endpoint, should be separate.

Specifically because the proxy server exposes a bunch of metrics from clients etc, not only proxy metrics.

iquidus: StartServer refactored in: [b1eee30](https://github.com/Layr-Labs/eigenda/pull/1863/commits/b1eee30d2637f9e741d375d35d4a2af942783ccd)

Noop instead of nil in: [cb9de4f](https://github.com/Layr-Labs/eigenda/pull/1863/commits/cb9de4fc1c375cf19f0854e41ec11bbef46d912f)

api/proxy/metrics/metrics.go
iquidus: Removed subsystem arg in: [e2f81e3](https://github.com/Layr-Labs/eigenda/pull/1863/commits/e2f81e3e1d1986a53d5f4439a28c93f3febc7b67)

api/proxy/metrics/metrics.go
```diff
@@ -15,6 +15,7 @@ import (
 
 const (
 	namespace           = "eigenda_proxy"
+	subsystem           = "default"
```
samlaf, line 11: let's change to namespace="eigenda" and subsystem="proxy" instead?

samlaf: hmm actually I'm not sure how people would want to run these in their infra... maybe should ask celo before changing this. nvm

api/proxy/metrics/server.go
```diff
@@ -0,0 +1,27 @@
...
+}
+
+func NewServer(registry *prometheus.Registry, cfg Config) *ophttp.HTTPServer {
+	address := net.JoinHostPort(cfg.Host, strconv.Itoa(cfg.Port))
+
+	h := promhttp.InstrumentMetricHandler(
+		registry, promhttp.HandlerFor(registry, promhttp.HandlerOpts{}),
+	)
+
+	return ophttp.NewHTTPServer(address, h)
```
samlaf, line 26: what does this ophttp provide? If its small can we copy the code into our repo? Otherwise can we just switch to using the golang http library directly

api/proxy/metrics/metrics.go
samlaf: lgtm

api/proxy/metrics/server.go
iquidus: https://github.com/ethereum-optimism/optimism/blob/develop/op-service/httputil/server.go

It wraps http.Server and adds some additional connivance functions. Plan to remove in a future PR.

PR #1864, opened 2025-08-12
feat(metrics): Add V2 accountant metrics

iquidus:
Closes DAINT-655

Adds the scaffolding for v2 accountant metrics, starting with cumulative payment. More to follow after feedback.

Note: Depends on https://github.com/Layr-Labs/eigenda/pull/1863

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Integration tests
   - [x] Local e2e tests

### DIFF
--- a/api/clients/v2/accountant.go
+++ b/api/clients/v2/accountant.go
@@ -47,6 +47,10 @@ type PeriodRecord struct {
 	Usage uint64
 }
 
+func NewUnpopulatedAccountant(accountID gethcommon.Address, metrics metrics.AccountantMetricer) *Accountant {
+	return NewAccountant(accountID, nil, nil, 0, 0, 0, 0, metrics)
+}
+
 func NewAccountant(
 	accountID gethcommon.Address,
 	reservation *core.ReservedPayment,
--- a/api/proxy/store/builder/storage_manager_builder.go
+++ b/api/proxy/store/builder/storage_manager_builder.go
@@ -583,7 +583,7 @@ func buildPayloadDisperser(
 
 	accountantMetrics := metrics_v2.NewAccountantMetrics(registry)
 	// The accountant is populated lazily by disperserClient.PopulateAccountant
-	accountant := clients_v2.NewAccountant(accountId, nil, nil, 0, 0, 0, 0, accountantMetrics)
+	accountant := clients_v2.NewUnpopulatedAccountant(accountId, accountantMetrics)
 
 	disperserClient, err := clients_v2.NewDisperserClient(
 		&clientConfigV2.DisperserClientCfg,
--- a/test/v2/client/test_client.go
+++ b/test/v2/client/test_client.go
@@ -149,7 +149,7 @@ func NewTestClient(
 		UseSecureGrpcFlag: true,
 	}
 
-	accountant := clientsv2.NewAccountant(accountId, nil, nil, 0, 0, 0, 0, metricsv2.NoopAccountantMetrics)
+	accountant := clientsv2.NewUnpopulatedAccountant(accountId, metricsv2.NoopAccountantMetrics)
 	disperserClient, err := clientsv2.NewDisperserClient(disperserConfig, signer, kzgProver, accountant)
 	if err != nil {
 		return nil, fmt.Errorf("failed to create disperser client: %w", err)
--- a/test/v2/live/live_network_test.go
+++ b/test/v2/live/live_network_test.go
@@ -499,7 +499,7 @@ func dispersalWithInvalidSignatureTest(t *testing.T, environment string) {
 		UseSecureGrpcFlag: true,
 	}
 
-	accountant := clients.NewAccountant(accountId, nil, nil, 0, 0, 0, 0, metrics.NoopAccountantMetrics)
+	accountant := clients.NewUnpopulatedAccountant(accountId, metrics.NoopAccountantMetrics)
 	disperserClient, err := clients.NewDisperserClient(disperserConfig, signer, nil, accountant)
 	require.NoError(t, err)
 


api/clients/v2/disperser_client.go
```diff
@@ -420,3 +424,7 @@ func (c *disperserClient) initOncePopulateAccountant(ctx context.Context) error
 	}
 	return nil
 }
+
+func (c *disperserClient) SetAccountantMetrics(metrics metrics.AccountantMetricer) {
+	c.accountantMetrics = metrics
```
ethenotethan: what's the reason for adding a setter vs having this field defined in the constructor function? iiuc we don't ever expect this field to change after account object init

api/clients/v2/metrics/accountant.go
```diff
@@ -0,0 +1,57 @@
...
+)
+
+const (
+	namespace           = "eigenda"
+	accountantSubsystem = "accountant"
+)
+
+var (
+	gweiFactor = big.NewInt(1000000000)
+)
```
ethenotethan: nit - can we add a comment explaining why we chose gwei and that users can perform transformations on the value via dashboard functions to change denomination if necessary?

api/proxy/metrics/metrics.go
```diff
@@ -57,12 +57,14 @@ type Metrics struct {
 
 var _ Metricer = (*Metrics)(nil)
 
-func NewMetrics(subsystem string) *Metrics {
+func NewMetrics(registry *prometheus.Registry, subsystem string) *Metrics {
+	if registry == nil {
+		return nil
+	}
```
ethenotethan: saw this in another one of your prs and think this is dangerous

api/clients/v2/metrics/accountant.go
```diff
@@ -0,0 +1,57 @@
...
+		}, []string{
+			"account_id",
+		}),
+	}
+}
+
+func (m *AccountantMetrics) RecordCumulativePayment(accountID string, wei *big.Int) {
+	// Convert wei to gwei
+	gwei := new(big.Int).Div(wei, gweiFactor)
+	gweiFloat, _ := gwei.Float64()
```
ethenotethan: why even convert - doing denomination changes on dashboard should be pretty intuitive for user. feels like we're making an arbitrary selection to do additional computations that aren't really necessary imo

api/proxy/metrics/metrics.go
samlaf: Agree this feels bug-prone

api/clients/v2/metrics/accountant.go
samlaf: We discussed this during standup a bit. Think the idea is that prometheus uses float64 for metrics, which starting at big numbers has "holes". Not sure if those holes become meaningful around 1e18 (weis) though? Agree that it would be good to at least document.

api/proxy/metrics/metrics.go
samlaf: Think it was just coming from the other PR actually? I changed the base branch of this PR to target `expose-disperser-probe`. @iquidus good practice to do this so we don't do double reviews.

api/clients/v2/disperser_client.go
iquidus: To avoid changing the function signature of the constructor for anyone importing the module (outside of proxy). If this is not an issue, will move to constructor.

api/clients/v2/metrics/accountant.go
iquidus: Yes, the go client for prometheus uses float64, so we need to convert the big.Int to a float64. After discussion gwei made the most sense. Will document.

api/proxy/metrics/metrics.go
iquidus: Thanks Sam. Yes, this commit was a duplicate. Will fix in other PR.

api/clients/v2/disperser_client.go
```diff
@@ -96,10 +98,11 @@ func NewDisperserClient(config *DisperserClientConfig, signer corev2.BlobRequest
...
 	return &disperserClient{
-		config:     config,
-		signer:     signer,
-		prover:     prover,
-		accountant: accountant,
+		config:            config,
+		signer:            signer,
+		prover:            prover,
+		accountant:        accountant,
+		accountantMetrics: metrics.NoopAccountantMetrics,
```
samlaf: shouldn't be stored here since its only relevant to accountant. inject it in accountant directly upstream

api/clients/v2/disperser_client.go
```diff
@@ -112,6 +115,7 @@ func (c *disperserClient) PopulateAccountant(ctx context.Context) error {
 			return fmt.Errorf("error getting account ID: %w", err)
 		}
 		c.accountant = NewAccountant(accountId, nil, nil, 0, 0, 0, 0)
+		c.accountant.SetMetrics(c.accountantMetrics)
```
samlaf: This reads wrong, but I'm guessing you wrote this because of the weirdness of this PopulateAccountant function?

Haven't analyzed this in depth, but I think in general we should stop doing implicit things when some pointers are set to nil, and instead error hard. The `if c.account == nil` above seems misguided and bug prone. Let's instead make sure the accountant is always set in the constructor, and then just return an error in PopulateAccountant if it isn't (should even panic imo to indicate its a bug, not a normal error).

api/clients/v2/disperser_client.go
samlaf: Agree, dependency injection is the way. Injected structs should be fully populated to prevent bugs.

samlaf: > To avoid changing the function signature of the constructor for anyone importing the module (outside of proxy). If this is not an issue, will move to constructor.

Ah just saw your message. I'd say let's not worry about these kinds of breaking changes. Accountant is being rewritten anyways, and our codebase just doesn't have the kind of maturity to be that stable just yet. Not worth making our APIs more complex and bug-prone.

iquidus:
Rebased on top of master.

api/clients/v2/disperser_client.go
iquidus: Addressed in [f7ca187](https://github.com/Layr-Labs/eigenda/pull/1864/commits/f7ca1876db40a291313cba58282239304ad6fd9e)

api/clients/v2/disperser_client.go
iquidus: See [f7ca187](https://github.com/Layr-Labs/eigenda/pull/1864/commits/f7ca1876db40a291313cba58282239304ad6fd9e)

api/clients/v2/disperser_client.go
iquidus: See [f7ca187](https://github.com/Layr-Labs/eigenda/pull/1864/commits/f7ca1876db40a291313cba58282239304ad6fd9e)

api/clients/v2/metrics/accountant.go
```diff
@@ -0,0 +1,57 @@
...
```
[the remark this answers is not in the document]
iquidus: Do we want this warning? I'm inclined to think no, but curious about other opinions.

api/clients/v2/metrics/accountant.go
iquidus: See [71db322](https://github.com/Layr-Labs/eigenda/pull/1864/commits/71db322c307cd33c8db853037a9e57514cae9930)

api/clients/v2/metrics/accountant.go
iquidus: See [71db322](https://github.com/Layr-Labs/eigenda/pull/1864/commits/71db322c307cd33c8db853037a9e57514cae9930)

api/clients/v2/metrics/accountant.go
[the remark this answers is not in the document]
samlaf: no think its fine without

PR #1866, opened 2025-08-12
fix: semaphore use for StoreChunks

cody-littley:
Fixes the way StoreChunks() uses a semaphore, old pattern never released semaphore.

### DIFF
--- a/node/config.go
+++ b/node/config.go
@@ -412,7 +412,7 @@ func NewConfig(ctx *cli.Context) (*Config, error) {
 		LittDBReadCacheSizeBytes:      uint64(ctx.GlobalFloat64(flags.LittDBReadCacheSizeGBFlag.Name) * units.GiB),
 		LittDBReadCacheSizeFraction:   ctx.GlobalFloat64(flags.LittDBReadCacheSizeFractionFlag.Name),
 		LittDBStoragePaths:            ctx.GlobalStringSlice(flags.LittDBStoragePathsFlag.Name),
-		LittRespectLocks:              ctx.GlobalBool(flags.LitRespectLocksFlag.Name),
+		LittRespectLocks:              ctx.GlobalBool(flags.LittRespectLocksFlag.Name),
 		DownloadPoolSize:              ctx.GlobalInt(flags.DownloadPoolSizeFlag.Name),
 		GetChunksHotCacheReadLimitMB:  ctx.GlobalFloat64(flags.GetChunksHotCacheReadLimitMBFlag.Name),
 		GetChunksHotBurstLimitMB:      ctx.GlobalFloat64(flags.GetChunksHotBurstLimitMBFlag.Name),
--- a/node/flags/flags.go
+++ b/node/flags/flags.go
@@ -451,7 +451,7 @@ var (
 		Required: false,
 		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "LITT_DB_STORAGE_PATHS"),
 	}
-	LitRespectLocksFlag = cli.BoolFlag{
+	LittRespectLocksFlag = cli.BoolFlag{
 		Name: common.PrefixFlag(FlagPrefix, "litt-respect-locks"),
 		Usage: "If set, LittDB will refuse to start if it can't acquire locks on the storage paths. " +
 			"Ideally this would always be enabled, but PID reuse in platforms like Kubernetes/Docker can make " +
@@ -670,7 +670,7 @@ var optionalFlags = []cli.Flag{
 	EigenDADirectoryFlag,
 	BlsOperatorStateRetrieverFlag,
 	EigenDAServiceManagerFlag,
-	LitRespectLocksFlag,
+	LittRespectLocksFlag,
 	StoreChunksBufferTimeoutFlag,
 	StoreChunksBufferSizeGBFlag,
 	StoreChunksBufferSizeFractionFlag,


node/node_v2.go
```diff
@@ -33,54 +33,43 @@ type RawBundle struct {
...
-	ctx context.Context,
+// Determines where to find the chunks we need to download for a given batch.
+func (n *Node) DetermineChunkLocations(
 	batch *corev2.Batch,
 	operatorState *core.OperatorState,
 	probe *common.SequenceProbe,
-) ([]*corev2.BlobShard, []*RawBundle, error) {
+) (downloadSizeInBytes uint64, relayRequests map[corev2.RelayKey]*relayRequest, err error) {
 
 	probe.SetStage("prepare_to_download")
```
litt3: is this still a good stage name?

node/node_v2.go
```diff
@@ -115,27 +104,38 @@ func (n *Node) DownloadBundles(
...
+	return downloadSizeInBytes, relayRequests, nil
+}
 
-		probe.SetStage("acquire_buffer_capacity")
-		semaphoreCtx, cancel := context.WithTimeout(ctx, n.Config.StoreChunksBufferTimeout)
-		defer cancel()
-		err := n.storeChunksSemaphore.Acquire(semaphoreCtx, int64(downloadSizeInBytes))
-		if err != nil {
-			return nil, nil, fmt.Errorf("failed to acquire buffer capacity: %w", err)
+func (n *Node) DownloadChunksFromRelays(
```
litt3, line 109: docs for `DownloadChunksFromRelays` and `DetermineChunkLocations` could use a little love, but the lack of docs is a preexisting condition, so up to you if you want to defer that

node/node_v2.go
```diff
@@ -33,54 +33,43 @@ type RawBundle struct {
...
+func (n *Node) DetermineChunkLocations(
 	batch *corev2.Batch,
 	operatorState *core.OperatorState,
 	probe *common.SequenceProbe,
-) ([]*corev2.BlobShard, []*RawBundle, error) {
+) (downloadSizeInBytes uint64, relayRequests map[corev2.RelayKey]*relayRequest, err error) {
 
 	probe.SetStage("prepare_to_download")
 
 	relayClient, ok := n.RelayClient.Load().(relay.RelayClient)
```
litt3: unused

node/grpc/server_v2.go
```diff
@@ -182,9 +182,33 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
...
+	// storeChunksSemaphore can be nil during unit tests, since there are a bunch of places where the Node struct
+	// is instantiated directly without using the constructor.
+	if s.node.StoreChunksSemaphore != nil {
+		// So far, we've only downloaded metadata for the blob. Before downloading the actual chunks, make sure there
+		// is capacity in the store chunks buffer. This is an OOM safety measure.
+
+		probe.SetStage("acquire_buffer_capacity")
+		semaphoreCtx, cancel := context.WithTimeout(ctx, s.node.Config.StoreChunksBufferTimeout)
+		defer cancel()
+		err = s.node.StoreChunksSemaphore.Acquire(semaphoreCtx, int64(downloadSizeInBytes))
```
anupsv, line 201: This blocks until the request size in bytes is free or the semaphoreCtx is `done` right. This essentially means the storechunks request from the disperser fails incase the previous requests are still being processed or so ? Does this mean disperser needs to retry on this error ?

litt3 (dismissed):

node/grpc/server_v2.go
cody-littley: > This essentially means the storechunks request from the disperser fails incase the previous requests are still being processed 

Correct. The intention of this change is to prevent too many chunks from being in memory at a single instant in time.

> Does this mean disperser needs to retry on this error ?

I've currently got dispersal retries disabled, as the current implementation is extremely inefficient. So if this triggers, the validator will not end up signing for some batches. I think this is ok though... if the disperser is sending more work than the validator can handle, it NEEDS to skip some batches or else it will accumulate a backlog that will eventually lead to OOM.

node/node_v2.go
cody-littley: changed to `determine_chunk_locations`

node/node_v2.go
cody-littley: expanded docs

node/node_v2.go
cody-littley: removed

node/grpc/server_v2.go
anupsv: So essentially failed dispersals is what we are agreeing to.

anupsv (dismissed):

litt3 (dismissed):

node/grpc/server_v2.go
cody-littley: Possibly worth having a discussion on this.

I'm not convinced it's possible to have a robust, high performance system without the ability shed load when under high stress. It's the difference between a system that recovers from a spike too large to handle, and a system that face plants when traffic spikes above a critical threshold.

node/config.go
```diff
@@ -424,7 +412,7 @@ func NewConfig(ctx *cli.Context) (*Config, error) {
 		LittDBReadCacheSizeBytes:      uint64(ctx.GlobalFloat64(flags.LittDBReadCacheSizeGBFlag.Name) * units.GiB),
 		LittDBReadCacheSizeFraction:   ctx.GlobalFloat64(flags.LittDBReadCacheSizeFractionFlag.Name),
 		LittDBStoragePaths:            ctx.GlobalStringSlice(flags.LittDBStoragePathsFlag.Name),
-		LittUnsafePurgeLocks:          ctx.GlobalBool(flags.LittUnsafePurgeLocksFlag.Name),
+		LittRespectLocks:              ctx.GlobalBool(flags.LitRespectLocksFlag.Name),
```
litt3:
```suggestion
		LittRespectLocks:              ctx.GlobalBool(flags.LittRespectLocksFlag.Name),
```

node/flags/flags.go
```diff
@@ -451,11 +451,13 @@ var (
 		Required: false,
 		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "LITT_DB_STORAGE_PATHS"),
 	}
-	LittUnsafePurgeLocksFlag = cli.BoolFlag{
-		Name:     common.PrefixFlag(FlagPrefix, "litt-unsafe-purge-locks"),
-		Usage:    "Unsafe flag to purge locks in LittDB. Use with caution, as it may lead to data loss or corruption.",
+	LitRespectLocksFlag = cli.BoolTFlag{
```
litt3:
```suggestion
	LittRespectLocksFlag = cli.BoolTFlag{
```

litt3: Why use `BoolTFlag` instead of `BoolFlag`?

node/config.go
cody-littley: fixed

node/flags/flags.go
cody-littley: fixed

cody-littley: This is now a `BoolFlag` (I caught this when I tested it)

PR #1870, opened 2025-08-14
fix: Use floats for leaky bucket

litt3:
- Previous impl had a bug to do with how monotonic clocks work
- Looking at the solution to that bug, the total complexity of the int based approach exceeded the benefit, IMO
- Since floats are tolerable for this utility, I decided to simply cut losses and get rid of the integer based solution

### DIFF
--- a/core/payments/reservation/leaky_bucket.go
+++ b/core/payments/reservation/leaky_bucket.go
@@ -159,7 +159,7 @@ func (lb *LeakyBucket) leak(now time.Time) error {
 	elapsed := now.Sub(lb.previousLeakTime)
 
 	if elapsed < 0 {
-		// This shouldn't happen since we're using monotonic time. But just in case...
+		// This can only happen if the user passes in time instances without monotonic timestamps
 		return &TimeMovedBackwardError{PreviousTime: lb.previousLeakTime, CurrentTime: now}
 	}
 


core/payments/reservation/leaky_bucket.go
```diff
@@ -64,8 +41,8 @@ func NewLeakyBucket(
 	symbolsPerSecondLeakRate uint64,
 	// bucketCapacityDuration * symbolsPerSecondLeakRate becomes the bucket capacity
 	bucketCapacityDuration time.Duration,
-	// whether to err on the side of permitting more or less throughput
-	biasBehavior BiasBehavior,
+	// whether the bucket should start full or empty
+	startFull bool,
```
samlaf, line 45: yay :)

core/payments/reservation/leaky_bucket.go
```diff
@@ -175,105 +137,39 @@ func (lb *LeakyBucket) RevertFill(now time.Time, symbolCount uint32) error {
...
-	lb.currentFillLevel = lb.currentFillLevel - uint64(symbolCount)
 	return nil
 }
 
 // Lets the correct number of symbols leak out of the bucket, based on when we last leaked
 //
 // Returns a TimeMovedBackwardError if input time is before previous leak time.
 func (lb *LeakyBucket) leak(now time.Time) error {
-	if now.Before(lb.previousLeakTime) {
+	elapsed := now.Sub(lb.previousLeakTime)
```
samlaf, line 159: nit: do we want to add the recommendation that users should create time instances that contain a monotonic clock reading, for better results?

core/payments/reservation/leaky_bucket.go
```diff
@@ -175,105 +137,39 @@ func (lb *LeakyBucket) RevertFill(now time.Time, symbolCount uint32) error {
...
+
+	if elapsed < 0 {
+		// This shouldn't happen since we're using monotonic time. But just in case...
```
samlaf: we aren't necessarily right? We would need to check if the passed in `now` time contains monotonic bits

core/payments/reservation/leaky_bucket.go
litt3: 8cb78b0d4e236dc279f20d1c84f56efd6896abc6

core/payments/reservation/leaky_bucket.go
litt3: tweaked comment 1850d017091496aafe3fd3680b455387a0d2dbef

PR #1873, opened 2025-08-14
feat(payments): Implement reservation ledger

litt3:
Implements a `ReservationLedger` wrapper around the `LeakyBucket`

Closes EGDA-1852

### DIFF
--- a/core/payments/reservation/errors.go
+++ b/core/payments/reservation/errors.go
@@ -1,25 +1,18 @@
 package reservation
 
 import (
-	"fmt"
-	"time"
+	"errors"
 )
 
-// TimeMovedBackwardError is returned when a timestamp is observed that is before a previously observed timestamp.
-//
-// This should not normally happen, but with clock drift and NTP adjustments, system clocks can occasionally jump
-// backward. This error allows the system to handle such cases gracefully rather than fatally erroring.
-type TimeMovedBackwardError struct {
-	// The current time that was provided
-	CurrentTime time.Time
-	// The previously observed time that is after CurrentTime
-	PreviousTime time.Time
-}
+// Sentinel errors for reservation operations
+var (
+	// ErrQuorumNotPermitted indicates that one or more requested quorums are not permitted by the reservation.
+	ErrQuorumNotPermitted = errors.New("quorum not permitted")
 
-// Implements the error interface
-func (e *TimeMovedBackwardError) Error() string {
-	return fmt.Sprintf("time moved backward: current time %s is before previous time %s (delta: %v)",
-		e.CurrentTime.Format(time.RFC3339Nano),
-		e.PreviousTime.Format(time.RFC3339Nano),
-		e.PreviousTime.Sub(e.CurrentTime))
-}
+	// ErrTimeOutOfRange indicates the dispersal time is outside the reservation's valid time range.
+	ErrTimeOutOfRange = errors.New("time outside reservation range")
+
+	// ErrTimeMovedBackward indicates a timestamp was observed that is before a previously observed timestamp.
+	// This is unlikely to happen in practice, but could be the result of clock drift and NTP adjustments.
+	ErrTimeMovedBackward = errors.New("time moved backward")
+)
--- a/core/payments/reservation/reservation.go
+++ b/core/payments/reservation/reservation.go
@@ -1,0 +1,94 @@
+package reservation
+
+import (
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/Layr-Labs/eigenda/core"
+)
+
+// Represents a reservation for a single user.
+//
+// TODO(litt3): I opted to duplicate the preexisting `ReservedPayment` struct, rather than using the old one. There
+// are nontrivial changes I wanted to make, and making those changes in a way that's compatible with the preexisting
+// usages was going to be messy. Instead, `ReservedPayment` can just be removed, when we remove the deprecated payment
+// system.
+type Reservation struct {
+	// The number of symbols / second that the holder of this reservation is entitled to disperse
+	symbolsPerSecond uint64
+
+	// The time at which the reservation becomes active
+	startTime time.Time
+
+	// The time at which the reservation expires
+	endTime time.Time
+
+	// The quorums that the holder of this reservation is entitled to disperse to
+	permittedQuorumIDs map[core.QuorumID]bool
+}
+
+// Create a representation of a single account Reservation.
+func NewReservation(
+	symbolsPerSecond uint64,
+	startTime time.Time,
+	endTime time.Time,
+	permittedQuorumIDs []core.QuorumID,
+) (*Reservation, error) {
+	if symbolsPerSecond == 0 {
+		return nil, errors.New("reservation must have >0 symbols per second")
+	}
+
+	if startTime.Equal(endTime) || endTime.Before(startTime) {
+		return nil, fmt.Errorf("start time (%v) must be before end time (%v)", startTime, endTime)
+	}
+
+	permittedQuorumIDsLen := len(permittedQuorumIDs)
+	if permittedQuorumIDsLen == 0 {
+		return nil, errors.New("reservation must permit at least one quorum")
+	}
+
+	permittedQuorumIDSet := make(map[core.QuorumID]bool, permittedQuorumIDsLen)
+	for _, quorumID := range permittedQuorumIDs {
+		permittedQuorumIDSet[quorumID] = true
+	}
+
+	return &Reservation{
+		symbolsPerSecond:   symbolsPerSecond,
+		startTime:          startTime,
+		endTime:            endTime,
+		permittedQuorumIDs: permittedQuorumIDSet,
+	}, nil
+}
+
+// Checks whether an input list of quorums are all permitted by the reservation.
+//
+// Returns nil if all input quorums are permitted, otherwise returns ErrQuorumNotPermitted.
+func (r *Reservation) CheckQuorumsPermitted(quorums []core.QuorumID) error {
+	for _, quorum := range quorums {
+		if !r.permittedQuorumIDs[quorum] {
+			permittedQuorums := make([]core.QuorumID, 0, len(r.permittedQuorumIDs))
+			for quorumID := range r.permittedQuorumIDs {
+				permittedQuorums = append(permittedQuorums, quorumID)
+			}
+			return fmt.Errorf("%w: quorum %d not in permitted set %v", ErrQuorumNotPermitted, quorum, permittedQuorums)
+		}
+	}
+
+	return nil
+}
+
+// Verifies that the given time falls within the reservation's valid time range.
+//
+// Returns ErrTimeOutOfRange if the time is outside the valid range.
+func (r *Reservation) CheckTime(timeToCheck time.Time) error {
+	if timeToCheck.Before(r.startTime) || timeToCheck.After(r.endTime) {
+		return fmt.Errorf("%w: dispersal time %s is outside range [%s, %s]",
+			ErrTimeOutOfRange,
+			timeToCheck.Format(time.RFC3339),
+			r.startTime.Format(time.RFC3339),
+			r.endTime.Format(time.RFC3339))
+	}
+
+	return nil
+}
--- a/core/payments/reservation/reservation_ledger.go
+++ b/core/payments/reservation/reservation_ledger.go
@@ -1,0 +1,98 @@
+package reservation
+
+import (
+	"fmt"
+	"sync"
+	"time"
+
+	"github.com/Layr-Labs/eigenda/core"
+)
+
+// Keeps track of the state of a given reservation
+//
+// This is a goroutine safe wrapper around the LeakyBucket algorithm.
+type ReservationLedger struct {
+	config ReservationLedgerConfig
+
+	// synchronizes access to the underlying leaky bucket algorithm
+	lock sync.Mutex
+
+	// an instance of the algorithm which tracks reservation usage
+	leakyBucket *LeakyBucket
+}
+
+// Creates a new reservation ledger, which represents the reservation of a single user with a leaky bucket
+func NewReservationLedger(
+	config ReservationLedgerConfig,
+	now time.Time,
+) (*ReservationLedger, error) {
+	leakyBucket, err := NewLeakyBucket(
+		config.reservation.symbolsPerSecond,
+		config.bucketCapacityDuration,
+		config.biasBehavior,
+		config.overfillBehavior,
+		now,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("new leaky bucket: %w", err)
+	}
+
+	return &ReservationLedger{
+		config:      config,
+		leakyBucket: leakyBucket,
+	}, nil
+}
+
+// Debit the reservation with a number of symbols.
+//
+// Algorithmically, that means adding a number of symbols to the leaky bucket.
+//
+// Returns (true, nil) if the reservation has enough capacity to perform the debit.
+// Returns (false, nil) if the bucket lacks capacity to permit the fill.
+// Returns (false, error) if an error occurs. Possible errors include:
+//   - ErrQuorumNotPermitted: requested quorums are not permitted by the reservation
+//   - ErrTimeOutOfRange: dispersal time is outside the reservation's valid time range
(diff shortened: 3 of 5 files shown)


core/payments/reservation/reservation.go
```diff
@@ -0,0 +1,96 @@
...
+	symbolsPerSecond uint64
+
+	// The time at which the reservation becomes active
+	startTime time.Time
+
+	// The time at which the reservation expires
+	endTime time.Time
+
+	// The quorums that the holder of this reservation is entitled to disperse to
+	permittedQuorumIDs map[core.QuorumID]bool
```
cody-littley:
```suggestion
	permittedQuorumIDs map[core.QuorumID]struct{}
```

I prefer `struct{}` instead of `bool` for creating the abstraction of a set. I'm under the impression that it's slightly more efficient (not that the difference makes a meaningful difference here). Additionally, it very clearly marks it as a set, as there is no way to convey any information in the map values.

core/payments/reservation/reservation.go
```diff
@@ -0,0 +1,96 @@
...
+		permittedQuorumIDs: permittedQuorumIDSet,
+	}, nil
+}
+
+// Checks whether an input list of quorums are all permitted by the reservation.
+//
+// Returns nil if all input quorums are permitted, otherwise returns ErrQuorumNotPermitted.
+func (r *Reservation) CheckQuorumsPermitted(quorums []core.QuorumID) error {
+	for _, quorum := range quorums {
+		if r.permittedQuorumIDs[quorum] {
```
cody-littley: If you switch to `struct{}`, this would become as follows:
```suggestion
		if _, ok := r.permittedQuorumIDs[quorum]; ok {
```

core/payments/reservation/reservation_ledger.go
```diff
@@ -0,0 +1,117 @@
...
+	"time"
+
+	"github.com/Layr-Labs/eigenda/core"
+)
+
+// Tracks usage of a single account reservation
+//
+// This struct is goroutine safe.
+type ReservationLedger struct {
+	config ReservationLedgerConfig
```
cody-littley, line 15: Any particular reason why you don't use a `*ReservationLedgerConfig`? Although copying a struct is usually not a big deal for tiny structs, I typically make a habit of always passing them around by reference. That way, I'm never surprised if I end up passing around a struct that is expensive to copy.

core/payments/reservation/reservation_ledger.go
```diff
@@ -0,0 +1,117 @@
...
+	// synchronizes access to the underlying leaky bucket algorithm
+	lock sync.Mutex
+
+	// an instance of the algorithm which tracks reservation usage
+	leakyBucket *LeakyBucket
+}
+
+// Creates a new reservation ledger, which represents the reservation of a single user with a leaky bucket
+func NewReservationLedger(
+	config ReservationLedgerConfig,
```
cody-littley, line 26:
```suggestion
	config *ReservationLedgerConfig,
```

core/payments/reservation/reservation_ledger.go
```diff
@@ -0,0 +1,117 @@
...
+	success, err := rl.leakyBucket.Fill(now, symbolCount)
+	if err != nil {
+		return false, fmt.Errorf("fill: %w", err)
+	}
+
+	return success, nil
+}
+
+// Credit the reservation with a number of symbols. This method "undoes" a previous debit, following a failed dispersal.
+func (rl *ReservationLedger) RevertDebit(now time.Time, symbolCount uint32) error {
```
cody-littley: Perhaps note that it might not perfectly undo things if time has passed, since a bucket is never allowed to go negative. So a revert does not necessarily mean you will actually be able to utilize all of the bandwidth you just reverted.

core/payments/reservation/reservation_ledger_config.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
+	"fmt"
+	"time"
+)
+
+// Configuration for a ReservationLedger, which manages the reservation of a single account
+type ReservationLedgerConfig struct {
+	// Contains the parameters of the reservation that the ReservationLedger is responsible for
+	reservation Reservation
```
cody-littley, line 11:
```suggestion
	reservation *Reservation
```

cody-littley (approved):

core/payments/reservation/reservation_ledger.go
litt3: @samlaf has expressed the opinion in the past that configs should *not* be pointers, so I was just following that.

I'm not super opinionated, and can't remember the precise reasoning. Sam, care to weigh in?

core/payments/reservation/reservation.go
litt3: 23c1d7a31d584af6f8824d6614b70204cb1546ec

core/payments/reservation/reservation.go
litt3: Done 23c1d7a31d584af6f8824d6614b70204cb1546ec

core/payments/reservation/reservation_ledger.go
litt3: Tried to clarify here df2cb309fdf1d118f3c6c661b77e0fc7ad2ce08c

core/payments/reservation/reservation_ledger.go
samlaf: I just favor immutability to prevent bugs from creeping in. If this was rust and we could pass an immutable reference that would be ideal, but passing a pointer in golang you never know whether the function will modify it or do whatever with it.

core/data.go
```diff
@@ -629,6 +629,10 @@ func ConvertToPaymentMetadata(ph *commonpbv2.PaymentHeader) (*PaymentMetadata, e
 }
 
 // ReservedPayment contains information the onchain state about a reserved payment
+//
+// TODO(litt3): this struct is in the process of being deprecated. It is used by the old accounting logic, but will
+// be replaced by the `reservation.Reservation` struct once the new accounting logic has superseded the old. At that
```
samlaf, line 634:
```suggestion
// be replaced by the [reservation.Reservation] struct once the new accounting logic has superseded the old. At that
```

samlaf: these links get picked up by IDE and you can click them.

core/payments/reservation/errors.go
```diff
@@ -1,25 +1,18 @@
...
-		e.PreviousTime.Format(time.RFC3339Nano),
-		e.PreviousTime.Sub(e.CurrentTime))
-}
+	// ErrTimeOutOfRange indicates the dispersal time is outside the reservation's valid time range.
+	ErrTimeOutOfRange = errors.New("time outside reservation range")
+
+	// ErrTimeMovedBackward indicates a timestamp was observed that is before a previously observed timestamp.
+	// This is unlikely to happen in practice, but could be the result of clock drift and NTP adjustments.
+	ErrTimeMovedBackward = errors.New("time moved backward")
+)
```
samlaf: curious why are you moving away from structured errors that have more information?

core/payments/reservation/leaky_bucket.go
```diff
@@ -154,13 +154,13 @@ func (lb *LeakyBucket) RevertFill(now time.Time, symbolCount uint32) error {
...
 //
-// Returns a TimeMovedBackwardError if input time is before previous leak time.
+// Returns ErrTimeMovedBackward if input time is before previous leak time.
 func (lb *LeakyBucket) leak(now time.Time) error {
 	elapsed := now.Sub(lb.previousLeakTime)
 
 	if elapsed < 0 {
 		// This can only happen if the user passes in time instances without monotonic timestamps
-		return &TimeMovedBackwardError{PreviousTime: lb.previousLeakTime, CurrentTime: now}
+		return fmt.Errorf("%w: previous=%v, current=%v", ErrTimeMovedBackward, lb.previousLeakTime, now)
```
samlaf: is your idea that clients will never need the timestamps so doesnt matter to include them as part of the error they will match on?

core/payments/reservation/reservation_ledger.go
litt3: Let's have the conversation here https://github.com/Layr-Labs/eigenda/pull/1873#discussion_r2277526418

core/payments/reservation/reservation_ledger_config.go
litt3: Let's have the conversation here https://github.com/Layr-Labs/eigenda/pull/1873#discussion_r2277526418

core/payments/reservation/reservation.go
```diff
@@ -0,0 +1,96 @@
...
+	for _, quorum := range quorums {
+		if _, ok := r.permittedQuorumIDs[quorum]; ok {
+			continue
+		}
+
+		permittedQuorums := make([]core.QuorumID, 0, len(r.permittedQuorumIDs))
+		for quorumID := range r.permittedQuorumIDs {
+			permittedQuorums = append(permittedQuorums, quorumID)
+		}
+		return fmt.Errorf("%w: quorum %d not in permitted set %v", ErrQuorumNotPermitted, quorum, permittedQuorums)
```
samlaf: the problem with wrapping like this is that when clients pattern match using errors.As() they'll get the ErrQuorumNotPermitted unwrapped and lost the quorum information

core/payments/reservation/reservation_ledger.go
```diff
@@ -0,0 +1,122 @@
...
+	return nil
+}
+
+// Checks whether a dispersal with given properties is permitted under the parameters of the reservation.
+//
+// This check should be called prior to calling Debit.
+//
+// Returns ErrQuorumNotPermitted if requested quorums are not permitted by the reservation
+// Returns ErrTimeOutOfRange if dispersal time is outside the reservation's valid time range
+func (rl *ReservationLedger) CheckInvariants(
```
samlaf: curious why you're doing it this way? I personally find it cleaner to have a single API (disperse) which returns structured errors.
Client can then pattern match the errors and decide which ones are worth retrying and which ones he should crash on, or log and continue, etc.

I guess you're basically separating errors into this API which checks for 400s errors and then any errors returned by dispersal() is a 500 which is worth retrying (possibly with backoff)?

core/payments/reservation/reservation_ledger_config.go
```diff
@@ -0,0 +1,64 @@
...
+	// This configuration parameter exists just in case we want to limit the cases that overfill is permitted in the
+	// future, but the current intention is for all entities to run with OverfillBehavior == OverfillOncePermitted
+	overfillBehavior OverfillBehavior
+	// Determines the maximum burst capacity of the LeakyBucket.
+	//
+	// The actual bucket capacity in symbols = symbolsPerSecond * bucketCapacityDuration
+	//
+	// This duration will be different for different parties, even for a given reservation. Clients will be configured
+	// to have a smaller bucket size than dispersers and validators, to account for latency in the dispersal process.
+	bucketCapacityDuration time.Duration
```
samlaf: ive thought about this some more and I'm not convinced that having bucket capacity in units of time makes sense anymore :D

Because it controls burstiness, which doesn't make sense to think in terms of seconds. It's the opposite, its the amount of data that I want to allow dispersing in a very (epsilon) short amount of time.

samlaf (dismissed):
LGTM outside of some nits

core/payments/reservation/reservation_ledger.go
litt3: There will still be a single `disperse` API that clients interact with. This `ReservationLedger` will be wrapped by a `ClientLedger`, which knows how to handle reservations, on demand payments, and all the required fallback logic.

The reason I split this up is for clarity in the parameters: we are dealing with *two* timestamps here: one timestamp is used for the leaky bucket, and a separate timestamp (passed into `CheckInvariants`), is the timestamp from the `PaymentHeader`, which is validated against the reservation parameters. To complicate things further, these timestamps will be _the same_ on the client side, but _different_ for validators and dispersers. 

To try to avoid confusion, I think it makes most sense to split these. My confidence isn't super high, and this could end up changing... but I think it makes sense to start here and then we can reconsider once we see how everything starts fitting together

core/payments/reservation/reservation_ledger_config.go
litt3: How else would you propose to configure it?

Each reservation has a rate, and we need to decide what multiple of that rate to set the bucket capacity at.

Thinking about this from the perspective of units, if we have a rate with unit `symbols / second`, and we want to convert that into a `bucketSize` with unit `symbols`, we need to multiply by a unit of time

core/payments/reservation/errors.go
litt3: Just because this seemed simpler I guess.

So based on your comments, you'd be in favor of implementing all the errors in a structured way, like I had before? I'd be ok with that strategy

core/payments/reservation/leaky_bucket.go
litt3: Yeah, that's pretty much the thought process. Let's continue the discussion here https://github.com/Layr-Labs/eigenda/pull/1873#discussion_r2277644748

core/data.go
litt3: 461cbf2143586d1cf6535c4dace9dc1486ed1c92

core/payments/reservation/errors.go
litt3: aa47492f3a342090c9388c1f4eb9324e92d51f30

core/payments/reservation/reservation.go
litt3: aa47492f3a342090c9388c1f4eb9324e92d51f30

core/payments/reservation/errors.go
samlaf: I really don't like the verbosity of golang structured errors either.. :\
But if we don't do this, then pattern matching on them loses all the context, so even if the call site logs the error, they won't see the time information.

Oh wait I guess this is not true if they log before starting the pattern matching... but it feels to finicky to me?

EDIT: having second-second thoughts lol. If clients pattern match using `errors.Is` then it doesn't unwrap, so your original sentinel errors were prob fine. It's only if they pattern match with `errors.As`... maybe we can revisit this discussion when we figure out how we want to pattern match the errors to do retries at the abstraction above? Even with structured errors I typically end up implementing a sentinel error for each type and implementing the `Is()` function so that I can pattern match more easily (don't need `errors.As`), so maybe your way is easier after all...

core/payments/reservation/reservation_ledger_config.go
samlaf: I guess my main point is I'm not convinced the bucketSize should be a multiple of the rate...
But this is bike-shedding at this point don't want to waste too much time on this, how about we discuss shortly during standup instead?

core/payments/reservation/reservation_ledger.go
samlaf: I see... hadn't realized they are 2 separate times. 2 things:
1. maybe rename `dispersalTime` to `paymentHeaderTimestamp`?
2. actually now thinking even more that having a single API (only Debit()) that takes both timestamps would be less bug-prone? I think (?) would make it obvious that those are potentially separate timestamps (because caller will be surprised 2 timestamps are being passed)?

core/payments/reservation/errors.go
litt3: Let's merge as-is, and revisit once we settle for sure on the right way

core/payments/reservation/reservation_ledger.go
cody-littley: Not 100% convinced, but not something worth blocking the PR over.

Lack of a language level support for const/immutable fields is easily in the top 5 of my golang gripes.

core/payments/reservation/reservation_ledger_config.go
litt3: > I guess my main point is I'm not convinced the bucketSize should be a multiple of the rate

Personally I can't think of any other way that would make sense. Curious if you come up with a different proposal, though

core/payments/reservation/reservation_ledger.go
litt3: 1. The problem is that the client will be doing this _before_ the `PaymentHeader` even exists, so I'm not sure  `paymentHeaderTimestamp` makes sense.
2. Let's revisit once this is actually used. I was unsure about the best way here, but I think actual usage will reveal the right choice, without needing to agonize too much right now

PR #1874, opened 2025-08-14
feat: litt rebase

cody-littley:
Adds the `litt rebase` command. Useful for changing the drive layout of a LittDB deployment.

### DIFF
--- a/litt/cli/rebase.go
+++ b/litt/cli/rebase.go
@@ -38,7 +38,7 @@ func rebaseCommand(ctx *cli.Context) error {
 		}
 	}
 
-	destinations := ctx.StringSlice("dest")
+	destinations := ctx.StringSlice("dst")
 	if len(destinations) == 0 {
 		return fmt.Errorf("no destinations provided")
 	}
@@ -303,7 +303,7 @@ func transferDataInTable(
 	}
 
 	if !preserveOriginal {
-		err = deleteSnapshotDirectory(logger, source, tableName, verbose)
+		err = deleteSnapshotDirectory(source, tableName)
 		if err != nil {
 			return fmt.Errorf("failed to delete snapshot directory for table %s: %w", tableName, err)
 		}
@@ -361,7 +361,7 @@ func deleteBoundaryFiles(logger logging.Logger, source string, tableName string,
 }
 
 // delete the old snapshot directory for a table. This will be reconstructed the next time the DB is loaded.
-func deleteSnapshotDirectory(logger logging.Logger, source string, tableName string, verbose bool) error {
+func deleteSnapshotDirectory(source string, tableName string) error {
 	snapshotDir := filepath.Join(source, tableName, segment.HardLinkDirectory)
 
 	exists, err := util.Exists(snapshotDir)
@@ -370,10 +370,6 @@ func deleteSnapshotDirectory(logger logging.Logger, source string, tableName str
 	}
 	if !exists {
 		return nil
-	}
-
-	if verbose {
-		logger.Infof("Deleting snapshot directory: %s", snapshotDir)
 	}
 
 	err = os.RemoveAll(snapshotDir)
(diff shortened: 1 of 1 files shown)


litt/cli/rebase.go
```diff
@@ -0,0 +1,591 @@
...
+
+	sources := ctx.StringSlice("src")
+	if len(sources) == 0 {
+		return fmt.Errorf("no sources provided")
+	}
+	for i, src := range sources {
+		var err error
+		sources[i], err = util.SanitizePath(src)
+		if err != nil {
+			return fmt.Errorf("invalid source path: %s", src)
```
litt3: nit: missing actual error from error message

litt/cli/rebase.go
```diff
@@ -0,0 +1,591 @@
...
+
+	destinations := ctx.StringSlice("dest")
+	if len(destinations) == 0 {
+		return fmt.Errorf("no destinations provided")
+	}
+	for i, dest := range destinations {
+		var err error
+		destinations[i], err = util.SanitizePath(dest)
+		if err != nil {
+			return fmt.Errorf("invalid destination path: %s", dest)
```
litt3: also missing actual error here

litt/cli/rebase.go
```diff
@@ -0,0 +1,591 @@
...
+			}
+
+			// Ignore "table.metadata" files, as they are not segment files.
+			if d.Name() == disktable.TableMetadataFileName {
+				return nil
+			}
+
+			// Check if the file is a segment file.
+			extension := filepath.Ext(path)
+			if extension == segment.MetadataFileExtension ||
```
litt3, line 174: Why is `MetadataFileExtension` in this condition, if you are `Ignoring "table.metadata" files`, as claimed above?

litt/cli/rebase.go
```diff
@@ -0,0 +1,591 @@
...
+	exists, err := util.Exists(source)
+	if err != nil {
+		return fmt.Errorf("failed to check if source directory %s exists: %w", source, err)
+	}
+	if !exists {
+		return nil
+	}
+
+	// Acquire a lock on the source directory.
+	lockPath := path.Join(source, util.LockfileName)
```
litt3, line 219: Why do you acquire a lock for all destinations at the start of the process, but acquire the lock for each source as you go along?

litt/cli/rebase.go
```diff
@@ -0,0 +1,591 @@
...
+		fsync,
+		verbose,
+		totalSegmentFileCount,
+		segmentFileCount)
+	if err != nil {
+		return fmt.Errorf("failed to transfer segment data for table %s: %w", tableName, err)
+	}
+
+	if !preserveOriginal {
+		err = deleteSnapshotDirectory(logger, source, tableName, true)
```
litt3: why `true` for `verbose` param, instead of respecting the input `verbose` parameter to `transferDataInTable`?

litt/cli/rebase.go
cody-littley: fixed

litt/cli/rebase.go
cody-littley: fixed

litt/cli/rebase.go
cody-littley: There are two flavors of metadata file:

- table metadata, which contains table configuration, there is exactly one of these per table
- segment metadata, one per segment in (so many will be present in a table), contains things like sharding factor, size of the segment, etc

This function specifically wants to count segment files, so it ignores the table metadata file, but counts the segment metadata files.

litt/cli/rebase.go
cody-littley: Added this comment:

```
	// Don't immediately take a lock on the source directories. Each source directory will be locked individually
	// before its data is transferred. Because source directories are deleted after their data is transferred,
	// it is inconvenient to hold the locks in this outer scope (since we need to release the lock to 
	// delete the directory).
```

litt/cli/rebase.go
cody-littley: This is an oversight, now uses the argument instead of hard coding `true`

PR #1876, opened 2025-08-15
fix: initOnceGrpc function bugs

samlaf:
Closes EGDA-1742.

These functions had a bug where the error was not stored, and hence would not be re-returned on future calls.

Furthermore, realized the function itself is completely unnecessary because creating a new grpc client doesn't actually make any IO calls. Those are only done the first time an rpc is submitted. So this logic was not needed at all. Removed it.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/disperser_client.go
+++ b/api/clients/disperser_client.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
-	"sync"
 	"time"
 
 	"github.com/Layr-Labs/eigenda/api"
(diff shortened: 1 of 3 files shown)


bxue-l2 (approved):
lgtm, how do you discover those bugs

samlaf:
> Ok, I've updated #1815 (fixed merge conflicts, light cleanup). Should be ready to merge with an approval.

merged master into this PR and fixed conflicts with your changes in v2/disperser_client.go, can you take a look and re-approve if it looks good?

litt3 (approved):

PR #1878, opened 2025-08-15
refactor: pass blobkey by value

samlaf:

Was trying to cleanup some of our conversion utils... and completely gave up. Going to take more work than I thought.
But while looking around I found this inconsistency in our APIs: blobkeys were sometimes passed by value, sometimes by pointer.

Passing by value is most likely better, and cleaner. Here is Claude's take:

```
⏺ For a [32]byte, returning by value is almost always better than returning a pointer. Here's why:

  Performance Comparison

  Passing [32]byte by value:
  - Stack allocation: No heap allocation needed
  - CPU cache friendly: 32 bytes fit in a single cache line (64 bytes)
  - Copy cost: ~1-2 CPU cycles with modern processors (SIMD instructions)

  Returning *[32]byte:
  - Heap allocation: Requires garbage collector involvement
  - Pointer indirection: Extra memory access to get the actual data
  - GC pressure: Creates work for the garbage collector
```

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/cert_builder.go
+++ b/api/clients/v2/cert_builder.go
@@ -133,7 +133,6 @@ func (cb *CertBuilder) getNonSignerStakesAndSignature(
 	}
 	checkSigIndices, err := cb.opsrCaller.GetCheckSignaturesIndices(&bind.CallOpts{Context: ctx, BlockNumber: big.NewInt(int64(rbn))},
 		cb.registryCoordinatorAddr, uint32(rbn), quorumNumbers, nonSignerOperatorIDs)
-
 	if err != nil {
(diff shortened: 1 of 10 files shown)


iquidus (approved):

bxue-l2 (approved):
lgtm

iquidus:
lgtm

Abusing pointers is sadly common practice in Go (I'm also guilty of this). If a value is referenced after the function that created it has returned, the compiler has to allocate it to the heap (otherwise it would get overwritten in the stack). We should only be returning pointers if it actually makes sense to do so. Less heap, less GC.

PR #1879, opened 2025-08-15
feat(metrics): add v2 disperser client metrics

iquidus:
Closes DAINT-692

Adds metrics for the v2 disperser client

Adding more metrics here gives use more data and allows users to optimize their dispersal costs.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/disperser_client.go
+++ b/api/clients/v2/disperser_client.go
@@ -7,6 +7,7 @@ import (
 	"time"
 
 	"github.com/Layr-Labs/eigenda/api"
+	"github.com/Layr-Labs/eigenda/api/clients/v2/metrics"
 	disperser_rpc "github.com/Layr-Labs/eigenda/api/grpc/disperser/v2"
 	"github.com/Layr-Labs/eigenda/common"
 	"github.com/Layr-Labs/eigenda/core"
@@ -15,6 +16,7 @@ import (
 	"github.com/Layr-Labs/eigenda/encoding"
 	"github.com/Layr-Labs/eigenda/encoding/rs"
 	"github.com/docker/go-units"
+	"github.com/prometheus/client_golang/prometheus"
 	"google.golang.org/grpc"
 )
 
@@ -57,6 +59,7 @@ type disperserClient struct {
 	prover             encoding.Prover
 	accountant         *Accountant
 	accountantLock     sync.Mutex
+	metrics            metrics.DispersalMetricer
 }
 
 var _ DisperserClient = &disperserClient{}
@@ -81,7 +84,13 @@ var _ DisperserClient = &disperserClient{}
 //
 //	// Subsequent calls will use the existing connection
 //	status2, blobKey2, err := client.DisperseBlob(ctx, data, blobHeader)
-func NewDisperserClient(config *DisperserClientConfig, signer corev2.BlobRequestSigner, prover encoding.Prover, accountant *Accountant) (*disperserClient, error) {
+func NewDisperserClient(
+	config *DisperserClientConfig,
+	signer corev2.BlobRequestSigner,
+	prover encoding.Prover,
+	accountant *Accountant,
+	registry *prometheus.Registry,
+) (*disperserClient, error) {
 	if config == nil {
 		return nil, api.NewErrorInvalidArg("config must be provided")
 	}
@@ -95,11 +104,14 @@ func NewDisperserClient(config *DisperserClientConfig, signer corev2.BlobRequest
 		return nil, api.NewErrorInvalidArg("signer must be provided")
 	}
 
+	metrics := metrics.NewDispersalMetrics(registry)
+
 	return &disperserClient{
 		config:     config,
 		signer:     signer,
 		prover:     prover,
 		accountant: accountant,
+		metrics:    metrics,
 		// conn and client are initialized lazily
 	}, nil
 }
@@ -283,6 +295,9 @@ func (c *disperserClient) DisperseBlobWithProbe(
 	if verifyReceivedBlobKey(blobHeader, reply) != nil {
 		return nil, [32]byte{}, fmt.Errorf("verify received blob key: %w", err)
 	}
+
+	c.metrics.RecordBlobSize(uint(len(data)))
+	c.metrics.RecordSymbolLength(symbolLength)
 
 	return &blobStatus, corev2.BlobKey(reply.GetBlobKey()), nil
 }
--- a/api/clients/v2/metrics/dispersal.go
+++ b/api/clients/v2/metrics/dispersal.go
@@ -1,0 +1,75 @@
+package metrics
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
+const (
+	dispersalSubsystem = "dispersal"
+)
+
+type DispersalMetricer interface {
+	RecordBlobSize(size uint)
+	RecordSymbolLength(length uint)
+}
+
+type DispersalMetrics struct {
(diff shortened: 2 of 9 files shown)


api/clients/v2/metrics/dispersal.go
```diff
@@ -0,0 +1,75 @@
...
+	sizeBuckets := []float64{
+		1024,     // 1KB
+		4096,     // 4KB
+		16384,    // 16KB
+		65536,    // 64KB
+		262144,   // 256KB
+		1048576,  // 1MB
+		4194304,  // 4MB
+		16777216, // 16MB
+	}
```
samlaf: think we should have all blob sizes, not skipping any category. 
[0, 128KiB, 256KiB, 512KiB, 1MiB, 2MiB,4MiB, 8MiB, 16MiB]
There's not that many anyways. Anything below 128KiB is charged for 128KiB anyways, so prob don't need fine grained buckets there (a single one would prob be fine imo).

api/clients/v2/metrics/dispersal.go
```diff
@@ -0,0 +1,75 @@
...
+			Help:      "Size of blobs created from payloads in bytes",
+			Buckets:   sizeBuckets,
+		}, []string{}),
+		SymbolLength: promauto.With(registry).NewHistogramVec(prometheus.HistogramOpts{
+			Name:      "blob_size_symbols",
+			Namespace: namespace,
+			Subsystem: dispersalSubsystem,
+			Help:      "Size of blobs created from payloads in symbols",
+			Buckets:   sizeBuckets,
+		}, []string{}),
```
samlaf: hmm this is the same metrics, except size_bytes is size_symbols*32. Don't think we need these as separate metris, and certainly don't think we need 2 separate functions to set them. I think the unit conversion could be done in grafana if one really wants symbols instead of bytes. And if we decide we really want to output both, I think they should be the same metric with a label, and both should be set from a single function RecordBlobSize()

api/clients/v2/metrics/dispersal.go
```diff
@@ -0,0 +1,75 @@
...
+	}
+}
+
+func (m *DispersalMetrics) RecordBlobSize(size uint) {
```
samlaf: document that size is in bytes. Prob change function to `RecordBlobSizeBytes`?

api/clients/v2/disperser_client.go
```diff
@@ -81,7 +84,13 @@ var _ DisperserClient = &disperserClient{}
 //
 //	// Subsequent calls will use the existing connection
 //	status2, blobKey2, err := client.DisperseBlob(ctx, data, blobHeader)
-func NewDisperserClient(config *DisperserClientConfig, signer corev2.BlobRequestSigner, prover encoding.Prover, accountant *Accountant) (*disperserClient, error) {
+func NewDisperserClient(
+	config *DisperserClientConfig,
+	signer corev2.BlobRequestSigner,
+	prover encoding.Prover,
+	accountant *Accountant,
+	registry *prometheus.Registry,
```
samlaf: I'm not a big fan of a nil registry pointer implicitly creating a noopMetricer. This should at the very least be documented, but otherwise any reason we don't take in a metricer interface instead, and let the client explicitly decide whether they want a real metrics instance or a noop one?

iquidus: Felt like this approach is cleaner (no need to create metrics before calling NewDisperserClient) and less error prone (no chance of nil metrics). But completely see where you are coming from. Have refactored, replacing the registry with the metrics interface and adding a nil check in [c9c3a42](https://github.com/Layr-Labs/eigenda/pull/1879/commits/c9c3a42a20a881a4b689b306ac63052128de5f82)

api/clients/v2/metrics/dispersal.go
iquidus: Done in: [3aebf9c](https://github.com/Layr-Labs/eigenda/pull/1879/commits/3aebf9cdbdf2f44e8f12835e3d6f174f1c891805)

api/clients/v2/metrics/dispersal.go
iquidus: Updated in [05c32aa](https://github.com/Layr-Labs/eigenda/pull/1879/commits/05c32aa57e8a0b724bb698dd593ec6b4bab998b0)

api/clients/v2/metrics/dispersal.go
iquidus: Removed symbol length as part of [3aebf9c](https://github.com/Layr-Labs/eigenda/pull/1879/commits/3aebf9cdbdf2f44e8f12835e3d6f174f1c891805)

api/clients/v2/disperser_client.go
samlaf: I see yeah. I still prefer readability at the callsites, even if its a bit more typing. imo code should be written to be easily read.
Also regarding nil checks in constructors, it's def fine to have, but I'm currently on the camp of thinking its not really needed. If a nil is passed, it'll panic with a stack trace very soon. Ideally in a test, and if not, then a test should be added anyways.

api/clients/v2/disperser_client.go
```diff
@@ -94,12 +102,17 @@ func NewDisperserClient(config *DisperserClientConfig, signer corev2.BlobRequest
 	if signer == nil {
 		return nil, api.NewErrorInvalidArg("signer must be provided")
 	}
+	if metrics == nil {
+		// nolint:wrapcheck
+		return nil, api.NewErrorInvalidArg("metrics must be provided")
+	}
```
samlaf: just realized all of these errors are wrong. These are not 400 errors, they are internal bugs. api.NewErrorInalidArg afaiu is meant to return 400s to users. Creating a new disperser client doesn't happen in a user handling goroutine. We should change these errors. Maybe in another PR?

api/clients/v2/metrics/dispersal.go
```diff
@@ -0,0 +1,60 @@
...
+const (
+	dispersalSubsystem = "dispersal"
+)
+
+type DispersalMetricer interface {
+	RecordBlobSizeBytes(size uint)
+}
+
+type DispersalMetrics struct {
+	BlobSize *prometheus.HistogramVec
```
samlaf: why are you using a Vec here btw? We only need a single histogram right?

api/clients/v2/metrics/dispersal.go
```diff
@@ -0,0 +1,60 @@
...
+			Namespace: namespace,
+			Subsystem: dispersalSubsystem,
+			Help:      "Size of blobs created from payloads in bytes",
+			Buckets:   sizeBuckets,
+		}, []string{}),
+	}
+}
+
+func (m *DispersalMetrics) RecordBlobSizeBytes(size uint) {
+	m.BlobSize.WithLabelValues().Observe(float64(size))
```
samlaf: You're not actually passing any label Values, so don't think we need to use a HistogramVec and can use a Histogram instead?

iquidus: Changed to Histrogram in [289a2dd](https://github.com/Layr-Labs/eigenda/pull/1879/commits/289a2dd95fac020119cc68dba9e4aa556269c5e7)

api/clients/v2/metrics/dispersal.go
iquidus: Good catch. Changed to Histogram in [289a2dd](https://github.com/Layr-Labs/eigenda/pull/1879/commits/289a2dd95fac020119cc68dba9e4aa556269c5e7)

samlaf (approved):

samlaf (approved):

PR #1880, opened 2025-08-15
refactor: refactor accountant metrics passing

iquidus:
Refactors the accountant constructors to accept a
`*prometheus.Registry` argument instead of a `Accountant.Metricer`

Moves `NewAccountantMetrics` call inside `NewAccountant`

Updates example

Passing the registry brings this inline with other metrics patterns.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/accountant.go
+++ b/api/clients/v2/accountant.go
@@ -12,6 +12,7 @@ import (
 	"github.com/Layr-Labs/eigenda/core"
 	"github.com/Layr-Labs/eigenda/core/meterer"
 	gethcommon "github.com/ethereum/go-ethereum/common"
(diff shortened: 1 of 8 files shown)


api/clients/v2/accountant.go
```diff
@@ -59,12 +60,14 @@ func NewAccountant(
 	pricePerSymbol uint64,
 	minNumSymbols uint64,
 	numBins uint32,
-	metrics metrics.AccountantMetricer,
+	registry *prometheus.Registry,
```
samlaf, line 63: why this change? See my comment here: https://github.com/Layr-Labs/eigenda/pull/1879#discussion_r2281118474

PR #1881, opened 2025-08-15
feat(payments): Implement `OnDemandLedger`

litt3:
part of EGDA-1853

It doesn't close that issue, since this PR doesn't include the DynamoDB implementation of the `CumulativePaymentStore`


core/payments/ondemand/cumulative_payment_store.go
```diff
@@ -0,0 +1,17 @@
+package ondemand
+
+import (
+	"context"
+	"math/big"
+)
+
+// Knows how to store and retrieve a cumulative payment value, stored as a *big.Int representing a number of wei
```
samlaf: Maybe use `set` and `get` instead of store and retrieve? idk my brain associates persistence with `store`, whereas your only implementation is inmemory.

Also "stored as a *big.Int is not necessarily true right? It most liely will be serialized before getting persisted to a fs or db right?

core/payments/ondemand/ephemeral_cumulative_payment_store.go
samlaf, line 1: I would prefer if each implementation of `CumulativePaymentStore` was in its own package. Otherwise the package gets cluttered with multiple implementations of the same interface, which has risks of bugs (using some var/function/etc that should only be used for a certain implementation with the other implementation)

core/payments/ondemand/errors.go
```diff
@@ -0,0 +1,34 @@
...
+	"fmt"
+	"math/big"
+
+	"github.com/Layr-Labs/eigenda/core"
+)
+
+// Indicates that a requested quorum is not supported for on-demand payments.
+type QuorumNotSupportedError struct {
+	RequestedQuorum  core.QuorumID
+	SupportedQuorums []core.QuorumID
```
samlaf, line 13: I thought you said we would only ever support quorums 0 and 1, and you were going to hardcode those values in the codebase?

core/payments/ondemand/on_demand_ledger.go
```diff
@@ -0,0 +1,207 @@
...
+	"github.com/Layr-Labs/eigenda/core"
+)
+
+// Keeps track of the cumulative payment state for on-demand dispersals for a single account.
+//
+// On-demand payments use a cumulative payment system where, each time a dispersal is made, we keep track of the total
+// amount paid by the account for that and all previous dispersals. The cumulative payment is chosen by the client
+// based on the state of its local accounting, and the chosen value can be verified by all other parties by checking:
+// 1. that the claimed value is <= the total deposits belonging to the account in the PaymentVault contract
+// 2. that the value has increased by at least the cost of the dispersal from the previously observed value
```
samlaf, line 18: saying that anyone can observe this is a bit of a stretch. Our disperser does not make this value public afaiu. Also this doesn't work in a world where there are multiple dispersers.

core/payments/ondemand/on_demand_ledger.go
```diff
@@ -0,0 +1,207 @@
...
+//
+// On-demand payments are currently limited to quorums 0 (ETH) and 1 (EIGEN) and can only be used when dispersing
+// through the EigenDA disperser.
+//
+// This is a goroutine safe struct, with some caveats. While the methods themselves can be called concurrently, there
+// are some practical concurrency limits based on the logic behind on-demand payments:
+// 1. Dispersals must be accounted for *in order*, since cumulative payment may only increase. Example:
+// - Assume there are two dispersals, N and N+1
+// - N+1 necessarily has a greater cumulative payment
+// - If N+1 is accounted for first, then the subsequent accounting for dispersal N will fail, since cumulative payment
```
samlaf: can this actually happen in practice? In practice, if N and N+1 are submitted in reverse order... then N+1 becomes N and N becomes N+1.

core/payments/ondemand/on_demand_ledger.go
```diff
@@ -0,0 +1,207 @@
...
+// 2. Debit reversion is only possible for a dispersal if no subsequent dispersals have been accounted for. Example:
+// - Assume there are two dispersals, N and N+1
+// - If N is accounted for, the debit for N can be reverted *only if* N+1 has not yet been accounted for
+// - If this is not respected, the entity doing the accounting will diverge from the correct ledger state
+//
+// TODO(litt3): There are some improvements that may be made to the on-demand logic in the future, to permit a larger
+// number of concurrent on-demand dispersals.
+type OnDemandLedger struct {
+	// total deposits available for the account in wei
+	totalDeposits *big.Int
```
samlaf, line 42: why not call this cumulativeDeposits, to match cumulativePayments?

core/payments/ondemand/on_demand_ledger.go
```diff
@@ -0,0 +1,207 @@
...
+	// price per symbol in wei
+	pricePerSymbol *big.Int
+	// minimum number of symbols to bill
+	minNumSymbols *big.Int
+	// synchronizes access to the cumulative payment store
+	lock sync.Mutex
+	// stores the cumulative payment for this ledger in wei
+	cumulativePaymentStore CumulativePaymentStore
```
samlaf: Don't think it makes sense for the OnDemandLedger to control the concurrency access to the cumulativePaymentStore. Different stores most likely have very differing concurrency access/control and would know better themselves how to manage that concurrency.

For eg your in-memory implementation would use a lock, but a dynamodb might have more sophisticated ways of managing concurrency. Some dbs might want to do optimistic concurrency control, others might use lockless approaches, etc.

core/payments/ondemand/on_demand_ledger.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	if pricePerSymbol.Sign() < 0 {
+		return nil, errors.New("pricePerSymbol cannot be negative")
+	}
+
+	if minNumSymbols == nil {
+		return nil, errors.New("minNumSymbols cannot be nil")
+	}
+	if minNumSymbols.Sign() < 0 {
+		return nil, errors.New("minNumSymbols cannot be negative")
```
samlaf: does this really need to be a *big.Int? or can it be a uint64?

core/payments/ondemand/on_demand_ledger.go
```diff
@@ -0,0 +1,207 @@
...
+
+	odl.lock.Lock()
+	defer odl.lock.Unlock()
+
+	currentCumulativePayment, err := odl.cumulativePaymentStore.GetCumulativePayment(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("get cumulative payment: %w", err)
+	}
+
+	newCumulativePayment := new(big.Int).Add(currentCumulativePayment, blobCost)
```
samlaf:
```suggestion
	newCumulativePayment := currentCumulativePayment.Add(currentCumulativePayment, blobCost)
```

samlaf: Any reason why you can't reuse the same memory?

core/payments/ondemand/on_demand_ledger.go
```diff
@@ -0,0 +1,207 @@
...
+// Computes the on demand cost of a number of symbols
+func (odl *OnDemandLedger) computeCost(symbolCount uint32) *big.Int {
+	symbolCountBig := big.NewInt(int64(symbolCount))
+
+	billableSymbols := symbolCountBig
+	if symbolCountBig.Cmp(odl.minNumSymbols) < 0 {
+		billableSymbols = odl.minNumSymbols
+	}
+
+	return new(big.Int).Mul(billableSymbols, odl.pricePerSymbol)
```
samlaf:
```suggestion
	return billableSymbols.Mul(billableSymbols, odl.pricePerSymbol)
```

samlaf: In many places you are needlessly allocating new memory.

core/payments/ondemand/on_demand_ledger.go
```diff
@@ -0,0 +1,207 @@
...
+
+	odl.lock.Lock()
+	defer odl.lock.Unlock()
+
+	currentCumulativePayment, err := odl.cumulativePaymentStore.GetCumulativePayment(ctx)
+	if err != nil {
+		return fmt.Errorf("get cumulative payment: %w", err)
+	}
+
+	newCumulativePayment := new(big.Int).Sub(currentCumulativePayment, blobCost)
```
samlaf:
```suggestion
	newCumulativePayment := currentCumulativePayment.Sub(currentCumulativePayment, blobCost)
```

core/payments/ondemand/on_demand_ledger.go
```diff
@@ -0,0 +1,207 @@
...
+
+	if newCumulativePayment.Sign() < 0 {
+		return fmt.Errorf("cannot revert debit: would result in negative cumulative payment")
```
samlaf: this looks like panic-worthy to me? This would only happen in case of a bug right? Since a revertDebit should only ever be called on an amount that has already been debitted?

core/payments/ondemand/on_demand_ledger.go
```diff
@@ -0,0 +1,207 @@
...
+		return nil, &InsufficientFundsError{
+			CurrentCumulativePayment: currentCumulativePayment,
+			TotalDeposits:            odl.totalDeposits,
+			BlobCost:                 blobCost,
+		}
+	}
+
+	if err := odl.cumulativePaymentStore.SetCumulativePayment(ctx, newCumulativePayment); err != nil {
+		return nil, fmt.Errorf("set cumulative payment: %w", err)
+	}
```
samlaf: This is getting into some fun territory! Can't help myself see this as a RMW (https://en.wikipedia.org/wiki/Read%E2%80%93modify%E2%80%93write) operation that we could optimize the crap out of. For example if we used https://github.com/lukechampine/uint128 instead of big.Ints, we could probably write a lockless implementation that uses hardware primitives like CAS to go brrr

samlaf: This btw, although of course very premature optimization, is another reason why I think the on_demand_ledger shouldnt control the concurrency of the underlying store. We could implement an ephemeral_cumulative_payment_store that uses a lock, and another one that is lockless with some atomic/CAS operations (this might require changing the interface though... since GET/SET don't permit this out of the box.. we might want an interface that has increment() for example, which seems to be the actual primitive we need?)

core/payments/ondemand/ephemeral_cumulative_payment_store.go
```diff
@@ -0,0 +1,43 @@
...
+// Constructs a new in-memory cumulative payment store
+func NewEphemeralCumulativePaymentStore() *EphemeralCumulativePaymentStore {
+	return &EphemeralCumulativePaymentStore{
+		cumulativePayment: big.NewInt(0),
+	}
+}
+
+// Gets the stored cumulative payment in wei
+func (e *EphemeralCumulativePaymentStore) GetCumulativePayment(_ context.Context) (*big.Int, error) {
+	return new(big.Int).Set(e.cumulativePayment), nil
```
samlaf: functions on big.Ints typically take the buffer as an argument, in order to allow the call-site to decide on memory allocation. For example if call site already has a big.Int buffer it could reuse, then we wouldn't need to allocate a new buffer inside the function (which btw you are not even using in your implementation and just recreating yet another buffer right away for the increment operation).

core/payments/ondemand/cumulative_payment_store.go
litt3: > Maybe use set and get instead of store and retrieve?

No strong opinion, I'll make this change

> Also "stored as a *big.Int is not necessarily true right?

It's true from the perspective of the interface user. The interface deals with big ints, and whatever it does under the hood is its own business

core/payments/ondemand/errors.go
litt3: > we would only ever support quorums 0 and 1

We only _currently_ support quorums 0 and 1. These values are hardcoded in `OnDemandLedger`, and the fewer places we have duplicated hard coded values, the easier it would be to change in the future, if needed.

core/payments/ondemand/on_demand_ledger.go
litt3: "All other parties" to me implies other participants in the protocol, not "anyone".

Looking at the specific points listed, # 1 truly is observable by anyone. And # 2 has the qualifier `from the previously observed value`, which seems correct to me. Do you have a suggestion for how to make this clearer?

> this doesn't work in a world where there are multiple dispersers.

Correct, this model doesn't work at all for multiple dispersers. On demand payments is limited to a single disperser

core/payments/ondemand/on_demand_ledger.go
litt3: It could happen in practice if we weren't aware of this property, and implemented the clients in a way that allowed it to happen, i.e. concurrent execution without synchronization guarantees.

I wrote this here to record the justification for why the clients must be implemented in a way that prevents this from happening. 

> if N and N+1 are submitted in reverse order... then N+1 becomes N and N becomes N+1.

The `N` here is about accounting order, not submission order. Curious if you have any suggestions to help address the confusion

core/payments/ondemand/on_demand_ledger.go
litt3: The name was chosen to match the `PaymentVault` naming

core/payments/ondemand/on_demand_ledger.go
litt3: The reason I was doing concurrency at the level of the `OnDemandLedger` is to to synchronize the actual accounting logic (get value, add, check, set new value)

If we don't have this synchronization logic in `OnDemandLedger`, we'd need to implement the whole accounting logic multiple times. But I guess this wouldn't be the worst thing, since it's pretty simple logic...

Will do the change as suggested 👍

core/payments/ondemand/cumulative_payment_store.go
litt3: 7579da28fd0ee48889f1370c58c22f8e76e5dfbd

core/payments/ondemand/ephemeral_cumulative_payment_store.go
litt3: 926d878e0c8ec123c6bf6ee9906ab4b357101d2f

core/payments/ondemand/errors.go
cody-littley: I think it's a good idea to move towards supporting on-demand payments for other quorums. I think it's mostly a technical limitation now, rather than a business decision, not to support on demand payments for custom quorums.

core/payments/ondemand/on_demand_ledger.go
```diff
@@ -0,0 +1,207 @@
...
+	lock sync.Mutex
+	// stores the cumulative payment for this ledger in wei
+	cumulativePaymentStore CumulativePaymentStore
+}
+
+// Constructs a new OnDemandLedger, backed by the input CumulativePaymentStore
+func NewOnDemandLedger(
+	totalDeposits *big.Int,
+	pricePerSymbol *big.Int,
+	minNumSymbols *big.Int,
```
cody-littley: I think Hedera made the right call when they decided to make the sum of all currency fit into 8 bytes. Not that we can do anything about that here, though.

core/payments/ondemand/on_demand_ledger.go
cody-littley: Not bad areas to instigate in the future. But I strongly suspect that this code will be orders of magnitude faster than it needs to be in order to handle extreme workload, so I agree that this is a premature optimization.

core/payments/ondemand/on_demand_ledger.go
litt3: 49821e571b09029fba568a1d13fd1fef2f70da3b

core/payments/ondemand/on_demand_ledger.go
litt3: 1c0d9374aa1c421d8a9bc4075022a79d6844c7aa

core/payments/ondemand/on_demand_ledger.go
litt3: There was no reason, but this doesn't exist anymore, due to other requested changes

core/payments/ondemand/on_demand_ledger.go
litt3: I think all should be fixed, in the process of making other requested changes

core/payments/ondemand/on_demand_ledger.go
litt3: Doesn't exist anymore, with other changes

core/payments/ondemand/on_demand_ledger.go
litt3: This is now the responsibility of the `CumulativePaymentStore`, instead of the `OnDemandLedger`.

In the `EphemeralCumulativePaymentStore` impl, I made this and other "impossible" bugs a panic

core/payments/ondemand/ephemeral_cumulative_payment_store.go
litt3: > functions on big.Ints typically take the buffer as an argument

I get that this can be a "more optimal" way to operate on big ints, but I lean toward thinking it's not worth the extra-complexity damage the the API in this case.

> For example if call site already has a big.Int buffer it could reuse

I'd argue that this sort of optimization would only be acceptable for cases that it's provably needed, since reusing an already-allocated big int for a new purpose is messy. Given the function of this API, I think I can confidently say that sort of optimization will never be necessary.

core/payments/ondemand/on_demand_ledger.go
cody-littley: I'm not a fan of using `panic()` for input validation like this. In order to determine if this is a safe thing to do, I first need to understand the entire workflow that led to this function being called. For example, does this function receive raw user data, or is it vetted first? By contrast, using standard errors allows me to reason about the safety of the function by looking at the function in isolation.

A related concern is that if we convert all of our input validation to use `panic()`, then inevitably a user will be able to force bad data into a function that causes a panic.

cody-littley (dismissed):

core/payments/ondemand/cumulative_payment_store.go
samlaf: > It's true from the perspective of the interface user. The interface deals with big ints, and whatever it does under the hood is its own business

bike-shed: I would say something like "stored a *big.Int" then, instead of "stored AS A *big.Int". But unimportant so resolving.

core/payments/ondemand/errors.go
samlaf: The real world constraint that @mooselumph was mentioning is that custom quorums will be created by rollups and represent their own token/quorum. Meaning every custom quorum has a single user. So its a lot easier for it to be governed by a reservation than to have to pay it in ETH on a per-dispersal basis.

core/payments/ondemand/on_demand_ledger.go
samlaf: > "All other parties" to me implies other participants in the protocol, not "anyone".

I see, but that seems like a pretty big restriction for a protocol that claims to be blockchain related. Probably worth mentioning explicitly.

> And # 2 has the qualifier from the previously observed value, which seems correct to me.

This is the one I had in mind. Only the person that dispersed and the disperser can observe this to be true, since other parties don't have access to the disperser's state. Directly related to my first comment above though.

core/payments/ondemand/on_demand_ledger.go
samlaf: Ahh I see so basically disperser-receiver-order must respect accounting-order. This means we really need to be careful with concurrent dispersals, especially if we start using a thread pool (if not its less scary because tcp respects send order).

> The N here is about accounting order, not submission order. Curious if you have any suggestions to help address the confusion

You say "there are two dispersals, N and N+1", which I think is what confused me. I would probably simplify the example to say "blob N is accounted first, and then blob N+1, but their dispersal (on the wire) order is reversed somehow"

core/payments/ondemand/cumulative_payment_store.go
```diff
@@ -0,0 +1,27 @@
...
+// Stores a cumulative payment, and can atomically add values to it while remaining within specified bounds.
+//
+// The cumulative payment is a *big.Int representing a number of wei
+type CumulativePaymentStore interface {
+	// Atomically adds a number of wei to the cumulative payment.
+	//
+	// May optionally support subtraction via negative amount parameter, depending on the implementation.
+	//
+	// Returns the new cumulative payment value after the addition.
+	// Returns An an [ondemand.InsufficientFundsError] if the addition would cause the cumulative payment to exceed
```
samlaf:
```suggestion
	// Returns an [ondemand.InsufficientFundsError] if the addition would cause the cumulative payment to exceed
```

core/payments/ondemand/ephemeral/ephemeral_cumulative_payment_store.go
```diff
@@ -0,0 +1,69 @@
...
+) (*big.Int, error) {
+	if amount == nil {
+		panic("amount cannot be nil")
+	}
+	if maxCumulativePayment == nil {
+		panic("maxCumulativePayment cannot be nil")
+	}
+	if maxCumulativePayment.Sign() < 0 {
+		panic(fmt.Sprintf("maxCumulativePayment cannot be negative: received %s", maxCumulativePayment.String()))
+	}
```
samlaf: afaiu @cody-littley is against panic'ing for input validation, and I agree with him.

Think we should only panic for invariants being broken. Here because its a public struct, anyone could import it and pass an invalid input. So I think public methods should return errors on invalid inputs, and call sites should panic if they receive such an error despite having already validated the inputs.

@cody-littley do you agree with this perspective?

core/payments/ondemand/errors.go
litt3: That makes sense, but I still think it makes sense to hardcode in as few places as possible

core/payments/ondemand/on_demand_ledger.go
litt3: 32aa8fbd3ba5661e6ffa004fe8200e8c4f85337a

I removed the the `all other parties` statement, to avoid the confusion: I don't think this is the right place to get into a deep `who knows what` explanation. That's a high level architecture kind of topic, which would become overwhelming to explain at all of the implementation leaves

litt3: 32aa8fbd3ba5661e6ffa004fe8200e8c4f85337a

I removed the the `all other parties` statement, to avoid the confusion: I don't think this is the right place to get into a deep `who knows what` explanation. That's a high level architecture kind of topic, which would become overwhelming to explain at all of the implementation leaves

core/payments/ondemand/on_demand_ledger.go
litt3: > This means we really need to be careful with concurrent dispersals

Right.

Ok, I see how it was confusing now. I tried rewording with `A` and `B`, which I think should be unambiguous now. f55b64cd746db95520a0c2b3278cbedd4cc8d7dd

core/payments/ondemand/cumulative_payment_store.go
litt3: d4d2ab170fce19d48044c84f967cf943c4c25775

core/payments/ondemand/ephemeral/ephemeral_cumulative_payment_store.go
litt3: dd3b81b048d9cb70052e86f7abe0995b54cfb1c7

panics removed

core/payments/ondemand/ephemeral_cumulative_payment_store.go
samlaf: I'm fine with merging as is. That being said, here are my bike-shed answers:

> I get that this can be a "more optimal" way to operate on big ints, but I lean toward thinking it's not worth the extra-complexity damage the the API in this case.

It's not that much more complex when you're used to it (for example if you have C background, or have used the golang big.Int API for a while). Literally every function there works that way. `add` is a method which takes 2 arguments, which means with the receiver it has 3 arguments! This broke my brain at first until I "got it".

core/payments/ondemand/on_demand_ledger.go
samlaf: This is the part I'm still confused by:
image

I thought the cumulative_payment WAS chosen by the accounting logic? But are you saying the current cumulative_payment is tracked outside of the accounting logic? So its set somewhere else?

Not worth blocking this PR since I might be the one misunderstanding, but for my understanding, can you list all the calls that need to be made in the same order? Is my understanding correct?
1. cumulative_payment first chosen somewhere, this decides on the initial ordering
2. accounting (debiting) must be called in the same order on the client
3. accounting (debiting) must be called in the same order on the disperser

core/payments/ondemand/on_demand_ledger.go
```diff
@@ -0,0 +1,167 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	blobCost := odl.computeCost(symbolCount)
+
+	newCumulativePayment, err := odl.cumulativePaymentStore.AddCumulativePayment(ctx, blobCost, odl.totalDeposits)
+	if err != nil {
+		return nil, fmt.Errorf("add cumulative payment: %w", err)
+	}
```
samlaf, line 115: should we panic here if a negativeBalanceError is returned? https://github.com/Layr-Labs/eigenda/pull/1881/files#diff-4e474963d9f8251df963bca81d3bc27580eec9e636c136552cb926afccc26bcbR54

After discussing with Cody, I agree that we shouldn't panic inside public methods because we don't know who will be using them. But on the outside, where its called from, if that error is returned, there's literally a bug right? We're trying to debit a negative value which is a bug? And at this point the software is literally broken and god knows that the accounting is doing? So we should panic iiuc?

samlaf (approved):
LGTM. Still have a question about a panic site but I don't think it should block merging this if you need to keep building on top.

core/payments/ondemand/on_demand_ledger.go
samlaf: Btw this is a follow-up to @cody-littley's answer here: https://github.com/Layr-Labs/eigenda/pull/1881#discussion_r2282803025

core/payments/ondemand/ephemeral_cumulative_payment_store.go
litt3: Yeah, that makes sense. Will keep and mind and follow the proposed pattern moving forward

core/payments/ondemand/on_demand_ledger.go
litt3: I think it would be _reasonable_ to panic in that case, but I'm not sure that it matters too much either way

core/payments/ondemand/on_demand_ledger.go
litt3: The answer here depends on the ongoing discussion surrounding `cumulativePayments`, which we're having in slack. Let's defer this discussion until our synchronous meeting tomorrow

PR #1882, opened 2025-08-17
feat: add grpc_recovery middleware on every server

samlaf:
This middleware will recover from any panics in any goroutine serving client requests, and instead of crashing the entire server, return a grpc INTERNAL error to that user.

This behavior is done by default on http servers, but for some reason is not default on grpc servers.

We have already seen our encoders crashing for a single request that was missing some srs table. This would have prevented those requests from crashing the entire encoder server.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


samlaf:
> I'm not convinced that this is the direction we want to go. In my mind, a `panic()` should always result in the teardown of the entire go process. This seems to subvert that pattern.

I agree that we should think through more whether any shared memory could be corrupted by requests. My impetus for doing this is that right now if a single request panics, the entire server crashes, which affects a lot of customers. Unclear whether this is worse than a shared memory corruption (since most of our services are mostly stateless afaiu).

golang's stdlib's http library has a recovery middleware behavior by default: https://pkg.go.dev/net/http#Handler
> If ServeHTTP panics, the server (the caller of ServeHTTP) assumes that the effect of the panic was isolated to the active request. It recovers the panic, logs a stack trace to the server error log, and either closes the network connection or sends an HTTP/2 RST_STREAM, depending on the HTTP protocol. To abort a handler so the client sees an interrupted response but the server doesn't log an error, panic with the value [ErrAbortHandler](https://pkg.go.dev/net/http#ErrAbortHandler).

samlaf:
Closing after discussions. Seems like there's no appetite for the risk introduced at this point in time.

PR #1884, opened 2025-08-17
feat(cli): add `doc metrics` cli command to proxy

iquidus:
Adds the `doc metrics` cli command back to the proxy without the `op-service` dependency.

Refactors existing proxy and v2 client metrics to use the added promauto factory wrapper for metric Document() support.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/metrics/metrics.go
+++ b/api/clients/v2/metrics/metrics.go
@@ -1,5 +1,138 @@
 package metrics
 
+import (
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promauto"
+)
+
 const (
 	namespace = "eigenda"
 )
+
+type Factory interface {
+	NewCounter(opts prometheus.CounterOpts) prometheus.Counter
+	NewCounterVec(opts prometheus.CounterOpts, labelNames []string) *prometheus.CounterVec
+	NewGauge(opts prometheus.GaugeOpts) prometheus.Gauge
+	NewGaugeFunc(opts prometheus.GaugeOpts, function func() float64) prometheus.GaugeFunc
+	NewGaugeVec(opts prometheus.GaugeOpts, labelNames []string) *prometheus.GaugeVec
+	NewHistogram(opts prometheus.HistogramOpts) prometheus.Histogram
+	NewHistogramVec(opts prometheus.HistogramOpts, labelNames []string) *prometheus.HistogramVec
+	NewSummary(opts prometheus.SummaryOpts) prometheus.Summary
+	NewSummaryVec(opts prometheus.SummaryOpts, labelNames []string) *prometheus.SummaryVec
+	Document() []DocumentedMetric
+}
+
+type DocumentedMetric struct {
+	Type   string   `json:"type"`
+	Name   string   `json:"name"`
+	Help   string   `json:"help"`
+	Labels []string `json:"labels"`
+}
+
+type documentor struct {
+	metrics []DocumentedMetric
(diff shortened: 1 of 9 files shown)


api/clients/v2/metrics/metrics.go
```diff
@@ -1,5 +1,138 @@
...
+	NewCounterVec(opts prometheus.CounterOpts, labelNames []string) *prometheus.CounterVec
+	NewGauge(opts prometheus.GaugeOpts) prometheus.Gauge
+	NewGaugeFunc(opts prometheus.GaugeOpts, function func() float64) prometheus.GaugeFunc
+	NewGaugeVec(opts prometheus.GaugeOpts, labelNames []string) *prometheus.GaugeVec
+	NewHistogram(opts prometheus.HistogramOpts) prometheus.Histogram
+	NewHistogramVec(opts prometheus.HistogramOpts, labelNames []string) *prometheus.HistogramVec
+	NewSummary(opts prometheus.SummaryOpts) prometheus.Summary
+	NewSummaryVec(opts prometheus.SummaryOpts, labelNames []string) *prometheus.SummaryVec
+	Document() []DocumentedMetric
+}
```
samlaf: Is there any reason we need this interface? If it's never mocked and only implemented by a single struct (documentor) then I don't think its useful, and I would prefer we don't create unneeded interfaces.

api/clients/v2/metrics/metrics.go
```diff
@@ -1,5 +1,138 @@
...
+	Help   string   `json:"help"`
+	Labels []string `json:"labels"`
+}
+
+type documentor struct {
+	metrics []DocumentedMetric
+	factory promauto.Factory
+}
+
+func With(registry *prometheus.Registry) Factory {
```
samlaf:
```suggestion
func With(registry *prometheus.Registry) *documentor {
```

samlaf: Functions should very rarely return interfaces imo. This is an application of the https://en.wikipedia.org/wiki/Robustness_principle, which in golang terms specifies to https://bryanftan.medium.com/accept-interfaces-return-structs-in-go-d4cab29a301b

api/clients/v2/metrics/metrics.go
samlaf, line 1: Feel like this should be moved somewhere top level like `common/metrics`? It feels like a utility that could be used throughout the eigenda codebase afaiu?

api/clients/v2/metrics/metrics.go
iquidus: No. It was copy paste from op-service. Cleaner without the interface. Removed in [ddff1d9](https://github.com/Layr-Labs/eigenda/pull/1884/commits/ddff1d9f704a77c9851a9a6a10abd20f34c5e31f)

api/clients/v2/metrics/metrics.go
iquidus: Updated in [ddff1d9](https://github.com/Layr-Labs/eigenda/pull/1884/commits/ddff1d9f704a77c9851a9a6a10abd20f34c5e31f)

api/clients/v2/metrics/metrics.go
iquidus: Moved to to level `common/metrics` package in [cf4f140](https://github.com/Layr-Labs/eigenda/pull/1884/commits/cf4f140225ffa906d18ed7f47b27b6e256b317ab)

samlaf (approved):
LGTM although there's a conflict now

PR #1887, opened 2025-08-18
feat: cache operator state

cody-littley:

Instead of fetching operator state for each batch, instead fetch operator state on a background thread, and use the latest made available when it comes time to create a batch.

Should be merged in tandem with https://github.com/Layr-Labs/eigenda-devops/pull/1883

### DIFF
--- a/disperser/cmd/controller/main.go
+++ b/disperser/cmd/controller/main.go
@@ -95,7 +95,7 @@ func RunController(ctx *cli.Context) error {
 	}
 
 	operatorStateRetrieverAddress, err :=
-		contractDirectory.GetContractAddress(context.Background(), directory.OperatorStateRetriever)
+		contractDirectory.GetContractAddress(context.Background(), directory.BLSOperatorStateRetriever)
 	if err != nil {
 		return fmt.Errorf("failed to get BLSOperatorStateRetriever address: %w", err)
 	}


disperser/cmd/controller/main.go
```diff
@@ -82,8 +84,37 @@ func RunController(ctx *cli.Context) error {
 		logger.Error("Cannot create chain.Client", "err", err)
 		return err
 	}
+
+	contractDirectory, err := directory.NewContractDirectory(
+		context.Background(),
```
litt3, line 89: For this and subsequent contract calls, probably should use the `ctx` param, instead of a new `context.Background()`

disperser/cmd/controller/main.go
```diff
@@ -82,8 +84,37 @@ func RunController(ctx *cli.Context) error {
...
+	}
+
 	chainReader, err := eth.NewReader(
-		logger, gethClient, config.BLSOperatorStateRetrieverAddr, config.EigenDAServiceManagerAddr)
+		logger,
+		gethClient,
+		operatorStateRetrieverAddress.Hex(),
+		serviceManagerAddress.Hex())
 	if err != nil {
 		return err
```
litt3: Missing error wrapping (errs returned from this constructor aren't all that informative)

disperser/cmd/controller/main.go
```diff
@@ -195,11 +226,26 @@ func RunController(ctx *cli.Context) error {
 		return nil
 	}
 	dispatcherBlobSet := controller.NewBlobSet()
+
+	batchMetadataManager, err := metadata.NewBatchMetadataManager(
+		context.Background(),
```
litt3, line 231: reuse existing ctx

disperser/controller/metadata/batch_metadata_manager.go
```diff
@@ -0,0 +1,199 @@
...
+	// with the most recent onchain data is done on a best effort basis.
+	GetMetadata() *BatchMetadata
+
+	// Release resources associated with this manager.
+	Close()
+}
+
+var _ BatchMetadataManager = (*batchMetadataManager)(nil)
+
+// A standard implementation of the BatchMetadataManager interface. Does all metadata fetching in a background\
```
litt3: `\` ?

disperser/controller/metadata/batch_metadata_manager.go
```diff
@@ -0,0 +1,199 @@
...
+// goroutine, guaranteeing that GetMetadata() never blocks.
+type batchMetadataManager struct {
+	ctx    context.Context
+	logger logging.Logger
+
+	// The underlying eth client used to interact with the blockchain.
+	ethClient bind.ContractBackend
```
litt3: is `ethClient` the correct name for this?

disperser/controller/metadata/batch_metadata_manager.go
```diff
@@ -0,0 +1,199 @@
...
+	// The time between updates to the metadata.
+	updatePeriod time.Duration
+
+	// The most recent batch metadata.
+	metadata atomic.Pointer[BatchMetadata]
+
+	alive atomic.Bool
+}
+
+// Create a new BatchMetadataManager with the specified quorums.
```
litt3: > with the specified quorums

quorums don't appear to be involved here

disperser/controller/metadata/batch_metadata_manager.go
```diff
@@ -0,0 +1,199 @@
...
+//
+// updatePeriod is the period at which the manager will update its metadata. Actual update timing may vary
+// depending on the amount of time it takes to successfully get new data.
```
litt3: A little more info in this constructor doc would be nice:

- synchronously does the first fetch
- starts the background process to periodically update

disperser/controller/metadata/batch_metadata_manager.go
```diff
@@ -0,0 +1,199 @@
...
+}
+
+// Close releases resources associated with this manager.
+func (m *batchMetadataManager) Close() {
+	m.alive.Store(false)
+}
+
+// Fetch the next reference block number (RBN) to use.
+func (m *batchMetadataManager) getNextReferenceBlockNumber() (uint64, error) {
+	// Get the latest block header to determine the current reference block number.
```
litt3: > the current reference block number

I don't think it makes sense to talk about the "current reference block number". This is just the "current block number", and from that we *choose* an RBN

disperser/controller/metadata/batch_metadata_manager.go
```diff
@@ -0,0 +1,199 @@
...
+// This method is called periodically to ensure that metadata reflects a recent(ish) reference block.
+func (m *batchMetadataManager) updateMetadata() error {
+	referenceBlockNumber, err := m.getNextReferenceBlockNumber()
+	if err != nil {
+		return fmt.Errorf("failed to get next reference block number: %w", err)
+	}
+
+	previousMetadata := m.metadata.Load()
+	if previousMetadata != nil && referenceBlockNumber < previousMetadata.referenceBlockNumber {
+		// Only update if the new RBN is greater than the most recent one.
```
litt3, line 148: This would probably indicate a bug if it actually happened, right? Would it make sense to log?

litt3 (approved):

disperser/cmd/controller/main.go
cody-littley: Fun fact: `ctx *cli.Context` is not a subtype of `context.Context`. Their only similarity is that they share similar names.

Not sure the way we want to handle this. Perhaps we should rename the urfave thing to something like `cliParser` or similar. This pattern shows up all over the place, so we should probably figure out a pattern to apply in this situation.

disperser/cmd/controller/main.go
cody-littley: fixed errors returned by this constructor

disperser/cmd/controller/main.go
litt3: Gotcha. Just ignore my comment then, not worth a lot of effort to overcome this IMO

disperser/cmd/controller/main.go
cody-littley: same comment as a above, existing `ctx` is not a real context

disperser/controller/metadata/batch_metadata_manager.go
cody-littley: removed

disperser/controller/metadata/batch_metadata_manager.go
cody-littley:
```
// ContractBackend defines the methods needed to work with contracts on a read-write basis.
type ContractBackend interface {
```

I'm more naming it for what it does. The name they give to the interface isn't very informative, IMO. Still willing to consider alternate names if you have any suggestions.

disperser/controller/metadata/batch_metadata_manager.go
cody-littley: Leftover from a previous iteration that did track quorums. Removed.

disperser/controller/metadata/batch_metadata_manager.go
cody-littley:
```
// Create a new BatchMetadataManager.
//
// This constructor does an initial blocking metadata fetch, so that any call to GetMetadata() after this constructor
// returns can immediately return valid metadata. It also starts a background goroutine that periodically updates the
// metadata at a rate defined by updatePeriod. Actual update timing may vary depending on the amount of time it
// takes to successfully get new data.
```

disperser/controller/metadata/batch_metadata_manager.go
cody-littley:
```
	// Get the latest block header to find the latest block number, 
	// which we use to choose a good current reference block number.
```

disperser/controller/metadata/batch_metadata_manager.go
cody-littley: Most likely this will trigger when there are no new blocks (e.g. if the blockchain stalls or if we get disconnected from the internet), but I agree that a log here is useful.

```
		m.logger.Errorf("new reference block number %d is not greater than the previous one %d, skipping update",
			referenceBlockNumber, previousMetadata.referenceBlockNumber)
```

core/thegraph/state.go
```diff
@@ -141,7 +141,8 @@ func (ics *indexedChainState) GetIndexedOperatorState(ctx context.Context, block
 		}
 	}
 	if len(aggKeys) == 0 {
-		return nil, fmt.Errorf("no aggregate public keys found for any of the specified quorums at block number %d", blockNumber)
+		ics.logger.Warnf("no aggregate public keys found for any of the specified quorums at block number %d",
```
litt3, line 144: Could you explain the rationale here? It seems like if we don't have the aggregate pubkeys, then we won't be able to interact correctly with the state. We _need_ the aggregate keys, right?

disperser/controller/metadata/batch_metadata_manager.go
litt3: Is there any reason you are using _that_ interface?

AFAICT, you could just change the type everywhere in this file to `common.EthClient`. What you're passing into the constructor implements that interface too, so you don't even need to change what you're passing in.

Then, when this is a `common.EthClient` instance, you'd have access to the `BlockNumber()` method, which will make actually getting the block number much more straightforward.

cody-littley: Per our discussion, I renamed `ethClient` to `contractBackend`. I also split the logic for looking up the RBN into a stand alone file so that we don't reimplement this utility functionality in the future.

core/thegraph/state.go
cody-littley: As far as I can tell, this can happen if the controller is started up before validators are registered. This happens during the inabox suite.

Previously, this wasn't a problem, since we would never attempt to get the operator state until after validators were first registered. But this is a problem with this PR, since we intentionally pre-fetch the operator state when we call controller's constructor.

I attempted to resolve this problem by looking at the inabox setup logic, but gave up after 30 minutes of unsuccessful spaghetti logic traversal. This is purely a change to get inabox testes passing.

litt3 (approved):

core/eth/directory/contract_names.go
```diff
@@ -6,16 +6,18 @@ package directory
 // When you add to this list, make sure you keep things in alphabetical order.
 
 const (
-	EigenDAEjectionManager ContractName = "EIGEN_DA_EJECTION_MANAGER"
-	OperatorStateRetriever ContractName = "OPERATOR_STATE_RETRIEVER"
-	RegistryCoordinator    ContractName = "REGISTRY_COORDINATOR"
-	ServiceManager         ContractName = "SERVICE_MANAGER"
+	BLSOperatorStateRetriever ContractName = "OPERATOR_STATE_RETRIEVER"
```
litt3, line 9: Are you sorting alphabetically based on variable name, or contract name?

cody-littley: Variable names. Although I think it's odd that the variable name doesn't match the contract name. Let's follow up on this in PR N+1.

PR #1890, opened 2025-08-18
feat(metrics): record CheckDACert gas usage

iquidus:
Closes DAINT-682

Adds `EstimateGasCheckDACert` to cert verifier for estimating gas usage by `CheckDACert`

Adds new metric to v2 test client for reporting gas estimate of the CheckDACert call

Adds `EstimateGasCheckDACert` to v2 test client.

Adds call to test client's `EstimateGasCheckDACert` to v2 load generator

We have no insight into how much gas is required for CheckDACert call. As it's a read method we can not see this on-chain.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/clients/v2/verification/cert_verifier.go
```diff
@@ -107,6 +108,60 @@ func (cv *CertVerifier) CheckDACert(
 	return nil
 }
 
+// EstimateGasCheckDACert uses eth_estimateGas to estimate the gas requirements for CheckDACert.
```
ethenotethan: nit: 
```suggestion
// EstimateGasCheckDACert uses eth_estimateGas to estimate the gas amount required for a CheckDACert call.
```

api/clients/v2/verification/cert_verifier.go
```diff
@@ -107,6 +108,60 @@ func (cv *CertVerifier) CheckDACert(
...
+		ctx,
+		certV3.ReferenceBlockNumber(),
+	)
+	if err != nil {
+		return 0, fmt.Errorf("get cert verifier address: %w", err)
+	}
+
+	certBytes, err := certV3.Serialize(coretypes.CertSerializationABI)
+	if err != nil {
+		return 0, fmt.Errorf("serialize cert: %w", err)
```
ethenotethan: nit - this is duplicate from the verify function. could be beneficial to create a shared method for this

iquidus: It's not all duplicate code, specifically the `certVerifierAddress` differs from the verify functions `certVerifierCaller`. I have added a `NormalizeCertV3` func in  [ce50098](https://github.com/Layr-Labs/eigenda/pull/1890/commits/ce500983b1deaafdc2e393c826c5dae2285dc9db) for the cert normalization logic.

test/v2/client/test_client_metrics.go
```diff
@@ -194,6 +195,14 @@ func newTestClientMetrics(logger logging.Logger, port int) *testClientMetrics {
 		[]string{},
 	)
 
+	gasCheckDACert := promauto.With(registry).NewGauge(
+		prometheus.GaugeOpts{
+			Namespace: namespace,
+			Name:      "gas_checkdacert",
+			Help:      "Gas estimate for CheckDACert call",
```
samlaf, line 202: nit: is the value we get from the estimate rpc call actually an estimate? Or is it the exact value? Might be useful to say that its an exact gas cost value coming from an estimate rpc call? Otherwise I can see myself in 2 months looking at this metric on grafana and not knowing whether its a good or bad estimate

test/v2/load/load_generator.go
```diff
@@ -257,6 +257,13 @@ func (l *LoadGenerator) disperseBlob(rand *random.TestRandom) (
 		return corev2.BlobKey{}, nil, nil, fmt.Errorf("failed to compute blob key: %w", err)
 	}
 
+	// Estimate gas for CheckDACert call
+	_, err = l.client.EstimateGasCheckDACert(ctx, eigenDAV3Cert)
+	if err != nil {
+		l.client.GetLogger().Errorf("failed to estimate gas for CheckDACert call: %v", err)
+		return corev2.BlobKey{}, nil, nil, fmt.Errorf("failed to estimate gas for CheckDACert call: %w", err)
+	}
```
samlaf: do we really want to return an error if metric failed? I think just logging and error should be fine?

test/v2/load/load_generator.go
```diff
@@ -257,6 +257,13 @@ func (l *LoadGenerator) disperseBlob(rand *random.TestRandom) (
 		return corev2.BlobKey{}, nil, nil, fmt.Errorf("failed to compute blob key: %w", err)
 	}
 
+	// Estimate gas for CheckDACert call
+	_, err = l.client.EstimateGasCheckDACert(ctx, eigenDAV3Cert)
```
samlaf: this will add latency to dispersals. Was thinking we might want to do this in a background goroutine. @cody-littley thoughts?

api/clients/v2/verification/cert_verifier.go
samlaf: Too bad the contractClient doesn't expose an estimate gas call

test/v2/load/load_generator.go
cody-littley: If this is making an eth RPC, doing it in a background goruoutine is preferable. I use this to measure system end to end latency, and so the less noise in that signal the better.

If this is just doing local computations, I'm ok leaving it on the main goroutine if you can pinky-swear that it takes less than a millisecond.

samlaf: it is making an eth rpc call. perhaps we should rename the function to make that more explicit?

api/clients/v2/verification/cert_verifier.go
```diff
@@ -107,6 +94,50 @@ func (cv *CertVerifier) CheckDACert(
...
+	}
+
+	// Estimate gas using eth_estimateGas
+	gasEstimate, err := cv.ethClient.EstimateGas(ctx, ethereum.CallMsg{
+		To:   &certVerifierAddress,
+		Data: callData,
+	})
+	if err != nil {
+		return 0, fmt.Errorf("estimate gas for checkDACert: %w", err)
+	}
```
samlaf: oh just remembered. add a log.Error with the CallMsg arguments. We unfortunately don't have this in most of the codebase but we're trying to add this logging everywhere we do an rpc call. So that when they fail we can debug by going to tenderly or whatever

test/v2/client/test_client_metrics.go
iquidus: Generally, the node reports the minimum gas required to just barely succeed. However it's based on state at the time of the call. It can also vary depending on what block tag it's called against, e.g latest vs pending (pending accounts for mempool txs). 

Gas refunds (clearing storage for example), reduces cost but are not perfectly predictable during the estimate.

Geth also adds a ~100k buffer.

test/v2/load/load_generator.go
iquidus: Moved to background goroutine in [c2435a1](https://github.com/Layr-Labs/eigenda/pull/1890/commits/c2435a110a1f049a288b6bd9f09afc0235fdfbc0)

api/clients/v2/verification/cert_verifier.go
iquidus: Added log in [af973e3](https://github.com/Layr-Labs/eigenda/pull/1890/commits/af973e31f61e8f2794c6f1cac572a2aa2d7d364d)

test/v2/load/load_generator.go
iquidus: Removed return in [6bdc79c](https://github.com/Layr-Labs/eigenda/pull/1890/commits/6bdc79ccfa93e8247c1320ddb08b5a283ebb18d0)

api/clients/v2/verification/cert_verifier.go
```diff
@@ -107,6 +95,57 @@ func (cv *CertVerifier) CheckDACert(
...
+		Data: callData,
+	}
+
+	// Estimate gas using eth_estimateGas
+	gasEstimate, err := cv.ethClient.EstimateGas(ctx, callMsg)
+	if err != nil {
+		cv.logger.Error(
+			"eth_estimateGas",
+			zap.String("to", callMsg.To.Hex()),
+			zap.String("data", fmt.Sprintf("0x%x", callMsg.Data)),
```
samlaf:
```suggestion
			"to", callMsg.To.Hex(),
			"data", fmt.Sprintf("0x%x", callMsg.Data),
```

samlaf: why are you using zap? Think its actually wrong here. We use slog most of the time, but also the interface for our logger.Error takes key value pairs, not a zap.String afaiu.

test/v2/client/test_client.go
```diff
@@ -914,3 +914,16 @@ func (c *TestClient) GetProxyWrapper() (*ProxyWrapper, error) {
 	}
 	return c.proxyWrapper, nil
 }
+
+func (c *TestClient) EstimateGasCheckDACert(
```
samlaf:
```suggestion
func (c *TestClient) EstimateGasAndReportCheckDACert(
```

test/v2/load/load_generator.go
```diff
@@ -257,9 +261,29 @@ func (l *LoadGenerator) disperseBlob(rand *random.TestRandom) (
...
 
+// estimateGasCheckDACertAsync performs gas estimation in a separate goroutine to avoid blocking blob dispersal.
+func (l *LoadGenerator) estimateGasCheckDACertAsync(eigenDAV3Cert *coretypes.EigenDACertV3) {
+	l.gasEstimationLimiter <- struct{}{}
+	defer func() {
+		<-l.gasEstimationLimiter
+	}()
+
+	gasTimeout := time.Duration(l.config.GasEstimationTimeout) * time.Second
+	ctx, cancel := context.WithTimeout(context.Background(), gasTimeout)
```
samlaf:
```suggestion
	ctx, cancel := context.WithTimeout(l.ctx, gasTimeout)
```

samlaf: Generally prefer not to store contexts in structs, following arguments in https://go.dev/blog/context-and-structs, but in any case the ctx does exist so we should def use it instead of context.Background

@cody-littley did you have a good reason for storing this one? Is it because its the main context that is used to cancel all goroutines when a sigterm signal is received?

test/v2/load/load_generator.go
```diff
@@ -257,9 +261,29 @@ func (l *LoadGenerator) disperseBlob(rand *random.TestRandom) (
...
 	}
 
+	// Estimate gas for CheckDACert call
+	go l.estimateGasCheckDACertAsync(eigenDAV3Cert)
+
 	return blobKey, payload, eigenDACert, nil
 }
 
+// estimateGasCheckDACertAsync performs gas estimation in a separate goroutine to avoid blocking blob dispersal.
+func (l *LoadGenerator) estimateGasCheckDACertAsync(eigenDAV3Cert *coretypes.EigenDACertV3) {
```
samlaf:
```suggestion
// estimateAndReportGasCheckDACert performs gas estimation and reports it as a metric.
// Make sure to call this in a separate goroutine to avoid blocking blob dispersal.
func (l *LoadGenerator) estimateAndReportGasCheckDACert(eigenDAV3Cert *coretypes.EigenDACertV3) {
```

samlaf: the function is not async right..? it is blocking the way written, so needs to be called in goroutine

api/clients/v2/verification/cert_verifier.go
iquidus: Oh my bad. Used zap due to this [README](https://github.com/Layr-Labs/eigensdk-go/blob/dev/logging/README.md)

samlaf: Oh yikes that's actually super outdated. We moved to using slog as soon as it came out in go1.21

iquidus: updated in [254d7fb](https://github.com/Layr-Labs/eigenda/pull/1890/commits/254d7fb6027bf41789fc7f4f7a2d3ee4a531f9d9)

test/v2/load/load_generator.go
iquidus: suggested change included in [85e3469](https://github.com/Layr-Labs/eigenda/pull/1890/commits/85e3469d534a655ed598270c9e9f43fcbcafae39)

test/v2/load/load_generator.go
iquidus: suggested change included in [85e3469](https://github.com/Layr-Labs/eigenda/pull/1890/commits/85e3469d534a655ed598270c9e9f43fcbcafae39)

test/v2/client/test_client.go
iquidus: done in [85e3469](https://github.com/Layr-Labs/eigenda/pull/1890/commits/85e3469d534a655ed598270c9e9f43fcbcafae39)

samlaf (approved):

PR #1895, opened 2025-08-19
Inabox using testcontainers

dmanc:

WIP: Attempt at using testcontainers for managing docker containers for the inabox test framework. Currently only added anvil, graph-node. and localstack.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/common/testinfra/README.md
+++ b/common/testinfra/README.md
@@ -1,0 +1,238 @@
+# EigenDA Test Infrastructure
+
+This package provides a modular, testcontainer-based infrastructure for EigenDA testing.
+
+## Features
+
+- **Modular Design**: Enable only the infrastructure components you need
+- **Testcontainers Integration**: Automatic container lifecycle management  
+- **Parallel Test Support**: Isolated containers for each test
+- **Type-Safe Configuration**: Go structs instead of YAML/env files
+- **Automatic Cleanup**: Containers destroyed after tests complete
+
+## Infrastructure Components
+
+### 1. Anvil Blockchain
+- Foundry's Anvil for local Ethereum testing
+- Configurable chain ID, block time, gas settings
+- Pre-funded accounts with deterministic private keys
+- Optional mainnet forking support
+
+### 2. LocalStack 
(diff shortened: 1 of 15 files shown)


inabox/deploy/utils.go
```diff
@@ -123,6 +123,7 @@ func execYarnCmd(command string, args ...string) {
 // Executes a forge script with a given rpc and private key
 func execForgeScript(script, privateKey string, deployer *ContractDeployer, extraArgs []string) {
 	log.Printf("Executing forge script with params %s, %s, %s", script, deployer.RPC, privateKey)
+	log.Printf("Full forge command will be: forge script %s --rpc-url %s --private-key %s --broadcast", script, deployer.RPC, privateKey)
```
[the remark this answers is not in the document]
dmanc: Test code

common/testinfra/README.md
```diff
@@ -0,0 +1,238 @@
...
+
+    // One line replaces all the above
+    manager, result, err := testinfra.StartMinimal(ctx)
+    require.NoError(t, err)
+    defer manager.Stop(ctx)
+    
+    // All services ready to use
+    runYourTestLogic(result.AnvilRPC, result.LocalStackURL)
+}
+```
```
samlaf: I think the reason why I liked docker compose approach more is that it can be used as a standalone local devnet. Whereas starting it like this via golang containers, how do you just let them run in the background, and eventually kill them? I guess we could have a main script which just starts everything and blocks, and when killed will kill all the containers. Doesn't seem super clean though. Wonder if we can have something like `docker compose up -d`

dmanc: It's definitely possible to let the containers live, I believe it's a matter of disabling the resource reaper https://golang.testcontainers.org/features/configuration/#customizing-ryuk-the-resource-reaper.

But also running a devnet is a separate problem that could be solved with testcontainers but possibly not the best solution.

samlaf: Oh nice, Claude saying we could clean up like this:
image

Not the cleanest but good to know its possible.

PR #1896, opened 2025-08-19
feat: litt push command

cody-littley:
Adds the `litt push` command, works similar to rsync but on littDB files.


litt/cli/push.go
```diff
@@ -0,0 +1,346 @@
...
+	// Lock source files. It would be nice to also lock the remote directories, but that's tricky given that
+	// we are interacting with the remote machine via SSH and rsync.
+	releaseSourceLocks, err := util.LockDirectories(logger, sources, util.LockfileName, fsync)
+	if err != nil {
+		return fmt.Errorf("failed to lock source directories: %v", err)
+	}
+	defer releaseSourceLocks()
+
+	// Create an SSH session to the remote host.
+	connection, err := util.NewSSHSession(logger, user, host, port, keyPath, "TODO fix before merge", verbose)
```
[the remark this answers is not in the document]
cody-littley: fixed

litt/cli/push.go
```diff
@@ -0,0 +1,366 @@
...
+		tableDestination := path.Join(dest, tableName, segment.SegmentDirectory)
+		filePaths, err := connection.FindFiles(tableDestination, extensions)
+		if err != nil {
+			return nil, fmt.Errorf("failed to list files in destination %s: %w", dest, err)
+		}
+
+		for _, filePath := range filePaths {
+			// Extract the file name from the path.
+			fileName := path.Base(filePath)
+			if _, exists := existingFiles[fileName]; !exists {
```
litt3: Is this logic backwards? Looks like you're adding to `existingFiles` all files that _don't_ exist

litt/cli/push.go
```diff
@@ -0,0 +1,366 @@
...
+	}
+
+	// Special handling if we are transferring data from a snapshot.
+	isSnapshot, err := segments[lowestSegmentIndex].IsSnapshot()
+	if err != nil {
+		return fmt.Errorf("failed to check if segment %d is a snapshot: %w", lowestSegmentIndex, err)
+	}
+	if isSnapshot {
+		if len(sources) > 1 {
+			return fmt.Errorf("table %s is a snapshot, but source directories found: %v", tableName, sources)
```
litt3: but _multiple_ source directories found

litt/cli/push.go
```diff
@@ -0,0 +1,366 @@
...
+	}
+
+	return nil
+}
+
+// Deletes local segments after they have been successfully transferred to the remote destination(s).
+func deleteLocalSegments(
+	segments map[uint32]*segment.Segment,
+	tableName string,
+	isSnapshot bool,
```
litt3, line 334: looks like in practice you don't permit this to be called _unless_ it's a snapshot, which seems to indicate that the `isSnapshot` param actually isn't necessary?

litt/cli/push.go
```diff
@@ -0,0 +1,366 @@
...
+		err := seg.BlockUntilFullyDeleted()
+		if err != nil {
+			return fmt.Errorf("failed to delete segment %d for table %s: %w",
+				seg.SegmentIndex(), tableName, err)
+		}
+	}
+
+	if isSnapshot {
+		// If we are dealing with a snapshot, update the lower bound file.
+		boundaryFile, err := disktable.LoadBoundaryFile(true, path.Join(sources[0], tableName))
```
litt3: You're using a raw bool parameter, when you have a `disktable.BoundaryType` defined. Also elsewhere in this file.

litt/cli/push.go
```diff
@@ -0,0 +1,366 @@
...
+	return nil
+}
+
+// Deletes local segments after they have been successfully transferred to the remote destination(s).
+func deleteLocalSegments(
+	segments map[uint32]*segment.Segment,
+	tableName string,
+	isSnapshot bool,
+	sources []string,
+	destinations []string,
```
litt3: I'm confused why you're passing in and logging `destinations` here. Seems like all you need are the `sources`

litt3 (approved):

litt/cli/push.go
cody-littley: ~I think this is backwards. I didn't test this case, which explains how the bug slipped by. I'll add a test for this.~

litt/cli/push.go
cody-littley:
```
			return fmt.Errorf("table %s is a snapshot, but source more than one source directories found: %v", 
				tableName, sources)
```

litt/cli/push.go
cody-littley:
```
enforce.True(isSnapshot, "we should have already returned an error if this is a non-snapshot table")
```

litt/cli/push.go
cody-littley: fixed

litt/cli/push.go
cody-littley: Ok, so it turns out I did test this specifically, and those tests started failing when I inverted this condition. The variable naming is definitely bad here, since it confused even me for a while.

I think it should be more clear with the reframing:

```
			enforce.MapDoesNotContainKey(existingFiles, fileName,
				"duplicate file found: %s and %s", fileName, existingFiles[fileName])
```

The former variable `exists` was only true if you find a file at the destination more than once. If it's not, it's probably a signal that there is data corruption, or will likely be data corruption if we continue.

litt/cli/push.go
cody-littley: You are correct... I was logging the wrong thing here. Fixed.

litt3 (approved):

PR #1897, opened 2025-08-19
feat(payments): Implement Controller GRPC to handle payments

litt3:
- closes EGDA-1786

DO NOT REVIEW YET


disperser/controller/metering/grpc_server.go
```diff
@@ -0,0 +1,130 @@
...
+func (s *GrpcServer) Stop() {
+	if s.server != nil {
+		s.server.GracefulStop()
+	}
+	if s.listener != nil {
+		_ = s.listener.Close()
+	}
+}
+
+func (s *GrpcServer) AuthorizePayment(ctx context.Context, req *pb.AuthorizePaymentRequest) (*pb.AuthorizePaymentReply, error) {
```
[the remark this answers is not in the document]
litt3: We are checking the signature from the API server. Since the API server and the controller are both considered part of the same entity, the controller can trust the API server. The API server is responsible for verifying the account signature before submitting a payment for authentication

api/proto/controller/v1/controller_service.proto
```diff
@@ -0,0 +1,36 @@
+syntax = "proto3";
+package controller.v1;
```
cody-littley: We probably shouldn't use a v1/v2 postfix for the controller.

api/proto/controller/v1/controller_service.proto
```diff
@@ -0,0 +1,36 @@
...
+
+import "common/v2/common_v2.proto";
+
+option go_package = "github.com/Layr-Labs/eigenda/api/grpc/controller/v1";
```
cody-littley: no `v1`

api/proto/controller/v1/controller_service.proto
cody-littley: Even if we eventually have new "versions" of the controller API, since this is internal we can make breaking changes as long as we update the entire disperser/full node at the same time.

api/proto/controller/v1/controller_service.proto
```diff
@@ -0,0 +1,36 @@
...
+
+  // Signature over the keccak hash of this AuthorizePaymentRequest (excluding this signature field).
+  bytes disperser_signature = 2;
+}
+
+// AuthorizePaymentResponse is returned after the controller does accounting and metering.
+//
+// A GRPC error indicates that there was a problem with the payment authorization. No error means everything succeeded.
+//
+// TODO(litt3): We should document the potential errors that could be returned.
```
cody-littley: In general, we should be careful about putting TODOs in protobuf docs, since this is public documentation. Perhaps add the TODO as a stand alone comment that isn't part of the message doc?

api/proto/controller/v1/controller_service.proto
```diff
@@ -0,0 +1,36 @@
...
+
+// Contains all information necessary for the controller to evaluate the validity of a dispersal payment
+message AuthorizePaymentRequest {
+  // The blob header is used for the following purposes:
+  // 1. Contains the PaymentHeader, which describes the payment being offered
+  // 2. Contains the quorums being dispersed to
+  common.v2.BlobHeader blob_header = 1;
+
+  // Signature over the keccak hash of this AuthorizePaymentRequest (excluding this signature field).
+  bytes disperser_signature = 2;
```
cody-littley: We should include a timestamp in any protobuf where we include a signature. This is needed by the replay gaurdian to prevent signature replay attacks. 

This is likely getting into overkill territory from a security point of view. But perhaps overkill is justified for gRPCs, since gRPCs are one of the most likely points of attack in a system.

cody-littley: For the type of the timestap, I suggest using a `uint64` holding the time in seconds since the epoch.

disperser/apiserver/controller_client.go
```diff
@@ -0,0 +1,123 @@
...
+}
+
+// Creates a client for communicating with the controller GRPC server
+func NewControllerClient(
+	ctx context.Context,
+	controllerAddress string,
+	kmsRegion string,
+	kmsEndpoint string,
+	disperserKMSKeyID string,
+) (*ControllerClient, error) {
```
cody-littley: Would it make sense to load the key from kms outside this constructor, and pass it in? Not sure how we do it elsewhere, but by loading it externally it will probably be easier to unit test.

disperser/apiserver/controller_client.go
cody-littley, line 1: I suggest putting this into `disperser/controller`, not `disperser/apiserver`.

disperser/apiserver/disperse_blob_v2.go
```diff
@@ -38,9 +38,19 @@ func (s *DispersalServerV2) DisperseBlob(ctx context.Context, req *pb.DisperseBl
...
+		if err != nil {
+			// Pass through the structured error from the controller
+			return nil, err
+		}
+	} else {
+		// Use the legacy payment metering system
+		// Check against payment meter to make sure there is quota remaining
+		if err := s.checkPaymentMeter(ctx, req, start); err != nil {
+			return nil, err
+		}
```
cody-littley: Once this feature is ready to ship, I suggest prioritizing the removal of the legacy pathway ASAP.

disperser/cmd/controller/main.go
```diff
@@ -275,6 +277,34 @@ func RunController(ctx *cli.Context) error {
...
+			paymentAuthorizationHandler)
+		if err != nil {
+			return fmt.Errorf("create gRPC server: %w", err)
+		}
+
+		go func() {
+			logger.Info("Starting controller gRPC server", "port", config.ServerConfig.GrpcPort)
+			if err := grpcServer.Start(); err != nil {
+				logger.Error("gRPC server failed", "error", err)
+			}
```
cody-littley: Perhaps a good time to panic? If the gRPC server is not online, the disperser as a whole is probably in a broken state. (It's possible you were following examples from elsewhere that don't handle this well.)

disperser/controller/grpcserver/server.go
```diff
@@ -0,0 +1,200 @@
...
+	metrics                     *metrics.ServerMetrics
+	replayGuardian              replay.ReplayGuardian
+	disperserAddress            gethcommon.Address
+}
+
+func NewServer(
+	ctx context.Context,
+	config Config,
+	awsClientConfig aws.ClientConfig,
+	disperserKMSKeyID string,
```
cody-littley: Consider moving kms stuff up one level, and just passing in keys.

disperser/controller/grpcserver/config.go
```diff
@@ -0,0 +1,84 @@
...
+	EnableServer bool
+
+	// If true, use the new payment authentication system running on the controller.
+	// If false, payment authentication is disabled and request validation will always fail
+	//
+	// Note: This flag requires EnableServer to be true in order to function.
+	EnablePaymentAuthentication bool
+
+	// Port that the gRPC server listens on
+	GrpcPort string
```
cody-littley: Pet peeve, we should always define ports as integers. Many places in the code record them as strings, and I've been changing them as I find them.

disperser/controller/grpcserver/server.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	err = s.verifyDisperserSignature(requestHash, request.GetDisperserSignature())
+	if err != nil {
+		s.metrics.ReportAuthorizePaymentSignatureFailure()
+		return nil, err
+	}
+	s.metrics.ReportAuthorizePaymentSignatureLatency(time.Since(start))
+
+	timestamp := time.Unix(0, request.GetBlobHeader().GetPaymentHeader().GetTimestamp())
```
cody-littley: Ah, ok. You are using the nested timestamp. I'm ok with this.

disperser/controller/metrics/server_metrics.go
```diff
@@ -0,0 +1,112 @@
...
+	authorizePaymentLatency           *prometheus.SummaryVec
+	authorizePaymentSignatureLatency  *prometheus.SummaryVec
+	authorizePaymentAuthFailures      *prometheus.CounterVec
+	authorizePaymentSignatureFailures *prometheus.CounterVec
+}
+
+func NewServerMetrics(registry *prometheus.Registry, logger logging.Logger) *ServerMetrics {
+	if registry == nil {
+		registry = prometheus.NewRegistry()
+	}
```
cody-littley:
```suggestion
	if registry == nil {
		return nil
	}
```

Then, make sure in each of the methods you handle the case where `m == nil`.

Creating new registries is a foot-gun pattern, since if it happens in a production environment then you may get funky results. Better to be a no-op than to register a second registry.

disperser/controller/metrics/server_metrics.go
```diff
@@ -0,0 +1,112 @@
...
+	registry.MustRegister(grpcMetrics)
+	grpcServerOption := grpc.UnaryInterceptor(
+		grpcMetrics.UnaryServerInterceptor(),
+	)
+
+	objectives := map[float64]float64{0.5: 0.05, 0.9: 0.01, 0.99: 0.001}
+
+	authorizePaymentLatency := promauto.With(registry).NewSummaryVec(
+		prometheus.SummaryOpts{
+			Namespace:  namespace,
```
dmanc, line 44: nit: There's another input called `Subsystem`, so you can define Subsystem = "authorize_payment".

```
	// Namespace, Subsystem, and Name are components of the fully-qualified
	// name of the Summary (created by joining these components with
	// "_"). Only Name is mandatory, the others merely help structuring the
	// name. Note that the fully-qualified name of the Summary must be a
	// valid Prometheus metric name.
```

api/proto/controller/v1/controller_service.proto
litt3: It's required by the `buf` tool we have running https://buf.build/docs/best-practices/style-guide/

Not sure when this was put in place, or what exactly it's doing

litt3: cc @samlaf Did you add this?

api/proto/controller/v1/controller_service.proto
litt3: e07d7e3a171f4739dbd914d7121ec118b76fc3b7

disperser/apiserver/controller_client.go
litt3: 40622a72c1d593c1c86019e4043b012cb5670409

Reorganized, but I need some help from the naming committee here (summoning @samlaf). I made the package `controller/service` which isn't great, but I haven't thought of anything better yet.

api/proto/controller/v1/controller_service.proto
litt3: see https://github.com/Layr-Labs/eigenda/pull/1897#discussion_r2340771296

api/proto/controller/v1/controller_service.proto
litt3: https://github.com/Layr-Labs/eigenda/pull/1897#discussion_r2338071474 for reference

disperser/controller/metrics/server_metrics.go
litt3: 67864e3b556e3f1a69a1654295c5f9d2f0d1920f

disperser/controller/grpcserver/config.go
litt3: 20264471b9245ad8b914b75ec452cbfcc4487bde

I went with `uint16` since that's the proper size for a port. I see you've used `uint64` elsewhere. Any particular reason you made that choice?

disperser/apiserver/controller_client.go
litt3: Much cleaner 👍 d08e0ce8df22fd8ab3a82981fc2b87c709b4181d

disperser/controller/grpcserver/server.go
litt3: d08e0ce8df22fd8ab3a82981fc2b87c709b4181d

disperser/cmd/controller/main.go
litt3: Just to verify, are you aware that `Start` blocks for the lifetime of the server? So a panic here would mean that the server crashing at any time crashes the whole controller (not just if it fails to start)

Maybe that's what we want? Seems reasonable enough

api/proto/controller/v1/controller_service.proto
cody-littley: Resolving this thread, we can have the discussion in the linked comment

disperser/apiserver/controller_client.go
cody-littley: Possible alternatives:

- `controller/server`
- `controller/grpc`

disperser/cmd/controller/main.go
cody-littley: If the controller stops being able to handle payments requests, the disperser as a whole is in a broken state that is unable to do new work. IMO, better at that point in time to tear everything down and restart the pod.

disperser/controller/grpcserver/config.go
cody-littley: I usually default to `uint64` unless I have reasons to use something different. Makes sense to use `uint16` for ports, I will start converting ports I come in contact with to this type.

disperser/apiserver/controller_client.go
litt3: - originally decided against `server`, since it feels weird to have `server.Client` verbiage, but I would be ok with it if the committee is for it
- decided against `grpc`, since it clashes with the actual `grpc` packages. but maybe we don't care about that?

disperser/cmd/controller/main.go
litt3: 51e3d62937e4f6592a5d22981f6e2757a87085c4

api/proto/controller/v1/controller_service.proto
samlaf: I did yes. its just a linter for protobufs. We can disable this rule if we haven't been following it. But typically it gives reasons and they make a lot of sense. aka everything should be versioned.

api/proto/controller/v1/controller_service.proto
```diff
@@ -0,0 +1,37 @@
...
+  // Client's ECDSA signature over the blob header's blobKey (keccak hash of the blob header).
+  // This signature can be verified against the account ID in the payment header.
+  bytes client_signature = 2;
+}
+
+// AuthorizePaymentResponse is returned after the controller does accounting and metering.
+//
+// A GRPC error indicates that there was a problem with the payment authorization. No error means everything succeeded.
+message AuthorizePaymentResponse {
+  // TODO(litt3): We should document the potential errors that could be returned.
```
samlaf: See https://grpc.github.io/grpc/core/md_doc_statuscodes.html
Maybe use UNAUTHENTICATED?
PERMISSION_DENIED we might want to keep open in case we ever introduce jwt authentication between microservices?

api/proto/controller/v1/controller_service.proto
```diff
@@ -0,0 +1,37 @@
...
+// ControllerService defines the APIs for the controller.
+//
+// Currently, this API is only intended for *internal* consumption: this is a way for different parts of the disperser
+// to communicate with each other
+service ControllerService {
+  // AuthorizePayment handles payment authorization for blob dispersal
+  //
+  // This is intended to be called by API server instances that are handling dispersal requests. The controller
+  // is responsible for accounting and metering for the dispersal.
+  rpc AuthorizePayment(AuthorizePaymentRequest) returns (AuthorizePaymentResponse) {}
```
samlaf: can you spell out what accounting and metering mean here? I keep forgetting.. :\
Also should we also mention why we decided to go without auth (other than validating the user request)?

api/proto/controller/v1/controller_service.proto
```diff
@@ -0,0 +1,37 @@
...
+
+import "common/v2/common_v2.proto";
+
+option go_package = "github.com/Layr-Labs/eigenda/api/grpc/controller/v1";
+
+// ControllerService defines the APIs for the controller.
+//
+// Currently, this API is only intended for *internal* consumption: this is a way for different parts of the disperser
+// to communicate with each other
+service ControllerService {
```
samlaf: wondering if we shouldn't reorganize our proto folder into external/ and internal/. Thoughts?

disperser/controller/service/server.go
```diff
@@ -0,0 +1,151 @@
...
+	success := false
+	defer func() {
+		if success {
+			s.metrics.ReportAuthorizePaymentLatency(time.Since(start))
+		} else {
+			s.metrics.ReportAuthorizePaymentAuthFailure()
+		}
+	}()
+
+	if s.paymentAuthorizationHandler == nil {
```
samlaf: why not force this to be non-nil in constructor?

disperser/controller/service/config.go
```diff
@@ -0,0 +1,81 @@
...
+import (
+	"fmt"
+	"time"
+)
+
+// Contains configuration for the controller gRPC server
+type Config struct {
+	// Enables the gRPC server
+	//
+	// TODO(litt3): the option to disable the server will be removed once any feature has been fully rolled out which
```
dmanc: Why do we need the option to disable in the first place?

api/hashing/authorize_payment_request_hashing.go
```diff
@@ -0,0 +1,26 @@
...
+	"fmt"
+
+	controller "github.com/Layr-Labs/eigenda/api/grpc/controller/v1"
+	"golang.org/x/crypto/sha3"
+)
+
+// ControllerAuthorizePaymentRequestDomain is the domain for hashing AuthorizePaymentRequest messages.
+const ControllerAuthorizePaymentRequestDomain = "controller.AuthorizePaymentRequest"
+
+func HashAuthorizePaymentRequest(request *controller.AuthorizePaymentRequest) ([]byte, error) {
```
samlaf, line 13: Why is this defined here instead of in the controller next to where its used (which iiuc is the only place we plan to use it)?

disperser/controller/service/server.go
```diff
@@ -0,0 +1,151 @@
...
+		opts = append(opts, grpc.MaxRecvMsgSize(s.config.MaxGRPCMessageSize))
+	}
+
+	if s.config.MaxIdleConnectionAge > 0 {
+		opts = append(opts, grpc.KeepaliveParams(keepalive.ServerParameters{
+			MaxConnectionIdle: s.config.MaxIdleConnectionAge,
+		}))
+	}
+
+	s.server = grpc.NewServer(opts...)
```
dmanc:
```
		grpc.ChainUnaryInterceptor(
			s.metrics.grpcMetrics.UnaryServerInterceptor(),
		), opt, keepAliveConfig)
```

Could add [this](https://pkg.go.dev/github.com/grpc-ecosystem/go-grpc-middleware/providers/prometheus@v1.0.1) and it will automatically give us some standard grpc metrics.

disperser/controller/service/config.go
litt3: This is just meant to make the rollout easier. With this config, we can merge the code without worrying about borking things if a release goes out with the feature partially implemented.

disperser/controller/service/server.go
litt3: This is being added via the metrics class:

```
opts = append(opts, s.metrics.GetGRPCServerOption())
```

```
	grpcServerOption := grpc.UnaryInterceptor(
		grpcMetrics.UnaryServerInterceptor(),
	)
```

Or is there something I'm missing here? What's the difference between `ChainUnaryInterceptor` and `UnaryInterceptor`?

disperser/controller/metrics/server_metrics.go
litt3: 8c3154dd6d44bd6d38989bdd961bfe0184be962e

disperser/controller/service/server.go
dmanc: Ah got it. The example I found uses a chain interceptor for some reason but I think if you're only using one inteceptor it's pretty much the same.

api/proto/controller/v1/controller_service.proto
cody-littley: I suggest we disable this rule for now. We don't have a good versioning schema, and until we do I think adding additional version numbers onto things will make things worse, not better.

Going forward, I'm open to the idea of discussing protobuf versioning strategies. But it's definitely something we should carefully consider before putting into practice.

cody-littley: Also, since this particular protobuf will be purely internal, we can always tack on version numbers in the future if we want.

disperser/apiserver/server_v2.go
```diff
@@ -369,3 +409,13 @@ func (s *DispersalServerV2) GetPaymentState(ctx context.Context, req *pb.GetPaym
 	}
 	return reply, nil
 }
+
+// Gracefully shuts down the server and closes any open connections
+func (s *DispersalServerV2) Stop() error {
```
samlaf, line 414: this doesn't seem like its called. do you have intention to wire it at some point?

disperser/apiserver/server_v2.go
```diff
@@ -63,6 +65,16 @@ type DispersalServerV2 struct {
...
+
+	// Determines which payment system to use. If true, use the new payments system running on the controller.
+	// If false, use the legacy payments system which executes on the API server.
+	//
+	// TODO(litt3): this field should be removed once the migration to the new payments system is complete
+	useControllerMediatedPayments bool
+
+	// Controller gRPC client connection and service client
+	controllerConnection *grpc.ClientConn
+	controllerClient     controller.ControllerServiceClient
```
samlaf, line 77: nit: this is more food for thought/discussion than an actual comment.
But overall (might be me..) I don't really like seeing a connection held like this just for closing.
Here's 2 alternate solutions proposed by Claude:
```
  2. Embed the connection in a wrapper struct:
  type ControllerClient struct {
      *controllerpb.ControllerServiceClient
      conn *grpc.ClientConn
  }

  func (c *ControllerClient) Close() error {
      return c.conn.Close()
  }

  3. Use context cancellation for cleanup instead of explicit Close():
  ctx, cancel := context.WithCancel(context.Background())
  defer cancel() // This will clean up the connection
  ```

File: api/hashing/authorize_payment_request_hashing.go (hunk 19 above)
Comment (samlaf) [reply api/hashing/authorize_payment_request_hashing.go:13]: Thinking about it this whole `api/hashing` directory doesn't make a lot of sense to me. Are these different hashes here actually related?
Curious for thoughts, am I the only one that's confused by this? "hashing" doesn't seem to be semantically relevant enough to warrant a high level `api/` subdir.

File: disperser/controller/metrics/server_metrics.go [hunk 23]
```diff
@@ -0,0 +1,96 @@
...
+)
+
+// Encapsulates metrics for the controller GRPC server
+type ServerMetrics struct {
+	logger           logging.Logger
+	grpcServerOption grpc.ServerOption
+
+	// AuthorizePayment metrics
+	authorizePaymentLatency      *prometheus.SummaryVec
+	authorizePaymentAuthFailures *prometheus.CounterVec
```
samlaf, line 26: nit: not fan of "Server" middlefix here because it made me think that this was an actual server exposed on some port.
Would prefer `metrics.NewControllerMetrics`, or putting it in controller namespace and just having controller.NewMetrics() or something

disperser/controller/service/config.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// Port that the gRPC server listens on
+	GrpcPort uint16
+
+	// Maximum size of a gRPC message that the server will accept (in bytes)
+	MaxGRPCMessageSize int
+
+	// Maximum time a connection can be idle before it is closed.
+	MaxIdleConnectionAge time.Duration
+
```
samlaf: nit: might be nice at some point to have a shared grpc server config that all our services inherit. Otherwise easy to forget some best practices on new services we add.

disperser/controller/service/server.go
```diff
@@ -0,0 +1,151 @@
...
+	}()
+
+	if s.paymentAuthorizationHandler == nil {
+		//nolint:wrapcheck
+		return nil, status.Error(codes.FailedPrecondition, "payment authorization handler not configured")
+	}
+
+	requestHash, err := hashing.HashAuthorizePaymentRequest(request)
+	if err != nil {
+		return nil, status.Errorf(codes.Internal, "failed to hash request: %v", err)
```
samlaf: nit: why not use our errors defined in api/errors.go?

api/proto/controller/v1/controller_service.proto
samlaf: I agree, mostly because we haven't followed that convention elsewhere, so it might confuse people more so than it helps. Also I believe its typical for versioning golang libraries for v1 to not be a suffix, and then v2 is. for example `json` vs `json/v2`.

litt3: ed81e100b356e40d0384051fc0c7b31083eb7e11

api/proto/controller/v1/controller_service.proto
litt3: Reading the definitions, it seems to me like it makes sense to use _both_ `UNAUTHENTICATED` and `PERMISSION_DENIED`, depending on the observed error.

Invalid client signature -> `UNAUTHENTICATED`
Valid signature, but invalid payment -> `PERMISSION_DENIED`

Not sure why jwt would need a monopoly on a given code- seems fine to share, and describe the exact mode of failure in the message.

Side note, I think failed jwt authentication would be `UNAUTHENTICATED`, not `PERMISSION_DENIED`

litt3: Another question: I had been planning to list all the potential errors here, and the reasons. But it seems like that might not be standard practice.

Thoughts?

disperser/controller/service/server.go
litt3: 870be19f29181d30c19b84683bae7f36218676b2

Didn't know they existed

api/proto/controller/v1/controller_service.proto
litt3: 5ab06bfd3291e1ecfc6d6b8f76b78ce0bbc5a272
370c365f966a9056c2ff91b7e3c22c3bc42d0d7e

api/proto/controller/v1/controller_service.proto
litt3: As we touched on yesterday, I think it might make more sense to have a `disperser` folder, and make that the backbone of our grouping. The disperser is the only thing that has `internal` services so far.

I'd prefer to kick this can down the road a bit- we can do this reorg as a standalone task, if we get consensus on it

disperser/controller/service/server.go
litt3: The server is being constructed in a way that explicitly permits nil handler

```
var paymentAuthorizationHandler *payments.PaymentAuthorizationHandler
if config.ServerConfig.EnablePaymentAuthentication {
	paymentAuthorizationHandler = payments.NewPaymentAuthorizationHandler()
}
```

Since @cody-littley will be working on a feature in the short term which also needs this server, it's possible that his feature may be done first. In such a case, we may want this server to be operational without payment authentication.

Once the migration to the new payments system is complete, I think it would make sense to force the authorization handler to be non-nil in the constructor

api/hashing/authorize_payment_request_hashing.go
litt3: I placed it in `hashing` because it depends on `hashBlobHeader`, which isn't exported. 

I don't have a super strong opinion here, but I don't think it's bad to have the `hashing` package. We *do* need some central spot for the hashing of common types, like the `blobHeader`. If that spot isn't `hashing`, then it would be something else that essentially fills the same role, right?

disperser/apiserver/server_v2.go
litt3: Agreed, I don't love this.

I originally had gone the wrapper struct route, but it didn't seem great to add an additional layer of indirection just for the sake of holding on to a connection, so I simplified.

For the context cancellation option, is that actually any cleaner? You still have to hold onto an additional object for the sake of cleanup, it's just that you're holding onto a cancel method instead of a connection.

I don't know, it seems like the problem is that the generated code doesn't have a `Close` method, so whatever way we go is non-ideal. 

Would you prefer the wrapper? My preference is to just hold onto the connection, but I don't feel super strongly

disperser/controller/metrics/server_metrics.go
litt3: > it made me think that this was an actual server exposed

To me, it would be called `MetricsServer` if a server was actually being exposed- I interpret `ServerMetrics` to mean "metrics for the server"

The problem is that these *aren't* the metrics for the entire controller: they are *just* for the grpc server. There are other preexisting metrics, in the controller package, so calling this new class `ControllerMetrics` doesn't seem right.

Thoughts? I'm not attached to `ServerMetrics`, if there is another option that works

disperser/controller/service/config.go
litt3: f0fe4297d7c140fd8af1d597e12f02515f63ceb0

Plopped it right into the `common` package you love so much. LMK if there's a different home you'd prefer

disperser/apiserver/server_v2.go
litt3: I think we probably _should_ be cleanly tearing things down. There are other elements of the `DispersalServer` which should be included in the `Stop` method, but I didn't want to expand scope to address these.

I figure it makes sense to implement proper teardown for the new feature, in case we want to use it eventually

api/proto/controller/v1/controller_service.proto
samlaf: > Invalid client signature -> UNAUTHENTICATED
Valid signature, but invalid payment -> PERMISSION_DENIED

This makes sense actually!

> Side note, I think failed jwt authentication would be UNAUTHENTICATED, not PERMISSION_DENIED

You might be right. I think JWT does both authn/authz (authentication==UNAUTHENTICATED and authorization =PERMISSION_DENIED). So might actually be able to return both depending on what exactly failed from the JWT. I don't know enough though.

> Another question: I had been planning to list all the potential errors here, and the reasons. But it seems like that might not be standard practice.

I actually don't know the standard practice here. I would also have assumed listing them here makes sense... but idk

api/proto/controller/v1/controller_service.proto
samlaf: Nice! I see so metering is global rate limiting

api/hashing/authorize_payment_request_hashing.go
samlaf: Yes. My reasoning was that iiuc there are very different reasons for hashing even inside api/hashing. Then other hashes are not here, like blobKey. Just seems a bit ad-hoc? I might have not looked deeply enough to understand its semantics/usecase though.

disperser/apiserver/server_v2.go
samlaf: No I think its fine as is. Don't have a super strong preference. Agree with your reasoning.

> For the context cancellation option, is that actually any cleaner? You still have to hold onto an additional object

Do you though? Isn't the point that the context cancellation is held at the very top and passed in everywhere? This instead of having to call Stop() on every object is my (small, not super strongly held) preference

disperser/controller/metrics/server_metrics.go
samlaf: no that works lol

disperser/controller/service/config.go
samlaf: haha no that works

api/hashing/authorize_payment_request_hashing.go
litt3: @samlaf All the methods in the package are for hashing protobuf objects. Would you find this more tolerable if the `hashing` package were nested beneath the `grpc` directory?

Thinking more on this, it seems the main reason that we need separate file(s) to handle protobuf hashing is... we can't add the hashing methods to the source file like we would do for normal golang

If we want to do this reorganization, let's do it in a later PR, to avoid holding up this one

disperser/apiserver/server_v2.go
litt3: Gotcha, I see your reasoning.

The problem is that you can't pass a cancelable context in when constructing the client connection, because it doesn't accept one. So I think we're stuck with non-ideal solutions here.

api/proto/controller/v1/controller_service.proto
litt3: ec3bf000d6475d8ae3dae990a57347d97d87db26

Left some notes

disperser/apiserver/controller_client.go
samlaf: how about we put it in `api/clients/controller`? Then we call it like `controller.NewClient()` which reads nice?

api/hashing/authorize_payment_request_hashing.go
samlaf: Yeah SGTM. I was wondering why we take the protobuf structs as argument to the hashing, since anyways we don't want to hash the protobuf struct (which contains other stuff). Feels like we should always convert the protobuf to an internal struct before hashing?

disperser/apiserver/controller_client.go
litt3: doh, I forgot I removed the wrapper client entirely 🙃 

Renamed the package to `server`, now that there isn't a client in there 8fdfe37fa049853d1a0a7439758ffa1754b31ba6

api/hashing/authorize_payment_request_hashing.go
cody-littley: > Feels like we should always convert the protobuf to an internal struct before hashing?

IMO, we should not convert protobufs to internal structs except where absolutely necessary. For example, if a protobuf keeps things in a list, but internally we need to access the data in a map.

We have lots of internal struct protobuf representations currently in our code that don't serve any specific purpose, and I think it makes the code harder to read and understand. It makes it especially difficult to make modifications, since each change is amplified and must be applied to multiple locations.

In the case of the protobuf hashing, I'm not sure converting to an internal struct makes things less error prone. It just adds another level of indirection that needs to be deeply documented. The mechanism for hashing a protobuf needs to be defined at a contractual level. That is to say, if there are different implementations in different languages, all parties should agree on the hash.

litt3: > we should not convert protobufs to internal structs except where absolutely necessary

💯

PR #1901, opened 2025-08-20
feat(payments): Implement dynamo cumulative payment store

litt3:
- closes EGDA-1853

### DIFF
--- a/core/payments/ondemand/dynamostore/dynamodb_cumulative_payment_store.go
+++ b/core/payments/ondemand/dynamostore/dynamodb_cumulative_payment_store.go
@@ -175,12 +175,12 @@ func (s *DynamoDBCumulativePaymentStore) AddCumulativePayment(
 // extractPaymentValue extracts and parses attributeCumulativePayment from a DynamoDB item
 func extractPaymentValue(item map[string]types.AttributeValue) (*big.Int, error) {
 	if len(item) == 0 {
-		return big.NewInt(0), nil
+		return nil, fmt.Errorf("empty item attributes")
 	}
 
 	attributeValue, ok := item[attributeCumulativePayment]
 	if !ok {
-		return big.NewInt(0), nil
+		return nil, fmt.Errorf("missing required attribute %s", attributeCumulativePayment)
 	}
 
 	attributeNumber, ok := attributeValue.(*types.AttributeValueMemberN)


core/payments/ondemand/dynamostore/dynamodb_cumulative_payment_store.go
```diff
@@ -0,0 +1,212 @@
...
+	}
+
+	updateItemOutput, err := s.dynamoClient.UpdateItem(ctx, updateItemInput)
+	if err != nil {
+		var conditionCheckFailedException *types.ConditionalCheckFailedException
+		if errors.As(err, &conditionCheckFailedException) {
+			currentValue, getErr := s.getCurrentPayment(ctx)
+			if getErr != nil {
+				return nil, fmt.Errorf("conditional check failed and couldn't get current value: %w", getErr)
+			}
```
cody-littley: Would it make sense to include both errors here?

cody-littley (dismissed):
I'm not familiar with dynamo syntax, but I'm happy with this change if tests are passing.

core/payments/ondemand/dynamostore/dynamodb_cumulative_payment_store.go
litt3: The dynamo error isn't very descriptive. But good point, I should include it for the sake of completeness

litt3: 698f90afc07e605786ea8a054c349097c46ca419

cody-littley (dismissed):

core/payments/ondemand/dynamostore/dynamodb_cumulative_payment_store.go
```diff
@@ -0,0 +1,213 @@
...
+// partition key and stores the CumulativePayment value as a number.
+//
+// This store represents a subset of the logic implemented in [meterer.DynamoDBMeteringStore]. It maintains the same
+// table structure for the sake of backwards compatibility, but otherwise is intended to replace the old class, as
+// part of the ongoing payments refactor.
+type DynamoDBCumulativePaymentStore struct {
+	// The DynamoDB client to use for storage operations
+	dynamoClient *dynamodb.Client
+	// The name of the DynamoDB table to store payments in, stored as *string for use in DynamoDB operations
+	tableName *string
```
samlaf, line 37: why do they use *string? What happens if its nil?

core/payments/ondemand/dynamostore/dynamodb_cumulative_payment_store.go
```diff
@@ -0,0 +1,213 @@
...
+
+	// Calculate the maximum allowed current cumulative payment value.
+	// If the cumulativePayment value that exists prior to the addition is greater than this, that means the account
+	// doesn't have enough funds to pay
+	maxAllowedCurrent := new(big.Int).Sub(maxCumulativePayment, amount)
+
+	// if maxAllowedCurrent is negative, that means the cost of this single dispersal alone exceeds total deposits
+	if maxAllowedCurrent.Sign() < 0 {
+		currentValue, err := s.getCurrentPayment(ctx)
+		if err != nil {
```
samlaf: took me a while to figure out what was happening here. The name of the variable is `maxCumulativePayment` but your comment here assumes its `totalDeposits`. This seems like an assumption that maxCumulativePayment will always be maxTotalDeposits. Worth changing the variable name to make this more obvious?

core/payments/ondemand/dynamostore/dynamodb_cumulative_payment_store.go
```diff
@@ -0,0 +1,213 @@
...
+	}
+
+	// Calculate the maximum allowed current cumulative payment value.
+	// If the cumulativePayment value that exists prior to the addition is greater than this, that means the account
+	// doesn't have enough funds to pay
+	maxAllowedCurrent := new(big.Int).Sub(maxCumulativePayment, amount)
+
+	// if maxAllowedCurrent is negative, that means the cost of this single dispersal alone exceeds total deposits
+	if maxAllowedCurrent.Sign() < 0 {
+		currentValue, err := s.getCurrentPayment(ctx)
```
samlaf: you're assuming no concurrent operation can modify the currentPayment. Is that accurate?

core/payments/ondemand/dynamostore/dynamodb_cumulative_payment_store.go
```diff
@@ -0,0 +1,213 @@
...
+// extractPaymentValue extracts and parses attributeCumulativePayment from a DynamoDB item
+func extractPaymentValue(item map[string]types.AttributeValue) (*big.Int, error) {
+	if len(item) == 0 {
+		return big.NewInt(0), nil
+	}
+
+	attributeValue, ok := item[attributeCumulativePayment]
+	if !ok {
+		return big.NewInt(0), nil
+	}
```
samlaf: should these not be errors? you check that the amount is > 0 so this value should always be set to a non-nil value no?

core/payments/ondemand/dynamostore/dynamodb_cumulative_payment_store.go
```diff
@@ -0,0 +1,213 @@
...
+		ReturnValues:              types.ReturnValueUpdatedNew,
+	}
+
+	updateItemOutput, err := s.dynamoClient.UpdateItem(ctx, updateItemInput)
+	if err != nil {
+		var conditionCheckFailedException *types.ConditionalCheckFailedException
+		if errors.As(err, &conditionCheckFailedException) {
+			currentValue, getErr := s.getCurrentPayment(ctx)
+			if getErr != nil {
+				return nil, fmt.Errorf(
```
samlaf: another assumption that no other concurrent call is modifying the value on dynamo.
Thinking.. could we instead use ReturnValueOld and manually add the updated amount locally? Then we wouldn't need to make a separate get call during error paths.

samlaf (approved):
LGTM outside of some nits/questions

core/payments/ondemand/dynamostore/dynamodb_cumulative_payment_store.go
litt3: ChatGPT says it's just the string strategy *in general*, to differentiate between `nil` and `""`, and that a `nil` tablename would yield an error. Haven't gone deeper than that, though

core/payments/ondemand/dynamostore/dynamodb_cumulative_payment_store.go
litt3: I'm not sure if you meant rename the variable in the error, or here in the code.

I went ahead and changed the name _in the error_, while leaving some explanation there for how `maxCumulativePayment` is determined

a3151567a03ab28bc4af07715f70cbf6665922e5

core/payments/ondemand/dynamostore/dynamodb_cumulative_payment_store.go
litt3: Turns out what we need is to set `ReturnValuesOnConditionCheckFailure`, so that the old item is returned when the condition fails. Claude steered me wrong and said this wasn't possible. Good callout

0306d4d2d4d277a2a3190fb6c0b3b4f5fc29e926

core/payments/ondemand/dynamostore/dynamodb_cumulative_payment_store.go
litt3: No longer being assumed with 0306d4d2d4d277a2a3190fb6c0b3b4f5fc29e926

core/payments/ondemand/dynamostore/dynamodb_cumulative_payment_store.go
litt3: > should these not be errors?

I don't think so. These blocks can be triggered if the account doesn't exist in the table at all, or if it exists, but doesn't have a cumulative payment defined.

For both of these cases, the correct interpretation is that the cumulative payment is `0`, since that's the default for all accounts not being tracked.

core/payments/ondemand/dynamostore/dynamodb_cumulative_payment_store.go
samlaf: I meant the actual name of the variable.
Your comment ends with "exceeds total deposits". Maybe just change it to "exceeds the max cumulative payment" to stay agnostic as to where the max comes from?
Otherwise I was suggesting to use totalDeposits everywhere, just embedding the fact that the max cumulative payment will ALWAYS come from the totalDeposits (I think..?)

Not sure what makes most sence though.

core/payments/ondemand/dynamostore/dynamodb_cumulative_payment_store.go
samlaf: hah nice

core/payments/ondemand/dynamostore/dynamodb_cumulative_payment_store.go
samlaf: image

Maybe I'm missing something, but this is called right after we update the table. that update table should (in all case..?) create the entry in the table, and make it non-zero, no?

core/payments/ondemand/dynamostore/dynamodb_cumulative_payment_store.go
litt3: > Maybe just change it to "exceeds the max cumulative payment" to stay agnostic as to where the max comes from?

👍 Missed that last reference to `deposits`. Fixed now, so this store doesn't refer to deposits at all

9a4b043dc9fc90ee9165c89f0887497edf784c37

core/payments/ondemand/dynamostore/dynamodb_cumulative_payment_store.go
litt3: My reasoning was that the value might not exist in the error case, since when a conditional update fails, the table isn't modified. 

But as currently implemented, the `attribute_not_exists` part of the condition guarantees that a failed condition can *only happen if* the CumulativePayment attribute exists. So I think you're right, this value should be expected to exist when parsing both error and standard cases.

85a9bb31302bb39d294a14ee1945ef010309cc08

samlaf (approved):
LGTM. Maybe add a comment somewhere mentioning the inconsistency risk of this implementation, linking to https://aws.amazon.com/blogs/database/implement-resource-counters-with-amazon-dynamodb/ though.

litt3:
Closing for now, because we've come upon a much better way of doing this, which avoids the painpoints observed by this impl

PR #1902, opened 2025-08-20
fix: update `NewDisperserClient` errors

iquidus:
Closes DAINT-704

Replaces the returned 400 errors with standard errors

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/disperser_client.go
+++ b/api/clients/v2/disperser_client.go
@@ -95,16 +95,16 @@ func NewDisperserClient(
 ) (*disperserClient, error) {
 
 	if config == nil {
-		return nil, api.NewErrorInvalidArg("config must be provided")
+		return nil, fmt.Errorf("config must be provided")
 	}
 	if config.Hostname == "" {
-		return nil, api.NewErrorInvalidArg("hostname must be provided")
+		return nil, fmt.Errorf("hostname must be provided")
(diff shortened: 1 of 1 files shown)


api/clients/v2/disperser_client.go
```diff
@@ -95,16 +95,16 @@ func NewDisperserClient(
...
+		return nil, fmt.Errorf("config must be provided")
 	}
 	if config.Hostname == "" {
-		return nil, api.NewErrorInvalidArg("hostname must be provided")
+		return nil, fmt.Errorf("hostname must be provided")
 	}
 	if config.Port == "" {
-		return nil, api.NewErrorInvalidArg("port must be provided")
+		return nil, fmt.Errorf("port must be provided")
 	}
```
ethenotethan: very nit picky and maybe unnecessary comment but why are we just checking against empty string. wouldn't we also want detection when a multi character empty string is provided; ie:
```
if strings.TrimSpace(val)
```

iquidus: Done: [a9d74c2](https://github.com/Layr-Labs/eigenda/pull/1902/commits/a9d74c2394b1a8fb9366489947b229dabef1c97b)

PR #1903, opened 2025-08-20
feat!: validator state cache

cody-littley:

Validators cache operator state.

In order to simplify setup logic, the contract directory is now a mandatory configuration field. This is a breaking change, and will need to be clearly communicated with node operators.

### DIFF
--- a/common/enforce/assertions.go
+++ b/common/enforce/assertions.go
@@ -1,8 +1,12 @@
 package enforce
 
-import "fmt"
+import (
+	"fmt"
 
-// If convenient, it's ok at add additional assertions to this collection, as long as those assertions are
+	"golang.org/x/exp/constraints"
+)
+
+// If convenient, it's ok to add additional assertions to this collection, as long as those assertions are
 // general purpose and not specific to a particular domain or use case. For example, don't import custom
 // types or packages that are not part of the standard library or common Go ecosystem.
 
@@ -13,7 +17,7 @@ func True(condition bool, message string, args ...any) {
 	}
 }
 
-// Asserts a condition is false and panics with an error message if the condition is false.
+// Asserts a condition is false and panics with an error message if the condition is true.
 func False(condition bool, message string, args ...any) {
 	if condition {
 		panic("Expected condition to be false: " + fmt.Sprintf(message, args...))
@@ -28,10 +32,48 @@ func Equals[T comparable](expected T, actual T, message string, args ...any) {
 }
 
 // Asserts that two values are not equal and panics with an error if they are equal.
+//
+// May not behave as expected for NaN values in floating point comparisons.
 func NotEquals[T comparable](notExpected T, actual T, message string, args ...any) {
 	if notExpected == actual {
 		panic(fmt.Sprintf("Expected inequality, %v == %v: %s", notExpected, actual,
 			fmt.Sprintf(message, args...)))
+	}
+}
+
+// Asserts a > b
+//
+// May not behave as expected for NaN values in floating point comparisons.
+func GreaterThan[T constraints.Ordered](a T, b T, message string, args ...any) {
+	if a <= b {
+		panic(fmt.Sprintf("Expected %v > %v: %s", a, b, fmt.Sprintf(message, args...)))
+	}
+}
+
+// Asserts a >= b
+//
+// May not behave as expected for NaN values in floating point comparisons.
+func GreaterThanOrEqual[T constraints.Ordered](a T, b T, message string, args ...any) {
+	if a < b {
+		panic(fmt.Sprintf("Expected %v >= %v: %s", a, b, fmt.Sprintf(message, args...)))
+	}
+}
+
+// Asserts a < b
+//
+// May not behave as expected for NaN values in floating point comparisons.
+func LessThan[T constraints.Ordered](a T, b T, message string, args ...any) {
+	if a >= b {
+		panic(fmt.Sprintf("Expected %v < %v: %s", a, b, fmt.Sprintf(message, args...)))
+	}
+}
+
+// Asserts a <= b
+//
+// May not behave as expected for NaN values in floating point comparisons.
+func LessThanOrEqual[T constraints.Ordered](a T, b T, message string, args ...any) {
+	if a > b {
+		panic(fmt.Sprintf("Expected %v <= %v: %s", a, b, fmt.Sprintf(message, args...)))
 	}
 }
 
--- a/common/enforce/assertions_test.go
+++ b/common/enforce/assertions_test.go
@@ -1,136 +1,147 @@
 package enforce
 
-import "testing"
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
 
 func TestTrue(t *testing.T) {
 	True(true, "This should not panic")
-	defer func() {
-		if r := recover(); r == nil {
-			t.Errorf("Expected panic, but did not panic")
-		}
-	}()
 
-	True(false, "This should panic, and the deferred function will catch it")
+	require.Panics(t, func() {
+		True(false, "This should panic")
+	})
 }
 
 func TestFalse(t *testing.T) {
 	False(false, "This should not panic")
-	defer func() {
-		if r := recover(); r == nil {
-			t.Errorf("Expected panic, but did not panic")
-		}
-	}()
 
-	False(true, "This should panic, and the deferred function will catch it")
+	require.Panics(t, func() {
+		False(true, "This should panic")
+	})
 }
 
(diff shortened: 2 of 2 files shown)


litt3 (approved):

node/cmd/main.go
```diff
@@ -84,14 +86,29 @@ func NodeMain(ctx *cli.Context) error {
...
+		gethcommon.HexToAddress(config.EigenDADirectory))
 	if err != nil {
-		return fmt.Errorf("cannot create eth.Reader: %w", err)
+		return fmt.Errorf("failed to create contract directory: %w", err)
+	}
+
+	operatorStateRetrieverAddress, err :=
+		contractDirectory.GetContractAddress(context.Background(), directory.OperatorStateRetriever)
+	if err != nil {
+		return fmt.Errorf("failed to get BLSOperatorStateRetriever address: %w", err)
```
litt3: `BLS` prefix to `BLSOperatorStateRetriever` is outdated

node/node.go
```diff
@@ -134,8 +139,24 @@ func NewNode(
 		return nil, fmt.Errorf("failed to get chainID: %w", err)
 	}
 
+	serviceManagerAddress, err := contractDirectory.GetContractAddress(ctx, directory.ServiceManager)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get EigenDAEjectionManager address from contract directory: %w", err)
```
litt3: error message is wrong, mentioning `EigenDAEjectionManager`

disperser/controller/metadata/batch_metadata_manager.go
```diff
@@ -150,7 +124,7 @@ func (m *batchMetadataManager) updateMetadata() error {
 		return nil
```
litt3: Your infof log above is missing the RBN argument

disperser/controller/metadata/batch_metadata_manager.go
```diff
@@ -150,7 +124,7 @@ func (m *batchMetadataManager) updateMetadata() error {
...
 	}
```
litt3, line 133: Question about this `if` block:

Why are you checking `referenceBlockNumber == previousMetadata.referenceBlockNumber` instead of `referenceBlockNumber <= previousMetadata.referenceBlockNumber`?

core/eth/operatorstate/operator_state_cache.go
```diff
@@ -0,0 +1,144 @@
...
+		indexLock:     common.NewIndexLock(indexLockSize),
+	}, nil
+}
+
+// GetOperatorState retrieves the operator state for a given reference block number and quorums.
+func (c *operatorStateCache) GetOperatorState(
+	ctx context.Context,
+	referenceBlockNumber uint64,
+	quorums []core.QuorumID,
+) (*core.OperatorState, error) {
```
litt3, line 82: So, if callers mutate the returned struct, it would mess up the values in the cache, right? Probably worth mentioning

node/node.go
cody-littley: fixed

node/cmd/main.go
cody-littley: fixed

disperser/controller/metadata/batch_metadata_manager.go
cody-littley: oops, updated

disperser/controller/metadata/batch_metadata_manager.go
cody-littley: The `ReferenceBlockProvider` implements this check. But in order to make the code easier to read, I added the following:

```
		// reference block provider prevents RBN from going backwards
		enforce.GreaterThanOrEqual(referenceBlockNumber, previousMetadata.referenceBlockNumber,
			"reference block number went backwards")
```

I had to copy the enforce assertions into this PR, since the PR that contains them has not yet merged.

core/eth/operatorstate/operator_state_cache.go
cody-littley:
```
// WARNING: do not modify the returned OperatorState or any of its contents, as this will corrupt the cached data.
```

Proper way to handle this would be to make the struct fields private and add getters. But this would change this into a 3k line PR.

common/enforce/assertions.go
```diff
@@ -0,0 +1,117 @@
+package enforce
+
+import (
+	"fmt"
+
+	"golang.org/x/exp/constraints"
+)
+
+// If convenient, it's ok at add additional assertions to this collection, as long as those assertions are
```
litt3: ...it's ok ~~at~~ _to_ add...

common/enforce/assertions.go
```diff
@@ -0,0 +1,117 @@
...
+// types or packages that are not part of the standard library or common Go ecosystem.
+
+// Asserts a condition is true and panics with a message if the condition is false.
+func True(condition bool, message string, args ...any) {
+	if !condition {
+		panic("Expected condition to be true: " + fmt.Sprintf(message, args...))
+	}
+}
+
+// Asserts a condition is false and panics with an error message if the condition is false.
```
litt3: if the condition is ~~false~~ true

common/enforce/assertions.go
```diff
@@ -0,0 +1,117 @@
...
+// Asserts that two values are not equal and panics with an error if they are equal.
+func NotEquals[T comparable](notExpected T, actual T, message string, args ...any) {
+	if notExpected == actual {
+		panic(fmt.Sprintf("Expected inequality, %v == %v: %s", notExpected, actual,
+			fmt.Sprintf(message, args...)))
+	}
+}
+
+// Asserts a > b
+func GreaterThan[T constraints.Ordered](a T, b T, message string, args ...any) {
```
litt3, line 47: gpt5 claims that this will incorrectly pass if either number is `NaN`, likely also the case for other similar methods

common/enforce/assertions_test.go
```diff
@@ -0,0 +1,182 @@
...
+
+func TestTrue(t *testing.T) {
+	True(true, "This should not panic")
+	defer func() {
+		if r := recover(); r == nil {
+			t.Errorf("Expected panic, but did not panic")
+		}
+	}()
+
+	True(false, "This should panic, and the deferred function will catch it")
```
litt3: any reason not to use `require.Panics` for these tests? Seems easier than the home rolled `defer` logic

common/enforce/assertions_test.go
```diff
@@ -0,0 +1,182 @@
...
+	defer func() {
+		if r := recover(); r == nil {
+			t.Errorf("Expected panic, but did not panic")
+		}
+	}()
+
+	NotEquals(1, 1, "This should panic, and the deferred function will catch it")
+}
+
+func TestGreaterThan(t *testing.T) {
```
litt3, line 41: consider adding a test case for ==, which should panic for `GreaterThan`

common/enforce/assertions_test.go
```diff
@@ -0,0 +1,182 @@
...
+	defer func() {
+		if r := recover(); r == nil {
+			t.Errorf("Expected panic, but did not panic")
+		}
+	}()
+
+	GreaterThanOrEqual(1, 2, "This should panic, and the deferred function will catch it")
+}
+
+func TestLessThan(t *testing.T) {
```
litt3, line 61: consider adding == test case

litt3 (approved):

common/enforce/assertions.go
cody-littley: fixed

common/enforce/assertions.go
cody-littley: fixed

common/enforce/assertions.go
cody-littley: As discussed offline, NaN is a dumb concept, and we should pretend it doesn't exist for this generic method. We can possibly create float-aware versions of this in the future that handle NaN, but it's premature do do so now. I have documented this behavior in the godoc though.

cody-littley:
```
//
// May not behave as expected for NaN values in floating point comparisons.
```

common/enforce/assertions_test.go
cody-littley: Did not know that existed. This is much cleaner.

common/enforce/assertions_test.go
cody-littley: added

common/enforce/assertions_test.go
cody-littley: added

litt3 (approved):

PR #1904, opened 2025-08-20
fix(payments): Simplify `ReservationLedger` debit process to a single method

litt3:
- @samlaf requested this change originally, and I pushed back because I wasn't certain
- He was right, so making this change now

### DIFF
--- a/core/payments/reservation/reservation_ledger.go
+++ b/core/payments/reservation/reservation_ledger.go
@@ -53,6 +53,8 @@ func NewReservationLedger(
 // Returns (true, nil) if the reservation has enough capacity to perform the debit.
 // Returns (false, nil) if the bucket lacks capacity to permit the fill.
 // Returns (false, error) if an error occurs. Possible errors include:
+//   - [QuorumNotPermittedError]: one or more of the requested quorums are not permitted by the reservation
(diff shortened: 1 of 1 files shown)


cody-littley (approved):

core/payments/reservation/reservation_ledger.go
```diff
@@ -64,9 +66,24 @@ func (rl *ReservationLedger) Debit(
 	// This a local time from the perspective of the entity that owns this ledger instance, to be used with the local
 	// leaky bucket: it should NOT be sourced from the PaymentHeader
 	now time.Time,
+	// the timestamp included, or planned to be included, in the PaymentHeader
+	dispersalTime time.Time,
```
samlaf, line 70: you mentioned when I first asked about this that the change didn't necessarily make sense because it was called before the PaymentHeader is created or something. Guess what changed is that you realized "planned to be included" solves the issue?

samlaf (approved):

core/payments/reservation/reservation_ledger.go
litt3: Right. It's a bit clunky like this, but not *too* bad, and makes other things simpler.

PR #1906, opened 2025-08-20
chore: better contract bindings makefile command

samlaf:
Fixes the issue where old bindings were not getting deleted. 
This deletes old bindings for MockRollup and EigenDAStateRetriever.

Also separate our previous `compile` make target which was both compiling the contracts AND generating bindings into two separate commands. IMO compiling contracts shouldn't have to wait for binding generation if it doesn't care for that. We will have to make sure to use `make bindings` instead now when we need bindings, such as if running inabox with local contract changes.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


cody-littley (approved):

contracts/generate-bindings.sh
```diff
@@ -23,9 +24,6 @@ function create_binding_abi_only {
   abigen --abi=data/tmp.abi --pkg=contract${contract} --out=$binding_dir/${contract}/binding.go
```
ethenotethan, line 24: should we add a ticket or potentially consider migrating to abigen v2 at some point? 
- https://github.com/ethereum/go-ethereum/pull/31379 
- http://geth.ethereum.org/docs/developers/dapp-developer/native-bindings-v2

contracts/README.md
```diff
@@ -15,7 +15,7 @@ forge install
 ### Compile
 To compile contracts and generate golang ABI bindings, run the following:
 ```
-make compile-contracts
+make bindings
```
ethenotethan: nit - could be useful to have structured like:
```
### Generate Bindings
...
### Compile
```

Currently the compile section refers to both bindings and compilation

contracts/Makefile
```diff
@@ -2,8 +2,12 @@ deps:
 	yarn
 
 compile: deps
-	# compile.sh compiles the contracts and creates go bindings
-	./compile.sh
+	forge build
+
+# make bindings compiles the contracts and creates go bindings
+bindings: compile
```
ethenotethan, line 8: nit - is the compile step cached at all or is it re-ran every-time `make bindings` is called?

contracts/Makefile
```diff
@@ -2,8 +2,12 @@ deps:
...
+	rm -rf ./bindings
```
ethenotethan, line 9: assuming this is to ensure that stale or de-scoped contracts don't dangle?

contracts/generate-bindings.sh
samlaf: yes please do so. would be a nice to have but I think it will be non-trivial change to apply so not really a priority for me atm.

contracts/Makefile
samlaf: the caching is done by forge. so it gets re-run by make but its instantaneous if already cached by forge.
We were previously always deleting the cache folder (by running `forge clean`) but I don't like this. Sometimes there'll be cache inconsistencies, but I think that should be on devs to figure out and clean their cache if/when needed. Too annoying to wait on long builds all the time imo.

contracts/Makefile
samlaf: yes exactly

contracts/generate-bindings.sh
samlaf: when doing so we'll be able to more easily log all eth-call arguments as well, which will be a nice debugging benefit

contracts/README.md
samlaf: [4374841](https://github.com/Layr-Labs/eigenda/pull/1906/commits/4374841877fe3053b93209f7ed7a83955f3129ba)

ethenotethan (approved):
LGTM

litt3 (approved):

pakim249CAL (approved):

PR #1908, opened 2025-08-21
feat: hoodi deployment

pakim249CAL:

The EigenDA directory and access control contract were not deployed previously as part of EigenDA, but were added on in the periphery. This PR changes the deployment script to deploy the address directory in tandem with EigenDA, and also adds the configurations for hoodi deployments that are used for preprod and testnet.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


contracts/script/deploy/eigenda/DeployEigenDA.s.sol
```diff
@@ -82,17 +74,32 @@ contract DeployEigenDA is Script {
         vm.startBroadcast();
 
         // DEPLOY PROXY ADMIN
-        deployed[PROXY_ADMIN] = address(new ProxyAdmin());
-        impl[PROXY_ADMIN] = deployed[PROXY_ADMIN];
-        upgraded[PROXY_ADMIN] = true;
+        proxyAdmin = address(new ProxyAdmin());
```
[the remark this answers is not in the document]
pakim249CAL: No, because the proxy admin lives in the proxy administrative layer which the eigenDA contracts should mostly be oblivious to.

contracts/script/deploy/eigenda/DeployEigenDA.s.sol
```diff
@@ -82,17 +74,32 @@ contract DeployEigenDA is Script {
...
+            address(
+                new TransparentUpgradeableProxy(
+                    address(new EigenDADirectory()),
+                    proxyAdmin,
+                    abi.encodeCall(EigenDADirectory.initialize, (address(new EigenDAAccessControl(msg.sender))))
+                )
+            )
+        );
+
+        console2.log("DIRECTORY:", address(directory));
```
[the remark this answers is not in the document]
pakim249CAL: It is not debug code, but it is the only contract that needs to be known to find all other contracts. So it is appropriate to log it.

pakim249CAL:
> Is this PR active?

Thanks for reminding me. I should update and merge this since it was used to deploy hoodi.

contracts/src/core/libraries/v3/address-directory/AddressDirectoryConstants.sol
```diff
@@ -2,6 +2,10 @@
 pragma solidity ^0.8.9;
 
 library AddressDirectoryConstants {
+    /// PROXY ADMIN
+
+    string internal constant PROXY_ADMIN_NAME = "PROXY_ADMIN";
```
samlaf, line 7: interesting. this wasn't in the directory before?

contracts/script/deploy/eigenda/mainnet.beta.config.toml
```diff
@@ -88,4 +88,10 @@ batchConfirmers = []
...
 adversaryThreshold = 33
-quorumNumbersRequired = [0, 1]
\ No newline at end of file
+quorumNumbersRequired = [0, 1]
+
+[initParams.eigenDA.disperser]
+dispersers = []
+
+[initParams.eigenDA.relay]
+relays = []
```
samlaf, line 97: is this empty because the dispersers/relays are already registered in our registries?

samlaf (approved):
Haven't checked the new deployment logic in detail but overall LGTM.
High-level question: what is the goal of these config files? Are we purely keeping them around for public visibility? Because they are only "initial" values and the ones onchain could be completely different right?

And for example mainnet config is missing the relay/disperser info, as I'm guessing those were already set outside of this script.

Is this DeployEigenDA.s.sol of any use going forward basically? When should it be used? Only for deploying to new chains like we did on Hoodi? Could/should we be using it in inabox?

pakim249CAL:
> High-level question: what is the goal of these config files? Are we purely keeping them around for public visibility? Because they are only "initial" values and the ones onchain could be completely different right?

Yes, they exist for public visibility so that it is known how I deployed the contracts. They may not necessarily reflect the state of the chain, but these configs should be kept somewhere. And I don't know of a better spot than here, regardless of how complete they are.

> And for example mainnet config is missing the relay/disperser info, as I'm guessing those were already set outside of this script.

Yes, they were set outside the script.

> Is this DeployEigenDA.s.sol of any use going forward basically? When should it be used? Only for deploying to new chains like we did on Hoodi? Could/should we be using it in inabox?

This is a separate deployment script used in CI that needs to be consolidated at some point. I think I will have to discuss with @dmanc some point how we can migrate over to the new deployment script. All the DeployOpenEigenLayer, EigenDADeployer, SetupEigenDA, etc should probably be consolidated with the new deployment script at some point. The current situation is not ideal, admittedly because we maintain two separate deploy scripts.

contracts/src/core/libraries/v3/address-directory/AddressDirectoryConstants.sol
pakim249CAL: No, and it is still debatable on whether it should be included or not, because implementation contracts are normally written to be completely oblivious to a proxy admin.

contracts/script/deploy/eigenda/mainnet.beta.config.toml
pakim249CAL: No, it was just registered separately after the deployment to save some time.

PR #1911, opened 2025-08-22
feat(clients-v2): Add abigen V2 to checkDACert call

ethenotethan:
Initially this PR was to introduce an "out-of-gas" check. after some more analysis doing this check in an optimal way will require further research and path finding. In the interim I've just updated the `checkDACert` call to leverage the ABI V2 binding for packing argument data and unpacking it

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/verification/cert_verifier.go
+++ b/api/clients/v2/verification/cert_verifier.go
@@ -57,22 +57,8 @@ func (cv *CertVerifier) CheckDACert(
 	ctx context.Context,
 	cert coretypes.EigenDACert,
 ) error {
-	// 1 - switch on the certificate type to determine which contract to call
-	var certV3 *coretypes.EigenDACertV3
-	var err error
-	switch cert := cert.(type) {
-	case *coretypes.EigenDACertV3:
-		certV3 = cert
-	case *coretypes.EigenDACertV2:
-		// EigenDACertV3 is the only version that is supported by the CheckDACert function
-		// but the V2 cert is a simple permutation of the V3 cert fields, so we convert it.
-		certV3 = cert.ToV3()
-	default:
-		// If golang had enums the world would be a better place.
-		panic(fmt.Sprintf("unsupported cert version: %T. All cert versions that we can "+
-			"construct offchain should have a CertVerifier contract which we can call to "+
-			"verify the certificate", cert))
-	}
+	// 1 - Normalize cert to V3
+	certV3 := NormalizeCertV3(cert)
 
 	// 2 - Call the contract method CheckDACert to verify the certificate
 	// TODO: Determine adequate future proofing strategy for EigenDACertVerifierRouter to be compliant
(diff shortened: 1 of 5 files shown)


api/clients/v2/verification/errors.go
```diff
@@ -45,3 +48,33 @@ type CertVerifierInvalidCertError struct {
...
+
+// containsOOGMessage returns true if the error message includes a known
+// go-ethereum or client variant string indicating insufficient gas.
+// Some messages originate outside strict EVM execution, but are included
+// for coverage of custom RPC provider behaviors.
+func containsOOGMessage(msg string) bool {
+	return strings.Contains(msg, "out of gas") ||
+		strings.Contains(msg, "gas required exceeds allowance") ||
+		strings.Contains(msg, "intrinsic gas too low")
+}
```
samlaf: were these LLM generated? If not do you know where they come from? Might be good to document which clients send which strings. Also with the multitude of eth execution clients (ethrex being a new one for example), are we sure this is going to work..?

api/clients/v2/verification/errors.go
```diff
@@ -45,3 +48,33 @@ type CertVerifierInvalidCertError struct {
...
+// returns a standard error message and does not malform the EVM return data in transit.
+func IsEVMOutOfGasError(err error) bool {
+	if err == nil {
+		return false
+	}
+
+	// Try to catch go-ethereum's RPC error (might be wrapped or modified based on provider)
+	var rerr rpc.Error
+	if errors.As(err, &rerr) {
+		msg := strings.ToLower(rerr.Error())
```
samlaf: wait this only works with go-ethereum's RPC? Then this is probably not an acceptable solution?

api/clients/v2/verification/cert_verifier.go
```diff
@@ -86,14 +94,35 @@ func (cv *CertVerifier) CheckDACert(
...
+		To:   &certVerifierAddr,
+		Gas:  MaxGasPerCheckDACert,
+		Data: callMsgBytes,
+	}, nil)
+
+	if err != nil && IsEVMOutOfGasError(err) {
+		panic(fmt.Errorf("eth_call to checkDACert ran out of gas. "+
+			"This should not occur with a 2/3 block gas limit allocation, "+
+			"indicating a malformed certificate, potential exploit, or RPC issue: %w", err))
+	}
```
samlaf: don't panic this will return a 500 to user in proxy. We prob want to return a structured error to decide what happens with this (prob drop using a derivation pipeline error?)

api/clients/v2/verification/cert_verifier.go
```diff
@@ -86,14 +94,35 @@ func (cv *CertVerifier) CheckDACert(
...
-	)
+
+	callMsgBytes, err := cv.v2VerifierBinding.TryPackCheckDACert(certBytes)
+	if err != nil {
+		return &CertVerifierInternalError{Msg: "pack checkDACert call", Err: err}
+	}
+
+	// Perform an eth_call with a gas limit of 20M or 2/3 of the current block gas limit.
+	// On-chain (e.g. in a rollup prover tx), budget additional gas since checkDACert adds
+	// to the overall execution path. The 2/3 factor is a heuristic and may be tuned further.
```
samlaf: wouldnt mention 2/3 here, but instead above the constant. 2/3 might become outdated very soon given clients are raising their max gas limit no?

samlaf (commented):
This is super nice! abigen v2 ftw
I would prob be fine with merging this if we make the MaxGasPerCheckDACert constant a variable instead, which can be configured as a proxy flag.

api/clients/v2/verification/errors.go
ethenotethan: what would an alternative solution look like? for clarification `go-ethereum` might not be the best term here but there is a fallback conditional check with asserts for substrings defined in node execution envs

api/clients/v2/verification/cert_verifier.go
ethenotethan: if we drop a cert based on this criteria then how would we prove this criteria within a fraud proof tx?

api/clients/v2/verification/cert_verifier.go
ethenotethan: yeah good callout - we should be adaptive to gas limit changes.  a bit curious about how we could extract this value though.. 

one way is potentially calling [eth_getBlockByNumber](https://www.quicknode.com/docs/ethereum/eth_getBlockByNumber) and extracting the `gasLimit` from the block. I don't really like this though cuz there's an edge case where a prover tx can execute up-to 2 weeks in-which an Ethereum upgrade can happen to raise the limit

api/clients/v2/verification/errors.go
ethenotethan: at the end of day its a heuristic - only way we could know for sure is if we re-execute the EVM within the client (just trust on RPC for prestate). I grabbed these values from go-ethereum directly and was gonna do further analysis which is why this PR is in a draft not fully ready for review state.

ethenotethan: generally speaking though the "out of gas" string is returned JSON RPC for every node provider iiuc

samlaf: Looks like there's an actual error code in the official schema? Not sure if its implemented by all eth-rpcs though... but probably? https://github.com/ethereum/execution-apis/blob/f91fd7b105b062fd49a3cc23063f5c7becff3afe/src/eth/execute.yaml#L137

api/clients/v2/verification/errors.go
samlaf: yeah I think go-ethereum is probably wrong?
My guess is this error would be returned by the go-ethereum client, which should be agnostic to the actual client that ran the rpc.
Their client doesn't implement structured errors though? Feels like it should given that there's an official schema for them: https://github.com/Layr-Labs/eigenda/pull/1911#discussion_r2301863420
Maybe we need to update our geth version?

api/clients/v2/verification/cert_verifier.go
samlaf: good question...

api/clients/v2/verification/cert_verifier.go
samlaf: I meant that writing 2/3 is fine, but it should live above near the constant, not in this place.
Agree we prob don't want to make extra rpc calls, and keep it simple.

samlaf (commented):
As I suggested during sprint planning, I'm very down to merge this for the benefits of using abigenv2.
Suggest we set gasLimit to be infinity by default, though make it configurable.
Then we can decide to turn on when we're ready.

api/clients/v2/verification/errors.go
ethenotethan: hmm that's defined for [eth_simulate](https://www.quicknode.com/docs/ethereum/eth_simulateV1) though - wonder if there's any benefits to using simulate over eth_call given it also does gas tracking and is better suited as as a full transaction simulation. I remember reading about this RPC method over a year ago when it was a proposal and presented an existential threat to the now failed startup I was at lol. Anyway happy to see if its at our disposal if we wanna leverage it

samlaf (approved):
:shipit:

PR #1913, opened 2025-08-22
refactor: eigendacert interface

samlaf:
We previously had 2 separate EigenDACert and RetrievableEigenDACert interface. I'm still not 100% sure what's the best way to do this but this became very clunky to use when trying to refactor our payload retriever clients to also do verification, since cert verifier takes an EigenDACert instead of a RetrievableEigenDACert.

The main thing I was trying to achieve is an ADT, so I merged both interfaces, which added the private method to our EigenDACert interface. This way no one else can implement this interface, and we are free to break the API in the future if we ever need to change our cert representation or the way this interface is handled/represented.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/clients/v2/verification/contract_status_codes.go
```diff
@@ -0,0 +1,42 @@
+package verification
+
+// VerificationStatusCode represents the status codes that are returned by
+// EigenDACertVerifier.checkDACert contract calls. The enum values below should match exactly
+// the status codes defined in the contract:
+// https://github.com/Layr-Labs/eigenda/blob/1091f460ba762b84019389cbb82d9b04bb2c2bdb/contracts/src/integrations/cert/libraries/EigenDACertVerificationLib.sol#L48-L54
+type VerificationStatusCode uint8
```
ethenotethan: nit - is verification status code the best name? curious if e.g `CheckDACertStatusCode` is better suited since it encapsulates the specificity of this enum (i.e, these status codes are mapped to only a contract view call return value and not e.g a large set verification module)

ethenotethan (dismissed):
LGTM

litt3 (approved):

api/clients/v2/verification/contract_status_codes.go
samlaf: [9682b6e](https://github.com/Layr-Labs/eigenda/pull/1913/commits/9682b6ef37fc0dcf3b2075469ad5d1c13b44f737)

litt3 (approved):

PR #1914, opened 2025-08-23
refactor(proxy): separate eigenda_mgr getEigenDAV1/V2

samlaf:
The previous coupling made it very hard to move the v2 functionality into a higher-level verifyingPayloadRetriever client that verifies the cert, the rbn recency check, and proceeds to retrieve from all configured backends.

Writing this first refactor PR to clearly separate v1 from v2 behavior in proxy, in order to enable this next refactor. This duplicates the secondary storage logic in both getEigenDAV1 and getEigenDAV2, but that both:
1. made me find some bug with the verify fct for v2 (see inline comments)
2. will soon be refactored for the v2 path into a higher-level client anyways

Also this refactor is another instance of one of my favorite programming mottos: https://matklad.github.io/2023/11/15/push-ifs-up-and-fors-down.html

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proxy/store/eigenda_manager.go
```diff
@@ -91,18 +92,115 @@ func (m *EigenDAManager) Get(ctx context.Context,
...
+		return m.getEigenDAV2(ctx, versionedCert, opts)
+	default:
+		return nil, fmt.Errorf("cert version unknown: %b", versionedCert.Version)
+	}
+}
+
+// getEigenDAV1 will attempt to retrieve a blob for the given versionedCert
+// from cache, EigenDA V1, and fallback storage.
+// TODO: we should also add the v1 RetrievalClient to retrieve from the validators directly
+// in case the v1 disperser is down, the same way we do for v2.
```
samlaf, line 114: important note. cc @ian-shim because we just talked about this after the cloudflare outage.

api/proxy/store/eigenda_manager.go
```diff
@@ -91,18 +92,115 @@ func (m *EigenDAManager) Get(ctx context.Context,
...
-	if versionedCert.Version == certs.V1VersionByte && m.eigendaV2 == nil {
-		return nil, errors.New("expected EigenDA V2 backend for DA commitment type with CertV1")
+
+	verifyFnForSecondary := func(ctx context.Context, cert []byte, payload []byte, l1InclusionBlockNumber uint64) error {
+		// We don't add the cert version because EigenDA V1 only supports [certs.V0VersionByte] Certs.
+		// We also don't use the l1InclusionBlockNumber because Recency check is only supported by EigenDA V2.
+		// TODO: we should decouple the Verify function into a VerifyCert and VerifyBlob function,
+		// and only verify the cert once here, before retrievals, and then only verify the blob commitment
+		// after retrievals.
+		return m.eigenda.Verify(ctx, cert, payload)
```
samlaf, line 125: not high priority but the way its currently done is a bit confusing and inefficient (we verify the cert possibly multiple times)

api/proxy/store/eigenda_manager.go
```diff
@@ -91,18 +92,115 @@ func (m *EigenDAManager) Get(ctx context.Context,
...
-	verifyMethod, err := m.getVerifyMethod(versionedCert.Version)
+	var readErrors []error
+	// 1 - read payload from cache if enabled
+	// Secondary storages (cache and fallback) store payloads instead of blobs.
+	// For simplicity, we bypass secondary storages when requesting encoded payloads,
+	// since those requests are only for secure integrations and run by provers/challengers.
+	// TODO: would be nice to store blobs instead of payloads in secondary storages, such that we could standardize all
+	// storages and make them all implement the [clients.PayloadRetriever] interface.
+	// We could then get rid of the proxy notion of caches/fallbacks and only have storages.
+	if m.secondary.CachingEnabled() && !opts.ReturnEncodedPayload {
```
samlaf: have not figured out how to do this yet but will think about it more in the subsequent PR that will try to abstract all storages into a single unified interface.

api/proxy/store/eigenda_manager.go
```diff
@@ -91,18 +92,115 @@ func (m *EigenDAManager) Get(ctx context.Context,
...
 	if err != nil {
-		return nil, fmt.Errorf("get verify method: %w", err)
+		return nil, fmt.Errorf("verify EigenDACert: %w", err)
+	}
+
+	verifyFnForSecondary := func(ctx context.Context, cert []byte, payload []byte, l1InclusionBlockNumber uint64) error {
+		// This was previously using the VerifyCert function, which is pointless because it is now verified above,
+		// and the cert only needs to be verified once.
+		// TODO: implement a verify blob function, the same way it is implemented in [payloadretrieval.RelayPayloadRetriever]
+		return nil
```
samlaf, line 192: this was an issue that already existed in our implementation. I could implement it as part of this PR though if you guys think its necessary.

api/proxy/store/eigenda_manager.go
```diff
@@ -91,18 +92,115 @@ func (m *EigenDAManager) Get(ctx context.Context,
...
 	}
 
-	// The eigendav2 verify function verifies the cert, but not the kzg commitment.
-	// TODO: should we be verifying those when retrieving from secondary storages like S3...?
-	verifyMethod, err := m.getVerifyMethod(versionedCert.Version)
+	var readErrors []error
+	// 1 - read payload from cache if enabled
+	// Secondary storages (cache and fallback) store payloads instead of blobs.
+	// For simplicity, we bypass secondary storages when requesting encoded payloads,
+	// since those requests are only for secure integrations and run by provers/challengers.
```
litt3: > 	// For simplicity, we bypass secondary storages when requesting encoded payloads,
>	// since those requests are only for secure integrations and run by provers/challengers.

This isn't relevant for v1, right? You early-out above if `ReturnEncodedPayload` is set

api/proxy/store/eigenda_manager.go
```diff
@@ -91,18 +92,115 @@ func (m *EigenDAManager) Get(ctx context.Context,
...
+	payload, err := m.eigenda.Get(ctx, versionedCert.SerializedCert)
+	if err == nil {
+		err = m.eigenda.Verify(ctx, versionedCert.SerializedCert, payload)
+		if err != nil {
+			return nil, fmt.Errorf("verify EigenDA V1 cert: %w", err)
+		}
+		// Only backup to secondary storage if we're returning the decoded payload
+		// since the secondary stores are currently hardcoded to store payloads only.
+		// TODO: we could consider also storing encoded payloads under separate keys?
+		if m.secondary.WriteOnCacheMissEnabled() && !opts.ReturnEncodedPayload {
```
litt3: throughout this method, you're doing checks of `ReturnEncodedPayload`, even though you explicitly error at the start of the method if this is true.

This seems like a simplification that was only partially executed?

api/proxy/store/eigenda_manager.go
```diff
@@ -91,18 +92,115 @@ func (m *EigenDAManager) Get(ctx context.Context,
...
+	// Secondary storages (cache and fallback) store payloads instead of blobs.
+	// For simplicity, we bypass secondary storages when requesting encoded payloads,
+	// since those requests are only for secure integrations and run by provers/challengers.
+	// TODO: would be nice to store blobs instead of payloads in secondary storages, such that we could standardize all
+	// storages and make them all implement the [clients.PayloadRetriever] interface.
+	// We could then get rid of the proxy notion of caches/fallbacks and only have storages.
+	if m.secondary.CachingEnabled() && !opts.ReturnEncodedPayload {
+		m.log.Debug("Retrieving payload from cached backends")
+		payload, err := m.secondary.MultiSourceRead(ctx,
+			versionedCert.SerializedCert, false, verifyFnForSecondary, opts.L1InclusionBlockNum)
```
litt3: It's a bit misleading to pass in `opts.L1InclusionBlockNum` here (and below), given you've stated in the doc above that this isn't a relevant concept for v1.

Would it make sense to pass in a 0 or something instead, to make it clear what's actually being checked here?

api/proxy/store/eigenda_manager.go
```diff
@@ -91,18 +92,115 @@ func (m *EigenDAManager) Get(ctx context.Context,
...
+		if err == nil {
+			return payload, nil
+		}
+		m.log.Warn("Failed to read payload from cache targets", "err", err)
+		readErrors = append(readErrors, fmt.Errorf("read from cache targets: %w", err))
+	}
+
+	// 2 - read payloadOrEncodedPayload from EigenDA
```
litt3: Doc again mentions encoded payload, which isn't actually an option here

api/proxy/store/eigenda_manager.go
litt3: I don't grasp the suggested change here. What exactly would be verified in a "verify blob" function?

It seems odd that we would implement something new here, I thought we were trying to move logic into the clients from the proxy, not the other way around?

api/proxy/store/eigenda_manager.go
```diff
@@ -236,47 +294,19 @@ func (m *EigenDAManager) putToCorrectEigenDABackend(ctx context.Context, value [
...
+	if m.secondary.AsyncWriteEntry() { // publish put notification to secondary's subscription on PutNotify topic
+		m.log.Debug("Publishing data to async secondary stores", "commitment", commitment)
+		m.secondary.Topic() <- secondary.PutNotify{
+			Commitment: commitment,
+			Value:      value,
 		}
-
-		m.log.Debug("Reading blob from EigenDAV2 backend", "returnEncodedPayload", opts.ReturnEncodedPayload)
-		payloadOrEncodedPayload, err := m.eigendaV2.Get(ctx, versionedCert, opts.ReturnEncodedPayload)
+		// secondary is available only for synchronous writes
```
litt3: I'm not understanding what this comment is trying to say. Why is it at the bottom of the `if` statement?

api/proxy/store/eigenda_manager.go
samlaf: Good callout! Basically we agree that this check needs to be implemented.
Prob makes sense to implement it only once we've moved this logic to be shared in the new higher-level v2 client that can accept any datastore. I don't think it'll be as clean as I had originally envisioned (in large part because s3 stores paylaods instead of blobs :'( ) but let's see how close to something clean we can get.

api/proxy/store/eigenda_manager.go
samlaf: [53c47f3](https://github.com/Layr-Labs/eigenda/pull/1914/commits/53c47f373a208dccce1fbd163f4b0729bc26e1f8)

api/proxy/store/eigenda_manager.go
samlaf: Thanks for noticing this!! Actually the verifyFn doesn't even need to take an L1InclusionBlockNumber anymore! See simplification in [2911613](https://github.com/Layr-Labs/eigenda/pull/1914/commits/291161388133693143dd7f50e0ff7d569cdf5241)

api/proxy/store/eigenda_manager.go
samlaf: Yes sir! opts should not even have been passed to the L1 function! Thanks for finding these
[5b1c69f](https://github.com/Layr-Labs/eigenda/pull/1914/commits/5b1c69ffb911227dfc02b3393513dde0617110af)

api/proxy/store/eigenda_manager.go
samlaf: [91be263](https://github.com/Layr-Labs/eigenda/pull/1914/commits/91be2632ad3f75ac7ea283b060ec6e5a55a90ab4)

api/proxy/store/eigenda_manager.go
samlaf: wasn't my code just moved it around. But I think this is meant to describe the else branch. Moved it next to the else statement: [b233f37](https://github.com/Layr-Labs/eigenda/pull/1914/commits/b233f37c568a574c9cad8b7e44bba5562a538097)

api/proxy/store/eigenda_manager.go
```diff
@@ -91,18 +92,110 @@ func (m *EigenDAManager) Get(ctx context.Context,
...
+		return m.eigenda.Verify(ctx, cert, payload)
 	}
 
-	// The eigendav2 verify function verifies the cert, but not the kzg commitment.
-	// TODO: should we be verifying those when retrieving from secondary storages like S3...?
-	verifyMethod, err := m.getVerifyMethod(versionedCert.Version)
+	var readErrors []error
+	// 1 - read payload from cache if enabled
+	// Secondary storages (cache and fallback) store payloads instead of blobs.
+	// For simplicity, we bypass secondary storages when requesting encoded payloads,
```
litt3: doc still referencing "requesting encoded payloads", even though that's not relevant to this method

api/proxy/store/eigenda_manager.go
```diff
@@ -91,18 +92,110 @@ func (m *EigenDAManager) Get(ctx context.Context,
...
+	}
+
+	// 2 - read payload from EigenDA
+	payload, err := m.eigenda.Get(ctx, versionedCert.SerializedCert)
+	if err == nil {
+		err = m.eigenda.Verify(ctx, versionedCert.SerializedCert, payload)
+		if err != nil {
+			return nil, fmt.Errorf("verify EigenDA V1 cert: %w", err)
+		}
+		// Only backup to secondary storage if we're returning the decoded payload
```
litt3: another doc which is talking about encoded vs decoded payload, even though retrieving an encoded payload isn't a relevant concept here

api/proxy/store/eigenda_manager.go
```diff
@@ -91,18 +92,110 @@ func (m *EigenDAManager) Get(ctx context.Context,
...
+		// TODO: we could consider also storing encoded payloads under separate keys?
+		if m.secondary.WriteOnCacheMissEnabled() {
+			m.backupToSecondary(ctx, versionedCert.SerializedCert, payload)
+		}
+		return payload, nil
+	}
+	readErrors = append(readErrors, fmt.Errorf("read from EigenDA backend: %w", err))
+
+	// 3 - read blob from fallbacks if enabled and data is non-retrievable from EigenDA
+	// Only use fallbacks if we're not requesting encoded payload
```
litt3: another stale doc

api/proxy/store/eigenda_manager.go
samlaf: [9de11c7](https://github.com/Layr-Labs/eigenda/pull/1914/commits/9de11c7e3ad1a9fa5caa7cbad02c180c85dceac4)

api/proxy/store/eigenda_manager.go
samlaf: [9de11c7](https://github.com/Layr-Labs/eigenda/pull/1914/commits/9de11c7e3ad1a9fa5caa7cbad02c180c85dceac4)

api/proxy/store/eigenda_manager.go
samlaf: [9de11c7](https://github.com/Layr-Labs/eigenda/pull/1914/commits/9de11c7e3ad1a9fa5caa7cbad02c180c85dceac4)

PR #1917, opened 2025-08-25
feat(payments): Simplify cumulative payments API

litt3:
closes EGDA-1853

replaces https://github.com/Layr-Labs/eigenda/pull/1901

### DIFF
--- a/core/payments/ondemand/on_demand_ledger.go
+++ b/core/payments/ondemand/on_demand_ledger.go
@@ -48,6 +48,9 @@ type OnDemandLedger struct {
 //
 // The CumulativePaymentStore is used in this constructor to get the current cumulative payment value. After each
 // debit, the latest cumulative payment will be stored in the CumulativePaymentStore.
+//
+// This is the constructor that should be used by those who persist on-demand payment data. Under the current
+// payment architecture, that means the disperser.
 func OnDemandLedgerFromStore(
 	ctx context.Context,
 	// the total deposits that have been made for the account to the PaymentVault
@@ -73,6 +76,10 @@ func OnDemandLedgerFromStore(
 
 // Creates a new OnDemandLedger, which *isn't* backed by a CumulativePayment store: the only representation of the
 // cumulative payment is in memory.
+//
+// This is the constructor that should be used by those who don't persist on-demand data. Under the current
+// payment architecture, that means the client. The client will get the latest cumulativePayment from the disperser
+// when starting up, and use that value to initialize the OnDemandLedger.
 func OnDemandLedgerFromValue(
 	// the total deposits that have been made for the account to the PaymentVault
 	totalDeposits *big.Int,


core/payments/ondemand/on_demand_ledger.go
```diff
@@ -23,37 +24,67 @@ import (
...
+}
+
+// Creates a new OnDemandLedger, which *isn't* backed by a CumulativePayment store: the only representation of the
+// cumulative payment is in memory.
+func OnDemandLedgerFromValue(
+	totalDeposits *big.Int,
+	pricePerSymbol *big.Int,
+	minNumSymbols uint64,
+	cumulativePayment *big.Int,
+) (*OnDemandLedger, error) {
```
cody-littley: nit, could you document these function arguments? Perhaps a bullet pointed list?

core/payments/ondemand/cumulative_payment_store.go
```diff
@@ -2,26 +2,130 @@ package ondemand
...
-		amount *big.Int,
-		// the maximum value for the cumulative payment, after the addition. Should be determined by the `TotalDeposits`
-		// for the account in the payment vault
-		maxCumulativePayment *big.Int,
-	) (*big.Int, error)
+// TODO(litt3): there are some potential avenues for optimization of this store:
+// 1. Use something other than DynamoDB. DynamoDB is being used for historical reasons, but there is only a single
+// writer now, which doesn't need any of the distributed DB properties provided by DynamoDB.
+// 2. Implement a write queue, so that the caller doesn't need to wait for the write to complete. The callers of the
+// CumulativePaymentStore just need *eventual* persistence of the cumulative payment, so using a queue would be
```
samlaf, line 32: just to be clear eventual persistence also means chance of breaking consistency, and undercharging. But we are also fine with that afaiu

core/payments/ondemand/on_demand_ledger.go
```diff
@@ -109,33 +147,60 @@ func (odl *OnDemandLedger) Debit(
...
+			BlobCost:                 blobCost,
+		}
+	}
+
+	if odl.cumulativePaymentStore != nil {
+		err := odl.cumulativePaymentStore.StoreCumulativePayment(ctx, newCumulativePayment)
+		if err != nil {
+			return nil, fmt.Errorf("store cumulative payment: %w", err)
+		}
 	}
```
cody-littley: The way I read this code, it seems like this struct is scoped to a single account... right? That is, each client will get its own OnDemandLedger. 

If that's not the case, then this lock while we access dynamo might cause problems.

core/payments/ondemand/on_demand_ledger.go
```diff
@@ -109,33 +147,60 @@ func (odl *OnDemandLedger) Debit(
...
+	newCumulativePayment := new(big.Int).Add(odl.cumulativePayment, blobCost)
+	if newCumulativePayment.Cmp(odl.totalDeposits) > 0 {
+		return nil, &InsufficientFundsError{
+			CurrentCumulativePayment: odl.cumulativePayment,
+			TotalDeposits:            odl.totalDeposits,
+			BlobCost:                 blobCost,
+		}
+	}
+
+	if odl.cumulativePaymentStore != nil {
```
cody-littley: Idea, feel free to skip it if you don't like it.

If we add the following to the `cumulativePaymentStore` implementation's functions:

```
if s == nil {
   // sane no-op behavior
}
```

You could then remove an indention level from a bunch of code in this class, since calling functions on nil would not panic.

cody-littley (dismissed):

core/payments/ondemand/cumulative_payment_store.go
```diff
@@ -2,26 +2,130 @@ package ondemand
...
+		return errors.New("newCumulativePayment cannot be nil")
+	}
+	if newCumulativePayment.Sign() < 0 {
+		return fmt.Errorf("cumulative payment cannot be negative: received %s", newCumulativePayment.String())
+	}
+
+	_, err := s.dynamoClient.UpdateItem(ctx, &dynamodb.UpdateItemInput{
+		TableName:        s.tableName,
+		Key:              s.accountKey,
+		UpdateExpression: aws.String("SET #cp = :new"),
```
samlaf:
```suggestion
		UpdateExpression: aws.String("SET #cp = :newValue"),
```

samlaf: was confused for a little while until I realized `:new` was just a variable name

core/payments/ondemand/on_demand_ledger.go
```diff
@@ -23,37 +24,67 @@ import (
...
 	pricePerSymbol *big.Int
 	// minimum number of symbols to bill
 	minNumSymbols uint64
-	// stores the cumulative payment for this ledger in wei
-	cumulativePaymentStore CumulativePaymentStore
+
+	// an optional store to back the cumulative payment for this account
+	//
+	// if non-nil, the new cumulative payment value will be stored here after each debit
+	cumulativePaymentStore *CumulativePaymentStore
```
samlaf, line 39: this feels a little bug-prone. is passing nil only used for tests?

samlaf: oh I see that you at least force non-nil in the constructor

core/payments/ondemand/on_demand_ledger.go
```diff
@@ -23,37 +24,67 @@ import (
...
+	if err != nil {
+		return nil, fmt.Errorf("get cumulative payment from store: %w", err)
+	}
+
+	return newOnDemandLedger(totalDeposits, pricePerSymbol, minNumSymbols, cumulativePaymentStore, cumulativePayment)
+}
+
+// Creates a new OnDemandLedger, which *isn't* backed by a CumulativePayment store: the only representation of the
+// cumulative payment is in memory.
+func OnDemandLedgerFromValue(
```
samlaf: hmm is this really necessary? Perhaps call it `OnDemandLedgerMemstore"? FromValue sounds like its a normal constructor, not only used for tests.

samlaf (approved):

core/payments/ondemand/on_demand_ledger.go
litt3: Correct, this struct is created per-account

core/payments/ondemand/cumulative_payment_store.go
litt3: Right. I'll note that caveat, but I'm pretty confident we're ok with tradeoffs like this for the sake of cleaner architecture + better performance 👍

core/payments/ondemand/on_demand_ledger.go
litt3: This will also be `nil` when the client runs this, i.e. client is running in-memory only, not backed by a DB. This will be non-nil when run on the disperser, which needs to persist the data

There are actually two separate public constructors, one for an instance that is backed by a `CumulativePaymentStore`, and one that isn't.

core/payments/ondemand/on_demand_ledger.go
litt3: It is a normal constructor- this is what a client will use to instantiate this.

litt3: Do you think this confusion is something that can be addressed with docs, or are you finding issue with the architecture in general?

core/payments/ondemand/on_demand_ledger.go
litt3: 1ddbd1d7858bdc901ea552ec7791762016b3143f

core/payments/ondemand/on_demand_ledger.go
litt3: 77b258a4a6b7070fe00004b1d62ed4144c6383b6

litt3: Note: I intentionally only added this to the `StoreCumulativePayment` method. `GetCumulativePayment` will only be called in places where the pointer isn't nil

core/payments/ondemand/cumulative_payment_store.go
litt3: 7669e6d191e3e85c75e9a677bb929d8dbfdc922e

core/payments/ondemand/cumulative_payment_store.go
litt3: e6b4f2c2df758e4268dfcfea048eb5f531cad72e

core/payments/ondemand/on_demand_ledger.go
litt3: 20fe052934395b734d04edc5c84c950e6310260a

I added some additional detail in this commit. LMK what you think

cody-littley (approved):

core/payments/ondemand/on_demand_ledger.go
samlaf: Oh wasn't thinking about the clients at all. Doc additions clear it up, thanks

PR #1919, opened 2025-08-26
feat: rapid LittDB flushes

cody-littley:

Rate limit and batch LittDB flushes, required to make flushes efficient with very rapid rate of flushing.

### DIFF
--- a/litt/disktable/disk_table.go
+++ b/litt/disktable/disk_table.go
@@ -496,8 +496,6 @@ func (d *DiskTable) Close() error {
 		return nil
 	}
 
-	d.flushCoordinator.Close()
-
 	if ok, err := d.errorMonitor.IsOk(); !ok {
 		return fmt.Errorf("cannot process Stop() request, DB is in panicked state due to error: %w", err)
 	}
--- a/litt/disktable/flush_coordinator.go
+++ b/litt/disktable/flush_coordinator.go
@@ -2,7 +2,6 @@ package disktable
 
 import (
 	"fmt"
-	"sync/atomic"
 	"time"
 
 	"github.com/Layr-Labs/eigenda/litt/util"
@@ -30,9 +29,6 @@ type flushCoordinator struct {
 
 	// used to rate limit flushes
 	rateLimiter *rate.Limiter
-
-	// set to false when the coordinator is closed
-	alive atomic.Bool
 }
 
 // A request to flush the underlying database. When the flush is eventually performed, a response is sent on
@@ -54,7 +50,6 @@ func newFlushCoordinator(
 		internalFlush: internalFlush,
 		requestChan:   make(chan any, requestChanBufferSize),
 	}
-	fc.alive.Store(true)
 
 	if flushPeriod > 0 {
 		fc.rateLimiter = rate.NewLimiter(rate.Every(flushPeriod), 1)
@@ -92,11 +87,6 @@ func (c *flushCoordinator) Flush() error {
 
 }
 
-// Closes the flush coordinator, cleaning up any resources.
-func (c *flushCoordinator) Close() {
-	c.alive.Store(false)
-}
-
 // The control loop that manages flush timing.
 func (c *flushCoordinator) controlLoop() {
 	defer close(c.requestChan)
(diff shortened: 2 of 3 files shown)


litt/disktable/flush_coordinator.go
```diff
@@ -0,0 +1,151 @@
...
+// caller's perspective, just the performance/timing.
+type flushCoordinator struct {
+	// Used to manage the lifecycle of LittDB threading resources.
+	errorMonitor *util.ErrorMonitor
+
+	// The function that actually performs the flush on the underlying database.
+	internalFlush func() error
+
+	// Channel to send flush requests to the control loop.
+	requestChan chan any
```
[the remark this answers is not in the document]
cody-littley: Yeah, but golang generics force me to do it (see the error monitor utility functions). Won't do.

litt/disktable/flush_coordinator.go
```diff
@@ -0,0 +1,151 @@
...
+				return
+			case request := <-c.requestChan:
+				waitingRequests.Enqueue(request)
+			case <-timer.C:
+
+				// we can now perform a flush
+				err := c.internalFlush()
+				// send a response to each waiting caller
+				for request, ok := waitingRequests.Dequeue(); ok; request, ok = waitingRequests.Dequeue() {
+					request.(flushCoordinatorRequest) <- err
```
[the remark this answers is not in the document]
cody-littley: Eh, not worth it. This is a small enough class that it's easy to reason about.

litt/disktable/disk_table.go
```diff
@@ -492,6 +496,9 @@ func (d *DiskTable) Close() error {
 		return nil
 	}
 
+	// TODO this should be integrated with the error monitor!
+	d.flushCoordinator.Close()
```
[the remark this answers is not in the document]
cody-littley: removed the TODO, this was already handled

litt/disktable/flush_coordinator.go
```diff
@@ -0,0 +1,151 @@
...
+
+// The control loop that manages flush timing.
+func (c *flushCoordinator) controlLoop() {
+	defer close(c.requestChan)
+
+	// requests that are waiting for a flush to be performed
+	waitingRequests := linkedlistqueue.New()
+
+	// timer used to wait until the next flush can be performed
+	timer := time.NewTimer(0)
```
litt3, line 98: should you defer a `timer.Stop`?

litt/disktable/flush_coordinator.go
```diff
@@ -0,0 +1,151 @@
...
+				for request, ok := waitingRequests.Dequeue(); ok; request, ok = waitingRequests.Dequeue() {
+					request.(flushCoordinatorRequest) <- err
+				}
+
+				timerActive = false
+			}
+		} else {
+			// There are pending flushes, we can handle requests immediately if the rate limiter allows it.
+			select {
+			case <-c.errorMonitor.ImmediateShutdownRequired():
```
litt3, line 126: would it ever be possible for `alive` to become false, without an immediate shutdown being required?

The concern is that you might be waiting at one of these `select` statements, `alive` becomes false, but there's nothing to break you out of the `select`

litt/disktable/flush_coordinator.go
```diff
@@ -0,0 +1,151 @@
...
+				err := c.internalFlush()
+				// send a response to each waiting caller
+				for request, ok := waitingRequests.Dequeue(); ok; request, ok = waitingRequests.Dequeue() {
+					request.(flushCoordinatorRequest) <- err
+				}
+
+				timerActive = false
+			}
+		} else {
+			// There are pending flushes, we can handle requests immediately if the rate limiter allows it.
```
litt3: Not sure I understand what this comment is getting at. Do you mean there are NO pending flushes? Or something else?

litt3 (approved):

litt/disktable/flush_coordinator.go
cody-littley: good catch, added

litt/disktable/flush_coordinator.go
cody-littley: This isn't possible in production code, but you make a good point -- this is a potential foot gun.

I've removed `Close()`, and rely on the `ErrorMonitor` shutdown sequence to clean up resources in unit tests.

litt/disktable/flush_coordinator.go
cody-littley: I agree this comment is confusing. Updated to the following:

```
			// We don't have any pending flush requests, so we aren't waiting on the timer. If we get a new request,
			// check to see if the rate limiter will allow it to be flushed immediately, and do so if possible.
```

PR #1920, opened 2025-08-26
feat(payments): Implement client ledger

litt3:
closes EGDA-1854


core/payments/client_ledger.go
```diff
@@ -0,0 +1,289 @@
...
+	if accountID == (gethcommon.Address{}) {
+		panic("account ID cannot be zero address")
+	}
+
+	switch clientLedgerMode {
+	case ClientLedgerModeReservationOnly:
+		if reservationLedger == nil || onDemandLedger != nil {
+			panic(fmt.Sprintf("in %s mode, expected reservation ledger to be non-nil and on-demand ledger to be nil",
+				ClientLedgerModeReservationOnly))
+		}
```
cody-littley: Nit, possibly a good place to use the new `enforce` library.

```suggestion
		enforce.True(reservationLedger == nil || onDemandLedger != nil,
		    "in %s mode, expected reservation ledger to be non-nil and on-demand ledger to be nil", ClientLedgerModeReservationOnly)
```

If you adopt this suggestion, I suggest you also make similar changes elsewhere in this file.

cody-littley: In the panic message you say `and`, but in the code you use `||`. Same is true for the cases below.

core/payments/client_ledger.go
```diff
@@ -0,0 +1,289 @@
...
+			panic(fmt.Sprintf("in %s mode, expected on-demand ledger to be non-nil and reservation ledger to be nil",
+				ClientLedgerModeOnDemandOnly))
+		}
+	case ClientLedgerModeReservationAndOnDemand:
+		if reservationLedger == nil || onDemandLedger == nil {
+			panic(fmt.Sprintf("in %s mode, expected reservation and on-demand ledgers to be non-nil",
+				ClientLedgerModeReservationAndOnDemand))
+		}
+	default:
+		panic(fmt.Sprintf("unknown clientLedgerMode %s", clientLedgerMode))
```
cody-littley, line 89: Perhaps we can add a new function `enforce.Unreachable(msg, args)`. Not that raw panic is bad, but it might be nice to have stylistic uniformity. Also, it would save you from having an extra `fmt.Sprintf` that you have to type.

cody-littley (dismissed):

core/payments/client_ledger.go
litt3: ca17b1f2f636446e0f99546ca7c5bdbf0745277d

core/payments/client_ledger.go
litt3: The problem with something like `Unreachable` is that it's not recognized as a `panic`, so the compiler still forces you to return a value after the call to `Unreachable`. Kinda messy

cody-littley: ok, let's keep the panic then

core/payments/client_ledger.go
```diff
@@ -69,33 +70,26 @@ func NewClientLedger(
...
+		enforce.NotNil(onDemandLedger, "in %s mode, on-demand ledger must be non-nil",
+			ClientLedgerModeReservationAndOnDemand)
 	default:
 		panic(fmt.Sprintf("unknown clientLedgerMode %s", clientLedgerMode))
 	}
 
-	if getNow == nil {
-		panic("getNow must not be nil")
-	}
+	enforce.True(getNow != nil, "getNow function must not be nil")
```
cody-littley, line 92:
```suggestion
	enforce.NotNil(getNow, "getNow function must not be nil")
```

cody-littley (dismissed):

common/enforce/assertions.go
```diff
@@ -125,3 +125,10 @@ func MapDoesNotContainKey[K comparable, V any](m map[K]V, key K, message string,
 		panic(fmt.Sprintf("Expected map to not contain key %v: %s", key, fmt.Sprintf(message, args...)))
 	}
 }
+
+// Asserts that an error is nil and panics with a message if it is not nil.
+func NilError(err error, message string, args ...any) {
+	if err != nil {
+		panic("Expected error to be nil: " + fmt.Sprintf(message, args...))
```
samlaf: prob also want to log the err?

core/data.go
```diff
@@ -505,6 +505,43 @@ type PaymentMetadata struct {
 	CumulativePayment *big.Int `json:"cumulative_payment"`
 }
 
+func NewPaymentMetadata(
```
samlaf, line 511: personally not a big fan of having a bunch of miscellaneous stuff under `core.`.
Do we really need this here or can we also keep it under `core/payments`?

core/data.go
```diff
@@ -505,6 +505,43 @@ type PaymentMetadata struct {
...
+	// account that the payment is for. must not be a 0 address
+	accountID gethcommon.Address,
+	// the time of the dispersal, which is stored as a unix nano timestamp
```
samlaf:
```suggestion
	// the time of the dispersal
```

samlaf: nit: technically not a unix nano ts right? its got all the other stuff, like the monotonic timestamp part for example

core/payments/client_ledger.go
```diff
@@ -0,0 +1,270 @@
...
+
+	// Though it would theoretically be possible to infer mode of operation based on on-chain state, it's important
+	// that this be directly configurable by the user, to ensure that reality matches intention.
+	//
+	// Consider, for example, if a user intends to operate with a reservation covering the majority of dispersals,
+	// with an on-demand balance as a backup. If there is a configuration issue which prevents the reservation from
+	// being used, the client could mistakenly burn through all backup funds before becoming aware of the
+	// misconfiguration. In such cases, it's better to fail early, to bring the misconfiguration to the attention of the
+	// user as soon as possible.
+	clientLedgerMode ClientLedgerMode
```
samlaf, line 47: thinking through this. the main user of this client will be proxy. I don't think (?) we want to add a proxy flag to set this? So we prob want to programmatically create the reservationLedger and onDemanderLedger anyways, and then set this mode also programmatically. In such a case I'm not sure I see the value.

Agree in general we prob should keep this and its generally good, just trying to fetch for your thoughts here and think through this a bit more.

core/payments/client_ledger.go
```diff
@@ -0,0 +1,270 @@
...
+		panic(fmt.Sprintf("reservation debit failed: %v", err))
+	}
+
+	if success {
+		paymentMetadata, err := core.NewPaymentMetadata(cl.accountID, now, nil)
+		enforce.NilError(err, "new payment metadata: %v", err)
+		return paymentMetadata, nil
+	}
+
+	cl.logger.Infof("Reservation lacks capacity for blob with %d symbols (%d bytes). Falling back to on-demand.",
```
samlaf, line 218: When reservation is turned on, are you checking that the reservation is actually >0 and panic'ing if not?

samlaf (dismissed):

core/payments/client_ledger.go
litt3: I've been mulling this over. I decided to implement it with this feature, so that it's available to us if we want it, but won't do any harm if not.

I was thinking it might make sense to have an *optional* proxy flag, which lets the user declare what sort of payments they want to use. If unset, it could be determined automatically.

Due to scenarios like the one I describe in the doc, I think there is value in letting users be _sure_ they are using the payment type they think they're using. It would be difficult to verify manually whether this is configured properly

core/data.go
litt3: The problem is that `PaymentMetadata` is already defined in this file, and it would seem weird to put the constructor in a different package.

But I also don't want to move the `PaymentMetadata` struct yet, because the old payment logic depends on it. I want to make as few changes to the old payment logic as possible, until it's time to just delete it.

f8f910ceacca4260b5a1f8ee2bc4d18fda696996

I added a TODO about the move.

core/data.go
litt3:
```
	// Timestamp represents the nanosecond of the dispersal request creation
	Timestamp int64 `json:"timestamp"`
```

It's stored as a unix timestamp in the `PaymentMetadata` struct. The constructor just takes a `time.Time` instance for "convenience", but I'm unsure if that's the right call. Thoughts?

common/enforce/assertions.go
litt3: 902c355b78f422f481e09dd7e98489f836674cbf

core/payments/client_ledger.go
litt3: It's not possible to create a `ReservationLedger` with 0 symbols per second: it's checked when creating a `Reservation` object, and when creating the underlying `LeakyBucket`

core/data.go
samlaf: Oh interesting... tbh my brain is a bit fried right now so wouldn't want to push code in wrong direction (agian) :\
How about we just make the comment more explicit like "the non-monotonic unix timestamp ns will be extracted from this timestamp." 
and perhaps add a TODO to consider whether we should just pass in ns directly? although passing ts seems fine tbh. I think (?) all time instances have to have a unix ts part right? its only the monotonic part that can be missing?

core/payments/client_ledger.go
samlaf: Agree. Actually thinking through this a bit more I also came to the conclusion that adding a proxy flag could actually make sense. can think through whether we want it optional or required later. if optional default would be to use both reservation and on-demand?

core/payments/client_ledger.go
samlaf: perfect thanks

samlaf (approved):

core/payments/client_ledger.go
litt3: I think the optional default would be "whatever the account has registered onchain"

i.e. if you have a reservation but no on-demand balance, use reservation. Or any other combination of the two payment types

core/data.go
litt3: > all time instances have to have a unix ts part right? its only the monotonic part that can be missing?

Right.

> although passing ts seems fine tbh

Alright, I decided just to make the doc more explicit, but skipped the todo.

7cac5af5732826a1a138c9e565feb27369dfb62b

PR #1922, opened 2025-08-27
feat(metrics): add v2 client retrieval metrics

iquidus:
Closes DAINT-693

Adds retrieval metrics for payload retrieval via both relay and validator paths

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/payloadretrieval/validator_payload_retriever.go
+++ b/api/clients/v2/payloadretrieval/validator_payload_retriever.go
@@ -79,7 +79,7 @@ func (pr *ValidatorPayloadRetriever) GetPayload(
 		return nil, coretypes.ErrBlobDecodingFailedDerivationError.WithMessage(err.Error())
 	}
 
-	pr.metrics.RecordPayloadSizeBytes(uint(len(payload)))
+	pr.metrics.RecordPayloadSizeBytes(len(payload))
 
 	return payload, nil
 }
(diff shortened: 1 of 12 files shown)


api/clients/v2/payloadretrieval/validator_payload_retriever.go
```diff
@@ -75,6 +79,8 @@ func (pr *ValidatorPayloadRetriever) GetPayload(
 		return nil, coretypes.ErrBlobDecodingFailedDerivationError.WithMessage(err.Error())
 	}
 
+	pr.metrics.RecordPayloadSizeBytes(uint(len(payload)))
```
ethenotethan: nit - if the final callee destination type is `float64` then why cast to `uint` and not instead just take `int` as the function arg and cast that to `float64`?
ie:
```
float64(len(payload))
```

ethenotethan (approved):
LGTM - just one minor nit

api/clients/v2/payloadretrieval/validator_payload_retriever.go
iquidus: Changed in [1a906a1](https://github.com/Layr-Labs/eigenda/pull/1922/commits/1a906a19aad3f12ff5b3d606dee90b3a45d4338c)

PR #1923, opened 2025-08-27
feat(payments): Implement `OnDemandPaymentValidator`

litt3:
Part of EGDA-1855

### DIFF
--- a/core/payments/ondemand/on_demand_payment_validator.go
+++ b/core/payments/ondemand/on_demand_payment_validator.go
@@ -1,0 +1,179 @@
+package ondemand
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"math/big"
+	"sync"
+
+	"github.com/Layr-Labs/eigenda/core/meterer"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	"github.com/aws/aws-sdk-go-v2/service/dynamodb"
+	gethcommon "github.com/ethereum/go-ethereum/common"
+	"github.com/hashicorp/go-multierror"
+	lru "github.com/hashicorp/golang-lru/v2"
+)
+
+// OnDemandPaymentValidator validates on-demand payments for multiple accounts
+type OnDemandPaymentValidator struct {
+	logger logging.Logger
+	// A cache of the ledgers being tracked.
+	//
+	// New OnDemandLedger entries are added to this cache as Debit requests are received from new accounts. Least
+	// recently used OnDemandLedger entries are removed if the cache gets above the configured size. Since on-demand
+	// payment data is stored in a persistent way, deleting an OnDemandLedger from memory doesn't result in data loss:
+	// it just means that a new OnDemandLedger object will need to be constructed if any future Debits must be handled
+	// from that account.
+	ledgers *lru.Cache[gethcommon.Address, *OnDemandLedger]
+	// protects concurrent access to the ledgers cache during ledger creation
+	ledgerCreationLock sync.Mutex
+	// Provides access to the values stored in the PaymentVault contract.
+	//
+	// The state of this object is updated on a background thread.
+	onChainState meterer.OnchainPayment
+	// the underlying dynamo client, which is used by all OnDemandLedger instances created by this struct
+	dynamoClient *dynamodb.Client
+	// the name of the dynamo table where on-demand payment information is stored
+	onDemandTableName string
+}
+
+// NewOnDemandPaymentValidator creates a new OnDemandPaymentValidator with specified cache size
+func NewOnDemandPaymentValidator(
+	logger logging.Logger,
+	// the maximum number of OnDemandLedger entries to be kept in the LRU cache
+	maxLedgers int,
+	// expected to be initialized and have its background update thread started
+	onChainState meterer.OnchainPayment,
+	dynamoClient *dynamodb.Client,
+	// the name of the dynamo table where on-demand payment information is stored
+	onDemandTableName string,
+) (*OnDemandPaymentValidator, error) {
+	if onChainState == nil {
+		return nil, errors.New("onChainState cannot be nil")
+	}
+	cache, err := lru.NewWithEvict(
+		maxLedgers,
+		func(accountAddress gethcommon.Address, _ *OnDemandLedger) {
+			logger.Infof("evicted account %s from LRU on-demand ledger cache", accountAddress.Hex())
+		},
+	)
+	if err != nil {
+		return nil, fmt.Errorf("new LRU cache with evict: %w", err)
+	}
+
+	if dynamoClient == nil {
+		return nil, errors.New("dynamo client cannot be nil")
+	}
+
+	if onDemandTableName == "" {
+		return nil, errors.New("on demand table name cannot be empty")
+	}
+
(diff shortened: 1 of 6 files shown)


core/payments/ondemand/on_demand_payment_validator.go
```diff
@@ -0,0 +1,179 @@
...
+	onChainState meterer.OnchainPayment,
+	dynamoClient *dynamodb.Client,
+	// the name of the dynamo table where on-demand payment information is stored
+	onDemandTableName string,
+) (*OnDemandPaymentValidator, error) {
+	if onChainState == nil {
+		return nil, errors.New("onChainState cannot be nil")
+	}
+	cache, err := lru.NewWithEvict(
+		maxLedgers,
```
samlaf, line 61: what happens if user sets this to 0 (default value for int)? Should we validate against this?

core/payments/ondemand/on_demand_payment_validator.go
```diff
@@ -0,0 +1,179 @@
...
+	// A cache of the ledgers being tracked.
+	//
+	// New OnDemandLedger entries are added to this cache as Debit requests are received from new accounts. Least
+	// recently used OnDemandLedger entries are removed if the cache gets above the configured size. Since on-demand
+	// payment data is stored in a persistent way, deleting an OnDemandLedger from memory doesn't result in data loss:
+	// it just means that a new OnDemandLedger object will need to be constructed if any future Debits must be handled
+	// from that account.
+	ledgers *lru.Cache[gethcommon.Address, *OnDemandLedger]
+	// protects concurrent access to the ledgers cache during ledger creation
+	ledgerCreationLock sync.Mutex
```
samlaf, line 35: https://github.com/hashicorp/golang-lru says its thread-safe, are you sure we need this mutex?

core/payments/ondemand/on_demand_payment_validator.go
```diff
@@ -0,0 +1,179 @@
...
+	accountID gethcommon.Address,
+) (*OnDemandLedger, error) {
+	// Fast path: check if ledger already exists in cache
+	if ledger, exists := pv.ledgers.Get(accountID); exists {
+		return ledger, nil
+	}
+
+	// Slow path: acquire lock and check again
+	pv.ledgerCreationLock.Lock()
+	defer pv.ledgerCreationLock.Unlock()
```
samlaf, line 120: are you sure this lock is needed?

core/payments/ondemand/on_demand_payment_validator.go
```diff
@@ -0,0 +1,179 @@
...
+
+	onDemandPayment, err := pv.onChainState.GetOnDemandPaymentByAccount(ctx, accountID)
+	if err != nil {
+		return nil, fmt.Errorf("get on-demand payment for account %v: %w", accountID.Hex(), err)
+	}
+
+	pricePerSymbol := big.NewInt(int64(pv.onChainState.GetPricePerSymbol()))
+	minNumSymbols := pv.onChainState.GetMinNumSymbols()
+
+	cumulativePaymentStore, err := NewCumulativePaymentStore(pv.dynamoClient, pv.onDemandTableName, accountID)
```
samlaf, line 134: bike-shed: wondering if we should consider adding a TTL to the LRU to force any cached ledger to refetch these items in case they have changed onchain?

core/payments/ondemand/on_demand_payment_validator.go
```diff
@@ -0,0 +1,179 @@
...
+	var result *multierror.Error
+	for _, update := range updates {
+		ledger, exists := pv.ledgers.Get(update.AccountAddress)
+		if !exists {
+			// if we aren't already tracking the account, there's nothing to do. we'll start tracking it if the
+			// account ever makes an on-demand dispersal
+			continue
+		}
+
+		err := ledger.UpdateTotalDeposits(update.NewTotalDeposit)
```
samlaf, line 172: each of these is making a network call to write to dynamodb correct?
If that's the case then consider using an https://pkg.go.dev/golang.org/x/sync/errgroup to make the requests in parallel instead.

core/payments/ondemand/on_demand_payment_validator.go
litt3: There are unit tests for this- `NewWithEvict` returns an error for values <= 0

core/payments/ondemand/on_demand_payment_validator.go
litt3: The structure itself is threadsafe, but this lock isn't to protect access to the struct.

Consider 2 callers fail to access a ledger via the fast path, and enter the slow path at the same moment. They will both construct a new ledger, add it to the cache, and then return it for use.

That means that whichever caller finishes construction first will return a ledger that gets used for the dispersal, but that instance will be overwritten in the cache by the second caller, effectively making the first dispersal "free"

core/payments/ondemand/on_demand_payment_validator.go
litt3: See https://github.com/Layr-Labs/eigenda/pull/1923#discussion_r2307440565

core/payments/ondemand/on_demand_payment_validator.go
litt3: There will be consideration for changing vault values in an upcoming PR.

MVP will use a periodic fetch of vault data, matching the current pattern that's used by the meterer. A fast follow will convert this to an event subscription model, to make updates more promptly

In both cases, evicting from the cache isn't necessary: we'll just update the parameters in-place

core/payments/ondemand/on_demand_payment_validator.go
litt3: I don't think the optimization is worth it in this case.

1. In the MVP, a thread will wake every 1 or 5 minutes, fetch up-to-date info, and then update these values. It is highly unlikely that there will be many updates at once: there just won't be that many users making vault deposits in a given 5 minute period
2. This update will be called in a background thread, and the lock acquired for the dynamo write only affects the single account being update. Since the update method triggers very infrequently, it doesn't matter if it takes a little longer to go through each account, and parallelization doesn't change the total impact on any given account.

core/payments/ondemand/on_demand_payment_validator.go
samlaf: Are you sure that's how it works? Wouldn't LRU instead add both callers' ledgers to the cache and evict too other ledgers that haven't been used in a while?

litt3: I'm talking about a case where two callers are accessing the same account. In this case, the two ledgers would share the same key, so it's not possible that they coexist in the cache

I suppose the creation lock _could_ be less restrictive, only acquiring an account specific lock. Not sure it's worth the added complexity for that, though

samlaf (approved):

core/payments/ondemand/on_demand_payment_validator.go
litt3: 54b735383db3074e3e0c3199b662f33960a80ff6

Tried to improve the doc to make this more clear

samlaf (approved):

PR #1924, opened 2025-08-27
feat(v2-clients): dispersal failover errors

samlaf:
Closes DAINT-458.

The v2 clients were not returning failover error signals for proxy to return teapot errors to tell rollup clients to failover to ethDA or some other DA layer.
This PR adds a few failover sites. It is most definitely not complete, and is still very coarse grained.

In the longer term, we might want to add more high-level proxy strategies such as "if none of the past 100 puts have gone through, failover"?

Contrary to the original intention in https://github.com/Layr-Labs/eigenda/pull/1938, decided to map BOTH INVALID_CERT and INTERNAL_ERROR cert verifier errors to failover signals. This is because the disperser will retry 3 times trying to construct a valid cert, and if it can't, then it should failover. For example, if it can't get enough signatures to pass its set thresholds, then eigenda nodes might be down.

### DIFF
--- a/api/clients/v2/payloaddispersal/payload_disperser.go
+++ b/api/clients/v2/payloaddispersal/payload_disperser.go
@@ -170,9 +170,13 @@ func (pd *PayloadDisperser) SendPayload(
 
 	err = pd.certVerifier.CheckDACert(timeoutCtx, eigenDACert)
 	if err != nil {
-		// TODO(samlaf): do we want to return an api.ErrorFailover here? In case the disperser is being malicious
-		// and returning invalid data? This could also just indicate that our cert building procedure above is buggy though...
-		// but perhaps also a reason to failover?
+		var errInvalidCert *verification.CertVerifierInvalidCertError
+		if errors.As(err, &errInvalidCert) {
+			// Regardless of whether the cert is invalid (400) or certVerifier contract has a bug (500),
+			// we send a failover signal. If we can't construct a valid cert after retrying a few times (proxy retry policy),
+			// then its safer for the rollup to failover to another DA layer.
+			return nil, api.NewErrorFailover(fmt.Errorf("checkDACert failed with blobKey %v: %w", blobKey.Hex(), err))
+		}
 		return nil, fmt.Errorf("verify cert for blobKey %v: %w", blobKey.Hex(), err)
 	}
 


api/clients/v2/payloaddispersal/payload_disperser.go
```diff
@@ -169,6 +170,9 @@ func (pd *PayloadDisperser) SendPayload(
 
 	err = pd.certVerifier.CheckDACert(timeoutCtx, eigenDACert)
 	if err != nil {
+		// TODO(samlaf): do we want to return an api.ErrorFailover here? In case the disperser is being malicious
+		// and returning invalid data? This could also just indicate that our cert building procedure above is buggy though...
+		// but perhaps also a reason to failover?
 		return nil, fmt.Errorf("verify cert for blobKey %v: %w", blobKey.Hex(), err)
```
samlaf: thoughts here?

api/clients/v2/payloaddispersal/payload_disperser.go
```diff
@@ -261,13 +266,19 @@ func (pd *PayloadDisperser) pollBlobStatusUntilSigned(
...
-					// returned error is verbose enough, no need to wrap it with additional context
-					return nil, err
+					// TODO(samlaf): checkThresholds should return more fine-grained errors
+					// For now, we only failover if thresholds were unmet, not anything else.
+					// The risk of failing over for everything is that eth-rpc calls could fail
+					// for networking reasons, which we don't want to failover to eth for!
+					var thresholdNotMetErr *thresholdNotMetError
+					if errors.As(err, &thresholdNotMetErr) {
+						return nil, api.NewErrorFailover(fmt.Errorf("check thresholds: %w", err))
+					}
```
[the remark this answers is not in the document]
samlaf: this is not true... its declared on line 277

api/clients/v2/payloaddispersal/payload_disperser.go
litt3: I think failover probably makes sense.

We could do some more fine-grained logic like you mentioned elsewhere, where we only failover if this happens multiple times... but it doesn't seem worth it right now

litt3 (dismissed):

api/clients/v2/disperser_client.go
```diff
@@ -213,7 +213,7 @@ func (c *disperserClient) DisperseBlobWithProbe(
 	err := c.initOncePopulateAccountant(ctx)
 	if err != nil {
 		c.accountantLock.Unlock()
-		return nil, [32]byte{}, api.NewErrorFailover(err)
+		return nil, [32]byte{}, fmt.Errorf("error initializing accountant: %w", err)
```
bxue-l2, line 216: Is it the idea that since initialization is not working, there is no reason to retry it?

But could it happen that the retry logic actually initiate it correctly, say Eth rpc connection?

api/proxy/store/generated_key/eigenda/eigenda.go
```diff
@@ -131,25 +132,38 @@ func (e Store) Put(ctx context.Context, value []byte) ([]byte, error) {
...
+				return false
+			}
+			if errors.Is(err, &api.ErrorFailover{}) {
+				// Failover errors should be retried before failing over.
+				return true
+			}
+			grpcStatus, isGRPCError := status.FromError(err)
 			if !isGRPCError {
-				// api.ErrorFailover is returned, so we should retry
+				// This should never happen because PutBlob only returns ErrorFailover as a non grpc error.
```
bxue-l2, line 147: It also InternalError, [here](https://github.com/Layr-Labs/eigenda/blob/2efb075083e87c94db1086183ca909054e9a56b6/api/clients/eigenda_client.go#L225)

api/clients/v2/payloaddispersal/payload_disperser.go
```diff
@@ -98,7 +99,7 @@ func (pd *PayloadDisperser) SendPayload(
 	timeoutCtx, cancel := context.WithTimeout(ctx, pd.config.ContractCallTimeout)
 	defer cancel()
 
-	// NOTE: there is a synchronization edge case where the disperser accredits a RBN that correlates
+	// NOTE: there is a synchronization edge case where the disperser accredits an RBN that correlates
```
bxue-l2, line 102: oh

api/clients/v2/payloaddispersal/payload_disperser.go
bxue-l2: This could be a RPC provider issue, then you might succeed at last by retrying it. But I think the current behavior is that you will retry almost all nonGRPC error even if it is no failoverError

About failover, if it is a network issue, if after retrying it still wouldn't help it, then Failover is right thing to do. Since ETH DA probably does not work either, so nothing hurts. But if it is not network issue, like you said, failover is right

api/clients/v2/payloaddispersal/payload_disperser.go
```diff
@@ -261,13 +266,19 @@ func (pd *PayloadDisperser) pollBlobStatusUntilSigned(
...
-			// TODO: we'll need to add more in-depth response status processing to derive failover errors
 			switch newStatus {
 			case dispgrpc.BlobStatus_COMPLETE:
 				err := checkThresholds(ctx, pd.certVerifier, blobStatusReply, blobKey.Hex())
 				if err != nil {
-					// returned error is verbose enough, no need to wrap it with additional context
-					return nil, err
+					// TODO(samlaf): checkThresholds should return more fine-grained errors
+					// For now, we only failover if thresholds were unmet, not anything else.
+					// The risk of failing over for everything is that eth-rpc calls could fail
```
bxue-l2, line 279: See comment, https://github.com/Layr-Labs/eigenda/pull/1924/files#r2305161314

About eth-rpc, we have retries for failover, sounds like a probability problem. Given a fail, p is network failure, 1-p is bug/ malicious batcher. The probably for our worst case is p^3, where 3 is number of retry. If we are confident we don't have bug, then we should fail for it.

bxue-l2: If I recall correctly, failover is on per dispersal basis. So it might be that bad if we mis-detect once or twice.

api/clients/v2/disperser_client.go
ethenotethan: what happens if this initialization fails? iiuc it puts the disperser client in a terminally ill state where no submissions can be made since the error is cached:
https://github.com/Layr-Labs/eigenda/blob/09f80609a11f2cd0a526bd4324ed0a96a2ceff50/api/clients/v2/disperser_client.go#L413-L427 

Iiuc we have to initialize this in the `Disperse` function but am curious why we don't retry if there's a cached error?  like @bxue-l2 pointed out - there are transient errors that can occur here

api/clients/v2/payloaddispersal/payload_disperser.go
ethenotethan: nit - is `an` correct lol? I'm down to byte shed 😤

api/clients/v2/payloaddispersal/payload_disperser.go
```diff
@@ -261,13 +266,19 @@ func (pd *PayloadDisperser) pollBlobStatusUntilSigned(
...
+					// for networking reasons, which we don't want to failover to eth for!
```
ethenotethan, line 280: time to retry 🫡

api/clients/v2/payloaddispersal/payload_disperser.go
litt3: `an` is correct- initialisms get the indefinite article which matches their pronunciation (so "arr-be-en" in this case)

api/clients/v2/payloaddispersal/payload_disperser.go
samlaf: Wow this was perfectly planted to be solved by https://github.com/Layr-Labs/eigenda/pull/1938.
Once that merges we'll map the `INTERNAL_ERROR` status code to failover here.

api/clients/v2/disperser_client.go
samlaf: yep you're absolutely right... I got rid of the initOnce functions for client, but forget why couldn't do it for accountant. Maybe just didn't know how to deal with it.
I'd prefer for that to be solved in a separate PR though?

api/proxy/store/generated_key/eigenda/eigenda.go
samlaf: image

InternalError is a grpc error

api/clients/v2/payloaddispersal/payload_disperser.go
samlaf: Yeah I've been caught red-handed by @litt3 many times haha

samlaf: But tbf I also feel like it depends on the reader. When I'm very used to the accronym I'll often expand them automatically in my brain. RBN here I'd read as "reference block number", in which case "a" is right.... so idk I think either is fine lol

api/clients/v2/payloaddispersal/payload_disperser.go
samlaf: Don't really believe in probabilities like that. Especially with fast dispersals, they are all going to be extremely correlated. If one fails for networking issue, chance is that the next 3 will also fail are much more likely.
Until we have very robust/low-level retry policies (say on eth client itself) then I'd prefer not too put too many failovers personally.

api/clients/v2/payloaddispersal/payload_disperser.go
ethenotethan: lmao don't have strong opinions here - personally I parse it as "reference block number" but given we're using acronym I'm fine with us making the change

samlaf: resolving then

api/clients/v2/disperser_client.go
samlaf: Actually just realized this whole accountant related rpc call is not needed in the new model with leaky buckets... believe its going away..? Created https://linear.app/eigenlabs/issue/DAINT-739/fix-accountant-initializationfailover-strategy-in-disperser-client to track this work.

api/clients/v2/payloaddispersal/payload_disperser.go
samlaf: [7326b9a](https://github.com/Layr-Labs/eigenda/pull/1924/commits/7326b9a55d1fd4e25faa566451983ea729efd6e5)
Actually opted to failover on both INVALID_CERT and INTERNAL_ERROR status codes.. I think @bxue-l2 this is what you had suggested?

api/clients/v2/disperser_client.go
bxue-l2: Read it one more time, I don't why I said retry. Because it indeed retried. 
- it is not a grpc error, so it will retry here https://github.com/Layr-Labs/eigenda/pull/1924/files#diff-a29fe93ac16dcc0e9c67ddbd336d1ddd93e45652c72e3c4cee6efaf82d31c74dR147

bxue-l2 (approved):

bxue-l2 (approved):
lgtm

PR #1925, opened 2025-08-27
feat(tools): add gas-exhaustion-cert-meter to measure the gas cost without signers

bxue-l2:

Introduces gas-exhaustion-cert-meter to estimate worst-case gas costs for EigenDA V3 certificate verification when all operators are non-signers.

Workflow
- User needs to input an existing cert as a template for removing signers and populating all the non-signers
- Based on the reference block number, the software pulls from ETH RPC for all operators and G1 pubkey, which are used for populating the non-signers
- G1 pubkey are pulled from BLSAPKRegistry, without a dependency of graph node

Sample data
- Includes mainnet and sepolia certificate examples in ./data/
- mainnet worst gas cost: 3,532,040
- sepolia gas cost: 593,619

More info on mainnet data
- there are 128 operators and two quorums in total
- 64 operators registered both quorum (so their signature counts twice)
- 64 operators registered a single quorum

Usage
```bash
  gas-exhaustion-cert-meter \
    --eigenda-network mainnet \
    --cert-rlp-file ./data/cert_v3.mainnet.rlp \
    --eth-rpc-url https://...
```
## Limitation

It is possible to remove the DAcert file requirement and craft everything by itself. But it is not done 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/tools/gas_exhaustion_cert_meter/config.go
+++ b/tools/gas_exhaustion_cert_meter/config.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 
+	proxycommon "github.com/Layr-Labs/eigenda/api/proxy/common"
 	"github.com/Layr-Labs/eigenda/common"
 	blsapkregistry "github.com/Layr-Labs/eigenda/contracts/bindings/BLSApkRegistry"
 	contractIEigenDADirectory "github.com/Layr-Labs/eigenda/contracts/bindings/IEigenDADirectory"
@@ -62,7 +63,13 @@ func ReadConfig(ctx *cli.Context, logger logging.Logger) (*Config, error) {
 		return nil, fmt.Errorf("dial Ethereum node at %s: %w", rpcURL, err)
 	}
 
-	directoryAddress := gethcommon.HexToAddress(ctx.GlobalString(flags.EigenDADirectoryFlag.Name))
+	networkString := ctx.String(flags.NetworkFlag.Name)
+	eigenDANetwork, err := proxycommon.EigenDANetworkFromString(networkString)
+	if err != nil {
+		return nil, fmt.Errorf("parse eigenDANetwork: %w", err)
+	}
+
+	directoryAddress := gethcommon.HexToAddress(eigenDANetwork.GetEigenDADirectory())
 
 	operatorStateRetrieverAddr, err := GetAddressByName(
 		ethContext, client, directoryAddress, "OPERATOR_STATE_RETRIEVER")
@@ -85,7 +92,7 @@ func ReadConfig(ctx *cli.Context, logger logging.Logger) (*Config, error) {
 		return nil, err
 	}
 
-	oStateRetrCaller, err := opstateretriever.NewContractOperatorStateRetrieverCaller(
+	opStateRetrCaller, err := opstateretriever.NewContractOperatorStateRetrieverCaller(
 		operatorStateRetrieverAddr, client)
 	if err != nil {
 		logger.Error("Failed to fetch OperatorStateRetriever contract", "err", err)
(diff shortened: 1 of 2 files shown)


tools/gas_exhaustion_cert_meter/flags/flags.go
```diff
@@ -0,0 +1,50 @@
...
+		Required: true,
+		EnvVar:   common.PrefixEnvVar(envPrefix, "EIGENDA_DIRECTORY"),
+	}
+	CertRlpFileFlag = cli.StringFlag{
+		Name: common.PrefixFlag(FlagPrefix, "cert-rlp-file"),
+		Usage: "Path to the RLP-encoded EigenDA V3 certificate file. " +
+			"Examples: ./data/cert_v3.mainnet.rlp, ./data/cert_v3.sepolia.rlp",
+		Required: true,
+		EnvVar:   common.PrefixEnvVar(envPrefix, "CERT_RLP_FILE"),
+	}
```
samlaf, line 38: nit: personally don't like using files, because its easier to copy-paste directly and create bash pipeline scripts than to have to copy them into a file.

Not a big deal right now but in general I'm against reading files as input other than for the main config.

tools/gas_exhaustion_cert_meter/flags/flags.go
```diff
@@ -0,0 +1,50 @@
...
+)
+
+var (
+	/* Required Flags*/
+	EigenDADirectoryFlag = cli.StringFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "eigenda-directory"),
+		Usage:    "Address of the EigenDA directory contract, which points to all other EigenDA contract addresses.",
+		Required: true,
+		EnvVar:   common.PrefixEnvVar(envPrefix, "EIGENDA_DIRECTORY"),
+	}
```
samlaf: in other tools I've been trying to use a --network-id which has hardcoded directory addresses. We can make that change when we refactor all our tools into an `eigenda` cli that contains all other tools as subcommands.

anupsv (dismissed):

tools/gas_exhaustion_cert_meter/config.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	certVerifierCaller, err := certVerifierBinding.NewContractEigenDACertVerifierCaller(v3CertVerifierAddr, client)
+	if err != nil {
+		return nil, fmt.Errorf("bind to verifier contract at %s: %w", v3CertVerifierAddr.Hex(), err)
+	}
+
+	return &Config{
+		EthClient:               client,
+		OpStateRetrCaller:       oStateRetrCaller,
```
bxue-l2: opStateRetrCaller

tools/gas_exhaustion_cert_meter/flags/flags.go
bxue-l2: yeah, that is easier

tools/gas_exhaustion_cert_meter/flags/flags.go
bxue-l2: I am thinking to have it accept the calldata you can parse from Eth, which is typically hex coded. But we have to build a parse library first for the tools, so next time but good callout

samlaf: oh I see, file is storing the binary data and not the hex encoded.

samlaf (approved):

tools/gas_exhaustion_cert_meter/flags/flags.go
bxue-l2: a68bad35391effee475ef11c398c5d1d98fb2f0d

samlaf (approved):

PR #1926, opened 2025-08-27
fix(cert-verifier): add error codes so fct doesnt revert on invalid certs

samlaf:
SUPERSEDED BY https://github.com/Layr-Labs/eigenda/pull/1938

WIP: this is a first draft to show what I'm thinking.. needs works.

We use Steel which is not able to prove functions that revert (see https://github.com/risc0/risc0-ethereum/issues/438), so we need to turn all reverts (for invalid cert reasons!) into error codes.

This commit is a first step in that direction, adding error codes for abi.decoding faling, and for the call to verifySignature failing.

This endeavor to reach no reverts for any invalid cert (that needs to be proven so using Steel in order to be discarded from a rollup's pipeline) will still require a lot more careful work and audit of our library and downstream dependencies.

TODOs:
- Do we really need the custom errors + revertOnError fct (that we're not using)? See TODO in code
- check all TODOs added by this PR
- add a check for operatorlist length < maxNumOperators in registry coordinator (malicious cert could contain a huge number of fake operator addresses which could cause an out of gas)
- go over the checkSignatureFunction require statements in detail
- extend unit tests to check all status codes

### DIFF
--- a/contracts/src/integrations/cert/EigenDACertVerifier.sol
+++ b/contracts/src/integrations/cert/EigenDACertVerifier.sol
@@ -34,7 +34,7 @@ contract EigenDACertVerifier is
     bytes internal _quorumNumbersRequired;
 
     uint8 internal constant MAJOR_VERSION = 3;
-    uint8 internal constant MINOR_VERSION = 0;
+    uint8 internal constant MINOR_VERSION = 1;
     uint8 internal constant PATCH_VERSION = 0;
 
     constructor(
--- a/contracts/src/integrations/cert/EigenDACertVerifier.sol
+++ b/contracts/src/integrations/cert/EigenDACertVerifier.sol
@@ -54,9 +54,8 @@ contract EigenDACertVerifier is
     }
 
     /// @notice Decodes a certificate from bytes to an EigenDACertV3
-    /// @dev This function should not be called directly. It is exposes as external
-    //       for the purpose of try/catch'ing it inside checkDACert.
-    function _decodeCert(bytes calldata data) external pure returns (CT.EigenDACertV3 memory cert) {
+    /// @dev This function is external for the purpose of try/catch'ing it inside checkDACert.
+    function decodeCert(bytes calldata data) external pure returns (CT.EigenDACertV3 memory cert) {
         return abi.decode(data, (CT.EigenDACertV3));
     }
 
--- a/contracts/src/integrations/cert/EigenDACertVerifier.sol
+++ b/contracts/src/integrations/cert/EigenDACertVerifier.sol
@@ -16,6 +16,7 @@ import {EigenDATypesV2 as DATypesV2} from "src/core/libraries/v2/EigenDATypesV2.
 import {IEigenDASemVer} from "src/core/interfaces/IEigenDASemVer.sol";
 
 import {EigenDACertVerificationLib as CertLib} from "src/integrations/cert/libraries/EigenDACertVerificationLib.sol";
+import {EigenDACertTypes as CT} from "src/integrations/cert/EigenDACertTypes.sol";
 
 contract EigenDACertVerifier is
     IEigenDACertVerifier,
@@ -52,14 +53,25 @@ contract EigenDACertVerifier is
         _quorumNumbersRequired = initQuorumNumbersRequired;
     }
 
+    /// @notice Decodes a certificate from bytes to an EigenDACertV3
+    /// @dev This function should not be called directly. It is exposes as external
+    //       for the purpose of try/catch'ing it inside checkDACert.
+    function _decodeCert(bytes calldata data) external pure returns (CT.EigenDACertV3 memory cert) {
+        return abi.decode(data, (CT.EigenDACertV3));
+    }
+
     /// @inheritdoc IEigenDACertVerifierBase
     function checkDACert(bytes calldata abiEncodedCert) external view returns (uint8) {
+        CT.EigenDACertV3 memory daCert;
+
+        try this._decodeCert(abiEncodedCert) returns (CT.EigenDACertV3 memory cert) {
+            daCert = cert;
+        } catch {
+            return uint8(CertLib.StatusCode.CERT_DECODE_REVERT);
+        }
+
         (CertLib.StatusCode status,) = CertLib.checkDACert(
-            _eigenDAThresholdRegistry,
-            _eigenDASignatureVerifier,
-            abiEncodedCert,
-            _securityThresholds,
-            _quorumNumbersRequired
+            _eigenDAThresholdRegistry, _eigenDASignatureVerifier, daCert, _securityThresholds, _quorumNumbersRequired
         );
         return uint8(status);
     }
--- a/contracts/src/integrations/cert/EigenDACertVerifier.sol
+++ b/contracts/src/integrations/cert/EigenDACertVerifier.sol
@@ -14,6 +14,14 @@ import {IEigenDASemVer} from "src/core/interfaces/IEigenDASemVer.sol";
 import {EigenDACertVerificationLib as CertLib} from "src/integrations/cert/libraries/EigenDACertVerificationLib.sol";
 import {EigenDACertTypes as CT} from "src/integrations/cert/EigenDACertTypes.sol";
 
+/// @title EigenDACertVerifier
+/// @notice Verifies EigenDA certificates
+/// @dev This contract's checkDACert function is designed to be zk provable by risczero's Steel library,
+/// which does not support zk proving reverting calls: https://github.com/risc0/risc0-ethereum/issues/438.
+/// For this reason, we avoid using revert statements and instead return error codes.
+/// The only acceptable reverts are from bugs, such as contract misconfiguration.
+/// The goal is to be able to perfectly classify into 3 categories: valid certs, invalid certs, and bugs.
+/// Bugs could be configuration mistakes (_eigenDAThresholdRegistry points to 0x0), or other unexpected logic issues.
 contract EigenDACertVerifier is
     IEigenDACertVerifier,
     IEigenDACertVerifierBase,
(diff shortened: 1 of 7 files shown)


contracts/src/integrations/cert/libraries/EigenDACertVerificationLib.sol
```diff
@@ -44,19 +44,30 @@ library EigenDACertVerificationLib {
...
+        REQUIRED_QUORUMS_NOT_SUBSET, // Required quorums not a subset of blob quorums
+        CERT_DECODE_REVERT, // Certificate abi.decoding reverted
+        SIGNATURE_VERIFICATION_CALL_REVERT // External call to signature verifier reverted
     }
 
     /// @notice Decodes a certificate from bytes to an EigenDACertV3
-    function decodeCert(bytes calldata data) internal pure returns (CT.EigenDACertV3 memory cert) {
+    /// @dev This function should not be called directly. It is exposes as external
+    //       for the purpose of try/catch'ing it inside checkDACert.
+    function _decodeCert(bytes calldata data) external pure returns (CT.EigenDACertV3 memory cert) {
```
pakim249CAL: This should not be done because this is a library. 

The "external" tag has very different meanings in libraries vs contracts. In libraries, external means that an external library will be deployed that must be linked to the contract that is using it. This functionality is largely unsupported by many important tools, and usually results in a maintenance nightmare.

So the correct solution IMO is to define in the top level CertVerifier contract an EXTERNAL function with the same name, invoking the INTERNAL function in this library.

contracts/src/integrations/cert/libraries/EigenDACertVerificationLib.sol
```diff
@@ -220,23 +234,41 @@ library EigenDACertVerificationLib {
...
 
-        return (StatusCode.SUCCESS, "", confirmedQuorumsBitmap);
+            return (StatusCode.SUCCESS, "", confirmedQuorumsBitmap);
+        } catch Error(string memory reason) {
+            // This would match any require(..., "string reason") revert that is pre custom errors,
+            // which earlier versions of BLSSignatureChecker used, and might still be deployed. See:
+            // https://github.com/Layr-Labs/eigenlayer-middleware/blob/fe5834371caed60c1d26ab62b5519b0cbdcb42fa/src/BLSSignatureChecker.sol#L96
+            return (StatusCode.SIGNATURE_VERIFICATION_CALL_REVERT, reason, 0);
+        } catch (bytes memory reason) {
+            if (reason.length < 4) {
```
0xClandestine, line 284: Custom errors can have a length greater than 4 if they have parameters.

contracts/src/integrations/cert/libraries/EigenDACertVerificationLib.sol
0xClandestine: If this is going to be external/public would suggest removing the underscore (`_`).

0xClandestine: Also not sure the try/catch comment is 100% needed. This seams like a generally useful method.

contracts/src/integrations/cert/libraries/EigenDACertVerificationLib.sol
samlaf: This is catching NON custom errors. custom errors have length >= 4 right? As they have a function signature + optional params

0xClandestine: Oh sorry, that's correct misread.

samlaf: thanks for confirming. resolving this thread then.

contracts/src/integrations/cert/libraries/EigenDACertVerificationLib.sol
samlaf: ethen addressed both of these comments in [145890a](https://github.com/Layr-Labs/eigenda/pull/1926/commits/145890a243353bc1191fd664f11a4b5ffd35aaa7)

contracts/src/integrations/cert/EigenDACertVerifier.sol
```diff
@@ -16,14 +12,24 @@ import {EigenDATypesV2 as DATypesV2} from "src/core/libraries/v2/EigenDATypesV2.
...
 
 import {EigenDACertVerificationLib as CertLib} from "src/integrations/cert/libraries/EigenDACertVerificationLib.sol";
-
+import {EigenDACertTypes as CT} from "src/integrations/cert/EigenDACertTypes.sol";
+
+/// @title EigenDACertVerifier
+/// @notice Verifies EigenDA certificates
+/// @dev This contract's checkDACert function is designed to be zk provable by risczero's Steel library,
+/// which does not support zk proving reverting calls: https://github.com/risc0/risc0-ethereum/issues/438.
+/// For this reason, we avoid using revert statements and instead return error codes.
```
ethenotethan, line 21: also compatibility with optimistic rollups for fraud proof resolution as well

contracts/src/integrations/cert/EigenDACertVerifier.sol
```diff
@@ -34,7 +40,7 @@ contract EigenDACertVerifier is
     bytes internal _quorumNumbersRequired;
 
     uint8 internal constant MAJOR_VERSION = 3;
-    uint8 internal constant MINOR_VERSION = 0;
+    uint8 internal constant MINOR_VERSION = 1;
```
ethenotethan, line 43: are we gonna have a github release correspond to this semver bump?

contracts/src/integrations/cert/EigenDACertVerifier.sol
```diff
@@ -43,39 +49,77 @@ contract EigenDACertVerifier is
...
+        }
         _eigenDAThresholdRegistry = initEigenDAThresholdRegistry;
         _eigenDASignatureVerifier = initEigenDASignatureVerifier;
         _securityThresholds = initSecurityThresholds;
         _quorumNumbersRequired = initQuorumNumbersRequired;
     }
 
+    /// @notice Decodes a certificate from bytes to an EigenDACertV3
+    /// @dev This function is external for the purpose of try/catch'ing it inside checkDACert.
+    function decodeCert(
```
ethenotethan, line 69: dumb idea but any merit in deferring to this when decoding the cert in the clients lol?

contracts/src/integrations/cert/libraries/EigenDACertVerificationLib.sol
```diff
@@ -44,48 +55,56 @@ library EigenDACertVerificationLib {
     /// @param blobQuorumsBitmap The bitmap of blob quorums
     error RequiredQuorumsNotSubset(uint256 requiredQuorumsBitmap, uint256 blobQuorumsBitmap);
 
-    /// @notice Status codes for certificate verification results
+    /// @notice Thrown when certificate decoding fails
```
ethenotethan:
```suggestion
    /// @notice Thrown when certificate decoding from bytes into struct fails
```

contracts/src/integrations/cert/libraries/EigenDACertVerificationLib.sol
```diff
@@ -117,6 +137,13 @@ library EigenDACertVerificationLib {
             return (err, errParams);
         }
 
+        // We validate that the cert's blob_version is valid. Otherwise the getBlobParams call below
+        // would return a codingRate=0 which will cause a divide by 0 in checkSecurityParams.
```
ethenotethan:
```suggestion
        // would return a codingRate=0 which will cause a divide by 0 error in checkSecurityParams.
```

contracts/src/integrations/cert/libraries/EigenDACertVerificationLib.sol
```diff
@@ -220,27 +258,45 @@ library EigenDACertVerificationLib {
...
+            return (StatusCode.SIGNATURE_VERIFICATION_CALL_REVERT, bytes(reason), 0);
+        } catch (bytes memory reason) {
+            if (reason.length < 4) {
+                // We re-throw any non custom-error that was caught here. For example,
+                // low-level evm reverts such as out-of-gas don't return any data.
+                // See https://rareskills.io/post/try-catch-solidity#gdvnie-9-what-gets-returned-during-an-out-of-gas?
+                // These generally mean there is a bug in our implementation, which should be addressed by a human debugger.
+                // TODO: figure out whether we can programmatically deal with out of gas, since that might happen from
+                // a maliciously constructed cert.
+                revert(string(reason));
```
ethenotethan, line 291: worth noting that this would be a liveness halting case for rollup DP

contracts/src/integrations/cert/libraries/EigenDACertVerificationLib.sol
```diff
@@ -387,4 +466,48 @@ library EigenDACertVerificationLib {
...
+
+    /**
+     * @notice Converts an ordered array of quorum numbers into a bitmap.
+     * @param orderedQuorums The array of quorum numbers to convert/compress into a bitmap. Must be in strictly ascending order.
+     * @return err The status code indicating success or failure.
+     * @return errParams Additional parameters related to the error, if any.
+     * @return The resulting bitmap.
+     * @dev Each byte in the input is processed as indicating a single bit to flip in the bitmap.
+     * @dev This function returns an error status code if there >256 quorums, or the quorums are not ordered or contain duplicates.
+     */
```
ethenotethan, line 492: should we link to the reference implementation that we vendor from?

ethenotethan (commented):
coming together nicely - I'll allocate some time later today to work on tests and continue to analyze this 🙏🏻

contracts/src/integrations/cert/EigenDACertVerifier.sol
samlaf: Hmm maybe we should huh. Let's discuss this with rest of team.

contracts/src/integrations/cert/EigenDACertVerifier.sol
samlaf: I think (?) golang is doing that for free for us since we only rlp decode to the certV3 interface offchain: https://github.com/Layr-Labs/eigenda/blob/18a6b479a2a392b1154b8443b030c94bbb054a18/api/proxy/store/generated_key/v2/eigenda.go#L90

Are you seeing issues with this?

samlaf:
Closing in favor of https://github.com/Layr-Labs/eigenda/pull/1938

PR #1927, opened 2025-08-28
feat(contracts): Fix compilation bugs from Sam's PR and add unit test for the decoding

ethenotethan:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/contracts/src/integrations/cert/EigenDACertVerifier.sol
+++ b/contracts/src/integrations/cert/EigenDACertVerifier.sol
@@ -54,9 +54,8 @@ contract EigenDACertVerifier is
     }
 
     /// @notice Decodes a certificate from bytes to an EigenDACertV3
-    /// @dev This function should not be called directly. It is exposes as external
-    //       for the purpose of try/catch'ing it inside checkDACert.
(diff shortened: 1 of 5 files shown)


contracts/src/integrations/cert/EigenDACertVerifier.sol
```diff
@@ -52,14 +53,25 @@ contract EigenDACertVerifier is
         _quorumNumbersRequired = initQuorumNumbersRequired;
     }
 
+    /// @notice Decodes a certificate from bytes to an EigenDACertV3
+    /// @dev This function should not be called directly. It is exposes as external
+    //       for the purpose of try/catch'ing it inside checkDACert.
+    function _decodeCert(bytes calldata data) external pure returns (CT.EigenDACertV3 memory cert) {
```
samlaf:
```suggestion
    /// @dev This function is external for the purpose of try/catch'ing it inside checkDACert.
    function decodeCert(bytes calldata data) external pure returns (CT.EigenDACertV3 memory cert) {
```

PR #1910, opened 2025-08-22
feat(contracts): Support abigen v2 migration path

ethenotethan:
## Why are these changes needed?
Enabling abigen v2 as a onetime migration would require breaking changes across the eigenda backend and services. this provides a lightweight framework that allows us to start leveraging v2 bindings without requiring a large scale one time migration. 

**Risk:** This does allow for v1 && v2 bindings to live in dual state for the same contract artifact

**Changes**
- Overhauled the compile.sh script with better usability and support for abigen v1 and v2 targets
- Updated mise.toml to support geth dependency with abigen v2 support
- Generated v2 bindings for the `EigenDACertVerifier` and `EigenDACertVerifierRouter` contracts 

**Example Run**
```
No files changed, compilation skipped
🔧 abigen v1 → ./bindings/PaymentVault/binding.go (PaymentVault)
🔧 abigen v1 → ./bindings/SocketRegistry/binding.go (SocketRegistry)
🔧 abigen v1 → ./bindings/AVSDirectory/binding.go (AVSDirectory)
🔧 abigen v1 → ./bindings/DelegationManager/binding.go (DelegationManager)
🔧 abigen v1 → ./bindings/BitmapUtils/binding.go (BitmapUtils)
🔧 abigen v1 → ./bindings/OperatorStateRetriever/binding.go (OperatorStateRetriever)
🔧 abigen v1 → ./bindings/RegistryCoordinator/binding.go (RegistryCoordinator)
🔧 abigen v1 → ./bindings/BLSApkRegistry/binding.go (BLSApkRegistry)
🔧 abigen v1 → ./bindings/IIndexRegistry/binding.go (IIndexRegistry)
🔧 abigen v1 → ./bindings/StakeRegistry/binding.go (StakeRegistry)
🔧 abigen v1 → ./bindings/BN254/binding.go (BN254)
🔧 abigen v1 → ./bindings/EigenDAServiceManager/binding.go (EigenDAServiceManager)
🔧 abigen v1 → ./bindings/IEigenDAServiceManager/binding.go (IEigenDAServiceManager)
🔧 abigen v1 → ./bindings/EjectionManager/binding.go (EjectionManager)
🔧 abigen v1 → ./bindings/EigenDACertVerifierV1/binding.go (EigenDACertVerifierV1)
🔧 abigen v1 → ./bindings/EigenDACertVerifierV2/binding.go (EigenDACertVerifierV2)
[...]

PR #1933, opened 2025-08-29
feat(payments): Rearchitect how OnDemand vault updates are made

litt3:
- The old code for keeping in sync with the `PaymentVault` was requiring too many changes. It was easier to just reimplement with a better architecture
- This PR separates out ledger caching logic into a separate struct
- The ledger cache is now responsible for its own periodic updates, rather than requiring an external entity to do polling and updates


core/payments/ondemand/on_demand_ledger_cache.go
```diff
@@ -0,0 +1,183 @@
...
+	)
+	if err != nil {
+		return nil, fmt.Errorf("new on-demand vault monitor: %w", err)
+	}
+
+	return ledgerCache, nil
+}
+
+// Retrieves an existing OnDemandLedger for the given account, or creates a new one if it doesn't exist
+func (c *OnDemandLedgerCache) GetOrCreate(ctx context.Context, accountID gethcommon.Address) (*OnDemandLedger, error) {
```
cody-littley, line 131: Potential race condition:

- create ledger for client A and hold onto it for a while
- receive a lot of requests, cache is blasted away
- create another ledger for client A
- Two ledgers exist for client A, both ledgers are updated concurrently

core/payments/ondemand/on_demand_ledger_cache.go
```diff
@@ -0,0 +1,183 @@
...
+	// Fast path: check if ledger already exists in cache
+	if ledger, exists := c.cache.Get(accountID); exists {
+		return ledger, nil
+	}
+
+	// Slow path: acquire lock and check again
+	c.ledgerCreationLock.Lock()
+	defer c.ledgerCreationLock.Unlock()
```
cody-littley: Even if you decide to just document the race condition without fixing it, this might be a good place to use an index lock.

core/payments/ondemand/on_demand_vault_monitor.go
```diff
@@ -0,0 +1,114 @@
...
+
+	return nil
+}
+
+// Runs the background update loop to periodically consume updates made to the PaymentVault
+func (vm *OnDemandVaultMonitor) runUpdateLoop(ctx context.Context) {
+	ticker := time.NewTicker(vm.updateInterval)
+	defer ticker.Stop()
+
+	vm.logger.Infof("Starting OnDemandPaymentVault background update thread with updateInterval %v", vm.updateInterval)
```
cody-littley: nit, for logging things like "X is configured with value Y", I usually try to use debug level logging.

```suggestion
	vm.logger.Debugf("Starting OnDemandPaymentVault background update thread with updateInterval %v", vm.updateInterval)
```

cody-littley (commented):
LGTM other than the admittedly unlikely race condition. Would be nice to fix, but willing to settle for thorough documentation if no simple solution is available.

core/payments/ondemand/on_demand_vault_monitor.go
litt3: 2e846a3a6b57be0695fd038489f932701ea8245a

core/payments/ondemand/on_demand_ledger_cache.go
litt3: Good callout.

As we discussed offline, I think it's ok to leave this unaddressed for now, for the sake of simplicity.

```
// Note: there exists a potential race condition with the access pattern of this method:
// 1. A ledger is retrieved from the cache
// 2. A large amount of activity (or a small configured cache size) causes the ledger to be evicted from the cache
// before the ledger operation has been completed
// 3. A different caller tries to retrieve the ledger for that account, gets a cache miss, and constructs a new instance
//
// With this sequence of events, there could be multiple existing ledger instances for the same account. The
// underlying cumulative payment store isn't designed to function with multiple instantiated ledger structs, so the
// operation of one instance would overwrite the operation of the other. Practically, this would mean that the user
// would get one free dispersal. The multiple instance problem would resolve itself after a single operation, since
// the LRU cache can only maintain a single instance, and the other instance would be destroyed.
//
// It is very unlikely for this race condition to take place if the cache has been configured with a sane size. Given
// the low probability of the occurrence, and the low severity of the race condition, we are not addressing it right
// now to avoid the complexity of the potential workarounds.
```

10a4f75c4a8e93ece206a28ee424eff027509bb5

core/payments/ondemand/on_demand_ledger_cache.go
litt3: I hadn't wanted to roll a bespoke solution, and didn't recall you already had implemented an index lock. Nice

e6bd09fdde19e3f20daeac2e27d26f42b0a0413c

cody-littley (approved):

core/payments/ondemand/on_demand_ledger_cache.go
```diff
@@ -0,0 +1,199 @@
...
+// the low probability of the occurrence, and the low severity of the race condition, we are not addressing it right
+// now to avoid the complexity of the potential workarounds.
+func (c *OnDemandLedgerCache) GetOrCreate(ctx context.Context, accountID gethcommon.Address) (*OnDemandLedger, error) {
+	// Fast path: check if ledger already exists in cache
+	if ledger, exists := c.cache.Get(accountID); exists {
+		return ledger, nil
+	}
+
+	// Slow path: acquire per-account lock and check again
+	accountIndex := binary.BigEndian.Uint64(accountID.Bytes()[:8])
```
cody-littley, line 138: Do we validate that the account ID is a valid account ID before this point? What if a user passes in fewer than 8 bytes?

litt3: This is safe, since the input type `gethcommon.Address` has a fixed array size of 20 bytes. So no input can have fewer than 8 bytes

core/payments/vault/payment_vault.go
```diff
@@ -0,0 +1,91 @@
...
+	"github.com/Layr-Labs/eigenda/core/payments"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	"github.com/ethereum/go-ethereum/accounts/abi/bind"
+	gethcommon "github.com/ethereum/go-ethereum/common"
+)
+
+// Provides access to PaymentVault contract
+type paymentVault struct {
+	logger              logging.Logger
+	paymentVaultBinding *bindings.ContractPaymentVault
```
samlaf, line 22: recommend you switch to using abigen v2 if you have the time/energy. Ethen has made it easy to switch with his latest PR https://github.com/Layr-Labs/eigenda/pull/1910. You can check https://github.com/Layr-Labs/eigenda/pull/1911 for an example migration

core/payments/vault/payment_vault.go
```diff
@@ -0,0 +1,91 @@
...
+	onDemandPayment, err := pv.paymentVaultBinding.GetOnDemandTotalDeposit(&bind.CallOpts{Context: ctx}, accountID)
+	if err != nil {
+		return nil, fmt.Errorf("get on demand total deposit for account %v eth call: %w", accountID.Hex(), err)
+	}
+	return onDemandPayment, nil
+}
+
+// Retrieves the global symbols per second parameter
+func (pv *paymentVault) GetGlobalSymbolsPerSecond(ctx context.Context) (uint64, error) {
+	globalSymbolsPerSecond, err := pv.paymentVaultBinding.GlobalSymbolsPerPeriod(&bind.CallOpts{Context: ctx})
```
samlaf: I feel like we should start anchoring all our eth calls in RBN. Is that possible? Otherwise we'll get inconsistent results if/when we change a value on-chain. Whether we care about these small inconsistencies is another questino though.

core/payments/ondemand/on_demand_vault_monitor.go
```diff
@@ -0,0 +1,114 @@
...
+		cancelFunc:          cancel,
+	}
+
+	go monitor.runUpdateLoop(ctxWithCancel)
+	return monitor, nil
+}
+
+func (vm *OnDemandVaultMonitor) Stop() {
+	vm.cancelFunc()
+}
```
samlaf: unimportant nit: isn't it simpler to just pass in a cancellable context to the constructor directly?

core/payments/ondemand/on_demand_vault_monitor.go
```diff
@@ -0,0 +1,114 @@
...
+	defer cancel()
+
+	newDeposits, err := vm.paymentVault.GetTotalDeposits(ctxWithTimeout, accountIDs)
+	if err != nil {
+		return fmt.Errorf("get total deposits: %w", err)
+	}
+
+	if len(newDeposits) != len(accountIDs) {
+		// this shouldn't be possible
+		return fmt.Errorf("deposit count mismatch: got %d deposits for %d accounts", len(newDeposits), len(accountIDs))
```
samlaf, line 73: panic?

core/payments/ondemand/on_demand_vault_monitor.go
```diff
@@ -0,0 +1,114 @@
...
+	}
+
+	for i, newDeposit := range newDeposits {
+		accountID := accountIDs[i]
+		err := vm.updateTotalDeposit(accountID, newDeposit)
+		if err != nil {
+			vm.logger.Errorf("update total deposit for account %v failed: %v", accountID.Hex(), err)
+		}
+	}
```
samlaf, line 84: nit: feel a bit uneasy personally reading all these for loops. There's a hidden assumption that the updates are being done against a cache, but the structure of this monitor doesn't enforce that (it just takes arbitrary functions).

Might be worth using errgroups? Honestly I feel like we should use errgroups everywhere lol, they are simple, clean, and will deal with slow backends nicely if ever needed.

samlaf (approved):
Didn't read through the whole thing but I trust you. LGTM other than some nits

core/payments/vault/payment_vault.go
litt3: I think defining RBN for these calls is only useful if we fully support dynamically changing these values. Unfortunately, I don't think it will be trivial to implement this... I wrote out some thoughts in [this ticket](https://linear.app/eigenlabs/issue/EGDA-1980/consider-updates-to-global-payment-parameters). This is a similar problem space to when we designed the system for choosing the correct cert verifier.

Given that the proper solution will be complicated, and the old payment code didn't support (safe) dynamic modification of these parameters either, I chose to keep things simple for now. I think we should address this at some point, but I don't think it's a requirement for the MVP.

Thoughts?

core/payments/vault/payment_vault.go
litt3: 5f289b463cb94346a3e2ec3024829c827ef004dc

core/payments/ondemand/on_demand_vault_monitor.go
litt3: ffef733cb79ab64cbe4821558662da359cf3914c

core/payments/ondemand/on_demand_vault_monitor.go
litt3: I'm not sure if there might be some edge case where this happens, due to network or rpc failures.

I don't _think_ there is, but it seems possible that such an error, if it were to arise, would/could be transitory. So it seems reasonable to just return an error and try again later, rather than killing the disperser / validator where this is running

core/payments/ondemand/on_demand_vault_monitor.go
litt3: 315e055a9283ea75b7eafd2a0b6a8f8727baf712

I tried out errgroups in this commit. I'm not sure how I feel about it... it seems that errgroups are meant for cases where failure of one operation ought to cancel all other operations. But that's not what we have here.

This solution does buy us concurrency for the updates, but is it worth the ugliness? Not sure

litt3: Reverted the change, and left a comment as we discussed 40abd8d974a7f2fef093c92b980feb510f2166e7

cody-littley: Using an `errgroup` doesn't eliminate the for loop, it just adds concurrency. I think adding concurrency here where we technically don't need it might make the code harder to reason about, not easier. If there is concern about a performance problem waiting to happen, my suggestion is to add a comment like this:

```
// Expected implementation in production environments will be updating a local cache. If that ever changes
// to something like a DB update, serial execution here may become a performance hot spot. If that
// ever happens, consider doing this operation asynchronously.
```

core/payments/vault/payment_vault.go
samlaf: I see… using ethereum as a config store is really a PITA.

PR #1935, opened 2025-09-02
feat: added random access deque

cody-littley:

Adds a new utility data structure: a deque (double ended queue) that has O(1) random access time.

### DIFF
--- a/common/random_access_deque.go
+++ b/common/random_access_deque.go
@@ -1,0 +1,400 @@
+package common
+
+import (
+	"fmt"
+	"math"
+
+	"github.com/Layr-Labs/eigenda/common/enforce"
+)
+
+// The minimum initial capacity of a RandomAccessDeque.
+const minimumInitialCapacity = 32
+
+// A double-ended queue (deque) that supports O(1) lookup by index.
+//
+// - Insertion time: O(1) average, O(n) worst-case (when resizing is needed)
+// - Deletion time: O(1) average, array space is not reclaimed
+// - Lookup time by index: O(1)
+// - Iteration: O(1) to build iterator, O(1) per step
+//
+// This data structure is not thread safe.
+type RandomAccessDeque[T any] struct {
+	// The current number of elements in the deque.
+	size uint64
+	// Underlying data storage
+	data []T
+	// The index in data that corresponds to the logical start of the deque.
+	startIndex uint64
+	// The index in data that corresponds to the logical end of the deque (one past the last element).
+	endIndex uint64
+	// The initial capacity of the deque. Used when calling Clear().
+	initialCapacity uint64
+}
+
+// Create a new RandomAccessDeque with the specified initial capacity. Queue can grow beyond this capacity if needed.
+func NewRandomAccessDeque[T any](initialCapacity uint64) *RandomAccessDeque[T] {
+
+	if initialCapacity < minimumInitialCapacity {
+		initialCapacity = minimumInitialCapacity
+	}
+
+	return &RandomAccessDeque[T]{
+		data:            make([]T, initialCapacity),
+		startIndex:      0,
+		endIndex:        0,
+		initialCapacity: initialCapacity,
+	}
+}
+
+// Get the number of elements in the deque.
+//
+// O(1)
+func (s *RandomAccessDeque[T]) Size() uint64 {
+	return s.size
+}
+
+// Insert a value at the front of the deque. This value will have index 0 after insertion, and all other values will
+// have their indices increased by 1.
+//
+// O(1) average, O(n) worst-case (when resizing is needed)
+func (s *RandomAccessDeque[T]) PushFront(value T) {
+	s.resizeForInsertion()
+
+	if s.startIndex == 0 {
+		// wrap around
+		s.startIndex = uint64(len(s.data)) - 1
+	} else {
+		s.startIndex--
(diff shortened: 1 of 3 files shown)


common/random_access_deque.go
```diff
@@ -0,0 +1,400 @@
...
+// Create a new RandomAccessDeque with the specified initial capacity. Queue can grow beyond this capacity if needed.
+func NewRandomAccessDeque[T any](initialCapacity uint64) *RandomAccessDeque[T] {
+
+	if initialCapacity < minimumInitialCapacity {
+		initialCapacity = minimumInitialCapacity
+	}
+
+	return &RandomAccessDeque[T]{
+		data:            make([]T, initialCapacity),
+		startIndex:      0,
```
litt3: nit: you explicitly init `startIndex` and `endIndex`, but leave `size` to be implicitly initialized

common/random_access_deque.go
```diff
@@ -0,0 +1,400 @@
...
+//
+// O(1) to call this method, O(1) per iteration step.
+func (s *RandomAccessDeque[T]) ReverseIteratorFrom(index uint64) (func(yield func(uint64, T) bool), error) {
+
+	if index >= s.size {
+		return nil, fmt.Errorf("index %d out of bounds (size %d)", index, s.size)
+	}
+
+	return func(yield func(uint64, T) bool) {
+		for i := index; i != math.MaxUint64; i-- {
```
litt3, line 305: this is a weird way to define the condition. any particular reason?

common/random_access_deque.go
```diff
@@ -0,0 +1,400 @@
...
+	if index >= s.size {
+		return nil, fmt.Errorf("index %d out of bounds (size %d)", index, s.size)
+	}
+
+	return func(yield func(uint64, T) bool) {
+		for i := index; i < s.size; i++ {
+			value, err := s.Get(i)
+			enforce.NilError(err, "Get failed, did you modify the deque while iterating?!?")
+
+			yield(i, value)
```
litt3: My understanding is that you need to check the returned value and exit early if false

```
  if !yield(i, value) {
      return
  }
```

Also applies for your reverse iterator

common/random_access_deque.go
```diff
@@ -0,0 +1,400 @@
...
+//
+// The compare function `compare(a V, b T) int` should return:
+//   - negative value if a < b
+//   - zero if a == b
+//   - positive value if a > b
+//
+// If the deque is not sorted or if the ordering is not a total ordering, the return value is undefined. This function
+// is not defined as a method on RandomAccessDeque due to this fact. Not all RandomAccessDeque instances will be sorted,
+// and so this function is not always valid to call.
+func BinarySearchInOrderedDeque[V any, T any](
```
litt3, line 348: Seems there's a bug here, which is observable with the following test

```
func TestBinarySearchUnderflowBug(t *testing.T) {
	// This test demonstrates the uint64 underflow bug in BinarySearchInOrderedDeque
	// when searching for a value smaller than the first element in a 2-element deque
	
	deque := common.NewRandomAccessDeque[int](10)
	deque.PushBack(10)
	deque.PushBack(20)
	// Deque now contains: [10, 20]
	
	comparator := func(a int, b int) int {
		if a < b {
			return -1
		} else if a > b {
			return 1
		}
		return 0
	}
	
	// Search for value 5, which is smaller than all elements
	// This should return index=0, exact=false (insertion point before first element)
	// But due to the underflow bug, it will panic or return incorrect results
	index, exact := common.BinarySearchInOrderedDeque(deque, 5, comparator)
	
	// Expected: value 5 should be inserted at index 0
	require.False(t, exact, "Should not find exact match for 5")
	require.Equal(t, uint64(0), index, "Value 5 should be inserted at index 0")
}
```

It appears that copilot also identified the bug above

common/random_access_deque.go
```diff
@@ -0,0 +1,400 @@
...
+		if cmp == 0 {
+			// We've found an exact match.
+			return targetIndex, true
+		} else if cmp < 0 {
+			// value < target, search left half
+			//
+			//      value is here
+			//  |-----------------------|-----------------------|
+			// left                   target                  right
+			right = targetIndex - 1
```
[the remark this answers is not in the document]
cody-littley: fixed

common/random_access_deque.go
cody-littley: deleted the entries for `startIndex` and `endIndex`

common/random_access_deque.go
cody-littley: Required for iterating backwards over an unsigned type, since `0-- == math.MaxUint64`. Can you think of a better way of defining this?

common/random_access_deque.go
cody-littley: Done.

common/random_access_deque.go
cody-littley: Good cach. I committed this unit test and fixed the issue.

common/random_access_deque.go
litt3: Can't you just do `i >= 0` as your condition?

cody-littley: The last iteration will do 0--, which wraps around to the max uint64 value. All unigned values are `>= 0`, so we'd end up looping forever.

PR #1936, opened 2025-09-02
feat(mdbook): Add last updated preprocessor

ethenotethan:
Introduces the `mdbook-last-updated` preprocessor which injects a footer to each page with a timestamp and gh commit reference; eg:
Screenshot 2025-09-03 at 4 12 55 AM

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


samlaf (approved):
Nice

PR #1937, opened 2025-09-02
Use mainnet srs in tests

dmanc:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


tools/srs-utils/README.md
```diff
@@ -26,6 +27,11 @@ Note that the G2 point files (`g2.point` and `g2.trailing.point`) are twice the
 points require twice as many bytes to represent as G1 points in the BN254 curve. Each G1 point requires 32 bytes 
 of storage, while each G2 point requires 64 bytes.
 
+The `g2.point.powerOf2` file contains only G2 points at power-of-2 indices (indices 1, 2, 4, 8, 16, ..., 2^27). This
```
dmanc, line 30: Not 100% confident about this comment from claude, cc: @bxue-l2 can you check if it's accurate statement?

core/v2/serialization_test.go
```diff
@@ -52,8 +52,8 @@ func TestBlobKeyFromHeader(t *testing.T) {
 	}
 	blobKey, err := bh.BlobKey()
 	assert.NoError(t, err)
-	// e2fc52cb6213041838c20164eac05a7660b741518d5c14060e47c89ed3dd175b has verified in solidity  with chisel
-	assert.Equal(t, "e2fc52cb6213041838c20164eac05a7660b741518d5c14060e47c89ed3dd175b", blobKey.Hex())
+	// 3e1669312e6af52b75300f738bc5a32d92f760cad98394822a040366ff007520 has verified in solidity  with chisel
+	assert.Equal(t, "3e1669312e6af52b75300f738bc5a32d92f760cad98394822a040366ff007520", blobKey.Hex())
```
samlaf, line 56: why is this changing?

tools/srs-utils/README.md
samlaf, line 1: Are these changes actually used for inabox? If not, then should this be in a separate PR?

core/v2/serialization_test.go
dmanc: My guess is that the points are different so the blobkey ends up being different (commitment?)

tools/srs-utils/README.md
dmanc: It's used to replace this config value that's used in inabox

```
      G2_POWER_OF_2_PATH: ../resources/srs/g2.point.powerOf2
```

dmanc:
> LGTM. Bit confused as to how this is all working. Is the srs-utils somehow being used somewhere to download subsets of the SRS files? Or are inabox processes now loading the entire 8GiB or whatever SRS files at startup time (which would make them much slower)?

It's not loading the entire 8 GiB since all the tests are setting these values pretty low:
```
		SRSOrder:        3000,
		SRSNumberToLoad: 3000,
```

core/v2/serialization_test.go
samlaf: The srs points were different? I’m guessing it doesn’t matter, but why were they different? And where did they come from?

dmanc: IIRC the inabox SRS was generated before the mainnet ones and we just never updated the tests to use the new points. @bxue-l2 should have the historical context though.

tools/srs-utils/README.md
bxue-l2: more document here, https://github.com/Layr-Labs/eigenda/blob/aaaa54711ff032acc8c9f3d485a645a32e4a0e19/encoding/kzg/pointsIO.go#L120-L124

looks like it should start with 0

PR #1938, opened 2025-09-03
fix(cert-verifier): add try/catch to never revert and classify all checkDACert calls as valid/invalid/bug

samlaf:
Closes DAINT-708.
This PR is a (hopefully cleaner) rewrite of #1926.

We use Steel which is not able to prove functions that revert (see https://github.com/risc0/risc0-ethereum/issues/438), so we need to turn all reverts (for invalid cert reasons!) into error codes.

This endeavor to reach no reverts for any invalid cert (that needs to be proven so using Steel in order to be discarded from a rollup's pipeline) will still require a lot more careful work and audit of our library and downstream dependencies, but this PR at least provides the correct framework for how to think about these issues.

## Note

This PR flips the CertVerificationLib on its head, and makes it revert with custom errors instead of returning status codes like before. The CertVerifier itself now try/catches around the CertVerificationLib.checkDACert call, and maps custom errors to a `INVALID_CERT` status code, and panics/evm-reverts to `BUG` status code.

The previous low-level status codes are no longer valid, but I kept them around for backwards compatibility (don't need to change off-chain logic).

Also had to update the ServiceManager to expose `NextBlobVersion` to conform to the new IEigenDAThresholdManager interface needed by the CertVerifier. I hope this is ok @pakim249CAL 

## Offchain usage intention

The intention for the different status codes are:
- read path: discard both INVALID_CERT and BUG status codes to ---> discard cert (proxy 418 http response)
- write path: return 400 for INVALID_CERT (which will force batcher to redisperse the blob and get a new cert... assuming something weird went wrong) and map BUG to 503 failover signal, in order to let rollups with misconfigured CertVerifier contracts failover to another DA layer while they fix the bug.

Added this intention in the offchain code, but plan to only implement the failover signal as part of https://github.com/Layr-Labs/eigenda/pull/1924


contracts/src/integrations/cert/EigenDACertVerifier.sol
```diff
@@ -34,9 +44,25 @@ contract EigenDACertVerifier is
...
 
     uint8 internal constant MAJOR_VERSION = 3;
-    uint8 internal constant MINOR_VERSION = 0;
+    uint8 internal constant MINOR_VERSION = 1;
     uint8 internal constant PATCH_VERSION = 0;
 
+    /// @notice Status codes for certificate verification results
+    enum StatusCode {
+        NULL_ERROR, // Unused error code. If this is returned, there is a bug in the code.
+        SUCCESS, // Verification succeeded
```
pakim249CAL: I don't think we should keep around the historical codes. The only backwards compatibility we should conserve is:

* 1 = success
* Any other code = failure.

contracts/src/integrations/cert/EigenDACertVerifier.sol
```diff
@@ -46,22 +72,76 @@ contract EigenDACertVerifier is
...
+        }
         _eigenDAThresholdRegistry = initEigenDAThresholdRegistry;
         _eigenDASignatureVerifier = initEigenDASignatureVerifier;
         _securityThresholds = initSecurityThresholds;
         _quorumNumbersRequired = initQuorumNumbersRequired;
     }
 
+    /// @notice Decodes a certificate from bytes to an EigenDACertV3
+    /// @dev This function is external for the purpose of try/catch'ing it inside checkDACert.
+    function decodeCert(bytes calldata data) external pure returns (CT.EigenDACertV3 memory cert) {
```
pakim249CAL: Because this is not meant to be exposed, I would prefix this with an underscore. This should make it more clear to people that find this function that it is not meant to be called directly, and they should not rely on it as part of the cert verifier interface.

contracts/src/integrations/cert/EigenDACertVerifier.sol
samlaf: This will make it easier to not have to worry about breaking changes on the golang code.
We want to map BUG to a failover signal, but if we get rid of old status codes, then BUG would be enum value 3 in new contracts, which is someone is using the certVerifier 3.0.0 is some other random status code.

So that creates some complications and we'd have to start using the semver and bump certverifier to 4.0.0. It's doable but don't think its worth the trouble? WDYT?

contracts/src/integrations/cert/EigenDACertVerifier.sol
samlaf: I had done that originally but @0xClandestine said that it could be called by someone offchain, as it was a perfectly valid function. I'm fine with either approach.
https://github.com/Layr-Labs/eigenda/pull/1926#discussion_r2305676682

pakim249CAL (commented):
As I look more at this, I'm starting to think that this really just ends up being a sort of remapping from solidity error types to a different categorization of error types. I ultimately do still question whether this belongs in the CertVerifier itself, or whether it can just be done outside of the scope of the CertVerifier contract. 

But I do think that it is good for the library to not handle the status codes anymore, and if we ever need a simpler or different entry point to a CertVerifier, it should not be too difficult to add. So overall I think we can continue with this approach.

contracts/src/core/interfaces/IEigenDAThresholdRegistry.sol
```diff
@@ -44,6 +44,10 @@ interface IEigenDAThresholdRegistry {
 
     ///////////////////////// V2 ///////////////////////////////
 
+    /// @notice Returns the next blob version
+    /// @dev Can be called before calling getBlobParams to verify that an input blobVersion actually exists
+    function nextBlobVersion() external view returns (uint16);
```
ethenotethan, line 49: is this getting called anywhere?

contracts/src/integrations/cert/EigenDACertVerifier.sol
```diff
@@ -34,9 +44,25 @@ contract EigenDACertVerifier is
...
     uint8 internal constant MAJOR_VERSION = 3;
-    uint8 internal constant MINOR_VERSION = 0;
+    uint8 internal constant MINOR_VERSION = 1;
     uint8 internal constant PATCH_VERSION = 0;
 
+    /// @notice Status codes for certificate verification results
+    enum StatusCode {
+        NULL_ERROR, // Unused error code. If this is returned, there is a bug in the code.
+        SUCCESS, // Verification succeeded
+        // The below 4 status codes are kept for backwards compatibility, but are no longer used.
```
ethenotethan, line 49: nit let's specify that these codes are mapped to semver version `3.0.0`

contracts/src/integrations/cert/EigenDACertVerifier.sol
```diff
@@ -46,22 +72,76 @@ contract EigenDACertVerifier is
...
+    /// @dev This function is external for the purpose of try/catch'ing it inside checkDACert.
+    function decodeCert(bytes calldata data) external pure returns (CT.EigenDACertV3 memory cert) {
+        return abi.decode(data, (CT.EigenDACertV3));
+    }
+
     /// @inheritdoc IEigenDACertVerifierBase
+    /// @dev This function try catches checkDACertReverts, and maps any reverts to status codes.
+    /// @dev Make sure to call this at a block number that is > RBN, otherwise this function will
+    /// return an INVALID_CERT status code because of a require in the BLSSignatureChecker library that we use.
+    /// TODO: we should return (uint8, bytes) instead and include the revert reason.
```
ethenotethan: should we file a linear ticket for this?

contracts/src/integrations/cert/EigenDACertVerifier.sol
```diff
@@ -46,22 +72,76 @@ contract EigenDACertVerifier is
...
+            // This matches any require(..., "string reason") revert that is pre custom errors,
+            // which many of our current eigenlayer-middleware dependencies like the BLSSignatureChecker still use. See:
+            // https://github.com/Layr-Labs/eigenlayer-middleware/blob/fe5834371caed60c1d26ab62b5519b0cbdcb42fa/src/BLSSignatureChecker.sol#L96
+            return uint8(StatusCode.INVALID_CERT);
+        } catch Panic(uint256) /*errorCode*/ {
+            // This matches any panic (e.g. arithmetic overflow, division by zero, invalid array access, etc.),
+            // which means a bug or misconfiguration of the CertVerifier contract itself.
+            return uint8(StatusCode.BUG);
+        } catch (bytes memory reason) {
+            if (reason.length == 0) {
```
ethenotethan, line 125: is this conditional still necessary?

contracts/src/integrations/cert/EigenDACertVerifier.sol
ethenotethan: I'd argue we should wait until the cert representation changes before having to introduce a major semver bump

contracts/src/integrations/cert/EigenDACertVerifier.sol
```diff
@@ -16,14 +16,24 @@ import {EigenDATypesV2 as DATypesV2} from "src/core/libraries/v2/EigenDATypesV2.
...
 import {EigenDACertVerificationLib as CertLib} from "src/integrations/cert/libraries/EigenDACertVerificationLib.sol";
-
+import {EigenDACertTypes as CT} from "src/integrations/cert/EigenDACertTypes.sol";
+
+/// @title EigenDACertVerifier
+/// @notice Verifies EigenDA certificates
+/// @dev This contract's checkDACert function is designed to be zk provable by risczero's Steel library,
+/// which does not support zk proving reverting calls: https://github.com/risc0/risc0-ethereum/issues/438.
+/// For this reason, we avoid using revert statements and instead return error codes.
+/// The only acceptable reverts are from bugs, such as contract misconfiguration, which require human intervention.
```
ethenotethan: nit - not true anymore

ethenotethan (commented):
this is 100x easier to reason about than the path we were going down before. Mostly LGTM with some smol comments

contracts/src/core/interfaces/IEigenDAThresholdRegistry.sol
samlaf: yes right here: https://github.com/Layr-Labs/eigenda/pull/1938/files#diff-571cbaa0087430087da84c9aaa19531041ccec8435c9a5e10eee16d415b55ce7R159

contracts/src/integrations/cert/EigenDACertVerifier.sol
samlaf: no but I added it to make it cleaner. Since length==0 maps to exact case thatn I know of, whereas length=1..4 I have no idea whether its even possible.

contracts/src/integrations/cert/EigenDACertVerifier.sol
samlaf: https://linear.app/eigenlabs/issue/DAINT-733/checkdacert-should-return-bytes-reason-along-with-status-code

contracts/src/integrations/cert/EigenDACertVerifier.sol
samlaf: [45f5c15](https://github.com/Layr-Labs/eigenda/pull/1938/commits/45f5c15309febcfeb600b0340d87d2d3a718863b)

contracts/src/integrations/cert/EigenDACertVerifier.sol
samlaf: [65ba385](https://github.com/Layr-Labs/eigenda/pull/1938/commits/65ba385e44945de204097b0171f2e5d48f4c11d0)

contracts/src/integrations/cert/EigenDACertVerifier.sol
samlaf: Created a 4.0.0 milestone on linear with all the breaking changes we should include when we bump: https://linear.app/eigenlabs/project/integrations-contracts-47228b81a260/issues?projectMilestoneId=a033c227-74da-44cc-9a12-9bfe57d7b600

samlaf:
> As I look more at this, I'm starting to think that this really just ends up being a sort of remapping from solidity error types to a different categorization of error types. I ultimately do still question whether this belongs in the CertVerifier itself, or whether it can just be done outside of the scope of the CertVerifier contract.
> 
> But I do think that it is good for the library to not handle the status codes anymore, and if we ever need a simpler or different entry point to a CertVerifier, it should not be too difficult to add. So overall I think we can continue with this approach.

I totally hear you. We can keep thinking about this. Our reasoning is that right now literally ALL of our usecases need this. And some rollups already plan to deploy the CertVerifier directly without a router. Not having it as ready-to-go would require a lot of handholding, education, pointing them to the right docs, having this try/catch example somewhere for them to copy-paste, and would ultimately imo lead to more misconfigurations by our clients. Better to KISS for now imo.

contracts/src/core/interfaces/IEigenDAThresholdRegistry.sol
ethenotethan: ah I see:
```
        uint16 nextBlobVersion = eigenDAThresholdRegistry.nextBlobVersion();
        if (blobVersion >= nextBlobVersion) {
            revert InvalidBlobVersion(blobVersion, nextBlobVersion);
        }
```

so what happens to the cert validity when `nextBlobVersion+=1`? If it was once failing on this check and the blob version updates, could it pass?

ethenotethan: I don't this changing matters for zk proving but for optimistic rollups it might

contracts/src/integrations/cert/EigenDACertVerifier.sol
ethenotethan: roger that makes sense

contracts/src/core/interfaces/IEigenDAThresholdRegistry.sol
samlaf: That's a good point........ another reason why anchoring calls @ RBN is important. But we don't know when blobVersion is introduced huh...

contracts/src/integrations/cert/libraries/EigenDACertVerificationLib.sol
```diff
@@ -44,48 +44,35 @@ library EigenDACertVerificationLib {
...
-    }
-
-    /// @notice Decodes a certificate from bytes to an EigenDACertV3
-    function decodeCert(bytes calldata data) internal pure returns (CT.EigenDACertV3 memory cert) {
-        return abi.decode(data, (CT.EigenDACertV3));
-    }
+    /// @notice Thrown when the blob version is invalid (doesn't exist in the ThresholdRegistry contract)
+    /// @param blobVersion The invalid blob version
+    /// @param nextBlobVersion The next blob version (valid versions need to be less than this number)
+    error InvalidBlobVersion(uint16 blobVersion, uint16 nextBlobVersion);
```
bxue-l2, line 50: As we talked in the FFT, invalid blob version shouldn't exist. 

When it is signed by operator nodes, they already checked it exists.

When it does not exist, the operators won't sign... unless they are colluding. Then we need blobParams is anchored against RBN

bxue-l2: but it is good we checked. The attack would be

Malicious opr signs a wrong blobParams, L2 node reject. Later blob params is added. But it will leave a trace because RBN is lower than new blobParams addition time

bxue-l2 (approved):

contracts/src/integrations/cert/libraries/EigenDACertVerificationLib.sol
samlaf: Problem is that blobParams are not tied to RBNs...they are just appended to a mapping whenever: https://github.com/Layr-Labs/eigenda/blob/6cd192ecbe5f0abfe73fc08df306cf00e32ef010/contracts/src/core/EigenDAThresholdRegistryStorage.sol#L26

contracts/src/integrations/cert/EigenDACertVerifier.sol
```diff
@@ -34,9 +38,26 @@ contract EigenDACertVerifier is
...
+        // The below 4 status codes are kept for backwards compatibility, but are no longer used.
+        // We previously had plans to have more granular error codes, but decided this was not necessary,
+        // and the only signal useful to offchain is to separate certs into: success, invalid (400), and bugs (500).
+        // TODO(4.0.0): get rid of these
+        UNUSED_HISTORICAL_INVALID_INCLUSION_PROOF,
+        UNUSED_HISTORICAL_SECURITY_ASSUMPTIONS_NOT_MET,
+        UNUSED_HISTORICAL_BLOB_QUORUMS_NOT_SUBSET,
+        UNUSED_HISTORICAL_REQUIRED_QUORUMS_NOT_SUBSET,
+        INVALID_CERT, // Certificate is invalid due to some revert from the verification library
+        BUG // Bug or misconfiguration in the CertVerifier contract itself. This includes solidity panics and evm reverts.
```
samlaf: TODO: rename to INTERNAL_ERROR

contracts/src/integrations/cert/libraries/EigenDACertVerificationLib.sol
samlaf: Created https://linear.app/eigenlabs/issue/DAINT-738/design-checkpointed-blobparams-or-hardcode-blobversion-in-certverifier to track this issue and potential solutions

contracts/src/core/interfaces/IEigenDAThresholdRegistry.sol
samlaf: Closing this thread and moving discussion to the more specific https://github.com/Layr-Labs/eigenda/pull/1938#discussion_r2323631729

contracts/src/integrations/cert/EigenDACertVerifier.sol
samlaf: [3adc82c](https://github.com/Layr-Labs/eigenda/pull/1938/commits/3adc82c2c6c44ccc3f1eac75cb0145896badee09)

contracts/src/integrations/cert/EigenDACertVerifier.sol
samlaf: Following our discussion this morning updated with the underscore prefix: [3ab00c2](https://github.com/Layr-Labs/eigenda/pull/1938/commits/3ab00c288a4bb2bc65435cb9ac4008b10c1299b1)

bxue-l2 (approved):
lgtm

pakim249CAL (approved):

PR #1939, opened 2025-09-03
feat: Use testcontainers for deploying localstack

dmanc:

Standarize on testcontainers for docker based tests. Some motivation behind it:
- Testcontainers has official module for localstack https://golang.testcontainers.org/modules/localstack/.
- Proxy uses testcontainers for deploying resources like minio which also has an official module https://golang.testcontainers.org/modules/minio/
- Testcontainers automatically spins up a reaper process called "Ryuk" for test cleanup https://github.com/testcontainers/moby-ryuk
- Configurable wait strategies and docker networking setup (will be useful as we continue to migrate components of inabox)

In addition this PR also does the following
- Delete inabox/deploy/localstack.go so that we can untangle the tests from depending on inabox package.
- Each unit test package deploys its own localstack rather than deploying a single localstack at the start through the test.sh file.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


common/aws/test/client_test.go
```diff
@@ -184,7 +179,11 @@ func TestReadNonExistentValue(t *testing.T) {
 
 		client, err := builder.build()
 		assert.NoError(t, err)
-		ReadNonExistentValueTest(t, client)
+		_, err = client.FragmentedDownloadObject(context.Background(), bucket, "nonexistent", 1000, 1000)
+		assert.Error(t, err)
```
cody-littley: Nit, I personally prefer using `require` instead of `assert`, e.g. `require.Error(t, err)`. The problem with `assert` is that it doesn't immediately stop the test. And when there is a first error, usually that first error is followed by many following errors. This ends up cluttering the output, and in practice we usually only care about the first failure.

testbed/deploy_resources.go
```diff
@@ -0,0 +1,197 @@
...
+	}
+
+	// Create metadata table
+	if config.MetadataTableName != "" {
+		_, err := test_utils.CreateTable(ctx, cfg, config.MetadataTableName,
+			blobstore.GenerateTableSchema(config.MetadataTableName, 10, 10))
+		if err != nil {
+			return fmt.Errorf("failed to create metadata table %s: %w", config.MetadataTableName, err)
+		}
+		fmt.Printf("Created metadata table: %s\n", config.MetadataTableName)
```
cody-littley: Would it be possible to use a logger instead of `fmt.Printf` in this file? If you decide to accept this suggestion, then I'd suggest searching for all prints in this PR and replacing them (not going to comment each location 😅).

cody-littley (approved):

common/aws/test/client_test.go
dmanc: 4e09b0586f1084f24739b2952f7a77a5978072f7, only changed the asserts -> require for the files inside this PR, otherwise it's gonna be a lot of changes.

testbed/deploy_resources.go
dmanc: a4f70bebbd6e5d2c05b28f5019d8058a6d10714a

cody-littley (approved):

samlaf (approved):

PR #1941, opened 2025-09-04
chore: remove testnet ci

pakim249CAL:

remove holesky testnet from CI. Reservation was yanked by us. It broke as expected and we should stop using it.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


samlaf (approved):
> What is the entire thing? What should I delete?

The makefile command, and all the test code that its triggering.
We can do as a follow-up though, happy to merge this to get rid of the big X checkmark on our CI.

PR #1945, opened 2025-09-04
feat(metering): Implement on demand meterer

litt3:
closes EGDA-1877

### DIFF
--- a/core/meterer/on_demand_meterer.go
+++ b/core/meterer/on_demand_meterer.go
@@ -1,0 +1,58 @@
+package meterer
+
+import (
+	"fmt"
+	"time"
+
+	"golang.org/x/time/rate"
+)
+
+// OnDemandMeterer handles global throughput rate limiting for on-demand payments.
+// It ensures that the global maximum throughput is observed across all on-demand dispersals.
+//
+// This struct is safe for use by multiple goroutines.
+type OnDemandMeterer struct {
+	limiter *rate.Limiter
+	getNow  func() time.Time
+}
+
+// Creates a new OnDemandMeterer with the specified rate limiting parameters.
+func NewOnDemandMeterer(
+	globalSymbolsPerSecond uint64,
+	globalRatePeriodInterval uint64,
+	getNow func() time.Time,
+) *OnDemandMeterer {
+	burstSize := int(globalSymbolsPerSecond * globalRatePeriodInterval)
+	limiter := rate.NewLimiter(rate.Limit(globalSymbolsPerSecond), burstSize)
+
+	return &OnDemandMeterer{
+		limiter: limiter,
+		getNow:  getNow,
+	}
+}
+
+// Reserves tokens for a dispersal with the given number of symbols.
+//
+// Returns a reservation that can be cancelled if the dispersal is not performed (e.g., if payment verification fails).
+// The reservation will automatically take effect if not cancelled.
+//
+// This method only succeeds if tokens are immediately available (no queueing/waiting). If a reservation is returned,
+// it is safe to proceed with dispersal without checking the delay.
+func (m *OnDemandMeterer) MeterDispersal(symbolCount uint32) (*rate.Reservation, error) {
+	reservation := m.limiter.ReserveN(m.getNow(), int(symbolCount))
+
+	if !reservation.OK() || reservation.Delay() > 0 {
+		reservation.Cancel()
+		return nil, fmt.Errorf("global rate limit exceeded: cannot reserve %d symbols", symbolCount)
+	}
+
+	return reservation, nil
+}
+
+// Cancels a reservation obtained by MeterDispersal, returning tokens to the rate limiter.
+// This should be called when a reserved dispersal will not be performed (e.g., payment verification failed).
+func (m *OnDemandMeterer) CancelDispersal(reservation *rate.Reservation) {
+	if reservation != nil {
+		reservation.Cancel()
+	}
+}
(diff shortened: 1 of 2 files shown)


core/meterer/on_demand_meterer_test.go
```diff
@@ -0,0 +1,33 @@
...
+	timeSource := func() time.Time { return startTime }
+	meterer := NewOnDemandMeterer(100, 10, timeSource)
+
+	reservation, err := meterer.MeterDispersal(500)
+	require.NoError(t, err)
+	require.NotNil(t, reservation)
+
+	require.NotPanics(t, func() {
+		meterer.CancelDispersal(reservation)
+	})
```
cody-littley: Nit, is there a point to wrapping it like this? If it panics, the test will fail, so I'm not sure why you specifically assert this.

cody-littley (approved):

core/meterer/on_demand_meterer_test.go
litt3: True. 3d1c8fd4e9246119f3bfc29918d8df3e717696f6

cody-littley (dismissed):

core/meterer/on_demand_meterer.go
```diff
@@ -0,0 +1,58 @@
+package meterer
+
+import (
+	"fmt"
+	"time"
+
+	"golang.org/x/time/rate"
+)
+
+// OnDemandMeterer handles global throughput rate limiting for on-demand payments.
```
samlaf, line 10: nit: only global b/c they all have to go through our disperser right? maybe we can use "enforce disperser-assigned on-demand throughput rate limits" instead? Which would keep being true if/when we decide to assign on-demand throughput to different dispersals?

core/meterer/on_demand_meterer.go
```diff
@@ -0,0 +1,58 @@
...
+	getNow  func() time.Time
+}
+
+// Creates a new OnDemandMeterer with the specified rate limiting parameters.
+func NewOnDemandMeterer(
+	globalSymbolsPerSecond uint64,
+	globalRatePeriodInterval uint64,
+	getNow func() time.Time,
+) *OnDemandMeterer {
+	burstSize := int(globalSymbolsPerSecond * globalRatePeriodInterval)
```
samlaf, line 25: This is probably a huge burst right? Can we even handle such a burst? Should we set this to something smaller?

core/meterer/on_demand_meterer.go
```diff
@@ -0,0 +1,58 @@
...
+
+	return reservation, nil
+}
+
+// Cancels a reservation obtained by MeterDispersal, returning tokens to the rate limiter.
+// This should be called when a reserved dispersal will not be performed (e.g., payment verification failed).
+func (m *OnDemandMeterer) CancelDispersal(reservation *rate.Reservation) {
+	if reservation != nil {
+		reservation.Cancel()
+	}
```
samlaf: is this a defensive check or are you purposefully intending to pass nils in sometimes?
Issue with these nil checks is that if there's a bug in the code that's passing a nil reservation itll just quitely chug along, and we'll never know. If this is not needed, I'd prefer to get rid of the nil check and just let it panic hard if a nil is passed.

core/meterer/on_demand_meterer.go
litt3: This mirrors how it's currently being done, albeit with a token bucket instead of a bin (so this is actually less permissive, due the "double burst" possibility with a bin algorithm)

globalSymbolsPerSecond = 131072
globalRatePeriodInterval = 30

131072 symbols * 30 seconds = 120 mb burst size

I don't know, is that too big? Doesn't seem crazy to me

core/meterer/on_demand_meterer.go
litt3: > only global b/c they all have to go through our disperser right?

Right. I think it makes most sense to keep this documentation explicitly applicable to the current system, for the sake of clarity.

There will need to be changes to this struct if / when on-demand dispersals are possible through any disperser other than our own. Since it's not clear what form those changes will take, I don't think it's too important to make this doc "forward compatible"

core/meterer/on_demand_meterer.go
litt3: I will say, I don't really see a reason that these values need to be on-chain. So I don't think there would be any issues, if we just decide to configure some different value

core/meterer/on_demand_meterer.go
litt3: Good point. Defensive check removed

dbc9702299176f0c7b0ff121db5980f3ae27e3b0

core/meterer/on_demand_meterer.go
samlaf: Oh yeah not as bad as I would have thought.
But iiuc this `globalRatePeriodInterval` was really only meaningful pre token bucket right?
We prob shouldn't be using it in this implementation? Feels even more arbitrary to me iiuc

samlaf (approved):
Approving to unblock but still unsure about https://github.com/Layr-Labs/eigenda/pull/1945#discussion_r2330502064

core/meterer/on_demand_meterer.go
litt3: I think that  `globalRatePeriodInterval` was *probably* originally intended just to specify the width of reservation bins, but was also used for the on-demand metering simply because the value was available.

I think its usage here is no more arbitrary than the previous usage in the context of on-demand metering. I opted for "arbitrary, but maximally consistent with previous implementation", for the sake of expediency. 

I'll raise the question in slack, to see if we want to choose a different value

PR #1946, opened 2025-09-04
feat(payments): add proxy flags to expose client ledger functionality

litt3:
- Needed to be able to configure the `ClientLedger`
- Does some reorganization with the config constants, to enable a better way of defining config flags


api/proxy/config/consts/const.go
```diff
@@ -0,0 +1,21 @@
...
+	MemstoreFlagsCategory   = "Memstore (for testing purposes - replaces EigenDA backend)"
+	StorageFlagsCategory    = "Storage"
+	S3Category              = "S3 Cache/Fallback"
+	VerifierCategory        = "Cert Verifier (V1 only)"
+	KZGCategory             = "KZG"
+	ProxyServerCategory     = "Proxy Server"
+	PaymentsCategory        = "Payments"
+
+	DeprecatedRedisCategory = "Redis Cache/Fallback"
+
```
samlaf: why do these need to live here? 
These categories are urfave CLI categories, so only elated to flags, not really config.
Can we keep them in flags.go or was there some other reasoning for reusing them?

samlaf: oh I see you're using them directly when defining the flags... hmmm...
can we maybe rename them to `VerifierFlagCategory` or something in that case?

api/proxy/config/v2/eigendaflags/cli.go
```diff
@@ -37,6 +39,23 @@ var (
 	NetworkFlagName                 = withFlagPrefix("network")
 	RBNRecencyWindowSizeFlagName    = withFlagPrefix("rbn-recency-window-size")
 	RelayConnectionPoolSizeFlagName = withFlagPrefix("relay-connection-pool-size")
+
+	ClientLedgerModeFlag = &cli.StringFlag{
+		Name:     withFlagPrefix("client-ledger-mode"),
+		Usage:    "Payment mode for the client. Options: 'legacy', 'reservation-only', 'on-demand-only', 'reservation-and-on-demand'.",
```
samlaf: can you define what legacy means given that its default

api/proxy/config/v2/eigendaflags/cli.go
```diff
@@ -37,6 +39,23 @@ var (
...
+		Value:    "legacy",
+		EnvVars:  []string{withEnvPrefix(consts.GlobalEnvVarPrefix, "CLIENT_LEDGER_MODE")},
+		Category: consts.PaymentsCategory,
+		Required: false,
+	}
+	PaymentVaultMonitorIntervalFlag = &cli.DurationFlag{
+		Name:     withFlagPrefix("payment-vault-monitor-interval"),
+		Usage:    "Interval for monitoring payment vault updates.",
```
samlaf: can you describe this a bit more, given that this will be read by rollup teams who are not familiar with our clietns? just mention that it keeps track of on-demand deposits and reservation statuses?

api/proxy/config/consts/const.go
litt3: Right, I had to move these out of `flags.go` into a separate package to avoid the import cycle.

One option would be to move the category name constants directly into the various `cli.go` files, and just remove the `category` parameter from the `CLIFlags` functions. This would be cleaner, but a substantially larger diff.

Thoughts?

api/proxy/config/v2/eigendaflags/cli.go
litt3: 3c4d13907f6240eccdadd625d12f8c6422c9a415

api/proxy/config/v2/eigendaflags/cli.go
litt3: edff87f5d01f2f597111bc398c379ebdade8ca8b

api/proxy/config/consts/const.go
cody-littley: Not relevant to this PR, but I've got some ideas for how we could massively clean up config cruft, to the point where we won't even need these sorts of constants in the first place. We should discuss how that might look for the proxy.

cody-littley (approved):

api/proxy/config/consts/const.go
litt3: I started down the path of the larger refactor, but it got too big. I pulled the plug, and just refactored the new fields to be in the same style as the existing fields.

PR #1947, opened 2025-09-05
docs: change from insecure to trusted

bxue-l2:

changed name from insecure -> trusted

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/docs/spec/src/integration/spec/6-secure-integration.md
+++ b/docs/spec/src/integration/spec/6-secure-integration.md
@@ -219,7 +219,7 @@ There is a synchronization risk that can temporarily cause dispersals to fail wh
 
 ### Rollup Stack Secure Integrations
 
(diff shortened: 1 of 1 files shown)


docs/spec/src/integration/spec/6-secure-integration.md
```diff
@@ -219,7 +219,7 @@ There is a synchronization risk that can temporarily cause dispersals to fail wh
 
 ### Rollup Stack Secure Integrations
```
ian-shim, line 220: should we call it trustless here

bxue-l2: I often use the language (secure + trusted). Trustless doesn't convey the sense of security, and that is my problem with that

PR #1949, opened 2025-09-05
docs: explain security thresholds

Lulu-Zhou-EigenLabs:
1. Add comments to explain how we check that the security threshold is satisfied in the CertVerifier.
2. Create documentation for security thresholds and blob parameters.


contracts/src/integrations/cert/EigenDACertVerifier.sol
```diff
@@ -29,6 +29,12 @@ contract EigenDACertVerifier is
 
     IEigenDASignatureVerifier internal immutable _eigenDASignatureVerifier;
 
+    /// @notice Default security thresholds used by {checkDACert}.
+    /// @dev Checked inside {EigenDACertVerificationLib-checkDACert}. Changing this default
+    ///      at deployment affects verification outcomes. Constraints to respect:
```
samlaf: There is no default specified in this contract. It's just an internal variable that is populated by the constructor. The "defaults" are injected by our deploy script

contracts/src/integrations/cert/EigenDACertVerifier.sol
```diff
@@ -29,6 +29,12 @@ contract EigenDACertVerifier is
...
+    ///      - confirmationThreshold > adversaryThreshold (constructor-enforced)
+    ///      - confirmationThreshold - adversaryThreshold > reconstructionThreshold
```
samlaf, line 39: reconstruction threshold is not defined anywhere in the contracts. Would be ideal to define it if we want to refer to it like this I think

docs/spec/src/protocol/architecture/security-parameters.md
```diff
@@ -0,0 +1,90 @@
...
+
+This page proves the relationship between blob parameters and security thresholds. 
+We also point readers to the code where security threshold constraints are implemented.
+
+## Encoding Rate and Reconstruction Threshold
+Recall the blob parameters defined in [Encoding](./encoding.md):
+- $c$: The total number of encoded chunks.  
+- $\gamma$: The ratio of original data to total encoded chunks, providing redundancy.
+- $r$ (`ReconstructionThreshold`): The minimum fraction of total stake required to reconstruct the blob.  
+- $n$: Maximum number of validators.
```
samlaf: do we explain intuitively why there is a bound on the number of validators? What's the intuition here?

docs/spec/src/protocol/architecture/encoding.md
```diff
@@ -24,6 +24,17 @@ Basic RS encoding is used to achieve the first requirement of *Adversarial toler
 
 Notice that given any number of chunks $M$ such that $M \times$`ChunkLength` >= `BlobLength`, via [polynomial interpolation](https://en.wikipedia.org/wiki/Polynomial_interpolation) it is possible to reconstruct the original polynomial, and therefore its coefficients which represent the original blob. 
 
+### Blob Encoding Parameters
+
+We define the **Blob encoding parameters** as a tuple **$(c, r, \gamma)$** where:
+
+- $c$ (`NumChunks`): The total number of encoded chunks after erasure coding (must be a power of 2).  
+- $\gamma$ (`1/CodingRate`): The ratio of original data to total encoded chunks, providing redundancy (must be an inverse power of 2). Note that for representational purposes, the `CodingRate` in our code is the inverse of  $\gamma$, the standard coding rate used in coding theory.
+- $r$ (`ReconstructionThreshold`): The minimum fraction of total stake required to reconstruct the blob.
```
samlaf: Onchain we don't have `r`, we have `maxNumOperators`. I'm assuming r is computed from the 3 params we have onchain?
Would be ideal if our spec here matches what we store onchain.

docs/spec/src/protocol/architecture/encoding.md
```diff
@@ -24,6 +24,17 @@ Basic RS encoding is used to achieve the first requirement of *Adversarial toler
 
 Notice that given any number of chunks $M$ such that $M \times$`ChunkLength` >= `BlobLength`, via [polynomial interpolation](https://en.wikipedia.org/wiki/Polynomial_interpolation) it is possible to reconstruct the original polynomial, and therefore its coefficients which represent the original blob. 
 
+### Blob Encoding Parameters
+
+We define the **Blob encoding parameters** as a tuple **$(c, r, \gamma)$** where:
+
+- $c$ (`NumChunks`): The total number of encoded chunks after erasure coding (must be a power of 2).  
+- $\gamma$ (`1/CodingRate`): The ratio of original data to total encoded chunks, providing redundancy (must be an inverse power of 2). Note that for representational purposes, the `CodingRate` in our code is the inverse of  $\gamma$, the standard coding rate used in coding theory.
```
samlaf: checked quickly, might be wrong. But I didn't see gamma or coding rate used in the RS section above. Shouldn't it?
I don't think we should assume people have coding theory experience. Would be better to provide simpler examples that they can extrapolate from (eg. coding rate of 2 means the original data will be doubled)

docs/spec/src/protocol/architecture/security-parameters.md
```diff
@@ -0,0 +1,90 @@
...
+
+The Byzantine liveness and safety properties of a blob are specified by a collection of `SecurityThresholds`:
+
+- `ConfirmationThreshold` - The confirmation threshold defines the minimum percentage of stake that must sign to make the DA certificate valid.
+- `SafetyThreshold` - The safety threshold refers to the minimum percentage of total stake an attacker must control to make a blob with a valid DA certificate unavailable.
+- `LivenessThreshold` - The liveness threshold refers to the minimum percentage of total stake an attacker must control to cause a liveness failure.
+
+### How to Set the Confirmation Threshold
+
+In the BFT security model, the `SafetyThreshold` and `LivenessThreshold` are estimated by the client. The `SafetyThreshold` is the maximum stake controlled by an adversary that signs the certificate but fails to serve the data, while the `LivenessThreshold` is the maximum stake controlled by an adversary that does not sign the certificates.
```
samlaf, line 100: wdym by estimated here? they are defined onchain

docs/spec/src/protocol/architecture/security-parameters.md
```diff
@@ -0,0 +1,90 @@
+# Security Parameters
+
+This page proves the relationship between blob parameters and security thresholds. 
+We also point readers to the code where security threshold constraints are implemented.
+
+## Encoding Rate and Reconstruction Threshold
+Recall the blob parameters defined in [Encoding](./encoding.md):
```
samlaf: the two lists don't match

docs/spec/src/protocol/architecture/assignment.md
samlaf, line 1: I know that you and @mooselumph really really like math formulas :P
But it would really help to have an excalidraw diagram giving an intuition for encoding process.
blob -> RS-encoded blob -> chunking -> assigning to operators

docs/spec/src/protocol/architecture/security-parameters.md
```diff
@@ -0,0 +1,90 @@
...
+
+Having established the relationship between encoding rate and reconstruction threshold, we now turn to the Byzantine Fault Tolerant (BFT) security model and how it relates to the blob parameters. 
+### Definition of Security Thresholds
+In this section, we define and prove the safety and liveness properties of EigenDA, building on the reconstruction property established above.
+
+The Byzantine liveness and safety properties of a blob are specified by a collection of `SecurityThresholds`:
+
+- `ConfirmationThreshold` - The confirmation threshold defines the minimum percentage of stake that must sign to make the DA certificate valid.
+- `SafetyThreshold` - The safety threshold refers to the minimum percentage of total stake an attacker must control to make a blob with a valid DA certificate unavailable.
+- `LivenessThreshold` - The liveness threshold refers to the minimum percentage of total stake an attacker must control to cause a liveness failure.
```
samlaf, line 96: 1. why don't you define reconstructionThreshold here?
2. in the contract we have adversary threshold. guessing that's safety threshold? (would be useful to also comment this out in the ThresholdRegistryStorage contract)
3. if possible a diagram to relate these thresholds would help

contracts/src/integrations/cert/EigenDACertVerifier.sol
Lulu-Zhou-EigenLabs: Could you point me to the deploy script with the defaults? Thanks!

docs/spec/src/protocol/architecture/encoding.md
Lulu-Zhou-EigenLabs: Yes, r is computed from r = [c/(c - n)] \gamma, where n is `maxNumOperators`.

docs/spec/src/protocol/architecture/assignment.md
Lulu-Zhou-EigenLabs: That's a good idea. I only added the parameters needed to explain the reconstruction threshold in assignment.md. I plan to update the `assignment.md ` with a code walkthrough and diagram in a separate PR.

docs/spec/src/protocol/architecture/security-parameters.md
Lulu-Zhou-EigenLabs: 1. The reconstructionThreshold is a constant after the blob parameters and chunk assignment algorithm are fixed. Therefore, it's more related to the encoding and assignment module.
2. Yes. This mapping is provided in the implementation section below. I'll add the comment in the code as well.
3. I'll show it in a stacked bar chart.

docs/spec/src/protocol/architecture/security-parameters.md
Lulu-Zhou-EigenLabs: The assignment may cause some rounding loss.

**Example:**  
There are 10 chunks, and 3 validators each with 1/3 of the stake.  
- Validator 1 gets 4 chunks  
- Validator 2 gets 3 chunks  
- Validator 3 gets 3 chunks  

Although Validator 2 holds 33% of the stake, it only gets 30% of the chunks.

As a result, the **reconstruction threshold** is larger than the **encoding rate**.

Recall that the relationship between the reconstruction threshold and encoding rate is:
r = (c / (c - n)) · γ  
- r = reconstruction threshold  
- γ = encoding rate  
- (c / (c - n)) > 1 is the coefficient

**Intuition:**  
As the number of operators increases, rounding loss in chunk assignment grows, which increases the coefficient.

docs/spec/src/protocol/architecture/security-parameters.md
Lulu-Zhou-EigenLabs: If I understand correctly, the `SafetyThreshold` is also an internal variable that is populated by the constructor, which can be chosen by the client.

pakim249CAL (approved):

contracts/src/integrations/cert/EigenDACertVerifier.sol
Lulu-Zhou-EigenLabs: Never mind, I found the configuration file.

docs/spec/src/protocol/architecture/encoding.md
Lulu-Zhou-EigenLabs: I've moved the blob parameter part out of `encoding.md`. Now the blob parameters are introduce in the `security-parameter.md` The encoding module will be update in a separate PR.

docs/spec/src/protocol/architecture/security-parameters.md
Lulu-Zhou-EigenLabs: https://github.com/Layr-Labs/eigenda/pull/1949/commits/37a81b88ef687701258737227d32e07c8e750c64

docs/spec/src/protocol/architecture/security-parameters.md
Lulu-Zhou-EigenLabs: For 3: diagram added in https://github.com/Layr-Labs/eigenda/pull/1949/commits/bbfce564707b06507128c54ddb67e454773cd226

Lulu-Zhou-EigenLabs: For 2: https://github.com/Layr-Labs/eigenda/pull/1949/commits/ea88efb731bd65fe525e3d94257c58d7ab80b325

contracts/src/integrations/cert/EigenDACertVerifier.sol
samlaf: I still think we should change this comment. Do you agree?

docs/spec/src/protocol/architecture/security-parameters.md
```diff
@@ -3,20 +3,48 @@
...
+We define the **Blob parameters** as a tuple **$(n, c, \gamma)$** where:
+
+
+- $n$ (`MaxNumOperators`): Maximum number of validators allowed in EigenDA.  
+- $c$ (`NumChunks`): The total number of encoded chunks after erasure coding (must be a power of 2).  
+- $\gamma$ (`1/CodingRate`): The ratio of original data to total encoded chunks, providing redundancy (must be an inverse power of 2). Note that for representational purposes, the `CodingRate` in our code is the inverse of  $\gamma$, the standard coding rate used in coding theory.
+
+Among the blob parameters, `CodingRate` and `NumChunks` are used in the [encoding](./encoding.md) process, while `NumChunks` and `MaxNumOperators` are used in the chunk [assignment](./assignment.md) process.
+
+This tumple is stored in the struct shown below ([see in the code](https://github.com/Layr-Labs/eigenda/blob/d8090af76ed69920983bb3781399a91d84d20d10/contracts/src/core/libraries/v1/EigenDATypesV1.sol#L7)):
```
samlaf:
```suggestion
This tuple is stored in the struct shown below ([see in the code](https://github.com/Layr-Labs/eigenda/blob/d8090af76ed69920983bb3781399a91d84d20d10/contracts/src/core/libraries/v1/EigenDATypesV1.sol#L7)):
```

docs/spec/src/protocol/architecture/security-parameters.md
```diff
@@ -3,20 +3,48 @@
...
+
+```solidity
+struct VersionedBlobParams {
+    uint32 maxNumOperators;
+    uint32 numChunks;
+    uint8 codingRate;
+}
+```
+The blob parameters for each version is stored in deploying the `EigenDAThresholdRegistry` contract.
```
samlaf:
```suggestion
The blob parameters for each version is stored in the `EigenDAThresholdRegistry` contract.
```

docs/spec/src/protocol/architecture/security-parameters.md
```diff
@@ -27,9 +55,37 @@ $$\geq \eta_i(c - n)$$
 Therefore, since $\sum_{i \in H} \eta_i \geq \frac{c}{c-n} \gamma$, we have:
 $$ \sum_{i \in H} c_i \geq \sum_{i \in H} \eta_i (c-n) \geq \frac{c}{c-n} \gamma \cdot (c - n) = \gamma c$$
 
+Now, we prove that any subset of validators with $r$ of the total stake own at least $\gamma c$ chunks, which is guaranteed to reconstruct the origianl blob due to the property of Reed-Solomon encoding.
+
+As we show in the previous subsection, by default, $n = 3537$, $c = 8192$ and $\gamma = 1/8$, which gives us the reconstruction threshold $r = 22\%$.
+
+### Intuition: Loss in Chunk Assignment
```
samlaf, line 62: very very nice and helpful section!

docs/spec/src/assets/security_thresholds.png
samlaf, line 1: Super helpful diagram, thanks for adding this!
weird reconstruction threshold formula (TR) of c/(c-n)*gamma finally clicked for me when I realized that we can look at reconstruction threshold in two different units:
1. RT_chunks (units of data/chunks): reconstruction threshold here IS gamma itself (which is just the % of encoded data that is the actual data... that makes intuitive sense)
2. RT_stake (units of stake): here the simple RT_chunks = gamma doesn't hold, because the data assignment gets distorted by the assignment formula. So we need to add the c/(c-n) factor to balance out

This finally clicked for me, thank you! Do feel like we could document this/draw a picture comparing the different thresholds in both units, but might be a bit hard to drawify.

docs/spec/src/protocol/architecture/assignment.md
```diff
@@ -40,7 +40,7 @@ The assignment algorithm `GetAssignments` works as follows:
    $$
    where $\mathbb{I}$ is the indicator function that adds one extra chunk to the first $c - c'$ validators to ensure the total number of assigned chunks equals $c$.
 
-
+We will prove that any subet of validators with sufficient combined stake can reconstruct the blob in [Security Parameters](./security-parameters.md).
```
samlaf:
```suggestion
We will prove that any subset of validators with sufficient combined stake can reconstruct the blob in [Security Parameters](./security-parameters.md).
```

docs/spec/src/protocol/architecture/security-parameters.md
```diff
@@ -62,29 +118,72 @@ The `ConfirmationThreshold` and `LivenessThreshold` satisfy the following inequa
...
+By substituting the variables using the notation mapping shown at the beginning of this section and simplifying, we get:  
+`ConfirmationThreshold - SafetyThreshold >= (c / (c - n)) * γ = ReconstructionThreshold`, 
+which is exactly inequality (1) shown in the previous subsection.  
 
 We strongly recommend that users set a `SafetyThreshold` >= 33% if they ever want to change the default settings.
 
 **2. Liveness Threshold**
 
-The `LivenessThreshold` does not appear in the code, but users should keep it in mind when changing the default settings. 
+The `LivenessThreshold` does not appear in the code, but users should keep the inequation (2) in mind when setting the confirmation `ConfirmationThreshold`.
```
samlaf:
```suggestion
The `LivenessThreshold` does not appear in the code, but users should keep the equation (2) in mind when setting the confirmation `ConfirmationThreshold`. 
```

docs/spec/src/protocol/architecture/security-parameters.md
```diff
@@ -62,29 +118,72 @@ The `ConfirmationThreshold` and `LivenessThreshold` satisfy the following inequa
...
+`(10000 - ((1_000_000 / (securityThresholds.confirmationThreshold - securityThresholds.adversaryThreshold)) / uint256(blobParams.codingRate))) * uint256(blobParams.numChunks) >= blobParams.maxNumOperators * 10000`  
 
-- `ConfirmationThreshold` : `securityThresholds.confirmationThreshold`
-- `SafetyThreshold` : `securityThresholds.adversaryThreshold`
-- $c$ : `blobParams.numChunks`
-- $n$ : `blobParams.maxNumOperators`
-- $\gamma$: 1 / `blobParams.codingRate`
+By substituting the variables using the notation mapping shown at the beginning of this section and simplifying, we get:  
+`ConfirmationThreshold - SafetyThreshold >= (c / (c - n)) * γ = ReconstructionThreshold`, 
+which is exactly inequality (1) shown in the previous subsection.
```
samlaf, line 175: I tried a few times to do this simplification and failed.... think we might need to actually spell it out a bit more.
The most confusing part for me is that we define [in the code](https://github.com/Layr-Labs/eigenda/blob/d5eec17de570d20d4e24ad53146ad8df4dce54cb/contracts/src/integrations/cert/libraries/EigenDACertVerificationLib.sol#L171) `gamma = confirmationThreshold - adversaryThreshold`, whereas gamma elsewhere is defined  as 1/codingRate... gamma is overloaded in 2 different places its very confusing. Can we rename one of them?

Also is there a way to rewrite the equation into a few simpler checks?
It's nice in mathematics to have one single equation, I get it, but for engineering purposes its not needed at all, plus its very unclear to me that this is optimal performance-wise (we're doing a bunch of divisions).

Can we rewrite this equation into a few different checks that don't use division for example? It would both be more performant, and also a lot cleaner/clearer. Might be ncie to also define reconstructionThreshold in the contract itself.

docs/spec/src/protocol/architecture/security-parameters.md
samlaf: That's correct, but there's no estimation being done. Unless the thresholds mean something more abstract in the theoretical BFT security model.

docs/spec/src/assets/security_thresholds.png
Lulu-Zhou-EigenLabs: That's a good observation! 

> draw a picture comparing the different thresholds in both units

Do you mean a graph that shows the relationship between RT_chunks (\gamma) and RT_stake, when n changes?

samlaf: Not necessarily when n changes, but just find some edge case n that makes it not perfectly align, and show why we need the correction c/(c-n). Honestly not super important... happy to merge already and keep iterating with future PRs

contracts/script/deploy/certverifier/CertVerifierDeployerV2.s.sol
```diff
@@ -35,6 +35,8 @@ contract CertVerifierDeployerV2 is Script, Test {
         raw = stdJson.parseRaw(data, ".eigenDAThresholdRegistry");
         eigenDAThresholdRegistry = abi.decode(raw, (address));
 
+        /// @dev read eigenda/docs/spec/src/protocol/architecture/security-parameters.md
```
pakim249CAL: I would juse use normal comments here, rather than adding "@dev". This flag is normally only used before things like function definitions.

pakim249CAL (approved):

PR #1950, opened 2025-09-05
feat: Anvil testbed + various improvements to tests

dmanc:

This PR did a couple of things (sorted by importance):
- Adds anvil to testbed `testbed/anvil.go`
- Use global test logger for tests.
- Test logger outputs JSON in CI and colored text when ran locally. In the future it would be interesting to write the logs to a file and add the test name as an attribute.
- Adds more reliable cleanup in `inabox/bin.sh`
- Configure localstack and anvil with options
- Fix `./core/thegraph` test 
- Refactor inabox/deploy/utils.go to return errors instead of panics, not sure if this was the right move though.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


Makefile
```diff
@@ -112,10 +112,8 @@ integration-tests:
 # Tests that require a build because they start local inabox infra:
 # either chain, subgraph, or localstack.
 integration-tests-inabox: build
+	go test -v ./core/thegraph
```
samlaf, line 115: nice

api/clients/v2/dispersal_request_signer_test.go
```diff
@@ -32,12 +34,12 @@ func setup(t *testing.T) {
 
 	if deployLocalStack {
 		var err error
-		cfg := testbed.DefaultLocalStackConfig()
-		cfg.Services = []string{"s3, kms"}
-		cfg.Port = localstackPort
-		cfg.Host = "0.0.0.0"
-
-		localstackContainer, err = testbed.NewLocalStackContainer(context.Background(), cfg)
+		localstackContainer, err = testbed.NewLocalStackContainerWithOptions(context.Background(), testbed.LocalStackOptions{
```
samlaf:
```suggestion
		localstackContainer, err = testbed.NewLocalStackContainerWithOptions(t.Context(), testbed.LocalStackOptions{
```

api/clients/v2/dispersal_request_signer_test.go
```diff
@@ -18,6 +19,7 @@ import (
 )
 
 var (
+	logger              = testutils.GetLogger()
 	localstackContainer *testbed.LocalStackContainer
```
samlaf: should we change setup() to return a testbed instead of having this global pointer?

api/clients/v2/dispersal_request_signer_test.go
```diff
@@ -32,12 +34,12 @@ func setup(t *testing.T) {
...
-		cfg := testbed.DefaultLocalStackConfig()
-		cfg.Services = []string{"s3, kms"}
-		cfg.Port = localstackPort
-		cfg.Host = "0.0.0.0"
-
-		localstackContainer, err = testbed.NewLocalStackContainer(context.Background(), cfg)
+		localstackContainer, err = testbed.NewLocalStackContainerWithOptions(context.Background(), testbed.LocalStackOptions{
+			ExposeHostPort: true,
+			HostPort:       localstackPort,
+			Services:       []string{"kms"},
```
samlaf: nit: should prob have an enum with the 3/4 services that we support only?

common/aws/dynamodb/client_test.go
```diff
@@ -44,32 +46,26 @@ func setup(_ *testing.M) {
 
 	if deployLocalStack {
 		var err error
-		cfg := testbed.DefaultLocalStackConfig()
-		cfg.Services = []string{"dynamodb"}
-		cfg.Port = localstackPort
-		cfg.Host = "0.0.0.0"
-
-		localstackContainer, err = testbed.NewLocalStackContainer(context.Background(), cfg)
+		localstackContainer, err = testbed.NewLocalStackContainerWithOptions(context.Background(), testbed.LocalStackOptions{
```
samlaf:
```suggestion
		localstackContainer, err = testbed.NewLocalStackContainerWithOptions(t.Context(), testbed.LocalStackOptions{
```

common/testutils/test_utils.go
```diff
@@ -124,6 +126,31 @@ func RandomString(length int) string {
 	return string(b)
 }
 
+// GetLogger returns a logger for use in tests.
+//
+// In CI ($CI set), logs are written in JSON format to stdout.
+// In local runs, logs are written in text format with colors enabled by default.
+// Colors can be disabled by setting $NO_COLOR.
```
samlaf: not a big fan of using env vars like this.
do we really need to output json in CI? text is always easier to read IMO.
Is it because github terminal doesnt support colors?

common/testutils/test_utils.go
```diff
@@ -124,6 +126,31 @@ func RandomString(length int) string {
...
+//
+// The logger always includes source information and logs at debug level.
+//
+// TODO: Future improvements like writing the test output to a file
+// and adding test metadata (e.g. test name) to log entries.
```
samlaf, line 134: not needed imo. there are a million ways to debug and reading file output is not in the top 5 afaik

common/testutils/test_utils.go
```diff
@@ -124,6 +126,31 @@ func RandomString(length int) string {
...
 func GetLogger() logging.Logger {
```
samlaf, line 135: Can we rename to `GetTestLogger`? Makes it easier to parse when reading test flags that we are using some test specific logger here.

core/indexer/indexer_suite_test.go
```diff
@@ -43,23 +46,31 @@ var _ = BeforeSuite(func() {
...
 		testConfig = deploy.NewTestConfig(testName, rootPath)
 		testConfig.Deployers[0].DeploySubgraphs = false
+		logger := testutils.GetLogger()
 
 		if testConfig.Environment.IsLocal() {
-			fmt.Println("Starting anvil")
-			testConfig.StartAnvil()
+			logger.Info("Starting anvil")
+			var err error
+			anvilContainer, err = testbed.NewAnvilContainerWithOptions(context.Background(), testbed.AnvilOptions{
```
samlaf, line 54: ginkgo BeforeSuite fct doesn't take a t.Context? another reason why we should move away from it

core/thegraph/state_integration_test.go
```diff
@@ -14,17 +13,27 @@ import (
...
-	graphUrl     string
-	testConfig   *deploy.Config
+	anvilContainer      *testbed.AnvilContainer
+	localstackContainer *testbed.LocalStackContainer
+	templateName        string
+	testName            string
+	graphUrl            string
+	testConfig          *deploy.Config
+
+	localstackPort      = "4570"
```
samlaf, line 28: imo we should never be hardcoding ports. Can we change the testbed to assign an arbitrary port and return it instead?

core/thegraph/state_integration_test.go
```diff
@@ -50,55 +59,69 @@ func setup() {
...
-	fmt.Println("Starting binaries")
+	logger.Info("Starting binaries")
 	testConfig.StartBinaries()
 }
 
 func teardown() {
-	fmt.Println("Stopping anvil")
-	testConfig.StopAnvil()
+	_ = localstackContainer.Terminate(context.Background())
+	_ = anvilContainer.Terminate(context.Background())
```
samlaf: update teardown to take a *t.Testing and pass t.Context() instead of context.Background

disperser/dataapi/v2/server_v2_test.go
```diff
@@ -269,7 +270,7 @@ func executeRequest(t *testing.T, router *gin.Engine, method, url string) *httpt
 
 func decodeResponseBody[T any](t *testing.T, w *httptest.ResponseRecorder) T {
 	body := w.Result().Body
-	defer core.CloseLogOnError(body, "response body", mockLogger)
+	defer core.CloseLogOnError(body, "response body", logger)
```
samlaf, line 282: just a test + reading so no issue here, but do want to point out that Jamie has mentioned that simply logging on error is prob not what we want when writing. We prob want writes that error when closing to be caught and reported as an error. Just writing this for general awareness, no action item needed here.

inabox/bin.sh
```diff
@@ -227,11 +227,24 @@ function start_detached {
...
+    # Force cleanup - kill any remaining processes by name
+    echo "Force cleaning up any remaining processes..."
+    pkill -f "churner/bin/server" || true
+    pkill -f "disperser/bin/server" || true
+    pkill -f "disperser/bin/encoder" || true
+    pkill -f "disperser/bin/batcher" || true
+    pkill -f "disperser/bin/controller" || true
+    pkill -f "relay/bin/relay" || true
+    pkill -f "node/bin/node" || true
+    pkill -f "retriever/bin/server" || true
```
samlaf, line 247: I gues we have the assumption that we can only run a single inabox at a time right? When we move to being able to run move than one, we'll have to update this otherwise itll kill all inabox containers

inabox/bin.sh
```diff
@@ -227,11 +227,24 @@ function start_detached {
...
-    kill_processes
-
-    rm -f $pid_file
+    
+    # Try to read PIDs from file if it exists
+    if [[ -f "$pid_file" ]]; then
+        pids=$(cat $pid_file)
+        kill_processes
+        rm -f $pid_file
+    fi
```
samlaf, line 236: at some point we should move away from this and start using process groups, such that killing the job thats leader of the process group kills all its dependencies. We should leverage kernel abstractions instead of this buggy approach where we keep track of stuff in a pid file

inabox/bin.sh
dmanc: Yeah this is a temporary solution. If we run them as containers then this whole bin.sh can go away.

inabox/bin.sh
dmanc: I'm planning to just move these to containers, lmk if any objections to that direction but I feel we should just deprecate bin.sh entirely.

core/thegraph/state_integration_test.go
dmanc: I'll add it as a future enhancement, so separate PR.

common/testutils/test_utils.go
dmanc: It's not us that would be reading file output it's more like you just dump it into Claude. That's currently what we do for inabox anyway and was thinking to preserve that functionality through this test logger.

api/clients/v2/dispersal_request_signer_test.go
dmanc: gonna not do this one, feels low ROI to me

common/testutils/test_utils.go
dmanc: This is gonna be a lot of changes. I'll do it in a separate PR.

core/indexer/indexer_suite_test.go
dmanc: 💯

core/thegraph/state_integration_test.go
dmanc: I tried to do this for most places but some test do TestMain pattern so we're unable to use t.Context()

common/aws/dynamodb/client_test.go
dmanc: Ended up doing a bunch of yak shaving. After this PR i'm going to submit a CLAUDE.md for testing guidelines. Like always use t.Context, etc.

common/testutils/test_utils.go
dmanc: I removed it, I'm pretty sure github does support colored output

dmanc: 9d2719f6a6b10742d8d6f6c0722b6a600f7275f2

common/aws/dynamodb/client_test.go
dmanc: 239a8aec168fc14d1b8d0f3f27079f4a90060494

api/clients/v2/dispersal_request_signer_test.go
dmanc: 239a8aec168fc14d1b8d0f3f27079f4a90060494, tried handling some here but might have missed some

samlaf (approved):

PR #1952, opened 2025-09-05
feat(payments): Implement reservation vault monitor

litt3:
- Creates a utility which monitors the PaymentVault for changes to reservations

### DIFF
--- a/core/payments/reservation/reservation.go
+++ b/core/payments/reservation/reservation.go
@@ -5,6 +5,7 @@ import (
 	"fmt"
 	"time"
 
+	bindings "github.com/Layr-Labs/eigenda/contracts/bindings/v2/PaymentVault"
 	"github.com/Layr-Labs/eigenda/core"
 )
 
@@ -59,6 +60,16 @@ func NewReservation(
 		endTime:            endTime,
 		permittedQuorumIDs: permittedQuorumIDSet,
 	}, nil
+}
+
+// Creates a Reservation from contract binding data
+func NewReservationFromBindings(bindingReservation *bindings.IPaymentVaultReservation) (*Reservation, error) {
+	return NewReservation(
+		bindingReservation.SymbolsPerSecond,
+		time.Unix(int64(bindingReservation.StartTimestamp), 0),
+		time.Unix(int64(bindingReservation.EndTimestamp), 0),
+		bindingReservation.QuorumNumbers,
+	)
 }
 
 // Checks whether an input list of quorums are all permitted by the reservation.
--- a/core/payments/reservation/reservation_vault_monitor.go
+++ b/core/payments/reservation/reservation_vault_monitor.go
@@ -1,0 +1,120 @@
+package reservation
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/Layr-Labs/eigenda/core/payments"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	gethcommon "github.com/ethereum/go-ethereum/common"
+)
+
+// Checks for updates to the PaymentVault contract, and updates ledgers with the new state
+type ReservationVaultMonitor struct {
+	logger logging.Logger
+	// fetches data from the PaymentVault
+	paymentVault payments.PaymentVault
+	// how frequently to fetch state from the PaymentVault to check for updates
+	updateInterval time.Duration
+	// function to get accounts that need to be updated
+	getAccountsToUpdate func() []gethcommon.Address
+	// function to update the reservation for an account
+	updateReservation func(accountID gethcommon.Address, newReservation *Reservation) error
+}
+
+// Creates a new ReservationVaultMonitor and starts a routine to periodically check for updates
+func NewReservationVaultMonitor(
+	ctx context.Context,
+	logger logging.Logger,
+	paymentVault payments.PaymentVault,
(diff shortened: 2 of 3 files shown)


core/payments/reservation/reservation.go
```diff
@@ -61,6 +62,16 @@ func NewReservation(
 	}, nil
 }
 
+// Creates a Reservation from contract binding data
+func NewReservationFromBindings(bindingReservation *bindings.IPaymentVaultReservation) (*Reservation, error) {
```
samlaf: Have we been using this `FromBindings` suffix everywhere?
Took me a min or two to realize that this `IPaymentVaultReservation` was a struct, and not the bindings themselves.
Find it confusing. Would prefer `FromContractsStruct` or something personally.
But maybe bike-shedding here.. your comment does say "from contract binding DATA"

core/payments/reservation/reservation_vault_monitor.go
```diff
@@ -0,0 +1,120 @@
...
+
+	monitor := &ReservationVaultMonitor{
+		logger:              logger,
+		paymentVault:        paymentVault,
+		updateInterval:      updateInterval,
+		getAccountsToUpdate: getAccountsToUpdate,
+		updateReservation:   updateReservation,
+	}
+
+	go monitor.runUpdateLoop(ctx)
```
samlaf, line 48: bike-shed: Ive read at some point that its best to have constructors that don't take a context, and instead let constructors just construct the state, and then force explicit starting/acquiring resources to be done in another method.

Don't have a super strong opinion, so curious for your thoughts. Main issue I see is that when people call a constructor that takes in a context, they aren't super sure what kind of context to pass: should this be a short context just to initialize the struct that needs to maek a single network call? Or does this need to be a process-lifetime context only used for cancelling.

Do think it would be best to at least document the context for this purpose.

core/payments/reservation/reservation_vault_monitor.go
```diff
@@ -0,0 +1,120 @@
...
+
+// Checks for updates to the PaymentVault contract, and updates ledgers with the new state
+type ReservationVaultMonitor struct {
+	logger logging.Logger
+	// fetches data from the PaymentVault
+	paymentVault payments.PaymentVault
+	// how frequently to fetch state from the PaymentVault to check for updates
+	updateInterval time.Duration
+	// function to get accounts that need to be updated
+	getAccountsToUpdate func() []gethcommon.Address
```
samlaf, line 22: are we expecting a lot of addresses to be returned by this? is it always going to be bounded?
Back of my mind keep thinking about the TigerBeetle motto of always having caps on everything, for backpressure and just overall better scalable design.

Way to do that here would prob be to pass in a channel to getAccountsToUpdate and receive the addresses over that. Allows a lot more flexibility for the caller, plus allows vaultMonitor to setup an upper bound on max addresses it can read at once (by having a bounded channel for example).

Thoughts?

core/payments/reservation/reservation.go
litt3: > Have we been using this FromBindings suffix everywhere?

I think I have used this elsewhere, but I agree with your comment.

Renamed as suggest (except I went with singular `Contract` instead of `Contracts`)

79c2abb995df37486f8bc99281b25b8c4d1d00fe

core/payments/reservation/reservation_vault_monitor.go
litt3: I think this makes a lot of sense.

If it's ok with you, I'll do this in a fast-follow PR. I went with this "bad" pattern for the `on demand` version of this utility, so I already need to make a PR addressing this concern anyway

core/payments/reservation/reservation_vault_monitor.go
litt3: This is a good discussion to have.

> is it always going to be bounded?

Yes. This will either be 
- a single account, when being used by a client
- or it will be a number of addresses which is bounded by the cache size configured for the `ReservationLedgerCache` (coming in the next PR)

The reason why I didn't seek to set any sort of limit at the level of the monitor is because I think it's reasonable for the users of this monitor to put sane limits on how they are using the utility.

I did some napkin math with claude's help:
- Using an RPC max response size of 5mb, which is the default for geth nodes, you could fetch >15k reservations while remaining under the response size limit
- Assuming we are using an RPC provider (which have much higher limits), we'd be able to fetch something like 100k reservation structs before hitting compute or gas limits

Therefore, I think this is something worth noting as a potential future optimization, but not something that needs to be addressed in the short term. Cache size will be substantially lower than these values.

(side note: I think I'd personally choose to implement a limit here by still passing in a slice of addresses, but just batching those addresses into multiple RPC calls)

litt3: I added some todos f9c9e7691987802e7102817d5661571891fe91c2

LMK if you're in agreement with my logic above, that this isn't a necessary change right now. If you disagree, I can go ahead and do this optimization now.

samlaf: logic this makes sense to me

litt3: To capture our offline discussion:

I will implement this optimization for both the reservation and on-demand monitors in a follow-up PR, since it's a relatively simple change.

PR #1953, opened 2025-09-05
feat(cert-verifier): contract enforced maxGasLimit

samlaf:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


contracts/src/integrations/cert/EigenDACertVerifier.sol
```diff
@@ -35,11 +35,28 @@ contract EigenDACertVerifier is
...
+    // See https://rareskills.io/post/try-catch-solidity#gdvnie-9-what-gets-returned-during-an-out-of-gas?
+    // The main gas cost associated with verifying cert is the for loop that iterates on all non-signing operators.
+    //
+    // We recommend setting this to 15M gas for optimistic rollups that need the checkDACert call to be executed onchain
+    // in the one step prover contract, to allow for some buffer room for other gas costs in the one step prover tx.
+    // This is because Fusaka is introducing a 16,777,216 per-tx gas limit: https://eips.ethereum.org/EIPS/eip-7825
+    //
+    // For zk rollups and optimistic rollups with zk fraud proofs, this limit is less crucial, but still useful to enforce
+    // a max limit on the amount of prover compute units needs to validate a cert.
+    uint32 internal _checkDACertGasLimit;
```
pakim249CAL: You may want to know that this particular variable may be set to immutable. What this would mean is that the variable would become part of the bytecode of the contract, rather than stored in contract storage. If this variable is not meant to be settable, it is probably better to label it as immutable.

pakim249CAL (approved):

contracts/src/integrations/cert/EigenDACertVerifier.sol
samlaf: Oh I see that makes sense. Guessing `_quorumNumbersRequired` and `_securityThresholds` are not immutable because you can't have immutable bytes and structs?

samlaf: [17f0477](https://github.com/Layr-Labs/eigenda/pull/1953/commits/17f0477b9cc1b783196e1bc70d8bcae37c4e5826)

contracts/src/integrations/cert/EigenDACertVerifier.sol
```diff
@@ -35,11 +35,28 @@ contract EigenDACertVerifier is
...
+    // See https://rareskills.io/post/try-catch-solidity#gdvnie-9-what-gets-returned-during-an-out-of-gas?
+    // The main gas cost associated with verifying cert is the for loop that iterates on all non-signing operators.
+    //
+    // We recommend setting this to 15M gas for optimistic rollups that need the checkDACert call to be executed onchain
+    // in the one step prover contract, to allow for some buffer room for other gas costs in the one step prover tx.
+    // This is because Fusaka is introducing a 16,777,216 per-tx gas limit: https://eips.ethereum.org/EIPS/eip-7825
+    //
+    // For zk rollups and optimistic rollups with zk fraud proofs, this limit is less crucial, but still useful to enforce
+    // a max limit on the amount of prover compute units needs to validate a cert.
+    uint32 internal immutable _checkDACertGasLimit;
```
ethenotethan, line 56: would it make sense to impose an upper bound?

contracts/script/deploy/eigenda/mainnet.beta.config.toml
```diff
@@ -88,4 +88,5 @@ batchConfirmers = []
 [initParams.eigenDA.certVerifier]
 confirmationThreshold = 55
 adversaryThreshold = 33
-quorumNumbersRequired = [0, 1]
\ No newline at end of file
+quorumNumbersRequired = [0, 1]
+checkDACertGasLimit = 15000000
```
ethenotethan:
```suggestion
checkDACertGasLimit = 15_000_000
```

contracts/src/integrations/cert/EigenDACertVerifier.sol
```diff
@@ -108,9 +127,7 @@ contract EigenDACertVerifier is
         // 1. success
         // 2. invalid cert (any failing require statement; we assume all require statements return either a string or custom error)
         // 3. internal error (everything else, including solidity panics and low-level evm reverts, basically anything unexpected)
-        // TODO(samlaf): certVerifier should be set with a maxGas param that will be passed here, to enforce deterministic behavior
-        // between different execution environments: EVM running onchain during optimistic rollup fraud proofs, zkVM, eth-call with higher gas limit.
-        try this.checkDACertReverts(daCert) {
+        try this.checkDACertReverts{gas: _checkDACertGasLimit}(daCert) {
```
ethenotethan, line 134: why pass through call stack vs doing a post-mortem check afterward; eg:
```
init_gas_available = gasLeft()
checkDACertReverts(daCert)
gas_used = init_gas_available - gasLeft()

if (gas_used >= checkDACertGasLimit) {

   return uint8(StatusCode.OutOfGas)
```


I guess we'd only really need this if we wanted explicit understanding via status code

contracts/src/integrations/cert/EigenDACertVerifier.sol
ethenotethan: also a lower bound - there should probably be a defensive check in the deploy script to stop a user from accidentally setting it to 0

contracts/src/integrations/cert/EigenDACertVerifier.sol
```diff
@@ -35,11 +35,28 @@ contract EigenDACertVerifier is
 
     DATypesV1.SecurityThresholds internal _securityThresholds;
 
+    // Each quorum that is part of this list MUST meet the confirmationThreshold defined in _securityThresholds
+    // in order for a cert to be considered valid.
+    // Recommended default value is 0x0001 to require the ETH and EIGEN quorums to sign.
```
ethenotethan, line 41: recommended or required defaults?

contracts/src/integrations/cert/EigenDACertVerifier.sol
```diff
@@ -35,11 +35,28 @@ contract EigenDACertVerifier is
 
     DATypesV1.SecurityThresholds internal _securityThresholds;
```
ethenotethan, line 37: let's also add a forge test which ensures that an amount set really low (e.g, 1) results in an INTERNAL_ERROR status code

contracts/src/integrations/cert/EigenDACertVerifier.sol
samlaf: recommended. they can do whatever they want.

contracts/src/integrations/cert/EigenDACertVerifier.sol
samlaf: do you mean amount of gas? youre pointing to security thresholds

contracts/src/integrations/cert/EigenDACertVerifier.sol
samlaf: hmm does this actually work? If a low-level evm revert happens it also consumes all gas passed to the call right? So how do you tell OOG apart from the other revert reasons?
image

contracts/src/integrations/cert/EigenDACertVerifier.sol
samlaf: Not sure how to test this... because the `_deployDA` in `setUp()` deploys the certVerifier with hardcoded 15M gaslimit.

image

Do you know of a workaround?

contracts/src/integrations/cert/EigenDACertVerifier.sol
ethenotethan: I think we might have to impose the assumption that we'd never hit an out-of-gas exception in the internal verification logic

ethenotethan: given we map out-of-gas to invalid cert

ethenotethan: I know the cert is invalid, you say it's valid. We diverge to a one step proof that **I have to issue**

ethenotethan: how do I prove to you the cert is invalid if the tx reverts due to out-of-gas?

contracts/src/integrations/cert/EigenDACertVerifier.sol
ethenotethan: couldn't you add an optional parameter to the setUp function?

contracts/src/integrations/cert/EigenDACertVerifier.sol
samlaf: [25a832f](https://github.com/Layr-Labs/eigenda/pull/1953/commits/25a832f2fe0599a09f63f70b998c46a1106bb144)

Don't feel like any upper bound would make sense, like no sane default

samlaf:
After chatting with @ethenotethan, the implementation in this PR still has some unresolved issues and is not a complete solution.
image

1. calldata gas costs are not taken into consideration (huge cert could consume a lot of gas)
2. abi.decode gas costs not taken into consideration
3. _checkDACertGasLimit is an upper bound, meaning if less than that amount is left at that point, that amount is passed (that amount or 63/64th of whats left...???)
   1. if all of it is passed, then the catch could revert, which would break the security assumption

Perhaps all we need then is an answer to question at point 3? Solidity [docs](https://docs.soliditylang.org/en/latest/control-structures.html#external-function-calls) don't seem to have an answer.

PR #1956, opened 2025-09-06
fix(mdbook): Add 3rd party latex dependency to fix display bug

ethenotethan:
**Before**
Screenshot 2025-09-07 at 1 37 12 AM

**After**
Screenshot 2025-09-07 at 1 38 47 AM

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/verification/cert_verifier.go
+++ b/api/clients/v2/verification/cert_verifier.go
@@ -77,6 +77,10 @@ func (cv *CertVerifier) CheckDACert(
 	// TODO(ethenotethan): determine if there's any merit in passing call context
 	// options (e.g, block number) to impose better determinism and safety on the simulation
 	// call
+	//
+	// imposing determinism here by binding a block reference would introduce the requirement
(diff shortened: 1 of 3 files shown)


api/clients/v2/verification/cert_verifier.go
```diff
@@ -77,6 +77,10 @@ func (cv *CertVerifier) CheckDACert(
 	// TODO(ethenotethan): determine if there's any merit in passing call context
 	// options (e.g, block number) to impose better determinism and safety on the simulation
 	// call
+	//
+	// imposing determinism here by binding a block reference would introduce the requirement
+	// for an archival node for rollups syncing which would introduce an additional operational cost
+	// to rollup operators
```
samlaf, line 83: move this entire comment down to where the actual call is made?

samlaf: btw this you actually want this in? Its very unrelated to the actual PR

PR #1957, opened 2025-09-08
feat(proxy): Arbitrum Custom DA Server Blueprint

ethenotethan:
Introduces scaffolding and blueprint for the Arbitrum Custom DA JSON RPC server that's gonna be compatible with EIgenDAV2. The handler interface is defined with all functions currently unimplemented (except the header byte check).  There is a lot of documentation and todos. 

This also refactors the existing server into a specific `rest` package with explicit renaming to `REST ALT DA Server` type.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proxy/cmd/server/entrypoint.go
```diff
@@ -28,7 +32,7 @@ func StartProxySvr(cliCtx *cli.Context) error {
 		return err
 	}
 
-	log.Info("Starting EigenDA Proxy Server", "version", Version, "date", Date, "commit", Commit)
+	log.Info("Starting EigenDA Proxy Service", "version", Version, "date", Date, "commit", Commit)
```
samlaf, line 35: not super opinionated but I personally prefer Server

samlaf: I think you're trying to say that the proxy "service" hosts different 'servers' (rest, rpc, metrics)?
Think the reserve: proxy "server" hosts "services"/endpoints makes more sense to me?

api/proxy/cmd/server/entrypoint.go
```diff
@@ -63,27 +67,47 @@ func StartProxySvr(cliCtx *cli.Context) error {
...
+		if err != nil {
+			return fmt.Errorf("new arbitrum custom da json rpc server: %w", err)
+		}
+
+		if err := arbitrumRpcServer.Start(); err != nil {
+			return fmt.Errorf("start arbitrum custom da json rpc server: %w", err)
+		}
+
+		defer func() {
+			if err := arbitrumRpcServer.Stop(); err != nil {
```
samlaf, line 102: hmm you prob don't want a defer here right? did you test this?
this function returns, so this'll immediately stop the rpc server

api/proxy/config/app_config.go
```diff
@@ -7,16 +7,19 @@ import (
...
 type AppConfig struct {
-	StoreBuilderConfig  builder.Config
-	SecretConfig        common.SecretConfigV2
-	ServerConfig        server.Config
+	StoreBuilderConfig builder.Config
+	SecretConfig       common.SecretConfigV2
+
+	ArbCustomDASvrCfg   arbitrum_altda.Config
+	RestSvrCfg          rest.Config
 	MetricsServerConfig metrics.Config
```
samlaf: can we rename all of them to `SvrCfg` or `ServerConfig`, but not mix and match?

api/proxy/docs/help_out.txt
```diff
@@ -16,6 +16,12 @@ GLOBAL OPTIONS:
    --help, -h     show help
    --version, -v  print the version
 
+   Arbitrum Custom DA JSON RPC Server
+
+   --arbitrum-da.addr value  Server listening address (default: "0.0.0.0") [$EIGENDA_PROXY_ARB_CUSTOM_DA_ADDR]
+   --arbitrum-da.enabled     Whether or not to enable Arbitrum Custom DA JSON RPC API (default: false) [$EIGENDA_PROXY_ARB_CUSTOM_DA_ENABLED]
+   --arbitrum-da.port value  Server listening port (default: 3100) [$EIGENDA_PROXY_ARB_CUSTOM_DA_PORT]
```
samlaf: are we sure that this json-rpc server is only arbitrum specific and that it couldnt ever be reused for other stacks?
Just wondering whether we should call these `--json-rpc.addr` instead of `--arbitrum-da.addr`

api/proxy/servers/arbitrum_altda/cli.go
```diff
@@ -0,0 +1,51 @@
...
+)
+
+const (
+	ListenAddrFlagName = "arbitrum-da.addr"
+	PortFlagName       = "arbitrum-da.port"
+	Enabled            = "arbitrum-da.enabled"
+)
+
+func withEnvPrefix(prefix, s string) []string {
+	return []string{prefix + "_ARB_CUSTOM_DA_" + s}
```
samlaf: why CUSTOM? Why not just `_ARBITRUM_DA_` to match the flag names?

api/proxy/servers/arbitrum_altda/cli.go
```diff
@@ -0,0 +1,51 @@
...
+			Name:     ListenAddrFlagName,
+			Usage:    "Server listening address",
+			Value:    "0.0.0.0",
+			EnvVars:  withEnvPrefix(envPrefix, "ADDR"),
+			Category: category,
+		},
+		&cli.IntFlag{
+			Name:     PortFlagName,
+			Usage:    "Server listening port",
+			Value:    3100,
```
samlaf: same port as the rest server right? don't think 2 services can listen on a single port

api/proxy/servers/arbitrum_altda/cli.go
```diff
@@ -0,0 +1,51 @@
...
+			EnvVars:  withEnvPrefix(envPrefix, "PORT"),
+			Category: category,
+		},
+		&cli.BoolFlag{
+			Name:     Enabled,
+			Usage:    "Whether or not to enable Arbitrum Custom DA JSON RPC API",
```
samlaf, line 35: curious why you use `custom` everywhere? is that part of arbitrum's spec? it sounds bespoke and not very serious. can we just name is arbitrum altda json rpc api?

api/proxy/servers/arbitrum_altda/common.go
```diff
@@ -0,0 +1,78 @@
...
+	"github.com/ethereum/go-ethereum/common/hexutil"
+)
+
+// EigenDAV2MessageHeaderByte is the unique EigenDAV2MessageHeaderByte.
+// The value chosen is purely arbitrary.
+//
+// TODO: See if there can be social consensus on the value we assume,
+// otherwise could eventually conflict with competitors or even OCL.
+// maybe we could reuse the existing OP social contract for DA layers
+// and assume 0x01?
```
samlaf: add link to op's forum post

api/proxy/servers/arbitrum_altda/common.go
```diff
@@ -0,0 +1,78 @@
...
+const EigenDAV2MessageHeaderByte byte = 0x42
+
+type PreimageType uint8
+
+// The ALT DA server only cares about type 3 or Custom DA preimage types
+const (
+	CustomDAPreimageType = 3
```
samlaf:
```suggestion
	CustomDAPreimageType PreimageType = 3
```

api/proxy/servers/arbitrum_altda/common.go
```diff
@@ -0,0 +1,78 @@
...
+//
+// TODO: See if there can be social consensus on the value we assume,
+// otherwise could eventually conflict with competitors or even OCL.
+// maybe we could reuse the existing OP social contract for DA layers
+// and assume 0x01?
+const EigenDAV2MessageHeaderByte byte = 0x42
+
+type PreimageType uint8
+
+// The ALT DA server only cares about type 3 or Custom DA preimage types
```
samlaf:
```suggestion
// The ALT DA server only cares about type 3 Custom DA preimage types
```

samlaf: or introduces ambiguity as to whether theres 2 things or not

api/proxy/servers/arbitrum_altda/common.go
samlaf, line 1: maybe good to add a `mod.go` or README in this section and describe the current state of arb rpc spec/implementation, pointing to links where those are defined/implemented

api/proxy/servers/arbitrum_altda/common.go
```diff
@@ -0,0 +1,78 @@
...
+const (
+	CustomDAPreimageType = 3
+)
+
+// PreimagesMap maintains a nested mapping:
+// //   preimage_type -> preimage_hash_key -> preimage bytes
+//
+// only the CustomDAPreimageType is used for EigenDAV2 batches
+//
+// TODO: Figure out what's getting passed over that wire. We're
```
samlaf:
```suggestion
// TODO: Figure out what's getting passed over the wire. We're
```

api/proxy/servers/arbitrum_altda/common.go
```diff
@@ -0,0 +1,78 @@
...
+// being passed a mapping and populating it with EigenDAV2 Arbitrum batch
+// context. It'd be good to know when an EigenDA batch is populated
+// during the preimage mapping population given we use 16mib-buffer_to_ensure_padding batch
```
samlaf: is this buffer_to_ensure_padding a variable?

api/proxy/servers/arbitrum_altda/common.go
```diff
@@ -0,0 +1,78 @@
...
+// sizing which arbitrum doesn't
+// natively support. If other preimage entries exist for the node prestate trie then data size limits
+// could unknowningly be hit/exceeded.
+type PreimagesMap map[PreimageType]map[common.Hash][]byte
+
+/*
+	These response types are copied verbatim (types, comments) from the upstream nitro reference implementation.
```
samlaf: link?

samlaf: also don't think we should say "upstream" since nitro is not an upstream dependency for eigenda repo.

api/proxy/servers/arbitrum_altda/server.go
```diff
@@ -0,0 +1,101 @@
...
+	return &Server{
+		cfg: cfg,
+		svr: svr,
+	}, nil
+
+}
+
+func (s *Server) Addr() string {
+	return s.addr
+}
```
samlaf, line 69: this is unused. is it really needed?

api/proxy/config/app_config.go
```diff
@@ -47,7 +50,7 @@ func ReadAppConfig(ctx *cli.Context) (AppConfig, error) {
 	return AppConfig{
 		StoreBuilderConfig:  storeBuilderConfig,
 		SecretConfig:        eigendaflags.ReadSecretConfigV2(ctx),
```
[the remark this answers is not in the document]
samlaf: Why did you resolve this? you're not reading the arb config

api/proxy/servers/arbitrum_altda/server.go
```diff
@@ -0,0 +1,101 @@
...
+//
+// Write toggling allows a user to use the ALT DA server in "read only" mode
+// which is typical behavior for all rollup operator nodes outside of the batch poster
+// TODO: Add support for "read only" mode
+//
+// Server request body limits are used to impose a maximum on the allowed request body size
+// TODO: Understand if this is actually used/respected in DA client anywhere
+// TODO: Add env ingestion for these values
+// TODO: Determine a proper default given EigenDA's large batch uniqueness. The defaultBodyLimit
+// //    is currently set to 5 mib which is too low for integration EigenDA
```
samlaf, line 31: just set it to be maxBlobSize, which is already a flag we have?

api/proxy/servers/arbitrum_altda/cli.go
```diff
@@ -0,0 +1,51 @@
+package arbitrum_altda
+
+import (
+	"github.com/urfave/cli/v2"
+)
+
+const (
+	ListenAddrFlagName = "arbitrum-da.addr"
```
samlaf, line 8: nit: might be better to use `host` instead of `addr`? Since `localhost` is a valid addr for example.

api/proxy/servers/arbitrum_altda/server.go
samlaf: oh I think you just forgot to use it in the log line where the server gets started

api/proxy/cmd/server/entrypoint.go
```diff
@@ -63,27 +67,47 @@ func StartProxySvr(cliCtx *cli.Context) error {
...
+				log.Error("failed to stop arbitrum custom da json rpc server", "err", err)
+			}
+			log.Info("Successfully shutdown Arbitrum Custom DA server")
+		}()
+
+		log.Info("Started Arbitrum Custom DA JSON RPC server")
```
samlaf:
```suggestion
		log.Info("Started Arbitrum Custom DA JSON RPC server", "addr", arbitrumRpcServer.Addr())
```

samlaf: mimicing op endpoint

api/proxy/servers/arbitrum_altda/handlers.go
```diff
@@ -0,0 +1,186 @@
...
+
+// Handlers defines the Arbitrum ALT DA server spec's JSON RPC methods
+// This method implementations should serve as a thin wrapper over the existing EigenDA manager construct
+// with translation mapping 503 (failover) and 418 (invalid_cert) status codes into error messages that
+// arbitrum nitro can understand to take actions preserving both rollup liveness and safety
+//
+// Some custom code / refactoring will likely be necessary for supporting the READPREIMAGE proof serialization logic
+type Handlers struct {
+	// TODO: Metrics support - makes sense to share metrics server between both rest and arbitrum alt da
+	//       servers. go-ethereum might suppprot
```
samlaf: incomplete sentence. but agree we could use a label in the metrics to specify which route the request was received on

api/proxy/servers/arbitrum_altda/handlers.go
```diff
@@ -0,0 +1,186 @@
...
+	// TODO: Add EigenDA manager here
+	// TODO: Add logging
+}
+
+// IsValidHeaderByte determines whether or not the sequencer message header byte is an EigenDAV2 cert type.
+// Arbitrum Nitro does this check via a bitwise AND which can cause overlapping and requires careful future
+// management. while we could determine a byte value with bits that don't overlap - it's more maintainable
+// to do a literal comparison and assume OCL would never introduce a conlficting byte value
+// OR our com
```
samlaf: ?

api/proxy/servers/arbitrum_altda/handlers.go
```diff
@@ -0,0 +1,186 @@
...
+//
+// // current encoding proposal:
+// // // SERIALIZED EigenDA V2 CUSTOM READ PREIMAGE PROOF:
+// // // kzg commitment and preimage length are extractable
+// // // from the existing DA Cert
+//
+// // // [0:32]  - root of unity @ field element offset
+// // // [32:64] - field element or preimageChunk being one step proven
+// // // [64:128] - point opening proof (g1 point)
+// // // [128:256] - g2TauMinusG2z
```
samlaf: nit: reads a bit weird with all the //

api/proxy/servers/arbitrum_altda/common.go
samlaf: also nit: I really dislike `common.go` as a name. Would prefer `types` here, or `msg_types` etc. common is just too big of a miscellaneous place where any kind of crap is added and becomes hard to maintain.

api/proxy/cmd/server/entrypoint.go
```diff
@@ -17,7 +18,10 @@ import (
 	"github.com/ethereum-optimism/optimism/op-service/ctxinterrupt"
 )
 
-func StartProxySvr(cliCtx *cli.Context) error {
+// TODO: Explore better enscapsulation patterns that binds common interfaces / usage patterns
```
litt3:
```suggestion
// TODO: Explore better encapsulation patterns that binds common interfaces / usage patterns
```

api/proxy/config/flags.go
```diff
@@ -25,7 +26,8 @@ const (
 	S3Category              = "S3 Cache/Fallback"
 	VerifierCategory        = "Cert Verifier (V1 only)"
 	KZGCategory             = "KZG"
-	ProxyServerCategory     = "Proxy Server"
+	ProxyRestServerCategory = "Proxy REST API Server (compatible with OP Stack ALT DA and standard commitment clients)"
```
litt3, line 29: The new wording you're using throughout is very rollup specific. Given that the proxy server is meant to work for non-rollup use cases, would it make more sense to keep the preexisting generic naming? It seems like it would be reasonable to keep the generic server type, and just add a more specific type of server which can be leveraged by certain usecases.

api/proxy/config/app_config.go
```diff
@@ -7,16 +7,19 @@ import (
...
 
 // AppConfig ... Highest order config. Stores all relevant fields necessary for running both proxy & metrics servers.
 type AppConfig struct {
-	StoreBuilderConfig  builder.Config
-	SecretConfig        common.SecretConfigV2
-	ServerConfig        server.Config
+	StoreBuilderConfig builder.Config
+	SecretConfig       common.SecretConfigV2
+
+	ArbCustomDASvrCfg   arbitrum_altda.Config
```
litt3: Does it actually make sense to add this new config into the `AppConfig`? 

I saw you had a TODO to explore better abstractions between the different server types, and I think that now is probably the time to do that. If we glom the custom arbitrum server into the existing structures for the sake of expediency, it's going to be very difficult to extract later.

Maybe we should be creating an entirely new entrypoint for this arbitrum server, instead of building it into the existing entrypoint. This would take some investment upfront to extract some of the preexisting logic into utilities that can be used by both the "standard" proxy server and the arbitrum one. But IMO now would be the time, if we want to avoid incurring a lot of tech debt

api/proxy/servers/arbitrum_altda/server.go
ethenotethan: Future proofing for integration tests where we pass in port 0 and it randomly assigns

api/proxy/servers/arbitrum_altda/common.go
ethenotethan: I couldn't find it when I tried looking - would you happen to have it handy?

api/proxy/cmd/server/entrypoint.go
ethenotethan: not sure I follow:
```
Sep 10 00:46:16.123 INF rest/server.go:83 Starting DA server endpoint=[::]:3100
Sep 10 00:46:16.134 INF server/entrypoint.go:81 Started EigenDA proxy REST ALT DA server
Sep 10 00:46:16.135 INF server/entrypoint.go:108 Started Arbitrum Custom DA JSON RPC server

^C

Sep 10 00:47:42.579 INF server/entrypoint.go:105 Successfully shutdown Arbitrum Custom DA server
Sep 10 00:47:42.579 INF server/entrypoint.go:88 Successfully shutdown REST ALT DA server
```

api/proxy/docs/help_out.txt
ethenotethan: hmm that is a good point - technically we could always expose other namespaced APIs on the server for other stacks/use-cases

api/proxy/servers/arbitrum_altda/cli.go
ethenotethan: yeah good find

api/proxy/servers/arbitrum_altda/cli.go
ethenotethan: https://github.com/OffchainLabs/nitro/pull/3548

api/proxy/servers/arbitrum_altda/cli.go
ethenotethan: probably makes more sense to align with what we already use for the REST ALT DA Server:
```
		&cli.StringFlag{
			Name:     ListenAddrFlagName,
			Usage:    "Server listening address",
			Value:    "0.0.0.0",
			EnvVars:  withEnvPrefix(envPrefix, "ADDR"),
			Category: category,
		},
```

api/proxy/servers/arbitrum_altda/common.go
ethenotethan: tbh I there won't be a canonical link for sometime given OCL closed their Custom DA PR and is doing piecewise changes to get it through

api/proxy/cmd/server/entrypoint.go
ethenotethan: hmm that breaks the package name of `servers` - can change it to services though if we feel this is better terminology for reasoning about the system

api/proxy/config/app_config.go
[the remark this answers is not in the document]
ethenotethan: huh no it's updated to read the config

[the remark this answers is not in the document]
ethenotethan: ab1faed7f6d20200c0819a3f7dd3aa96fff3b5b8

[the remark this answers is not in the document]
ethenotethan: looks I forgot to push my latest head up with some minor patches

api/proxy/config/app_config.go
ethenotethan: [79f3514](https://github.com/Layr-Labs/eigenda/pull/1957/commits/79f35142afd63a6d67e4cb6efc4091a2d8c0e58c)

api/proxy/servers/arbitrum_altda/handlers.go
ethenotethan: a21569eeba4021d16cda6954ef25809eecc77974

api/proxy/servers/arbitrum_altda/handlers.go
ethenotethan: a21569eeba4021d16cda6954ef25809eecc77974

api/proxy/servers/arbitrum_altda/handlers.go
ethenotethan: a21569eeba4021d16cda6954ef25809eecc77974

api/proxy/servers/arbitrum_altda/cli.go
ethenotethan: [e269999](https://github.com/Layr-Labs/eigenda/pull/1957/commits/e269999c94bc2261e4d890cfb4b705f45b830c17)

api/proxy/servers/arbitrum_altda/common.go
ethenotethan: [c11e0f7](https://github.com/Layr-Labs/eigenda/pull/1957/commits/c11e0f7f88bcf362ae1823f287a25748ecb2b54e)

api/proxy/servers/arbitrum_altda/server.go
ethenotethan: Don't think `maxBlobSize` is technically the correct value for this given the maximum allowed value would be a bit less to compensate for additional bytes added in the payload -> blob encoding which is `16_252_897`

api/proxy/config/app_config.go
ethenotethan: > Maybe we should be creating an entirely new entrypoint for this arbitrum server, instead of building it into the existing entrypoint

Are you referring to an additional entry-point abstraction or separate binary? I fail to see how the dependency injection pattern for how the ARB Custom DA server would differ from the existing metrics one

api/proxy/servers/arbitrum_altda/cli.go
ethenotethan: https://github.com/Layr-Labs/eigenda/pull/1957/commits/7cccd3cb6b51137f26677e1ce1e085cc115f9f50

api/proxy/servers/arbitrum_altda/common.go
ethenotethan: just referring to MaxPayloadSize - updated

ethenotethan: [76a8c51](https://github.com/Layr-Labs/eigenda/pull/1957/commits/76a8c518a40f1878cc83426a18eac8a05f07a026)

api/proxy/config/app_config.go
litt3: I was misunderstanding some stuff, ignore what I said!

api/proxy/cmd/server/entrypoint.go
samlaf: ah nvm just understood how the ctxinterrupt.Wait() at the end works.. I thought it was returning control to the urface framework or something, but it just blocks waiting for signals. Cool this works then, nvm.

api/proxy/docs/help_out.txt
samlaf: after discussing this, this is good. very unlikely that the same exact json-rpc will be used by other stacks.

api/proxy/servers/arbitrum_altda/cli.go
samlaf: ah I see

api/proxy/servers/arbitrum_altda/common.go
samlaf: https://github.com/ethereum-optimism/specs/discussions/135

api/proxy/servers/arbitrum_altda/server.go
samlaf: oh right we had talked about computing maxPayloadSize from maxBlobSize (we have a util function for that somewhere). Could use that then

api/proxy/servers/arbitrum_altda/cli.go
samlaf: snap good catch. don't like it but it is what it is

PR #1962, opened 2025-09-08
docs(proxy): explain g2 srs flags better

samlaf:
Closes DAINT-745.

The explanation was lacking before, which caused some confusion. Took the explanation from the KZG flags that we use for the operator nodes, and ammended them to match proxy use case.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/proxy/docs/help_out.txt
+++ b/api/proxy/docs/help_out.txt
@@ -86,10 +86,10 @@ This check is optional and will be skipped when set to 0. (default: 0) [$EIGENDA
 
    KZG
 
-   --eigenda.cache-path value        path to SRS tables for caching. This resource is not currently used, but needed because of the shared eigenda KZG library that we use. We will eventually fix this. (default: "resources/SRSTables/") [$EIGENDA_PROXY_EIGENDA_TARGET_CACHE_PATH]
+   --eigenda.cache-path value        Path to SRS tables for caching. This resource is not currently used, but needed because of the shared eigenda KZG library that we use. We will eventually fix this. (default: "resources/SRSTables/") [$EIGENDA_PROXY_EIGENDA_TARGET_CACHE_PATH]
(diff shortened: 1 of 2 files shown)


api/proxy/docs/help_out.txt
```diff
@@ -86,10 +86,10 @@ This check is optional and will be skipped when set to 0. (default: 0) [$EIGENDA
...
-   --eigenda.g2-path value           path to g2.point file. (default: "resources/g2.point") [$EIGENDA_PROXY_EIGENDA_TARGET_KZG_G2_PATH]
-   --eigenda.g2-path-trailing value  path to g2.trailing.point file. (default: "resources/g2.trailing.point") [$EIGENDA_PROXY_EIGENDA_TARGET_KZG_G2_TRAILING_PATH]
+   --eigenda.g2-path value           Path to g2.point file, which is needed to generate and verify blob length proofs. If this flag doesn't point to a complete 16GiB G2 SRS file, then the --eigenda.g2-path-trailing flag must be used to specify a trailing segment of that file, of equivalent length to the initial segment pointed to by this flag. (default: "resources/g2.point") [$EIGENDA_PROXY_EIGENDA_TARGET_KZG_G2_PATH]
+   --eigenda.g2-path-trailing value  Path to g2.trailing.point file, which is needed to generate and verify blob length proofs. If --eigenda.g2-path is pointing to the entire 16GiB G2 SRS file which contains 268435456 G2 points, this flag is not needed. Otherwise, users can truncate the G2 file by taking an initial segment, and a trailing segment (both of the same size). (default: "resources/g2.trailing.point") [$EIGENDA_PROXY_EIGENDA_TARGET_KZG_G2_TRAILING_PATH]
```
anupsv, line 92: We should probably make it simpler and have them use the full file ? Makes it easier ?

samlaf: Full file is 16GiB lol, doesn't really make sense. I think they just want the explanation to understand what trailing file is.

PR #1963, opened 2025-09-09
feat: replace churning

pakim249CAL:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/contracts/src/core/EigenDARegistryCoordinator.sol
+++ b/contracts/src/core/EigenDARegistryCoordinator.sol
@@ -147,7 +147,7 @@ contract EigenDARegistryCoordinator is
         }).numOperatorsPerQuorum;
 
         // For each quorum, validate that the new operator count does not exceed the maximum
-        // If it does, churns an operator via an exhaustive search through the operator set.
+        // If it does, churns the operator with the lowest stake via an exhaustive search through the operator set.
         for (uint256 i; i < quorumNumbers.length; i++) {
             uint8 quorumNumber = uint8(quorumNumbers[i]);
 
@@ -157,7 +157,8 @@ contract EigenDARegistryCoordinator is
         }
     }
 
-    /// @notice Deprecated function. Use `registerOperator` instead.
+    /// @notice Deprecated function. Use `registerOperator` instead, which implements churning without a churn approver.
+    ///         Kept for backwards compatibility purposes only.
     function registerOperatorWithChurn(
         bytes calldata quorumNumbers,
         string calldata socket,
@@ -805,6 +806,7 @@ contract EigenDARegistryCoordinator is
     }
 
     /// @notice Deprecated function.
+    /// @dev    Kept for backwards compatibility purposes, and will be deleted when the migration to the new churning process is completed.
     function calculateOperatorChurnApprovalDigestHash(address, bytes32, OperatorKickParam[] memory, bytes32, uint256)
         external
         pure


contracts/src/core/EigenDARegistryCoordinator.sol
```diff
@@ -161,6 +161,66 @@ contract EigenDARegistryCoordinator is
         }
     }
 
+    function registerOperatorWithChurn(
```
anupsv, line 162: Is this for operators registering with churn enabled ?

contracts/src/core/EigenDARegistryCoordinator.sol
```diff
@@ -161,6 +161,66 @@ contract EigenDARegistryCoordinator is
...
+    function _churnOperator(uint8 quorumNumber) internal {
+        bytes32[] memory operatorList = indexRegistry.getOperatorListAtBlockNumber(quorumNumber, uint32(block.number));
+        require(operatorList.length > 0, "RegCoord._churnOperator: no operators to churn");
+
+        // Find the operator with the lowest stake
+        bytes32 operatorToChurn;
+        uint96 lowestStake = type(uint96).max;
+        for (uint256 i; i < operatorList.length; i++) {
+            uint96 operatorStake = stakeRegistry.getCurrentStake(operatorList[i], quorumNumber);
+            if (operatorStake < lowestStake) {
```
anupsv, line 182: wasn't there a requirement of the new stake to be a certain % above the lowest stake or so ?

contracts/src/core/EigenDARegistryCoordinator.sol
pakim249CAL: The current situation is this:

To avoid removing any functions that might break anything else in the rest of the repository, we wanted to add a new function (this function) that does the churning logic without touching the old logic.

Unfortunately we cannot overload registerOperator because none of the function arguments change. So here I just choose to overload registerOperatorWithChurn.

I don't think this will be the final solution, because it simply does not work. Adding a new function increases the contract size, and the contract is already at its size limit.

contracts/src/core/EigenDARegistryCoordinator.sol
pakim249CAL: Can you link me to this requirement? I don't see it anywhere.

anupsv: @Lulu-Zhou-EigenLabs might know but i'll try to find it.

contracts/src/core/EigenDARegistryCoordinator.sol
```diff
@@ -150,91 +147,50 @@ contract EigenDARegistryCoordinator is
         }).numOperatorsPerQuorum;
 
         // For each quorum, validate that the new operator count does not exceed the maximum
-        // (If it does, an operator needs to be replaced -- see `registerOperatorWithChurn`)
-        for (uint256 i = 0; i < quorumNumbers.length; i++) {
+        // If it does, churns an operator via an exhaustive search through the operator set.
+        for (uint256 i; i < quorumNumbers.length; i++) {
```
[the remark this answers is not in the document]
pakim249CAL: I know it's not consistent, but I prefer this style for changes I make. I don't think it's a problem.

contracts/src/core/EigenDARegistryCoordinator.sol
pakim249CAL: The approach decided on was to simply change this function to use the new churning logic to keep things as backwards compatible as possible.

contracts/src/core/EigenDARegistryCoordinator.sol
```diff
@@ -150,91 +147,50 @@ contract EigenDARegistryCoordinator is
...
-                    setParams: operatorSetParams
-                });
-
-                _deregisterOperator(operatorKickParams[i].operator, quorumNumbers[i:i + 1]);
+    ) external virtual {
+        registerOperator(quorumNumbers, socket, params, operatorSignature);
+    }
+
+    function _churnOperator(uint8 quorumNumber) internal {
+        bytes32[] memory operatorList = indexRegistry.getOperatorListAtBlockNumber(quorumNumber, uint32(block.number));
```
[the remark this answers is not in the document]
pakim249CAL: IDK what this AI is talkin about. Querying the current block number is just fine.

contracts/src/core/EigenDARegistryCoordinator.sol
```diff
@@ -150,91 +147,50 @@ contract EigenDARegistryCoordinator is
...
+            uint96 operatorStake = stakeRegistry.getCurrentStake(operatorList[i], quorumNumber);
+            if (operatorStake < lowestStake) {
+                lowestStake = operatorStake;
+                operatorToChurn = operatorList[i];
             }
         }
+
+        // Deregister the operator with the lowest stake
+        bytes memory quorumNumbers = new bytes(1);
+        quorumNumbers[0] = bytes1(uint8(quorumNumber));
```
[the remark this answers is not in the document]
pakim249CAL: I think it's better the current way since it's more explicit about parsing out the bytes, rather than relying on abi encoding behavior.

contracts/src/core/EigenDARegistryCoordinator.sol
```diff
@@ -150,91 +147,50 @@ contract EigenDARegistryCoordinator is
         }).numOperatorsPerQuorum;
 
         // For each quorum, validate that the new operator count does not exceed the maximum
-        // (If it does, an operator needs to be replaced -- see `registerOperatorWithChurn`)
-        for (uint256 i = 0; i < quorumNumbers.length; i++) {
+        // If it does, churns an operator via an exhaustive search through the operator set.
```
samlaf: whats the policy it uses? churns the operator with lowest stake?
recommend mentioning it in this comment instead of just saying "an operator"

contracts/src/core/EigenDARegistryCoordinator.sol
```diff
@@ -150,91 +147,50 @@ contract EigenDARegistryCoordinator is
...
-     * @param quorumNumbers is an ordered byte array containing the quorum numbers being registered for
-     * @param params contains the G1 & G2 public keys of the operator, and a signature proving their ownership
-     * @param operatorKickParams used to determine which operator is removed to maintain quorum capacity as the
-     * operator registers for quorums
-     * @param churnApproverSignature is the signature of the churnApprover over the `operatorKickParams`
-     * @param operatorSignature is the signature of the operator used by the AVS to register the operator in the delegation manager
-     * @dev `params` is ignored if the caller has previously registered a public key
-     * @dev `operatorSignature` is ignored if the operator's status is already REGISTERED
-     */
+    /// @notice Deprecated function. Use `registerOperator` instead.
```
samlaf: Might be good to add why its deprecated, explaining how we've moved from offchain churning to onchain. Also mention that we're keeping this function temporarily for backwards compatibility until operators have all upgraded to using the new offchain code, and then we'll be free to delete this function.

contracts/src/core/EigenDARegistryCoordinator.sol
```diff
@@ -940,33 +804,13 @@ contract EigenDARegistryCoordinator is
...
-                    OPERATOR_CHURN_APPROVAL_TYPEHASH,
-                    registeringOperator,
-                    registeringOperatorId,
-                    operatorKickParams,
-                    salt,
-                    expiry
-                )
-            )
-        );
+    /// @notice Deprecated function.
```
samlaf, line 808: nit: think its good in general to add info as to why its deprecated and when we plan to remove it totally.

samlaf: also makes sure someone doesn't just come in and delete it accidentaly before we have finished the offchain migration period.

pakim249CAL: I don't actually know when we plan to remove it entirely. Have we  defined an offchain migration period? Is it necessary to yet? Maybe this is a question for whoever is implementing the off-chain changes. In the meanwhile, I will try to include something more generic.

pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1963/commits/f8cc53f3fe4f8721e0186e1fe2d757c21f6385b9

contracts/src/core/EigenDARegistryCoordinator.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1963/commits/f8cc53f3fe4f8721e0186e1fe2d757c21f6385b9

contracts/src/core/EigenDARegistryCoordinator.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/1963/commits/f8cc53f3fe4f8721e0186e1fe2d757c21f6385b9

samlaf (approved):

contracts/src/core/EigenDARegistryCoordinator.sol
ian-shim: Are we still keeping these conditions?

https://github.com/Layr-Labs/eigenda/blob/master/operators/churner/churner.go#L234

https://github.com/Layr-Labs/eigenda/blob/master/operators/churner/churner.go#L251

pakim249CAL: I cannot fully speak to whether those conditions should be kept or not. However, one of the goals of this PR was to avoid needing to make these kinds of changes in the off-chain code.

There is no difference in whether registerOperator or registerOperatorWithChurn is called. And so, I think any changes to these lines of code should be considered alongside the full set of offchain changes needed to eliminate old churning logic.

PR #1964, opened 2025-09-09
feat(payments): Add reservation remaining capacity metric

litt3:
- Adds reservation remaining capacity metric
- Makes necessary underlying changes to support the new metric
- Fixes a bug where the minimum number of billable symbols wasn't being respected by the `ReservationLedger`

### DIFF
--- a/api/clients/v2/metrics/accountant.go
+++ b/api/clients/v2/metrics/accountant.go
@@ -68,6 +68,8 @@ func (m *AccountantMetrics) RecordCumulativePayment(accountID string, wei *big.I
 }
 
 func (m *AccountantMetrics) RecordReservationPayment(accountID string, remainingCapacity float64) {
+	// We are expecting at most 100s of reservation, so the number of accountIDs (prom dimensions) won't be excessively
+	// large
 	m.ReservationRemainingCapacity.WithLabelValues(accountID).Set(remainingCapacity)
 }
 
--- a/api/clients/v2/metrics/accountant.go
+++ b/api/clients/v2/metrics/accountant.go
@@ -17,14 +17,14 @@ var (
 
 type AccountantMetricer interface {
 	RecordCumulativePayment(accountID string, wei *big.Int)
-	RecordReservationPayment(accountID string, remainingCapacity float64)
+	RecordReservationPayment(remainingCapacity float64)
 
 	Document() []metrics.DocumentedMetric
 }
 
 type AccountantMetrics struct {
 	CumulativePayment            *prometheus.GaugeVec
-	ReservationRemainingCapacity *prometheus.GaugeVec
+	ReservationRemainingCapacity prometheus.Gauge
 
 	factory *metrics.Documentor
 }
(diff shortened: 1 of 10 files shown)


api/clients/v2/metrics/accountant.go
```diff
@@ -57,6 +67,10 @@ func (m *AccountantMetrics) RecordCumulativePayment(accountID string, wei *big.I
 	m.CumulativePayment.WithLabelValues(accountID).Set(gweiFloat64)
 }
 
+func (m *AccountantMetrics) RecordReservationPayment(accountID string, remainingCapacity float64) {
+	m.ReservationRemainingCapacity.WithLabelValues(accountID).Set(remainingCapacity)
```
samlaf: maybe add a note that we are only expecting max 100s of reservation, so the amount of accountID (prom dimensions) won't be excessively large

core/payments/clientledger/client_ledger.go
```diff
@@ -154,6 +154,8 @@ func (cl *ClientLedger) debitReservationOnly(
 		panic(fmt.Sprintf("reservation debit failed: %v", err))
 	}
 
+	cl.accountantMetricer.RecordReservationPayment(cl.accountID.Hex(), remainingCapacity)
```
samlaf: nit: could it make sense to do this inside reservationLedger.Debit() instead?

samlaf (dismissed):

api/clients/v2/metrics/accountant.go
litt3: Hmm. Given this is only intended to be run on a client, it probably doesn't even make sense to report an `accountID` at all. There will always only be ONE account

core/payments/clientledger/client_ledger.go
litt3: Since these metrics are for clients only, they cannot be pushed down into the `ReservationLedger`, which will also run on dispersers and validators

api/clients/v2/metrics/accountant.go
litt3: 9f13783a458a5a9855cec2516f29ea2c32803c7a

Went ahead and changed to a gauge

api/clients/v2/metrics/accountant.go
```diff
@@ -17,12 +17,14 @@ var (
...
 	RecordCumulativePayment(accountID string, wei *big.Int)
+	RecordReservationPayment(remainingCapacity float64)
 
 	Document() []metrics.DocumentedMetric
 }
 
 type AccountantMetrics struct {
-	CumulativePayment *prometheus.GaugeVec
+	CumulativePayment            *prometheus.GaugeVec
+	ReservationRemainingCapacity prometheus.Gauge
```
cody-littley, line 27: Why is one a pointer and one not?

api/clients/v2/metrics/accountant.go
```diff
@@ -69,6 +81,9 @@ var NoopAccountantMetrics AccountantMetricer = new(noopAccountantMetricer)
 func (n *noopAccountantMetricer) RecordCumulativePayment(_ string, _ *big.Int) {
 }
 
+func (n *noopAccountantMetricer) RecordReservationPayment(_ float64) {
+}
+
```
cody-littley, line 86: I don't mind if you leave this alone since there is pre-existing code. I usually don't create mock metrics instances, I just make it so that metrics functions become no-ops if `this` is `nil`.

api/clients/v2/metrics/accountant.go
litt3: `GaugeVec` is a struct, but `Gauge` is an interface

api/clients/v2/metrics/accountant.go
litt3: Makes sense. Let's leave as-is for now, to avoid changing pre-existing code

PR #1970, opened 2025-09-10
feat(payments): Chunk fetching of PaymentVault reservation updates

litt3:
- Allows us to configure a maximum number of accountIDs to fetch updates for in a given call from the reservation vault monitor
- Addresses the concern brought up [here](https://github.com/Layr-Labs/eigenda/pull/1952#discussion_r2331066327)
- The corresponding change for the on demand monitor will be made in a subsequent PR, once this approach has gone through review


core/payments/reservation/reservation_vault_monitor.go
```diff
@@ -101,6 +97,54 @@ func (vm *ReservationVaultMonitor) refreshReservations(ctx context.Context) erro
...
+// Fetches reservations from the PaymentVault. If number of accountIDs exceeds configured rpcBatchSize, multiple RPC
+// calls will be made in series to fetch all reservation data. If rpcBatchSize is configured to be 0, all data
+// will be fetched in a single call, no matter how many accounts are passed in.
+func (vm *ReservationVaultMonitor) fetchReservations(
+	ctx context.Context,
+	accountIDs []gethcommon.Address,
+) ([]*bindings.IPaymentVaultReservation, error) {
+	if len(accountIDs) == 0 {
+		return nil, nil
+	}
```
samlaf: is this really what we want? or is this a bug?

core/payments/reservation/reservation_vault_monitor.go
```diff
@@ -101,6 +97,54 @@ func (vm *ReservationVaultMonitor) refreshReservations(ctx context.Context) erro
...
+			if end > len(accountIDs) {
+				end = len(accountIDs)
+			}
+			accountBatches = append(accountBatches, accountIDs[i:end])
+		}
+	}
+
+	allReservations := make([]*bindings.IPaymentVaultReservation, 0, len(accountIDs))
+	for batchIndex, batch := range accountBatches {
+		newReservations, err := vm.paymentVault.GetReservations(ctx, batch)
```
samlaf: nit: this would be a good usecase for errgroup no? Fetch everything in parallel, if anything errors error the whole thing since we need all the accounts anyways.

samlaf (approved):

core/payments/reservation/reservation_vault_monitor.go
litt3: You're right, this shouldn't be here. It's being checked that `accountIDs != 0` in `refreshReservations`, prior to calling `fetchReservations`. 

I removed it
cd82cc02a50bedbbcf034e02368fd080fa95c059

core/payments/reservation/reservation_vault_monitor.go
litt3: bc278dd92fde8f5c57bcd28032363cf64b80cfc8

👍 Done. Feels maybe a bit overengineered, but not too bad to reason about, and makes this fully future proof

core/payments/reservation/reservation_vault_monitor.go
samlaf: hmm but prob still better to keep just in case no? I like to have checks in both caller in callee, following https://tigerbeetle.com/blog/2023-12-27-it-takes-two-to-contract/

But perhaps if 0 len accountIDs is passed itll just work fine anyways since the for loop wont do anything?

core/payments/reservation/reservation_vault_monitor.go
samlaf: lgtm. My only issue with this now is that if someone uses public rpc with rate limits and it sends a gazillion number of requests at once, then it'll get rate limited and never end up creating the whole array. :\ 
I guess nothing is ever "fully future proof" lol

core/payments/reservation/reservation_vault_monitor.go
litt3: > But perhaps if 0 len accountIDs is passed itll just work fine anyways since the for loop wont do anything?

That's right. If empty accounts _did_ manage to get passed in, the whole thing would just be a noop

PR #1972, opened 2025-09-10
refactor(encoding): introduce parametrized_verifier

samlaf:
Miscellaneous `encoding/` improvement PR.

Contains a bunch of small improvements, but the main one is that we have a parametrized_prover, but not the equivalent parametrized_verifier.
Created this, and refactored into its own file. refactored some methods to be private instead of public.
Also had to fix a bunch of tests to use the public methods instead.

Also includes a random little cleanup of the unused stuff in encoding/test.


encoding/README.md
dmanc, line 1: We should improve this, no action needed in this PR though

encoding/test/main.go
```diff
@@ -12,44 +12,15 @@ import (
 	"github.com/Layr-Labs/eigenda/encoding/kzg"
 	"github.com/Layr-Labs/eigenda/encoding/kzg/prover"
 	"github.com/Layr-Labs/eigenda/encoding/kzg/verifier"
-	"github.com/Layr-Labs/eigenda/encoding/rs"
 
 	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
 )
 
+// TODO(samlaf): is anyone using this...? Can we delete it?
```
dmanc: I don't think this is being ran on our tests. I would vote to delete it, might have been useful during the development period of the library.

encoding/kzg/verifier/parametrized_verifier.go
```diff
@@ -0,0 +1,86 @@
...
+	"github.com/consensys/gnark-crypto/ecc/bn254"
+	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
+)
+
+type ParametrizedVerifier struct {
+	*kzg.KzgConfig
+	Srs *kzg.SRS
+
+	Fs *fft.FFTSettings
+	Ks *kzg.KZGSettings
```
dmanc: Do we need KZG Setting? Like it seems like it just has FFTSettings and SRS.

encoding/kzg/verifier/parametrized_verifier.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+
+	err = verifyFrame(frame, v.Ks, commitment, &v.Ks.ExpandedRootsOfUnity[j], &g2Atn)
+	if err != nil {
+		return fmt.Errorf("VerifyFrame: %w", err)
+	}
+	return nil
+}
+
+// Verify function assumes the Data stored is coefficients of coset's interpolating poly
```
dmanc, line 42: ChatGPT 5 gave me this:

```
// verifyFrame checks the KZG opening for the frame at point x.
//
// Assumptions:
//   - frame.Coeffs are the coefficients (little-endian by degree) of the
//     coset’s interpolating polynomial I(X) in Fr.
//   - ks.Srs.G1[:n] provides [s^0]₁,…,[s^{n-1}]₁, where n = len(frame.Coeffs).
//
// Steps (high level):
//   1) Compute x^n, where n = len(frame.Coeffs).
//   2) Form [x^n]₂ and then [s^n − x^n]₂.
//   3) Commit to I(s) in G1 via multiexp against the SRS: [I(s)]₁.
//   4) Compute [commitment − I(s)]₁.
//   5) Verify the pairing equation:
//
//        e([commitment − I(s)]₁, g₂) == e([proof]₁, [s^n − x^n]₂)
//
//      (equivalently: e(−[commitment − I(s)]₁, g₂) * e([proof]₁, [s^n − x^n]₂) = 1_T)
//
// Returns nil on success; otherwise wraps the underlying error.
```

dmanc: @bxue-l2 pls verify lol

encoding/kzg/verifier/verifier.go
```diff
@@ -215,9 +199,9 @@ func (v *Verifier) VerifyFrames(
 
 	for ind := range frames {
 		err = verifier.VerifyFrame(
-			(*bn254.G1Affine)(commitments.Commitment),
 			frames[ind],
 			uint64(indices[ind]),
+			(*bn254.G1Affine)(commitments.Commitment),
 			params.NumChunks,
```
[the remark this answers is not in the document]
samlaf: not sure what this is saying... clearly its wrong, but is copilot on to something? is index really a frameIndex (like I renamed it to)? Some places are passing in a chunk index it seems... are chunk index and frame index the same thing?

encoding/test/main.go
samlaf: [7447795](https://github.com/Layr-Labs/eigenda/pull/1972/commits/74477958814baee3771ecebaaef35c6e0b82a2f7)

encoding/kzg/verifier/parametrized_verifier.go
samlaf: Good callout..! Ended up cleaning a lot more stuff thanks to this. Hope it's ok @bxue-l2 :D
[24fe439](https://github.com/Layr-Labs/eigenda/pull/1972/commits/24fe439fbb2491209fc290182fffc4cfb6e8672c)

encoding/kzg/verifier/parametrized_verifier.go
samlaf: ks no longer valid because passing srs directly now.
But otherwise I feel like these math eqns are already described in the code, and would be better suited for that.
I'd prefer a high level explanation of what this is doing, in words.

encoding/kzg/prover/prover.go
```diff
@@ -443,14 +437,15 @@ func (p *Prover) createGnarkBackendProver(params encoding.EncodingParams, fs *ff
...
-		Ks:                    ks,
 		KzgMultiProofBackend:  multiproofBackend,
 		KzgCommitmentsBackend: commitmentsBackend,
 	}, nil
 }
 
-func (p *Prover) createIcicleBackendProver(params encoding.EncodingParams, fs *fft.FFTSettings, ks *kzg.KZGSettings) (*ParametrizedProver, error) {
-	return CreateIcicleBackendProver(p, params, fs, ks)
+func (p *Prover) createIcicleBackendProver(
+	params encoding.EncodingParams, fs *fft.FFTSettings,
```
bxue-l2, line 440: alright, it does not seem we use ks at all in the prover

encoding/kzg/verifier/parametrized_verifier.go
bxue-l2: Fyi, this function is not used by validator, they most likely use batch verification. 
It is possible, the distributed retrivier is using this. but I haven't checked.

The reason is that there are 8192 frames per blob, it is pretty inefficient to verify 819 frames individually for 10% stake

samlaf: Yes its used by:
1. retrieval_client [here](https://github.com/Layr-Labs/eigenda/blob/24fe439fbb2491209fc290182fffc4cfb6e8672c/api/clients/retrieval_client.go#L207): could we use the universal eqn as well here?
2. its also used by the validator's retrieval_worker [here](https://github.com/Layr-Labs/eigenda/blob/24fe439fbb2491209fc290182fffc4cfb6e8672c/api/clients/v2/validator/internal/chunk_deserializer.go#L77) (I think to retrieve chunks in the background from other validators...? not sure): should this also be using the universal eqn then?

bxue-l2: If the sender of the chunks are operators, it makes sense to universally verify per operators batch.

bxue-l2: Btw, the comment above looks right. Useful to cite
1. https://dankradfeist.de/ethereum/2020/06/16/kate-polynomial-commitments.html
2. https://github.com/khovratovich/Kate/blob/master/Kate_amortized.pdf

encoding/kzg/verifier/parametrized_verifier.go
```diff
@@ -0,0 +1,84 @@
...
+	if err != nil {
+		return fmt.Errorf("GetLeadingCosetIndex: %w", err)
+	}
+
+	g2Atn, err := kzg.ReadG2Point(uint64(len(frame.Coeffs)), v.SRSOrder, v.G2Path)
+	if err != nil {
+		return fmt.Errorf("ReadG2Point: %w", err)
+	}
+
+	err = verifyFrame(frame, v.Srs, commitment, &v.Fs.ExpandedRootsOfUnity[j], &g2Atn)
```
bxue-l2, line 35: nice, it actually uses the right FS that has a large enough size of Roots of unities.

We have 2 FS, one for RS, one for KZG. The KZG one is larger than the FS for RS one. Important not to confuse both

encoding/kzg/verifier/verifier.go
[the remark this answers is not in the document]
bxue-l2: Bullish on AI. But Sam you are right

encoding/kzg/prover/prover.go
```diff
@@ -20,15 +20,16 @@ import (
 	_ "go.uber.org/automaxprocs"
 )
 
+// Prover
```
dmanc: nit: Could we add documentation here or just remove the comment.

samlaf: [31a17b4](https://github.com/Layr-Labs/eigenda/pull/1972/commits/31a17b457fe6c917dd46a1ca82875d6130c5038f)

encoding/kzg/verifier/parametrized_verifier.go
dmanc: Let's do comprehensive revamp of documentation in later PR, maybe once all the refactors are done.

PR #1975, opened 2025-09-11
feat(payments): Test on demand ledger

litt3:
- I was lacking some unit test coverage for the `OnDemandLedger`

### DIFF
--- a/core/payments/ondemand/on_demand_ledger_test.go
+++ b/core/payments/ondemand/on_demand_ledger_test.go
@@ -1,7 +1,6 @@
 package ondemand_test
 
 import (
-	"context"
 	"errors"
 	"math/big"
 	"testing"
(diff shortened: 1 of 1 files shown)


core/payments/ondemand/on_demand_ledger_test.go
```diff
@@ -0,0 +1,245 @@
...
+	"github.com/stretchr/testify/require"
+)
+
+func TestDebit(t *testing.T) {
+	t.Run("successful debit", func(t *testing.T) {
+		store, cleanup := createTestStore(t, "TestDebitSuccessful")
+		defer cleanup()
+
+		ledger, err := ondemand.OnDemandLedgerFromStore(
+			context.Background(), big.NewInt(1000), big.NewInt(1), 10, store)
```
samlaf:
```suggestion
			t.Context(), big.NewInt(1000), big.NewInt(1), 10, store)
```

samlaf: we gotta start telling LLMs to stop using context.Background() as much as possible

litt3: Can't blame the LLM for it, that's just me not knowing `t.Context` is a thing 🙃

litt3: 59b7323a07b6afa1692eeb93d6e6172ba9f561fe

litt3:
> I still like the table driven better even when its verbose like this. Might be me though... def won't force you to change it. Perhaps TDD is easier for external folks to get a quick glimpse of the coverage, but (in this case) its harder to maintain/understand the implementation of the test?

I personally tend to find TDD hard to understand, maybe just me. Will definitely keep an eye out for cases when it makes most sense, I'm sure there are other examples that are more cut and dry in favor of a table based design, which I normally don't stop to consider at all

PR #1976, opened 2025-09-11
feat: Add test/CLAUDE.md

dmanc:

Adds testing guidelines with common guidelines and testing patterns for the EigenDA codebase.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/test/CLAUDE.md
+++ b/test/CLAUDE.md
@@ -310,7 +310,7 @@ func setupTestEnvironment(t *testing.T) *Environment {
 }
 ```
 
-Using t.Helper() will make it so that ...
+Using t.Helper() will make it so that the stack trace will point to the actual test function that called it — not the helper itself.
 
 ### 7.2 Common Test Utilities
 


test/CLAUDE.md
```diff
@@ -0,0 +1,445 @@
...
+|-----------|-------------|
+| **Test Independence** | Each test must be self-contained and not rely on execution order. |
+| **Clear Failures** | Failure output should make it obvious what went wrong and where |
+| **Resource Cleanup** | Containers, files, and connections must be cleaned up deterministically |
+| **Subtests** | Use t.Run() for logical grouping and granular execution (go test -run) |
+| **Descriptive Names** | Name tests for the behavior under test, not the implementation |
+
+---
+
+## 2. Anti-Patterns & Fixes
```
litt3, line 20: Could we add checking error strings with `require.Contains` as a known antipattern? It's super brittle, and doesn't really add much. (side note- lots of string matching assertions is a pretty good indicator that the tests are AI written. for some reason agents LOVE this pattern)

test/CLAUDE.md
```diff
@@ -0,0 +1,445 @@
...
+
+| Anti-Pattern | Problem | Prefer |
+|--------------|--------------|-----------------|
+| Global test state | Interdependent tests; order sensitivity | Test-scoped setup/fixtures |
+| Silent failures | Hard to debug | Always include failure messages |
+| Missing cleanup | Resource leaks; flakiness | Use t.Cleanup() with timeouts |
+| Testing internals | Brittle/overfitted tests | Test behavior via public interfaces |
+| Fixed sleeps | Flaky under load | Polling with timeout/backoff |
```
litt3, line 28: The problem with polling+backoff is that it's super verbose... maybe if we had some good helper utilities available, that would improve things?

Sometimes I'll find myself doing something like sleeping some multiple of the time I expect something to take. So if an action should be done in 1 milli, I'll sleep for 10 millis. Not the most robust pattern, but gets the job done without a bunch of boilerplate

Thoughts on this?

test/CLAUDE.md
```diff
@@ -0,0 +1,445 @@
...
+    require.NotNil(t, reply, "reply should not be nil")
+}
+```
+
+## 4. Assertions
+
+### 4.1 require vs assert
+
+- **`require`**: Stop immediately on failure (use in setup or preconditions).
+- **`assert`**: Continue to gather multiple failures in one run.
```
litt3, line 168: I don't know that I see the argument for using two separate libraries for this.

I guess you're saying that `assert` lets all the assertions execute before failing the test. But does that really matter much? Failure is failure. IMO having multiple assertion libraries is too ugly to be worth it

test/CLAUDE.md
```diff
@@ -0,0 +1,445 @@
...
+container, err := StartContainer()
+require.NoError(t, err, "failed to start container")
+
+// Use assert for multiple checks that should all run
+assert.Equal(t, expected1, actual1, "first check failed")
+assert.Equal(t, expected2, actual2, "second check failed")
+assert.Equal(t, expected3, actual3, "third check failed")
+```
+
+### 4.2 Always Include Descriptive Messages
```
litt3: Do we want to _always_ require error messages? I tend to follow the pattern of only including a message when necessary, but could be convinced otherwise.

A lot of times the message is redundant... for example:

```
require.Equal(t, a, b, "a should equal b")
```

Is the string actually helping anything here?

test/CLAUDE.md
```diff
@@ -0,0 +1,445 @@
...
+    // ...
+}
+```
+
+### 6.2 Cleanup with Timeouts
+
+```go
+t.Cleanup(func() {
+    logger.Info("Stopping container")
+    ctx, cancel := context.WithTimeout(context.Background(), 30*time.Second)
```
litt3, line 288: Is `context.Background()` actually correct here? Is there no way to get access to `t.Context()`?

test/CLAUDE.md
```diff
@@ -0,0 +1,445 @@
...
+func TestMain(m *testing.M) {
+    // Must use context.Background() here - no test context available
+    ctx := context.Background()
+    container, err := startContainer(ctx)
+    // ...
+}
+
+func TestSomething(t *testing.T) {
+    // Use t.Context() in actual test functions
+    ctx := t.Context()
```
litt3, line 278: TIL- didn't know this was a thing

CLAUDE.md
litt3, line 1: Consider adding the `test` directory to the `### 2.2 Project Subdirectories` table

test/CLAUDE.md
dmanc: Kudos to @samlaf for introducing it to me.

test/CLAUDE.md
dmanc: So I ran into issues when I tried to use t.Context() but my current understanding is that t.Context() is cancelled after the test finishes and cleanup runs after that.

test/CLAUDE.md
dmanc: In this particular example I don't think messages like "a should equal b" would be good but something like "hash mismatch" gives useful context.

But yeah it depends on the message and if we end up with a "test nitpicker" we can simply ignore the suggestions or tune it further.

test/CLAUDE.md
dmanc: It's the same library just two different helper functions. @cody-littley Argues that we should highly prefer `require` over `assert`. There are really only certain cases where using an `assert` is more ergonomic e.g. something like:

```
    assert.NotEmpty(t, got.ID)
    assert.Equal(t, "alice@example.com", got.Email)
    assert.WithinDuration(t, time.Now(), got.CreatedAt, 2*time.Second)
```

So when this test fails I can see them all at once and fix them all at once.

test/CLAUDE.md
dmanc: Yeah this is something claude noticed as an anti pattern but agree with your point that we need better test utilities. Our package is actually quite light right now https://github.com/Layr-Labs/eigenda/blob/master/common/testutils/test_utils.go.

But suppose there was a test utility that already did this type of operation, it's hard to tell what test utilities exists right now. So we would probably want a test nitpicker that suggests some common utilities?

CLAUDE.md
dmanc: 75f5e8bc908d4a9603a7e9a028ec904e02e44897

test/CLAUDE.md
dmanc: 75f5e8bc908d4a9603a7e9a028ec904e02e44897

test/CLAUDE.md
dmanc: Changed the wording a bit f866b940906b5f9ec23f557eea5debba8bb798db

test/CLAUDE.md
litt3: I suppose the `AssertEventuallyX` methods probably cover a lot of these cases- I haven't really been considering these as available tools writing tests. I'm ok with the current wording. 

> So we would probably want a test nitpicker that suggests some common utilities?

I think this is a good idea. If we put a list of what test utilities are available in this file, claude would have a good idea of what to recommend

test/CLAUDE.md
litt3: Gotcha. I'm wondering if this should maybe be something left unsaid in this doc, then? The worry would be that claude will nitpick us with `use the other method!` comments, when it's really just a question of personal taste.

I'm not too opinionated I guess. But maybe Cody is, I'll leave this unresolved till he responds

test/CLAUDE.md
```diff
@@ -0,0 +1,457 @@
...
+
+## 1. Core Testing Principles
+
+| Principle | Description |
+|-----------|-------------|
+| **Test Independence** | Each test must be self-contained and not rely on execution order. |
+| **Clear Failures** | Failure output should make it obvious what went wrong and where |
+| **Resource Cleanup** | Containers, files, and connections must be cleaned up deterministically |
+| **Subtests** | Use t.Run() for logical grouping and granular execution (go test -run) |
+| **Descriptive Names** | Name tests for the behavior under test, not the implementation |
```
samlaf, line 16: I think https://go.dev/wiki/TableDrivenTests should be part of this list. IMO its WAY more important than using subtests (which can still be written in an unreadable way without TDTs)

test/CLAUDE.md
samlaf: I don't really like assert. for me the ingrained "assert" semantic is to panic and crash the process if assert is wrong, not to keep going.

I've never used test asserts,but guessing they are only useful for super long slow tests like inabox? Otherwise just require and fail one error at a time imo.

test/CLAUDE.md
```diff
@@ -0,0 +1,457 @@
...
+// Acceptable - self-explanatory
+require.NotNil(t, client)
+require.Len(t, results, 3)
+```
+
+---
+
+## 5. Organizing Tests with Subtests
+
+### 5.1 Table-Driven Tests
```
samlaf, line 207: imo these should be flipped. subtests are not important to highlight. TDT is important, and should be the main thing to focus on. All tests, whenever possible, should be TDT imo.

test/CLAUDE.md
```diff
@@ -0,0 +1,457 @@
...
+
+```go
+// Bad
+ctx := context.Background()
+
+// Good
+ctx := t.Context()
+```
+
+Exception: In TestMain, use context.Background() (no test context exists).
```
samlaf, line 266: nit: best is probably to add some known timeout, not just Background(), since if for example docker compose fails to start or something the test will just hang forever potentially.

PR #1978, opened 2025-09-11
perf(srs): embed g2PowerOf2 srs file into binaries

samlaf:
This is both of a perf and a simplification. My goal is to make our KZG config loading easier to understand.

The power of 2 points are currently loaded dynamically one at a time from a 1.8KiB file whenver needed. This doesn't make sense, and its a lot more hassle, and the powerOf2 file needs to be carried around. 1.8KiB is nothing, so better to just embed in all binaries that need it and call it a day. Multiframe verification will also run faster thanks to this.

Deprecated the flags to load these dynamically, and removed from configs everywhere.

### DIFF
--- a/resources/srs/srs.go
+++ b/resources/srs/srs.go
@@ -22,7 +22,7 @@ var serializedG2PowerOf2Data []byte
 var G2PowerOf2SRS []bn254.G2Affine
 
 func init() {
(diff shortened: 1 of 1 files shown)


resources/srs/srs.go
```diff
@@ -0,0 +1,37 @@
+package srs
+
+import (
+	_ "embed"
+	"fmt"
+
+	"github.com/consensys/gnark-crypto/ecc/bn254"
+)
+
+//go:embed g2.point.powerOf2
```
bxue-l2, line 10: interesting

PR #1983, opened 2025-09-12
docs(payments): Document new payment system

litt3:
- Provides documentation for the new payment system


docs/spec/src/protocol/payments/payment_system.md
```diff
@@ -0,0 +1,178 @@
...
+    - `minNumSymbols`: dispersals smaller than this threshold are billed as if they were `minNumSymbols` in size
+    - `pricePerSymbol`: the price per symbol (in wei) for on-demand payments
+
+## 2. Payment Methods
+
+### 2.1 Reservation Payments
+
+- Reservations provide guaranteed bandwidth for a specified time period.
+   - Users reserve capacity in advance, and must "use it or lose it".
+   - Reservations are procured out-of-band, through EigenDA
```
cody-littley: I suggest checking with Kydo if this is the correct way of phrasing it

```suggestion
   - Reservations are procured out-of-band, through Eigen Labs
```

docs/spec/src/protocol/payments/payment_system.md
```diff
@@ -0,0 +1,178 @@
...
+   - Each validator keeps track of the dispersals from each user account, and will reject dispersals if the user doesn't
+   have enough capacity
+   - Clients keep a local reckoning of their own reservation usage, so that they can stay within the bounds of their
+   reserved bandwidth
+   - Dispersers also keep a local reckoning of client reservation usage, but a malicious client can bypass this check
+   by intentionally dispersing too much data spread out over multiple dispersers. From the perspective of any given
+   disperser, the client is within reservation limits. But in total, the client is over the limit. This isn't a problem,
+   because validator nodes will catch the misbehavior. By having dispersers keep track of reservation usage, we are 
+   imposing a limit on how severely a client can misbehave in this way: in a system with N dispersers, a malicious
+   client can disperse at most N * reservation rate
```
cody-littley: Nit, many lines end without a ".", even though they are otherwise complete sentences. Probably something claude can fix for you.

docs/spec/src/protocol/payments/payment_system.md
```diff
@@ -0,0 +1,178 @@
...
+validator are naturally smoothed out.
+- If a dishonest client tries to disperse more data than allowed, the behavior will be permitted by validators for
+a short time, but eventually even the larger validator bucket will fill. At that point, validators will limit new
+dispersals from the dishonest client to the rate of the reservation, and no additional dispersals may be stolen.
+- The capacity difference between client and validator buckets must be chosen to accommodate the maximum
+expected latency of the system. Specifically:
+`validatorBucketCapacity - clientBucketCapacity = reservationRate * maxSystemLatency`
+- This ensures that honest clients operating at full capacity won't be rejected due to timing discrepancies.
+- Proposed bucket sizing (actual configuration may vary): Client buckets will use 1 minute duration while validator
+buckets use 6 minutes, accommodating up to 5 minutes of system latency.
```
cody-littley, line 71: Yikes, that's a big burst window.

docs/spec/src/protocol/payments/payment_system.md
```diff
@@ -0,0 +1,178 @@
...
+The cumulative payment is a field set in the PaymentHeader by the client when making a dispersal. It represents the
+total cost (in wei) of all previous dispersals, plus the new dispersal.
+
+- **Historical context:** In a prior payments implementation, the cumulative payment field included by the client in
+  the PaymentHeader had to be monotonically increasing, and the disperser would verify that each new cumulative payment
+  received exceeded the previous one by at least the cost of the new blob. This severely limited concurrency, since
+  clients had to make sure that all on-demand dispersals were handled by the disperser in strict order. In practice,
+  that meant waiting for the entire network roundtrip, for dispersal N to be confirmed before submitting dispersal N+1.
+- **Current implementation:** The system has been simplified to improve concurrency:
+   - Clients still populate the `cumulative_payment` field with their local calculation of cumulative payment
```
cody-littley: I suggest adding something along the lines of `remove cumulative_payment field` to this document: https://linear.app/eigenlabs/issue/EGDA-2004/protobufgrpc-change-wishlist

docs/spec/src/protocol/payments/payment_system.md
litt3: This was a shot from the hip on my part, we can put anything we want here. OR, we can put nothing at all here, and kick the can down the road.

If you had to choose a starting point for these values, what would you use?

docs/spec/src/protocol/payments/payment_system.md
litt3: Recalling when we discussed this with @mooselumph , we were going to leave this here for now, just in case we need to use it in the future.

As it stands, we need _some_ field to indicate whether you're using on-demand or reservation payments. We might decide to change the _type_ of the field at some point in the future

docs/spec/src/protocol/payments/payment_system.md
```diff
@@ -0,0 +1,178 @@
+# EigenDA Payment System
+
+## 1. Overview
+
+The EigenDA payment system allows users to pay for blob dispersals through two methods: reservations and on-demand
+payments. All payment logic is implemented in the [`core/payments`](../../../../../core/payments/) package.
+
+**Key Concepts:**
+- Blob sizes are measured in *symbols*, where each symbol contains 32 bytes of data
```
samlaf: nit: each symbol contains 31 bytes of payload data. Not sure if worth mentioning here though..

docs/spec/src/protocol/payments/payment_system.md
```diff
@@ -0,0 +1,178 @@
...
+- Blob sizes are measured **post-blob encoding**: user payloads expand to some degree during blob encoding
```
samlaf: is "to some degree" used colloquially here? sounds like its expanding to some polynomial degree. maybe just remove?

docs/spec/src/protocol/payments/payment_system.md
```diff
@@ -0,0 +1,178 @@
...
+
+### 2.1 Reservation Payments
+
+- Reservations provide guaranteed bandwidth for a specified time period.
+   - Users reserve capacity in advance, and must "use it or lose it".
+   - Reservations are procured out-of-band, through EigenDA
+- The system uses a [leaky bucket algorithm](../../../../../core/payments/reservation/leaky_bucket.go) to manage usage:
+  symbols are added to the bucket each time a blob is dispersed, and these leak out over time. A user can only make a
+  dispersal if the leaky bucket has available capacity.
+   - The total capacity of the leaky bucket is defined in terms of a *duration*. The size of the bucket in symbols is
```
samlaf: total capacity is not in duration, its in symbols/s.
It is parametrized by BOTH duration AND rate.

docs/spec/src/protocol/payments/payment_system.md
```diff
@@ -0,0 +1,178 @@
...
+
+#### 2.1.1 Source of Truth
+
+- Validator nodes are the source of truth for reservation usage.
+   - Each validator keeps track of the dispersals from each user account, and will reject dispersals if the user doesn't
+   have enough capacity
+   - Clients keep a local reckoning of their own reservation usage, so that they can stay within the bounds of their
+   reserved bandwidth
+   - Dispersers also keep a local reckoning of client reservation usage, but a malicious client can bypass this check
+   by intentionally dispersing too much data spread out over multiple dispersers. From the perspective of any given
```
samlaf, line 47: I thought we didn't want our docs to include multiple dispersers since its not currently a thing?

docs/spec/src/protocol/payments/payment_system.md
samlaf: isnt our p90 (?) system latency around 10sec?

samlaf (commented):
First pass, didn't read the whole thing yet. Generally very happy to see this written down! Thinking some subset of this, with user guidelines, might be worth putting on our docs page to update https://docs.eigencloud.xyz/products/eigenda/core-concepts/payments

docs/spec/src/protocol/payments/payment_system.md
litt3: IMO, you can't fully explain the implementation decisions here without touching on distributed dispersal in some areas, since the design was built to satisfy distributed dispersal requirements

> we didn't want our docs to include multiple dispersers since its not currently a thing?

When was this decided? IMO it's not a problem to start referencing an upcoming feature in docs, when the docs cover designs created to support the upcoming feature

samlaf: You're right! nvm, I was probably thinking of the protobuf files instead.

docs/spec/src/protocol/payments/payment_system.md
litt3: We need to be much more confident than p90 for this. p99.9+ I'd say

docs/spec/src/protocol/payments/payment_system.md
litt3: 968aca707e2eb686963c489fb0f78d39fcf0e11e

docs/spec/src/protocol/payments/payment_system.md
litt3: e9e026d46b063f7127fdeb1a676b9d73f8b1ac2b

docs/spec/src/protocol/payments/payment_system.md
litt3: b279c02ab587272e8b5e6733ec8807f59e0d64ba

tweaked wording

docs/spec/src/protocol/payments/payment_system.md
litt3: 94d31a166c709c33ff55d6bafae6585e7ad6e574

docs/spec/src/protocol/payments/payment_system.md
litt3: It is defined in code by passing in a duration, but I see how this could be confusing. Reworded.

005a8438a6d29128d9bb1027c5ad47a71cc18493

docs/spec/src/protocol/payments/payment_system.md
cody-littley: Under "normal" circumstances, it's nearly impossible for latency to exceed one minute. We stop attempting to collect signatures after 50 seconds. So if it takes that long, the dispersal just fails.

However, due to lack of backpressure, it's theoretically plausible to build up a very large backlog of work. I've seen that happen when the controller crashes (the API servers keep adding to the queue without pausing), or when the encoders can't keep up.

When we build up a backlog of this size, it's theoretically possible to have arbitrarily long latencies. I'm talking hours or more. I've actually seen this sort of behavior during extreme test conditions.

samlaf: Can we/should we just put an upper bound on queue length and start dropping requests after that? I feel like we'eve talked about this a million times, but never implemented it?

docs/spec/src/protocol/payments/payment_system.md
```diff
@@ -0,0 +1,179 @@
...
+  in the future if needed.
+
+## 3. Client Payment Strategy
+
+### 3.1 Payment Header
+
+Each dispersal request includes a [PaymentHeader](../../../../../api/proto/common/v2/common_v2.proto#L111) containing:
+- `account_id`: Ethereum address identifying the payment account
+- `timestamp`: Nanosecond UNIX timestamp (serves as nonce)
+- `cumulative_payment`: Variable-length big-endian uint for on-demand dispersal (or empty for reservation dispersal)
```
samlaf, line 158: nit: prob good to specify that its at most a uint256, since the source of truth is found onchain.

docs/spec/src/protocol/payments/payment_system.md
```diff
@@ -0,0 +1,179 @@
...
+  to populate this field with meaningful values. This preserves the option to reintroduce cumulative payment validation
+  in the future if needed.
+
+## 3. Client Payment Strategy
+
+### 3.1 Payment Header
+
+Each dispersal request includes a [PaymentHeader](../../../../../api/proto/common/v2/common_v2.proto#L111) containing:
+- `account_id`: Ethereum address identifying the payment account
+- `timestamp`: Nanosecond UNIX timestamp (serves as nonce)
```
samlaf, line 157: @bxue-l2 had questions from megaeth wrt nonce vs timestamp. Just writing this comment to flag it to him, no action item.

samlaf (approved):

docs/spec/src/protocol/payments/payment_system.md
litt3: Technically only the _total deposits_ are limited to max uint256... you can choose an arbitrarily high cumulative payment, and it would be a valid message. It's just not possible to deposit enough funds to cover the dispersal 

In any case, not sure it's necessary to specify at this level of documentation. Max uint256 wei exceeds the total amount of eth that exists, so users will likely hit practical limits before arriving at the technical limits 🙃

docs/spec/src/protocol/payments/payment_system.md
litt3: > Can we/should we just put an upper bound on queue length and start dropping requests after that?

Yes. I have it on my todos to make this change.

PR #1985, opened 2025-09-13
feat(proxy): All user APIs disabled by default with enablement through `EIGENDA_PROXY_APIS_TO_ENABLE` env var

ethenotethan:
Given the introduction of an Arbitrum specific CustomDA server, we need to update how we "toggle" different APIs across the proxy application. 

**Proposed Taxonomy**
At a high level there are three supported servers that a user can choose to turn on:
- `metrics`
- `rest` --> `APIs`
- `arb`

Under the `rest` API server, there are 4 different modes or API expressions that can be enabled:
- `standard`
- `op-generic`
- `op-keccak`
- `admin`

Enablement of any of the following REST APIs would result in the construction of the REST Server (assuming a user didn't just provide `admin`). **IF a Rest ALTDA API is triggered that is disabled then the server returns a 403 unauthorized status code**

This is a **breaking change** for customers given:
- All service related codepaths are **disabled by default** on the proxy
- Two flags are now deprecated (`METRICS_ENABLED`, `SERVER_API_ENABLED`)

This forces customers to explicitly state which servers and REST handler processing paths they need for their use case. From a security perspective we're by default providing a minimal attack surface which customers have to be conscious of.  

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proxy/cmd/server/entrypoint.go
```diff
@@ -74,21 +75,27 @@ func StartProxyService(cliCtx *cli.Context) error {
...
-			log.Error("failed to stop REST ALT DA server", "err", err)
-		}
+		defer func() {
+			if err := restServer.Stop(); err != nil {
+				log.Error("failed to stop REST ALT DA server", "err", err)
+			}
 
-		log.Info("Successfully shutdown REST ALT DA server")
-	}()
+			log.Info("Successfully shutdown REST ALT DA server")
```
samlaf: this needs to be inside else branch right?

api/proxy/config/enabled_apis/enabled_apis.go
```diff
@@ -0,0 +1,172 @@
...
+		apis,
+	}
+}
+
+func (e EnabledAPIs) has(api API) bool {
+	return slices.Contains(e.apis, api)
+}
+
+// API represents the different APIs that can be exposed on the proxy application
+type API uint8
```
samlaf: why not use string?

api/proxy/config/enabled_apis/cli.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+func CLIFlags(category string, envPrefix string) []cli.Flag {
+	return []cli.Flag{&cli.StringSliceFlag{
+		Name: EnabledAPIsFlagName,
+		Usage: fmt.Sprintf("Which proxy application APIs to enable. supported options are "+
+			"%s, %s, %s, %s, %s, %s", Admin.ToString(), StandardCommitment.ToString(),
+			OpGenericCommitment.ToString(), OpKeccakCommitment.ToString(),
+			ArbCustomDAServer.ToString(), MetricsServer.ToString()),
+		Value:    cli.NewStringSlice(),
```
samlaf: by default no api is turned on? why not make it required instead?

api/proxy/config/enabled_apis/cli.go
```diff
@@ -0,0 +1,39 @@
...
+	return apis
+}
+
+func CLIFlags(category string, envPrefix string) []cli.Flag {
+	return []cli.Flag{&cli.StringSliceFlag{
+		Name: EnabledAPIsFlagName,
+		Usage: fmt.Sprintf("Which proxy application APIs to enable. supported options are "+
+			"%s, %s, %s, %s, %s, %s", Admin.ToString(), StandardCommitment.ToString(),
+			OpGenericCommitment.ToString(), OpKeccakCommitment.ToString(),
+			ArbCustomDAServer.ToString(), MetricsServer.ToString()),
```
samlaf: prefer to have a function defined on the API type to list all values. we do this for network types

api/proxy/config/enabled_apis/enabled_apis.go
```diff
@@ -0,0 +1,172 @@
...
+		return fmt.Errorf("expected at least one \"apis.enabled\" value to be provided")
+	}
+
+	if common.ContainsDuplicates(e.apis) {
+		return fmt.Errorf("enabled apis contains duplicate: %+v", e.apis)
+	}
+
+	if e.Metrics() && (!e.RestALTDA() && !e.ArbCustomDA()) {
+		return fmt.Errorf("metrics cannot be enabled unless `arb` and/or `rest` also is")
+	}
```
samlaf: nit: delete this and replace it with a simpler check "either REST or arb custom must be provided. having metrics turned on/off should be orthogonal, and not the main important point.

api/proxy/config/enabled_apis/enabled_apis.go
```diff
@@ -0,0 +1,172 @@
...
+
+	if e.RestALTDAWithAdmin() &&
+		(!e.RestALTDAOPGeneric() && !e.RestALTDAOPKeccak() && !e.RestALTStandard()) {
+		return fmt.Errorf("admin mode for REST ALTDA server cannot be enabled without also " +
+			"setting one of `op-generic`, `op-keccak`, `standard`")
+	}
```
samlaf: here also the admin is not the main point. main point is one of the data routes needs to be enabled.

api/proxy/config/app_config.go
```diff
@@ -47,11 +57,22 @@ func ReadAppConfig(ctx *cli.Context) (AppConfig, error) {
...
 	}
 
+	// NOTE: double passing of the `enabledAPIs` type to both the
+	//       the AppConfig and Rest server config is an abstraction leak.
+	//       given rest endpoint logic is guarded by the enabledAPIs it
+	//       makes sense to be like this. otherwise a separate env var
+	//       could be introduced that provides enablement toggles purely
+	//       for the REST server's code paths.
+	//
+	enabledAPIs := enabled_apis.ReadEnabledAPIs(ctx)
```
samlaf: hmm can't you embed the rest config inside the top level enabledAPIs config? what didn't work when you tried this?

api/proxy/servers/rest/handlers_cert.go
```diff
@@ -56,11 +61,21 @@ func (svr *Server) handleGetOPKeccakCommitment(w http.ResponseWriter, r *http.Re
 
 // handleGetOPGenericCommitment handles the GET request for optimism generic commitments.
 func (svr *Server) handleGetOPGenericCommitment(w http.ResponseWriter, r *http.Request) error {
+	if !svr.config.EnabledAPIs.RestALTDAOPGeneric() {
+		w.WriteHeader(http.StatusForbidden)
+		return fmt.Errorf("op-generic DA Commitment type detected but `op-generic` API is not enabled")
+	}
```
samlaf: guessing you've considered turning these routes off altogether in router.go, but opted to keep them with more explicit error msg?

ethenotethan: Precisely - otherwise a user wouldn't have good introspection into why their request was failing

api/proxy/cmd/server/entrypoint.go
ethenotethan: it's in the deferred function logic triggered at shutdown:
```
		defer func() {
			if err := restServer.Stop(); err != nil {
				log.Error("failed to stop REST ALT DA server", "err", err)
			}

			log.Info("Successfully shutdown REST ALT DA server")
		}()
```

api/proxy/servers/rest/cli.go
```diff
@@ -18,6 +22,22 @@ func withEnvPrefix(prefix, s string) []string {
 	return []string{prefix + "_" + s}
 }
 
+func DeprecatedCLIFlags(envPrefix string, category string) []cli.Flag {
+	return []cli.Flag{
+		&cli.StringSliceFlag{
```
litt3, line 27: Why did you choose to deprecate this flag, and add `APIS_TO_ENABLE` which seems like the same thing?

api/proxy/config/enabled_apis/enabled_apis.go
ethenotethan: agree with the first comment but why would admin exposed without not be the main check here for a proper api types expression for the REST ALT DA Server?

api/proxy/servers/rest/cli.go
ethenotethan: felt it was better suited using a plural env definition

api/proxy/config/app_config.go
ethenotethan: are you referring to updating the `enabledAPIs` struct to store notion of two slices (i.e, servers_enabled, rest_apis_enabled)?

api/proxy/config/enabled_apis/enabled_apis.go
ethenotethan: [590c2bf](https://github.com/Layr-Labs/eigenda/pull/1985/commits/590c2bf6949db20e430762e6210ed43e34919af5)

api/proxy/config/enabled_apis/cli.go
ethenotethan: eh I prefer this though because it complies with the logging format of key|value which specifically showcases enablement and gives a user direct introspection in all available REST API options

api/proxy/config/enabled_apis/cli.go
ethenotethan: [590c2bf](https://github.com/Layr-Labs/eigenda/pull/1985/commits/590c2bf6949db20e430762e6210ed43e34919af5)

ethenotethan: initially I didn't make it required since we don't actually have any required CLI flags for the proxy but enforce a notion of "required" when doing expressional config invariance

api/proxy/config/enabled_apis/enabled_apis.go
ethenotethan: [590c2bf](https://github.com/Layr-Labs/eigenda/pull/1985/commits/590c2bf6949db20e430762e6210ed43e34919af5)

api/proxy/config/enabled_apis/cli.go
ethenotethan: because if you make it required then all subcommand calls (e.g, generate metrics) from the binary will fail unless it's provided. iiuc there are no required fields across all proxy CLI flags

ethenotethan:
```
./bin/eigenda-proxy doc 
2025/09/16 21:11:40 maxprocs: Leaving GOMAXPROCS=16: CPU quota undefined
NAME:
   eigenda-proxy doc

USAGE:
   eigenda-proxy doc command [command options]

t=2025-09-16T21:11:40+0700 lvl=crit msg="Application failed" message="Required flag \"apis.enabled\" not set"
```

api/proxy/cmd/server/entrypoint.go
samlaf: I know, but my point is log.Info will run even if the log.Error branch was hit. I think we only one want OR the other right?

api/proxy/config/enabled_apis/cli.go
samlaf: wtf lol never realized. This is kind of dumb. I'm sure there's a way to disable global flags for a subcommand no?
Otherwise we might want to make starting the proxy another subcommand with its own flags.. and then make that subcommand default (also not sure if this is possible...)

Anyways all of this is doesnt need to happen in this PR of course.

api/proxy/config/enabled_apis/cli.go
samlaf: not following. I'm not saying NOT have it in this. I'm just saying to have a function which creates a list of all enum values, and then use that here.

The point is if we ever add another API value its very likely we'll just forget to update it here, so the flag docs will be outdated. But if we define that function right below where the enum values are defined, its much easier to remember to update it, and itll propagate everywhere.

api/proxy/config/enabled_apis/enabled_apis.go
samlaf: I'm just saying the check should instead enforce one of op-generic, op-keccak, OR standard to be turned on. ALWAYS.
Then if we have that you can turn on/off admin API orthogonally. We don't need this check.

api/proxy/cmd/server/entrypoint.go
ethenotethan: this is how it's done for all of servers rn though (probably shouldn't be this way)

api/proxy/config/app_config.go
samlaf: Yes. Actually just use embedde structs:
```
type RESTAPIs struct {
   admin AdminAPI
   opGeneric OPGenericAPI
   ...
}
type EnabledAPIs struct {
   RESTAPIs,
   ArbAPIs,
   m MetricsAPI
```

then EnabledAPIs contains all the apis as direct fields (via the struct embedding), but we can also ONLY pass RESTAPIs struct to REST handler, instead of the entire EnabledAPIs struct.

api/proxy/servers/rest/cli.go
samlaf: Oh can we rename APIsEnabledFlagName->DeprecatedAPIsEnabledFlagName then?

api/proxy/cmd/server/entrypoint.go
samlaf: I'm just saying we should change it everywhere to
```
		defer func() {
			if err := restServer.Stop(); err != nil {
				log.Error("failed to stop REST ALT DA server", "err", err)
			} else {
			   log.Info("Successfully shutdown REST ALT DA server")
                         }
		}()
```

api/proxy/config/enabled_apis/cli.go
ethenotethan: will set `Args` to false

api/proxy/config/enabled_apis/enabled_apis.go
ethenotethan: after discussing it makes the most sense to remove this conditional given there will likely be admin support in the future for ARB Custom DA

api/proxy/config/app_config.go
ethenotethan: 8bdf3fb070ad1ccc61af03d085e493574cc81fec

api/proxy/config/enabled_apis/enabled_apis.go
ethenotethan: 8bdf3fb070ad1ccc61af03d085e493574cc81fec

api/proxy/servers/rest/cli.go
ethenotethan: 302f7e5e1cb0ea1f28258f45510430a802b099b9

api/proxy/config/enabled_apis/cli.go
ethenotethan: 25c7d7c3b5e56ed2c6c370384c510e108e57ff1b

api/proxy/cmd/server/entrypoint.go
ethenotethan: [80c09c8](https://github.com/Layr-Labs/eigenda/pull/1985/commits/80c09c8d5432991272058306eba86fb77580a197)

api/proxy/config/enabled_apis/cli.go
ethenotethan: [f3dc62a](https://github.com/Layr-Labs/eigenda/pull/1985/commits/f3dc62a10772ed652d0da4593f6bc90c4560157b)

api/proxy/config/enabled_apis/cli.go
ethenotethan: actually have to revert: [0b3d315](https://github.com/Layr-Labs/eigenda/pull/1985/commits/0b3d31525bb1c84fd45d018fcfeea28eebf864d0)

ethenotethan: played around the CLI subcommand options (i.e, SkipFlagParse=true, Args=false) and couldn't find anything that resulted in a bypass. If I had to guess the flags we're providing are at the application level meaning and are always preprocessed despite what a subcommand's definition says

ethenotethan: [2f88d86](https://github.com/Layr-Labs/eigenda/pull/1985/commits/2f88d86d59d9e147234fd95842a0e82b5e36754d)

samlaf (approved):

PR #1987, opened 2025-09-14
Add graph node to testbed

dmanc:

Spins up graph node using testcontainers. Removes `inabox//subgraph.go`

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


inabox/Makefile
```diff
@@ -26,12 +26,14 @@ deploy-all:
...
-	./bin.sh stop-graph
 # Stop anvil
 	docker ps -q --filter "ancestor=ghcr.io/foundry-rs/foundry" | xargs -r docker stop 2>/dev/null || true
 # Stop localstack based on container name
 	docker ps -q --filter "ancestor=localstack/localstack" | xargs -r docker stop 2>/dev/null || true
-
+# Stop graph node stack based on container names
+	docker ps -q --filter "ancestor=graphprotocol/graph-node:v0.35.0" | xargs -r docker stop 2>/dev/null || true
+	docker ps -q --filter "ancestor=ipfs/kubo:v0.24.0" | xargs -r docker stop 2>/dev/null || true
+	docker ps -q --filter "ancestor=postgres:13" | xargs -r docker stop 2>/dev/null || true
```
samlaf, line 32: nit: we shouldnt need any of these things. when running inside test suite, they should be killed by test framework. When running via inabox, then I think its ryuk that's managing them? So we probably want to be using ryuk to GC them?

inabox/deploy/cmd/main.go
```diff
@@ -150,7 +150,24 @@ func chainInfra(ctx *cli.Context, config *deploy.Config) error {
...
+			HostAdminPort:  "8020",
+			HostIPFSPort:   "5001",
+			Logger:         logger,
+		})
+		if err != nil {
+			return fmt.Errorf("failed to start graph node: %w", err)
+		}
+		// Wait for Graph Node to be ready
+		fmt.Println("Waiting for Graph Node to be ready...")
+		time.Sleep(10 * time.Second)
```
samlaf: can we use testcontainers utils that wait for a port to be ready or something instead of just waiting 10 seconds?

dmanc: Removed logs, was already doing that with this
```
		WaitingFor: wait.ForAll(
			wait.ForListeningPort("8000/tcp"),
			wait.ForLog("Starting GraphQL HTTP server").WithStartupTimeout(90*time.Second),
		),
```

inabox/Makefile
dmanc: So inabox has two modes of operation: test or devnet.

Ryuk enabled in test mode (cleanup when test ends)
Ryuk disabled in devnet mode (no cleanup, need to run make stop-infra)

It might be possible to have Ryuk enabled in devnet mode and have it do the garbage collection, will take a look in a separate diff.

samlaf: i see... i thought in inabox mode Ryuk was just outliving the test process. Makes sense then if Ryuk is just not spun up at all.

inabox/tests/integration_suite_test.go
```diff
@@ -117,13 +118,21 @@ var _ = BeforeSuite(func() {
 	testConfig = deploy.NewTestConfig(testName, rootPath)
 
 	if testConfig.Environment.IsLocal() {
+		// Create a shared Docker network for all containers
+		nw, err := network.New(context.Background(),
+			network.WithDriver("bridge"),
+			network.WithAttachable())
+		Expect(err).To(BeNil(), "failed to create Docker network")
+		logger.Info("Created Docker network", "name", nw.Name)
```
samlaf: is this just going to create a random named docker network? Looks like you never delete it anywhere iiuc? So when test fails or inabox mode is launched, how do we delete these networks?

samlaf (approved):

inabox/tests/integration_suite_test.go
dmanc: Yeah seems random
```
Sep 19 11:54:09.125 INF tests/integration_suite_test.go:128 Created Docker network name=194e306d-8eed-4366-b077-3f602811607a
```

c87c7fbb413217e831d589ebe65ae92358cd32f5

PR #1992, opened 2025-09-15
feat: data structures for tracking signing rates

cody-littley:

Creates data structures for efficiently tracking validator signing rates. I will integrate these data structures into our code in a following PR.


core/signingrate/threadsafe_signing_rate_tracker.go
```diff
@@ -0,0 +1,363 @@
...
+		return nil, errors.New("signing rate tracker is shutting down")
+	case t.requests <- request:
+	}
+
+	// await the response
+	select {
+	case <-t.ctx.Done():
+		return nil, errors.New("signing rate tracker is shutting down")
+	case response := <-request.responseChan:
+		return response.result, nil
```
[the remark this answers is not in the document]
cody-littley: fixed

core/signingrate/signing_rate_bucket.go
```diff
@@ -0,0 +1,196 @@
...
+		return b.cachedProtobuf
+	}
+
+	start := uint64(b.startTimestamp.Unix())
+	end := uint64(b.endTimestamp.Unix())
+
+	quorumSigningRates := make([]*validator.QuorumSigningRate, len(b.signingRateInfo))
+
+	for quorumID, quorumInfo := range b.signingRateInfo {
+		validatorSigningRates := make([]*validator.ValidatorSigningRate, 0, len(b.signingRateInfo))
```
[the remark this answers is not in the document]
cody-littley: fixed

api/proto/validator/signing_rate.proto
```diff
@@ -0,0 +1,51 @@
...
+// Records information about validator signing rate during a time period.
+message ValidatorSigningRate {
+
+  // The unique identifier of the validator (i.e. the operator ID).
+  bytes id = 1;
+
+  // The number of signed by the validator during the period.
+  uint64 signed_batches = 2;
+
+  // The number of unsigned by the validator during the period.
```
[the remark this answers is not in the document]
cody-littley: fixed

api/proto/validator/signing_rate.proto
```diff
@@ -0,0 +1,49 @@
...
+  // Contains the sum of the time spent by the validator waiting for signing requests to be processed, in nanoseconds.
+  // Only batches that are signed are considered (i.e. if the validator does not succeed in signing a batch,
+  // the time spend in the attempt is not counted).
+  uint64 signing_latency = 6;
+}
+
+// Contains signing rate information about a specific quorum.
+message QuorumSigningRate {
+  // The unique identifier of the quorum. Technically quorums are stored in a a uint8,
+  // but using uint64 here to future-proof things.
```
litt3: > to future-proof things

By choosing uint64, it adds the burden of checking the field when handling the message. Do we actually need to future proof in this way?

core/signingrate/signing_rate_tracker.go
```diff
@@ -0,0 +1,75 @@
...
+// Tracks signing rates for validators and serves queries about signing rates.
+type SigningRateTracker interface {
+
+	// Get the signing rate for a validator over the specified time range. Start time is rounded forwards/backwards
+	// to the nearest bucket boundaries.
+	//
+	// Returned data threadsafe to read, but should not be modified.
+	GetValidatorSigningRate(
+		quorum core.QuorumID,
+		id core.OperatorID,
```
litt3: given there are different IDs in play, consider potentially renaming this parameter to `operatorID` (also similar parameters named `id`)

Up to you

core/signingrate/signing_rate_tracker.go
```diff
@@ -0,0 +1,75 @@
...
+	// last bucket.
+	//
+	// The intended use of this method is to set up a SigningRateTracker that mirrors a remote SigningRateTracker.
+	// The remote tracker is the source of truth, and this local tracker is just a cache. Periodically, get data
+	// from the remote tracker using GetSigningRateDump(), and then insert the data returned into this tracker using
+	// UpdateLastBucket().
+	//
+	// This operation doesn't mark a bucket as unflushed. A bucket is only marked as unflushed when it is modified,
+	// not when it is provided whole-sale from an external source.
+	UpdateLastBucket(now time.Time, bucket *validator.SigningRateBucket) // TODO should this accept a timestamp?
```
litt3: TODO

core/signingrate/signing_rate_bucket.go
```diff
@@ -0,0 +1,196 @@
...
+	info.SignedBatches += 1
+	info.SignedBytes += batchSize
+	info.SigningLatency += uint64(signingLatency.Nanoseconds())
+
+	b.cachedProtobuf = nil
+}
+
+// Report that a validator has failed to sign a batch of the given size.
+//
+// If the validator was previously Up, it will be marked as Down.
```
litt3: > If the validator was previously Up, it will be marked as Down.

I don't see anything in the impl related to this

core/signingrate/signing_rate_bucket.go
```diff
@@ -0,0 +1,196 @@
...
+// The resulting protobuf is a deep copy, and is therefore threadsafe to use concurrently with this SigningRateBucket.
+func (b *SigningRateBucket) ToProtobuf() *validator.SigningRateBucket {
+	if b.cachedProtobuf != nil {
+		return b.cachedProtobuf
+	}
+
+	start := uint64(b.startTimestamp.Unix())
+	end := uint64(b.endTimestamp.Unix())
+
+	quorumSigningRates := make([]*validator.QuorumSigningRate, len(b.signingRateInfo))
```
litt3:
```suggestion
	quorumSigningRates := make([]*validator.QuorumSigningRate, 0, len(b.signingRateInfo))
```

Looks like you're setting the length to what you wanted the capacity to be. Since you're appending, this will result in len(b.signingRateInfo) `nil` elements, followed by your actual elements. Thanks, golang footgun!

core/signingrate/signing_rate_bucket_test.go
```diff
@@ -0,0 +1,199 @@
...
+	"time"
+
+	"github.com/Layr-Labs/eigenda/api/grpc/validator"
+	"github.com/Layr-Labs/eigenda/common/testutils/random"
+	"github.com/Layr-Labs/eigenda/core"
+	"github.com/stretchr/testify/require"
+)
+
+// Returns true if two validator.ValidatorSigningRate messages are equal.
+func areSigningRatesEqual(a *validator.ValidatorSigningRate, b *validator.ValidatorSigningRate) bool {
```
litt3, line 16: missing comparison of unsigned batches

core/signingrate/signing_rate_tracker_impl.go
```diff
@@ -0,0 +1,299 @@
...
+	}
+
+	totalSigningRate := &validator.ValidatorSigningRate{
+		Id: id[:],
+	}
+
+	iterator, err := s.buckets.IteratorFrom(startIndex)
+	enforce.NilError(err, "should be impossible with a valid index")
+	for _, bucket := range iterator {
+		if bucket.startTimestamp.After(endTime) {
```
litt3:
```suggestion
		if !bucket.startTimestamp.Before(endTime) {
```

core/signingrate/signing_rate_tracker_impl.go
```diff
@@ -0,0 +1,299 @@
...
+	// always have this span.
+	bucketSpan time.Duration
+
+	// A function that returns the current time.
+	timeSource func() time.Time
+}
+
+// Create a new SigningRateTracker.
+//
+//   - signingRateDatabase: The database to use for storing historical signing rate information.
```
litt3: stale doc. consider inlining param docs with params below, to help keep things in sync

core/signingrate/signing_rate_tracker_impl.go
```diff
@@ -0,0 +1,299 @@
...
+//   - timeSpan: The amount of time to keep in memory. Queries are only supported for this timeSpan.
+//   - bucketSpan: The duration of each bucket.
+func NewSigningRateTracker(
+	logger logging.Logger,
+	timeSpan time.Duration,
+	bucketSpan time.Duration,
```
litt3, line 44: Any reason to validate time span, but not bucket span?

core/signingrate/signing_rate_tracker.go
```diff
@@ -0,0 +1,75 @@
+package signingrate
+
+import (
+	"time"
+
+	"github.com/Layr-Labs/eigenda/api/grpc/validator"
+	"github.com/Layr-Labs/eigenda/core"
+)
+
+// Tracks signing rates for validators and serves queries about signing rates.
```
litt3, line 10: This would be a good place to roughly describe what purposes the tracker serves, i.e. who is using it, and for what.

core/signingrate/signing_rate_tracker_impl.go
```diff
@@ -0,0 +1,299 @@
...
+
+func (s *signingRateTracker) GetSigningRateDump(
+	startTime time.Time,
+) ([]*validator.SigningRateBucket, error) {
+
+	buckets := make([]*validator.SigningRateBucket, 0, s.buckets.Size())
+
+	// Iterate backwards. In general, dump requests will only be used to fetch recent data, so
+	// we should optimize the case where we are requesting a few buckets from the end of the deque.
+	// Wost case scenario, we iterate the entire deque. If we do that, we are about to transmit the contents
```
litt3:
```suggestion
	// Worst case scenario, we iterate the entire deque. If we do that, we are about to transmit the contents
```

core/signingrate/signing_rate_tracker_impl.go
```diff
@@ -0,0 +1,299 @@
...
+}
+
+func (s *signingRateTracker) GetValidatorSigningRate(
+	quorum core.QuorumID,
+	id core.OperatorID,
+	startTime time.Time,
+	endTime time.Time,
+) (*validator.ValidatorSigningRate, error) {
+
+	if endTime.Before(startTime) {
```
litt3: Should this be !After instead of Before?

core/signingrate/signing_rate_tracker_impl.go
```diff
@@ -0,0 +1,299 @@
...
+			Id: id[:],
+		}, nil
+	}
+
+	comparator := func(timestamp time.Time, bucket *SigningRateBucket) int {
+		unixTimestamp := timestamp.Unix()
+
+		if unixTimestamp < bucket.startTimestamp.Unix() {
+			return -1
+		} else if unixTimestamp >= bucket.endTimestamp.Unix() {
```
litt3, line 121: Should this be >, instead of >=?

core/signingrate/signing_rate_tracker_impl.go
```diff
@@ -0,0 +1,299 @@
...
+) ([]*validator.SigningRateBucket, error) {
+
+	buckets := make([]*validator.SigningRateBucket, 0, s.buckets.Size())
+
+	// Iterate backwards. In general, dump requests will only be used to fetch recent data, so
+	// we should optimize the case where we are requesting a few buckets from the end of the deque.
+	// Wost case scenario, we iterate the entire deque. If we do that, we are about to transmit the contents
+	// of the deque over a network connection. And so in that case, the cost of iteration doesn't really matter.
+	for _, bucket := range s.buckets.ReverseIterator() {
+		if bucket.EndTimestamp().Before(startTime) || bucket.EndTimestamp().Equal(startTime) {
```
litt3:
```suggestion
		if !bucket.EndTimestamp().After(startTime) {
```

core/signingrate/threadsafe_signing_rate_tracker.go
```diff
@@ -0,0 +1,363 @@
...
+
+	// The base signing rate tracker that does the actual work.
+	base SigningRateTracker
+
+	// Channel for sending requests to the internal goroutine.
+	requests chan any
+}
+
+func NewThreadsafeSigningRateTracker(base SigningRateTracker) SigningRateTracker {
+	ctx, cancel := context.WithCancel(context.Background())
```
litt3: bike shed: curious your thoughts on an alternate pattern, instead creating a cancelable context inside the constructor, which gets stored as a member variable and called in `Close()`:

You could replace `Close()` with `Start(ctx)`. The input context could be cancelable if needed by the caller, or not if not. Then, you'd use the passed in `ctx` in your control loop, just like you're currently using the existing context.

Change not required, just curious if you think either pattern is preferable over the other

litt3 (approved):

api/proto/validator/signing_rate.proto
cody-littley: Protobufs don't have a `uint8` type. Changed this to a `uint32`, since that's how it's stored in other protobufs. I also removed the comment about future proofing.

Note that there is no meaningful attack when sending a quorum number larger than 2^8:

- the disperser can already lie about signing rate metrics
- anybody querying for a non-existent quorum is just going to get back an empty response (i.e. "I don't have any data for that quorum")

core/signingrate/signing_rate_tracker.go
cody-littley: Renamed to `validatorID`. I'd like to rename "operator ID" to "validator ID", since we are more or less deprecating the "operator" terminology in our offchain code. Eventually I'll also rename the `core.OperatorID` type.

core/signingrate/signing_rate_tracker.go
cody-littley: removed argument

core/signingrate/signing_rate_bucket.go
cody-littley: related to a thing that I ended up deleting, removed the comment

core/signingrate/signing_rate_bucket.go
cody-littley: Good catch, fixed. Not the first time I've blasted myself in the foot with this particular API.

cody-littley: If it were me designing the API, I'd so something fancy like `make([]*validator.QuorumSigningRate, 0).WithCapacity(len(b.signingRateInfo), 0)`

core/signingrate/signing_rate_bucket_test.go
cody-littley: added

core/signingrate/signing_rate_tracker_impl.go
cody-littley: done

core/signingrate/signing_rate_tracker_impl.go
cody-littley: added missing validation

core/signingrate/signing_rate_tracker.go
cody-littley:
```
// This data structure is used by two main components:
// 1. The controller keeps track of signing rates for all validators it disperses to.
// 2. The API servers periodically download signing rate data from the controller to serve API requests.
```

core/signingrate/signing_rate_tracker_impl.go
cody-littley: fixed

core/signingrate/signing_rate_tracker_impl.go
cody-littley: fixed

core/signingrate/signing_rate_tracker_impl.go
cody-littley: The start time is inclusive, the end time is exclusive. Meaning that if you request data exactly at the end timestamp, you are requesting for data that is not in the bucket. Added a comment to clarify.

```
// unixTimestamp == bucket.endTimestamp.Unix(), then timestamp is "after" the bucket since end is exclusive
```

core/signingrate/signing_rate_tracker_impl.go
cody-littley: fixed

I really wish go would let us do operator overloading. Being able to use operators like `<=` on basic data types like timestamps would be SO convenient.

core/signingrate/threadsafe_signing_rate_tracker.go
cody-littley: There are some pros and cons. Lets discuss this on a call.

core/signingrate/signing_rate_tracker_impl.go
cody-littley: fixed

core/signingrate/threadsafe_signing_rate_tracker.go
cody-littley: Fixed as we discussed on the call (use outer context, in tests use t.Context() since it gets auto-cancelled).

core/signingrate/signing_rate_tracker_impl.go
```diff
@@ -169,10 +167,10 @@ func (s *signingRateTracker) GetSigningRateDump(
 
 	// Iterate backwards. In general, dump requests will only be used to fetch recent data, so
 	// we should optimize the case where we are requesting a few buckets from the end of the deque.
-	// Wost case scenario, we iterate the entire deque. If we do that, we are about to transmit the contents
+	// Worst case scenario, we iterate the entire deque. If we do that, we are about to transmit the contents
 	// of the deque over a network connection. And so in that case, the cost of iteration doesn't really matter.
 	for _, bucket := range s.buckets.ReverseIterator() {
-		if bucket.EndTimestamp().Before(startTime) || bucket.EndTimestamp().Equal(startTime) {
+		if !bucket.EndTimestamp().After(startTime) || bucket.EndTimestamp().Equal(startTime) {
```
litt3: `bucket.EndTimestamp().Equal(startTime)` is redundant here, `!After` should cover it, right?

cody-littley: removed

PR #1993, opened 2025-09-15
feat: metrics for signing rate info

cody-littley:

A metrics struct for reporting signing rate metrics. Will allow us to see long term signing rates without looking at dynamo.


litt3 (approved):

litt3 (commented):
As we discussed offline, consider moving the metrics to the `controller` package

PR #1995, opened 2025-09-15
feat(payments): Implement reservation ledger cache

litt3:
- Implement a cache for reservation ledgers
- In service of EGDA-1855

### DIFF
--- a/core/payments/reservation/reservation_ledger.go
+++ b/core/payments/reservation/reservation_ledger.go
@@ -121,6 +121,14 @@ func (rl *ReservationLedger) RevertDebit(now time.Time, symbolCount uint32) (flo
 	return remainingCapacity, nil
 }
 
+// Checks if the underlying leaky bucket is empty.
+func (rl *ReservationLedger) IsBucketEmpty(now time.Time) (bool, error) {
+	rl.lock.Lock()
+	defer rl.lock.Unlock()
+
+	return rl.leakyBucket.CheckFillLevel(now) <= 0, nil
+}
+
 // UpdateReservation updates the reservation parameters and recreates the leaky bucket, if necessary
 //
 // This method replaces the current reservation with a new one if the new reservation differs from the old.
--- a/core/payments/reservation/reservation_ledger_cache.go
+++ b/core/payments/reservation/reservation_ledger_cache.go
@@ -1,0 +1,211 @@
+package reservation
+
+import (
+	"context"
+	"encoding/binary"
+	"errors"
+	"fmt"
+	"time"
+
+	"github.com/Layr-Labs/eigenda/common"
+	"github.com/Layr-Labs/eigenda/core/payments"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	gethcommon "github.com/ethereum/go-ethereum/common"
+	lru "github.com/hashicorp/golang-lru/v2"
+)
+
+// Stores a collection of ReservationLedgers in an LRU cache
+type ReservationLedgerCache struct {
+	logger logging.Logger
+	// A cache of the ledgers being tracked.
+	//
+	// Least recently used ReservationLedger entries are removed if the cache gets above the configured size. Since
+	// the LeakyBuckets that underlie the reservation ledgers are only in memory, evicting a ReservationLedger
+	// from this cache can result in data loss! If a ledger is deleted and then re-added, the new instance is
+	// instantiated *empty*, as if the user has had no recent dispersals. This errs on the side of permitting more
+	// throughput, since being more permissive is preferable to the alternative of *not* providing the amount of
+	// throughput guaranteed by a reservation.
+	//
+	// IMPORTANT: If the cache size is configured to be too small and there is a lot of churn, then
+	// dishonest clients may be able to utilize more than their allotted reservations! Be sure to configure a large
+	// enough cache. If any ReservationLedgers are removed from this cache with non-empty buckets, the occurrence will
+	// be logged as an error. If such error logs are observed, this cache size must be increased.
+	cache *lru.Cache[gethcommon.Address, *ReservationLedger]
+	// can access state of the PaymentVault contract
+	paymentVault payments.PaymentVault
+	// source of current time for the leaky bucket algorithm
+	timeSource func() time.Time
+	// how to handle requests that would overfill the bucket
+	overfillBehavior OverfillBehavior
+	// duration used to calculate bucket capacity
+	bucketCapacityPeriod time.Duration
+	// minimum number of symbols to bill, from the PaymentVault
+	minNumSymbols uint32
+	// protects concurrent access to the ledgers cache during ledger creation
+	//
+	// The lru.Cache object itself is threadsafe, as are the ReservationLedger values contained in the cache. This lock
+	// is to make sure that only one caller is constructing a new ReservationLedger at a time for a specific account.
+	// Otherwise, it would be possible for two separate callers to get a cache miss for the same account, create the
+	// new object for the same account key, and try to add them to the cache.
+	ledgerCreationLock *common.IndexLock
+	// monitors the PaymentVault for changes, and updates cached ledgers accordingly
+	vaultMonitor *ReservationVaultMonitor
+}
+
+func NewReservationLedgerCache(
+	ctx context.Context,
+	logger logging.Logger,
+	maxLedgers int,
+	paymentVault payments.PaymentVault,
+	timeSource func() time.Time,
+	overfillBehavior OverfillBehavior,
+	bucketCapacityPeriod time.Duration,
+	updateInterval time.Duration,
+) (*ReservationLedgerCache, error) {
+	if paymentVault == nil {
+		return nil, errors.New("payment vault must be non-nil")
+	}
+
+	if timeSource == nil {
+		return nil, errors.New("time source must be non-nil")
+	}
+
+	if bucketCapacityPeriod <= 0 {
+		return nil, errors.New("bucket capacity period must be > 0")
+	}
+
+	cache, err := lru.NewWithEvict(
+		maxLedgers,
+		func(accountAddress gethcommon.Address, reservationLedger *ReservationLedger) {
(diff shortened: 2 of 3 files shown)


core/payments/reservation/reservation_ledger.go
```diff
@@ -121,6 +121,14 @@ func (rl *ReservationLedger) RevertDebit(now time.Time, symbolCount uint32) (flo
 	return remainingCapacity, nil
 }
 
+// Checks if the underlying leaky bucket is empty.
+func (rl *ReservationLedger) IsBucketEmpty(now time.Time) bool {
```
samlaf, line 128: nit: would this not work with any time? Or does it really only work with a "now" for some reason?

Aka could I use this method to ask questions about past and/or future times? Would that even make sense? Probably not because the dispersed blobs that affect bucket do not time travel..?

core/payments/reservation/reservation_ledger_cache.go
```diff
@@ -0,0 +1,206 @@
...
+	// the LeakyBuckets that underlie the reservation ledgers are only in memory, evicting a ReservationLedger
+	// from this cache can result in data loss! If a ledger is deleted and then re-added, the new instance is
+	// instantiated *empty*, as if the user has had no recent dispersals. This errs on the side of permitting more
+	// throughput, since being more permissive is preferable to the alternative of *not* providing the amount of
+	// throughput guaranteed by a reservation.
+	//
+	// IMPORTANT: If the cache size is configured to be too small and there is a lot of churn, then
+	// dishonest clients may be able to utilize more than their allotted reservations! Be sure to configure a large
+	// enough cache. If any ReservationLedgers are removed from this cache with non-empty buckets, the occurrence will
+	// be logged as an error. If such error logs are observed, this cache size must be increased.
```
samlaf: do we have a metric tracking this as well?

core/payments/reservation/reservation_ledger_cache.go
```diff
@@ -0,0 +1,206 @@
...
+	cache *lru.Cache[gethcommon.Address, *ReservationLedger]
+	// can access state of the PaymentVault contract
+	paymentVault payments.PaymentVault
+	// source of current time for the leaky bucket algorithm
+	timeSource func() time.Time
```
samlaf, line 49: nit: might be nice to create a `Clock` interface, which could only have `now()` as a method for now (assuming this is all that leaky buckets need?)

core/payments/reservation/reservation_ledger_cache.go
```diff
@@ -0,0 +1,206 @@
...
+	// how to handle requests that would overfill the bucket
+	overfillBehavior OverfillBehavior
+	// duration used to calculate bucket capacity
+	bucketCapacityPeriod time.Duration
+	// minimum number of symbols to bill, from the PaymentVault
```
samlaf:
```suggestion
	// minimum number of symbols to bill each dispersal, from the PaymentVault
```

core/payments/reservation/reservation_ledger_cache.go
```diff
@@ -0,0 +1,206 @@
...
+	minNumSymbols uint32
+	// protects concurrent access to the ledgers cache during ledger creation
+	//
+	// The lru.Cache object itself is threadsafe, as are the ReservationLedger values contained in the cache. This lock
+	// is to make sure that only one caller is constructing a new ReservationLedger at a time for a specific account.
+	// Otherwise, it would be possible for two separate callers to get a cache miss for the same account, create the
+	// new object for the same account key, and try to add them to the cache.
+	ledgerCreationLock *common.IndexLock
```
samlaf, line 62: how are we expecting this reservation ledger to be used again? Why would multiple goroutines try to create new ledgers? oh is each request received by the validator running in a separate goroutine, and each are using this shared reservation ledger?

core/payments/reservation/reservation_ledger_cache.go
```diff
@@ -0,0 +1,206 @@
...
+	ctx context.Context,
+	accountID gethcommon.Address,
+) (*ReservationLedger, error) {
+	// Fast path: check if ledger already exists in cache
+	if ledger, exists := c.cache.Get(accountID); exists {
+		return ledger, nil
+	}
+
+	// Slow path: acquire per-account lock and check again
+	accountIndex := binary.BigEndian.Uint64(accountID.Bytes()[:8])
```
samlaf: nit: constructor for IndexLock takes a uint32 but then its lock/unlock methods take a uint64. Feels inconsistent? Also not important but we could consider adding a bunch more typed lock/unlock methods on the accountIndex, such that we could pass in a geth-address, or a bigInt, etc. thoughts?

samlaf (approved):

core/payments/reservation/reservation_ledger_cache.go
```diff
@@ -0,0 +1,252 @@
...
+) {
+	if reservationLedger.IsBucketEmpty(c.timeSource()) {
+		c.logger.Debugf("evicted account %s from LRU reservation ledger cache", accountID.Hex())
+		return
+	}
+
+	// The bucket is not empty!!! This was a premature eviction: we must resize the cache
+
+	newSize := c.maxLedgers * 2
+	if newSize > maxReservationLRUCacheSize {
```
cody-littley, line 227: Is it possible for `handleEviction()` to be called concurrently on multiple threads? If "no", might be good to comment why. If "yes", then you probably need to hold a lock while you resize the cache.

core/payments/reservation/reservation_ledger.go
litt3: It MUST be the true `now`

Under the hood, this time instance is being used to `leak` the bucket. We need to compute the most up-to-date state of the leaky bucket to check if it's empty. And to do that, we need to know when "now" is.

Added doc to try to clarify 60ffef5afa381a4caa0f0e560a63e79801a5859c

core/payments/reservation/reservation_ledger_cache.go
litt3: e7abc82e70773853b24f4a44f42108118516ebef

I added an eviction lock. To remove the possibility of a deadlock, I removed the acquisition of the `ledgerCreationLock` within the evict method.

core/payments/reservation/reservation_ledger_cache.go
litt3: Were I to keep the original implementation, I'd agree that a metric is needed.

But, as @cody-littley rightfully pointed out in his review, the original strategy of this PR, permitting premature evictions, was something of a foot gun which shouldn't be tolerated.

With the most recent implementation, the cache will be dynamically resized up to a very large value. Therefore, I don't think the metric is necessary anymore.

core/payments/reservation/reservation_ledger_cache.go
litt3: Interesting- stylistically you'd prefer that?

My philosophy using `func() time.Time` in lots of places is that it's easy enough, and avoids a layer of indirection. But I don't feel super strongly. Maybe something to consider for a future PR

core/payments/reservation/reservation_ledger_cache.go
litt3: Added clarification, with slightly different wording 

12d1abcc0419ff6a71ecaffa60b4049c04af5969

core/payments/reservation/reservation_ledger_cache.go
litt3: > request received by the validator running in a separate goroutine, and each are using this shared reservation ledger

That's right. Note that this cache will be used on both the `disperser`, and on `validators`. In both cases, requests from a single account may arrive and be handled in separate routines.

Imagine a new client comes online and starts pumping out dispersals. It's possible/likely that the `disperser` or a `validator` may be handling multiple of these requests before the first request has been fully handled and the ledger added to the queue. In such a case, the ledger creation lock ensures only one instance is created

core/payments/reservation/reservation_ledger_cache.go
litt3: > Feels inconsistent?

The total number of indices will exceed the number of locks, probably by a lot (otherwise, why use the index lock at all?). With that in mind, having the index be a larger int type seems reasonable to me.

> we could consider adding a bunch more typed lock/unlock methods on the accountIndex, such that we could pass in a geth-address, or a bigInt, etc.

Hmm, interesting point. Maybe makes sense? On the one hand, it's kinda nice that the existing index lock impl is super simple and pure.... index in, lock out. On the other hand, it's unsightly to be manually converting geth addresses to uint64 indices all over the place.

cc @cody-littley for discussion on the index lock util

samlaf (approved):

core/payments/reservation/reservation_ledger_cache.go
cody-littley: I'd be ok adding functions with the following pattern for common index types (e.g. byte arrays, uint32, etc.). Internally, they would convert to uint64 and call into `Lock()`/`Unlock()`. Up to you if you want to do this in this PR, or a follow up.

```
func LockBytes(index byte[])
func UnlockBytes(index byte[])
```

core/payments/reservation/reservation_ledger_cache.go
cody-littley: I'm not sure if using an interface instead of a function is necessary... do we ever need a clock object to do more than provide the current time? Perhaps we could add an alias to make it look pretty, though: `type Clock func() time.Time`. Not sure if the introduction of this type alias belongs in this PR or another.

samlaf: I'd be happy with the type def (it does have one less level of indirection, which is not needed if we only use a single function). Does not need to be done in this PR, or ever tbh. Just a personal preference that I can live without.

core/payments/reservation/reservation_ledger_cache.go
litt3: I'll add this next time I use the index lock for something similar

PR #1997, opened 2025-09-16
refactor(srs)!: fix SRSOrder to 2^28 + hardcode 28 G1PowerOf2 pts we need

samlaf:
## Breaking Change

V1 does not enforce that blob lengths are power of 2. So we would need to deploy to our preprod/testnet/mainnet disperser before we distribute this code to any validators/clients.

TODO: try to duplicate VerifyCommit into VerifyCommitV1/V2 and only make this change across V1 codepaths.

## What
Two main things:
1. hardcode SRSOrder everywhere, and remove it from KzgConfig (afaiu it was only configurable for tests)
2. Embed 28 G1 points that are needed to verify blob length commitments/proofs

## Benefits of this PR
1. disperser, encoder, inabox, proxy, etc. will all use consistent ways of loading and using SRS points (right now its not possible to spin up proxy against inabox for example)
4. operators will only need to download 16MiBs of G1 SRS points, instead of the current 8GiB file that they have to download

## Implementation details / things to be careful

In order to make sure that the VerifyCommitment fct (see this [discussion](https://github.com/Layr-Labs/eigenda/pull/1997#discussion_r2350684724)) really only ever needs to read one of the embedded 28 G1 points, we have to make sure that the blob length passed to it is a power of 2 (since there are only 28 powers of 2 < 2^28):
image

My current understanding is that we:
1. don't want to enforce "blobs" dispersed to be powers of 2, BUT
2. we want to enforce that [blobLengths](https://github.com/Layr-Labs/eigenda/blob/master/encoding/data.go#L32-L33) are always a power of 2 (note that is already part of the comment in the code!!)

However a lot of our current tests didn't enforce this, and were using non power of 2 lengths. I fixed some of them, but am unsure how exactly this behaves across the entire architecture, and whether there are other hidden assumptions (for example how does it affect V1 chunk assignment logic).

I also made a change to the low level ToFrArray function (see this [discussion](https://github.com/Layr-Labs/eigenda/pull/1997/files#r2350686402)), such that it always pads to the next power of 2. **This is probably not what we want?**

## Notes
The SRSOrder is currently configurable in the KzgConfig. Afaiu that was just to allow for "hacks" in tests and inabox. The real srs order is 2^28 for bn254, and its a lot more principled to stick to the real order, and instead load fewer points in tests/inabox. This however requires loading points from the "trailing" part of the large (8Gib and 16GiB) G1 and G2 SRS files. We already have that functionality for G2,  should instead be passing the g2 trailing file like we now do elsewhere.

For G1, the previous code was reading trailing points of the G1 file (G1file[2^28 - 2^i]) one by one whenever needed. By enforcing that all blobs are powers of 2, and fixed the SRS order, we really only ever need to read 28 G1 points. We hardcode those and embed them into the binaries.

One weird thing I realized is that the prover doesn't check and/or enforce that the data is receives is a power of 2 length. And the tests were not enforcing that. Changes the ToFrArray to pad the data it encodes to the next power of 2. That loses some performance since we might be sending some 0s over the wire, but at least it allows the other optimization of not needing to read the entire G1 file like operators do today. Also we can always change that function and only pad on the disperser if ever needed.

TODO: still some tests to fix


encoding/kzg/verifier/verifier.go
```diff
@@ -104,14 +106,18 @@ func (v *Verifier) VerifyBlobLength(commitments encoding.BlobCommitments) error
...
+func (v *Verifier) VerifyCommit(lengthCommit *bn254.G2Affine, lengthProof *bn254.G2Affine, commitmentLength uint64) error {
+	if !encoding.IsPowerOfTwo(commitmentLength) {
+		return fmt.Errorf("commitment length %d is not a power of 2", commitmentLength)
+	}
+	commitmentLengthLog := bits.TrailingZeros64(commitmentLength)
+	if commitmentLengthLog > 28 {
+		return fmt.Errorf("commitment length %d is > max possible 2^28", commitmentLength)
 	}
+	// TODO(samlaf): document how this works...
+	g1Challenge := srs.G1PowerOf2SRS[commitmentLengthLog]
```
samlaf, line 118: TODO: @bxue-l2 need your help

encoding/rs/utils.go
```diff
@@ -19,7 +19,12 @@ func ToFrArray(inputData []byte) ([]fr.Element, error) {
 	bytes := padToBytesPerSymbolMultiple(inputData)
 
 	elementCount := len(bytes) / encoding.BYTES_PER_SYMBOL
-	outputElements := make([]fr.Element, elementCount)
+	// We pad to the next power of 2, because blobs are always a power of 2 symbols,
+	// and our prover currently doesn't pad to the next power of 2.
+	// TODO(samlaf): should we pad in the prover only instead?
+	// Padding here will force all dispersed data over the wire to be a power of 2,
+	// possibly sending a bunch of zeros over the wire.
+	outputElements := make([]fr.Element, NextPowerOf2(uint64(elementCount)))
```
samlaf: TODO: discuss whether we are ok with this, and whether there are other possible solutions. Also are we sure the disperser pads all data it receives to power of 2? Prover should probably enforce that... it currently doesn't.

samlaf: In particular @bxue-l2 PTAL at the TestLengthProof in encoding/kzg/verifier/length_test.go.

It was failing which required making this change here. But previously it was working without enforcing that blobs are powers of 2 (blobLength was not). This is kind of scary... are we sure we enforce on testnet/mainnet that blobs are always powers of 2?? Also you'll need to explain to me how the g1Challenge point chosen makes sense, since it was working before with a point that is not of the form G1SRSFile[2^28-2^i].

encoding/kzg/verifier/verifier.go
bxue-l2: ah, I might have said something wrong before. The length pairing check only does
 e( g1Challenge, lengthCommit) = e( G_1, length_proof )

Notice this the eq above does not have kzg commitment, but we need to check equivalence of lengthCommit and kzg blob commitment.

We used another trick, https://github.com/Layr-Labs/eigenda/blob/fec8aa5da665bfad80795190e08ce785305b13eb/core/v2/validator.go#L185. And we the verification in batch

e(kzg blob commitment, G2) = (G1, length Commitment)

encoding/rs/utils.go
bxue-l2: Oh, for AVS users, it might be true.

For rollup, I think they all take IFFT. but it is hard to assert. Good callout

bxue-l2: I think for v2, we started with blob to be power of 2. Maybe a phased transition if it worths the time.

encoding/kzg/verifier/verifier.go
samlaf: This is not what our spec says the length commitment/proof are at all... should we update it? https://layr-labs.github.io/eigenda/protocol/architecture/encoding.html#validation-via-kzg

It says that `length_commitment = [x^{n-m}p]_1`, and that with this iiuc you dont even need a length proof? You can just directly check `e([p]_1, [x^{n-m}]_2) = e(length_commitment, G_2)`. What am I missing? Why isn't this what the code is doing?

bxue-l2: Because the srs we use have more g2 points than g1. It forces us to have this change, otherwise it is hackable

PR #1795, opened 2025-07-23
feat: add grpc logging interceptor in v2 api server

ian-shim:
## Why are these changes needed?
Redoing https://github.com/Layr-Labs/eigenda/pull/1690

## Checks

- [ ] I've made sure the tests are passing. Note that there might be a few flaky tests, in that case, please comment that they are not relevant.
- [ ] I've checked the new test coverage and the coverage percentage didn't drop.
- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests
   - [ ] This PR is not tested :(

PR #1998, opened 2025-09-16
feat: log grpc response status from api server

ian-shim:
Redoing https://github.com/Layr-Labs/eigenda/pull/1795 without interceptors. 
Wraps a method that returns status and logs the status with corresponding level before returning the result

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/apiserver/disperse_blob_v2.go
+++ b/disperser/apiserver/disperse_blob_v2.go
@@ -65,7 +65,6 @@ func (s *DispersalServerV2) disperseBlob(
 		}
 		_, err := s.controllerClient.AuthorizePayment(ctx, authorizePaymentRequest)
 		if err != nil {
-			s.logger.Warn("payment authorization failed", "err", err)
 			return nil, status.Convert(err)
 		}
 	} else {
--- a/disperser/apiserver/get_blob_status_v2.go
+++ b/disperser/apiserver/get_blob_status_v2.go
@@ -67,13 +67,6 @@ func (s *DispersalServerV2) getBlobStatus(
 
 	cert, _, err := s.blobMetadataStore.GetBlobCertificate(ctx, blobKey)
 	if err != nil {
-		s.logger.Error(
-			"failed to get blob certificate for blob in GatheringSignatures/Complete status",
-			"err",
-			err,
-			"blobKey",
-			blobKey.Hex(),
-		)
 		if errors.Is(err, dispcommon.ErrMetadataNotFound) {
 			return nil, status.New(codes.NotFound, "no such blob certificate found")
 		}
@@ -83,13 +76,11 @@ func (s *DispersalServerV2) getBlobStatus(
 	// For blobs in GatheringSignatures/Complete status, include signed batch and blob inclusion info
 	blobInclusionInfos, err := s.blobMetadataStore.GetBlobInclusionInfos(ctx, blobKey)
 	if err != nil {
-		s.logger.Error("failed to get blob inclusion info for blob", "err", err, "blobKey", blobKey.Hex())
-		return nil, status.Newf(codes.Internal, "failed to get blob inclusion info: %v", err)
+		return nil, status.Newf(codes.Internal, "failed to get blob inclusion info for blob %s: %v", blobKey.Hex(), err)
 	}
 
 	if len(blobInclusionInfos) == 0 {
-		s.logger.Error("no blob inclusion info found for blob", "blobKey", blobKey.Hex())
-		return nil, status.New(codes.Internal, "no blob inclusion info found")
+		return nil, status.Newf(codes.Internal, "no blob inclusion info found for blob %s", blobKey.Hex())
 	}
 
 	if len(blobInclusionInfos) > 1 {
@@ -143,6 +134,5 @@ func (s *DispersalServerV2) getBlobStatus(
 		}, status.New(codes.OK, "")
 	}
 
-	s.logger.Error("no signed batch found for blob", "blobKey", blobKey.Hex())
-	return nil, status.New(codes.Internal, "no signed batch found")
+	return nil, status.Newf(codes.Internal, "no signed batch found for blob %s", blobKey.Hex())
 }


cody-littley (approved):

disperser/apiserver/disperse_blob_v2.go
```diff
@@ -28,15 +46,15 @@ func (s *DispersalServerV2) DisperseBlob(ctx context.Context, req *pb.DisperseBl
 	// Validate the request
 	onchainState := s.onchainState.Load()
 	if onchainState == nil {
-		return nil, api.NewErrorInternal("onchain state is nil")
+		return nil, status.New(codes.Internal, "onchain state is not available")
```
litt3, line 49: Building statuses this way bypasses the utility methods introduced by @samlaf. If we want to build statuses instead of errors, should we have equivalent utilities for statuses?

disperser/apiserver/disperse_blob_v2.go
```diff
@@ -17,9 +17,27 @@ import (
...
+		// nolint:wrapcheck
+		return reply, st.Err()
+	}
+	return reply, nil
+}
+
+func (s *DispersalServerV2) disperseBlob(
+	ctx context.Context,
+	req *pb.DisperseBlobRequest,
+) (*pb.DisperseBlobReply, *status.Status) {
```
litt3, line 40: What is the reasoning being switching the return types of all these internal methods to `*status.Status`?

It seems like we could accomplish the same thing if we left the return types as errors, and then log the errors within `DisperseBlob`. Are we gaining anything by wrapping these grpc errors up into statuses, just to turn around and unpack them back into errors in `DisperseBlob`?

disperser/apiserver/disperse_blob_v2.go
```diff
@@ -17,9 +17,27 @@ import (
...
 )
 
-func (s *DispersalServerV2) DisperseBlob(ctx context.Context, req *pb.DisperseBlobRequest) (*pb.DisperseBlobReply, error) {
+func (s *DispersalServerV2) DisperseBlob(
+	ctx context.Context,
+	req *pb.DisperseBlobRequest,
+) (*pb.DisperseBlobReply, error) {
+	reply, st := s.disperseBlob(ctx, req)
+	api.LogResponseStatus(s.logger, st)
+	if st != nil {
```
litt3, line 30: Here you're checking whether the status is non-nil, and returning an error.

But looking at `disperseBlob`, the status is ALWAYS non-nil, even in the success case. This works because you're returning the `reply` even in the error case, but this doesn't seem like what you were going for

disperser/apiserver/disperse_blob_v2.go
```diff
@@ -47,14 +65,14 @@ func (s *DispersalServerV2) DisperseBlob(ctx context.Context, req *pb.DisperseBl
 		}
 		_, err := s.controllerClient.AuthorizePayment(ctx, authorizePaymentRequest)
 		if err != nil {
-			// nolint:wrapcheck // Pass through the structured error from the controller
-			return nil, err
+			s.logger.Warn("payment authorization failed", "err", err)
```
litt3: Won't this case be logged like all the other errors? Does it need to be logged manually here?

api/logging.go
```diff
@@ -0,0 +1,36 @@
...
+		codes.FailedPrecondition,
+		codes.Aborted,
+		codes.OutOfRange,
+		codes.Unimplemented,
+		codes.Internal,
+		codes.Unavailable,
+		codes.DataLoss:
+		logger.Error("gRPC response status", "code", s.Code(), "message", s.Message())
+	case codes.Canceled,
+		codes.InvalidArgument,
```
litt3, line 27: Curious how you decided on the categorizations here. For example, it seems like `InvalidArgument` should often be an error, but it's part of the warnings here. Seems like any way we try to slice these will be tricky, since these codes will have different severity, depending on the situations where they were triggered. 

Would it make sense to simply log _all_ of these as errors?

disperser/apiserver/get_blob_status_v2.go
```diff
@@ -13,59 +13,83 @@ import (
...
 		return &pb.BlobStatusReply{
 			Status: metadata.BlobStatus.ToProfobuf(),
-		}, nil
+		}, status.New(codes.OK, "")
 	}
 
 	cert, _, err := s.blobMetadataStore.GetBlobCertificate(ctx, blobKey)
 	if err != nil {
-		s.logger.Error("failed to get blob certificate for blob in GatheringSignatures/Complete status", "err", err, "blobKey", blobKey.Hex())
+		s.logger.Error(
```
litt3: Any reason why the standard logging above isn't sufficient, and this is logged manually here?

(Same question goes for other manual logs within these utilities)

samlaf:
Curious why did you decide to move away from using interceptors? Feel like we could have added this logging utility as an interceptor no?

disperser/apiserver/server_v2.go
```diff
@@ -232,38 +234,53 @@ func (s *DispersalServerV2) Start(ctx context.Context) error {
...
+	req *pb.BlobCommitmentRequest,
+) (*pb.BlobCommitmentReply, *status.Status) {
 	start := time.Now()
 	defer func() {
 		s.metrics.reportGetBlobCommitmentLatency(time.Since(start))
 	}()
 
 	if s.prover == nil {
-		return nil, api.NewErrorUnimplemented()
+		return nil, status.New(codes.Internal, "prover is not configured")
```
litt3, line 259: Did you intentionally change this from an `Unimplemented` error to an `Internal` error?

disperser/apiserver/server_v2.go
```diff
@@ -319,25 +336,41 @@ func (s *DispersalServerV2) RefreshOnchainState(ctx context.Context) error {
...
+		return nil, status.New(codes.InvalidArgument, "invalid account ID")
 	}
 
 	accountID := gethcommon.HexToAddress(req.GetAccountId())
 
 	// validate the signature
 	if err := s.blobRequestAuthenticator.AuthenticatePaymentStateRequest(accountID, req); err != nil {
 		s.logger.Debug("failed to validate signature", "err", err, "accountID", accountID)
-		return nil, api.NewErrorInvalidArg(fmt.Sprintf("authentication failed: %s", err.Error()))
+		return nil, status.Newf(codes.Unauthenticated, "failed to validate signature: %s", err.Error())
```
litt3, line 373: You changed from `InvalidArgument` to `Unauthenticated`. Was this intentional?

disperser/apiserver/disperse_blob_v2.go
ian-shim: I thought about doing that, but the utilities are thin wrappers around the statuses anyways. So I'm not sure it's worth creating a wrapper for each code

api/logging.go
ian-shim: The codes will have different severity but I think they should be consistent. 
Log should be at `error` level only if it's unexpected and deserves something like an alert. I categorized any codes that are "expected" as `warn` here. i.e. if the request fails due to invalid arguments, the failure is expected

disperser/apiserver/disperse_blob_v2.go
ian-shim: Right, this is just to cover any future mistakes where we introduce new return path that returns `nil` instead of an `OK` status. Maybe I should make the child method's return type not a pointer?

disperser/apiserver/disperse_blob_v2.go
ian-shim: Yeah, I wanted the wrapper method to inspect the severity (status code) and log at different levels

disperser/apiserver/disperse_blob_v2.go
ian-shim: You're right. Will remove

disperser/apiserver/server_v2.go
ian-shim: Yes. Thought `Internal` is more appropriate because missing prover is probably a misconfiguration, not that the method is `Unimplemented`. Wdyt?

disperser/apiserver/server_v2.go
ian-shim: Yes, changed to `Unauthenticated` since it's an authentication failure, but i could see how this can be `InvalidArgument`.

disperser/apiserver/server_v2.go
litt3: 👍 Makes sense to me

disperser/apiserver/server_v2.go
litt3: Unauthenticated makes sense to me, just wanted to make sure it wasn't inadvertent

litt3 (approved):

PR #2000, opened 2025-09-16
test: run LittDB unit tests in a separate panel

cody-littley:

Split LittDB unit tests into stand alone panel.


.github/workflows/litt-unit-tests.yaml
```diff
@@ -0,0 +1,74 @@
...
+        run: make build
+
+      - name: Test LittDB
+        run: make litt-unit-tests
+
+      - name: Upload coverage artifact
+        uses: actions/upload-artifact@65c4c4a1ddee5b72f698fdd19549f0f0fb45cf08 # v4.6
+        with:
+          name: coverage
+          path: coverage.out
```
litt3: I think this will conflict with the `coverage.out` from the main unit tests. I think you need to have separate unique artifact names here. Consider whether this might apply to any other parts of these tests as well

Makefile
```diff
@@ -92,11 +92,15 @@ docker-release-build:
 	BUILD_TAG=${SEMVER} SEMVER=${SEMVER} GITDATE=${GITDATE} GIT_SHA=${GITSHA} GIT_SHORT_SHA=${GITCOMMIT} \
 	docker buildx bake proxy-release ${PUSH_FLAG}
 
-# Some of the unit test suites take > 1 min to run (e.g. relay and littdb tests).
-# TODO: we should break these up into short and long unit-tests.
+# Run all tests that don't have their own panel.
 unit-tests:
```
litt3, line 96: looks like you removed `CI=true`. Not sure if that should live here or within the test scripts themselves, but I think it should be added somewhere

Makefile
```diff
@@ -92,11 +92,15 @@ docker-release-build:
...
 	go clean -testcache
-	CI=true go test -short ./... -coverprofile=coverage.out
+	./test/scripts/test-with-blacklist.sh . ./litt
+
+# Run the unit tests in litt/ only.
+litt-unit-tests:
+	go clean -testcache
+	./test/scripts/test-with-whitelist.sh . ./litt
```
litt3, line 103: Your whitelist script cleans the cache, so the `go clean -testcache` here is redundant.

Your blacklist script _doesn't_ clean the cache. Probably you should choose a consistent strategy for both, and remove any duplicate cleanings

.github/workflows/litt-unit-tests.yaml
litt3: Also consider how choosing a different artifact name will affect action badges

test/scripts/test-with-whitelist.sh
```diff
@@ -0,0 +1,48 @@
...
+# Trim leading/trailing spaces
+PKGS=$(echo "$PKGS" | xargs)
+
+if [ -z "$PKGS" ]; then
+  echo "No packages matched the whitelist." >&2
+  exit 0
+fi
+
+echo "Running tests for whitelist:"
+printf '%s\n' $PKGS
```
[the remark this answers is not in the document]
cody-littley: done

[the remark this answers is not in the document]
cody-littley: packages can't have spaces (and even if they can, we shouldn't let them past code review!)

test/scripts/test-with-blacklist.sh
```diff
@@ -0,0 +1,61 @@
...
+# Trim whitespace.
+PKGS=$(echo "$PKGS" | xargs || true)
+
+if [ -z "$PKGS" ]; then
+  echo "No packages left to test after applying blacklist." >&2
+  exit 0
+fi
+
+echo "Running tests for:"
+printf '%s\n' $PKGS
```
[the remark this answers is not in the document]
cody-littley: packages can't have spaces (and even if they can, we shouldn't let them past code review!)

test/scripts/test-with-whitelist.sh
```diff
@@ -0,0 +1,48 @@
...
+
+go clean -testcache
+go test -short $PKGS -coverprofile=coverage.out
```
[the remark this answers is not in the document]
cody-littley: packages can't have spaces (and even if they can, we shouldn't let them past code review!)

test/scripts/test-with-blacklist.sh
```diff
@@ -0,0 +1,61 @@
...
+
+# Run tests (coverage output like your prior script)
+go test -short $PKGS -coverprofile=coverage.out
```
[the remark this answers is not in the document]
cody-littley: packages can't have spaces (and even if they can, we shouldn't let them past code review!)

Makefile
cody-littley: removed double clean, left it in the makefile

Makefile
cody-littley: Moved the flags into the scripts. Concerned that if I define them in the makefiles, they may not get passed down into the script's context.

litt3 (dismissed):
Nothing looks wrong to me, but I'm not familiar enough with github yaml format to be super confident here.

PR #2002, opened 2025-09-16
feat: config registry

pakim249CAL:

Adds a configuration registry to the EigenDADirectory, with the initial goal of supporting a simple key value store keyed by a string hash, and values of fixed and dynamically sized bytes, settable by EigenDA governance. This configuration registry can be expanded at a later date to include more configuration formats.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
```diff
@@ -0,0 +1,21 @@
...
+
+library ConfigRegistryTypes {
+    struct NameSet {
+        mapping(bytes32 => string) names;
+        string[] nameList;
+    }
+
+    struct Bytes32Cfg {
+        mapping(bytes32 => bytes32) values;
+        mapping(bytes32 => string) extraInfo;
```
samlaf, line 12: curious, why not store them as as single mapping to a struct{values, extraInfo} and have a single getter that returns both?

contracts/src/core/interfaces/IEigenDADirectory.sol
```diff
@@ -42,3 +39,10 @@ interface IEigenDADirectory {
     /// @notice Gets all names in the directory.
     function getAllNames() external view returns (string[] memory);
 }
+
+interface IEigenDAConfigRegistry {}
```
samlaf: is this meant to be empty?

pakim249CAL: Eventually I will copy over the fns I added in the EigenDADirectory here. I am just not completely sure about the interface yet.

contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
pakim249CAL: In this particular case, I don't think structifying this makes the code any clearer, and bundling it doesn't really serve any on-chain use case. extraInfo is not meant to be consumed on-chain at all, and will typically just be a wasted SLOAD if it is bundled into a getter.

contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
```diff
@@ -0,0 +1,21 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
```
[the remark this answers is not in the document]
pakim249CAL: This inconsistency is intentional because libraries like this that simply define types do not need strict versioning like test files do. This allows this library to be used in other projects more easily.

contracts/src/core/interfaces/IEigenDADirectory.sol
pakim249CAL: I've added a more full interface here.

contracts/src/core/interfaces/IEigenDADirectory.sol
```diff
@@ -42,3 +39,84 @@ interface IEigenDADirectory {
...
+
+    /// @notice Gets the registered key for bytes32 configuration entries at a specific index.
+    function getRegisteredKeyBytes32(uint256 index) external view returns (string memory);
+
+    /// @notice Gets the registered key for bytes configuration entries at a specific index.
+    function getRegisteredKeyBytes(uint256 index) external view returns (string memory);
+}
+
+/// @notice Interface for the EigenDA Directory
+///         This interface currently only includes functions for managing a directory of addresses by name.
```
litt3: > only includes functions for managing a directory of addresses by name.

Doc is stale, since this now deals with more than just addresses

litt3 (dismissed):
LGTM

contracts/src/core/interfaces/IEigenDADirectory.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/2002/commits/7e2c4b699490b2c406c5c9b7808a32b7688904ea

litt3 (approved):

contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
samlaf: I was assuming the compiler would be smart enough to only load the part of the struct that is actually used, but maybe not.

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
```diff
@@ -0,0 +1,112 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+import {ConfigRegistryStorage as S} from "src/core/libraries/v3/config-registry/ConfigRegistryStorage.sol";
+import {ConfigRegistryTypes as T} from "src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol";
+
+library ConfigRegistryLib {
```
samlaf, line 7: Might be nice to add a bit of documentation saying that this only implements setters and getters, and updates+deletes+access-control all need to be done on top of this.

PR #2003, opened 2025-09-16
feat(payments): Implement `ReservationPaymentValidator`

litt3:
- closes EGDA-1855


core/payments/reservation/reservation_payment_validator.go
```diff
@@ -0,0 +1,85 @@
...
+	// provides access to payment vault contract
+	paymentVault payments.PaymentVault,
+	// source of current time for the leaky bucket algorithm
+	timeSource func() time.Time,
+	// how to handle requests that would overfill the bucket
+	overfillBehavior OverfillBehavior,
+	// duration used to calculate bucket capacity
+	bucketCapacityPeriod time.Duration,
+	// interval for checking for payment updates
+	updateInterval time.Duration,
```
samlaf: nit: why aren't you using a config struct?

core/payments/reservation/reservation_payment_validator.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+
+	now := pv.timeSource()
+	success, _, err := ledger.Debit(now, dispersalTime, symbolCount, quorumNumbers)
+	if err != nil {
+		return fmt.Errorf("debit reservation payment: %w", err)
+	}
+
+	if !success {
+		return fmt.Errorf("reservation debit failed: insufficient capacity")
```
samlaf: is there no more info we could give to users?
Does this mean global limit is reached? Or it means their bucket is overfilled atm?

core/payments/reservation/reservation_payment_validator.go
```diff
@@ -0,0 +1,85 @@
...
+	quorumNumbers []uint8,
+	dispersalTime time.Time,
+) error {
+	ledger, err := pv.ledgerCache.GetOrCreate(ctx, accountID)
+	if err != nil {
+		return fmt.Errorf("get or create ledger: %w", err)
+	}
+
+	now := pv.timeSource()
+	success, _, err := ledger.Debit(now, dispersalTime, symbolCount, quorumNumbers)
```
cody-littley: Bit of a race here, but you already called it out in a comment.

```
	// Don't bother checking if another routine already re-created this ledger. Even if another routine *did* create
	// a new instance, it's reasonable to preference the old instance over the new. There may be some small discrepancy
	// here, but there would be no feasible way for a malicious client to exploit this. In the worst case, the leaky
	// bucket will be slightly less filled than it ought to have been. Since it's incredibly unlikely to happen in the
	// first place, it's not worth contorting the design to address.
	c.cache.Add(accountID, reservationLedger)
```

cody-littley (dismissed):

core/payments/reservation/reservation_payment_validator.go
cody-littley:
```suggestion
		return fmt.Errorf("go pound sand")
```

litt3: Note that this isn't a component that will be used by users. This is used by the disperser, and by validators. Users interact with the `ClientLedger`, which has more user friendly messages.

> Does this mean global limit is reached?

There is no concept of a global limit for reservation based dispersals. This just means the account bucket cannot permit the debit

core/payments/reservation/reservation_payment_validator.go
litt3: Do you think I need to add anything to this struct? Or is the doc in the cache enough?

core/payments/reservation/reservation_payment_validator.go
litt3: No specific reason

9dda7ab4d2a7237c6204d83c421a0953b6fcb7fb

core/payments/reservation/reservation_payment_validator.go
cody-littley: I'm ok with this in its current form

cody-littley (approved):

core/payments/reservation/reservation_payment_validator.go
samlaf: I see. We prob still want informative logs in case users complain and/or ask aboutit. Is this error wrapped with client account_id and bucket size info at a higher level? If not, think we should add that info here.

samlaf (dismissed):
Happy with this PR contingent on answer to https://github.com/Layr-Labs/eigenda/pull/2003#discussion_r2356241456

core/payments/reservation/reservation_payment_validator.go
litt3: Theoretically, it shouldn't be possible for a user to trigger this insufficient capacity error, unless they modify their client to bypass local accounting. 

I agree it makes sense to add additional context, since it's easy to do so. I don't think there's anything to gain adding account ID to the log, though- clients know their own account ID

5f96f5275bc514d4ceb10bae32224c0376306ddb

cody-littley (approved):

core/payments/reservation/reservation_payment_validator.go
samlaf: this was for the log on the server side, so that we can know which accountID didnt have capacity and correlate with user log

PR #2005, opened 2025-09-17
fix: deprecate built-in indexer with error messages

cosmoknightt:
Following up on @litt3's suggestion in #1894 to address the unused indexer code.

- Replace indexer creation with deprecation error messages in batcher and controller
- Error directs users to use `UseGraph=true` instead

I didn't remove the indexer module itself in this PR to keep changes minimal. The indexer code remains but can't be instantiated. If you'd prefer a complete removal of the indexer directory and related code, I can do that in a follow-up PR.

**Question:**
Are there other areas of dead code you'd recommend cleaning up while we're at it? Happy to tackle more deprecation/cleanup work.


dmanc (approved):

litt3 (approved):
LGTM. Thanks @cosmoknightt!

litt3:
> > LGTM. Thanks @cosmoknightt!
> 
> Thanks! BTW should we remove the indexer directory and related code. And if its okay would you create some first-good-issue? or if there is any chore or test I can further assist you guys @litt3

I think it probably makes sense to remove the dead code- though I'm not super familiar with this area of the code. If you'd like to take a crack at it and ping me for a review, I'm happy to take a look

Regarding getting some "good first issues"- that came up recently internally as well. Hopefully we can get some hashed out soon to help external contributors get started

PR #2009, opened 2025-09-17
feat(payments): Instantiate client ledger

litt3:
- Actually builds the client ledger when starting the proxy, if configured to do so
- I included a bit of payment docs in a `CLAUDE.md`, not directly related to the other changes 🙃

### DIFF
--- a/api/proxy/store/builder/storage_manager_builder.go
+++ b/api/proxy/store/builder/storage_manager_builder.go
@@ -606,6 +606,12 @@ func buildPayloadDisperser(
 	dispersalMetrics := metrics_v2.NewDispersalMetrics(registry)
 
 	var accountant *clients_v2.Accountant
+	// The legacy `Accountant` is only initialized if using legacy payments.
+	//
+	// There isn't an `else` statement here, because `ClientLedger` (responsible for the new payment system)
+	// construction is handled below by the `buildClientLedger` helper function. The `ClientLedger` cannot be built
+	// here in the same place as the `Accountant` because it requires the `disperserClient` be already built, and the
+	// `Accountant`, if being used, is a part of the `disperserClient`
 	if clientConfigV2.ClientLedgerMode == clientledger.ClientLedgerModeLegacy {
 		// The accountant is populated lazily by disperserClient.PopulateAccountant
 		accountant = clients_v2.NewUnpopulatedAccountant(accountId, accountantMetrics)


api/proxy/store/builder/storage_manager_builder.go
```diff
@@ -679,3 +705,171 @@ func buildLocalSigner(
 
 	return signer, nil
 }
+
+// buildReservationLedger creates a reservation ledger for a given account
+func buildReservationLedger(
+	ctx context.Context,
+	paymentVault payments.PaymentVault,
+	accountID geth_common.Address,
+	now time.Time,
```
anupsv, line 720: would this need to be passed in ?

api/proxy/store/builder/storage_manager_builder.go
```diff
@@ -679,3 +705,171 @@ func buildLocalSigner(
...
+	accountID geth_common.Address,
+	minNumSymbols uint32,
+	disperserClient *clients_v2.DisperserClient,
+) (*ondemand.OnDemandLedger, error) {
+	pricePerSymbol, err := paymentVault.GetPricePerSymbol(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("get price per symbol: %w", err)
+	}
+
+	totalDeposits, err := paymentVault.GetTotalDeposit(ctx, accountID)
```
anupsv, line 777: How often is this updated from onchain changes ?

litt3: It is configurable via flag. Default is 30s

```
		&cli.DurationFlag{
			Name: PaymentVaultMonitorIntervalFlagName,
			Usage: "Interval at which clients poll to check for changes to the PaymentVault contract (relevant " +
				"updates include changes to reservation parameters, and new on-demand payment deposits)",
			Value:    30 * time.Second,
			EnvVars:  []string{withEnvPrefix(envPrefix, "PAYMENT_VAULT_MONITOR_INTERVAL")},
			Category: category,
			Required: false,
		},
```

api/proxy/store/builder/storage_manager_builder.go
litt3: Could you clarify what you mean? Are you asking why it's necessary?

cody-littley: Although I'm not always consistent about it, as a general pattern I think passing in timestamps is a good idea. It makes some types of unit tests a lot easier, since you can provide mock times instead of relying on wall clock timing.

api/proxy/store/builder/storage_manager_builder.go
```diff
@@ -679,3 +705,171 @@ func buildLocalSigner(
...
+	reservationConfig, err := reservation.NewReservationLedgerConfig(
+		*clientReservation,
+		minNumSymbols,
+		// start full since reservation usage isn't persisted: assume the worst case (heavy usage before startup)
+		true,
+		// this is a parameter for flexibility, but there aren't plans to operate with anything other than this value
+		reservation.OverfillOncePermitted,
+		// TODO(litt3): is there a different place we should define this? hardcoding makes sense... it's just a
+		// question of *where*
+		time.Minute,
```
cody-littley, line 750: Is  this not a config option because we want to make it more tricky for somebody to tamper with it?

api/proxy/store/builder/storage_manager_builder.go
```diff
@@ -597,8 +605,11 @@ func buildPayloadDisperser(
 	accountantMetrics := metrics_v2.NewAccountantMetrics(registry)
 	dispersalMetrics := metrics_v2.NewDispersalMetrics(registry)
 
-	// The accountant is populated lazily by disperserClient.PopulateAccountant
-	accountant := clients_v2.NewUnpopulatedAccountant(accountId, accountantMetrics)
+	var accountant *clients_v2.Accountant
+	if clientConfigV2.ClientLedgerMode == clientledger.ClientLedgerModeLegacy {
+		// The accountant is populated lazily by disperserClient.PopulateAccountant
+		accountant = clients_v2.NewUnpopulatedAccountant(accountId, accountantMetrics)
+	}
```
samlaf, line 618: reading this quickly, why are there no else cases here? is it the buildClientLedger that checks the other cases? Are they not mutually exclusive?

Not very important since its just a builder so doesnt have to be best quality code but just not following when reading linearly what happens when the other modes are chosen. Maybe add a small comment?

api/proxy/store/builder/storage_manager_builder.go
litt3: This is a value that we should be choosing. 

For the sake of simplicity, I was planning on hard-coding this. (though like the comment says, I'm not really sure _where_).

An alternative would be to configure this on-chain.

api/proxy/store/builder/storage_manager_builder.go
litt3: b41a91bd8f53f8d10b5534aa089abf3aac81f164

The are mutually exclusive, but they're needed at different types during the construction flow. Left a comment to try to explain, not sure how helpful it will end up being

PR #2010, opened 2025-09-17
refactor(verifier): create verifier v2 (currently just copy of v1)

samlaf:
Part of stack of 3 PRs:
1. https://github.com/Layr-Labs/eigenda/pull/2010
2. https://github.com/Layr-Labs/eigenda/pull/2011
3. https://github.com/Layr-Labs/eigenda/pull/2014

This stack of PRs achieves what I attempted to achieve in https://github.com/Layr-Labs/eigenda/pull/1997, but realized that it wasn't going to work without first forking verifier/prover into v2 versions, since only EigenDA V2 enforces that blobs are powers of 2, which allows for the embedded 28 G1PowerOf2 pts trick.

### Description 

Currently just a copy of v1 verifier. Plan to make modifications in a future PR.

Also decided to get rid of encoding Prover/Verifier interfaces in order to make sure that v2 places don't use the v1 implementation by detail (which wouldn't work because of the powerOf2 blob lengths assumption!). Kept the prover interface in disperser/encoder which was the only place it was being used. Moved the mock implementation there as well. Then had to move a few things around to fix import cycles.

This came as a requirement while trying to introduce https://github.com/Layr-Labs/eigenda/pull/1997, and realizing that the (performance/simplification) modifications introduced there only work for v2.

PR #2011, opened 2025-09-17
refactor(prover):  create prover v2 (currently just copy of v1) 

samlaf:
Part of stack of 3 PRs:
1. https://github.com/Layr-Labs/eigenda/pull/2010
2. ~~https://github.com/Layr-Labs/eigenda/pull/2011~~ https://github.com/Layr-Labs/eigenda/pull/2035
3. https://github.com/Layr-Labs/eigenda/pull/2014

This stack of PRs achieves what I attempted to achieve in https://github.com/Layr-Labs/eigenda/pull/1997, but realized that it wasn't going to work without first forking verifier/prover into v2 versions, since only EigenDA V2 enforces that blobs are powers of 2, which allows for the embedded 28 G1PowerOf2 pts trick.

### Description 

Turns out will also need a prover v2 with the same assumptions as v1 (only deals with powers of 2 blob lengths, which allows for some optimizations/simplifications).


encoding/kzg/prover/v2/prover.go
```diff
@@ -0,0 +1,462 @@
+package prover
```
dmanc, line 1: Alot of this would just be the same code copy pasted into v2 folder?

samlaf: Are you asking if this PR copies the prover code into the v2 folder? If so then yes, see PR title.

samlaf:
Closing in favor of https://github.com/Layr-Labs/eigenda/pull/2035

PR #2013, opened 2025-09-18
feat(proxy): Incorporate ARB server into E2E framework

ethenotethan:
- Introduces lightweight test cases for ARB Custom DA Server using go-ethereum low level RPC client. Vendoring the `DAClient` object maintained by OCL would be overly cumbersome given it relies on many native nitro packages which we'd have to vendor directly. 
- Removes stale `Enabled` field from ARB Custom DA Server config

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


litt3 (dismissed):

api/proxy/servers/arbitrum_altda/handlers.go
```diff
@@ -12,17 +12,17 @@ import (
 	"github.com/ethereum/go-ethereum/common/hexutil"
 )
 
+const (
+	ArbDANamespacedPrefix         = "daprovider"
+	MethodIsValidHeaderByte       = ArbDANamespacedPrefix + "_" + "isValidHeaderByte"
+	MethodStore                   = ArbDANamespacedPrefix + "_" + "store"
+	MethodRecoverBatchFromPayload = ArbDANamespacedPrefix + "_" + "recoverPayloadFromBatch"
```
samlaf: nit: honestly this is one of those cases where I think we prob don't need the `ArbDANamespacedPrefix` variable. Especially now with LLMs that can edit a bunch of repeated code in an instant, I find it much more readable if all the Method names were spelled out in full, with the repeated prefix.

api/proxy/servers/arbitrum_altda/server.go
```diff
@@ -76,6 +76,13 @@ func NewServer(ctx context.Context, cfg *Config, h *Handlers) (*Server, error) {
 
 }
 
+func (svr *Server) Port() int {
```
samlaf, line 81:
```suggestion
// Port returns the port that the server is listening on.
// Useful in case Config.Port was set to 0 to let the OS assign a random port.
func (svr *Server) Port() int {
```

api/proxy/test/testutils/test_suite.go
```diff
@@ -78,35 +84,73 @@ func CreateTestSuite(
 		panic(fmt.Sprintf("build storage managers: %v", err.Error()))
 	}
 
-	proxyServer := rest.NewServer(appConfig.RestSvrCfg, certMgr, keccakMgr, logger, metrics)
-	router := mux.NewRouter()
-	proxyServer.RegisterRoutes(router)
-	if appConfig.StoreBuilderConfig.MemstoreEnabled {
-		memconfig.NewHandlerHTTP(logger, appConfig.StoreBuilderConfig.MemstoreConfig).
-			RegisterMemstoreConfigHandlers(router)
+	if appConfig.EnabledServersConfig.RestAPIConfig.Enabled() {
```
samlaf, line 87: nit: this reads a bit clunky. if we never use the EnabledServerConfig and RestAPIConfig as standalone outside of the appConfig, then might be worth removing the `Config` suffixes?

samlaf (dismissed):

api/proxy/test/testutils/test_suite.go
ethenotethan: true but that would be a larger refactor outside of this PR

api/proxy/servers/arbitrum_altda/server.go
ethenotethan: [2213ee6](https://github.com/Layr-Labs/eigenda/pull/2013/commits/2213ee6838ebb759667449a3ff3ec3649ad47ed5)

api/proxy/servers/arbitrum_altda/handlers.go
ethenotethan: [2213ee6](https://github.com/Layr-Labs/eigenda/pull/2013/commits/2213ee6838ebb759667449a3ff3ec3649ad47ed5)

PR #2015, opened 2025-09-18
feat(payment): Do payment validation on controller

litt3:
- Actually does the payments logic on the controller


cody-littley (approved):

core/payments/reservation/reservation_payment_validator.go
```diff
@@ -48,28 +48,27 @@ func NewReservationPaymentValidator(
 
 // Validates a reservation payment for a blob dispersal
 // The caller is responsible for verifying the signature before calling this method
+//
+// Returns (true, nil) if the reservation has enough capacity to perform the debit.
+// Returns (false, nil) if the bucket lacks capacity to permit the dispersal.
+// Returns (false, error) if an error occurs during validation.
```
samlaf, line 54: nit: you could also use a structured error when bucket lacks capacity and pattern match on it. I guess doing the bool approach is easier though?

core/payments/vault/payment_vault.go
```diff
@@ -114,6 +114,29 @@ func (pv *paymentVault) GetGlobalSymbolsPerSecond(ctx context.Context) (uint64,
...
+	if err != nil {
+		return 0, fmt.Errorf("pack GlobalRatePeriodInterval call: %w", err)
+	}
+
+	returnData, err := pv.ethClient.CallContract(ctx, ethereum.CallMsg{
+		To:   &pv.paymentVaultAddress,
+		Data: callData,
+	}, nil)
+	if err != nil {
+		return 0, fmt.Errorf("global rate period interval eth call: %w", err)
```
samlaf, line 129: please add log with callMsg params (to and calldata) so we can debug if needed

samlaf: this should be added after every eth calls imo. Might actually be worth making the ethClient do this actually..?

samlaf (approved):

core/payments/reservation/reservation_payment_validator.go
samlaf: (please dont change just want to discuss)

litt3: The reason I went with a bool is because it doesn't quite seem right to use an error for a case that is expected and not really an "error". Lots of reservation usages could come back without capacity (at least in theory), in which case it's standard procedure to fall back to on-demand payments

Lots of ways to skin the cat here, but I think it makes sense to use a bool for the `ReservationPaymentValidator`, since that's the strategy that was already used for the underlying `ReservationLedger`

core/payments/vault/payment_vault.go
litt3: Would it be ok if I do a followup PR to address this for all the payment vault methods? (or maybe add to the ethclient, if that makes sense)

core/payments/reservation/reservation_payment_validator.go
samlaf: 🐱

core/payments/vault/payment_vault.go
samlaf: if you have time to create this in the eth client, that would solve the issue once and for all and I would really reslly appreciate you, even more than I currently do :D

litt3: Let me see what it will take, but I'll plan on going that route unless it turns out to be super hard or something 👍

PR #2016, opened 2025-09-18
fix: Copy everything in icicle Dockerfile

dmanc:

Fixes some build issue introduced recently (not sure when)

```
3.813 ../encoding/kzg/cli.go:8:2: no required module provides package github.com/Layr-Labs/eigenda/resources/srs; to add it:
3.813 	go get github.com/Layr-Labs/eigenda/resources/srs
------
icicle.Dockerfile:58
--------------------
  56 |     # Build the server with icicle backend
  57 |     WORKDIR /app/disperser
  58 | >>> RUN go build -tags=icicle -o ./bin/server ./cmd/encoder
  59 |     
  60 |     # Start a new stage for the base image
--------------------
ERROR: failed to solve: process "/bin/sh -c go build -tags=icicle -o ./bin/server ./cmd/encoder" did not complete successfully: exit code: 1
Build references
Check build summary support
Error: buildx bake failed with: ERROR: failed to solve: process "/bin/sh -c go build -tags=icicle -o ./bin/server ./cmd/encoder" did not complete successfully: exit code: 1
```

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/cmd/encoder/icicle.Dockerfile
+++ b/disperser/cmd/encoder/icicle.Dockerfile
@@ -25,14 +25,7 @@ COPY api/proxy/clients ./api/proxy/clients
 RUN go mod download
 
 # Copy the rest of the source code
-COPY ./disperser /app/disperser
-COPY common /app/common
-COPY contracts /app/contracts
-COPY core /app/core
-COPY api /app/api
-COPY indexer /app/indexer
-COPY encoding /app/encoding
-COPY relay /app/relay
+COPY . .
 
 # Define Icicle versions and checksums
 ENV ICICLE_VERSION=3.4


samlaf (commented):
Do we know why this was failing? What changed?
Copying everything in might work but it'll also bloat the image potentially. Do we mind?
Do you have access to the build logs? Maybe we can try to figure out what went wrong in order to fix at the root cause potentially?

If related to SRS, do we just need to copy the resources/srs folder in?

samlaf (approved):
LGTM. I think what's failing is probably because of my change in https://github.com/Layr-Labs/eigenda/pull/1978/files, where the srs package now embeds the g2PowerOf2 srs file. So what we were missing is probably just to COPY the srs files into the container, which your `COPY . .` is now doing.

PR #2017, opened 2025-09-18
feat(payments): Implement basic e2e payments tests

litt3:
- Adds one proxy test each, for reservation and on-demand payments
- On-demand test is disabled for now. I might enable it at somepoint, if I just make a large sepolia deposit for the CI account


api/proxy/test/testutils/setup.go
```diff
@@ -172,16 +175,18 @@ func NewTestConfig(
...
+		Backend:              backend,
+		Retrievers:           []common.RetrieverType{common.RelayRetrieverType, common.ValidatorRetrieverType},
+		Expiration:           14 * 24 * time.Hour,
+		UseKeccak256ModeS3:   false,
+		UseS3Caching:         false,
+		UseS3Fallback:        false,
+		WriteThreadCount:     0,
+		WriteOnCacheMiss:     false,
+		ClientLedgerMode:     clientledger.ClientLedgerModeLegacy,
+		VaultMonitorInterval: 30 * time.Second,
```
ethenotethan, line 189: nit - any reason to set to 30 seconds vs aligning with ETH block times? iiuc this value doesn't even matter for your tests

api/proxy/test/e2e/server_test.go
```diff
@@ -500,6 +501,47 @@ func TestV2ValidatorRetrieverOnly(t *testing.T) {
...
+	requireStandardClientSetGet(t, ts, blob)
+
+	// Verify that dispersal and retrieval succeeded
+	requireDispersalRetrievalEigenDA(t, ts.Metrics.HTTPServerRequestsTotal, commitments.StandardCommitmentMode)
+
+	t.Log("Successfully dispersed and retrieved blob using reservation-only payments")
+}
+
+func TestOnDemandPayments(t *testing.T) {
+	t.Skip("Manual only for now, since we don't have a way of topping up on demand funds automatically")
```
ethenotethan, line 525: probably not possible or worth it but wonder if there's any virtualization we could introduce to proxy's memstore for maintaining a "stubbed view" of payment vault state

api/proxy/test/e2e/server_test.go
```diff
@@ -500,6 +501,47 @@ func TestV2ValidatorRetrieverOnly(t *testing.T) {
...
+	defer kill()
+
+	// Test basic dispersal and retrieval with on-demand payments
+	blob := testutils.RandBytes(1000)
+	requireStandardClientSetGet(t, ts, blob)
+
+	// Verify that dispersal and retrieval succeeded
+	requireDispersalRetrievalEigenDA(t, ts.Metrics.HTTPServerRequestsTotal, commitments.StandardCommitmentMode)
+
+	t.Log("Successfully dispersed and retrieved blob using on-demand-only payments")
```
ethenotethan, line 542: nit any value in also asserting that the cumulative payment value was deducted appropriately?

api/proxy/test/testutils/setup.go
litt3: This is just the same default value as is used elsewhere in the code. Not sure what we would gain from aligning to eth block times?

> iiuc this value doesn't even matter for your tests

That's right, won't actually be used for anything (but needs to not be 0).

api/proxy/test/e2e/server_test.go
litt3: Memstore doesn't even use the clients, right? Trying to do payments without the clients operating would essentially mean re-implementing everything, which probably isn't worth it

api/proxy/test/testutils/setup.go
samlaf: It does make sense to align to block time in test setting because you'll pick up on new values faster though.

api/proxy/test/e2e/server_test.go
samlaf: This could be tested against inabox though, which would prob be better anyways? In inabox we could make a deposit at the beginning (we already set a reservation)

samlaf: Not needed in this PR of course. But maybe worth creating an issue to start considering moving some of our e2e proxy tests to run against inabox?

Do you see any extra value in this running against a real testnet as opposed to inabox?

api/proxy/test/e2e/server_test.go
samlaf: maybe worth it..? guess it depends on whether we want to treat the ledger/accounting as a black box here or not? It is an e2e test after all.. so maybe worth also asserting that if its value is exposed? If not then don't think its worth it personally.

samlaf (approved):

api/proxy/test/e2e/server_test.go
litt3: I'm adding it here because it's quick and easy to get the first bit of automatic testing. I'll be adding inabox testing next.

We've talked of eventually merging these tests into the other testing suites, which I think would be good in the long run. There's no great reason to have these separate, it just so happens that these tests are the easiest to tack onto

api/proxy/test/e2e/server_test.go
litt3: Yeah, this data isn't exposed to be inspected at this level.

If we added endpoints to query payment state it might make sense to add in the future

api/proxy/test/testutils/setup.go
litt3: If we cared to pick up new values, I'd agree. But I don't anticipate adding any tests to this suite that rely on dynamic `PaymentVault` state, just due to how impossible that would be to instrument.

I do intend to add those kind of tests to run in inabox, though

samlaf (approved):

Issue #2022, opened 2025-09-19
[Enhancement]: Remove indexer test files and mock implementations


cosmoknightt:
### Use case and current behavior

This is Phase 1 of removing the deprecated indexer functionality.

Remove all test files and mock implementations related to indexer. This is the safest first step since it only affects test code.

### Enhancement

Based on #2005

PR #2026, opened 2025-09-19
fix: remove indexer test files and mocks

cosmoknightt:
Closes https://github.com/Layr-Labs/eigenda/issues/2022

Remove all test files and mock implementations related to the deprecated
indexer functionality. This is Phase 1 of indexer cleanup.

Files removed:
- indexer/test/ directory with all test utilities
- indexer test files: header_service_test.go, header_store_test.go
- core/indexer test files: indexer_suite_test.go, state_test.go
- indexer/mock/ directory
- core/mock/operator_sockets_filterer.go
- api/clients/retrieval_client_test.go (depends on indexer mocks)


api/clients/retrieval_client_test.go
litt3, line 1: @cosmoknightt The `indexer` isn't used in production code, but the mock implementation is apparently a key part of tests, such as this file.

My guess would be that the `indexer` implementation of the `Indexer` interface may be deletable, but not the interface itself, nor the mock implementation. This may change at some point, but for now we need to be careful to not delete any test coverage of system components

cosmoknightt: @litt3 Thank you for the feedback

I've updated the PR to be more conservative:
- Only removed test files specific to the indexer implementation
- Preserved `api/clients/retrieval_client_test.go` and all mock files used by other components
- This ensures no test coverage is lost for non-indexer components

The updated PR now only removes:
- Core indexer test files (indexer_suite_test.go, state_test.go)
- Indexer storage implementation tests (leveldb/header_store_test.go, inmem/header_store_test.go)
- Indexer service tests (eth/header_service_test.go)
- The indexer test from Makefile integration-tests target

Would you mind taking another look?

PR #2028, opened 2025-09-19
feat: remove indexer internal implementations and fix compile issues

cosmoknightt:
Remove all internal implementation files while preserving interface definitions. Also fix compile issues by removing node.go dependency on indexer. This is Phase 3 of indexer cleanup.

Core indexer implementations removed:
- core/indexer/indexer.go (factory function)
- core/indexer/operator_*.go (operator implementations)
- core/indexer/state.go (state management)
- core/mock/operator_sockets_filterer.go (mock file)

Main indexer implementations removed:
- indexer/inmem/ (in-memory storage implementation)
- indexer/leveldb/ (LevelDB storage implementation)
- indexer/eth/ (Ethereum integration implementation)
- indexer/upgrader/ (upgrader implementation)

Node.go fixes (partial Phase 4 content):
- Remove OperatorSocketsFilterer field and import
- Disable socket monitoring in checkRegisteredNodeIpOnChain()
- Add TODO comments for future thegraph implementation

Interface definitions preserved:
- indexer/indexer.go (main interfaces)
- indexer/config.go (configuration structs)
- indexer/accumulator.go, filterer.go, etc. (interface definitions)
- core/indexer/errors.go, upgrader.go (base definitions)


node/node.go
```diff
@@ -227,10 +225,8 @@ func NewNode(
 		return nil, fmt.Errorf("failed to create new store: %w", err)
 	}
 
-	socketsFilterer, err := indexer.NewOperatorSocketsFilterer(serviceManagerAddress, client)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create new operator sockets filterer: %w", err)
-	}
+	// TODO: OperatorSocketsFilterer functionality was removed due to indexer deprecation.
```
dmanc, line 228: Interesting. So this seems to be the only production use case of the inmem indexer right now? Graph node would not be a good option because that would require every operator to take a dependency on that infrastructure.

cosmoknightt: Ah yeah, make sense, havnt't consider this aspect, emm TBH this is the third PR of a series of index  deleting, Im making this a draft now until the previous PR has been merged. Maybe I will redesign this later. Anyway, thanks for feedback @dmanc

PR #2033, opened 2025-09-19
ci: Add codecov to unit-test.yml

dmanc:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


codecov.yml
```diff
@@ -0,0 +1,138 @@
...
+
+  litt-tests:
+    paths:
+      - "litt/"
+    carryforward: false
+
+# Codecov CI configuration
+codecov:
+  notify:
+    # Wait for all CI builds before sending notifications
```
samlaf: ci builds? or ci runs?

codecov.yml
```diff
@@ -0,0 +1,138 @@
...
+    wait_for_ci: true
+
+  # Require the codecov token for security
+  require_ci_to_pass: true
```
samlaf: doesnt seem related to token...?

codecov.yml
```diff
@@ -0,0 +1,138 @@
...
+  behavior: default
+  require_changes: false
+  require_base: false
+  require_head: true
+
+# Flag configuration for different test types
+flags:
+  unit-tests:
+    paths:
+      - "!contracts/"
```
samlaf: why not contracts?

README.md
```diff
@@ -2,7 +2,7 @@
 ![Integration Tests](https://github.com/Layr-Labs/eigenda/actions/workflows/integration-tests.yml/badge.svg)
 ![Linter](https://github.com/Layr-Labs/eigenda/actions/workflows/golangci-lint.yml/badge.svg)
 ![Contracts](https://github.com/Layr-Labs/eigenda/actions/workflows/test-contracts.yml/badge.svg)
-![Go Coverage](https://github.com/Layr-Labs/eigenda/wiki/coverage.svg)
+[![codecov](https://codecov.io/github/Layr-Labs/eigenda/graph/badge.svg?token=EKLGVKW1VN)](https://codecov.io/github/Layr-Labs/eigenda)
```
samlaf, line 5: which coverage will this show? we output two separate ones right? for littdb and the other tests

samlaf (approved):

README.md
dmanc: This will be the combined coverage, we upload the artifacts separately but codecov handles merging them into one coverage %.

codecov.yml
dmanc: contracts has bindings.go and it ends up making our coverage lower since it's 0% coverage

codecov.yml
dmanc: removed a3db980b50b254adb3e8fc3a681cf529656c01f1, it's already true by default

codecov.yml
dmanc: a3db980b50b254adb3e8fc3a681cf529656c01f1, already true by default so i just removed

codecov.yml
samlaf: is code coverage only for golang stuff? or do we also have code coverage for our solidity contracts?

samlaf (dismissed):

codecov.yml
dmanc: I didn't consider that there's probably code coverage files for solidity. We could add that in a future PR?

samlaf (approved):

PR #2037, opened 2025-09-19
chore: Deprecate ginko

dmanc:

Standarize on using the standard go testing framework for inabox + random indexer test.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/inabox/tests/integration_suite_test.go
+++ b/inabox/tests/integration_suite_test.go
@@ -99,7 +99,11 @@ func TestMain(m *testing.M) {
 	}
 
 	// Run suite setup
-	setupSuite()
+	if err := setupSuite(); err != nil {
+		logger.Error("Setup failed:", err)
+		teardownSuite()
+		os.Exit(1)
+	}
 
 	// Run all tests
 	code := m.Run()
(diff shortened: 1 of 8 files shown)


inabox/tests/integration_suite_test.go
```diff
@@ -91,26 +90,61 @@ func init() {
...
+	// Wait for setup to complete or timeout
+	select {
+	case err := <-setupDone:
+		if err != nil {
+			logger.Fatal("Setup failed:", err)
+		}
+	case <-ctx.Done():
+		logger.Fatal("Setup timed out after 1 minute")
+	}
+}
```
samlaf: hmm this feels wrong. setup() function itself should deal with timeout. Doing it like this will leak resources if setup times out. so all the docker stuff you created will be left open. setup function itself should manage the select and delete what it started when ctx.Done() is reached. Also probably should avoid logger.Fatal in tests since that prevents shutdown defer statements from running to cleanup resources

inabox/tests/integration_suite_test.go
```diff
@@ -435,24 +502,57 @@ func setupRetrievalClients(testConfig *deploy.Config) error {
...
+	}()
+
+	// Wait for teardown to complete or timeout
+	select {
+	case <-teardownDone:
+		logger.Info("Teardown completed successfully")
+	case <-ctx.Done():
+		logger.Warn("Teardown timed out after 1 minute - some resources may not have been cleaned up")
+	}
+}
```
samlaf: same thing here doesn't feel like the right way. But maybe our setup/shutdown code is not good enough to actually handle timesouts probably. Feel like we should just let it hang, and force us to make it better, instead of using this kind of workaround though

inabox/tests/integration_suite_test.go
dmanc: I simplified the code a bit https://github.com/Layr-Labs/eigenda/pull/2037/commits/faf932b0e7ca9d8f582bbec968ab97a2f1c36d9e, lmk if this is what you had in mind

samlaf: Yeah I think this is cleaner. Liked having the timeout defined in main and passed in to setup suite and tear down fcts though, cause then we can see it explicitly.

also why change 1min to 5min?

samlaf (approved):

inabox/tests/integration_suite_test.go
dmanc: 1 min wasn't enough when I tried it in CI so I just did 5 min

PR #2041, opened 2025-09-22
chore: Remove inabox dependency from plugin_test.go

dmanc:

Removes the `inabox/deploy/deploy.go` dependency in plugin_test.go by replacing it with testbed based deployment. Goal is to eventually replace `inabox/deploy/deploy.go` with the testbed package.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


inabox/deploy/deploy.go
```diff
@@ -388,38 +387,3 @@ func (env *Config) StopBinaries() {
 		logger.Fatal("Failed to stop binaries", "error", err)
 	}
 }
-
-func (env *Config) RunNodePluginBinary(operation string, operator OperatorVars) {
```
samlaf, line 392: am I understanding correctly that this was never used for inabox, and was only used for the plugin_test.go test. Any idea why it was added here..?

node/plugin/tests/plugin_test.go
```diff
@@ -20,24 +22,38 @@ import (
...
-	testName     string
-
 	logger = test.GetLogger()
 
 	// Shared test resources
 	anvilContainer *testbed.AnvilContainer
-	testConfig     *deploy.Config
+	deployResult   *testbed.DeploymentResult
+	privateKeys    *testbed.PrivateKeyMaps
+	testOperator   OperatorConfig
```
samlaf, line 32: for a follow-up PR: we should stop using these global vars and instead have the testbed return an actual testbed... which we use in each Test separately.

node/plugin/tests/plugin_test.go
```diff
@@ -50,34 +66,48 @@ func TestMain(m *testing.M) {
...
-		}
-	}
-
-	testConfig = deploy.NewTestConfig(testName, rootPath)
-	testConfig.Deployers[0].DeploySubgraphs = false
 
-	logger.Info("Starting anvil")
 	var err error
 	anvilContainer, err = testbed.NewAnvilContainerWithOptions(ctx, testbed.AnvilOptions{
-		ExposeHostPort: true, // This will bind container port 8545 to host port 8545
```
samlaf, line 69: why remove this comment? is it no longer true?

node/plugin/tests/plugin_test.go
```diff
@@ -266,3 +317,49 @@ func getTransactor(t *testing.T, operator deploy.OperatorVars) *eth.Writer {
...
+		"--ecdsa-key-password", operator.NODE_ECDSA_KEY_PASSWORD,
+		"--bls-key-password", operator.NODE_BLS_KEY_PASSWORD,
+		"--socket", socket,
+		"--quorum-id-list", operator.NODE_QUORUM_ID_LIST,
+		"--chain-rpc", operator.NODE_CHAIN_RPC,
+		"--eigenda-directory", operator.NODE_EIGENDA_DIRECTORY,
+		"--bls-operator-state-retriever", operator.NODE_BLS_OPERATOR_STATE_RETRIVER,
+		"--eigenda-service-manager", operator.NODE_EIGENDA_SERVICE_MANAGER,
+		"--churner-url", operator.NODE_CHURNER_URL,
+		"--num-confirmations", "0",
```
samlaf, line 354: what exactly are we trying to test with this? that the flags work properly? If not then why don't we just import the plugin function and call it directly, instead of subshelling like this? Do we gain much from doing this this way?

node/plugin/tests/plugin_test.go
dmanc: Yeah I want to do this once everything is deployed through testbed (still missing EigenDA components for example).

inabox/deploy/deploy.go
dmanc: Not sure to be honest, definitely felt out of place.

node/plugin/tests/plugin_test.go
dmanc: Not sure either, this tests is pretty old but lemme see if I can simplify it a bit

node/plugin/tests/plugin_test.go
dmanc: d9783ff0c8a26b11f37b73d1a1e981dec018fc2f

node/plugin/tests/plugin_test.go
dmanc: Going to leave it as is, the motivation from the original test writer was to test that the CLI functions properly.

samlaf (approved):

PR #2042, opened 2025-09-22
feat(payments): Add client ledger to example

litt3:
- Actually instantiate client ledger instead of using `nil`

### DIFF
--- a/api/clients/v2/examples/client_construction.go
+++ b/api/clients/v2/examples/client_construction.go
@@ -82,9 +82,26 @@ func createPayloadDisperser(privateKeyHex string) (*payloaddispersal.PayloadDisp
 		return nil, fmt.Errorf("create cert verifier: %w", err)
 	}
 
-	certBuilder, err := createCertBuilder(contractDirectory)
+	operatorStateRetrieverAddr, err := contractDirectory.GetContractAddress(
+		context.Background(), directory.OperatorStateRetriever)
 	if err != nil {
-		return nil, fmt.Errorf("create cert builder: %w", err)
+		return nil, fmt.Errorf("get OperatorStateRetriever address: %w", err)
+	}
+
+	registryCoordinatorAddr, err := contractDirectory.GetContractAddress(
+		context.Background(), directory.RegistryCoordinator)
+	if err != nil {
+		return nil, fmt.Errorf("get RegistryCoordinator address: %w", err)
+	}
+
+	certBuilder, err := clients.NewCertBuilder(
+		logger,
+		operatorStateRetrieverAddr,
+		registryCoordinatorAddr,
+		ethClient,
+	)
+	if err != nil {
+		return nil, fmt.Errorf("new cert builder: %w", err)
 	}
 
 	blockNumMonitor, err := createBlockNumberMonitor()
@@ -135,12 +152,18 @@ func createPayloadDisperser(privateKeyHex string) (*payloaddispersal.PayloadDisp
 func createRelayPayloadRetriever(
 	logger logging.Logger,
 	ethClient *geth.EthClient,
-	contractDirectory *directory.ContractDirectory,
+	operatorStateRetrieverAddr gethcommon.Address,
+	serviceManagerAddr gethcommon.Address,
 ) (*payloadretrieval.RelayPayloadRetriever, error) {
 
-	reader, err := createEthReader(logger, ethClient, contractDirectory)
+	reader, err := eth.NewReader(
+		logger,
+		ethClient,
+		operatorStateRetrieverAddr.Hex(),
+		serviceManagerAddr.Hex(),
+	)
 	if err != nil {
-		return nil, fmt.Errorf("create eth reader: %w", err)
+		return nil, fmt.Errorf("new reader: %w", err)
 	}
 
 	relayClient, err := createRelayClient(logger, ethClient, reader.GetRelayRegistryAddress())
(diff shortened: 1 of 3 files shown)


api/clients/v2/examples/client_construction.go
```diff
@@ -18,28 +21,33 @@ import (
...
-	certVerifierRouterAddress  = "0x7F40A8e1B62aa1c8Afed23f6E8bAe0D340A4BC4e"
-	registryCoordinatorAddress = "0x53012C69A189cfA2D9d29eb6F19B32e0A2EA3490"
-	// These two addresses are no longer required for the Eth Client, but parameter is still being taken until we deprecate the flags
-	eigenDAServiceManagerAddress = ""
-	// operatorStateRetrieverAddress is still used for CertBuilder
-	operatorStateRetrieverAddress = "0x003497Dd77E5B73C40e8aCbB562C8bb0410320E7"
+	ethRPCURL         = "https://ethereum-sepolia-rpc.publicnode.com"
+	disperserHostname = "disperser-testnet-sepolia.eigenda.xyz"
+	// Contract registry address for Sepolia - this allows fetching all other contract addresses
+	contractRegistryAddress = "0x9620dC4B3564198554e4D2b06dEFB7A369D90257"
```
samlaf:
```suggestion
	eigenDADirectoryAddress = "0x9620dC4B3564198554e4D2b06dEFB7A369D90257"
```

samlaf: let's use the actual contract names otherwise we'll confuse people (I'm actually confused myself right now given we actually have contracts named registryXYZ)

api/clients/v2/examples/client_construction.go
```diff
@@ -18,28 +21,33 @@ import (
...
+	// CertVerifierRouter is not available from the contract registry and must be specified directly
+	certVerifierRouterAddress = "0x58D2B844a894f00b7E6F9F492b9F43aD54Cd4429"
```
samlaf: don't think this is true? pretty sure we have the default router in the directory?

api/clients/v2/examples/client_construction.go
```diff
@@ -142,8 +185,13 @@ func createValidatorPayloadRetriever() (*payloadretrieval.ValidatorPayloadRetrie
 		return nil, fmt.Errorf("create eth client: %w", err)
 	}
 
+	contractDirectory, err := createContractDirectory(context.Background(), logger, ethClient)
+	if err != nil {
+		return nil, fmt.Errorf("create contract directory: %w", err)
+	}
+
```
samlaf: nit: not related to this PR but I've noticed in general we seem to call the contractDirectory at random places whenever an address is needed.

I think overall I would prefer if we fetched all addresses that we need at the very beginning of build processes (whether this be in prod code, or test code, or examples like this) for consistency. This way we know from the start if it fails and theres a config issue. Also I prefer the dependency injection way of passing down the address that a service needs only, not the directory and then letting each contract understand how to fetch the addresses it needs. Thoughts?

api/clients/v2/examples/client_construction.go
```diff
@@ -363,3 +438,153 @@ func createLogger() (logging.Logger, error) {
...
+	if err != nil {
+		return nil, fmt.Errorf("get PaymentVault address: %w", err)
+	}
+
+	paymentVault, err := vault.NewPaymentVault(logger, ethClient, paymentVaultAddr)
+	if err != nil {
+		return nil, fmt.Errorf("new payment vault: %w", err)
+	}
+
+	minNumSymbols, err := paymentVault.GetMinNumSymbols(ctx)
```
samlaf, line 411: does the client monitor for changes to this?
Also for my own knowledge who owns the source of truth wrt this? is it validators or dispersers? afaiu this is only used for disperser charging client, which the disperser doesn't really need to follow the onchain logic right? So it could set/use whatever, and following the contract is just a convention?

api/clients/v2/examples/client_construction.go
cody-littley: I agree that we should avoid passing the contract directory into constructors, and should instead pass in contract addresses. Makes things a lot easier to unit test. Otherwise, we'd need a mock implementation of the contract directory.

api/clients/v2/examples/client_construction.go
litt3: f80cb43f02abc4240b158df3c4f0a645687a3296

api/clients/v2/examples/client_construction.go
litt3: 0fdb7d56770f32e5e7b38e17a1dc1e62f284583f

api/clients/v2/examples/client_construction.go
litt3: I agree in general, passing in the address is the better pattern, which we should use.

For the case of the examples... I feel like any way I slice this it ends up not-great. TBH I regret writing these examples- they are hard to maintain, and we have plenty of examples within our code and tests, which would help people construct these manually, if they wanted to.

Modified this PR, to get the addresses earlier in the construction process, and pass them in as parameters
7c2fc3ded9354befbe044ef7d324be9fe141c35d
243acdd1ef32b60227305c90f581e7b16f34555b

api/clients/v2/examples/client_construction.go
litt3: > does the client monitor for changes to this?

No. This falls into the category of parameters that we cannot change with the current system

The disperser is the source of truth for on-demand payments, and the validators are the source of truth for reservation payments. Nevertheless, we rely on different parties independently tracking payment state, so it's important for everyone to precisely agree on what this value is. Without anchoring the value to RBN (or wallclock time, in the case of reservations), there's no way to get everyone to make the change at the same time, therefore this parameter cannot be safely modified

api/clients/v2/examples/client_construction.go
samlaf: > we have plenty of examples within our code and tests, which would help people construct these manually, if they wanted to.

I think I agree with you. We can discuss this. That being said, is it only painful now because you're making a huge number of changes, but once its there they'll be stable and a nice reference for others? I'm not sure...

PR #2044, opened 2025-09-22
feat(payments): Add controller payment metrics

litt3:
- Adds a bunch of metrics to the controller stack of payment utilities
- Adds better context for payment authorization failures

### DIFF
--- a/api/proxy/test/e2e/server_test.go
+++ b/api/proxy/test/e2e/server_test.go
@@ -14,6 +14,7 @@ import (
 	"github.com/Layr-Labs/eigenda/api/proxy/metrics"
 	"github.com/Layr-Labs/eigenda/api/proxy/store/secondary"
 	"github.com/Layr-Labs/eigenda/api/proxy/test/testutils"
+	"github.com/Layr-Labs/eigenda/core/payments/clientledger"
 	"github.com/Layr-Labs/eigenda/encoding"
 	"github.com/Layr-Labs/eigenda/encoding/utils/codec"
 	"github.com/Layr-Labs/eigenda/test/random"
@@ -498,6 +499,47 @@ func TestV2ValidatorRetrieverOnly(t *testing.T) {
 
 	requireStandardClientSetGet(t, ts, testutils.RandBytes(1000))
 	requireDispersalRetrievalEigenDA(t, ts.Metrics.HTTPServerRequestsTotal, commitments.StandardCommitmentMode)
+}
+
+func TestReservationPayments(t *testing.T) {
+	t.Parallel()
+
+	testCfg := testutils.NewTestConfig(testutils.GetBackend(), common.V2EigenDABackend, nil)
+	testCfg.ClientLedgerMode = clientledger.ClientLedgerModeReservationOnly
+
+	tsConfig := testutils.BuildTestSuiteConfig(testCfg)
+	ts, kill := testutils.CreateTestSuite(tsConfig)
+	defer kill()
+
+	// Test basic dispersal and retrieval with reservation payments
+	blob := testutils.RandBytes(1000)
+	requireStandardClientSetGet(t, ts, blob)
+
+	// Verify that dispersal and retrieval succeeded
+	requireDispersalRetrievalEigenDA(t, ts.Metrics.HTTPServerRequestsTotal, commitments.StandardCommitmentMode)
+
+	t.Log("Successfully dispersed and retrieved blob using reservation-only payments")
+}
+
+func TestOnDemandPayments(t *testing.T) {
+	t.Skip("Manual only for now, since we don't have a way of topping up on demand funds automatically")
+	t.Parallel()
+
+	testCfg := testutils.NewTestConfig(testutils.GetBackend(), common.V2EigenDABackend, nil)
+	testCfg.ClientLedgerMode = clientledger.ClientLedgerModeOnDemandOnly
+
+	tsConfig := testutils.BuildTestSuiteConfig(testCfg)
+	ts, kill := testutils.CreateTestSuite(tsConfig)
+	defer kill()
+
+	// Test basic dispersal and retrieval with on-demand payments
+	blob := testutils.RandBytes(1000)
+	requireStandardClientSetGet(t, ts, blob)
+
+	// Verify that dispersal and retrieval succeeded
+	requireDispersalRetrievalEigenDA(t, ts.Metrics.HTTPServerRequestsTotal, commitments.StandardCommitmentMode)
+
+	t.Log("Successfully dispersed and retrieved blob using on-demand-only payments")
 }
 
 // requireDispersalRetrievalEigenDA ... ensure that blob was successfully dispersed/read to/from EigenDA
--- a/api/proxy/test/testutils/setup.go
+++ b/api/proxy/test/testutils/setup.go
@@ -148,6 +148,9 @@ type TestConfig struct {
 	UseKeccak256ModeS3 bool
 	UseS3Caching       bool
 	UseS3Fallback      bool
+
+	ClientLedgerMode     clientledger.ClientLedgerMode
+	VaultMonitorInterval time.Duration
 }
 
 // NewTestConfig returns a new TestConfig
@@ -172,16 +175,18 @@ func NewTestConfig(
 			OpKeccakCommitment:  true,
 			StandardCommitment:  true,
 		},
-		BackendsToEnable:   backendsToEnable,
-		DispersalBackend:   dispersalBackend,
-		Backend:            backend,
-		Retrievers:         []common.RetrieverType{common.RelayRetrieverType, common.ValidatorRetrieverType},
-		Expiration:         14 * 24 * time.Hour,
-		UseKeccak256ModeS3: false,
-		UseS3Caching:       false,
-		UseS3Fallback:      false,
-		WriteThreadCount:   0,
-		WriteOnCacheMiss:   false,
+		BackendsToEnable:     backendsToEnable,
+		DispersalBackend:     dispersalBackend,
+		Backend:              backend,
+		Retrievers:           []common.RetrieverType{common.RelayRetrieverType, common.ValidatorRetrieverType},
+		Expiration:           14 * 24 * time.Hour,
+		UseKeccak256ModeS3:   false,
+		UseS3Caching:         false,
+		UseS3Fallback:        false,
+		WriteThreadCount:     0,
+		WriteOnCacheMiss:     false,
+		ClientLedgerMode:     clientledger.ClientLedgerModeLegacy,
+		VaultMonitorInterval: 30 * time.Second,
 	}
 }
 
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -42,6 +42,7 @@ group "all" {
     "batcher",
     "disperser",
     "encoder",
+    "encoder-icicle",
     "retriever",
     "churner",
     "dataapi",
@@ -122,6 +123,9 @@ target "encoder" {
 target "encoder-icicle" {
   context    = "."
   dockerfile = "./disperser/cmd/encoder/icicle.Dockerfile"
+  // Currently needed because Dockerfile has amd64 hardcoded in a few places.
+  // TODO: make Dockerfile also work for arm.
+  platforms  = ["linux/amd64"]
   tags       = ["${REGISTRY}/${REPO}/encoder-icicle:${BUILD_TAG}"]
 }
 
--- a/encoding/kzg/prover/v2/icicle/multiframe_proof.go
+++ b/encoding/kzg/prover/v2/icicle/multiframe_proof.go
@@ -38,7 +38,7 @@ type WorkerResult struct {
 
 // This function supports batching over multiple blobs.
 // All blobs must have same size and concatenated passed as polyFr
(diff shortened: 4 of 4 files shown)


core/meterer/on_demand_meterer_metrics.go
```diff
@@ -0,0 +1,55 @@
...
+func NewOnDemandMetererMetrics(
+	registry *prometheus.Registry,
+	namespace string,
+	subsystem string,
+) *OnDemandMetererMetrics {
+	if registry == nil {
+		return nil
+	}
+
+	onDemandGlobalMeterExhaustedRequests := prometheus.NewCounter(
```
cody-littley: Nit, I personally prefer the following pattern:

```suggestion
	onDemandGlobalMeterExhaustedRequests := promauto.With(registry).NewCounter(
```

This way, you can't accidentally forget to register a metric.

core/meterer/on_demand_meterer_metrics.go
```diff
@@ -0,0 +1,55 @@
...
+	)
+
+	onDemandGlobalMeterExhaustedSymbols := prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "on_demand_global_meter_exhausted_symbols_count",
+			Subsystem: subsystem,
+			Help:      "Total number of symbols rejected due to global rate limit",
+		},
+	)
```
samlaf: nit: could have made these labels on one single metric

core/meterer/on_demand_meterer_metrics.go
```diff
@@ -0,0 +1,55 @@
...
+		onDemandGlobalMeterExhaustedRequests: onDemandGlobalMeterExhaustedRequests,
+		onDemandGlobalMeterExhaustedSymbols:  onDemandGlobalMeterExhaustedSymbols,
+	}
+}
+
+// RecordGlobalMeterExhaustion records a request rejection due to global rate limit
+func (m *OnDemandMetererMetrics) RecordGlobalMeterExhaustion(symbolCount uint32) {
+	if m == nil {
+		return
+	}
```
samlaf, line 51: I'm really not a big fan of these checks. I don't think a pointer should ever be allowed to be nil silently like this, as it will prevent catching actual misconfigurations. Prefer to not have these, let the program panic, and we fix the thing. If you actually don't want a metrics instance, then that should be checked somewhere else above (either use an interface, or skip this call altogether in call sites).

core/payments/ondemand/on_demand_cache_metrics.go
```diff
@@ -0,0 +1,88 @@
...
+	m.cacheSize = prometheus.NewGaugeFunc(
+		prometheus.GaugeOpts{
+			Namespace: m.namespace,
+			Name:      "on_demand_ledger_cache_size",
+			Subsystem: m.subsystem,
+			Help:      "Current number of entries in the on-demand ledger cache",
+		},
+		func() float64 {
+			return float64(sizeGetter())
+		},
```
samlaf, line 67: this is a lot of gymnastics. does sizeGetter really need to be a function? this is only ever registered once, so wouldnt change right? Can we just pass a float instead?

Passing a function that returns an int which gets mapped to a float64 which gets wrapped in another function...... what's happening here exactly?

samlaf: Can this function be called multiple times? is that safe? I'm a big confused by this thing

core/payments/ondemand/on_demand_validator_metrics.go
```diff
@@ -0,0 +1,107 @@
...
+import (
+	"github.com/prometheus/client_golang/prometheus"
+)
+
+// Tracks metrics for the [OnDemandPaymentValidator]
+type OnDemandValidatorMetrics struct {
+	onDemandSymbolCount        prometheus.Histogram
+	onDemandInsufficientFunds  prometheus.Counter
+	onDemandQuorumNotSupported prometheus.Counter
+	onDemandUnexpectedErrors   prometheus.Counter
```
samlaf, line 13: do we really need a metrics for this? or is an error log sufficient? Are we expecting this to actually happen more than once in a blue moon? is the purpose here that alerting based on a metrics is simpler than alerting based on log and we don't want to create metrics in grafana based on logs?

core/payments/reservation/reservation_cache_metrics.go
```diff
@@ -0,0 +1,130 @@
...
+}
+
+func NewReservationCacheMetrics(
+	registry *prometheus.Registry,
+	namespace string,
+	subsystem string,
+) *ReservationCacheMetrics {
+	if registry == nil {
+		return nil
+	}
```
samlaf, line 27: not a fan of all of these nil checks, as I've mentioned elsewhere.

core/payments/reservation/reservation_validator_metrics.go
```diff
@@ -0,0 +1,147 @@
...
+)
+
+// Tracks metrics for the [ReservationPaymentValidator]
+type ReservationValidatorMetrics struct {
+	reservationSymbolCount           prometheus.Histogram
+	reservationInsufficientBandwidth prometheus.Counter
+	reservationQuorumNotPermitted    prometheus.Counter
+	reservationTimeOutOfRange        prometheus.Counter
+	reservationTimeMovedBackward     prometheus.Counter
+	reservationUnexpectedErrors      prometheus.Counter
```
samlaf, line 15: should we just bundle these into a single error count? and users can look at logs for error specifics?

core/meterer/on_demand_meterer_metrics.go
litt3: This is a preference I've been following at the request of @cody-littley.

I agree with Cody and prefer this current strategy over no-op impls of an interface: metrics are often nil, and it's much less verbose this way. I wouldn't like the nil receiver pattern for most cases, but for metrics it doesn't bother me. I don't feel strongly, though.

core/meterer/on_demand_meterer_metrics.go
cody-littley: Several other places where you'd want to adopt this pattern as well, if you take the change.

cody-littley (dismissed):

core/payments/ondemand/on_demand_cache_metrics.go
litt3: The cache size is being defined as a _pull_ metric, instead of _push_. I don't want to be calling `cache.Len()` and reporting the value for every cache access: registering a gauge with prometheus, which just fetches the size at the rate of normal metrics polling seems to make sense.

> Passing a function that returns an int which gets mapped to a float64 which gets wrapped in another function

- cache length is defined as an integer
- the metric is reported as a float

The current impl lets the cache register the function gauge without knowing that the metric needs to be a float. Alternatively, we could define the function gauge parameter as `sizeGetter func() float64`, and let the cache do the casting when registering the gauge. The way I currently have it seems clean to me, but I wouldn't mind changing where the cast takes place if you have a preference.

core/meterer/on_demand_meterer_metrics.go
cody-littley: Perhaps worth having a bike shed conversation on this.

The point you bring up about silently not doing something when you actually want to do it is valid.

There are a few reasons why I typically treat metric classes like this:

- metrics produce no return values, and the program can function in a sane way if they are no-ops
- An interface + mock is extremely verbose for metrics, especially since metrics classes often have lots of methods.
- I'm not a fan of doing nil checks in the caller
    - adds complexity to the business logic part of the code :reach-for-club:
    - allows you to silently fail with nil the same way as if you did the check inside the metrics instance
- the nil checks inside the metrics class don't add too much complexity, since most of these functions are basically just one-liners.

One possible compromise might be to omit an error level log message whenever the metrics are nil. That way if it is that way in production we at least make some noise about it.

@samlaf, would you like to resolve this prior @litt3 merging, or would it be ok to discuss asyncronously? (There are many places in core code currently that do metrics like this.)

core/payments/ondemand/on_demand_validator_metrics.go
litt3: > do we really need a metrics for this?

I'm not sure this is the question we should ask when deciding to add a metric. It seems to imply that a new metric has a high cost, which it really doesn't, right?

Looking at data in a chart is vastly easier than trying to go through logs, and it lets you stack the data up against other metrics to establish correlations.

Metrics are cheap, and having good metric coverage for all the edge cases means I can make a single dashboard that tells me everything I need to know about the payment system, and that would be really valuable to have if there's ever an incident in prod. If the once-in-a-blue-moon actually happens, I'd be very glad to have packaged this into a metric

core/payments/reservation/reservation_cache_metrics.go
litt3: https://github.com/Layr-Labs/eigenda/pull/2044#discussion_r2370049239

Let's have the conversation here

core/payments/reservation/reservation_validator_metrics.go
litt3: This gets at the same root issue as [this conversation](https://github.com/Layr-Labs/eigenda/pull/2044#discussion_r2370237624)

Looking at metrics for prod incidents is way better than looking at logs, IMO.

core/meterer/on_demand_meterer_metrics.go
litt3: Seems a bit odd to have multiple totally different things in a single metric. Labels seem like they should be used for multiple instances of the same metric, i.e. `node A`'s value, and also `node B`s.

Maybe I'm wrong though- do people bundle metrics together like that? What's the benefit?

core/meterer/on_demand_meterer_metrics.go
litt3: 6b195efe3173fda89faf744454654e1e5bd2bfe4

core/meterer/on_demand_meterer_metrics.go
samlaf: They are the same thing though in my brain, just different units on the same event (count vs bytes).

core/meterer/on_demand_meterer_metrics.go
samlaf: I see. No need to block on this discussion, we should def have a bike shed on this.

I'm not sure I understand what the issue with the noop interface implementation is though, especially with LLMs that can generate those very easily.

I think the way I would be convinced for an approach over another is purely performance wise (even though it doesn't matter much in most of our codebase, it at least feels like A way to resolve this issue). IMO the noop implementation is both faster performance wise (not sure but I'm thinking the compiler might even be able to completely inline + remove the function calls). AND its also cleaner and less chance of misconfiguration bugs.

core/payments/ondemand/on_demand_cache_metrics.go
samlaf: > The cache size is being defined as a pull metric, instead of push.

hmm might be worth having a quick call tomorrow, as I'm not sure I'm following.
iiuc what you're trying to do though prometheus has ways to do those, and they are called exported metrics. You register a function and its called whenever the /metrics endpoint is called: https://prometheus.io/docs/instrumenting/exporters/
Typically its done to (re)export metrics from another binary, but I guess it could also be used to just poll an internal service.

core/payments/ondemand/on_demand_validator_metrics.go
samlaf: Weird you seemed to have completely changed your perspective since luke's PR a while ago where he was adding a lot of metrics for each client errors, duplicating logs, and iirc you were also not agreeing with that.

we can add all the metrics we want but at the end of the day someone needs to arrange a diagram to show them, and we need to have alerts setup for them. metrics should be **statistical** averages. They are used to just get an alert/see something went wrong, and then you go digging for what, in the logs.

core/payments/reservation/reservation_validator_metrics.go
samlaf: I very much disagree. The log has all the info and you'll always end up reading the logs. You'll want to understand WHY a time moved backward for example, which will require context. These metrics by themselves are not useful anyways is my main point.

core/meterer/on_demand_meterer_metrics.go
litt3: It's the same event, but I really don't see what we gain be combining values here. IMO the simple way is to have 2 metrics. Less description, easier to see a value at a glance, easier to make dashboards.

cody-littley: They stem from the same event, but they are not the same thing. In many other places in core code, we also split metrics like this.

core/payments/ondemand/on_demand_cache_metrics.go
litt3: > prometheus has ways to do those

The prometheus library has gauge metrics that accept a getter param, which is what I'm using. I'm not rolling my own solution here, this is just using a feature that's included in the library.

> useful for cases where it is not feasible to instrument a given system with Prometheus metrics directly

This description for exporters doesn't seem to match what I'm doing here. We _are_ instrumenting this with prometheus directly

core/payments/ondemand/on_demand_validator_metrics.go
litt3: > someone needs to arrange a diagram to show them

Right, that's me. I implemented the metrics here that I want in the dashboard that I'll be looking at if there's a production issue with payments.

> lot of metrics for each client errors

It all comes down to specifics, and it's certainly possible to go overboard.

PR #2045, opened 2025-09-22
ci: Add encoder-icicle to docker build pipeline

dmanc:

Adds encoder-icicle to "all" group so that we build it in the `docker-publish-internal` pipeline. This will help catch any issues with the icicle encoder build earlier.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/docker-bake.hcl
+++ b/docker-bake.hcl
@@ -123,6 +123,8 @@ target "encoder" {
 target "encoder-icicle" {
   context    = "."
   dockerfile = "./disperser/cmd/encoder/icicle.Dockerfile"
+  // Currently needed because Dockerfile has amd64 hardcoded in a few places.
+  // TODO: make Dockerfile also work for arm.
   platforms  = ["linux/amd64"]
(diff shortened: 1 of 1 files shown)


docker-bake.hcl
```diff
@@ -122,6 +123,7 @@ target "encoder" {
 target "encoder-icicle" {
   context    = "."
   dockerfile = "./disperser/cmd/encoder/icicle.Dockerfile"
+  platforms  = ["linux/amd64"]
```
samlaf, line 128:
```suggestion
  // Currently needed because Dockerfile has amd64 hardcoded in a few places.
  // TODO: make Dockerfile also work for arm.
  platforms  = ["linux/amd64"]
```

samlaf (commented):
LGTM. https://github.com/Layr-Labs/eigenda/pull/2046 should fix the pipeline issue

samlaf (approved):

PR #2048, opened 2025-09-22
feat(proxy): default V2 storage backend

ethenotethan:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/proxy/config/v2/eigendaflags/cli.go
+++ b/api/proxy/config/v2/eigendaflags/cli.go
@@ -237,6 +237,21 @@ This check is optional and will be skipped when set to 0.`,
 	}
 }
 
+func ReadClientConfigV2Memstore(ctx *cli.Context) (common.ClientConfigV2, error) {
+
+	maxBlobLengthFlagContents := ctx.String(MaxBlobLengthFlagName)
+	maxBlobLengthBytes, err := eigendaflags.ParseMaxBlobLength(maxBlobLengthFlagContents)
+	if err != nil {
+		return common.ClientConfigV2{}, fmt.Errorf(
+			"parse max blob length flag \"%v\": %w", maxBlobLengthFlagContents, err)
+	}
+
+	return common.ClientConfigV2{
+		PutTries:         ctx.Int(PutRetriesFlagName),
+		MaxBlobSizeBytes: maxBlobLengthBytes,
+	}, nil
+}
+
 func ReadClientConfigV2(ctx *cli.Context) (common.ClientConfigV2, error) {
 	disperserConfig, err := readDisperserCfg(ctx)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


litt3 (dismissed):

api/proxy/store/cli.go
```diff
@@ -33,7 +33,7 @@ func CLIFlags(envPrefix, category string) []cli.Flag {
 			Name:     BackendsToEnableFlagName,
 			Usage:    "Comma separated list of eigenDA backends to enable (e.g. V1,V2)",
 			EnvVars:  withEnvPrefix(envPrefix, "BACKENDS_TO_ENABLE"),
-			Value:    cli.NewStringSlice("V1"),
+			Value:    cli.NewStringSlice("V2"),
```
samlaf, line 36: should we enable both by default? To not break v1 users until we actually deprecate

api/proxy/config/v2/eigendaflags/cli.go
```diff
@@ -237,6 +237,21 @@ This check is optional and will be skipped when set to 0.`,
...
+	maxBlobLengthBytes, err := eigendaflags.ParseMaxBlobLength(maxBlobLengthFlagContents)
+	if err != nil {
+		return common.ClientConfigV2{}, fmt.Errorf(
+			"parse max blob length flag \"%v\": %w", maxBlobLengthFlagContents, err)
+	}
+
+	return common.ClientConfigV2{
+		PutTries:         ctx.Int(PutRetriesFlagName),
+		MaxBlobSizeBytes: maxBlobLengthBytes,
+	}, nil
```
samlaf, line 252: so guessing we're reusing these somehow in memstore config? Is there another way to read these variables? Feels super bug-prone to set these here and then re-use them somewhere else. Could we just populate a common.MemstoreConfig or something instead?

api/proxy/store/builder/config.go
```diff
@@ -17,34 +17,41 @@ import (
...
 
-	// secondary storage cfgs
+	// secondary storage cfg
 	S3Config s3.Config
 }
 
 // ReadConfig ... parses the Config from the provided flags or environment variables.
+// client config ingestion is ignored if `memstore=true`
 func ReadConfig(ctx *cli.Context) (Config, error) {
+	useMemStore := ctx.Bool(memstore.EnabledFlagName)
```
samlaf, line 45: you say that "client config ingestion is ignored if memstore=true", but thats not really true because you populate 2 fields, which I'm guessing are read somewhere else to populate some other config or whatnot.

Would prefer if we take a bit more time to actually split those things apart completely. We just just not even have to call client.ReadConfig at all when useMemStore is set to true. Like this should just return an empty config right here

PR #2053, opened 2025-09-23
fix(bloblength): use uint32 instead of uint

samlaf:
We've been using uint/int in a bunch of data types which are fundamental to our protocol, and influence hashing and possibly other things. We shouldn't.

bloblength is once instance where we use uint32 onchain, uint offchain (and actually uint64 in the encoder proto!).

This PR changes the main offchain bloblength type from uint to uint32, and propagates the changes everywhere.

I decided not to change all the v1 chunk encoding/assignment logic which still uses uint, and so just convert uint32->uint when calling those functions. This is fine because golang uint type is always at least uint32.

### DIFF
--- a/disperser/batcher/encoding_streamer_test.go
+++ b/disperser/batcher/encoding_streamer_test.go
@@ -452,7 +452,7 @@ func TestPartialBlob(t *testing.T) {
 	assert.NotNil(t, encodedBlob1.BlobHeader.BlobCommitments)
 	assert.NotNil(t, encodedBlob1.BlobHeader.BlobCommitments.Commitment)
 	assert.NotNil(t, encodedBlob1.BlobHeader.BlobCommitments.LengthProof)
-	assert.Equal(t, encodedBlob1.BlobHeader.BlobCommitments.Length, uint32(48))
+	assert.Equal(t, encodedBlob1.BlobHeader.BlobCommitments.Length, uint32(48)) //nolint: staticcheck
 	assert.Len(t, encodedBlob1.BlobHeader.QuorumInfos, 1)
 	assert.ElementsMatch(t, encodedBlob1.BlobHeader.QuorumInfos, []*core.BlobQuorumInfo{{
 		SecurityParam: core.SecurityParam{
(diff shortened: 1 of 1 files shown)


api/proto/encoder/v2/encoder_v2.proto
```diff
@@ -15,6 +15,9 @@ service Encoder {
 message EncodeBlobRequest {
   bytes blob_key = 1;
   EncodingParams encoding_params = 2;
+  // TODO(samlaf): we should change this to uint32, since blobLengths are uint32 everywhere.
```
litt3, line 18: @cody-littley has expressed reservations in the past about putting TODOs into proto docs like this, since these are very visible. Not sure where this sort of info should be captured, though...

disperser/cmd/apiserver/lib/apiserver.go
```diff
@@ -157,7 +157,7 @@ func RunDisperserServer(ctx *cli.Context) error {
 			meterer,
 			authv2.NewPaymentStateAuthenticator(config.AuthPmtStateRequestMaxPastAge, config.AuthPmtStateRequestMaxFutureAge),
 			prover,
-			uint64(config.MaxNumSymbolsPerBlob),
+			uint32(config.MaxNumSymbolsPerBlob),
```
litt3: unnecessary cast

api/proto/encoder/v2/encoder_v2.proto
samlaf: whats wrong with this being very visible? Also its in the encoder proto, which is internal.

disperser/cmd/apiserver/lib/apiserver.go
samlaf: [050099d](https://github.com/Layr-Labs/eigenda/pull/2053/commits/050099dcb134f375b875d5007b18710a8799c354)

encoding/kzg/prover/v2/gnark/commitments.go
```diff
@@ -16,19 +16,19 @@ type KzgCommitmentsGnarkBackend struct {
...
 }
 
 func (p *KzgCommitmentsGnarkBackend) ComputeLengthProofForLengthV2(
-	coeffs []fr.Element, length uint64,
+	coeffs []fr.Element, length uint32,
 ) (*bn254.G2Affine, error) {
-	if length < uint64(len(coeffs)) {
+	if length < uint32(len(coeffs)) {
 		return nil, fmt.Errorf("length is less than the number of coefficients")
 	}
```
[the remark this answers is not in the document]
samlaf: this is fine. if it panics then theres a config bug anyways so just fix it and restart the binary.

encoding/kzg/prover/gnark/commitments.go
```diff
@@ -16,17 +16,19 @@ type KzgCommitmentsGnarkBackend struct {
...
 }
 
-func (p *KzgCommitmentsGnarkBackend) ComputeLengthProofForLength(coeffs []fr.Element, length uint64) (*bn254.G2Affine, error) {
-	if length < uint64(len(coeffs)) {
+func (p *KzgCommitmentsGnarkBackend) ComputeLengthProofForLength(
+	coeffs []fr.Element, length uint32,
+) (*bn254.G2Affine, error) {
+	if length < uint32(len(coeffs)) {
 		return nil, fmt.Errorf("length is less than the number of coefficients")
 	}
```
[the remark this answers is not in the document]
samlaf: this is fine. if it panics then theres a config bug anyways so just fix it and restart the binary.

litt3 (approved):

api/proto/encoder/v2/encoder_v2.proto
litt3: I suppose so that the docs seem polished.

Given this is internal, I think it seems reasonable to permit such docs, whether or not we avoid them for public facing protos

PR #2055, opened 2025-09-23
fix(payments): Fix on-demand insufficient funds behavior

litt3:
- Old panic behavior was implemented when `ClientLedger` didn't have a way to see new deposits. Now that `ClientLedger` has that ability, it makes sense to stay alive
- Enables on-demand proxy test on sepolia. I'm just gonna deposit a bunch of sepeth for the CI's account, so we get this coverage. No need to try to complicate things with auto-refill, I'll just make a new deposit if it runs out

### DIFF
--- a/core/payments/clientledger/client_ledger_test.go
+++ b/core/payments/clientledger/client_ledger_test.go
@@ -324,7 +324,7 @@ func TestOnDemandOnly(t *testing.T) {
 		require.Equal(t, accountID, paymentMetadata.AccountID)
 	})
 
-	t.Run("insufficient funds panic", func(t *testing.T) {
+	t.Run("insufficient funds returns error", func(t *testing.T) {
 		clientLedger := NewClientLedger(
 			ctx,
 			test.GetLogger(),
@@ -339,9 +339,9 @@ func TestOnDemandOnly(t *testing.T) {
 		)
 		require.NotNil(t, clientLedger)
 
-		require.Panics(t, func() {
-			_, _ = clientLedger.Debit(ctx, 1001, []core.QuorumID{0, 1})
-		}, "insufficient funds should cause fatal panic in on-demand only mode")
+		_, err := clientLedger.Debit(ctx, 1001, []core.QuorumID{0, 1})
+		var insufficientFundsErr *ondemand.InsufficientFundsError
+		require.ErrorAs(t, err, &insufficientFundsErr)
 	})
 
 	t.Run("fatal errors cause panic", func(t *testing.T) {


api/clients/v2/examples/client_construction.go
```diff
@@ -510,15 +509,18 @@ func createOnDemandLedger(
 		return nil, fmt.Errorf("get payment state from disperser: %w", err)
 	}
 
+	var cumulativePayment *big.Int
 	if paymentState.GetCumulativePayment() == nil {
-		return nil, errors.New("received nil cumulative payment from disperser")
+		cumulativePayment = big.NewInt(0)
```
samlaf, line 514: why is it fine to set to 0 when disperser sent up nil?

core/payments/clientledger/client_ledger.go
```diff
@@ -204,8 +204,24 @@ func (cl *ClientLedger) debitOnDemandOnly(
...
 	cumulativePayment, err := cl.onDemandLedger.Debit(ctx, blobLengthSymbols, quorums)
-	enforce.NilError(err, "on-demand debit failed. reservations aren't configured, and the ledger won't become "+
-		"aware of new on-chain deposits without a restart")
+	if err != nil {
+		var insufficientFundsErr *ondemand.InsufficientFundsError
+		if errors.As(err, &insufficientFundsErr) {
+			// Don't panic if insufficient funds occurs: new deposits will be observed by the client ledger, so it's
+			// possible to recover from this.
+			// nolint:wrapcheck // the returned error message is informative
+			return nil, err
```
samlaf, line 213: prob want to add a big log.Warn here though? "Your ondemand balance is 0. please deposit more funds"

api/clients/v2/examples/client_construction.go
litt3: The disperser sends a `nil` if there isn't any record of a cumulative payment found, which is the default state if the user has never submitted an on-demand dispersal before.

I were to reimplement the RPC, I'd have the disperser return a zero in this case- but for now I think it's easiest to just work with the current implementation

core/payments/clientledger/client_ledger.go
litt3: The error returned is logged higher up in the stack:


> Sep 24 09:07:44.410 ERR request completed with error method=POST url="/put?commitment_mode=standard" commitment_mode=standard cert_version=unknown status=500 duration=1.048471875s error="post request failed: debit: insufficient on-demand funds: current cumulative payment: 0 wei, total deposits: 0 wei, blob cost: 1830912000000 wei"

samlaf: This is only if they use proxy though right? I never know the right balance between libraries introducing logs or not. Rust does it nice because each library logs under a namespace which can be shut up easily.

samlaf: That being said, fine to merge with or without the extra log.

litt3: > This is only if they use proxy though right?

Right, but if someone rolls their own integration, it seems reasonable to assume they will be responsible for logging any returned errors, as opposed to silently discarding them.

I think it makes sense for a library to do logging iff a given occurrence is important, and would otherwise not be visible to the library user as an error.

PR #2056, opened 2025-09-23
fix(payments): Properly synchronize vault monitor tests

litt3:
- Old test code was flaky. I added mutexes, so that the concurrent map access is actually safe


litt3:
> Interesting. How did you catch this? Did you run `go test -race`? We should add `go test -race` to our CI somewhere, to catch these upfront.

@samlaf I saw the test flake, and went digging. Interestingly, `-race` _did_ identify the issue. I agree it would make sense to add, because why not

PR #2057, opened 2025-09-23
refactor(prover-v2): cleanup unused fcts

samlaf:
Continuing cleanup of v2 prover/verifier. This time cleaning up prover.
Main reason for this cleanup is that the v1 prover was both generating commitments + proofs (frames).
But in EigenDA V2, commitments are generated either on client side, or on the apiserver (via grpc endpoint).
Hence, it doesn't make sense to have this large prover struct that can do both commitments and frames.
It makes more sense to split them off, which will make documentation, specifically to SRS and FFT config/requirements, a lot easier to understand.

After this merges, I plan to split off kzg commitment generation into its own struct (left a few TODOs related to this in the code).

This PR might look big, but half of it is moving things around, the other half is deleting unused things and fixing the tests. Suggest looking at the commit list and looking at each commit in isolation (they are self-contained and should be very to look at in isolation).

### DIFF
--- a/encoding/kzg/prover/v2/icicle.go
+++ b/encoding/kzg/prover/v2/icicle.go
@@ -22,7 +22,7 @@ const (
 )
 
 func CreateIcicleBackendProver(p *Prover, params encoding.EncodingParams, fs *fft.FFTSettings) (*ParametrizedProver, error) {
-	_, fftPointsT, err := p.SetupFFTPoints(params)
+	_, fftPointsT, err := p.setupFFTPoints(params)
 	if err != nil {
 		return nil, err
 	}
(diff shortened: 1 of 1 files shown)


encoding/kzg/prover/v2/prover.go
```diff
@@ -434,8 +291,72 @@ func (p *Prover) createIcicleBackendProver(
...
+		if err != nil {
+			return nil, fmt.Errorf("parse dimension part of the table: %w", err)
+		}
+		cosetSizeValue, err := strconv.Atoi(tokens[1][5:])
+		if err != nil {
+			return nil, fmt.Errorf("parse coset size part of the table: %w", err)
+		}
+
+		params := encoding.EncodingParams{
+			NumChunks:   uint64(cosetSizeValue),
```
dmanc: Actually I noticed today that this is wrong. NumChunks should be dimE and chunkLength should be cosetSize, even the comment above says that. I don't think this code path has ever been used...

dmanc (approved):
:shipit:

encoding/kzg/prover/v2/parametrized_prover.go
```diff
@@ -14,14 +14,19 @@ import (
...
-	*rs.Encoder
+	srsNumberToLoad uint64
 
-	KzgConfig *KzgConfig
+	encodingParams encoding.EncodingParams
+	encoder        *rs.Encoder
 
-	KzgMultiProofBackend  KzgMultiProofsBackendV2
-	KzgCommitmentsBackend KzgCommitmentsBackendV2
+	computeMultiprooNumWorker uint64
```
litt3:
```suggestion
	computeMultiproofNumWorker uint64
```

litt3 (approved):

encoding/kzg/prover/v2/parametrized_prover.go
samlaf: [967e8b0](https://github.com/Layr-Labs/eigenda/pull/2057/commits/967e8b073b2ad3c9a57432adca74509e321e70d0)

litt3 (approved):

encoding/kzg/prover/v2/prover.go
samlaf: [f845b1c](https://github.com/Layr-Labs/eigenda/pull/2057/commits/f845b1c7c110eaed5f09864786423f1433072a61)

PR #2058, opened 2025-09-24
ci: Add churner deployment to testbed package and fix inabox manual deployment

dmanc:

This PR adds the churner to the testbed package. Also removes inabox dependency from `churner_test.go`.

In addition this PR fixes issues with the inabox devnet deployment so it's runnable and attempts to make the README.md clearer. I wouldn't spend time reviewing it since the deployment process will be refactored after every component is migrated to containers.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


inabox/bin.sh
```diff
@@ -19,13 +19,17 @@ function kill_processes {
 function start_trap {
     trap kill_processes SIGINT
 
-    set -a
-    source $testpath/envs/churner.env
-    set +a
-    ../operators/churner/bin/server &
+    if [ "$SKIP_CHURNER" != "true" ]; then
```
samlaf: So the end goal is that bin.sh is still used for local devnet inabox, but for e2e tests all processes are started as goroutines right?

If this is the case, I'd prefer we just create a new temporary function in bin.sh that starts everything that is not deployed by golang. That makes it easier to cleanup and delete when we don't need it. Also really dislike using env_vars like this.

inabox/deploy/deploy.go
```diff
@@ -371,7 +371,8 @@ func (env *Config) StartBinaries() {
 	}
 
 	logger.Info("Starting binaries")
-	err := execCmd("./bin.sh", []string{"start-detached"}, []string{}, true)
+	// Skip churner since tests will start it as a goroutine using testbed
+	err := execCmd("./bin.sh", []string{"start-detached"}, []string{"SKIP_CHURNER=true"}, true)
```
samlaf: think instead we should have a flag in `StartBinaries` that will control whether these are started for tests or devnet, and when starting for tests we can call the new fct `start-detached-for-tests` or something

test/testbed/deploy_eigenda_churner_goroutine.go
```diff
@@ -0,0 +1,316 @@
...
+// ChurnerConfig defines configuration for the churner container
+//
+//nolint:lll // struct field documentation
+type ChurnerConfig struct {
+	// Enable churner service
+	Enabled bool `env:"-"` // Skip in env mapping
+
+	// Log configuration
+	LogFormat string `env:"CHURNER_LOG_FORMAT"`
+	LogLevel  string `env:"CHURNER_LOG_LEVEL"`
```
samlaf: doesn't the churner have its own config? Don't think we should be using the .env config to spin up these goroutines, we want to move everything to golang, including config generation

test/testbed/deploy_eigenda_churner_goroutine.go
```diff
@@ -0,0 +1,316 @@
...
+	MockIndexer interface{} `env:"-"` // Will be cast to thegraph.IndexedChainState
+}
+
+type ChurnerGoroutine struct {
+	server   *grpc.Server
+	listener net.Listener
+	url      string
+	cancel   context.CancelFunc
+	config   ChurnerConfig
+}
```
samlaf: im confused... doesnt churner have a struct like this already?

inabox/tests/integration_suite_test.go
```diff
@@ -235,6 +237,14 @@ func setupSuite() error {
 			return fmt.Errorf("failed to mine block: %w", err)
 		}
 
+		logger.Info("Starting churner goroutine")
+		churnerConfig := setupChurnerConfig(testConfig)
+		churnerGoroutine, err = testbed.StartChurnerGoroutine(churnerConfig, logger)
```
samlaf: why do we need testbed for churner? can't we just use the churner code itself?

operators/churner/tests/churner_test.go
```diff
@@ -96,14 +131,57 @@ func setupTest(t *testing.T) (*testbed.AnvilContainer, *testbed.LocalStackContai
...
 func TestChurner(t *testing.T) {
 	ctx := t.Context()
-	_, _, testConfig := setupTest(t)
+	testSetup := setupTest(t)
+
+	// Create mock indexer
+	mockIndexer := &indexermock.MockIndexedChainState{}
+
+	// Configure and start churner using testbed pattern with mock indexer
+	churnerConfig := testbed.ChurnerConfig{
```
samlaf: why not use the actual churner config?

inabox/bin.sh
dmanc: 636b39d0bee91bfb24472ee9cd04d933b3e897de

inabox/deploy/deploy.go
dmanc: 636b39d0bee91bfb24472ee9cd04d933b3e897de

operators/churner/tests/churner_test.go
dmanc: 81a6d724e9dd96ad072728450fab8f4b18b27b40

inabox/tests/integration_suite_test.go
dmanc: 81a6d724e9dd96ad072728450fab8f4b18b27b40

test/testbed/deploy_eigenda_churner_goroutine.go
dmanc: 81a6d724e9dd96ad072728450fab8f4b18b27b40

test/testbed/deploy_eigenda_churner_goroutine.go
dmanc: 81a6d724e9dd96ad072728450fab8f4b18b27b40

inabox/deploy/cmd/main.go
```diff
@@ -158,7 +158,6 @@ func chainInfra(ctx *cli.Context, config *deploy.Config) error {
 		ExposeHostPort: true,
 		HostPort:       "8545",
 		Logger:         logger,
-		Network:        dockerNetwork,
```
samlaf, line 161: why remove this?

inabox/tests/integration_suite_test.go
```diff
@@ -518,6 +536,126 @@ func setupRetrievalClients(testConfig *deploy.Config) error {
...
+		EigenDADirectory:           testConfig.EigenDA.EigenDADirectory,
+		ChurnApprovalInterval:      15 * time.Minute,
+		PerPublicKeyRateLimit:      1 * time.Second,
+	}
+
+	// Set graph URL if graph node is enabled
+	if deployer.DeploySubgraphs && graphNodeContainer != nil {
+		churnerConfig.ChainStateConfig = thegraph.Config{
+			Endpoint: "http://localhost:8000/subgraphs/name/Layr-Labs/eigenda-operator-state",
+		}
```
samlaf, line 589: why isn't this url part of testconfig?

inabox/tests/integration_suite_test.go
```diff
@@ -518,6 +536,126 @@ func setupRetrievalClients(testConfig *deploy.Config) error {
 	return err
 }
 
+func startChurnerServer(testConfig *deploy.Config) error {
+	// Start listener on the specified port
+	grpcPort := "32002"
```
samlaf: this should prob be part of churner config?

inabox/tests/integration_suite_test.go
```diff
@@ -518,6 +536,126 @@ func setupRetrievalClients(testConfig *deploy.Config) error {
...
+	listener, err := net.Listen("tcp", fmt.Sprintf(":%s", grpcPort))
+	if err != nil {
+		return fmt.Errorf("failed to listen on port %s: %w", grpcPort, err)
+	}
+	churnerListener = listener
```
samlaf: I know churner also does this, but imo this listener should be constructed at as the very last thing. No point in binding a port if we can't construct the churner struct.

inabox/tests/integration_suite_test.go
```diff
@@ -518,6 +536,126 @@ func setupRetrievalClients(testConfig *deploy.Config) error {
 	return err
 }
 
+func startChurnerServer(testConfig *deploy.Config) error {
```
samlaf, line 539: The high-level structure I would like to see is:
```
churnerConfig := churnerConfigFromTestConfig(testConfig)
churner := churner.New(churnerConfig)
churner.serve()
```

This would require some refactors to the churner object itself, so I'm fine with this being done in another PR. But having too many separate places where stuff are built is error-prone and just less maintainable.

inabox/tests/integration_suite_test.go
```diff
@@ -518,6 +536,126 @@ func setupRetrievalClients(testConfig *deploy.Config) error {
...
+		gethClient,
+		churnerConfig.OperatorStateRetrieverAddr,
+		churnerConfig.EigenDAServiceManagerAddr)
+	if err != nil {
+		return fmt.Errorf("failed to create writer: %w", err)
+	}
+
+	// Create indexer
+	cs := coreeth.NewChainState(churnerTx, gethClient)
+	indexer := thegraph.MakeIndexedChainState(churnerConfig.ChainStateConfig, cs, churnerLogger)
```
samlaf: personally prefer no 1/2 letter vars. makes parsing code harder when you're not right next to the place where that var was defined, such as ChainStateConfig(g,cs,s,pr)... like what are those 4 things?

inabox/tests/integration_suite_test.go
```diff
@@ -518,6 +536,126 @@ func setupRetrievalClients(testConfig *deploy.Config) error {
...
+	// Start serving in goroutine
+	go func() {
+		churnerLogger.Info("Starting churner gRPC server", "port", grpcPort)
+		if err := churnerServer.Serve(churnerListener); err != nil {
+			churnerLogger.Info("Churner gRPC server stopped", "error", err)
+		}
+	}()
+
+	// Give server time to start
+	time.Sleep(100 * time.Millisecond)
```
samlaf, line 653: is this really needed? would personally prefer a TODO comment saying to create a proper health check

inabox/tests/integration_suite_test.go
```diff
@@ -52,6 +61,8 @@ TODO: Put these into a testSuite object which is initialized per inabox E2E test
 var (
 	anvilContainer     *testbed.AnvilContainer
 	graphNodeContainer *testbed.GraphNodeContainer
+	churnerServer      *grpc.Server
+	churnerListener    net.Listener
```
samlaf, line 65: might as well start now and stop using global variables. Let's just define a testHarness with all the binaries defined in there. And after starting we return that struct, which for now would only contain the churner. I really hate global variables.

operators/churner/tests/churner_test.go
```diff
@@ -50,26 +53,26 @@ var (
...
 
-func setupTest(t *testing.T) (*testbed.AnvilContainer, *testbed.LocalStackContainer, *deploy.Config) {
+// TestSetup contains all the test infrastructure needed for churner tests
+type TestSetup struct {
+	AnvilContainer      *testbed.AnvilContainer
+	LocalstackContainer *testbed.LocalStackContainer
+	Contracts           *testbed.DeploymentResult
+	Operators           []operatorInfo
+	PrivateKeys         *testbed.PrivateKeyMaps
+}
```
samlaf: nice..!
nit: I prefer to call these testHarness, for consistency with rest of codebase.

operators/churner/tests/churner_test.go
```diff
@@ -78,15 +81,52 @@ func setupTest(t *testing.T) (*testbed.AnvilContainer, *testbed.LocalStackContai
...
+	operators := make([]operatorInfo, numOperators)
+	for i := 0; i < numOperators; i++ {
+		operatorKey := fmt.Sprintf("opr%d", i)
+		operators[i] = operatorInfo{
+			ECDSAPrivateKey: privateKeys.EcdsaMap[operatorKey].PrivateKey,
+			ECDSAKeyFile:    privateKeys.EcdsaMap[operatorKey].KeyFile,
+			ECDSAPassword:   privateKeys.EcdsaMap[operatorKey].Password,
+			BLSKeyPath:      privateKeys.BlsMap[operatorKey].KeyFile,
+			BLSPassword:     privateKeys.BlsMap[operatorKey].Password,
+		}
```
samlaf: why doesn't testbed.LoadPrivateKeys directly give us back OperatorInfo structs like this?
like shouldnt the testbed have a higher level function like testbed.GenerateOperators()?

operators/churner/tests/churner_test.go
```diff
@@ -96,30 +136,126 @@ func setupTest(t *testing.T) (*testbed.AnvilContainer, *testbed.LocalStackContai
...
+
+	// Start churner server directly using the churner package
+	grpcPort := "32002"
+	listener, err := net.Listen("tcp", fmt.Sprintf(":%s", grpcPort))
+	require.NoError(t, err, "failed to listen on port")
+	defer func() {
+		if err := listener.Close(); err != nil {
+			t.Logf("failed to close listener: %v", err)
+		}
+	}()
```
samlaf: yet another place where we replicate all of this logic. This should all be abstracted in a churner struct with a Serve() method.

operators/churner/tests/churner_test.go
```diff
@@ -96,30 +136,126 @@ func setupTest(t *testing.T) (*testbed.AnvilContainer, *testbed.LocalStackContai
...
+	}
 }
 
 func TestChurner(t *testing.T) {
 	ctx := t.Context()
-	_, _, testConfig := setupTest(t)
+	testSetup := setupTest(t)
+
+	// Create mock indexer
+	mockIndexer := &indexermock.MockIndexedChainState{}
```
samlaf, line 134: not related to this PR but do you know why we use a mock indexer for these tests instead of using an actual indexer? we have a graph node spun up.. couldnt we use a real indexedChainState instance?

samlaf: Like what's the point of spinning up anvil and thegraph if we use a mock indexed chain state?

dmanc: We don't spin up the graph for this particular test

inabox/deploy/cmd/main.go
dmanc: a090e5e5e54be710ade3dedb822c2d81834bd0c4

dmanc: bad rebase

inabox/tests/integration_suite_test.go
dmanc: a090e5e5e54be710ade3dedb822c2d81834bd0c4

inabox/tests/integration_suite_test.go
dmanc: a090e5e5e54be710ade3dedb822c2d81834bd0c4

inabox/tests/integration_suite_test.go
dmanc: a090e5e5e54be710ade3dedb822c2d81834bd0c4

inabox/tests/integration_suite_test.go
dmanc: a090e5e5e54be710ade3dedb822c2d81834bd0c4

operators/churner/tests/churner_test.go
dmanc: a090e5e5e54be710ade3dedb822c2d81834bd0c4

operators/churner/tests/churner_test.go
dmanc: Seemed reasonable a090e5e5e54be710ade3dedb822c2d81834bd0c4

operators/churner/tests/churner_test.go
dmanc: Don't feel like doing it in this PR

inabox/tests/integration_suite_test.go
dmanc: Agree that seems good, will probably save it for another PR

inabox/tests/integration_suite_test.go
dmanc: I'll do it in a followup PR and I can remove most globals that way

inabox/tests/integration_suite_test.go
dmanc: Will do it in followup where i introduce a test harness to inabox

operators/churner/tests/churner_test.go
samlaf: oh interesting.

PR #2059, opened 2025-09-24
feat(srs): Add srs table download utility

dmanc:

Adds the ability to download the required SRSTables for EigenDA V2 encoding (proof generation). Also adds some documentation around them.

Also added the tools directory to the code coverage ignore list.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/payments/ondemand/test/on_demand_vault_monitor_test.go
+++ b/core/payments/ondemand/test/on_demand_vault_monitor_test.go
@@ -101,26 +101,15 @@ func TestOnDemandVaultMonitor(t *testing.T) {
 	testAccount := accounts[2]
 	testVault.SetTotalDeposit(testAccount, big.NewInt(9999)) // Changed
 
-	// Clear captured updates to verify new updates
-	mu.Lock()
-	capturedUpdates = make(map[gethcommon.Address]*big.Int)
-	mu.Unlock()
-
-	// Wait for the monitor to fetch the updated deposits
-	test.AssertEventuallyEquals(t, len(accounts), func() int {
+	// Wait for the monitor to fetch the updated deposit
+	test.AssertEventuallyEquals(t, big.NewInt(9999), func() *big.Int {
 		mu.Lock()
 		defer mu.Unlock()
-		return len(capturedUpdates)
+		return capturedUpdates[testAccount]
 	}, time.Second)
 
-	// Check that the specific account was updated correctly
-	mu.Lock()
-	updatedDeposit, ok := capturedUpdates[testAccount]
-	require.True(t, ok, "account %s should have been updated", testAccount.Hex())
-	require.NotNil(t, updatedDeposit)
-	require.Equal(t, big.NewInt(9999), updatedDeposit)
-
 	// Other accounts should remain unchanged
+	mu.Lock()
 	for i, addr := range accounts {
 		if addr != testAccount {
 			deposit, ok := capturedUpdates[addr]
--- a/core/payments/reservation/reservation_vault_monitor_test.go
+++ b/core/payments/reservation/reservation_vault_monitor_test.go
@@ -113,26 +113,15 @@ func TestReservationVaultMonitor(t *testing.T) {
 		QuorumSplits:     []byte{100},
 	})
 
-	// Clear captured updates to verify new updates
-	mu.Lock()
-	capturedUpdates = make(map[gethcommon.Address]*Reservation)
-	mu.Unlock()
-
 	// Wait for the monitor to fetch the updated reservation
-	test.AssertEventuallyEquals(t, len(accounts), func() int {
+	test.AssertEventuallyEquals(t, uint64(999), func() uint64 {
 		mu.Lock()
 		defer mu.Unlock()
-		return len(capturedUpdates)
+		return capturedUpdates[testAccount].symbolsPerSecond
 	}, time.Second)
 
-	// Check that the specific account was updated correctly
-	mu.Lock()
-	updatedReservation, ok := capturedUpdates[testAccount]
-	require.True(t, ok, "account %s should have been updated", testAccount.Hex())
-	require.NotNil(t, updatedReservation)
-	require.Equal(t, uint64(999), updatedReservation.symbolsPerSecond)
-
 	// Other accounts should remain unchanged
+	mu.Lock()
 	for i, addr := range accounts {
 		if addr != testAccount {
 			reservation, ok := capturedUpdates[addr]
--- a/core/payments/vault/test_payment_vault.go
+++ b/core/payments/vault/test_payment_vault.go
@@ -3,6 +3,7 @@ package vault
 import (
 	"context"
 	"math/big"
+	"sync"
 
 	bindings "github.com/Layr-Labs/eigenda/contracts/bindings/v2/PaymentVault"
 	"github.com/Layr-Labs/eigenda/core/payments"
@@ -11,6 +12,8 @@ import (
 
 // TestPaymentVault is a test implementation of the PaymentVault interface
 type TestPaymentVault struct {
+	mu sync.Mutex
+
 	// Storage for individual account deposits
 	totalDeposits map[gethcommon.Address]*big.Int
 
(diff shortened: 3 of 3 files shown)


tools/srs-utils/README.md
```diff
@@ -27,11 +28,54 @@ Note that the G2 point files (`g2.point` and `g2.trailing.point`) are twice the
 points require twice as many bytes to represent as G1 points in the BN254 curve. Each G1 point requires 32 bytes 
 of storage, while each G2 point requires 64 bytes.
 
-The `g2.point.powerOf2` file contains only G2 points at power-of-2 indices (indices 1, 2, 4, 8, 16, ..., 2^27). This 
-optimized file contains just 28 G2 points instead of the full set, significantly reducing memory usage for operator 
-nodes. Since operators only perform multi-reveal proofs on blobs with power-of-2 polynomial degrees, they don't need 
+The `g2.point.powerOf2` file contains only G2 points at power-of-2 indices (indices 1, 2, 4, 8, 16, ..., 2^27). This
+optimized file contains just 28 G2 points instead of the full set, significantly reducing memory usage for operator
+nodes. Since operators only perform multi-reveal proofs on blobs with power-of-2 polynomial degrees, they don't need
```
samlaf, line 33: is this true actually? what about v1? afaict v1 doesn't enforce lengths to be powers of 2. Does v1 not use multi-reveal proofs?

tools/srs-utils/README.md
```diff
@@ -27,11 +28,54 @@ Note that the G2 point files (`g2.point` and `g2.trailing.point`) are twice the
...
+
+```
+Blob Size = (numChunks × cosetSize × 32 bytes) / codingRatio
+```
+
+Where:
+- `numChunks` = 8192 (for blob version 0)
+- `cosetSize` = chunk length (varies based on blob size)
+- `32 bytes` = size of each BN254 field element
+- `codingRatio` = 8 (fixed erasure coding expansion factor)
```
samlaf: its not fixed its part of the blobEncodingParams. version 0 uses 8, but it could possibly change in the future. @Lulu-Zhou-EigenLabs and @mooselumph did mention that it has to be a power of 2 for some reason, and we would never go to 16, so the only likely change would be to 4.

tools/srs-utils/README.md
```diff
@@ -27,11 +28,54 @@ Note that the G2 point files (`g2.point` and `g2.trailing.point`) are twice the
...
+
+The supported blob size depends on the coset size (chunk length) used:
+
+```
+Blob Size = (numChunks × cosetSize × 32 bytes) / codingRatio
+```
+
+Where:
+- `numChunks` = 8192 (for blob version 0)
+- `cosetSize` = chunk length (varies based on blob size)
```
samlaf, line 64: how is chunk length related to cosetSize? Actually can someone explain what cosetSize even means here.

Also we should add a section to explain what's even stored in these SRSTable files. I'm assuming each of them contains points that form a (cyclic?) subgroup. And these are used for... FFT only? So each SRSTable can be generated by taking the first point in it and adding it to itself?

Also would the SRSTable for dimE8192.coset4 == that for dimE4096.coset8 for example?

tools/srs-utils/README.md
```diff
@@ -70,6 +114,35 @@ To download with the power-of-2 points file:
...
+srs-utils download-tables
+```
+
+This will download all coset tables for the default dimension (dimE8192). The files will be saved to
+`resources/srs/SRSTables` directory by default.
+
+Options:
+- `--dimension`: The dimension to download (default: "dimE8192")
+- `--output-dir`: Directory where the tables will be saved (default: "resources/srs/SRSTables")
+- `--base-url`: Base URL for downloading (default: "https://srs-mainnet.s3.amazonaws.com/kzg/SRSTables")
```
samlaf, line 130: were these always stored? Or did you just upload them? How long does it take to generate these files for dimE8192?

tools/srs-utils/README.md
```diff
@@ -27,11 +28,54 @@ Note that the G2 point files (`g2.point` and `g2.trailing.point`) are twice the
...
+In EigenDA V2, **blob version 0** specifically sets `numChunks=8192`, which is why the dimE8192 tables are the
+primary SRS tables used in production.
+
+#### Available Table Files
+
+The SRS tables are organized by dimension (numChunks) and coset size:
+
+| Dimension | Coset Sizes | Total Size | Description |
+|-----------|-------------|------------|-------------|
+| dimE8192  | 4, 8, 16, 32, 64, 128, 256, 512, 1024 | ~1 GB | Tables for numChunks=8192 (blob version 0) |
```
samlaf: where does E come from?

tools/srs-utils/table_downloader/flags.go
```diff
@@ -0,0 +1,62 @@
...
+// ReadTablesConfig reads command line flags into a config struct
+func ReadTablesConfig(cCtx *cli.Context) (TablesDownloaderConfig, error) {
+	cosetSizesStr := cCtx.String(flagCosetSizes)
+	var cosetSizes []int
+	if cosetSizesStr != "" {
+		parts := strings.Split(cosetSizesStr, ",")
+		for _, part := range parts {
+			part = strings.TrimSpace(part)
+			var size int
+			if _, err := fmt.Sscanf(part, "%d", &size); err == nil {
```
samlaf, line 50: this is some crazy LLM gymnastics. Just use a `cli.IntFlag` instead.

tools/srs-utils/downloader/tables_flags.go
```diff
@@ -0,0 +1,62 @@
...
+	},
+	cli.StringFlag{
+		Name:  flagCosetSizes,
+		Usage: "Comma-separated list of coset sizes to download (e.g., 4,8,16,32,64,128,256,512,1024)",
+		Value: "4,8,16,32,64,128,256,512,1024",
+	},
+}
+
+// ReadTablesConfig reads command line flags into a config struct
+func ReadTablesConfig(cCtx *cli.Context) (TablesDownloaderConfig, error) {
```
samlaf: nit: use same function name that we use everywhere in the codebase (forget the exact fct but look up any CLI.go or flags.go file)

samlaf (commented):
This is pretty nice! Have a bunch of questions, but most important ones are:
1. how long does it take to generate these files?
2. who is the target user for this? Assuming its only needed for our disperser? What were we using up to today? Were they just generated on startup?
3. Might be good to flag somewhere (readme at least) that these SRSTables are not needed for operators.

tools/srs-utils/README.md
dmanc: @bxue-l2 What does E mean

tools/srs-utils/README.md
dmanc: No these have always been in the S3 bucket. Let me try generating them manually and get back to you on timing.

Also is it worth adding a generate subcommand or nah?

tools/srs-utils/README.md
dmanc: @bxue-l2

tools/srs-utils/table_downloader/flags.go
dmanc: Discussed on slack to keep this since cli.IntSliceFlag is a worse experience

tools/srs-utils/README.md
dmanc: Trying to generate with SRS_LOAD=524288 caused a crash

2025/09/24 20:10:51 Table with params: DimE=8192 CosetSize=128 does not exist
2025/09/24 20:10:51 Failed to get KZG encoder: new prover: failed to get sub tables: cannot precompute table: insufficient SRS points loaded (have 524288, need at least 1048576). Consider increasing loaded SRS points or using precomputed tables

dmanc: Anyways it takes a long time to generate them, the previous param (8192,64) already took 20s

tools/srs-utils/README.md
dmanc: resolved

tools/srs-utils/README.md
dmanc: extension

tools/srs-utils/README.md
dmanc: - SRSTables are all g1 points
- Subsection of SRS and take ECNTT

samlaf (dismissed):

samlaf (approved):

PR #2063, opened 2025-09-24
feat(payments): Implement validator payment validation

litt3:
- Uses payment utilities on validator to check payments
- Payment validation disabled by default
- closes EGDA-2008

### DIFF
--- a/node/flags/flags.go
+++ b/node/flags/flags.go
@@ -569,6 +569,33 @@ var (
 		Required: false,
 		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "IGNORE_VERSION_FOR_EJECTION_DEFENSE"),
 	}
+	EnablePaymentValidationFlag = cli.BoolFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "enable-payment-validation"),
+		Usage:    "Whether the validator should perform payment validation. Temporary flag that will be removed once the new payments system is fully in place.",
+		Required: false,
+		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "ENABLE_PAYMENT_VALIDATION"),
+	}
+	ReservationMaxLedgersFlag = cli.IntFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "reservation-max-ledgers"),
+		Usage:    "Initial size for the reservation ledger LRU cache. This increases dynamically if premature evictions are detected.",
+		Required: false,
+		Value:    1024,
+		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "RESERVATION_MAX_LEDGERS"),
+	}
+	ReservationBucketCapacityPeriodFlag = cli.DurationFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "reservation-bucket-capacity-period"),
+		Usage:    "Duration used to calculate bucket capacity when creating new reservation ledgers.",
+		Required: false,
+		Value:    180 * time.Second,
+		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "RESERVATION_BUCKET_CAPACITY_PERIOD"),
+	}
+	PaymentVaultUpdateIntervalFlag = cli.DurationFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "payment-vault-update-interval"),
+		Usage:    "Interval for checking for payment vault updates.",
+		Required: false,
+		Value:    30 * time.Second,
+		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "PAYMENT_VAULT_UPDATE_INTERVAL"),
+	}
 
 	/////////////////////////////////////////////////////////////////////////////
 	// TEST FLAGS SECTION
@@ -715,6 +742,10 @@ var optionalFlags = []cli.Flag{
 	EjectionSentinelPeriodFlag,
 	EjectionDefenseEnabledFlag,
 	IgnoreVersionForEjectionDefenseFlag,
+	EnablePaymentValidationFlag,
+	ReservationMaxLedgersFlag,
+	ReservationBucketCapacityPeriodFlag,
+	PaymentVaultUpdateIntervalFlag,
 }
 
 func init() {
--- a/node/grpc/server_v2.go
+++ b/node/grpc/server_v2.go
@@ -146,6 +146,7 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
 			return nil, api.NewErrorInvalidArg(fmt.Sprintf("failed to verify request: %v", err))
 		}
 	}
+	// TODO(litt3): why would we permit the blob authenticator to be nil?
 	if s.blobAuthenticator != nil {
 		// TODO: check the latency of request validation later; could be parallelized to avoid significant
 		// impact to the request latency
@@ -156,6 +157,13 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
 			}
 		}
 	}
+
+	// Validate reservation payments (on-demand payments are validated on the controller)
+	err = s.node.ValidateReservationPayment(ctx, batch, probe)
+	if err != nil {
+		return nil, fmt.Errorf("validate reservation payment: %w", err)
+	}
+
 	probe.SetStage("get_operator_state")
 	s.logger.Info("new StoreChunks request",
 		"batchHeaderHash", hex.EncodeToString(batchHeaderHash[:]),
--- a/node/node.go
+++ b/node/node.go
@@ -39,6 +39,9 @@ import (
 	"github.com/Layr-Labs/eigenda/core"
 	"github.com/Layr-Labs/eigenda/core/eth"
 	"github.com/Layr-Labs/eigenda/core/indexer"
+	"github.com/Layr-Labs/eigenda/core/payments/reservation"
+	"github.com/Layr-Labs/eigenda/core/payments/reservation/reservationvalidation"
+	"github.com/Layr-Labs/eigenda/core/payments/vault"
 	corev2 "github.com/Layr-Labs/eigenda/core/v2"
 
 	"github.com/Layr-Labs/eigensdk-go/logging"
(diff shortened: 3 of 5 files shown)


node/flags/flags.go
```diff
@@ -569,6 +569,34 @@ var (
...
+		Required: false,
+		Value:    1024,
+		EnvVar:   common.PrefixEnvVar(EnvVarPrefix, "RESERVATION_MAX_LEDGERS"),
+	}
+	ReservationBucketCapacityPeriodFlag = cli.DurationFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "reservation-bucket-capacity-period"),
+		Usage:    "Duration used to calculate bucket capacity when creating new reservation ledgers.",
+		Required: false,
+		// TODO(litt3): we need to decide whether this is the default value we actually want to ship
+		Value:  180 * time.Second,
```
samlaf, line 590: scary... this should really be on-chain right?
what happens if we have some validators that update to this version and then stick to it forever, while we update this value to some other value?

node/grpc/server_v2.go
```diff
@@ -146,6 +146,7 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
 			return nil, api.NewErrorInvalidArg(fmt.Sprintf("failed to verify request: %v", err))
 		}
 	}
+	// TODO(litt3): why would we permit the blob authenticator to be nil?
 	if s.blobAuthenticator != nil {
```
samlaf, line 150: lol wut? yeah why?

samlaf: oh my guess is that this is just "being nice" to the clients. If their blobCert isn't valid it would fail somewhere else, but this is catching stuff early. I also think this should always be turned on.

samlaf: oh wait this is on the node not the disperser... wtf. This absolutely always needs to be validated then? Not sure what that validateDispersalRequest does though

node/node.go
```diff
@@ -290,27 +296,68 @@ func NewNode(
...
+			return nil, fmt.Errorf("create payment vault: %w", err)
+		}
+
+		reservationConfig, err := reservationvalidation.NewReservationLedgerCacheConfig(
+			config.ReservationMaxLedgers,
+			config.ReservationBucketCapacityPeriod,
+			// this is hardcoded: it's a parameter just in case, but it's never expected to change
+			reservation.OverfillOncePermitted,
+			config.PaymentVaultUpdateInterval,
+		)
```
samlaf: curious organization wise, why don't you instead embed the CacheConfig as part of the node config? Instead of copy-pasting the fields and then calling this constructor here?

node/grpc/server_v2.go
```diff
@@ -156,6 +157,13 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
 			}
 		}
 	}
+
+	// Validate reservation payments (on-demand payments are validated on the controller)
```
samlaf:
```suggestion
	// Validate reservation payments (on-demand payments are validated by the disperser only)
```

samlaf: think this is clearer? I was a bit confused, thinking the nodes were making a grpc call to the controller. Also most people won't know what the controller is or where it lives. Maybe if you want to mention controller for internal devs then at least mention "the disperser's controller service" or something?

node/flags/flags.go
litt3: Should be on-chain, if we had the mechanism in place to synchronize updates to the value.

> what happens if we have some validators that update to this version

Unless they decide to manually specify this initial default, to override whatever future default we ship, it wouldn't be a problem.

But yeah, I don't like this. The alternative would be to literally hardcode it somewhere, so it's not even something we have to worry about operators messing with

node/grpc/server_v2.go
litt3: I'm not sure if any of the checks in `validateDispersalRequest` are security critical... But yeah, I'm pretty sure there's absolutely no reason this should be disablable. cc @anupsv

node/flags/flags.go
cody-littley: IMO, this is something that would work really nicely to put into the new config directory. Perhaps as a follow up PR after the config directory merges, but before we ship the next release?

I think the criteria for putting something in the config directory should be something like this:
- value should be synchronized across entire validator fleet (i.e. validators should not set it individually)
- value is something we plausibly may want to change in the future
- we are ok with the time lag required to get governance to agree on the change

If it's not in the config directory, I don't think it really makes that big of a difference if we hard code vs. make it a setting. It's unlikely that a large number of validators are going to tamper with this setting, and anybody who really wanted to do so could trivially change code and compile.

cody-littley: I suggest a default value of 60 seconds. We allow 50 seconds for validators to return signatures. If our end-to-end latency ever grows beyond 60 seconds, we are in a SEVERE degraded state.

node/grpc/server_v2.go
cody-littley: Agree. If I had to guess, there's probably a test somewhere that wants to disable it. But better to drop tests than to allow a bypass like this in production code.

node/node.go
cody-littley: Perhaps a good topic of discussion as a small group as we consider how to make our configuration less cruddy.

Using nested config with the urfave library is legitimately terrible. Nest the same config in two places, and you have to manually define CLI flags for each place. It makes things a nightmare to maintain.

Nested configs may get better if we ditch urfave though.

cody-littley (approved):

node/node.go
litt3: No good reason

1b826f69805c067620aaedb86de7e7623235a89a

node/flags/flags.go
litt3: > I suggest a default value of 60 seconds

Note that the requirement here is that `validatorBucketDuration - clientBucketDuration >= maxLatency`.

- Client duration is currently configured to 1 minute.
- I chose 3 minutes here, so that we can tolerate up to 2 minutes client->validator latency
- I had intended to set the disperser bucket duration to 2 minutes, to tolerate up to 1 minute latency on either side of the disperser.

What would you recommend setting for the three durations we need to configure?

litt3: > this is something that would work really nicely to put into the new config directory. Perhaps as a follow up PR after the config directory merges, but before we ship the next release?

The problem is that this field needs a timestamp associated with it, to coordinate updates to the value. We could theoretically _increase_ the bucket size without the timestamp, but wouldn't be able to _decrease_ it. 

The question is, do we want to configure this in the directory before we can do so with an associated timestamp, and then switch to a checkpointed field when the feature becomes available?

node/grpc/server_v2.go
litt3: > most people won't know what the controller is or where it lives

I think most people reading inline code docs would know what this is. But I agree your suggestion is clearer, done

55ba65890f0fb2f0b410b4a5bb31965374e40f6a

node/node.go
litt3: > Perhaps a good topic of discussion as a small group as we consider how to make our configuration less cruddy.

💯 

I implemented the suggestion before I saw your comment, @cody-littley. But in this case, I don't detect any additional pain due to the fact that we're using urfave. Could you clarify why you'd have to define multiple sets of CLI flags?

node/flags/flags.go
litt3: We discussed this offline. I will merge this as-is for now, with the understanding that we cannot release the feature prior to having on-chain checkpointed configs.

node/node.go
cody-littley: not a blocker, ok as is

cody-littley (dismissed):

node/node.go
samlaf: agree that we should discuss config improvements across the board. Im specifically interested in proxy, but of course we should have good principles everywhere. will resolve this thread though, as im happy with this change here.

node/grpc/server_v2.go
anupsv: @litt3 if i remember correctly it was something to do with V1 and V2 compatibility but i could be wrong. But agree, no reason the validation should be nil for anything moving forward.

node/grpc/server_v2.go
```diff
@@ -156,6 +157,13 @@ func (s *ServerV2) StoreChunks(ctx context.Context, in *pb.StoreChunksRequest) (
 			}
 		}
 	}
+
+	// Validate reservation payments (on-demand payments are validated on the the disperser's controller service)
+	err = s.node.ValidateReservationPayment(ctx, batch, probe)
```
samlaf, line 175: as we discussed offline this morning, I have a bunch of questions about how this plays with our general dispersal retry strategy.

We debit here but there's a million different errors that can happen below. Afaiu @cody-littley said we don't currently have retries (at all?) on disperser, so this should be fine. But what if we ever change the disperser strategy and start retrying for certain classes of errors? Would you not want to undebit the value then? Or in other words why not only debit as the very last thing in this method, after we're sure we're returning a 200 to the disperser?

Feel like this architecture choice should be documented, either here or somewhere else. But its something that tricky and easy to forget, and could lead to some annoying bugs down the line if we make changes.

litt3: > why not only debit as the very last thing in this method

We specifically want to avoid doing work if the payment isn't valid. Otherwise, a malicious client can play tricks using multiple dispersers to get validators to do extra work. Therefore, we must validate payments before doing the hard work

> we don't currently have retries (at all?) on disperser

Not only do we have retries disabled on the disperser, but we explicitly [protect against replays](https://github.com/Layr-Labs/eigenda/blob/validator-payments/node/grpc/server_v2.go#L144) on the validator. So even if a disperser _tried_ to resubmit, it would be rejected prior to the payment being reapplied

```
		hash, err := s.chunkAuthenticator.AuthenticateStoreChunksRequest(ctx, in, time.Now())
		if err != nil {
			return nil, api.NewErrorInvalidArg(fmt.Sprintf("failed to authenticate request: %v", err))
		}

		timestamp := time.Unix(int64(in.GetTimestamp()), 0)
		err = s.replayGuardian.VerifyRequest(hash, timestamp)
		if err != nil {
		}
```


I added some inline docs here:
5fbc2d71e9c9fff74e06a95b06bd3a6eb0f53e7e

PR #2064, opened 2025-09-24
fix: version number

cody-littley:

Introduce proper internal tracking of the software version. Although this will have to be updated manually, we will get CI failures if the version number is wrong in a release branch.


common/version/semver.go
```diff
@@ -0,0 +1,147 @@
...
+	if s.major != other.major {
+		return s.major < other.major
+	}
+	if s.minor != other.minor {
+		return s.minor < other.minor
+	}
+	if s.patch != other.patch {
+		return s.patch < other.patch
+	}
+	return s.errata < other.errata
```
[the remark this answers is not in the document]
cody-littley: fixed

common/version/semver.go
```diff
@@ -0,0 +1,147 @@
...
+// Semver represents a semantic version.
+type Semver struct {
+	major  int
+	minor  int
+	patch  int
+	errata string
+}
+
+// NewSemver creates a new Semver instance.
+func NewSemver(major int, minor int, patch int, errata string) (*Semver, error) {
```
litt3: Why use `int`?

common/version/semver.go
```diff
@@ -0,0 +1,147 @@
...
+		// Try with errata
+		n, err := fmt.Sscanf(versionStr, "%d.%d.%d-%s", &major, &minor, &patch, &errata)
+		if err != nil {
+			return nil, fmt.Errorf("invalid version format: %w", err)
+		}
+		if n != 4 {
+			return nil, fmt.Errorf("invalid version format")
+		}
+	} else {
+		var extra string
```
litt3, line 55: Do we really want to simply discard the extra?

common/version/semver.go
```diff
@@ -0,0 +1,147 @@
...
+
+	return &Semver{
+		major:  major,
+		minor:  minor,
+		patch:  patch,
+		errata: errata,
+	}, nil
+}
+
+// Parses a semantic version string and returns a Semver instance.
```
litt3, line 30: Consider describing here how this handles errata and extra, at a high level

common/version/semver.go
```diff
@@ -0,0 +1,147 @@
...
+		if n != 4 {
+			return nil, fmt.Errorf("invalid version format")
+		}
+	} else {
+		var extra string
+		n, err := fmt.Sscanf(versionStr, "%d.%d.%d%s", &major, &minor, &patch, &extra)
+		if err != nil && !errors.Is(err, io.EOF) {
+			return nil, fmt.Errorf("invalid version format: %w", err)
+		}
+		if n != 3 {
```
litt3: Why is this 3, instead of 4?

common/version/semver.go
```diff
@@ -0,0 +1,147 @@
...
+	return s.patch
+}
+
+// Get the errata string.
+func (s *Semver) Errata() string {
+	return s.errata
+}
+
+// Compares two Semver instances for equality.
+func (s *Semver) Equals(other *Semver) bool {
```
litt3: We don't want to compare errata here? Should probably be mentioned in the doc

common/version/semver.go
```diff
@@ -0,0 +1,147 @@
...
+	if s == nil || other == nil {
+		return false
+	}
+	return s.major == other.major && s.minor == other.minor && s.patch == other.patch
+}
+
+// Compares two Semver instances to see if this one is less than the other. Ignores errata.
+func (s *Semver) LessThan(other *Semver) bool {
```
litt3: Would it make sense to just define a comparator, and then thinly wrap the comparator to provide `LessThan`, `GreaterThan`, etc.?

common/version/semver_test.go
```diff
@@ -0,0 +1,175 @@
...
+	semver, err := SemverFromString(str)
+	require.NoError(t, err)
+	require.Equal(t, major, semver.Major())
+	require.Equal(t, minor, semver.Minor())
+	require.Equal(t, patch, semver.Patch())
+	require.Equal(t, errata, semver.Errata())
+	require.Equal(t, str, semver.String())
+}
+
+func TestSerialization(t *testing.T) {
```
litt3, line 20: Looks like you're testing eratta, but not the "extra" style which is defined within `semver`

node/version.go
```diff
@@ -1,7 +1,5 @@
 package node
 
 var (
-	SemVer    = ">=0.9.0-rc.1"
-	GitCommit = ""
-	GitDate   = ""
+	SemVer = ">=0.9.0-rc.1"
```
litt3: I thought you said you wanted to update this to `2.4`?

common/version/semver.go
cody-littley: converted to `uint64`.

common/version/semver.go
cody-littley: Added a comment.

```
		// "extra" will catch any trailing characters after the last integer. If we have trailing characters, they
		// should always be preceded by a hyphen. Since in this branch we don't have a hyphen, consider any trailing
		// characters to be an error.
		var extra string
```

We just needed a place holder, since if there is anything in it then n won't equal 3. But for the sake of completeness, I also added this check.

```
if n != 3 || extra != "" {
```

common/version/semver.go
cody-littley:
```
// Parses a semantic version string and returns a Semver instance.
//
// Requires the string to have the following format: X.Y.Z[-errata], where X, Y, and Z are
// non-negative integers, and errata is an optional arbitrary string. Note that if
// errata is present, it must be preceded by a hyphen, e.g. "1.2.3-alpha.1".
```

common/version/semver.go
cody-littley: Because `extra` is just to catch if sscanf finds extra characters at the end. Without it, extra characters get ignored.

common/version/semver.go
cody-littley: added

common/version/semver.go
cody-littley: Updated the doc. Also added a `StrictEquals()` method if you do want to check the errata field.

common/version/semver_test.go
cody-littley: The "extra" field is to catch junk characters at the end of the string, which is exercised.

node/version.go
cody-littley: Oops, meant to delete this file. Fixed.

cody-littley: Although I should point out that this version is in fact technically correct. 😜 

correct-technically-correct

litt3 (dismissed):

litt3 (approved):

common/version/semver.go
```diff
@@ -9,7 +9,7 @@ import (
 
 var _ fmt.Stringer = (*Semver)(nil)
 
-// Semver represents a semantic version.
+// Semver represents a semantic defaultVersion.
```
litt3: Rename mistake, should still be "version"

common/version/default_version_test.go
```diff
@@ -26,7 +21,7 @@ func TestCurrentVersion(t *testing.T) {
 	// Check if branch follows the release/SEMVER pattern
 	const releasePrefix = "release/"
 	if !strings.HasPrefix(branch, releasePrefix) {
-		t.Skipf("Current branch '%s' is not a release branch, skipping version check", branch)
+		t.Skipf("Current branch '%s' is not a release branch, skipping defaultVersion check", branch)
```
litt3: seems like rename mistake here too

common/version/semver.go
```diff
@@ -27,7 +27,7 @@ func NewSemver(major uint64, minor uint64, patch uint64, errata string) *Semver
 	}
 }
 
-// Parses a semantic version string and returns a Semver instance.
+// Parses a semantic defaultVersion string and returns a Semver instance.
```
litt3: here too

common/version/semver.go
cody-littley: fixed here and elsewhere

common/version/default_version_test.go
cody-littley: fixed here and elsewhere

common/version/semver.go
cody-littley: fixed here and elsewhere

litt3 (approved):

PR #2065, opened 2025-09-24
feat(payments): Add inabox test coverage for payments system

litt3:
- Creates inabox payments tests
- Makes necessary changes to inabox to support the new tests

NOTE:
- Currently, none of the new payments tests run automatically. They take a long time, and cannot be run in parallel yet
- Eventually, once inabox pain points are resolved, these tests should be run in CI. Until then, at least we can get manual coverage.


inabox/tests/payments/reservation_only_test.go
```diff
@@ -0,0 +1,234 @@
...
+	"github.com/Layr-Labs/eigenda/core/payments/clientledger"
+	"github.com/Layr-Labs/eigenda/core/payments/reservation"
+	integration "github.com/Layr-Labs/eigenda/inabox/tests"
+	"github.com/Layr-Labs/eigenda/test"
+	"github.com/Layr-Labs/eigenda/test/random"
+	"github.com/stretchr/testify/require"
+)
+
+// NOTE: Currently, it doesn't work to run these tests in sequence. Each test must be run as a separate command.
+// The problem is that the cleanup logic sometimes randomly fails to free docker ports, so subsequent setups fail.
```
dmanc, line 18: Is it possible to run the tests sequentially but with the same inabox deployment (no cleanup)?

litt3: The problem is that we need to configure the system in different ways: the point of these tests is to check different combinations of controller + client configurations

If we could redeploy individual parts of the deployment, e.g. keep all components except redeploy a new controller, then this could be much faster, and could be done without any cleanup between tests.

Since deployment is all-or-nothing for now, this was the only way I could think to accomplish the needed behavior

dmanc: I see thanks for the context. Redeployment sounds like a good feature but I think in the short term we would just run them in parallel once we can spin up inabox with OS allocated ports.

inabox/tests/payments/reservation_only_test.go
```diff
@@ -0,0 +1,234 @@
...
+// Once we figure out why resources aren't being freed, then these tests will be runnable the "normal" way.
+
+func TestReservationOnly_LegacyClient_LegacyController(t *testing.T) {
```
dmanc, line 21: Might want to add something like 	t.Skip("Manual test for now"), not sure if the CI is going to try to run the tests automatically

contracts/script/EigenDADeployer.s.sol
```diff
@@ -93,7 +93,7 @@ contract EigenDADeployer is DeployOpenEigenLayer {
     uint64 _pricePerSymbol = 0.447 gwei;
     uint64 _priceUpdateCooldown = 1;
     uint64 _globalSymbolsPerPeriod = 131072;
-    uint64 _reservationPeriodInterval = 300;
+    uint64 _reservationPeriodInterval = 10;
```
samlaf, line 96: Should we move this script to some contracts/script/inabox/ folder?

contracts/script/SetUpEigenDA.s.sol
```diff
@@ -158,7 +158,7 @@ contract SetupEigenDA is EigenDADeployer, EigenLayerUtils {
 
         // Register Reservations for client as the eigenDACommunityMultisig
         IPaymentVault.Reservation memory reservation = IPaymentVault.Reservation({
-            symbolsPerSecond: 452198,
+            symbolsPerSecond: uint64(vm.envOr("USER_RESERVATION_SYMBOLS_PER_SECOND", uint256(452198))),
```
samlaf, line 161: You know how much I hate env vars. There is a config file for this script afair, we could add it there instead?

Or even better do we have golang code to setup a reservation? I feel like we should setup reservations from golang code, then we can set them up whenever/however we wish in your golang tests.

contracts/script/SetUpEigenDA.s.sol
```diff
@@ -167,8 +167,6 @@ contract SetupEigenDA is EigenDADeployer, EigenLayerUtils {
         address clientAddress = address(0x1aa8226f6d354380dDE75eE6B634875c4203e522);
         vm.startBroadcast(msg.sender);
         paymentVault.setReservation(clientAddress, reservation);
-        // Deposit OnDemand
-        paymentVault.depositOnDemand{value: 0.1 ether}(clientAddress);
```
samlaf, line 171: why remove this?

core/thegraph/state_integration_test.go
```diff
@@ -47,6 +47,7 @@ func setupTest(t *testing.T) *inaboxtests.InfrastructureHarness {
 		TestName:          testName,
 		InMemoryBlobStore: true, // Graph indexer test doesn't need blob store
 		Logger:            logger,
+		RootPath:          "../../",
```
samlaf, line 50: ehh we really shouldn't be doing this! This is so error prone. If you ever run this test from some other folder for whatever reason (like from root) it'll break.

@dmanc we really need to fix these paths issues. The right way is to have a file either in root or in some common/paths package, read where it is using golang stdlib, then path.join(../..) to it, and make that a global var that other packages can import. Then regardless of where the package live, or where the code is executed, that RootPath variable will ALWAYS be valid.

disperser/apiserver/server_v2.go
```diff
@@ -143,7 +143,7 @@ func NewDispersalServerV2(
 		controllerConnection = connection
 		controllerClient = controller.NewControllerServiceClient(connection)
 
-		logger.Info("Using controller-based payment system")
+		logger.Info("Using controller-based payment system", "controller", controllerAddress)
```
samlaf, line 146: nit which def doesnt belong in this PR: we should move away from using "address" to mean URL. Whenever I see address I think of ethereum address. We should use "host" and "port" and "url" for rpc "addresses".

inabox/Makefile
```diff
@@ -1,10 +1,30 @@
...
+run-payments-tests:
+	@for test in \
+		TestReservationOnly_LegacyClient_LegacyController \
+		TestReservationOnly_LegacyClient_NewController \
+		TestReservationOnly_NewClient_LegacyController \
+		TestReservationOnly_NewClient_NewController; do \
+		echo "Running $$test..."; \
+		if ! go test ./tests/payments -v -run "^$$test$$"; then \
+			echo "❌ TEST FAILED: $$test"; \
+			$(MAKE) -s stop-services stop-infra; \
```
samlaf: nit: you can use `stop-inabox` to stop both infra and services

inabox/deploy/config.go
```diff
@@ -221,12 +221,20 @@ func (env *Config) generateDisperserV2Vars(ind int, logPath, dbPath, grpcPort st
 		DISPERSER_SERVER_RESERVATIONS_TABLE_NAME: "e2e_v2_reservation",
 		DISPERSER_SERVER_ON_DEMAND_TABLE_NAME:    "e2e_v2_ondemand",
 		DISPERSER_SERVER_GLOBAL_RATE_TABLE_NAME:  "e2e_v2_global_reservation",
+		DISPERSER_SERVER_CONTROLLER_ADDRESS:      "localhost:30000",
```
samlaf: we need to move away from hardcoded ports, but at least for now could you make it a const at the top of the file? I see it reused in at least 2 places.

inabox/tests/setup_infra.go
```diff
@@ -5,20 +5,25 @@ import (
...
+	TestName                        string
+	InMemoryBlobStore               bool
+	Logger                          logging.Logger
+	RootPath                        string
+	MetadataTableName               string
+	BucketTableName                 string
+	MetadataTableNameV2             string
+	UserReservationSymbolsPerSecond uint64
+	ClientLedgerMode                clientledger.ClientLedgerMode
+	ControllerUseNewPayments        bool
```
samlaf, line 33: yikes I can see this getting out of hands again.
This was meant to be an infra only config. But we're now questioning what "infra" means, given that thegraph and localstack are actually disperser-specific infra (potentially, although not sure its needed for inabox...)

In any case these disperser specific config stuff should live in some other ServicesConfig or something. Haven't figured out the right organization but want to avoid these bloated 40+fields configs.

inabox/tests/payments/payload_submitter.go
```diff
@@ -0,0 +1,87 @@
...
+
+	"github.com/Layr-Labs/eigenda/api/clients/v2/coretypes"
+	"github.com/Layr-Labs/eigenda/api/clients/v2/payloaddispersal"
+	"github.com/Layr-Labs/eigenda/test/random"
+)
+
+// Submits payloads at a certain rate for a duration.
+//
+// Returns dispersal errors on a channel.
+func SubmitPayloads(
```
samlaf:
```suggestion
func submitPayloads(
```

samlaf: Q: I know that this is very simple, but could it reuse the load generator code instead?

inabox/tests/setup_test_harness.go
```diff
@@ -117,10 +127,14 @@ func NewTestHarnessWithSetup(infra *InfrastructureHarness) (*TestHarness, error)
 	}
 
 	// Setup payload disperser
-	if err := setupPayloadDisperserForContext(testCtx, infra); err != nil {
+	if err := setupPayloadDisperserForContext(ctx, testCtx, infra); err != nil {
```
samlaf, line 130: ok so imo (havent checked all the dependencies and whether this works cleanly given current state of inabox), clients should NOT be part of the inabox infra. They should be built on a per-test basis when needed.

This is the right separation of concern imo. This way you can control your payment tests, especially parallelization. Run 4 tests in parallel, they each start by creating a new reservation, and then build the clients using that private key.

inabox/tests/setup_test_harness.go
```diff
@@ -362,3 +553,30 @@ func newTransactOptsFromPrivateKey(privateKeyHex string, chainID *big.Int) *bind
 
 	return opts
 }
+
+func setupPaymentVaultTransactor(
```
samlaf, line 557: this is also a very generic/useful component right, that we could reuse as part of a CLI tool? Why have this under inabox?

inabox/tests/payments/reservation_only_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+	infra, err := integration.SetupInfrastructure(infraConfig)
+	require.NoError(t, err)
+
+	testHarness, err := integration.NewTestHarnessWithSetup(infra)
+	require.NoError(t, err)
+
+	t.Cleanup(func() {
+		testHarness.Cleanup()
+		integration.TeardownInfrastructure(infra)
```
samlaf: you prob want to call teardownInfra right after SetupInfra, otherwise if NewTestHarnessWithSetup errors, infra will not get cleaned.

inabox/tests/payments/reservation_only_test.go
```diff
@@ -0,0 +1,234 @@
...
+
+	t.Run("Within reservation limits", func(t *testing.T) {
+		initialReservation, err := reservation.NewReservation(
+			1024,
+			time.Now().Add(-1*time.Hour),
+			time.Now().Add(24*time.Hour),
+			[]core.QuorumID{0, 1},
+		)
+		require.NoError(t, err)
+		err = testHarness.UpdateReservation(t.Context(), t, initialReservation)
```
samlaf: nit: can we rename `UpdateReservationOnchain` or something to make it obvious we're making an eth tx? otherwise it could read as "read from onchain and update the test harness' reservation"

inabox/tests/payments/reservation_only_test.go
litt3: d21cb0cbd1b8a2fc725f70de55467d7e56265558

contracts/script/SetUpEigenDA.s.sol
litt3: > You know how much I hate env vars

I share your dislike, but I'm not sure there's a better solution in the short term.

> There is a config file for this script afair, we could add it there instead?

The problem is that I need to define this at test time. So short of duplicating the config file with a single value changed, I'm not sure how this would be accomplished. Even if we had duplicate config files, there would have to be some mechanism of injecting the correct config file path to this deployment script, and I'm not sure how else we'd do it, other than env vars. 

> Or even better do we have golang code to setup a reservation? I feel like we should setup reservations from golang code, then we can set them up whenever/however we wish in your golang tests.

I wanted to go that way, but it's not trivial to do based on how the infrastructure is all deployed together. The flow we would _want_ is this:

1. deploy infra
2. submit some smart contract transactions
3. spin up eigenDA components

This is further complicated by the fact that the old `Accountant` doesn't observe `PaymentVault` updates in real time, so the reservation must be in place before the clients are created.

contracts/script/SetUpEigenDA.s.sol
litt3: I need to test edge cases which include NOT having enough on demand funds. This is a huge deposit, and it's not possible to un-deposit funds at test time.

I removed this, and we can just deposit on demand funds as needed, on a test by test bases. Most tests will probably rely on reservations, anyway.

core/thegraph/state_integration_test.go
litt3: > ehh we really shouldn't be doing this! This is so error prone.

💯 

Current code had this hardcoded, which made it impossible to put tests in a subdir. This is just my workaround to get these tests written, without overhauling inabox right away

disperser/apiserver/server_v2.go
litt3: I feel like it's generally clear from context, but I agree it's not ideal.

> We should use "host" and "port" and "url" for rpc "addresses".

Makes sense, I'm on board with that. As long as we don't try to use `socket` 🤮

inabox/tests/payments/payload_submitter.go
litt3: > could it reuse the load generator code instead?

I took a look- I don't think that could be reused here with very substantial modification. The load generator is doing a ton of stuff, and here we only need to do a very small fraction.

It's also nice to be able to add logs without affecting other things, and to have access to a `testing.T` instance, which we wouldn't want to add to the load generator

inabox/Makefile
litt3: 7c86143fe9d43c8c0d6082fe6f23f2a0dcb7db60

inabox/tests/setup_infra.go
litt3: I anticipated some discussion on this point. Some things I considered:

- All three of these fields are temporary: once the old payment system is removed, they will be deleted
- Each of these three values belongs to a different component. So if I put these into specific config structs, and nested those inside of the infra config, we would still have 3 fields here. But each field would be a config struct that contains only a single value. 
- I think the correct way to do this would be to have separate setup steps for each part of the infrastructure stack. So there would be a `ControllerConfig`, but not nested inside the `InfraConfig`. It would be totally separate, passed into some kind of `SetupController` function. Since reworking the flow of instantiation is a much larger change, I opted to keep it simple for now.

contracts/script/EigenDADeployer.s.sol
litt3: Makes sense to me (but in a separate PR)

inabox/deploy/config.go
litt3: > we need to move away from hardcoded ports

👍 

Not sure what exactly that would look like in this PR. So for now:
9a0ba5f1648e15d9cb27d1108c8330e9d0f97be9

inabox/tests/payments/payload_submitter.go
litt3: 964ff49b7eded5c9461ee5b53ee674d728c3b735

edit:
whoops, missed some changes. 87dcf91655d89f014d5d4507b669ed85df0c241a

inabox/tests/setup_test_harness.go
litt3: Definitely on board. 

As I've mentioned elsewhere, I feel like our infra setup should be split in to multiple parts _in general_. But it's especially the case for clients, that they shouldn't be part of the "infra".

I don't think this will be hugely difficult to pull off, but there will be some level of reorganization, which I think exceeds the amount of inabox tweaks that fit into this PR.

inabox/tests/setup_test_harness.go
litt3: The transactor itself is provided by the bindings. This method is just to instantiate the existing transactor, so that tests can access it through the `TestHarness`. 

We might want a CLI to update the `PaymentVault` at some point, but this method here isn't directly applicable to anything beyond the scope of the `TestHarness`

inabox/tests/payments/reservation_only_test.go
litt3: 4735210e6558883977655f0695e953d0f9a9b1a7

inabox/tests/payments/reservation_only_test.go
litt3: 30e6d8a2b5085ce5b515ff30845d3cb46c6730e6

dmanc (approved):

contracts/script/SetUpEigenDA.s.sol
pakim249CAL: I think a deployment generally should NOT be setting stuff like this, when it can be accomplished with a regular tx. So I'm not against this change.

pakim249CAL (dismissed):

contracts/script/SetUpEigenDA.s.sol
samlaf: @dmanc here's a good reason for separating infra deployment (at least chain) from eigenda services!

contracts/script/SetUpEigenDA.s.sol
samlaf: Issue is that this is used in inabox the devnet environment, where we most likely do want a deposit.

@litt3 do you have an example test where you do a deposit offchain? Didn't see one.

@pakim249CAL this is a setup script though, not a deployment script. It first deploys Eiegenlayer and EigenDA by calling some other script. And then it registers operators and clients. Tbh I agree we should have offchain code to setup operators/clients, but until we do, I prefer we keep this script working for devnet.

core/thegraph/state_integration_test.go
samlaf: Oh you introduced RootPath in this PR? If it was already hardcoded, then it was probably hardcoded as a relative path, which was the issue. Could you just change that hardcoded path to be an absolute path using the strategy I suggested above. I'd prefer we don't introduce more custom relative path logic like this.

inabox/tests/setup_infra.go
samlaf: agree with your 3rd bullet point, that's also how I'm envisioning it. In the meantime if this is just temporary stuff could you add a "legacy payment stuff; will soon be deprecated" comment above those 3 fields and keep them separate from the rest?

contracts/script/SetUpEigenDA.s.sol
litt3: > Issue is that this is used in inabox the devnet environment, where we most likely do want a deposit.

Why would anyone care to have an on-demand deposit? There is a massive ~16MiB/sec default reservation, which can cover any conceivable rate of dispersal. As currently configured, this default on-demand deposit has never once actually been used. I don't see any reason to use on-demand payments for tests, other than for cases where you're specifically testing payment behavior.

> do you have an example test where you do a deposit offchain? Didn't see one.

Additional tests are coming for on-demand. The tests in this PR are only for reservation based payments.

That aside, even these reservation-only tests rely on _not_ having a huge on-demand deposit for the test account. These tests make assertions about failed dispersals when the reservation cannot cover a dispersal, so having a big on-demand deposit removes the ability to make any assertions about the limits of a reservation (for old clients, that don't declare payment mode).

inabox/tests/setup_infra.go
litt3: 27f08b8a869385d0ca0d8319c5bb3e2858830829

core/thegraph/state_integration_test.go
litt3: > Could you just change that hardcoded path to be an absolute path using the strategy I suggested above

I implemented this, and it's too big to include in this PR. There are _lots_ of touch points with root path within inabox.

> I'd prefer we don't introduce more custom relative path logic like this.

This isn't introducing any custom logic, it's just the existing logic but slightly expanded to support different nesting of the test location.

I'm on board with the change in general, but it's not something that belongs tacked onto this PR

contracts/script/SetUpEigenDA.s.sol
samlaf: I see. makes sense. resolving

samlaf (approved):

PR #2067, opened 2025-09-24
Feat/cert verifier configs

pakim249CAL:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


samlaf (dismissed):
LGTM. I wonder if we should just use a single config for all networks to make sure they are always the same (I can't think of a reason why we would want them to diverge).

Although I do guess it makes sense to have them separate if we ever plan to change them in lockstep (say hoodi before others).

ethenotethan (approved):
lgtm

PR #2069, opened 2025-09-24
fix(inabox): network issues

samlaf:
There was actually a few errors that we ran into. graph couldnt connect to chain. added better error handling as well in deploy-all.

Most importantly creating a network instance for anvil and graph to run in. test-containers doesn't allow naming this network... so we end up with a lot of garbage UUID docker networks... these will need to be pruned individually (or by `docker system prune -A`).


dmanc (approved):
thx for fixing

PR #2070, opened 2025-09-24
refactor(inabox): cleanup inabox deployer + makefile

samlaf:
Simplify inabox deployment. Everyone is always confused as to how to start it. There's too many commands. We always want to start all of it, so no need for all the subcommands that start portions of it. Got rid of all of that and only left with simple makefile commands:
1. make new-anvil (prob can find a better name...)
2. make start-infra
3. make start-services (calls bin.sh)


dmanc (approved):
🧹

dmanc (approved):

PR #2071, opened 2025-09-24
relax encoded payload header

bxue-l2:

The spec is too restrictive on how the header in the encoded payload should be set (which has to be 0). This PR makes it less restrictive, while putting the recommended option.

Currently on the read path, both proxy and hokulea does not check if they are zero-filled.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


samlaf (approved):
Add docs: as prefix to your PR title

bxue-l2:
specifying zero-fill has some advantage
we can be worryfree to use those remaining bytes in the future.

PR #2073, opened 2025-09-25
chore: generalize the leaky bucket so it can be used elsewhere in the codebase

cody-littley:

The leaky bucket utility is useful, but right now it is tightly coupled to payments code.


common/ratelimit/leaky_bucket.go
```diff
@@ -1,45 +1,55 @@
...
-	symbolsPerSecondLeakRate float64
+	// The quantity of "water" that leaks out of the bucket each second, as determined by the configuration.
+	leakRate float64
 
-	// The number of symbols currently in the bucket
+	// The amount of "water" currently in the bucket
 	currentFillLevel float64
 
 	// The time at which the previous leak calculation was made
 	previousLeakTime time.Time
```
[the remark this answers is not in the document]
cody-littley: um... no, these fields are not in fact exported

common/ratelimit/overfill_behavior.go
```diff
@@ -1,4 +1,4 @@
-package reservation
+package ratelimit
```
litt3, line 1: Documentation in this file hasn't been generalized

common/ratelimit/leaky_bucket.go
```diff
@@ -1,45 +1,55 @@
...
+// might be too small to fit the largest permissible quantity of work.
 //
 // NOTE: This struct doesn't do any synchronization! The caller is responsible for making sure that only one goroutine
 // is using it at a time.
 type LeakyBucket struct {
 	// Defines different ways that overfilling the bucket should be handled
 	overfillBehavior OverfillBehavior
 
-	// The total number of symbols that fit in the bucket
+	// The total quantity of "water" that fit in the bucket
```
litt3:
```suggestion
	// The total quantity of "water" that fits in the bucket
```

common/ratelimit/leaky_bucket.go
```diff
@@ -1,45 +1,55 @@
...
 	previousLeakTime time.Time
 }
 
-// Creates a new instance of the [LeakyBucket] algorithm
+// Creates a new instance of the LeakyBucket algorithm
 func NewLeakyBucket(
-	// how fast symbols leak out of the bucket
-	symbolsPerSecondLeakRate uint64,
-	// bucketCapacityDuration * symbolsPerSecondLeakRate becomes the bucket capacity
+	// how fast "water" leaks out of the bucket per second
```
litt3: "how fast" and "per second" seem redundant.

Either it's "how fast", or it's "how _much_ ... per second".

common/ratelimit/leaky_bucket.go
```diff
@@ -48,14 +58,14 @@ func NewLeakyBucket(
 	// the current time, when this is being constructed
 	now time.Time,
 ) (*LeakyBucket, error) {
-	if symbolsPerSecondLeakRate == 0 {
-		return nil, errors.New("symbolsPerSecondLeakRate must be > 0")
+	if leakRate == 0 {
```
litt3: since it's a float now instead of an unsigned int, you need to assert it's not negative.

common/ratelimit/leaky_bucket.go
```diff
@@ -133,7 +143,23 @@ func (lb *LeakyBucket) CheckFillLevel(now time.Time) float64 {
...
+	if fillLevel < 0 {
+		return errors.New("fill level must be >= 0")
+	}
+
+	if fillLevel > lb.bucketCapacity && lb.overfillBehavior == OverfillNotPermitted {
+		return fmt.Errorf("fill level %f exceeds bucket capacity %f, but overfilling is not permitted",
+			fillLevel, lb.bucketCapacity)
+	}
+
+	lb.previousLeakTime = now
```
litt3: You're adjusting the `previousLeakTime`, but not actually leaking. I don't think this is correct.

Either you need to call `leak`, and let that method be responsible for `previousLeakTime`. Or you need to simply set the fill level, and leave leakage alone.

I lean towards a simple setter, since that's the feature that was actually required by the payments implementation.

common/ratelimit/leaky_bucket.go
```diff
@@ -133,7 +143,23 @@ func (lb *LeakyBucket) CheckFillLevel(now time.Time) float64 {
...
 }
 
-// Reverts a previous fill, i.e. removes the number of symbols that got added to the bucket
+// Overrides the current fill level of the bucket, setting it to the specified value.
+func (lb *LeakyBucket) SetFillLevel(now time.Time, fillLevel float64) error {
+	if fillLevel < 0 {
+		return errors.New("fill level must be >= 0")
+	}
+
+	if fillLevel > lb.bucketCapacity && lb.overfillBehavior == OverfillNotPermitted {
```
litt3: I don't think this is the correct error case to have.

Consider the case where overfill _isn't_ permitted, and you want to resize the bucket, while maintaining the same fill level (this is the usecase in payments). Let's say an on-chain configuration is decreasing the size of the bucket, but the previous fill level is greater than the new configured capacity.

It doesn't seem like you should error in that case- you should simply modify the capacity, and then be forced to wait for the overfill to drain.

common/ratelimit/leaky_bucket.go
```diff
@@ -133,7 +143,23 @@ func (lb *LeakyBucket) CheckFillLevel(now time.Time) float64 {
 	return lb.currentFillLevel
 }
 
-// Reverts a previous fill, i.e. removes the number of symbols that got added to the bucket
+// Overrides the current fill level of the bucket, setting it to the specified value.
+func (lb *LeakyBucket) SetFillLevel(now time.Time, fillLevel float64) error {
+	if fillLevel < 0 {
+		return errors.New("fill level must be >= 0")
```
litt3: including the actual value would be useful IMO, to help track down a bug if something unexpected is ending up here.

This also goes for other float checking in this class.

common/ratelimit/overfill_behavior.go
cody-littley: fixed

common/ratelimit/leaky_bucket.go
cody-littley: fixed

common/ratelimit/leaky_bucket.go
cody-littley: fixed

common/ratelimit/leaky_bucket.go
cody-littley: check updated

common/ratelimit/leaky_bucket.go
cody-littley: added

common/ratelimit/leaky_bucket.go
cody-littley: This method is now removed, per our discussion

common/ratelimit/leaky_bucket.go
cody-littley: This method is now removed, per our discussion

core/payments/reservation/reservation_ledger.go
```diff
@@ -61,15 +61,15 @@ func NewReservationLedger(
 // If the bucket doesn't have enough capacity to accommodate the fill, symbolCount IS NOT added to the bucket, i.e. a
 // failed debit doesn't count against the meter.
 func (rl *ReservationLedger) Debit(
-	// now should be from a source that includes monotonic timestamp for best results.
-	// This a local time from the perspective of the entity that owns this ledger instance, to be used with the local
-	// leaky bucket: it should NOT be sourced from the PaymentHeader
+// now should be from a source that includes monotonic timestamp for best results.
```
litt3: consider re-saving this file with vscode, to fix the messed up formatting introduced by goland

common/ratelimit/leaky_bucket.go
```diff
@@ -227,6 +213,37 @@ func (lb *LeakyBucket) GetRemainingCapacity() float64 {
 }
 
 // Gets the total capacity of the bucket.
-func (lb *LeakyBucket) GetBucketCapacity() float64 {
+func (lb *LeakyBucket) GetCapacity() float64 {
 	return lb.bucketCapacity
 }
+
+// Reconfigure bucket parameters. Preservers fill level. If the new bucket capacity is smaller than the current
```
litt3:
```suggestion
// Reconfigure bucket parameters. Preserves fill level. If the new bucket capacity is smaller than the current
```

litt3 (dismissed):

core/payments/reservation/reservation_ledger.go
cody-littley: done

common/ratelimit/leaky_bucket.go
cody-littley: fixed

litt3 (approved):

PR #2074, opened 2025-09-25
refactor(inabox): Create a test harness for inabox tests

dmanc:

Creates a test harness that contains all the utilities and services related to both V1 and V2 integration tests.
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


inabox/tests/integration_suite_test.go
```diff
@@ -51,53 +51,68 @@ import (
...
-*/
-var (
-	anvilContainer     *testbed.AnvilContainer
-	graphNodeContainer *testbed.GraphNodeContainer
-	churnerServer      *grpc.Server
-	churnerListener    net.Listener
-	// chainDockerNetwork is only used by anvil and graphNode
-	chainDockerNetwork *testcontainers.DockerNetwork
+// Global test harness instance - will be initialized once per test suite
+var testHarness *TestHarness
```
samlaf: hmm this doesn't seem like its doing much?
The actual goal should be to separate actually global vars (coming from inabox infra 3 components) from other configs that should be rebuilt for each test (this is actually what a testHarness is).

I'm fine with having global vars for the stuff that is actually global (external dependencies), but I think clients, configs, etc, should all be rebuilt for every test.

Otherwise the issue is to change/fix a test it's very scary to go change a config. And the config above is impossible to understand, there's just too many different random things bunched together.

inabox/tests/test_harness.go
```diff
@@ -0,0 +1,98 @@
...
+	MetadataTableNameV2 string
+
+	// Deployment configuration (shared)
+	TestConfig *deploy.Config
+
+	// Logger
+	Logger logging.Logger
+
+	// Context management
+	Cancel context.CancelFunc
```
samlaf: what context is this managing?

inabox/tests/test_harness.go
```diff
@@ -0,0 +1,98 @@
...
+	// DynamoDB table names (global for the test suite)
+	MetadataTableName   string
+	BucketTableName     string
+	MetadataTableNameV2 string
+
+	// Deployment configuration (shared)
+	TestConfig *deploy.Config
+
+	// Logger
+	Logger logging.Logger
```
samlaf, line 53: don't think we need this here. this should be per test actually. different tests can use different log levels.

inabox/tests/integration_v2_test.go
```diff
@@ -178,23 +183,23 @@ func TestEndToEndV2Scenario(t *testing.T) {
...
 	require.True(t, errors.As(err, &certErr))
 	// TODO(samlaf): after we update to CertVerifier 4.0.0 whose checkDACert will return error bytes,
 	// we should check that extra bytes returned start with signature of the InvalidInclusionProof error
 	require.Equal(t, verification.StatusInvalidCert, certErr.StatusCode)
 }
 
-func validateTxReceipt(ctx context.Context, txHash gethcommon.Hash) error {
-	receipt, err := ethClient.TransactionReceipt(ctx, txHash)
+func validateTxReceipt(ctx context.Context, testHarness *TestHarness, txHash gethcommon.Hash) error {
+	receipt, err := testHarness.EthClient.TransactionReceipt(ctx, txHash)
```
samlaf, line 202: only pass the ethClient, not the whole testHarness

inabox/tests/integration_test.go
```diff
@@ -20,15 +20,20 @@ import (
...
 	for i := 0; i < numBlocks; i++ {
-		err := rpcClient.CallContext(t.Context(), nil, "evm_mine")
+		err := testHarness.RPCClient.CallContext(t.Context(), nil, "evm_mine")
 		require.NoError(t, err)
 	}
 }
 
 func TestEndToEndScenario(t *testing.T) {
+	// Create a fresh test context for this test
+	testCtx, err := NewTestHarnessWithSetup(globalInfra)
```
samlaf: why call this testCtx instead of testHarness? Its not a ctx right?

inabox/tests/setup_infra.go
```diff
@@ -0,0 +1,211 @@
...
+) (*InfrastructureHarness, error) {
+	infra := &InfrastructureHarness{
+		TemplateName:        templateName,
+		TestName:            testName,
+		InMemoryBlobStore:   inMemoryBlobStore,
+		Logger:              logger,
+		MetadataTableName:   "test-BlobMetadata",
+		BucketTableName:     "test-BucketStore",
+		MetadataTableNameV2: "test-BlobMetadata-v2",
+	}
```
samlaf: these things are input to the building of a infraHarness. take them out of the infraHarness and into a infraHarnessConfig or something. Then setupLocalInfrastructure below will take an infraHarnessConfig and output an infraHarness. Way cleaner than passing in a half-built infraHarness that gets modified in-place.

inabox/tests/setup_infra.go
```diff
@@ -0,0 +1,211 @@
...
+
+	ctx, cancel := context.WithTimeout(context.Background(), 5*time.Minute)
```
samlaf, line 39: should probably have this injected instead of using random 5 minutes. tests might end up taking longer than 5 minutes at some point. we should just use t.Context() for tests.

inabox/tests/integration_test.go
```diff
@@ -20,15 +20,20 @@ import (
 	"github.com/stretchr/testify/require"
 )
 
-func mineAnvilBlocks(t *testing.T, numBlocks int) {
+func mineAnvilBlocks(t *testing.T, testHarness *TestHarness, numBlocks int) {
```
samlaf: better to take minimal set of dependencies. only pass the rpcClient if thats only what is needed, not the whole testHarness.

inabox/tests/setup_infra.go
dmanc: I can't use t.Context() here because it's part of TestMain

samlaf: I see. I was reading quickly. Keeping the hardcoded 5 mins is prob fine then given that this function would know better how long it takes to spin up the infra then its callers. Should prob still take a ctx as input to the function instead of using a context.Background() like this

inabox/tests/test_harness.go
dmanc: This logger is for infrastructure deployment related, not used inside the actual tests currently.

inabox/tests/integration_test.go
dmanc: 4b937c0748862cef91cb6aa74ab6e4745383d157

inabox/tests/setup_infra.go
dmanc: 4b937c0748862cef91cb6aa74ab6e4745383d157, not sure if this is what you meant.

inabox/tests/integration_test.go
dmanc: 4b937c0748862cef91cb6aa74ab6e4745383d157

inabox/tests/integration_v2_test.go
dmanc: 4b937c0748862cef91cb6aa74ab6e4745383d157

inabox/tests/test_harness.go
dmanc: 4b937c0748862cef91cb6aa74ab6e4745383d157, removed

PR #2075, opened 2025-09-25
feat: new and improved ejector

cody-littley:

A new ejector that is compatible with the new ejection pattern.


ejector/ejector_config.go
```diff
@@ -69,21 +69,19 @@ type EjectorConfig struct {
 
 	// The period at which to periodically attempt to finalize ejections that have been started.
 	EjectionFinalizationPeriod time.Duration
-}
 
-var _ config.VerifiableConfig = (*EjectorSecretConfig)(nil)
+	// The number of blocks to wait before using a reference block number for quorum.
```
litt3: > using a reference block number for quorum.

Not understanding this comment fully. How does a specific quorum factor in here?

core/eth/validator_quorum_lookup.go
```diff
@@ -0,0 +1,124 @@
...
+
+var _ ValidatorQuorumLookup = (*validatorQuorumLookup)(nil)
+
+// A standard implementation of the ValidatorQuorumLookup interface.
+type validatorQuorumLookup struct {
+	registryCoordinator *regcoordinator.ContractEigenDARegistryCoordinator
+}
+
+// Create a new ValidatorQuorumLookup instance.
+func NewValidatorQuorumLookup(
```
litt3, line 33: Did you intentionally not implement this constructor?

core/eth/validator_quorum_lookup.go
```diff
@@ -0,0 +1,124 @@
...
+) ([]core.QuorumID, error) {
+
+	blockNumber := big.NewInt(int64(referenceBlockNumber))
+
+	opts := &bind.CallOpts{
+		Context:     ctx,
+		BlockNumber: blockNumber,
+	}
+
+	// This method returns a bitmap as a big.Int, where the first 255 bits represent membership in quorums 0-254.
```
litt3, line 66:
```
        :$x+;+XX+XXXX&&&&Xxx                                                  
      .X+++++;:;x+++++;;++++X:                                                
      :     ;+; ++++;+x+;;+;;X:                                               
      ; .. ;XX;+$x+;;;;+x++;;+X                                               
     &;;+&&$+: + ;x+++;;++;;:xx                                               
   .&+;+;    ;X$X:+;++;+;+x+;:;                                               
  &.  ;:     Xx;++x++++;;;+++;:;                                              
 :    :.  .+$x;;:;+++++xx++x++++;                                             
 +&;+.:x&&&X++;;+x;++++++++;;+++:$                                            
  + +;Xx;. ;:;x+::+;:;+++++x+;+++;$                                           
  . +$+;;+;;.+:;x;:;x::+++;+++;;+;+X                                          
    &;:;:;;;:::x+++:;x+;+;++++++;;++                                          
  ;$X++;::++;::.:Xx+;:;x++;;++++x+;;+                                         
    ;x+++;::+;;::; X;+:;+++++++++x+;;x:                                       
      &++++;:.;;;+; x;;:;;+++++++++x;;Xx                                      
     ::;XXxxx;++;:;;:X+;;;;+++++++++++:Xx                                     
      + .;;;;;+;:+;:::;X;;+++++++++++x+;xx                                    
      ;  ;:;+;;;;;;++::;x;;;;+++++++++++++:                                   
        :x&XX++++;:;:;;;;++;;;;;;+++++x+++x:                                  
            :X+xxx+++;::;;;+;;;;;;;++x;:::;x$+                                
                :Xxxx+++;::;+;;;;;;;;+X:;;;X;X;                               
                  ;+++++;;+::;++;;;+;;xx++;;;+$:                              
                    ;X+++++++;::+x++;;;++x+++;;X;                             
                     . &Xx++++++;;+xx;:;+++++;;;:+                            
                          Xx++++++;:;+++++++++++:.+.                          
                            ;X+++++++;;:;;+++++x+;;$                          
                              X&X++;++++;;::;++++++;&                         
                                 +&X+;+++++;:;;+xx+;;X.                       
                                    :&X++++X+;.::;++;X+$x                     
                                       :&X+++xx;::;++X.+:&                    
                                          :X+;+x+;;+$x;+;++                   
                                           .&&&$$XX;;   x :X                  
                                                 ::.   && ;;&                 
                                                   X; x+; ;xXX+               
                                                     +$$.  +&XX;              
                                                      ;&:;&& ::+;:            
                                                       X+    ;X;;$;           
                                                         $x  $:X+:+:          
                                                          &X&X   ;X$          
                                                           :&.  .$+xx         
                                                             &  ++x+;;        
                                                              &+$x;::+x       
                                                               X$.  X&x+$     
                                                                ++ +: :;;&    
                                                                 XX;  +x++$   
                                                                  XXXx++++;X; 
                                                                   $+++++++;& 
                                                                   :$+;;;;;.& 
                                                                     :&x+;+;& 
                                                                              

```

core/eth/validator_quorum_lookup.go
```diff
@@ -0,0 +1,124 @@
...
+	bigIntBitmap, err := v.registryCoordinator.GetCurrentQuorumBitmap(opts, validatorID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get quorum bitmap: %w", err)
+	}
+	bitmap := bigIntBitmap.Bytes()
+
+	quorumIDs := make([]core.QuorumID, 0)
+
+	// Although technically 254 is the max quorum ID (due to an embarrassing off-by-one typo), it doesn't hurt
```
litt3, line 74: I thought we determined that the max was something like 192 or something like that?

core/eth/validator_quorum_lookup.go
```diff
@@ -0,0 +1,124 @@
...
+}
+
+// GetQuorumsForValidator implements ValidatorQuorumLookup.
+func (c *cachedValidatorQuorumLookup) GetQuorumsForValidator(
+	ctx context.Context,
+	validatorAddress core.OperatorID,
+	referenceBlockNumber uint64,
+) ([]core.QuorumID, error) {
+
+	if quorums, ok := c.cache.Get(validatorAddress); ok {
```
litt3: This `Get` isn't taking the rbn into account at all. Seems like that can't be correct

ejector/ejection_criteria.go
```diff
@@ -0,0 +1,16 @@
...
+)
+
+// Checks to see if a validator is eligible for ejection. A validator is considered to be eligible for ejection if
+// 1. It has not signed for any batches in the time period (i.e. signed_batches == 0), and
+// 2. There were batches it COULD have signed (i.e. unsigned_batches > 0)
+//
+// Note that the SLAs for validators requiest a much higher standard for ejection (i.e. 99.5% signing rate over a day).
+// But due to constraints in the ejection protocol, there is little point in attempting to eject a validator that
+// is currently online (even if it isn't doing a good job).
+func IsEjectable(signingRate *validator.ValidatorSigningRate) bool {
```
litt3: This is a one liner, called in only one location. Does it actually need to be a utility function in its own file?

ejector/main/main.go
```diff
@@ -0,0 +1,157 @@
...
+		geth.EthClientConfig{
+			RPCURLs:          secretConfig.EthRpcUrls,
+			PrivateKeyString: secretConfig.PrivateKey,
+			NumConfirmations: ejectorConfig.EthBlockConfirmations,
+			NumRetries:       ejectorConfig.EthRpcRetryCount,
+		},
+		senderAddress,
+		logger)
+	if err != nil {
+		logger.Error("Cannot create chain.Client", "err", err)
```
litt3: Why does this error case log, and all the other ones don't?

ejector/main/main.go
```diff
@@ -0,0 +1,157 @@
...
+		return fmt.Errorf("failed to create geth client: %w", err)
+	}
+
+	contractDirectory, err := directory.NewContractDirectory(
+		context.Background(),
```
litt3: any reason not to use the preexisting ctx?

ejector/ejection_transactor.go
```diff
@@ -22,3 +31,223 @@ type EjectionTransactor interface {
...
+	selfAddress gethcommon.Address
+
+	// Used to execute eth reads
+	caller *contractEigenDAEjectionManager.ContractIEigenDAEjectionManagerCaller
+
+	// Used to execute eth writes
+	transactor *contractEigenDAEjectionManager.ContractIEigenDAEjectionManagerTransactor
+
+	// A function that can sign transactions from selfAddress.
+	signer bind.SignerFn
```
litt3: Where'd you dig up this type? It looks like it's part of an optimism libarary- is that actually what we should use?

ejector/ejector_config.go
```diff
@@ -69,21 +69,19 @@ type EjectorConfig struct {
...
-type EjectorSecretConfig struct {
-	// The Ethereum RPC URL(s) to use for connecting to the blockchain.
-	EthRpcUrls []string `docs:"required"`
+	// The interval at which to poll for a new reference block number.
+	ReferenceBlockNumberPollInterval time.Duration
 
-	// The private key to use for signing ejection transactions.
-	PrivateKey string `docs:"required"`
+	// The size of the cache to use for Ethereum-related caching layers.
+	EthCacheSize int
```
litt3: I'm not crazy about the name `EthCacheSize`, but I don't know if I can think of any better alternatives

ejector/ejection_transactor.go
```diff
@@ -22,3 +31,223 @@ type EjectionTransactor interface {
...
+		transactor:                    transactor,
+		signer:                        transactOpts.Signer,
+		referenceBlockProvider:        referenceBlockProvider,
+		quorumScanner:                 quorumScanner,
+		validatorQuorumLookup:         validatorQuorumLookup,
+		validatorIDToAddressConverter: validatorIDToAddressConverter,
+	}, nil
+}
+
+// CompleteEjection implements EjectionTransactor.
```
litt3: There are lots of these types of autogen comments you mentioned you normally delete

ejector/ejection_transactor.go
```diff
@@ -22,3 +31,223 @@ type EjectionTransactor interface {
...
+	// A utility for looking up which quorums a given validator is a member of at a specific reference block number.
+	validatorQuorumLookup eth.ValidatorQuorumLookup
+
+	// A utility for converting between validator IDs and addresses.
+	validatorIDToAddressConverter eth.ValidatorIDToAddressConverter
+}
+
+// Create a new EjectionTransactor.
+func NewEjectionTransactor(
+	ctx context.Context,
```
litt3: unused

core/eth/reference_block_provider.go
```diff
@@ -77,3 +78,46 @@ func (r *referenceBlockProvider) GetReferenceBlockNumber(ctx context.Context) (u
...
+	// The last time we updated the reference block number.
+	lastUpdate time.Time
+}
+
+// NewPeriodicReferenceBlockProvider creates a new ReferenceBlockProvider that wraps the given base
+// ReferenceBlockProvider. The returned implementation will only call the base provider once every updatePeriod, and
+// will return the last value it saw in between updates.
+func NewPeriodicReferenceBlockProvider(
+	base ReferenceBlockProvider,
+	updatePeriod time.Duration,
```
litt3, line 104: consider checking validity of `updatePeriod` (>=0)

ejector/ejection_transactor.go
```diff
@@ -22,3 +31,223 @@ type EjectionTransactor interface {
...
+		return fmt.Errorf("failed to get reference block number: %w", err)
+	}
+
+	quorums, err := e.quorumScanner.GetQuorums(ctx, rbn)
+	if err != nil {
+		return fmt.Errorf("failed to get quorums: %w", err)
+	}
+	quorumBytes := eth.QuorumListToBytes(quorums)
+
+	opts := &bind.TransactOpts{
```
litt3, line 171: Is there any reason you aren't passing on ctx when creating `TransactOpts`? Here and elsewhere

ejector/ejector.go
```diff
@@ -0,0 +1,205 @@
...
+	period time.Duration
+
+	// Defines the time window over which to evaluate signing metrics when deciding whether to eject a validator.
+	ejectionCriteriaTimeWindow time.Duration
+
+	// Used to convert validator IDs to validator addresses.
+	validatorIDToAddressCache eth.ValidatorIDToAddressConverter
+
+	// Used to look up the latest reference number.
+	referenceBlockProvider eth.ReferenceBlockProvider
```
litt3, line 39: referenceBlockProvider and operatorStateCache aren't actually being initialized in the constructor

ejector/ejector.go
```diff
@@ -0,0 +1,205 @@
...
+}
+
+// evaluateValidators looks up signing rates and evaluates which validators should be ejected.
+func (e *Ejector) evaluateValidators() {
+
+	v1SigningRates, err := e.signingRateLookupV1.GetSigningRates(
+		e.ejectionCriteriaTimeWindow,
+		nil, // all quorums
+		ProtocolVersionV1,
+		true, // omit perfect signers if possible
```
litt3: "if possible"? What does that mean precisely?

ejector/ejector.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+
+	v2SigningRates, err := e.signingRateLookupV2.GetSigningRates(
+		e.ejectionCriteriaTimeWindow,
+		nil, // all quorums
+		ProtocolVersionV2,
+		true, // omit perfect signers if possible
+	)
+	if err != nil {
+		e.logger.Error("error looking up v2 signing rates", "error", err)
```
litt3: nit: it's a bit strange to have no return value for `evaluateValidators`, and log all errors internally. Why not just return an error, and log it in the main loop if non-nil?

Same comment applies to `evaluateValidator`

core/eth/validator_quorum_lookup.go
```diff
@@ -0,0 +1,124 @@
...
+		Context:     ctx,
+		BlockNumber: blockNumber,
+	}
+
+	// This method returns a bitmap as a big.Int, where the first 255 bits represent membership in quorums 0-254.
+	bigIntBitmap, err := v.registryCoordinator.GetCurrentQuorumBitmap(opts, validatorID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get quorum bitmap: %w", err)
+	}
+	bitmap := bigIntBitmap.Bytes()
```
litt3, line 71: The logic here doesn't check out for me, but maybe I'm missing something. I think bigint.Bytes returns a variable length byte slice, so it won't be any larger than necessary for the value stored there

So if a node is only a member of quorum 0, for example, I think indexing into the bitmap within the loop below is gonna end up panicking, since the size of `bitmap` won't be large enough

ejector/ejection_transactor.go
```diff
@@ -22,3 +31,223 @@ type EjectionTransactor interface {
 	// Complete the ejection proceedings against the operator with the given address.
 	CompleteEjection(ctx context.Context, addressToEject gethcommon.Address) error
 }
+
+var _ EjectionTransactor = &ejectionTransactor{}
+
+// ejectionTransactor is the production implementation of the EjectionTransactor interface.
+type ejectionTransactor struct {
```
litt3, line 41: Your docs don't mention thread safety- would be good to explicitly state for all structs

core/eth/validator_quorum_lookup.go
```diff
@@ -0,0 +1,124 @@
...
+	gethcommon "github.com/ethereum/go-ethereum/common"
+	lru "github.com/hashicorp/golang-lru/v2"
+)
+
+// A utility for looking up which quorums a given validator is a member of at a specific reference block number.
+type ValidatorQuorumLookup interface {
+	// Get the list of quorums that the given validator is a member of, at the specified reference block number.
+	GetQuorumsForValidator(
+		ctx context.Context,
+		validatorAddress core.OperatorID,
```
litt3, line 21: param is incorrectly named. the type says `ID`, but the name says `address`

ejector/ejector.go
```diff
@@ -0,0 +1,205 @@
...
+	e.logger.Info("Validator is eligible for ejection",
+		"validatorID", signingRate.GetValidatorId(),
+		"validatorAddress", validatorAddress.Hex(),
+		"signedBatches", signingRate.GetSignedBatches(),
+		"unsignedBatches", signingRate.GetUnsignedBatches(),
+		"signedBytes", signingRate.GetSignedBytes(),
+		"unsignedBytes", signingRate.GetUnsignedBytes(),
+	)
+
+	rbn, error := e.referenceBlockProvider.GetReferenceBlockNumber(e.ctx)
```
litt3:
```suggestion
	rbn, err := e.referenceBlockProvider.GetReferenceBlockNumber(e.ctx)
```

ejector/ejector.go
```diff
@@ -0,0 +1,205 @@
...
+
+	stakeFractions, err := getStakeFractionMap(validatorID, operatorState)
+	if err != nil {
+		e.logger.Error("error calculating stake fractions", "validatorID", signingRate.GetValidatorId(), "error", err)
+		return
+	}
+
+	// The ejection manager is responsible for deduplicating ejection requests, and deciding if
+	// there are other factors that may prevent ejection (e.g. too many ejection attempts, etc.).
+	_ = e.ejectionManager.EjectValidator(validatorAddress, stakeFractions)
```
litt3: why discard this error?

ejector/ejector.go
```diff
@@ -0,0 +1,205 @@
...
+	}
+}
+
+// evaluateValidator evaluates a single validator's signing rate and decides whether to eject it.
+func (e *Ejector) evaluateValidator(signingRate *validator.ValidatorSigningRate) {
+	if !IsEjectable(signingRate) {
+		return
+	}
+
+	validatorID := core.OperatorID(signingRate.GetValidatorId()[:])
```
litt3, line 150: looks like `signingRate.GetValidatorId()[:]` would panic if `GetValidatorId` returns `nil`, or if the length isn't `32` bytes

Are we sure that that could never happen?

core/eth/validator_quorum_lookup.go
```diff
@@ -0,0 +1,124 @@
...
+	}
+	bitmap := bigIntBitmap.Bytes()
+
+	quorumIDs := make([]core.QuorumID, 0)
+
+	// Although technically 254 is the max quorum ID (due to an embarrassing off-by-one typo), it doesn't hurt
+	// to check bit 255. It's possible that the typo will be fixed in the future, and this
+	// code should still work if that happens.
+	for i := 0; i <= math.MaxUint8; i++ {
+		byteIndex := i / 8
```
litt3, line 78: Bigint returns big endian, and Claude claims that your logic here is assuming little endian. I haven't taken the time to consider deeply, but his argument seems reasonable to me at first glance

core/eth/validator_quorum_lookup.go
```diff
@@ -0,0 +1,124 @@
...
+	registryCoordinator *regcoordinator.ContractEigenDARegistryCoordinator
+}
+
+// Create a new ValidatorQuorumLookup instance.
+func NewValidatorQuorumLookup(
+	backend bind.ContractBackend,
+	registryCoordinatorAddress gethcommon.Address,
+) (ValidatorQuorumLookup, error) {
+
+	return nil, nil
```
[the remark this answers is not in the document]
cody-littley: yikes, fixed

ejector/ejector_config.go
cody-littley:
```
	// The number of blocks to wait before using a reference block number for quorum. That is to say, do not always
	// use the latest block number we know about, but rather use the block number that is sufficiently old as to make
	// choosing the wrong fork unlikely. If this config value is X, and we know about block 100, we will use a RBN
	// 100-X.
	ReferenceBlockNumberOffset uint64
```

core/eth/validator_quorum_lookup.go
cody-littley: Not intentional, this was an oversight. Now fixed.

core/eth/validator_quorum_lookup.go
cody-littley: Screenshot 2025-10-17 at 2 18 31 PM

core/eth/validator_quorum_lookup.go
cody-littley: I wasn't aware of this... in any case, the big int type has enough bits for 256 total quorum IDs. No harm in making this code more permissive than it has to be.

core/eth/validator_quorum_lookup.go
cody-littley: Doh! Fixed.

ejector/ejection_criteria.go
cody-littley: removed utility funciton

ejector/main/main.go
cody-littley: Good callout, removed

ejector/ejector_config.go
cody-littley: How about this?

```
// The size for the caches for on-chain data.
	ChainDataCacheSize int
```

ejector/ejection_transactor.go
cody-littley: removed from this file

ejector/main/main.go
cody-littley: fixed

ejector/ejection_transactor.go
cody-littley: removed parameter

core/eth/reference_block_provider.go
cody-littley:
```	
if updatePeriod < 0 {
		return nil, fmt.Errorf("updatePeriod must be positive")
	}
```

IMO, makes sense to support 0, which basically means "look it up each time", unless we happen to look twice in the same nanosecond.

ejector/ejection_transactor.go
cody-littley: fixed

ejector/ejector.go
cody-littley: fixed

ejector/ejector.go
cody-littley: Omitting perfect signers is an optimization, but one we can't depend on. This works differently between v1 and v2 stuff.

```
// omit perfect signers if possible (data API has inconsistent behavior across v1 and v2)
```

ejector/ejector.go
cody-littley: fixed both

core/eth/validator_quorum_lookup.go
cody-littley: The base solidity function has a type `uint192`, so I'm assuming it's going to give me 192 bytes?

I honestly have no idea how this is going to work, as the API is completely undocumented. This is something I was planning on debugging via testing on preprod.

ejector/ejection_transactor.go
cody-littley:
```
// The ejection transactor is thread safe, although parallel calls may result in duplicate work (i.e. two calls might
// end up putting the same data in a cache).
```

core/eth/validator_quorum_lookup.go
cody-littley: fixed

ejector/ejector.go
cody-littley: fixed

ejector/ejector.go
cody-littley: odd, I can't remember why I did this. Fixed.

ejector/ejector.go
cody-littley: I think we can trust it, since it comes from things we control. But added a check just in case.

ejector/ejector_config.go
litt3: > using a reference block number for quorum

I guess my point is that I don't see how the `EjectorConfig` is for _a quorum_. Isn't this the some configuration for all quorums?

core/eth/validator_quorum_lookup.go
litt3: > so I'm assuming it's going to give me 192 bytes

My understanding is that the size of the output array depends on the value stored, and it will not be any larger than necessary

cody-littley:
Due to a merge gone wrong, this PR has been superseded by https://github.com/Layr-Labs/eigenda/pull/2075

PR #2076, opened 2025-09-25
chore: refactor makefile build/clean commands

samlaf:
What this refactor achieves is that running `make` in the root directory builds everything (protobufs, binaries, contracts, bindings).

build was always cleaning, which makes building slow, and makes it impossible to add build as a make pre-target (eg inabox).

Also root makefile build command was not building proxy.

And clean was not cleaning everything.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -54,7 +54,9 @@ clean:
 	$(MAKE) -C relay clean
 	$(MAKE) -C litt clean
 	$(MAKE) -C api/proxy clean
+	$(MAKE) -C contracts clean
 
+# Compiles the contracts and builds the golang bindings.
 contract-bindings:
 	$(MAKE) -C contracts bindings
 
--- a/contracts/Makefile
+++ b/contracts/Makefile
@@ -1,8 +1,9 @@
-deps:
-	yarn
-
 compile: deps
 	forge build
+
+# clean doesn't remove the bindings, because they are committed to the repo.
+clean:
+	forge clean
 
 # make bindings compiles the contracts and creates go bindings
 bindings: compile
@@ -14,3 +15,6 @@ fmt:
 
 fmt-check:
 	forge fmt --check
+
+deps:
+	yarn
--- a/node/Makefile
+++ b/node/Makefile
@@ -25,10 +25,6 @@ build-plugin: clean
 	go mod tidy
 	go build -o ./bin/node_plugin ./plugin/cmd
 
-proto:
-	cd .. && make protoc
-
-
 lint:
 	golangci-lint run
 


disperser/Makefile
```diff
@@ -65,9 +53,6 @@ run_encoder: build_encoder
   --disperser-encoder.log.level-std debug \
   --disperser-encoder.log.level-file debug
 
-run_dataapi_tests:
-	go test -v disperser/dataapi
-
 # You can override these defaults via CLI or environment variables
 run_blobapi: build_blobapi
```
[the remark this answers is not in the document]
samlaf: [b7b4ebb](https://github.com/Layr-Labs/eigenda/pull/2076/commits/b7b4ebbbe2935b79556232f48037ef580dd8e4c7)

disperser/Makefile
```diff
@@ -1,25 +1,13 @@
...
-build_controller:
-	go build -o ./bin/controller ./cmd/controller
+clean:
+	rm -rf ./bin
 
-build_blobapi:
-	go build -o ./bin/blobapi ./cmd/blobapi
+# Below are example run commands. They are not maintained so likely to be out of date.
 
 run_batcher: build_batcher
```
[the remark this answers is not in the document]
samlaf: [b7b4ebb](https://github.com/Layr-Labs/eigenda/pull/2076/commits/b7b4ebbbe2935b79556232f48037ef580dd8e4c7)

dmanc (approved):

pakim249CAL (approved):

disperser/Makefile
```diff
@@ -1,27 +1,15 @@
...
-	rm -rf ./bin
-
-build: build_server build_batcher build_encoder build_dataapi build_controller build_blobapi
-
-build_batcher:
-	go build -o ./bin/batcher ./cmd/batcher
-
-build_server:
+build:
+# We build the apiserver individually to change its name to "server" instead of "apiserver"
```
litt3, line 2: > change its name to "server" instead of "apiserver"

Curious why we care to do that?

relay/Makefile
```diff
@@ -1,14 +1,13 @@
 SHELL := /bin/bash
 
+# Build the light node.
```
litt3, line 3: "light node" 👀

litt3 (approved):

disperser/Makefile
samlaf: no idea... it was like this and there's a few dependencies on this name (dockerfiles, CI, probably some devops repo stuff) so decided against changing it.

There's actually a bunch of places where we name binaries "server" (like node/, relay/ etc). I don't really like it.
I think what would actually be neat is if we built all the binaries into `root/bin/` with their proper names. THen in scripts no need to cd into directories etc, all the binaries are in the same place.

relay/Makefile
samlaf: wth lol I didn't even notice... we should change this.

PR #2077, opened 2025-09-25
fix: use the router for live tests and traffic generator

cody-littley:

We should be exercising the router in the live test panel and traffic generator.

### DIFF
--- a/test/v2/client/test_client.go
+++ b/test/v2/client/test_client.go
@@ -18,7 +18,6 @@ import (
 	"github.com/Layr-Labs/eigenda/api/clients/v2/validator"
 	"github.com/Layr-Labs/eigenda/api/clients/v2/validator/mock"
 	"github.com/Layr-Labs/eigenda/api/clients/v2/verification"
-	"github.com/Layr-Labs/eigenda/api/clients/v2/verification/test"
 	proxycommon "github.com/Layr-Labs/eigenda/api/proxy/common"
 	proxyconfig "github.com/Layr-Labs/eigenda/api/proxy/config"
 	"github.com/Layr-Labs/eigenda/api/proxy/config/enablement"
@@ -56,7 +55,7 @@ type TestClient struct {
 	config                      *TestClientConfig
 	payloadClientConfig         *clientsv2.PayloadClientConfig
 	logger                      logging.Logger
-	certVerifierAddressProvider *test.TestCertVerifierAddressProvider
+	certVerifierAddressProvider clientsv2.CertVerifierAddressProvider
 	disperserClient             *clientsv2.DisperserClient
 	payloadDisperser            *payloaddispersal.PayloadDisperser
 	relayClient                 relay.RelayClient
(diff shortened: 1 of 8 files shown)


test/v2/live/live_network_test.go
```diff
@@ -298,28 +262,23 @@ func smallBlobDispersalAllQuorumsSetsTest(t *testing.T, environment string) {
...
+		err := testBasicDispersal(t, c, payload)
 		require.NoError(t, err)
 	})
 
-	t.Run("2", func(t *testing.T) {
-		checkAndSetCertVerifierAddress(t, c, config.EigenDACertVerifierAddressQuorums2)
-		err = testBasicDispersal(t, c, payload)
-		require.NoError(t, err)
-	})
+	// We need to eventually re-enable testing with quorum set {0,1,2} and {2}
```
litt3, line 272: Optionally, you could keep these tests, and make them simply use the raw `CertVerifier` addresses, just like they were doing before they got disabled. Presence of the `CertVerifierRouter` doesn't prevent usage of the `CertVerifier` directly, thought this would complicate the test logic.

Up to you

litt3 (approved):

test/v2/client/test_client.go
```diff
@@ -428,7 +435,7 @@ func NewTestClient(
 					VaultMonitorInterval:               time.Second * 30,
 					PutTries:                           3,
 					MaxBlobSizeBytes:                   16 * units.MiB,
-					EigenDACertVerifierOrRouterAddress: config.EigenDACertVerifierAddressQuorums0_1,
+					EigenDACertVerifierOrRouterAddress: routerAddress.Hex(),
```
dmanc, line 438: Do you know if the router outputs the cert verifier that's being used? Would be nice to be able to check in the logs that we're using the new version.

test/v2/live/live_network_test.go
```diff
@@ -298,28 +262,23 @@ func smallBlobDispersalAllQuorumsSetsTest(t *testing.T, environment string) {
...
+	//t.Run("0 1 2", func(t *testing.T) {
+	//	err := testBasicDispersal(t, c, payload)
+	//	require.NoError(t, err)
+	//})
+	//
+	//t.Run("2", func(t *testing.T) {
+	//	err := testBasicDispersal(t, c, payload)
+	//	require.NoError(t, err)
+	//})
```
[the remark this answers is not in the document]
cody-littley: nah

test/v2/live/live_network_test.go
cody-littley: I'd prefer to disable them for now, since I can't actually test them without an actual cert verifier.

test/v2/client/test_client.go
cody-littley: Yes. If we are running a proxy, it logs this:

```
	case err != nil && isExecutionReverted(err):
		log.Warnf("EigenDA cert verifier router address was detected to not be a router at address (%s), "+
			"using it as an immutable cert verifier instead", routerOrImmutableVerifierAddr.Hex())
		isRouter = false
	case err != nil:
		return nil, fmt.Errorf("failed to determine whether cert verifier is immutable or "+
			"deployed behind a router at address (%s) : %w", routerOrImmutableVerifierAddr.Hex(), err)
	default:
		log.Infof("EigenDA cert verifier address was detected as an EigenDACertVerifierRouter "+
			"at address (%s), using it as such", routerOrImmutableVerifierAddr.Hex())
	}
```

We should see these logs in the live test panel results. By default, the traffic generator doesn't use the proxy. But if we run in proxy mode, we should expect to see this log.

PR #2082, opened 2025-09-27
perf(fft): pre-allocate roots of unity slice

samlaf:
We know exactly how many roots of unity are going to be expanded, yet we were previously starting with a cap=2 buffer. Fixed this.

Also added a benchmark and test to make sure we only allocate the amount of memory necessary.

### DIFF
--- a/encoding/fft/fft.go
+++ b/encoding/fft/fft.go
@@ -27,6 +27,8 @@
 package fft
 
 import (
+	"fmt"
+
 	"github.com/Layr-Labs/eigenda/encoding"
 	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
 )
@@ -43,10 +45,13 @@ type FFTSettings struct {
 }
 
 // NewFFTSettings creates FFTSettings for a given maximum scale (log2 of max width).
+// Precomputes the roots of unity for all widths up to 2^maxScale.
+// Note that MaxWith is in units of Fr elements, so the actual byte size is 32 * MaxWidth.
+// In order to FFT a blob of size 16MiB, you thus need maxScale=19 (2^19 * 32 = 16MiB).
 func NewFFTSettings(maxScale uint8) *FFTSettings {
 	width := uint64(1) << maxScale
 	root := &encoding.Scale2RootOfUnity[maxScale]
-	rootz := expandRootOfUnity(&encoding.Scale2RootOfUnity[maxScale])
+	rootz := expandRootOfUnity(maxScale)
 
 	// reverse roots of unity
 	rootzReverse := make([]fr.Element, len(rootz))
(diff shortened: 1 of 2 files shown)


encoding/fft/fft.go
```diff
@@ -66,16 +71,19 @@ func NewFFTSettings(maxScale uint8) *FFTSettings {
...
 
-	for i := 1; !rootz[i].IsOne(); {
-		rootz = append(rootz, fr.Element{})
-		this := &rootz[i]
-		i++
-		rootz[i].Mul(this, rootOfUnity)
+	for i := 2; i < len(rootz); i++ {
+		rootz[i].Mul(&rootz[i-1], &rootOfUnity)
+	}
+	if rootz[len(rootz)-1].Cmp(new(fr.Element).SetOne()) != 0 {
```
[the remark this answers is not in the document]
samlaf: ehhh this is dumb

encoding/fft/fft.go
```diff
@@ -43,10 +45,13 @@ type FFTSettings struct {
 }
 
 // NewFFTSettings creates FFTSettings for a given maximum scale (log2 of max width).
+// Precomputes the roots of unity for all widths up to 2^maxScale.
```
bxue-l2, line 48: why it is called maxScale (suggesting there is no higher), why not just scale.

dmanc (approved):

encoding/fft/fft_test.go
```diff
@@ -0,0 +1,49 @@
...
+)
+
+const (
+	// Change this to benchmark different maxScales.
+	maxScale = uint8(22) // 2^22 * 32 = 128MiB
+)
+
+// BenchmarkFFTSettings benchmarks the creation of FFTSettings for a given maxScale.
+// This maxScale of 22 allows FFTs of up to 128MiB (2^22 * 32 bytes).
+// This in turn allows blobs of up to 16MiB, given that our RS encoding uses a 8x expansion
```
bxue-l2, line 17: I am a bit suspicious that the coding ratio is always 1/8. I don't remember well, does it fluctuate based on stake distribution. @Lulu-Zhou-EigenLabs

samlaf: coding rate is 8 for blob param = 0, see https://layr-labs.github.io/eigenda/protocol/architecture/security-parameters.html#blob-parameters

dmanc: This code is also used in V1 right? I believe the coding ratio is variable there but not sure if it's ever > 8

encoding/fft/fft.go
samlaf: idk I didn't change it it's always been called maxScale... prob because you can use this FFTSettings to fft blobs smaller than this scale? Is that true?

samlaf: I'll merge but we can discuss and change it later.

encoding/fft/fft_test.go
samlaf: Oh good point keep forgetting about god damn V1. I should have just forked all of encoding into a v2 folder...
Weren't we using 4.something in V1? I have no idea..

In any case all of this doesn't really matter for this benchmark/test tbh. The benchmark/test works for all scales.

PR #2083, opened 2025-09-28
feat(inabox): Spin up operator nodes as goroutines in inabox

dmanc:

This PR introduces `operator_setup.go` which is capable of setting up `N` operators with similar config (asides from ports). 

In addition this PR does some other stuff
- Context management (setup vs long-lived infrastructure)
- Node graceful shutdown / stop command for server (will make it a separate PR if it makes sense)
- Add some missing context done in node code (should be separate PR too)

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


core/thegraph/state_integration_test.go
```diff
@@ -70,25 +64,6 @@ func setupTest(t *testing.T) (
 	require.NoError(t, err, "failed to create Docker network")
 	logger.Info("Created Docker network", "name", nw.Name)
 
-	localstackContainer, err := testbed.NewLocalStackContainerWithOptions(ctx, testbed.LocalStackOptions{
-		ExposeHostPort: true,
```
samlaf, line 74: we were spinning up localstack but its not needed?

core/thegraph/state_integration_test.go
```diff
@@ -123,33 +98,45 @@ func setupTest(t *testing.T) (
...
 
-	logger.Info("Starting binaries")
-	testConfig.StartBinaries(true)
+	// Create a minimal InfrastructureHarness
+	infraHarness := &inaboxtests.InfrastructureHarness{
+		TestName:   testName,
+		TestConfig: testConfig,
+		Logger:     logger,
+		Ctx:        ctx,
+	}
```
samlaf: dont think we should ever be creating this by hand. dont we have a fct to deploy the infra? we should be calling that instead of spinning up things individually above.

core/thegraph/state_integration_test.go
```diff
@@ -123,33 +98,45 @@ func setupTest(t *testing.T) (
...
+
+	// For graph indexer test, we don't need churner functionality
+	churnerURL := ""
+
+	// Start operator nodes as goroutines using the inabox operator setup
+	logger.Info("Starting operator nodes as goroutines")
+	err = inaboxtests.StartOperatorsForInfrastructure(infraHarness, anvilContainerPort, churnerURL)
+	require.NoError(t, err, "failed to start operator nodes")
```
samlaf: are these only spun up for registration? Feel like we should have a separate fct that does that. can you add a todo

inabox/tests/churner_setup.go
```diff
@@ -23,7 +23,8 @@ import (
 
 // StartChurnerForInfrastructure starts the churner server as part of the global infrastructure.
 // This should be called after Anvil and other containers are started.
-func StartChurnerForInfrastructure(infra *InfrastructureHarness) error {
+// Returns the churner RPC address that can be used by operators.
+func StartChurnerForInfrastructure(infra *InfrastructureHarness, anvilRPC string) (string, error) {
```
samlaf: hmm why isnt anvilRPC part of infraHarness?

inabox/tests/setup_infra.go
```diff
@@ -53,8 +54,11 @@ func SetupGlobalInfrastructure(config *InfrastructureConfig) (*InfrastructureHar
 
 	testConfig := deploy.ReadTestConfig(testName, rootPath)
 
+	// Create a long-lived context for the infrastructure lifecycle
+	infraCtx, infraCancel := context.WithCancel(context.Background())
+
 	if testConfig.Environment.IsLocal() {
-		return setupLocalInfrastructure(ctx, config, testName, testConfig)
+		return setupLocalInfrastructure(setupCtx, infraCtx, infraCancel, config, testName, testConfig)
```
samlaf: this is getting a bit crazy. setup, infra, config, testconfig. do we really need all of these? i have no idea what each of these is doing when looking at this. maybe ok but justwant to mke sure theres no consolidation we could do

samlaf: also whats local vs global infra

inabox/tests/setup_infra.go
```diff
@@ -169,30 +188,47 @@ func setupLocalInfrastructure(
...
+	churnerURL, err := StartChurnerForInfrastructure(infra, anvilRPC)
 	if err != nil {
 		return nil, fmt.Errorf("failed to start churner server: %w", err)
 	}
-	infra.Logger.Info("Churner server started", "port", "32002")
+	infra.Logger.Info("Churner server started", "address", churnerURL)
+
+	// Start operator goroutines
+	infra.Logger.Info("Starting operator goroutines")
+	err = StartOperatorsForInfrastructure(infra, anvilRPC, churnerURL)
```
samlaf: think all binaries should be started in a different fct as infra. they are very different objects

inabox/tests/test_harness.go
```diff
@@ -32,8 +33,9 @@ type InfrastructureHarness struct {
 
 	// EigenDA components
 	// TODO: Should EigenDA components be their own test harness?
```
samlaf: yes they should imo. can be embedded in here if we so decide but imo should have their own namespace.

samlaf: hmm actually now come to think of it chain is shared but graphnode is eigenda disperser instance specific right?

samlaf: maybe worth a live chat

node/grpc/run.go
samlaf, line 1: node refactor should be separate pr imo

dmanc: Agreed, lets discuss in this PR https://github.com/Layr-Labs/eigenda/pull/2086

core/thegraph/state_integration_test.go
dmanc: Yeah graph node tests really just needs anvil, graph node, and operator registration.

node/grpc/run.go
dmanc: removed code here to make it independent of #2086 122cd1ee27f09babf5426f5b61d3bccb06675e2c

core/thegraph/state_integration_test.go
dmanc: Added todo 122cd1ee27f09babf5426f5b61d3bccb06675e2c

inabox/tests/churner_setup.go
dmanc: 3ac1591e10406913406d1a1cca423842f0ce2f18

core/thegraph/state_integration_test.go
dmanc: 786879bb39656a0d8cde509ee785a276a6d9fbe9

core/thegraph/state_integration_test.go
dmanc: Added a comment in 786879bb39656a0d8cde509ee785a276a6d9fbe9

inabox/tests/setup_infra.go
dmanc: Let's discuss it sync

core/thegraph/state_integration_test.go
```diff
@@ -41,115 +32,46 @@ func init() {
...
+	// Start all the necessary infrastructure like anvil, graph node, and eigenda components
+	// TODO(dmanc): We really only need to register operators on chain, maybe add some sort of
+	// configuration to allow that mode.
+	infraHarness, err := inaboxtests.SetupInfrastructure(config)
+	require.NoError(t, err, "failed to setup global infrastructure")
 
-	logger.Info("Starting binaries")
-	testConfig.StartBinaries(true)
+	// Update the graph URL to use the container from infrastructure
+	if infraHarness.ChainHarness.GraphNode != nil {
```
samlaf, line 59: why != nil check? what if its nil? is that ok?

inabox/tests/test_harness.go
```diff
@@ -14,43 +14,42 @@ import (
...
 // InfrastructureHarness contains the shared infrastructure components
 // that are global across all tests (external dependencies)
 type InfrastructureHarness struct {
-	// Infrastructure containers - truly global
-	AnvilContainer      *testbed.AnvilContainer
-	GraphNodeContainer  *testbed.GraphNodeContainer
-	LocalstackContainer *testbed.LocalStackContainer
-	ChainDockerNetwork  *testcontainers.DockerNetwork
+	// Shared docker network
+	SharedNetwork *testcontainers.DockerNetwork
```
samlaf, line 26: only shared by chain and graph though right? prob worth documenting

PR #2086, opened 2025-09-29
feat(validator): Validator graceful shutdown

dmanc:

Adds graceful shutdown for validator node along with improvements to logging and context management.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/node/grpc/run.go
+++ b/node/grpc/run.go
@@ -104,7 +104,7 @@ func RunServers(serverV1 *Server, serverV2 *ServerV2, config *node.Config, logge
 				healthcheck.RegisterHealthServer("node.Dispersal", gs)
 
 				logger.Info("v1 dispersal enabled on port", config.InternalDispersalPort, "address", listener.Addr().String(), "GRPC Listening")
-				if err := gs.Serve(listener); err != nil && err != grpc.ErrServerStopped {
+				if err := gs.Serve(listener); err != nil && !errors.Is(err, grpc.ErrServerStopped) {
 					select {
 					case <-runner.ctx.Done():
 						logger.Info("v1 dispersal server stopping due to context cancellation")
@@ -153,7 +153,7 @@ func RunServers(serverV1 *Server, serverV2 *ServerV2, config *node.Config, logge
 				healthcheck.RegisterHealthServer("node.v2.Dispersal", gs)
 
 				logger.Info("v2 dispersal enabled on port", config.InternalV2DispersalPort, "address", listener.Addr().String(), "GRPC Listening")
-				if err := gs.Serve(listener); err != nil && err != grpc.ErrServerStopped {
+				if err := gs.Serve(listener); err != nil && !errors.Is(err, grpc.ErrServerStopped) {
 					select {
 					case <-runner.ctx.Done():
 						logger.Info("v2 dispersal server stopping due to context cancellation")
@@ -201,7 +201,7 @@ func RunServers(serverV1 *Server, serverV2 *ServerV2, config *node.Config, logge
 				healthcheck.RegisterHealthServer("node.Retrieval", gs)
 
 				logger.Info("v1 retrieval enabled on port", config.InternalRetrievalPort, "address", listener.Addr().String(), "GRPC Listening")
-				if err := gs.Serve(listener); err != nil && err != grpc.ErrServerStopped {
+				if err := gs.Serve(listener); err != nil && !errors.Is(err, grpc.ErrServerStopped) {
 					select {
 					case <-runner.ctx.Done():
 						logger.Info("v1 retrieval server stopping due to context cancellation")
@@ -249,7 +249,7 @@ func RunServers(serverV1 *Server, serverV2 *ServerV2, config *node.Config, logge
 				healthcheck.RegisterHealthServer("node.v2.Retrieval", gs)
 
 				logger.Info("v2 retrieval enabled on port", config.InternalV2RetrievalPort, "address", listener.Addr().String(), "GRPC Listening")
-				if err := gs.Serve(listener); err != nil && err != grpc.ErrServerStopped {
+				if err := gs.Serve(listener); err != nil && !errors.Is(err, grpc.ErrServerStopped) {
 					select {
 					case <-runner.ctx.Done():
 						logger.Info("v2 retrieval server stopping due to context cancellation")


node/cmd/main.go
```diff
@@ -114,9 +114,13 @@ func NodeMain(ctx *cli.Context, softwareVersion *version.Semver) error {
 		return fmt.Errorf("failed to get ServiceManager address: %w", err)
 	}
 
+	// Create a cancellable context for the node
+	ctx, cancel := context.WithCancel(context.Background())
```
samlaf, line 118: cliCtx is a context itself, that you could use instead of context.Background. I'm honestly not sure whether its got anything registered on it though. Perhaps it already catches ctrl-c SIGINT signals?

node/cmd/main.go
```diff
@@ -127,6 +131,7 @@ func NodeMain(ctx *cli.Context, softwareVersion *version.Semver) error {
 	if err != nil {
 		return err
 	}
+	defer node.Shutdown() // Ensure node is always properly shut down (idempotent)
```
samlaf, line 134: I was talking with @litt3 about this on another of his PRs. For me it doesn't really make sense to pass in a cancellable context AND to have a node shutdown. I feel like we should choose one way and stick to it.

node.Shutdown typically just internally cancels the internal context it saved.

node/cmd/main.go
```diff
@@ -156,7 +161,35 @@ func NodeMain(ctx *cli.Context, softwareVersion *version.Semver) error {
...
+	sigChan := make(chan os.Signal, 1)
+	signal.Notify(sigChan, syscall.SIGINT, syscall.SIGTERM)
+
+	go func() {
+		sig := <-sigChan
+		logger.Infof("Received signal %v, initiating graceful shutdown", sig)
+
+		// Shutdown in correct order
+		runner.Stop()
+		node.Shutdown()
```
samlaf, line 183: why do we need these 2 things? why doesn't the node hold the grpc servers and node.Shutdown stops the servers?

node/grpc/run.go
```diff
@@ -16,138 +18,249 @@ import (
...
-func RunServers(serverV1 *Server, serverV2 *ServerV2, config *node.Config, logger logging.Logger) error {
+// Runner manages the lifecycle of gRPC servers
+type Runner struct {
+	servers []*grpc.Server
+	logger  logging.Logger
+	ctx     context.Context
+	cancel  context.CancelFunc
+	wg      sync.WaitGroup
+	mu      sync.Mutex
+}
```
samlaf, line 29: curious: why do we need this instead of this being inside Node itself?

test/v1/integration_test.go
```diff
@@ -159,15 +159,26 @@ func TestDispersalAndRetrieval(t *testing.T) {
 
 	gethClient, _ := mustMakeRetriever()
 
+	var runners []*nodegrpc.Runner
 	for _, op := range ops {
```
samlaf, line 163: what is ops here? operators?

node/node.go
```diff
@@ -1137,59 +1184,63 @@ func (n *Node) checkNodeReachability(checkPath string) {
...
+				}
+
+				if len(v2ResponseObject.Operators) > 0 {
+					// Process the first operator from the array
+					n.processReachabilityResponse(version, v2ResponseObject.Operators[0])
+				} else {
+					n.Logger.Error("Reachability check v2 returned empty operators array")
+				}
 			}
+		case <-n.CTX.Done():
```
samlaf, line 1241: why caps lock CTX?

node/node.go
```diff
@@ -978,6 +984,45 @@ func (n *Node) ProcessBatch(
 	return signature, nil
 }
 
+// Shutdown gracefully shuts down the node and its components
+func (n *Node) Shutdown() {
+	n.shutdownOnce.Do(func() {
```
samlaf, line 989: do we expect Shutdown to be called concurrently from multiple goroutines??
If not, maybe just keep it simple and use a boolean? Or is there really an issue with calling Shutdown more than once? Seems like everything is check if != nil?

node/cmd/main.go
cody-littley: > cliCtx is a context itself

Is it? If you try to store that in a `context.Context`, the compiler says the types are incompatible.

node/grpc/run.go
```diff
@@ -16,138 +18,249 @@ import (
...
+				default:
+				}
+
+				addr := fmt.Sprintf("%s:%s", localhost, config.InternalDispersalPort)
+				listener, err := net.Listen("tcp", addr)
+				if err != nil {
+					logger.Fatalf("Could not start tcp listener: %v", err)
+				}
+
+				opt := grpc.MaxRecvMsgSize(60 * 1024 * 1024 * 1024) // 60 GiB
```
cody-littley, line 90: See offline slack discussion

node/grpc/run.go
```diff
@@ -16,138 +18,249 @@ import (
...
+
+				addr := fmt.Sprintf("%s:%s", localhost, config.InternalRetrievalPort)
+				listener, err := net.Listen("tcp", addr)
+				if err != nil {
+					logger.Fatalf("Could not start tcp listener: %v", err)
+				}
 
-			opt := grpc.MaxRecvMsgSize(1024 * 1024 * 300) // 300 MiB
-			gs := grpc.NewServer(opt)
+				opt := grpc.MaxRecvMsgSize(1024 * 1024 * 300) // 300 MiB
```
cody-littley, line 188: see offline thread

node/node.go
```diff
@@ -978,6 +984,45 @@ func (n *Node) ProcessBatch(
 	return signature, nil
 }
 
+// Shutdown gracefully shuts down the node and its components
+func (n *Node) Shutdown() {
```
cody-littley, line 988: This method should also shut down LittDB (`db.Close()`). The LittDB instances is embedded inside the `ValidatorStore`.

Although not strictly necessary from a correctness point of view, shutting down the DB allows it to fully flush data to disk, which means it doesn't need to do recovery operations at the next startup. A hard shutdown can result in partially corrupted files (recoverable), and could conceivably take nontrivial time for LittDB to parse and fix the problem.

When doing this, important to wait for all threads that might be writing to the DB to be stopped. If you stop the DB then write to it, the DB will clutter the log with nasty messages.

node/node.go
```diff
@@ -978,6 +984,45 @@ func (n *Node) ProcessBatch(
...
+
+		// Stop Node API
+		if n.NodeApi != nil && n.Config.EnableNodeApi {
+			// TODO: The NodeApi struct doesn't have a Stop method currently,
+			// but we should add one if the API server needs cleanup
+			n.Logger.Info("Node API shutdown complete")
+		}
+
+		// Close the store if it has a close method
+		if n.Store != nil {
```
cody-littley, line 1017: `Store` contains v1 data, so adding a close won't here won't shut down LittDB.

node/node.go
cody-littley: The `Node` struct has some severe style problems. All of its variables are publicly exported, and unit tests in other packages directly access pretty much all of the fields willy-nilly. We are aware of the issue, but there is a lot of tech debt we need to wade through before we can fix this.

node/cmd/main.go
dmanc: might be an urfave/cli v2 feature?

samlaf: oh I see... interesting. Indeed looks like only the v2 cli.COntext is an actual context

node/node.go
samlaf: kek

PR #2089, opened 2025-09-29
feat(contracts): use address directory in registry coordinator

pakim249CAL:

Use the address directory in the registry coordinator

https://linear.app/eigenlabs/issue/EGDA-2075/change-references-to-address-directory-in-registry-coordinator

Also changes one of the scripts to use a directory, as it wasn't before and was blocking this change.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


contracts/script/deploy/eigenda/DeployEigenDA.s.sol
```diff
@@ -249,55 +294,32 @@ contract DeployEigenDA is Script {
...
     }
 
     function upgrade(string memory contractName, bytes memory initData) internal {
         require(!upgraded[contractName], string.concat("Contract already upgraded: ", contractName));
-        require(deployed[contractName] != address(0), string.concat("Contract not deployed: ", contractName));
 
-        ProxyAdmin proxyAdmin = ProxyAdmin(deployed[PROXY_ADMIN]);
         address implementation = impl[contractName];
-        TransparentUpgradeableProxy proxy = TransparentUpgradeableProxy(payable(deployed[contractName]));
+        TransparentUpgradeableProxy proxy = TransparentUpgradeableProxy(payable(directory.getAddress(contractName)));
```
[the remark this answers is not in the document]
pakim249CAL: Fixed in this [commit](https://github.com/Layr-Labs/eigenda/pull/2089/commits/5c71ddefd59ab5a17bfd0242228656dbe4467e9a) by simply not using names not in the constants.

contracts/src/core/EigenDARegistryCoordinatorStorage.sol
```diff
@@ -63,8 +55,7 @@ abstract contract EigenDARegistryCoordinatorStorage is IRegistryCoordinator {
     /// @notice mapping from quorum number to the latest block that all quorums were updated all at once
     mapping(uint8 => uint256) public quorumUpdateBlockNumber;
 
-    /// @notice the dynamic-length array of the registries this coordinator is coordinating
-    address[] public registries;
+    address[] private _deprecated_2;
```
samlaf, line 58: just to be sure, dynamic arrays like this get written to the storage slot keccak(_deprecated_2 SLOT) or something like this right? And we're sure nothing else will ever be written/read from there?

contracts/src/core/EigenDARegistryCoordinator.sol
```diff
@@ -827,4 +830,24 @@ contract EigenDARegistryCoordinator is
...
+        return IServiceManager(directory.getAddress(AddressDirectoryConstants.SERVICE_MANAGER_NAME.getKey()));
+    }
+
+    function blsApkRegistry() public view returns (IBLSApkRegistry) {
+        return IBLSApkRegistry(directory.getAddress(AddressDirectoryConstants.BLS_APK_REGISTRY_NAME.getKey()));
+    }
+
+    function stakeRegistry() public view returns (IStakeRegistry) {
+        return IStakeRegistry(directory.getAddress(AddressDirectoryConstants.STAKE_REGISTRY_NAME.getKey()));
+    }
```
samlaf, line 847: do we need these to be public? Is this only for backwards compatibility?
If we really want to lean in and use directory everywhere then we prob would want to completely get rid of these public functions and make them internal no?

contracts/src/core/EigenDARegistryCoordinator.sol
```diff
@@ -800,9 +796,16 @@ contract EigenDARegistryCoordinator is
...
+    /// @dev DEPRECATED. Use the address directory instead.
+    function registries(uint256) external pure returns (address) {
+        return address(0);
+    }
+
     /// @notice Returns the number of registries
-    function numRegistries() external view returns (uint256) {
-        return registries.length;
+    /// @dev DEPRECATED. Use the address directory instead.
+    function numRegistries() external pure returns (uint256) {
```
samlaf, line 807: just to be sure, do we know how this function was used? Was it purely for info/debugging? Or were any contracts / offchain processes relying on this?

contracts/src/core/EigenDARegistryCoordinator.sol
```diff
@@ -699,7 +695,7 @@ contract EigenDARegistryCoordinator is
     }
 
     function _setOperatorSocket(bytes32 operatorId, string memory socket) internal {
-        socketRegistry.setOperatorSocket(operatorId, socket);
+        socketRegistry().setOperatorSocket(operatorId, socket);
         emit OperatorSocketUpdate(operatorId, socket);
     }
```
samlaf, line 700: curious for your thoughts as to this function. Would you prefer if we didn't have this setOperatorSocket in the coordinator and instead have operators talk to the socketRegistry directly (given that we've made that easy with the directory)?

contracts/src/core/EigenDARegistryCoordinatorStorage.sol
pakim249CAL: Dynamic arrays are structured in a similar way to a mapping. In fact, a dynamic array is actually a mapping itself, but also just keeps track of the length. 

There is certainly data that is already written here. The purpose of marking this variable as deprecated is actually to achieve what you state, which is to make sure that nothing is ever written/read from this array in the future.

contracts/src/core/EigenDARegistryCoordinator.sol
pakim249CAL: The problem is that these functions exist already, and so removing "public" would essentially make this upgrade NOT backwards compatible.

Note that in the storage contract, these external functions were actually included in the contract before, because the storage variables are labelled public. This automatically creates a getter for that particular variable, in a similar manner to what I've defined here.

contracts/src/core/EigenDARegistryCoordinator.sol
pakim249CAL: It was purely for info/debugging. None of our contracts ever find a registry by "index" or ever check for "number of registries" AFAIK. I think this is safe to say. And if we ever do, then I would like to see that code obliterated.

contracts/src/core/EigenDARegistryCoordinator.sol
pakim249CAL: Yes. In general, I would like to see the RegistryCoordinator do less. We can perhaps do those kinds of changes in a future improvement.

contracts/src/core/EigenDARegistryCoordinator.sol
samlaf: same lol

contracts/src/core/EigenDARegistryCoordinator.sol
samlaf: Right. I was just saying might be good to add a comment that these are deprecated and we plan to eventually get rid of them after all offchain code moves to using EigenDADirectory.

PR #2095, opened 2025-09-30
fix(payments): Use proper `ReservationLedger` constructor

litt3:
- Old code was directly instantiating new config struct
- It was missing an important field
- The fix is to use the constructor, instead of raw struct instantiation

### DIFF
--- a/core/payments/reservation/reservation_ledger.go
+++ b/core/payments/reservation/reservation_ledger.go
@@ -157,11 +157,14 @@ func (rl *ReservationLedger) UpdateReservation(newReservation *Reservation, now 
 	}
 
 	// Create new config with the updated reservation
-	newConfig := ReservationLedgerConfig{
-		reservation:            *newReservation,
-		startFull:              rl.config.startFull,
-		overfillBehavior:       rl.config.overfillBehavior,
-		bucketCapacityDuration: rl.config.bucketCapacityDuration,
+	newConfig, err := NewReservationLedgerConfig(
+		*newReservation,
+		rl.config.minNumSymbols,
+		rl.config.startFull,
+		rl.config.overfillBehavior,
+		rl.config.bucketCapacityDuration)
+	if err != nil {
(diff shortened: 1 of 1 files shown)


core/payments/reservation/reservation_ledger.go
```diff
@@ -157,11 +157,14 @@ func (rl *ReservationLedger) UpdateReservation(newReservation *Reservation, now
...
-		reservation:            *newReservation,
-		startFull:              rl.config.startFull,
-		overfillBehavior:       rl.config.overfillBehavior,
-		bucketCapacityDuration: rl.config.bucketCapacityDuration,
+	newConfig, err := NewReservationLedgerConfig(
+		*newReservation,
+		rl.config.minNumSymbols,
+		rl.config.startFull,
+		rl.config.overfillBehavior,
+		rl.config.bucketCapacityDuration)
```
samlaf, line 165: Def not against merging this, just brainstorming for fun, and also b/c I generally dont like config constructors. Mostly coming from a perspective of writing public libraries, because config constructors break clients whenever they need to be upudated, which is no bueno. For internal stuff its less bad, but I still generally dislike them (though open to being convinced otherwise).

would it make sense instead to have NewReservationLedgerConfigFromOld or something that just takes another LedgerConfig, copies it, and replaces the reservation with another?
That seems to be the issue here, you are trying to copy another reservationLedger, not create a new one from scratch?

Or you could do something like:
```
newConfig := rl.config
newConfig.reservation = *newReservation
```
?

litt3:
```
newConfig := rl.config
newConfig.reservation = *newReservation
```

I considered this alternative, but decided against it because I really don't like mutating config structs. Config structs _should_ be immutable, IMO. If golang would give us to tools to accomplish it, all the ones I write _would_ be.

litt3: > NewReservationLedgerConfigFromOld

This seems more complicated than it needs to be, IMO

PR #2096, opened 2025-09-30
feat(inabox): Deploy relays as goroutines inside inabox integration tests

dmanc:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/inabox/tests/integration_suite_test.go
+++ b/inabox/tests/integration_suite_test.go
@@ -64,6 +64,12 @@ func setupSuite(logger logging.Logger) error {
 		TestName:          testName,
 		InMemoryBlobStore: inMemoryBlobStore,
 		Logger:            logger,
+		RelayURLs: []string{
+			"localhost:32035",
+			"localhost:32037",
+			"localhost:32039",
+			"localhost:32041",
+		},
 	}
 	var err error
 	globalInfra, err = integration.SetupInfrastructure(config)
--- a/inabox/tests/setup_disperser_harness.go
+++ b/inabox/tests/setup_disperser_harness.go
@@ -22,6 +22,7 @@ type DisperserHarnessConfig struct {
 	BucketTableName     string
 	MetadataTableNameV2 string
 	EthClient           common.EthClient
+	RelayURLs           []string // URLs to register for relays
 }
 
 // TODO: Add encoder, api server, relay, controller, batcher
@@ -81,9 +82,15 @@ func setupDisperserKeypairAndRegistrations(config *DisperserHarnessConfig) error
 		return fmt.Errorf("failed to generate disperser keypair: %w", err)
 	}
 
-	// Register blob versions, relays, and disperser keypair
+	// Register disperser keypair and relays
 	if config.TestConfig.EigenDA.Deployer != "" && config.TestConfig.IsEigenDADeployed() {
 		config.TestConfig.PerformDisperserRegistrations(config.EthClient)
+
+		// Register relay URLs if provided
+		if len(config.RelayURLs) > 0 {
+			config.Logger.Info("Registering relay URLs", "count", len(config.RelayURLs))
+			config.TestConfig.RegisterRelays(config.EthClient, config.RelayURLs, config.EthClient.GetAccountAddress())
+		}
 	}
 
 	return nil
--- a/inabox/tests/setup_disperser_harness.go
+++ b/inabox/tests/setup_disperser_harness.go
@@ -3,11 +3,27 @@ package integration
 import (
 	"context"
 	"fmt"
+	"io"
+	"log/slog"
+	"net"
+	"os"
+	"strings"
+	"time"
 
 	"github.com/Layr-Labs/eigenda/common"
+	"github.com/Layr-Labs/eigenda/common/aws/dynamodb"
+	"github.com/Layr-Labs/eigenda/common/aws/s3"
+	coreeth "github.com/Layr-Labs/eigenda/core/eth"
+	"github.com/Layr-Labs/eigenda/core/thegraph"
+	v2 "github.com/Layr-Labs/eigenda/core/v2"
+	"github.com/Layr-Labs/eigenda/disperser/common/v2/blobstore"
 	"github.com/Layr-Labs/eigenda/inabox/deploy"
+	"github.com/Layr-Labs/eigenda/relay"
+	"github.com/Layr-Labs/eigenda/relay/chunkstore"
+	"github.com/Layr-Labs/eigenda/relay/limiter"
 	"github.com/Layr-Labs/eigenda/test/testbed"
 	"github.com/Layr-Labs/eigensdk-go/logging"
+	"github.com/prometheus/client_golang/prometheus"
 	"github.com/testcontainers/testcontainers-go"
 )
 
@@ -16,16 +32,19 @@ type DisperserHarnessConfig struct {
 	Logger              logging.Logger
 	Network             *testcontainers.DockerNetwork
 	TestConfig          *deploy.Config
+	TestName            string
 	InMemoryBlobStore   bool
 	LocalStackPort      string
 	MetadataTableName   string
 	BucketTableName     string
+	S3BucketName        string // S3 bucket name for blob storage
 	MetadataTableNameV2 string
 	EthClient           common.EthClient
 	RelayURLs           []string // URLs to register for relays
+	infraCtx            context.Context
 }
 
-// TODO: Add encoder, api server, relay, controller, batcher
+// TODO: Add encoder, api server, controller, batcher
 type DisperserHarness struct {
 	LocalStack     *testbed.LocalStackContainer
 	DynamoDBTables struct {
@@ -35,11 +54,12 @@ type DisperserHarness struct {
 	S3Buckets struct {
 		BlobStore string
 	}
+	RelayInstances []*RelayInstance
 }
 
 // setupLocalStackResources initializes LocalStack and deploys AWS resources
 func setupLocalStackResources(
-	ctx context.Context, config *DisperserHarnessConfig,
+	ctx context.Context, config DisperserHarnessConfig,
 ) (*testbed.LocalStackContainer, error) {
 	config.Logger.Info("Setting up LocalStack for blob store")
 	localstackContainer, err := testbed.NewLocalStackContainerWithOptions(
@@ -60,6 +80,7 @@ func setupLocalStackResources(
 		LocalStackEndpoint:  fmt.Sprintf("http://0.0.0.0:%s", config.LocalStackPort),
 		MetadataTableName:   config.MetadataTableName,
 		BucketTableName:     config.BucketTableName,
+		BucketName:          config.S3BucketName,
 		V2MetadataTableName: config.MetadataTableNameV2,
 		Logger:              config.Logger,
 	}
@@ -72,7 +93,7 @@ func setupLocalStackResources(
 }
 
 // setupDisperserKeypairAndRegistrations generates disperser keypair and performs registrations
-func setupDisperserKeypairAndRegistrations(config *DisperserHarnessConfig) error {
+func setupDisperserKeypairAndRegistrations(config DisperserHarnessConfig) error {
 	if config.TestConfig == nil {
 		return nil
 	}
@@ -96,9 +117,11 @@ func setupDisperserKeypairAndRegistrations(config *DisperserHarnessConfig) error
 	return nil
 }
 
-// SetupDisperserHarness creates and initializes the disperser infrastructure (LocalStack, DynamoDB tables, S3 buckets)
-func SetupDisperserHarness(ctx context.Context, config *DisperserHarnessConfig) (*DisperserHarness, error) {
-	harness := &DisperserHarness{}
+// SetupDisperserHarness creates and initializes the disperser infrastructure (LocalStack, DynamoDB tables, S3 buckets, relays)
+func SetupDisperserHarness(ctx context.Context, config DisperserHarnessConfig) (*DisperserHarness, error) {
+	harness := &DisperserHarness{
+		RelayInstances: make([]*RelayInstance, 0),
+	}
 
 	// Set default values if not provided
 	if config.LocalStackPort == "" {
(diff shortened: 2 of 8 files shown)


inabox/tests/integration_suite_test.go
```diff
@@ -64,6 +64,12 @@ func setupSuite(logger logging.Logger) error {
 		TestName:          testName,
 		InMemoryBlobStore: inMemoryBlobStore,
 		Logger:            logger,
+		RelayURLs: []string{
+			"localhost:32035",
+			"localhost:32037",
+			"localhost:32039",
+			"localhost:32041",
```
samlaf: why is it using odd numbers only?

inabox/tests/setup_infra.go
```diff
@@ -18,7 +18,9 @@ type InfrastructureConfig struct {
 	Logger              logging.Logger
 	MetadataTableName   string
 	BucketTableName     string
+	S3BucketName        string
 	MetadataTableNameV2 string
+	RelayURLs           []string
```
samlaf: Id much prefer if this wasn't here. Users should not be expected to configured relayURLs. Nobody should have to care about this. We should move to these being os generated asap. But in the meantime we can prob just hardcode them somewhere?

Or is there idea that different tests spin up different number of relays?

inabox/tests/setup_infra.go
```diff
@@ -104,41 +106,45 @@ func SetupInfrastructure(config *InfrastructureConfig) (*InfrastructureHarness,
...
 		return nil, setupErr
 	}
 	infra.DisperserHarness = *disperserHarness
 
+	// Setup Operator Harness third (requires chain and disperser to be ready)
+	operatorHarnessConfig := &OperatorHarnessConfig{
+		TestConfig:   testConfig,
+		TestName:     testName,
+		Logger:       logger,
+		ChainHarness: &infra.ChainHarness,
```
samlaf: I wouldn't put this here. I'd just pass it as a separate argument to SetupOperatorHarness. It's not configuration

inabox/tests/setup_infra.go
```diff
@@ -104,41 +106,45 @@ func SetupInfrastructure(config *InfrastructureConfig) (*InfrastructureHarness,
...
+		Ctx:          infraCtx,
```
samlaf: infraCtx also doesn't belong in a config struct. Why is this passed here at all actually?

test/testbed/deploy_localstack_resources.go
```diff
@@ -25,6 +25,7 @@ type DeployResourcesConfig struct {
 	LocalStackEndpoint  string
 	MetadataTableName   string
 	BucketTableName     string
+	BucketName          string // Optional: S3 bucket name, defaults to "test-eigenda-blobstore"
 	V2MetadataTableName string
 	V2PaymentPrefix     string         // Optional: prefix for v2 payment tables, defaults to "e2e_v2_"
 	Region              string         // Optional: AWS region, defaults to "us-east-1"
```
samlaf: nit: prefer comments above the field, since they appear when you hover over the config type in IDEs.
These comments on the side don't

test/testbed/deploy_subgraphs.go
```diff
@@ -166,6 +166,16 @@ func deploySubgraph(config SubgraphDeploymentConfig, updater SubgraphUpdater, pa
...
+	// Save the current directory so we can restore it later
+	originalDir, err := os.Getwd()
+	if err != nil {
+		return fmt.Errorf("failed to get current directory: %w", err)
+	}
+	// Ensure we restore the directory when we're done
+	defer func() {
+		_ = os.Chdir(originalDir)
+	}()
+
```
samlaf: instead of doing this can we just get rid of the `os.Chdir` call below and instead use `path.join` for the execBashCmd call sites? We shouldn't be changing directory like this.

test/testbed/localstack.go
```diff
@@ -144,13 +145,13 @@ func (ls *LocalStackContainer) GetServiceEndpoint(service string) string {
...
-		"AWS_SECRET_ACCESS_KEY": "test",
-		"AWS_DEFAULT_REGION":    ls.options.Region,
-		"AWS_ENDPOINT_URL":      ls.Endpoint(),
+// GetAWSClientConfig returns AWS client configuration for connecting to LocalStack
+func (ls *LocalStackContainer) GetAWSClientConfig() aws.ClientConfig {
+	return aws.ClientConfig{
+		Region:          ls.options.Region,
+		EndpointURL:     ls.Endpoint(),
+		AccessKey:       "test",
+		SecretAccessKey: "test",
```
samlaf, line 154: good improvement

inabox/tests/setup_disperser_harness.go
```diff
@@ -81,17 +103,26 @@ func setupDisperserKeypairAndRegistrations(config *DisperserHarnessConfig) error
...
-	// Register blob versions, relays, and disperser keypair
+	// Register disperser keypair and relays
 	if config.TestConfig.EigenDA.Deployer != "" && config.TestConfig.IsEigenDADeployed() {
 		config.TestConfig.PerformDisperserRegistrations(config.EthClient)
+
+		// Register relay URLs if provided
+		if len(config.RelayURLs) > 0 {
+			config.Logger.Info("Registering relay URLs", "count", len(config.RelayURLs))
+			config.TestConfig.RegisterRelays(config.EthClient, config.RelayURLs, config.EthClient.GetAccountAddress())
+		}
```
samlaf: if not provided what happens...? shouldn't we panic?

inabox/tests/integration_suite_test.go
samlaf: also I see that these relayURLs are used to determine how many relayURLs to spin up. If a nil array is passed nothing is spun up. Should be documented as the very least, though I'd personally prefer if we find a better way to configure this.. like just specifying number of relays to spin up

inabox/tests/setup_disperser_harness.go
```diff
@@ -141,12 +183,255 @@ func SetupDisperserHarness(ctx context.Context, config *DisperserHarnessConfig)
...
 }
 
+// RelayInstance holds the state for a single relay
+type RelayInstance struct {
+	Server   *relay.Server
+	Listener net.Listener
+	Port     string
+	URL      string
+	Logger   logging.Logger
+}
```
samlaf, line 191: gosh we did this for the relay as well? :(
This struct should live in the relay package itself, not here.

inabox/tests/setup_disperser_harness.go
```diff
@@ -141,12 +183,255 @@ func SetupDisperserHarness(ctx context.Context, config *DisperserHarnessConfig)
...
+	logFilePath := fmt.Sprintf("%s/relay_%d.log", logsDir, relayIndex)
+	logFile, err := os.OpenFile(logFilePath, os.O_CREATE|os.O_APPEND|os.O_WRONLY, 0644)
+	if err != nil {
+		return nil, fmt.Errorf("failed to open relay log file: %w", err)
+	}
+
+	// Create relay logger
+	loggerConfig := common.LoggerConfig{
+		Format:       common.TextLogFormat,
+		OutputWriter: io.MultiWriter(os.Stdout, logFile),
```
samlaf, line 293: eventually we should refactor this to have a centralized place where all the logs are generated and held. This is how op's framework works. There's a map of component name to logger instance in the testHarness.

inabox/tests/setup_disperser_harness.go
```diff
@@ -141,12 +183,255 @@ func SetupDisperserHarness(ctx context.Context, config *DisperserHarnessConfig)
...
+	s3Client, err := s3.NewClient(config.infraCtx, awsConfig, relayLogger)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create s3 client: %w", err)
+	}
+
+	// Test S3 connection by listing objects (works even if bucket is empty)
+	relayLogger.Info("Testing S3 connection", "bucket", harness.S3Buckets.BlobStore)
+	testCtx2, testCancel2 := context.WithTimeout(context.Background(), 30*time.Second)
+	defer testCancel2()
+	if _, err := s3Client.ListObjects(testCtx2, harness.S3Buckets.BlobStore, ""); err != nil {
```
samlaf: why do we have all these tests here? Aren't these things the relay itself would want to do in its constructor/initializer?

inabox/tests/setup_disperser_harness.go
```diff
@@ -141,12 +183,255 @@ func SetupDisperserHarness(ctx context.Context, config *DisperserHarnessConfig)
 	return harness, nil
 }
 
+// RelayInstance holds the state for a single relay
+type RelayInstance struct {
+	Server   *relay.Server
+	Listener net.Listener
```
samlaf, line 187: ah I kept wondering how we can wire together all the services if we use port 0 to let the OS assign port numbers to each, think this just gave me the answer.

If all the different services get injected a listener instead of creating one in their constructor, then a test harness can create all the listeners, and then bind all the assigned ports to the difference services, build a map of service-name -> address (b/c its got all the ports now), and then inject them into all the constructors.

We should do this!

inabox/tests/setup_disperser_harness.go
```diff
@@ -141,12 +183,255 @@ func SetupDisperserHarness(ctx context.Context, config *DisperserHarnessConfig)
...
+		GRPCPort:                   mustParsePort(port),
+		MaxGRPCMessageSize:         1024 * 1024 * 300,
+		MetadataCacheSize:          1024 * 1024,
+		MetadataMaxConcurrency:     32,
+		BlobCacheBytes:             32 * 1024 * 1024,
+		BlobMaxConcurrency:         32,
+		ChunkCacheBytes:            32 * 1024 * 1024,
+		ChunkMaxConcurrency:        32,
+		MaxKeysPerGetChunksRequest: 1024,
+		RateLimits: limiter.Config{
```
samlaf, line 360: same thing as for logs, I think all configs should be created in a centralized place and injected afterwards.

Think a really nice refactor would be to:
1. create backing test infra (chain, thats it?)
2. create all the listeners (see my previous comment; from this we get the ports for all the services)
3. create the configs for all services (inject the listeners from step 2; they contain the service addresses)
4. start all services (inject the configs from step 3 + the contexts to manage shutdown (!!!!!))

test/testbed/deploy_localstack_resources.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2096/commits/8bdf95fce552d7d4f68369507b077a9e65ef5bcb

test/testbed/deploy_subgraphs.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2096/commits/1f7086c5887cdfced4575532df0e851f94e61938

inabox/tests/setup_infra.go
dmanc: ed113e2b531cd8efdefcf19e3f1981430351ecf7 but would like to discuss context management

inabox/tests/setup_infra.go
dmanc: ed113e2b531cd8efdefcf19e3f1981430351ecf7

inabox/tests/setup_disperser_harness.go
dmanc: nice lmk if this is what you had in mind: 77270f297deac28b194300812620f05efc7b0713

inabox/tests/setup_disperser_harness.go
dmanc: removed in some commit since I was using it to debug

inabox/tests/setup_disperser_harness.go
dmanc: 77270f297deac28b194300812620f05efc7b0713, i opted to just log a warning since some test may not need relays

inabox/tests/setup_infra.go
dmanc: 77270f297deac28b194300812620f05efc7b0713, I left number of relays as a config option. It's possible some may even want to control the sharding configuration for each relay too but will leave the config simple for now.

inabox/tests/integration_suite_test.go
dmanc: 77270f297deac28b194300812620f05efc7b0713, with the change you suggested to use OS allocated ports this is no longer hardcoded

inabox/tests/setup_disperser_harness.go
dmanc: added as a todo https://github.com/Layr-Labs/eigenda/pull/2096/commits/922ccd1c135f91f5924cc6477c1435cd2a2dcb6c

inabox/tests/setup_disperser_harness.go
dmanc: added as a todo https://github.com/Layr-Labs/eigenda/pull/2096/commits/922ccd1c135f91f5924cc6477c1435cd2a2dcb6c

relay/server.go
```diff
@@ -159,6 +159,7 @@ func NewServer(
 		authenticator:    authenticator,
 		replayGuardian:   replayGuardian,
 		metrics:          relayMetrics,
+		chainReader:      chainReader,
```
dmanc, line 162: This might be a bug lol

inabox/tests/setup_disperser_harness.go
```diff
@@ -141,12 +180,278 @@ func SetupDisperserHarness(ctx context.Context, config *DisperserHarnessConfig)
...
+// startRelays starts all relay goroutines
+func startRelays(ctx context.Context, harness *DisperserHarness, config DisperserHarnessConfig) error {
+	config.Logger.Info("Pre-creating listeners for relay goroutines", "count", config.RelayCount)
+
+	// Pre-create all listeners with port 0 (OS assigns ports)
+	listeners := make([]net.Listener, config.RelayCount)
+	actualURLs := make([]string, config.RelayCount)
+
+	for i := range config.RelayCount {
+		listener, err := net.Listen("tcp", "localhost:0")
```
samlaf:
```suggestion
		listener, err := net.Listen("tcp", "0.0.0.0:0")
```

samlaf: not sure but this just seems safer in case warp does some weirdness with iptables.

relay/server.go
samlaf: ...?? how/where does relay use chainReader. and how was it working without this?

samlaf (approved):

relay/server.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2100

inabox/tests/setup_disperser_harness.go
dmanc: https://github.com/Layr-Labs/eigenda/commit/9c0f778ba5784123bc9bfd32cf433e48c09ca473

PR #2098, opened 2025-10-01
feat(proxy): Add `error-on-secondary-insert-failure` flag for strict secondary storage error handling

shyam-patel-kira:
Added `secondary.error-on-secondary-insert-failure` CLI flag to enable strict error handling for secondary storage write failures. When enabled:
- secondary storage write failures return HTTP 500 to the client instead of being silently logged
- Prevents flag from being enabled with async writes through configuration validation

Default is `false` to maintain backward compatibility


api/proxy/store/secondary/cli.go
```diff
@@ -0,0 +1,48 @@
+package secondary
+
+import (
```
ethenotethan: why are we introducing a separate CLI package vs just extending the existing one which already supports secondary storage env vars? https://github.com/Layr-Labs/eigenda/blob/master/api/proxy/store/cli.go

api/proxy/store/cli.go
```diff
@@ -117,12 +123,15 @@ func ReadConfig(ctx *cli.Context) (Config, error) {
 		}
 	}
 
+	secondaryCfg := secondary.ReadConfig(ctx)
```
ethenotethan: there's an abstraction leak with this pattern given other env values being ingested here (e.g, WriteOnCacheMiss) are unique to secondary storage logic. Having this additional sub-config feels like a regression and unnecessarily increases code overhead imo. 

If we do want to introduce an additional sub-config or CLI group then there should be a better taxonomy or hierarchical definition here. E.g. having some "policy" sub-config which defines how secondary interaction logic works for writing and reading (e.g, `WriteOnCacheMiss`, `ConcurrentWriteThreads`, `ErrorOnSecondaryInsertFailure`).

api/proxy/store/config.go
```diff
@@ -60,5 +61,11 @@ func (cfg *Config) Check() error {
 		return fmt.Errorf("number of secondary write workers can't be greater than 100")
 	}
 
+	// verify that ErrorOnSecondaryInsertFailure is not enabled with async writes
+	if cfg.ErrorOnSecondaryInsertFailure && cfg.AsyncPutWorkers > 0 {
+		return fmt.Errorf("error-on-secondary-insert-failure requires synchronous writes " +
+			"(concurrent-write-routines must be 0)")
```
ethenotethan: nit:
```suggestion
		return fmt.Errorf("error-on-secondary-insert-failure requires synchronous writes " +
			"(i.e, storage.concurrent-write-routines must be 0)")
```

api/proxy/store/validation_test.go
```diff
@@ -0,0 +1,64 @@
+package store
```
ethenotethan: why roll an entirely new test vs extend the existing one which are already is used to enforce expressional config invariants? https://github.com/Layr-Labs/eigenda/blob/master/api/proxy/store/config_test.go

api/proxy/store/secondary/secondary.go
```diff
@@ -125,14 +139,22 @@ func (sm *SecondaryManager) HandleRedundantWrites(ctx context.Context, commitmen
...
+	// If no writes succeeded at all, always return error
 	if successes == 0 {
-		return errors.New("failed to write blob to any redundant targets")
+		return fmt.Errorf("failed to write blob to any redundant targets: %w", errors.Join(errs...))
+	}
+
+	// If errorOnInsertFailure is enabled and any writes failed (partial success), return error
+	if sm.errorOnInsertFailure && len(errs) > 0 {
+		return fmt.Errorf("failed to write to %d of %d secondary targets (strict mode enabled): %w",
+			len(errs), len(sources), errors.Join(errs...))
```
ethenotethan: nit:
what is `(strict mode enabled)` mean? Why not just use refer to the boolean field verbatim?

api/proxy/store/eigenda_manager.go
```diff
@@ -283,18 +292,32 @@ func (m *EigenDAManager) putToCorrectEigenDABackend(ctx context.Context, value [
...
+
+	// Synchronous writes
+	m.log.Debug("Publishing data to single threaded secondary stores")
+	err := m.secondary.HandleRedundantWrites(ctx, commitment, value)
+	if err != nil {
+		m.log.Error("Secondary insertions failed", "error", err.Error())
+		// Only propagate the error if errorOnInsertFailure is enabled.
+		// This allows the caller to return HTTP 500 to the client.
+		if m.secondary.ErrorOnInsertFailure() {
+			return fmt.Errorf("secondary storage write failed: %w", err)
```
ethenotethan:
```suggestion
			return fmt.Errorf("a secondary storage write failed and error-on-secondary-insert-failure is enabled: %w", err)
```

api/proxy/store/eigenda_manager.go
```diff
@@ -283,18 +292,32 @@ func (m *EigenDAManager) putToCorrectEigenDABackend(ctx context.Context, value [
 	return nil, fmt.Errorf("unsupported dispersal backend: %v", backend)
 }
 
-func (m *EigenDAManager) backupToSecondary(ctx context.Context, commitment []byte, value []byte) {
+// backupToSecondary writes data to secondary storage backends (caches and fallbacks).
+// When errorOnInsertFailure is enabled and writes are synchronous, errors are returned
+// to the caller to propagate as HTTP 500 responses. For async writes, errors are only logged.
+func (m *EigenDAManager) backupToSecondary(ctx context.Context, commitment []byte, value []byte) error {
```
ethenotethan, line 298: @samlaf curious if it'd make better sense to yield a `503` instead? This feature ensures an all-or-error pattern where we enforce distributed consistency between primary (EigenDA) and redundant secondaries. But this isn't actually atomic, i.e. there's a cost paid for each of these successful EigenDA dispersals with no reversion mechanisms in place. 

From the perspective of the rollup this is a DoS cost vector since a batch poster would continuously pay for blobs which never get accredited to sequencer inbox. Imo this case getting triggered justifies a `503` instead to mitigate unnecessary cost drain

api/proxy/store/validation_test.go
```diff
@@ -0,0 +1,64 @@
...
+
+import (
```
ethenotethan: can we add an E2E test which ensures the correctness of this feature? https://github.com/Layr-Labs/eigenda/tree/master/api/proxy/test/e2e

api/proxy/store/secondary/cli.go
shyam-patel-kira: Noted, thanks for pointing out, fixed `054b021`

api/proxy/store/cli.go
shyam-patel-kira: `054b021`

api/proxy/store/config.go
shyam-patel-kira: `054b021`

shyam-patel-kira: `068c0a9`

api/proxy/store/validation_test.go
shyam-patel-kira: removed this in `054b021`

api/proxy/store/secondary/secondary.go
shyam-patel-kira: makes sense, fixed
`054b021`

api/proxy/store/validation_test.go
shyam-patel-kira: Added e2e test, `3fe9fef`

api/proxy/store/cli.go
```diff
@@ -73,7 +74,17 @@ func CLIFlags(envPrefix, category string) []cli.Flag {
 			EnvVars:  withEnvPrefix(envPrefix, "WRITE_ON_CACHE_MISS"),
 			Category: category,
 		},
+		&cli.BoolFlag{
+			Name: ErrorOnSecondaryInsertFailureFlagName,
+			Usage: "Return HTTP 500 if any secondary storage write fails. " +
```
ethenotethan, line 79: Can we also articulate the risk where a rollup batch poster can infinite death loop since this mode binds the liveness of the batch poster to that of the secondary?

api/proxy/store/secondary/secondary_test.go
```diff
@@ -0,0 +1,137 @@
...
+
+import (
+	"context"
+	"errors"
+	"os"
+	"testing"
+
+	"github.com/Layr-Labs/eigenda/api/proxy/common"
+	"github.com/Layr-Labs/eigenda/api/proxy/metrics"
+	"github.com/Layr-Labs/eigenda/api/proxy/store/secondary"
```
ethenotethan: Id rather we not mock the storage layer for usage in the E2E tests and instead just produce the failure case verbatim. Ie by setting an invalid secondary S3 URL or tearing down the instance so it's not available.

api/proxy/test/e2e/server_rest_test.go
```diff
@@ -346,6 +348,197 @@ func testProxyWriteCacheOnMiss(t *testing.T, dispersalBackend common.EigenDABack
...
+// secondary storage write failures are logged but don't cause the PUT to fail.
+func TestErrorOnSecondaryInsertFailureFlagOffV1(t *testing.T) {
+	testErrorOnSecondaryInsertFailureFlagOff(t, common.V1EigenDABackend)
+}
+
+func TestErrorOnSecondaryInsertFailureFlagOffV2(t *testing.T) {
+	testErrorOnSecondaryInsertFailureFlagOff(t, common.V2EigenDABackend)
+}
+
+func testErrorOnSecondaryInsertFailureFlagOff(t *testing.T, dispersalBackend common.EigenDABackend) {
```
ethenotethan: Why is this test needed given the default  value would be used in other pre-existing tests?

api/proxy/test/e2e/server_rest_test.go
```diff
@@ -346,6 +348,197 @@ func testProxyWriteCacheOnMiss(t *testing.T, dispersalBackend common.EigenDABack
...
+	testCfg.ErrorOnSecondaryInsertFailure = true // Enable flag
+	testCfg.WriteThreadCount = 5                 // Enable async writes (INVALID)
+
+	tsConfig := testutils.BuildTestSuiteConfig(testCfg)
+
+	// Creating test suite should fail due to config validation
+	defer func() {
+		if r := recover(); r != nil {
+			errMsg := fmt.Sprintf("%v", r)
+			require.Contains(t, errMsg, "requires synchronous writes",
```
ethenotethan: Are there any additional guarantees that are provided by duplicating this check here vs just relying on the existing config test? This would also be a regression to the testing strategy already introduced

api/proxy/store/config_test.go
```diff
@@ -65,4 +65,41 @@ func TestConfigVerification(t *testing.T) {
 		err := cfg.Check()
 		require.Error(t, err)
 	})
+
+	t.Run("ErrorOnSecondaryInsertFailure: flag OFF, async OFF", func(t *testing.T) {
```
ethenotethan: There shouldn't be a need to have this test

api/proxy/store/config_test.go
```diff
@@ -65,4 +65,41 @@ func TestConfigVerification(t *testing.T) {
...
+		cfg := validCfg()
+		cfg.AsyncPutWorkers = 0
+		cfg.ErrorOnSecondaryInsertFailure = false
+
+		err := cfg.Check()
+		require.NoError(t, err)
+	})
+
+	t.Run("ErrorOnSecondaryInsertFailure: flag OFF, async ON", func(t *testing.T) {
```
ethenotethan: Again here as well

api/proxy/store/config_test.go
shyam-patel-kira: I added these to increase coverage, but probably doesn't matter

ethenotethan: yeah makes sense - I've burned myself in past projects by squeezing the coverage metric to 90%+ to then realize that:
1. the number of lines covered doesn't matter as more specific assumptions and invariants are
2. there's a maintenance burden when introducing new change sets given a lot of tests will need reworking

api/proxy/store/cli.go
shyam-patel-kira: `63f00ad8`

api/proxy/test/e2e/server_rest_test.go
shyam-patel-kira: nope, removed this `63f00ad8`

api/proxy/store/secondary/secondary_test.go
shyam-patel-kira: I just noticed there already something like this present in the codebase

api/proxy/store/config_test.go
ethenotethan: any chance we could remove the erroneous tests?

api/proxy/test/e2e/server_rest_test.go
```diff
@@ -346,6 +347,110 @@ func testProxyWriteCacheOnMiss(t *testing.T, dispersalBackend common.EigenDABack
...
+// and all secondary writes succeed, the PUT succeeds normally.
+func TestErrorOnSecondaryInsertFailureSuccessV1(t *testing.T) {
+	testErrorOnSecondaryInsertFailureSuccess(t, common.V1EigenDABackend)
+}
+
+func TestErrorOnSecondaryInsertFailureSuccessV2(t *testing.T) {
+	testErrorOnSecondaryInsertFailureSuccess(t, common.V2EigenDABackend)
+}
+
+func testErrorOnSecondaryInsertFailureSuccess(t *testing.T, dispersalBackend common.EigenDABackend) {
```
ethenotethan: what exactly is this test guaranteeing? iiuc this is just a normal secondary happy case

api/proxy/store/secondary/secondary.go
```diff
@@ -125,14 +139,22 @@ func (sm *SecondaryManager) HandleRedundantWrites(ctx context.Context, commitmen
 		if err != nil {
 			sm.log.Warn("Failed to write to redundant target", "backend", src.BackendType(), "err", err)
 			cb(Failed)
+			errs = append(errs, fmt.Errorf("write to %s failed: %w", src.BackendType(), err))
```
litt3, line 146: Do we actually want to continue if `errorOnInsertFailure` is `true`, and an error happens? Or would it make sense to immediately return any error that's encountered?

I'm lacking background here, but my default assumption with a configuration field like this would be to expect an early return if an error is encountered. It might be worth explicitly describing the behavior in the docs of the config field and the `HandleRedundantWrites` method.

shyam-patel-kira: good catch, if do a fast-fail we need to consider DoS attack vector where Adversary could intentionally make first backend (cache) unavailable to trigger errors:
1. Attacker DoS's cache backend (e.g., S3)
2. All PUT requests fail immediately with HTTP 500
3. Batch poster enters infinite retry loop
4. System effectively down even though primary EigenDA + fallback are healthy

I'm not sure if it's probably okay to have it since nodes would be running these. I have added fast-fail here [`6a7f8d7`](https://github.com/Layr-Labs/eigenda/pull/2098/commits/6a7f8d77e6883341892dc282d9ee47c140e749e0) for now, can remove if not required wyt?
cc: @samlaf @ethenotethan

api/proxy/test/e2e/server_rest_test.go
shyam-patel-kira: removed these here: [6a7f8d7](https://github.com/Layr-Labs/eigenda/pull/2098/commits/6a7f8d77e6883341892dc282d9ee47c140e749e0) and added more relevant ones here [87ae7e1](https://github.com/Layr-Labs/eigenda/pull/2098/commits/87ae7e18747444b424dce0a66dc2f95ab22568ba)

api/proxy/store/secondary/secondary.go
litt3: > System effectively down even though primary EigenDA + fallback are healthy

Isn't that essentially what this new feature is doing anyway?

shyam-patel-kira: > System effectively down even though primary EigenDA + fallback are healthy

refers to a specific DoS attack scenario, not the normal operation of the flag.

Default behavior is: 
- Primary EigenDA succeeds, secondary storage fails → HTTP 500 (with flag ON)
- This couples batch poster liveness to secondary storage availability (documented in CLI warning)

DoS scenario:
  - Attacker targets only the cache backend (tried first in code at line 123-124)
  - Cache fails → fail-fast triggers immediate HTTP 500
  - Fallback backend is never attempted (even though it's healthy)
  - System appears down despite both primary EigenDA AND fallback being operational

Issue is Fail-fast stops iteration after first failure, so attacking just the cache (first secondary backend) can make the entire system unavailable, even when primary + fallback are healthy. I've documented this already in CLI warning. Users enabling this flag must understand they're accepting this trade-off for strict consistency guarantees, that is If we decide to keep the fast fail.

litt3: I don't understand the argument that failing fast changes the potential DoS profile- it seems someone with the ability to bring down the s3 bucket could do the same attack, whether this loop fails fast or not, if the flag is `true`.

In any case, the new fail-fast logic LGTM

shyam-patel-kira: The fail-fast behavior doesn't introduce a new vulnerability, but it does make the DoS attack easier/cheaper in scenarios with multiple secondary backends: Without fail-fast, Attacker must take down BOTH cache AND fallback to cause system failure, with fail-fast, Attacker only needs to take down cache (tried first)

PR #2102, opened 2025-10-01
feat(inabox): Spin up encoder + encoder v2 as goroutines in inabox integration tests

dmanc:

This PR also introduces `DisperserV1Harness` to deploy V1 disperser components. In addition we do some minor refactoring around localstack and resource deployment.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/encoder/server.go
+++ b/disperser/encoder/server.go
@@ -128,6 +128,34 @@ func (s *EncoderServer) Start() error {
 	return gs.Serve(listener)
 }
 
+// StartWithListener starts the server using the provided listener. This method will block until the server is stopped.
+func (s *EncoderServer) StartWithListener(listener net.Listener) error {
+	opt := grpc.MaxRecvMsgSize(1024 * 1024 * 300) // 300 MiB
+	gs := grpc.NewServer(opt,
+		grpc.UnaryInterceptor(
+			s.grpcMetrics.UnaryServerInterceptor(),
+		),
+	)
+	reflection.Register(gs)
+	pb.RegisterEncoderServer(gs, s)
+	s.grpcMetrics.InitializeMetrics(gs)
+
+	// Register Server for Health Checks
+	name := pb.Encoder_ServiceDesc.ServiceName
+	healthcheck.RegisterHealthServer(name, gs)
+
+	s.close = func() {
+		err := listener.Close()
+		if err != nil {
+			log.Printf("failed to close listener: %v", err)
+		}
+		gs.GracefulStop()
+	}
+
+	s.logger.Info("port", s.config.GrpcPort, "address", listener.Addr().String(), "GRPC Listening")
+	return gs.Serve(listener)
+}
+
 func (s *EncoderServer) EncodeBlob(ctx context.Context, req *pb.EncodeBlobRequest) (*pb.EncodeBlobReply, error) {
 	startTime := time.Now()
 	blobSize := len(req.GetData())
--- a/disperser/encoder/server.go
+++ b/disperser/encoder/server.go
@@ -125,7 +125,10 @@ func (s *EncoderServer) Start() error {
 	}
 
 	s.logger.Info("port", s.config.GrpcPort, "address", listener.Addr().String(), "GRPC Listening")
-	return gs.Serve(listener)
+	if err := gs.Serve(listener); err != nil {
+		return fmt.Errorf("failed to serve grpc: %w", err)
+	}
+	return nil
 }
 
 // StartWithListener starts the server using the provided listener. This method will block until the server is stopped.
@@ -153,7 +156,10 @@ func (s *EncoderServer) StartWithListener(listener net.Listener) error {
 	}
 
 	s.logger.Info("port", s.config.GrpcPort, "address", listener.Addr().String(), "GRPC Listening")
-	return gs.Serve(listener)
+	if err := gs.Serve(listener); err != nil {
+		return fmt.Errorf("failed to serve grpc: %w", err)
+	}
+	return nil
 }
 
 func (s *EncoderServer) EncodeBlob(ctx context.Context, req *pb.EncodeBlobRequest) (*pb.EncodeBlobReply, error) {
--- a/disperser/encoder/server_v2.go
+++ b/disperser/encoder/server_v2.go
@@ -86,6 +86,11 @@ func (s *EncoderServerV2) Start() error {
 		log.Fatalf("Could not start tcp listener: %v", err)
 	}
 
+	return s.StartWithListener(listener)
+}
+
+// StartWithListener starts the server using the provided listener. This method will block until the server is stopped.
+func (s *EncoderServerV2) StartWithListener(listener net.Listener) error {
 	gs := grpc.NewServer(
 		grpc.UnaryInterceptor(
 			s.grpcMetrics.UnaryServerInterceptor(),
--- a/inabox/deploy/cmd/main.go
+++ b/inabox/deploy/cmd/main.go
@@ -122,7 +122,7 @@ func DeployAll(ctx *cli.Context) error {
 	config.RegisterRelays(ethClient, relayURLs, ethClient.GetAccountAddress())
 
 	logger.Info("Generating variables")
-	err = config.GenerateAllVariables("0.0.0.0:34001")
+	err = config.GenerateAllVariables("0.0.0.0:34000", "0.0.0.0:34001")
 	if err != nil {
 		logger.Errorf("could not generate environment variables: %v", err)
 		panic(err)
--- a/inabox/deploy/cmd/main.go
+++ b/inabox/deploy/cmd/main.go
@@ -122,7 +122,7 @@ func DeployAll(ctx *cli.Context) error {
 	config.RegisterRelays(ethClient, relayURLs, ethClient.GetAccountAddress())
 
 	logger.Info("Generating variables")
-	err = config.GenerateAllVariables()
+	err = config.GenerateAllVariables("0.0.0.0:34001")
 	if err != nil {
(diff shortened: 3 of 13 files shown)


disperser/encoder/server.go
```diff
@@ -125,7 +125,41 @@ func (s *EncoderServer) Start() error {
...
-	return gs.Serve(listener)
+	if err := gs.Serve(listener); err != nil {
+		return fmt.Errorf("failed to serve grpc: %w", err)
+	}
+	return nil
+}
+
+// StartWithListener starts the server using the provided listener. This method will block until the server is stopped.
+func (s *EncoderServer) StartWithListener(listener net.Listener) error {
+	opt := grpc.MaxRecvMsgSize(1024 * 1024 * 300) // 300 MiB
```
samlaf, line 90: why duplicate code like this? can't you change Start() to use this one like you did in the previous component?

inabox/deploy/cmd/main.go
```diff
@@ -122,7 +123,7 @@ func DeployAll(ctx *cli.Context) error {
 	config.RegisterRelays(ethClient, relayURLs, ethClient.GetAccountAddress())
 
 	logger.Info("Generating variables")
-	err = config.GenerateAllVariables()
+	err = config.GenerateAllVariables("0.0.0.0:34000", "0.0.0.0:34001")
```
samlaf: can you make this a config or something? bit random to pass addresses like this, and its hard to read. also why can't we use port 0 like we did for relays?

samlaf: also not sure we get any benefit from passing these here. Might as well keep them hardcoded all in the config.go file, and then swap all of them for generated ports once we're ready. Best to keep the hardcoded ports all in one place

inabox/tests/helpers.go
```diff
@@ -0,0 +1,29 @@
...
+import (
+	"fmt"
+	"path/filepath"
+	"runtime"
+)
+
+// getSRSPaths returns the correct paths to SRS files based on the source file location.
+// This uses runtime.Caller to determine where this file is located and calculates
+// the relative path to the resources/srs directory from there.
+func getSRSPaths() (g1Path, g2Path string, err error) {
```
samlaf, line 25: nice!. but I don't think this belongs in inabox. we should be using it for all our tests everywhere. don't even think its test only related. I'd prob want to use this in proxy

inabox/tests/setup_disperser_v1_harness.go
```diff
@@ -0,0 +1,264 @@
...
+type DisperserV1HarnessConfig struct {
+	Logger              logging.Logger
+	Network             *testcontainers.DockerNetwork
+	TestConfig          *deploy.Config
+	TestName            string
+	InMemoryBlobStore   bool
+	BucketTableName     string
+	V1MetadataTableName string
+	BlobStoreBucketName string
+	EthClient           common.EthClient
```
samlaf: client should not be part of config. tbh neither should network.

config is just config params. everything else is injected resources. they should be passed to constructors as individual objects imo

samlaf (commented):
Mostly LGTM although I am starting to fear that we're adding a lot of tech debt by not taking the time to refactor our services' constructors/start functions and are duplicating a lot of code. Might be worth taking the time to do a proper refactor of service code at the same time.. wydt?

inabox/tests/setup_disperser_v1_harness.go
dmanc: 5c3fc183ca466f46138aff27875cb666488ec777

disperser/encoder/server.go
dmanc: 5c3fc183ca466f46138aff27875cb666488ec777 + https://github.com/Layr-Labs/eigenda/pull/2102/commits/98c925ad5cb04cc76f6c57ca2f9bec63d630c0fe

dmanc:
> Mostly LGTM although I am starting to fear that we're adding a lot of tech debt by not taking the time to refactor our services' constructors/start functions and are duplicating a lot of code. Might be worth taking the time to do a proper refactor of service code at the same time.. wydt?

I could do that but it will just take more time to integrate each component -> inabox tests rely on startbinaries longer. I have 4 components left to integrate (batcher, controller, API servers, retriever) and they likely each have the same problem.

inabox/deploy/cmd/main.go
dmanc: The only two pieces that depend on this function are the devnet mode (the one you reviewed here) and the integration tests. Might be better to pull out all the hardcoded stuff imo since it will make it easier to replace.

samlaf:
> I could do that but it will just take more time to integrate each component -> inabox tests rely on startbinaries longer. I have 4 components left to integrate (batcher, controller, API servers, retriever) and they likely each have the same problem.

But the OKR is to reduce overall tech debt right? Not just to improve inabox. I'm just starting to fear inabox will be hard to refactor into a nice structure if we keep piling more golang tech debt into it. Maybe we should just plow through this as you suggest, and after that only having to deal with golang code will somehow make it easier to refactor/do small improvements. Maybe that's the correct strategy.. I'm not sure.

samlaf (dismissed):

disperser/cmd/encoder/main.go
```diff
@@ -83,22 +92,35 @@ func RunEncoderServer(ctx *cli.Context) error {
...
 		config.EncoderConfig.LoadG2Points = false
 		prover, err := proverv2.NewProver(logger, proverv2.KzgConfigFromV1Config(&config.EncoderConfig), encodingConfig)
 		if err != nil {
+			_ = listener.Close()
 			return fmt.Errorf("failed to create encoder: %w", err)
 		}
 
 		s3Client, err := s3.NewClient(context.Background(), config.AwsClientConfig, logger)
 		if err != nil {
+			_ = listener.Close()
```
samlaf: Can you defer listener.Close() instead of calling it at every err site? easy to forget to do when someone does a refactor like this.

disperser/cmd/encoder/main.go
```diff
@@ -119,16 +141,19 @@ func RunEncoderServer(ctx *cli.Context) error {
...
 	if err != nil {
+		_ = listener.Close()
 		return fmt.Errorf("failed to create encoder: %w", err)
 	}
 
 	server := encoder.NewEncoderServer(*config.ServerConfig, logger, prover, metrics, grpcMetrics)
 
-	return server.Start()
+	logger.Info("Starting encoder v1 server", "address", listener.Addr().String())
+	return fmt.Errorf("failed to start encoder v1 server: %w", server.StartWithListener(listener))
```
samlaf: this will return an error even if StartWithListener is shutdown normally. I think that's not what we want?

disperser/encoder/server.go
```diff
@@ -124,7 +109,7 @@ func (s *EncoderServer) Start() error {
 		gs.GracefulStop()
 	}
 
-	s.logger.Info("port", s.config.GrpcPort, "address", listener.Addr().String(), "GRPC Listening")
+	s.logger.Info("address", listener.Addr().String(), "GRPC Listening")
```
samlaf: wtf this looks completely backwards lol

samlaf: prob meant:
```
s.logger.Info("GRPC Listening", "address", listener.Addr().String())
```

disperser/encoder/server_v2.go
```diff
@@ -107,7 +100,7 @@ func (s *EncoderServerV2) Start() error {
 		gs.GracefulStop()
 	}
 
-	s.logger.Info("port", s.config.GrpcPort, "address", listener.Addr().String(), "GRPC Listening")
+	s.logger.Info("address", listener.Addr().String(), "GRPC Listening")
```
samlaf: also backwards

inabox/tests/setup_disperser_harness.go
```diff
@@ -207,19 +214,34 @@ func SetupDisperserHarness(
...
 	// Start controller goroutine
-	if err := startController(ctx, ethClient, config.OperatorStateSubgraphURL, harness, config); err != nil {
+	if err := startController(
+		ctx,
+		ethClient,
+		config.OperatorStateSubgraphURL,
+		encoderAddress,
+		harness,
+		config,
+	); err != nil {
```
samlaf, line 231: still think all of these should be refactored to return the created structs which we can then manually add to the harness. Prefer when inplace modifications are explicit. But can all be done in a future PR.

disperser/encoder/server_v2.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2102/commits/df1b76d047cec42135f4b639dab78e88e8cff692

disperser/encoder/server.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2102/commits/df1b76d047cec42135f4b639dab78e88e8cff692

disperser/cmd/encoder/main.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2102/commits/df1b76d047cec42135f4b639dab78e88e8cff692, I just removed the wrapped error. This is what we were doing before so I think it's fine.

disperser/cmd/encoder/main.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2102/commits/df1b76d047cec42135f4b639dab78e88e8cff692

inabox/tests/setup_disperser_harness.go
dmanc: Thanks I'll make a larger refactor once I get API server in

PR #2104, opened 2025-10-02
feat: simple config framework

cody-littley:

Introduces a new configuration framework that's WAY easier to work with than urfave. See the [README](https://github.com/Layr-Labs/eigenda/blob/2086-config/common/config/README.md) for a more description.

### DIFF
--- a/common/config/README.md
+++ b/common/config/README.md
@@ -1,4 +1,4 @@
-# Configuration Managmenet
+# Configuration Management
 
 This configuration "framework" attempts to achieve maximal simplicity when it comes to creating, modifying, and
 and maintaining configuration. Configuration is inherently a simple concept, and so the execution of configuration
@@ -31,7 +31,7 @@ type VerifiableConfig interface {
 }
 ```
 
-Although in theory the `Verify()` method can be a no-op, it is highly reccomended to implement basic santity checking.
+Although in theory the `Verify()` method can be a no-op, it is highly recommended to implement basic sanity checking.
 
 The "constructor" for a config object must satisfy the interface `func() T` where `T` 
 implements `config.VerifiableConfig`.
@@ -75,7 +75,7 @@ cfg, err := config.ParseConfig(DefaultMyConfig, MyAppPrefix, "path/to/config1.to
 
 ## ParseConfigFromCLI()
 
-An alternate way of parsing configuaration is with the following method:
+An alternate way of parsing configuration is with the following method:
 
 ```
 ParseConfigFromCLI[T VerifiableConfig](constructor func() T, envPrefix string) (T, error)
@@ -85,10 +85,10 @@ The primary difference between this method and `ParseConfig()` is that `ParseCon
 command line arguments provided to the process are configuration file paths. Using this method is functionally
 equivalent to parsing for file paths from the CLI arguments, then passing those file paths into `ParseConfig()`.
 
-Although use of `ParseConfigFromCLI()` is not required to use this framwork, it is highly encouraged. Any time
+Although use of `ParseConfigFromCLI()` is not required to use this framework, it is highly encouraged. Any time
 configuration is sourced through multiple pathways, complexity grows. If configuration is large enough that the
-config framework is needed, then it's best if all configuration flows through the config framwork. A hodge-podge
-of CLI arguments mixed with the configuration framework adds a lot of unecessary complexity.
+config framework is needed, then it's best if all configuration flows through the config framework. A hodge-podge
+of CLI arguments mixed with the configuration framework adds a lot of unnecessary complexity.
 
 # Documenting Config
 
@@ -116,7 +116,7 @@ grug no like yaml, yaml look simple sometimes, but grug know yaml only pretend b
 but grug not reach for club if not use toml.
 ```
 
-In order to set a varaible in a config file, simply mirror the struct and variable names in "the obvious way".
+In order to set a variable in a config file, simply mirror the struct and variable names in "the obvious way".
 Below is an example using toml.
 
 ```go
@@ -167,7 +167,7 @@ should contain only upper case letters and underscores.
 
 For each entry in a config struct, there is an environment variable that is mapped to that entry. The name of the
 environment variable is `PREFIX_NAMEOFVARIABLE`. If the variable is in a nested struct, for each "parent variable",
-add the name of the parent variable in uppercase, and seperate parent variables with underscores. 
+add the name of the parent variable in uppercase, and separate parent variables with underscores. 
 
 The following example shows the names of the environment variables that could be used to configure the following
 struct.
@@ -195,7 +195,7 @@ type Baz struct {
 
 ## Mistyped Environment Variables
 
-The config framework looks at all enviornment variables that begin with the prefix. If it finds any environment
+The config framework looks at all environment variables that begin with the prefix. If it finds any environment
 variable with the prefix that does not map to an entry in the config struct, it returns an error. This is intentional.
 Similar to mistyped config, an environment variable that doesn't map to a config entry is likely to be a bug.
 
@@ -203,7 +203,7 @@ Similar to mistyped config, an environment variable that doesn't map to a config
 
 The purpose of the constructor is to set default values in the struct. The config API requires a constructor method
 in order to strongly encourage users of this framework to set sane default values where possible. In general,
-the fewer values that are requried to be set, the easier it is to configure something.
+the fewer values that are required to be set, the easier it is to configure something.
 
 # Required Values
 
--- a/common/config/config_parser.go
+++ b/common/config/config_parser.go
@@ -12,8 +12,8 @@ import (
 )
 
 // ParseConfig parses the configuration from the given paths and environment variables. Configuration files are
-// loaded in order, with later files overriding earlier ones. Environment variables are loaded last, and overrid values
-// from all configuration files. If there are default values in the config, the counstructor should return an instance
+// loaded in order, with later files overriding earlier ones. Environment variables are loaded last, and override values
+// from all configuration files. If there are default values in the config, the constructor should return an instance
 // initialized with those values.
 func ParseConfig[T VerifiableConfig](
 	// A function that returns a new instance of the config struct to be populated.
--- a/common/config/config_test.go
+++ b/common/config/config_test.go
@@ -449,7 +449,7 @@ func TestInvalidEnvironmentVariable(t *testing.T) {
 	require.NoError(t, os.Unsetenv("PREFIX_THIS_VARIABLE_WAS_MISTYPED"))
 }
 
-func TestVerificaitonFailure(t *testing.T) {
+func TestVerificationFailure(t *testing.T) {
 	configFile := "test/config.toml"
 
 	// Set environment variables to override some config values.


go.mod
```diff
@@ -245,18 +245,33 @@ require (
...
 	google.golang.org/genproto/googleapis/rpc v0.0.0-20250528174236-200df99c418a // indirect
 	google.golang.org/protobuf v1.36.6
 	gopkg.in/natefinch/lumberjack.v2 v2.2.1 // indirect
 	gopkg.in/yaml.v2 v2.4.0 // indirect
 	gopkg.in/yaml.v3 v3.0.1
 )
+
+require (
+	github.com/go-viper/mapstructure/v2 v2.4.0
+	github.com/spf13/viper v1.21.0
```
dmanc, line 266: Did we consider https://github.com/spf13/cobra for CLI? Might play nice with viper

common/config/README.md
```diff
@@ -0,0 +1,207 @@
+# Configuration Management
+
+This configuration "framework" attempts to achieve maximal simplicity when it comes to creating, modifying, and
+and maintaining configuration. Configuration is inherently a simple concept, and so the execution of configuration
+should likewise be simple.
+
+# Config is Struct
+
+```
+grug say, me want config in this struct. why config need be more than struct?
```
samlaf, line 10: Let's do theory first: b/c config is a denormalized view, which means theres risk for misconfiguration/inconsistencies. You're basically asking "why don't we just use mongodb and dump our data is json?". Well relational DBs exist for a reason, and normalized structure is useful (although people will argue about the impedance mismatch and yadi yada).

In practice: I'm fully fully onboard with this. I do think we need to think harder about presenting an external config vs internal config (had written some thoughts about this long time ago in this [proxy issue](https://github.com/Layr-Labs/eigenda-proxy/issues/328)) and how to make it simple for external consumers to not misconfigure services. This internal vs external config is also why there are [config DSLs](https://samlaf.github.io/programming/config-dsls.html).

In actual practice: let's do this! start simple and iterate.

common/config/README.md
```diff
@@ -0,0 +1,207 @@
...
+// VerifiableConfig is an interface for configurations that can be validated.
+type VerifiableConfig interface {
+	// Verify checks that the configuration is valid, returning an error if it is not.
+	Verify() error
+}
+```
+
+Although in theory the `Verify()` method can be a no-op, it is highly recommended to implement basic sanity checking.
+
+The "constructor" for a config object must satisfy the interface `func() T` where `T`
```
samlaf, line 36: shall we talk about whether we want to use T or *T? We're generally pretty inconsistent in the repo afaict. I personally prefer T, as moving around Ts typically only happens at startup time anyways, so the costs outweight the benefits IMO.

Only reaon for using *T is if its EXPLICITLY shared between different structs AND requires being updated dynamically (in which case it needs a mutex etc, perhaps we could define a `ShareableUpdatableConfig` interface or something).

common/config/README.md
```diff
@@ -0,0 +1,207 @@
...
+There are two ways config can be loaded.
+
+The first way is to pass in a list of zero or more configuration files to `ParseConfig()`.
+
+```go
+ParseConfig[T VerifiableConfig](constructor func() T, envPrefix string, configPaths ...string) (T, error)
+```
+
+`ParseConfig()` will load data from the configuration files in order (later files override values from earlier files).
+After loading configuration files, `ParseConfig()` loads environment variables (overriding values set by config files).
```
samlaf, line 52: do we have actual use cases for overrides? Feels bug prone, and is why CUE is all about type checked inheritance with values+types being in the same hierarchy (https://cuelang.org/docs/concept/the-logic-of-cue/).

Also what about getting rid of env vars altogether? Do we absolutely need them?

common/config/README.md
```diff
@@ -0,0 +1,207 @@
...
+# How to Load Config
+
+## ParseConfig()
+
+There are two ways config can be loaded.
+
+The first way is to pass in a list of zero or more configuration files to `ParseConfig()`.
+
+```go
+ParseConfig[T VerifiableConfig](constructor func() T, envPrefix string, configPaths ...string) (T, error)
```
samlaf, line 48:
```suggestion
ParseConfig[T VerifiableConfig](constructor func() T, envVarPrefix string, configPaths ...string) (T, error)
```

samlaf: nit: took me a while to realize how this MYAPP was being used.

common/config/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+The following TOML file can be loaded into the structs above.
+
+```toml
+X = 1234
+Y = 3.14159265359
+Z = "this is a string"
+Bar.A = "this is another string"
+Bar.B = "5s"
+Bar.C.ValueStoredInAVariableWithALongName = "yet another string"
```
samlaf, line 149: even nicer is to use [[Bar]] sections

common/config/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+If there is a config value that does not have a corresponding entry in the config struct, the configuration framework
+will return an error when it attempts to parse the config. This is very intentional. Unmatched config file entries
+almost always signal a mistake in the configuration files. At the very least, returning an error for unmatched config
+forces config files to be kept clean and well maintained.
+
+# Environment Variables
+
+The config framework supports loading config from environment variables. Although the primary intended use case for
+environment variables is for loading secrets, there is nothing stopping configuration from being loaded entirely
```
samlaf, line 162: personally think secrets should not be stored in the config file anyways. In proxy we have a separate `config` and `secrets` structs. This is very nice because you always know you can safely just dump the parsed config to stdout during service startup, which helps with debugging.

common/config/README.md
```diff
@@ -0,0 +1,207 @@
...
+type Baz struct {
+    ValueStoredInAVariableWithALongName String // PREFIX_BAR_C_VALUESTOREDINAVARIABLEWITHALONGNAME
+}
+```
+
+## Mistyped Environment Variables
+
+The config framework looks at all environment variables that begin with the prefix. If it finds any environment
+variable with the prefix that does not map to an entry in the config struct, it returns an error. This is intentional.
+Similar to mistyped config, an environment variable that doesn't map to a config entry is likely to be a bug.
```
samlaf, line 200: ohhhhhhh this is nice! I really really really like this. Always hated env vars for this exact reason, but thought I couldn't do something like this b/c there's always a bunch of env vars set by shell.

But only checking env vars with the prefix solves this. nice

samlaf: one question though, how do you plan to deal with breaking change of deleting an entry from a config? It can be nice to clients that they can just keep reusing their old config, which in this case would mean allow correct prefix env vars that dont match config. though I'm not against just creating a breaking change either..

common/config/README.md
```diff
@@ -0,0 +1,207 @@
...
+
+# Default Values
+
+The purpose of the constructor is to set default values in the struct. The config API requires a constructor method
+in order to strongly encourage users of this framework to set sane default values where possible. In general,
+the fewer values that are required to be set, the easier it is to configure something.
```
samlaf, line 206: beautiful

samlaf (commented):
I love you Cody

common/config/README.md
pschork: > Also what about getting rid of env vars altogether? Do we absolutely need them?

Noooo! They are not absolutely required, but I like being able to describe a k8 deployment to examine the env var config. If you put everything in a config file, it's possible to mutate it and diverge from the config committed to repo.

The way traffic-generator-v2 helm chart [pipes it's config](https://github.com/Layr-Labs/eigenda-devops/blob/master/charts/traffic-generator-v2/values/eigenda-mainnet-beta/us-east-1/ethereum/values.yaml#L55) into json files is ugly imo.

go.mod
cody-littley: I'm personally not convinced that we need a CLI. We currently use urfave to create CLI arguments, but do we actually ever invoke things using the CLI arguments? I'm in favor of dropping support for setting configuration via CLI arguments, and forcing all config to go through config files and environment variables.

cody-littley:
@samlaf 

> I love you Cody

I'd prefer to just be friends.

go.mod
cody-littley: My personal opinion of CLIs is that they should generally be used when you expect a tool to be used directly by a human, or if the number of potential arguments is small (i.e. not hundreds of possible args). Things like the LittDB utility for backing up the database makes sense as a CLI, since it's a small self contained tool with just a few configurations, and since it's easy to believe that it may be used directly by a human in a terminal.

The validator, on the other hand, is a large and complex beast. It's got dozens of required arguments, and hundreds of potential arguments. Nobody is going to be starting a validator by typing a command by hand in a terminal. So a CLI doesn't seem very useful.

common/config/README.md
cody-littley: This new framework doesn't change our configuration model. The previous framework was just a way to shove a bunch of data into a struct. The new framework does the same, but with less cruft.

With respect to internal vs. external configurations, definitely willing to discuss! I wonder if that problem is something we might solve more by convention rather than framework level support. For example, we could split internal configuration into its own struct that's nested inside the main config struct.

e.g. 

```
const MyAppPrefix = "MYAPP"

type MyAppInternalConfig struct {
    ExperimentalSetting int
}


type MyAppConfig struct {
    EthRPCURL String
    Unsafe MyAppInternalConfig
}
```

This could be configured as follows:

```
// via flags
export MYAPP_ETHRPCURL = "http://..."
export MYAPP_UNSAFE_EXPERIMENTAL_SETTING=42

// via toml
EthRPCURL = "http://..."
Unsafe.ExperimentalSetting = 42
```

common/config/README.md
cody-littley: Open to discussing this. But would it make sense to have that discussion in parallel to this PR? Hoping to get the core of the framework approved by the team without getting bogged down in details we may not all yet agree on. 😅 

The new framework supports both patterns, and doesn't attempt to force the user to of the framework to use one pattern over another. If we do decide to force users into a particular pattern, modifying the framework to enforce it would be a simple change and could be done as a follow up PR.

common/config/README.md
cody-littley: # Env Vars

I'm personally not a fan of env vars, since I prefer there to be one way of doing something. But I'm also not convinced it's simple to drop support in the short term, even if we could get @pschork on board. Main problem is that the technique we use to inject secrets uses environment variables, and I'm worried it might get messy if we try to embed secrets in config files. Possibly not secure as well, since now there is a file with the secret in plaintext floating around.

# Config Files

> The way traffic-generator-v2 helm chart [pipes it's config](https://github.com/Layr-Labs/eigenda-devops/blob/master/charts/traffic-generator-v2/values/eigenda-mainnet-beta/us-east-1/ethereum/values.yaml#L55) into json files is ugly imo.

Yeah, I think the current execution is ugly. But I do wish we used config files in general. I wonder how much of the current ugliness is due to the configuration being in a file vs. the way I hacked together the helm charts (I'm not convinced I did a good job on the helm chart).

If I had to sacrifice the ability to handle config files in order to get this framework PR through, I'd still consider the framework a major improvement over what we currently have.

# Overrides

> do we have actual use cases for overrides? Feels bug prone,

Overrides are hard to prevent if we support both environment variables and config files, due to how the viper framework does its thing. And once we support overrides, I thought "why not lean into it"?

Once again, willing to sacrifice this feature in order to build consensus. Having only flags in this framework would still be a massive improvement over urfave.

# Next Steps

@samlaf @pschork Do you think I should remove support for config files (and therefore overrides)? If we can't come to consensus on config files in the short term, I'm willing to drop it for now and take up the issue later in #da-bike-shed.

common/config/README.md
cody-littley: This new framework supports separate config structs like you describe. Two possible patterns:

1. Use different env var prefixes (e.g. `CONTROLLER` vs `SECRET_CONTROLLER`)
2. Use a parent struct that contains both, then discard the parent struct immediately after you parse config.

common/config/README.md
cody-littley: If we wanted to allow legacy flags to stick around, we'd have to keep that variable inside the struct and mark it as deprecated, kind of like it works with protobufs.

# Alternate Approach

A potential alternate strategy we could bake into the framework would be to pass in a list of strings which we explicitly marking as previously valid but no longer valid. You'd have to add a string to the list any time you deprecated a setting. We could then log an error/warning if we see those flags, but allow the program to continue without crashing.

@samlaf is this a feature you'd like to see?

samlaf: Tbh the more I think about this the more I feel like breaking config changes are absolutely acceptable and shouldn’t be shied away like for library breaking changes. I think this is also how you feel right?

cody-littley: I agree, requiring config updates at upgrade times doesn't seem unreasonable. 

I think the key is going to be to make this transition require as little pain as possible. For big transitions like this will bring, we should have tooling to partially automate the process. For smaller changes, thorough documentation should be sufficient. And we should always be highly available for Q&A to help with the transition.

common/config/README.md
cody-littley: Updated explanation to show both TOML patterns.

common/config/README.md
cody-littley: Added the following paragraph:

```
The `envPrefix` argument is used when parsing environment variables. All environment variables without the specified
prefix are ignored. If `envPrefix` is an empty string, then environment variable parsing is skipped.
```

Also updated the example:
```
// All environment variables will start with "MYAPP_".
const MyAppPrefix = "MYAPP"
```

common/config/README.md
```diff
@@ -0,0 +1,211 @@
+# Configuration Managmenet
```
litt3: Managmenet

common/config/README.md
```diff
@@ -0,0 +1,211 @@
...
+
+```go
+// VerifiableConfig is an interface for configurations that can be validated.
+type VerifiableConfig interface {
+	// Verify checks that the configuration is valid, returning an error if it is not.
+	Verify() error
+}
+```
+
+Although in theory the `Verify()` method can be a no-op, it is highly reccomended to implement basic santity checking.
```
litt3: spellcheck throughout file

common/config/config_parser.go
```diff
@@ -0,0 +1,248 @@
...
+	"reflect"
+	"strings"
+
+	"github.com/Layr-Labs/eigenda/litt/util"
+	"github.com/go-viper/mapstructure/v2"
+	"github.com/spf13/viper"
+)
+
+// ParseConfigFromCLI parses configuration, pulling config paths from command line arguments. Assumes that any command
+// line argument, if present, is a config path. The resulting config is written into target. If there should be default
```
litt3: docs are still stale- `target` is from a previous iteration, as I understand. This applies to other documentation too

common/config/README.md
cody-littley: fixed

common/config/README.md
cody-littley: had claude find and fix

common/config/config_parser.go
cody-littley: fixed

PR #2106, opened 2025-10-02
feat(proxy): read-only mode when no privateKey provided (v2 only)

samlaf:
Proxy was previously forced to be provided with a signerPrivateKey. In order to start in read-only mode, we were passing a bogus privateKey (see our .env.example).

Changed our config/flags/examples to allow not passing a SignerPrivateKey and explicitly marking that flag as optional.

This is a request from Celo, who found a few of their operators were confused and kept asking about this.

Note that this change was only made for EigenDA v2 clients, not V1.

### Notes

I do think adding an explicit --read-only flag would be valuable. If people think this would be better to be more explicit, then I can add it. I opted for keeping it simple, but I think the read-only flag would probably be easy to add. We do plan on moving our config to structured config instead of urfave flags like we current use (see this [proposal](https://github.com/Layr-Labs/eigenda/pull/2104/files)), so I personally think it would make more sense to rethink the proxy config structure then (explicitly separate write/read/verify config params).

### DIFF
--- a/api/proxy/store/generated_key/v2/eigenda.go
+++ b/api/proxy/store/generated_key/v2/eigenda.go
@@ -26,11 +26,16 @@ import (
 type Store struct {
 	log logging.Logger
 
+	// Dispersal related fields. disperser is is optional, and PUT routes will return 500s if not set.
+	disperser *payloaddispersal.PayloadDisperser
 	// Number of times to try blob dispersals:
 	// - If > 0: Try N times total
 	// - If < 0: Retry indefinitely until success
 	// - If = 0: Not permitted
 	putTries int
+
+	// Verification related fields.
+	certVerifier *verification.CertVerifier
 	// Allowed distance (in L1 blocks) between the eigenDA cert's reference block number (RBN)
 	// and the L1 block number at which the cert was included in the rollup's batch inbox.
 	// If cert.L1InclusionBlock > batch.RBN + rbnRecencyWindowSize, an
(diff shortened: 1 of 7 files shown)


litt3 (approved):

iquidus (approved):

api/proxy/store/generated_key/v2/eigenda.go
```diff
@@ -26,32 +26,36 @@ import (
 type Store struct {
 	log logging.Logger
 
+	// Dispersal related fields. disperser is is optional, and PUT routes will return 500s if not set.
```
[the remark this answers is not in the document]
samlaf: [7ee90c5](https://github.com/Layr-Labs/eigenda/pull/2106/commits/7ee90c54f376464341024fb327085f4b53ccbc86)

litt3 (approved):

api/proxy/test/e2e/configuration_test.go
```diff
@@ -0,0 +1,53 @@
...
+	"github.com/Layr-Labs/eigenda/api/proxy/common"
+	"github.com/Layr-Labs/eigenda/api/proxy/test/testutils"
+	"github.com/stretchr/testify/require"
+)
+
+// Tests that a proxy started with V2 EigenDA backend and without a signer private key
+// is in read-only mode, meaning that POST routes return 500 errors, while GET routes work as expected.
+// TODO(samlaf): Feels a bit dumb to run a simple test like this in e2e framework,
+// since it takes 9 seconds, requires an actual eth-rpc (adds ci flakiness), etc.
+// We don't really have an alternative right now however afaiu.
```
ethenotethan: couldn't we just mark the test as "memstore only"?

samlaf: I only implemented the feature for eigendav2 (since it only looks at the `eigenda.v2.signer-private-key` flag).
Alternatively we'd have to add an explicit read-only flag, which could turn off read paths for all the routes (memstore, eigenda, std, etc).

PR #2107, opened 2025-10-02
refactor(relay): Relay server to take in dependencies

dmanc:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/common/geth/multihoming_client.go
+++ b/common/geth/multihoming_client.go
@@ -69,6 +69,11 @@ func NewMultiHomingClient(config EthClientConfig, senderAddress gethcommon.Addre
 func (m *MultiHomingClient) GetRPCInstance() (int, dacommon.EthClient) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
+	if len(m.RPCs) == 0 {
+		m.Logger.Error("[MultiHomingClient] No RPC clients available - please check EthClientConfig.RPCURLs configuration")
+		panic("no RPC clients available")
+	}
+
 	index := m.GetTotalNumberRpcFault() % uint64(len(m.RPCs))
 	if index != m.lastRPCIndex {
 		m.Logger.Info("[MultiHomingClient] Switch RPC", "new index", index, "old index", m.lastRPCIndex)
--- a/inabox/tests/setup_disperser_harness.go
+++ b/inabox/tests/setup_disperser_harness.go
@@ -10,19 +10,12 @@ import (
 	"time"
 
 	"github.com/Layr-Labs/eigenda/common"
-	"github.com/Layr-Labs/eigenda/common/aws/dynamodb"
-	"github.com/Layr-Labs/eigenda/common/aws/s3"
-	coreeth "github.com/Layr-Labs/eigenda/core/eth"
+	"github.com/Layr-Labs/eigenda/common/geth"
 	"github.com/Layr-Labs/eigenda/core/thegraph"
-	v2 "github.com/Layr-Labs/eigenda/core/v2"
-	"github.com/Layr-Labs/eigenda/disperser/common/v2/blobstore"
 	"github.com/Layr-Labs/eigenda/inabox/deploy"
 	"github.com/Layr-Labs/eigenda/relay"
-	"github.com/Layr-Labs/eigenda/relay/chunkstore"
-	"github.com/Layr-Labs/eigenda/relay/limiter"
 	"github.com/Layr-Labs/eigenda/test/testbed"
 	"github.com/Layr-Labs/eigensdk-go/logging"
-	"github.com/prometheus/client_golang/prometheus"
 	"github.com/testcontainers/testcontainers-go"
 )
 
@@ -54,7 +47,7 @@ type DisperserHarness struct {
 	S3Buckets struct {
 		BlobStore string
 	}
-	RelayInstances []*RelayInstance
+	RelayInstances []*relay.Instance
 }
 
 // setupLocalStackResources initializes LocalStack and deploys AWS resources
@@ -116,7 +109,7 @@ func setupDisperserKeypairAndRegistrations(config DisperserHarnessConfig) error 
 // (LocalStack, DynamoDB tables, S3 buckets, relays)
 func SetupDisperserHarness(ctx context.Context, config DisperserHarnessConfig) (*DisperserHarness, error) {
 	harness := &DisperserHarness{
-		RelayInstances: make([]*RelayInstance, 0),
+		RelayInstances: make([]*relay.Instance, 0),
 	}
 
 	// Set default values if not provided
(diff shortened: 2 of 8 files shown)


common/geth/multihoming_client.go
```diff
@@ -69,6 +69,11 @@ func NewMultiHomingClient(config EthClientConfig, senderAddress gethcommon.Addre
 func (m *MultiHomingClient) GetRPCInstance() (int, dacommon.EthClient) {
 	m.mu.Lock()
 	defer m.mu.Unlock()
+	if len(m.RPCs) == 0 {
+		m.Logger.Error("[MultiHomingClient] No RPC clients available - please check EthClientConfig.RPCURLs configuration")
+		panic("no RPC clients available")
```
samlaf:
```suggestion
		m.Logger.Fatal("[MultiHomingClient] No RPC clients available - please check EthClientConfig.RPCURLs configuration")
```

inabox/tests/setup_disperser_harness.go
```diff
@@ -298,160 +280,60 @@ func startRelayWithListener(
...
-	if err != nil {
-		return nil, fmt.Errorf("failed to create dynamodb client: %w", err)
-	}
 
-	s3Client, err := s3.NewClient(ctx, awsConfig, relayLogger)
-	if err != nil {
-		return nil, fmt.Errorf("failed to create s3 client: %w", err)
+	// Create eth client config with RPC URL from deployer
+	ethClientConfig := geth.EthClientConfig{
+		RPCURLs:    []string{config.TestConfig.Deployers[0].RPC},
```
samlaf: why is this under Deployers? Feel like the RPCURLs should live much higher up in the config?

relay/cmd/lib/relay.go
```diff
@@ -3,84 +3,69 @@ package lib
...
+		relay.ServerDependenciesConfig{
+			AWSConfig:                  config.AWS,
+			MetadataTableName:          config.MetadataTableName,
+			BucketName:                 config.BucketName,
+			OperatorStateRetrieverAddr: config.OperatorStateRetrieverAddr,
+			ServiceManagerAddr:         config.EigenDAServiceManagerAddr,
+			ChainStateConfig:           config.ChainStateConfig,
+			EthClientConfig:            config.EthClientConfig,
+			LoggerConfig:               config.Log,
+		},
```
samlaf: why isnt this already a substruct of the config?

relay/server.go
```diff
@@ -163,6 +162,45 @@ func NewServer(
...
 
+// NewInstanceWithDependencies creates and starts a relay server instance with the given dependencies.
+// This is a convenience function that combines NewServer + StartAsInstanceWithListener.
+// The listener must be pre-created by the caller before calling this function.
+func NewInstanceWithDependencies(
+	ctx context.Context,
+	config *Config,
+	deps *ServerDependencies,
+	listener net.Listener,
+) (*Instance, error) {
```
samlaf: I don't think we need this function. I don't think we even need the ServerDependencies struct. Just call NewServer directly and let the dependencies be explicit in the constructor.

relay/server.go
```diff
@@ -163,6 +162,45 @@ func NewServer(
 	}, nil
 }
 
+// NewInstanceWithDependencies creates and starts a relay server instance with the given dependencies.
+// This is a convenience function that combines NewServer + StartAsInstanceWithListener.
```
samlaf: There's too many similar convenience functions. I'd vote for just keeping NewServer. Let the dependencies be constructed and passed at call sites. We don't need NewInstanceWithDependencies or StartAsInstance imo. The listener should prob just be passed to the constructor, and then Start() would use it.

relay/instance.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+// Instance holds the state for a single relay server instance.
+type Instance struct {
+	Server   *Server
+	Listener net.Listener
+	Port     string
+	URL      string
+	Logger   logging.Logger
+}
```
samlaf: why do we need this? There's too many structs flowing around. It should be simple:
1. create config
2. create relay struct
3. start()

The "server" here in instance is not a "server", its like... the whole relay thing.
Is there a way to start/use the relay without the grpc server? If not, then I'd say just stick the server in the Relay (Server) struct itself.

relay/deps.go
```diff
@@ -0,0 +1,107 @@
...
+	OperatorStateRetrieverAddr string
+	ServiceManagerAddr         string
+	ChainStateConfig           thegraph.Config
+	EthClientConfig            geth.EthClientConfig
+	LoggerConfig               common.LoggerConfig
+}
+
+// NewServerDependencies creates all the dependencies needed for a relay server.
+// This function is shared between the CLI entrypoint and test harnesses.
+func NewServerDependencies(
```
samlaf: My personal preference is to not hide away the dependency building in these OOP abstractions, and instead just have it directly inside cmd/lib/relay.go's main start function. Even if its 300 lines its fine, all the dependency construction should be easy to read through / copy-paste around, since most services probably do very similar things.

common/geth/multihoming_client.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2107/commits/a8c93ca0c24cfb0d2bea58b847e45a20ab69c91c

inabox/tests/setup_disperser_harness.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2107/commits/c82c18c7f995840f43ebb5845601d87c83c0de00

I pass the chain harness ethclient now

relay/cmd/lib/relay.go
dmanc: reverted after discussion https://github.com/Layr-Labs/eigenda/pull/2107/commits/a8c93ca0c24cfb0d2bea58b847e45a20ab69c91c

relay/server.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2107/commits/a8c93ca0c24cfb0d2bea58b847e45a20ab69c91c

relay/server.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2107/commits/a8c93ca0c24cfb0d2bea58b847e45a20ab69c91c

relay/instance.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2107/commits/a8c93ca0c24cfb0d2bea58b847e45a20ab69c91c

relay/deps.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2107/commits/a8c93ca0c24cfb0d2bea58b847e45a20ab69c91c

PR #2109, opened 2025-10-03
refactor(validator): Refactor validator server code to allow precreated listeners

dmanc:

Adds listeners to the server struct so that we can precreate them with os-allocated ports in tests.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/inabox/tests/setup_operator_harness.go
+++ b/inabox/tests/setup_operator_harness.go
@@ -77,8 +77,8 @@ func SetupOperatorHarness(
 
 // operatorListeners holds the network listeners for a single operator
 type operatorListeners struct {
-	v1 grpc.V1Listeners
-	v2 grpc.V2Listeners
+	v1 grpc.Listeners
+	v2 grpc.Listeners
 }
 
 // StartOperators starts all operator nodes configured in the test config
@@ -119,12 +119,12 @@ func (oh *OperatorHarness) StartOperators(ctx context.Context, logger logging.Lo
 
 	// Create listeners and start each operator
 	for i := range operatorCount {
-		v1Listeners, err := grpc.CreateV1Listeners("0.0.0.0:0", "0.0.0.0:0")
+		v1Listeners, err := grpc.CreateListeners("0", "0")
 		if err != nil {
 			return fmt.Errorf("failed to create v1 listeners for operator %d: %w", i, err)
 		}
 
-		v2Listeners, err := grpc.CreateV2Listeners("0.0.0.0:0", "0.0.0.0:0")
+		v2Listeners, err := grpc.CreateListeners("0", "0")
 		if err != nil {
 			v1Listeners.Close()
 			return fmt.Errorf("failed to create v2 listeners for opersator %d: %w", i, err)
--- a/node/cmd/main.go
+++ b/node/cmd/main.go
@@ -129,11 +129,11 @@ func NodeMain(ctx *cli.Context, softwareVersion *version.Semver) error {
 	}
 
 	var serverV1 *nodegrpc.Server
-	var v1Listeners nodegrpc.V1Listeners
+	var v1Listeners nodegrpc.Listeners
 	if config.EnableV1 {
-		v1Listeners, err = nodegrpc.CreateV1Listeners(
-			fmt.Sprintf("0.0.0.0:%s", config.InternalDispersalPort),
-			fmt.Sprintf("0.0.0.0:%s", config.InternalRetrievalPort))
+		v1Listeners, err = nodegrpc.CreateListeners(
+			config.InternalDispersalPort,
+			config.InternalRetrievalPort)
 		if err != nil {
 			return fmt.Errorf("failed to create v1 listeners: %w", err)
 		}
@@ -152,13 +152,15 @@ func NodeMain(ctx *cli.Context, softwareVersion *version.Semver) error {
 	}
 
 	var serverV2 *nodegrpc.ServerV2
-	var v2Listeners nodegrpc.V2Listeners
+	var v2Listeners nodegrpc.Listeners
 	if config.EnableV2 {
-		v2Listeners, err = nodegrpc.CreateV2Listeners(
-			fmt.Sprintf("0.0.0.0:%s", config.InternalV2DispersalPort),
-			fmt.Sprintf("0.0.0.0:%s", config.InternalV2RetrievalPort))
+		v2Listeners, err = nodegrpc.CreateListeners(
+			config.InternalV2DispersalPort,
+			config.InternalV2RetrievalPort)
 		if err != nil {
-			v1Listeners.Close()
+			if config.EnableV1 {
+				v1Listeners.Close()
+			}
 			return fmt.Errorf("failed to create v2 listeners: %w", err)
 		}
 
@@ -168,7 +170,9 @@ func NodeMain(ctx *cli.Context, softwareVersion *version.Semver) error {
 			operatorStateRetrieverAddress.Hex(),
 			eigenDAServiceManagerAddress.Hex())
 		if err != nil {
-			v1Listeners.Close()
+			if config.EnableV1 {
+				v1Listeners.Close()
+			}
 			v2Listeners.Close()
 			return fmt.Errorf("cannot create eth.Reader: %w", err)
 		}
@@ -185,7 +189,9 @@ func NodeMain(ctx *cli.Context, softwareVersion *version.Semver) error {
 			v2Listeners.Dispersal,
 			v2Listeners.Retrieval)
 		if err != nil {
-			v1Listeners.Close()
+			if config.EnableV1 {
+				v1Listeners.Close()
+			}
 			v2Listeners.Close()
 			return fmt.Errorf("failed to create server v2: %v", err)
 		}
@@ -193,8 +199,12 @@ func NodeMain(ctx *cli.Context, softwareVersion *version.Semver) error {
 
 	err = nodegrpc.RunServers(serverV1, serverV2, config, logger)
 	if err != nil {
-		v1Listeners.Close()
-		v2Listeners.Close()
+		if config.EnableV1 {
+			v1Listeners.Close()
+		}
+		if config.EnableV2 {
+			v2Listeners.Close()
+		}
 		return fmt.Errorf("failed to start gRPC servers: %w", err)
 	}
 
--- a/node/grpc/listeners.go
+++ b/node/grpc/listeners.go
@@ -5,14 +5,14 @@ import (
 	"net"
 )
 
-// V1Listeners holds the network listeners for v1 gRPC servers.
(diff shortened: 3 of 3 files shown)


node/grpc/server_test.go
```diff
@@ -184,7 +184,26 @@ func newTestServerWithConfig(t *testing.T, mockValidator bool, config *node.Conf
 		Validator:      val,
 		ValidationPool: workerpool.New(1),
 	}
-	return grpc.NewServer(config, node, logger, ratelimiter, version.DefaultVersion())
+
+	// Create listeners with OS-allocated ports for testing
+	v1DispersalListener, err := net.Listen("tcp", "0.0.0.0:0")
+	if err != nil {
+		panic("failed to create v1 dispersal listener")
```
cody-littley: My personal preference for failing a test is `require.NoError(t, err)` instead of panic.

node/grpc/run.go
```diff
@@ -65,11 +57,7 @@ func RunServers(serverV1 *Server, serverV2 *ServerV2, config *node.Config, logge
 			return
 		}
 		for {
-			addr := fmt.Sprintf("%s:%s", localhost, config.InternalV2DispersalPort)
-			listener, err := net.Listen("tcp", addr)
-			if err != nil {
-				logger.Fatalf("Could not start tcp listener: %v", err)
-			}
+			listener := serverV2.dispersalListener
```
cody-littley: In the PR description, you mention we are making this change in order to use OS allocated ports. Is there a reason why we can't update `config` to point to those OS allocated ports instead of adding new constructor arguments?

dmanc: Right now the way we get a port allocated by the OS is to create a listener with port 0.
```go
	v1DispersalListener, err := net.Listen("tcp", "0.0.0.0:0")
```

Not sure of the alternative for getting a random port.

node/grpc/server_test.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2109/commits/53e0f6617289de66c5a0e37b7e4fecc09473904b

inabox/tests/setup_operator_harness.go
```diff
@@ -146,63 +142,155 @@ func (oh *OperatorHarness) StartOperators(ctx context.Context) error {
...
-	for i := 0; i < operatorCount; i++ {
-		instance, err := oh.startOperator(ctx, i)
+	// Now start each operator with its pre-created listeners
+	for i, listeners := range allListeners {
+		server, serverV2, err := oh.startOperator(ctx, logger, i, listeners)
 		if err != nil {
-			// Clean up any operators we started before failing
-			oh.Cleanup(context.Background(), oh.logger)
+			// Clean up any operators we started and all remaining listeners
+			oh.Cleanup(context.Background(), logger)
```
litt3: any reason not to use the `ctx` param here?

inabox/tests/setup_operator_harness.go
```diff
@@ -146,63 +142,155 @@ func (oh *OperatorHarness) StartOperators(ctx context.Context) error {
...
+			oh.closeAllListeners(logger, allListeners[i:])
```
litt3: Is the idea here that any listener already handled will be closed by `oh.Cleanup(context.Background(), logger)`?

litt3 (commented):
Ok, weird that github now requires you to put something in this box

inabox/tests/setup_operator_harness.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2109/commits/79201dcae736d922ec5ce77c55a00b510ded86b1, wasn't actually used so it not longer takes in a context.

inabox/tests/setup_operator_harness.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2109/commits/79201dcae736d922ec5ce77c55a00b510ded86b1

Actually the Stop command wasn't doing anything so it was not cleaning up the listeners appropriately. I made some changes so that the Stop command will clean up the servers listeners. This also meant I had to remove the infinite for loop restart logic in run.go.

```go
		for {
```

node/grpc/run.go
cody-littley: Ah, ok. Not a huge fan of passing in test-only things to the constructor, but I'm ok moving forward with this PR to unblock inabox improvements. We can work on fixing the constructor arguments in the future. (Perhaps some sort of interface responsible for building sockets, and we can pass in a mock implementation?)

cody-littley (dismissed):

inabox/tests/setup_operator_harness.go
```diff
@@ -146,63 +142,144 @@ func (oh *OperatorHarness) StartOperators(ctx context.Context) error {
...
 	}
 
-	oh.logger.Info("Starting operator goroutines", "count", operatorCount)
+	logger.Info("Starting operators", "count", operatorCount)
+
+	// Create listeners and start each operator
+	for i := range operatorCount {
+		listeners, err := oh.createOperatorListeners(i)
+		if err != nil {
+			// Clean up any operators already started (Stop() will close their listeners)
```
litt3: No method `Stop` is being called here, so I'm not sure how to interpret this comment

Also, `closeListeners` is _not_ called by `Cleanup`- did you intend to do that?

node/cmd/main.go
```diff
@@ -128,9 +129,41 @@ func NodeMain(ctx *cli.Context, softwareVersion *version.Semver) error {
 		return err
 	}
 
+	// Create listeners for v1 and v2 servers with configured ports
+	v1DispersalListener, err := net.Listen("tcp", fmt.Sprintf("0.0.0.0:%s", config.InternalDispersalPort))
```
litt3: It seems like there is duplicate logic here, with what's included in `setup_operator_harness`. Should this be part of a utility, to avoid the duplication?

node/cmd/main.go
```diff
@@ -128,9 +129,41 @@ func NodeMain(ctx *cli.Context, softwareVersion *version.Semver) error {
...
 	// TODO(cody-littley): the metrics server is currently started by eigenmetrics, which is in another repo.
 	//  When we fully remove v1 support, we need to start the metrics server inside the v2 metrics code.
-	server := nodegrpc.NewServer(config, node, logger, ratelimiter, softwareVersion)
+	server := nodegrpc.NewServer(
+		config,
+		node,
+		logger,
+		ratelimiter,
+		softwareVersion,
+		v1DispersalListener,
```
litt3: These listeners are only used conditionally, but they're created unconditionally. Consider only creating them when they're actually needed

dmanc: https://github.com/Layr-Labs/eigenda/pull/2109/commits/2606b1d0df5feb13eceb6501ea0d8c2281528a3a

node/cmd/main.go
dmanc: Yup tried to do that here https://github.com/Layr-Labs/eigenda/pull/2109/commits/2606b1d0df5feb13eceb6501ea0d8c2281528a3a

inabox/tests/setup_operator_harness.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2109/commits/2606b1d0df5feb13eceb6501ea0d8c2281528a3a

I reorganized a bit of the logic here, should be simpler I think

node/cmd/main.go
```diff
@@ -128,21 +128,51 @@ func NodeMain(ctx *cli.Context, softwareVersion *version.Semver) error {
...
 	}
 
 	var serverV2 *nodegrpc.ServerV2
+	var v2Listeners nodegrpc.V2Listeners
 	if config.EnableV2 {
+		v2Listeners, err = nodegrpc.CreateV2Listeners(
+			fmt.Sprintf("0.0.0.0:%s", config.InternalV2DispersalPort),
+			fmt.Sprintf("0.0.0.0:%s", config.InternalV2RetrievalPort))
+		if err != nil {
+			v1Listeners.Close()
```
litt3: This will panic if `!config.EnableV1`

(also the other similar calls below)

node/grpc/listeners.go
```diff
@@ -0,0 +1,78 @@
...
+	if err != nil {
+		listeners.Close()
+		return listeners, fmt.Errorf("failed to create v1 retrieval listener: %w", err)
+	}
+
+	return listeners, nil
+}
+
+// V2Listeners holds the network listeners for v2 gRPC servers.
+type V2Listeners struct {
```
litt3: We don't really need to have separate structs for v1 and v2, since they're identical, right?

node/cmd/main.go
```diff
@@ -128,21 +128,51 @@ func NodeMain(ctx *cli.Context, softwareVersion *version.Semver) error {
...
+			v1Listeners.Dispersal,
+			v1Listeners.Retrieval,
+		)
 	}
 
 	var serverV2 *nodegrpc.ServerV2
+	var v2Listeners nodegrpc.V2Listeners
 	if config.EnableV2 {
+		v2Listeners, err = nodegrpc.CreateV2Listeners(
+			fmt.Sprintf("0.0.0.0:%s", config.InternalV2DispersalPort),
```
litt3: Would it potentially make sense to just make the parameters ports, and populate the `0.0.0.0` address within the method? Seems like that's what's used in every case.

node/grpc/run.go
samlaf: Didn't look at how this is actually implemented here, but @cody-littley the port would actually be part of the (external) config! If a user sets port 0 as part of his config, then the OS will spin up the service on a random allocated port. This is most useful for testing, but not strictly for that use case.

The alternative to doing listener injection is to let the constructor create the listener, and then expose a public function `Port()`. I personally think the former is cleaner, but open to bike-shed debating this :)

node/cmd/main.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2109/commits/37ba45d4c829ccbb07b7a4e092297eaac9bee7c1

So I don't think it will actually panic since `v1Listeners` is just an empty struct and Close checks if the internal listeners are null before calling close. I added some safety checks anyway since it makes it more explicit and clear.

node/cmd/main.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2109/commits/37ba45d4c829ccbb07b7a4e092297eaac9bee7c1

node/grpc/listeners.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2109/commits/37ba45d4c829ccbb07b7a4e092297eaac9bee7c1

PR #2113, opened 2025-10-03
feat(inabox): Support manual client construction

litt3:
- Provides utilities for creating `PayloadDispersers` at test time, separate from the rest of the infra stack
- Still leaves the preexisting default `PayloadDisperser`, which will work for many test cases
- Reworks the recently added payment tests, to take advantage of the new utility


inabox/tests/payments/payload_submitter.go
```diff
@@ -10,9 +10,10 @@ import (
...
 	"github.com/Layr-Labs/eigenda/api/clients/v2/payloaddispersal"
 	"github.com/Layr-Labs/eigenda/test/random"
+	"github.com/stretchr/testify/require"
 )
 
-// Submits payloads at a certain rate for a duration.
+// Submits payloads at a certain rate for a duration. Asserts the actual success rate is within tolerance of expected.
 //
 // Returns dispersal errors on a channel.
 func submitPayloads(
```
samlaf:
```suggestion
func mustSubmitPayloads(
```

samlaf: recommend we start using the `must` prefix for test functions that also include some require statements

inabox/tests/setup_infra.go
```diff
@@ -26,11 +25,9 @@ type InfrastructureConfig struct {
 	// Number of relay instances to start, if not specified, no relays will be started.
 	RelayCount int
 
-	// The following fields are temporary, to be able to test different payments configurations. They will be removed
+	// The following field is temporary, to be able to test different payments configurations. It will be removed
 	// once legacy payments are removed.
-	UserReservationSymbolsPerSecond uint64
-	ClientLedgerMode                clientledger.ClientLedgerMode
-	ControllerUseNewPayments        bool
+	ControllerUseNewPayments bool
```
samlaf, line 29: how is the deploy.Config config related to this Infrastructure Config? which is used where? is the other one only for devnet inabox and this one for e2e tests inabox?

inabox/tests/setup_test_harness.go
```diff
@@ -291,255 +287,22 @@ func setupRetrievalClientsForContext(testHarness *TestHarness, infraHarness *Inf
 	return nil
 }
 
-func setupPayloadDisperserForContext(
+func setupDefaultPayloadDisperser(
```
samlaf, line 293: guessing there's another function somewhere to create non default test payload dispersers? Could you add a comment pointing to that function here por favor?

inabox/tests/payments/reservation_only_test.go
```diff
@@ -78,168 +68,183 @@ func testReservationOnly(t *testing.T, clientLedgerMode clientledger.ClientLedge
...
-			if err != nil {
-				failureCount++
-			} else {
-				successCount++
-			}
-		}
+// - Submit blobs at a rate that is supported by the reservation, and assert that all dispersals succeed
+// - Make the reservation smaller
+// - Submit blobs at the same rate, and assert some dispersals fail
+func testWithinThenOver(
```
dmanc: nit: Name of this function seems a bit confusing at first glance. Maybe something like `testReservationReduction`.

inabox/tests/test_harness.go
```diff
@@ -154,3 +183,248 @@ func calculateQuorumSplits(numQuorums int) []byte {
...
+	signer, err := auth.NewLocalBlobRequestSigner(config.PrivateKey)
+	if err != nil {
+		return nil, fmt.Errorf("create blob request signer: %w", err)
+	}
+
+	disperserClientConfig := &clientsv2.DisperserClientConfig{
+		Hostname: "localhost",
+		// TODO(litt3): there needs to be a better way to configure this. I'm not addressing this for now, since this
+		// is how it is currently being done, and dynamic port definition will be a larger change.
+		Port: "32005",
```
dmanc: Could we pass this in as an argument?

inabox/tests/payments/reservation_only_test.go
```diff
@@ -78,168 +68,183 @@ func testReservationOnly(t *testing.T, clientLedgerMode clientledger.ClientLedge
...
+	resultChan = submitPayloads(
+		t, testRandom, payloadDisperser, blobsPerSecond, blobBytes, submissionDuration, 0.5, 0.25)
+	for range resultChan {
+	}
+}
+
+// - Submit blobs at a rate that is larger than the reservation, and assert some dispersals fail
+// - Make the reservation larger
+// - Submit blobs at the same rate, and assert that all dispersals succeed
+func testOverThenWithin(
```
dmanc: nit: similar comment as above, maybe rename to `testReservationOverflow`?

inabox/tests/utils.go
```diff
@@ -15,3 +17,19 @@ func MineAnvilBlocks(t *testing.T, rpcClient common.RPCEthClient, numBlocks int)
...
 	}
 }
+
+// Generates a random ECDSA private key and returns it as a hex string (without the "0x" prefix)
+func GenerateRandomPrivateKeyHex(t *testing.T) string {
+	t.Helper()
+	privateKey, err := crypto.GenerateKey()
+	require.NoError(t, err)
+	return gethcommon.Bytes2Hex(crypto.FromECDSA(privateKey))
+}
```
samlaf: don't we have this function in the random util your brother wrote?

inabox/tests/utils.go
```diff
@@ -15,3 +17,19 @@ func MineAnvilBlocks(t *testing.T, rpcClient common.RPCEthClient, numBlocks int)
...
+
+// Derives the Ethereum address (account ID) from a private key hex string
+func GetAccountIDFromPrivateKeyHex(t *testing.T, privateKeyHex string) gethcommon.Address {
```
samlaf:
```suggestion
func MustGetAccountIDFromPrivateKeyHex(t *testing.T, privateKeyHex string) gethcommon.Address {
```

inabox/tests/payments/reservation_only_test.go
```diff
@@ -78,168 +68,183 @@ func testReservationOnly(t *testing.T, clientLedgerMode clientledger.ClientLedge
...
+	// reservation required to exactly support blobsPerSecond
+	reservationRequiredForRate := float32(minSizeBlobPerSecondReservationSize) * blobsPerSecond
+
+	testRandom := random.NewTestRandom()
+	privateKey := integration.GenerateRandomPrivateKeyHex(t)
+	accountID := integration.GetAccountIDFromPrivateKeyHex(t, privateKey)
+
+	payloadDisperserConfig := integration.GetDefaultTestPayloadDisperserConfig()
+	payloadDisperserConfig.ClientLedgerMode = clientLedgerMode
+	payloadDisperserConfig.PrivateKey = privateKey
```
samlaf: nit: don't use GetDefault if you're changing the only 2 fields in the config

inabox/tests/payments/reservation_only_test.go
```diff
@@ -78,168 +68,183 @@ func testReservationOnly(t *testing.T, clientLedgerMode clientledger.ClientLedge
...
+		uint64(reservationRequiredForRate*2.0),
+		time.Now().Add(-1*time.Hour),
+		time.Now().Add(24*time.Hour),
+		[]core.QuorumID{0, 1},
+	)
+	require.NoError(t, err)
+	registerReservation(t, testHarness, clientReservation, accountID)
 
-		quarter := (successCount + failureCount) / 4
+	payloadDisperser, err := testHarness.CreatePayloadDisperser(t.Context(), logger, payloadDisperserConfig)
```
samlaf, line 134: nit: feel like the CreateTestPayloadDisperser should also create the reservation and/or on-demand deposit? Perhaps extend the TestPayloadDisperserConfig with reservation/on-demand deposit amounts?

inabox/tests/test_harness.go
```diff
@@ -103,30 +116,46 @@ func (tc *TestHarness) Cleanup() {
 	// Most will be garbage collected, but connections will be closed when EthClient is garbage collected
 }
 
-// Updates the reservation for the test account on the PaymentVault contract
-func (tc *TestHarness) UpdateReservationOnChain(ctx context.Context, t *testing.T, r *reservation.Reservation) error {
-	quorumNumbers := r.GetQuorumNumbers()
+// Provides thread-safe access to the deployer TransactOpts.
+//
+// Returns the TransactOpts and an unlock function that MUST be called when done.
+func (tc *TestHarness) GetDeployerTransactOpts() (*bind.TransactOpts, func()) {
```
samlaf, line 128: the returned TransactOpts doesn't actually contain a nonce. The nonce is populated by the ethereum client, so feels a bit weird to have an unlock function on the TransactOpts itself. It works... but its a bit confusing I feel.

We could change DeployerTransactorOpts to have an internal lock with Nonce and increase it locally. That would hide the lock and make it much simpler I feel?

Anyways not strictly necessary but would be good to at least document since I had to go digging in the abigen code for 5 mins to figure this out.

samlaf: speaking of which, feel like we should switch inabox to start using abigen v2. Saw that we only have certVerifier using abigenv2 but not the router.

inabox/tests/payments/reservation_only_test.go
litt3: 782b76f6ffcbdf3237a91b95b574f299d4000343

inabox/tests/payments/reservation_only_test.go
litt3: 782b76f6ffcbdf3237a91b95b574f299d4000343

inabox/tests/payments/payload_submitter.go
litt3: 90b94a28b705803900e1b866914a586b4ac2f3cb

inabox/tests/setup_test_harness.go
litt3: Right, it's the method used by this method to construct the default `PayloadDisperser`.

Added a doc af598548a055f7848270bca3fb0ffe2fd0dbfc2b

inabox/tests/utils.go
litt3: I don't know, is that really helpful? Pretty much _every_ method that accepts a `testing.T` instance is gonna have an assertion of some sort... what are we buying ourselves by prefixing every testing method with `must`?

inabox/tests/payments/reservation_only_test.go
litt3: Why not?

It's redundant in the current state, but if another field is added to the struct, I'd want to use that default here, rather than breaking. That's one of the big benefits of the `get default` pattern

inabox/tests/utils.go
litt3: Not exactly, but close enough

0ae2e566b7afa4d009a681d2317313a1c8ba08b6

inabox/tests/payments/reservation_only_test.go
litt3: The problem with that approach as that the setting of `PaymentVault` conditions is pretty nuanced.

Various components have async monitor threads that look at the `PaymentVault` state, so if you make an update, you need to _either_ make it before the relevant objects are constructed, _or_ you need to sleep some multiple of the poll rate to make sure it's picked up.

The nuance makes designing a general purpose API messy, and the result would invite misusage, IMO. I think simpler is better here

inabox/tests/test_harness.go
litt3: c91c81e31a1f7c3ea83110c45733a472047d06a0

inabox/tests/setup_infra.go
litt3: According to @dmanc, `deploy.Config` is just tech debt, and this `InfrastructureConfig` is what we should be using.

But in this case, it's necessary to pass the value through both, based on the different places where the value is needed

dmanc (dismissed):

inabox/tests/setup_infra.go
dmanc: Yes we still need `deploy.Config` for now since that's what `GenerateAllVariables` function depends on.

inabox/tests/test_harness.go
litt3: > We could change DeployerTransactorOpts to have an internal lock with Nonce and increase it locally. That would hide the lock and make it much simpler I feel?

I originally spun my wheels for a while trying to get this to work, tracking the nonce internally. Probably I was just missing something small, but I ended up going with this simple strategy... it works and I didn't want to spend any more time on this for now.

Added a doc 0de7f37eccc41b6dd0cf7e9df7bc283095c66935

litt3: > speaking of which, feel like we should switch inabox to start using abigen v2. Saw that we only have certVerifier using abigenv2 but not the router.

Agreed, that would be a good change at some point here

inabox/tests/setup_infra.go
samlaf: But GeneratAllVariables is used in devnet inabox right? So it's not really tech debt.. its needed right?

inabox/tests/utils.go
samlaf: fair enough. t could be used for other purposes like logging or whatever, but kinda agree. I guess I wish we minimized the places where we pass testing.T so we can see them in tests though. I personally find it cleaner.

inabox/tests/setup_infra.go
dmanc: Yup it's still needed until we figure out how to consolidate

inabox/tests/payments/reservation_only_test.go
samlaf: Good point! I guess I prefer the pattern where the default config is created by the constructor and overridden by passed config, but I do agree with you this is clean as well.

inabox/tests/payments/reservation_only_test.go
samlaf: But you are already sleeping inside registerReservation no?
I'm saying that testHarness.CreatePayloadDisperser should also do (in the case of reservation being in its config)
```
	clientReservation, err := reservation.NewReservation(
		// reservation larger than it needs to be
		uint64(reservationRequiredForRate*2.0),
		time.Now().Add(-1*time.Hour),
		time.Now().Add(24*time.Hour),
		[]core.QuorumID{0, 1},
	)
	require.NoError(t, err)
	registerReservation(t, testHarness, clientReservation, accountID)
```

I might be missing something...

inabox/tests/test_harness.go
samlaf: that works, thanks!

inabox/tests/setup_infra.go
samlaf: Gotcha, @litt3 can you add a small comment on top of Config to this effect? I know it doesn't really belong in this PR but given we have context here might as well, otherwise we'll forget.

```
// Config is used by devnet inabox, whereas inabox when spun up for tests uses InfrastructureConfig instead.
// TODO: We should eventually find a way to consolidate them.
```

litt3: fbd04345c82d5e0d64f449bb907e20047c8c0a85

inabox/tests/payments/reservation_only_test.go
litt3: `registerReservation` is just a little helper method within this test file, it's not intended to be part of the testing framework. It's needed in this file because we're explicitly testing what happens when reservations change, but the vast majority of tests will just want active reservations, _not_ changing ones. The sleep is not desirable in the standard test case, since it makes tests take longer, and makes them potentially more flaky.

I think the correct place to be registering PaymentVault state will be _before_ constructing the controller and payload disperser(s). We can't do that now, since the infrastructure stack isn't split apart. But for now, I think it makes sense to avoid baking the sleep strategy into the `PayloadDisperser` utilities.

samlaf: That makes sense!

samlaf (dismissed):

samlaf (approved):

PR #2114, opened 2025-10-04
feat(controller) adds oci kms signer

pschork:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


common/oci/object_storage.go
```diff
@@ -0,0 +1,461 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create OCI Object Storage client: %w", err)
+	}
+
+	// Fall back to standard OCI environment variables if application config is empty
+	finalCfg := cfg
+	if finalCfg.Region == "" {
+		if region := os.Getenv("OCI_REGION"); region != "" {
+			finalCfg.Region = region
+		}
```
cody-littley: We should avoid direct environment variable access inside business logic, and instead route all configuration through the config framework.

common/oci/object_storage.go
```diff
@@ -0,0 +1,461 @@
...
+	}
+	if finalCfg.CompartmentID == "" {
+		if compartmentID := os.Getenv("OCI_COMPARTMENT_ID"); compartmentID != "" {
+			finalCfg.CompartmentID = compartmentID
+		}
```
cody-littley: same comment as above, avoid direct env var access

common/oci/object_storage.go
```diff
@@ -0,0 +1,461 @@
...
+	}
+	if finalCfg.Namespace == "" {
+		if namespace := os.Getenv("OCI_NAMESPACE"); namespace != "" {
+			finalCfg.Namespace = namespace
```
cody-littley: same comment as above, avoid direct env var access

common/oci/object_storage.go
```diff
@@ -0,0 +1,461 @@
...
+		err = <-resultChannel
+		if err != nil {
+			return err
+		}
+	}
+	if err := ctx.Err(); err != nil {
+		return fmt.Errorf("context error during fragmented upload: %w", err)
+	}
+	return nil
+}
```
cody-littley, line 278: Note that the fragmented upload/download is a feature I am going to be deleting in the very near future. I'm not saying we can't merge this PR with this feature, just letting you know so you don't spend too long iterating on this part of the API.

The reason why this feature is going away is because relays will now do partial reads instead of always reading the entire file. This was a necessary optimization to reach 1gb/s on preprod.

common/oci/object_storage.go
```diff
@@ -24,12 +23,12 @@ var (
...
-	CompartmentID               string
-	BucketName                  string
-	FragmentParallelismConstant int
-	FragmentParallelismFactor   int
+	Namespace                   string `mapstructure:"namespace"`
+	Region                      string `mapstructure:"region"`
+	CompartmentID               string `mapstructure:"compartment_id"`
+	BucketName                  string `mapstructure:"bucket_name"`
+	FragmentParallelismConstant int    `mapstructure:"fragment_parallelism_constant"`
+	FragmentParallelismFactor   int    `mapstructure:"fragment_parallelism_factor"`
```
cody-littley: What's up with these tags? The `mapstructure` tags are not needed by urfave or by the new framework. Not a blocker, just curious.

pschork: claude hallucinations. removed

common/oci/object_storage.go
```diff
@@ -0,0 +1,450 @@
...
+	}
+
+	if workers == 0 {
+		workers = 1
+	}
+
+	return &ociClient{
+		cfg:                 &finalCfg,
+		objectStorageClient: objectStorageClient,
+		concurrencyLimiter:  make(chan struct{}, workers),
```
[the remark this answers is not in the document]
pschork: lol accepting this caused deadlock https://github.com/Layr-Labs/eigenda/pull/2221/files

PR #2117, opened 2025-10-06
chore(gitignore): path should have been **/testdata/fuzz

samlaf:
**/testdata is used to put testdata inputs that we want to actually commit to the repo.
Corrected to use the /fuzz subdir, which is what I should have used in the first place.


samlaf:
> Are there any unit tests you need to modify to point to this new directory?

No. This was not directories I was manually using in unit tests. I only added **/testdata to ignore the golang generated fuzz directories.

PR #2118, opened 2025-10-06
feat: utility for looking up signing rate using DataAPI

cody-littley:

This is code that will be used by the new ejector. Adds utility methods for looking up signing rates utilizing the DataAPI. Eventually, this data will be sourced from the controller, but in the short term we will continue using the DataAPI.

The code in this PR was modeled after https://github.com/Layr-Labs/eigenda-devops/blob/master/watchdog/common/dataapi_client.go

The returned data type is a protobuf. Out of context, this may might not make sense. The protobuf used to return signing rates is the same protobuf that will eventually be returned by the controller when we start using the controller to look up signing rates. By using the protobuf format now, we can make the eventual switchover to the controller have a very small impact on the ejector's code.

### DIFF
--- a/inabox/Makefile
+++ b/inabox/Makefile
@@ -10,10 +10,8 @@ run-e2e-tests:
 # be removed as soon as we figure out why in-test teardown isn't always succeeding.
 run-payments-tests:
 	@for test in \
-		TestReservationOnly_LegacyClient_LegacyController \
-		TestReservationOnly_LegacyClient_NewController \
-		TestReservationOnly_NewClient_LegacyController \
-		TestReservationOnly_NewClient_NewController; do \
+		TestReservationOnly_LegacyController \
+		TestReservationOnly_NewController; do \
 		echo "Running $$test..."; \
 		if ! go test ./tests/payments -v -run "^$$test$$"; then \
 			echo "❌ TEST FAILED: $$test"; \
--- a/inabox/deploy/config_types.go
+++ b/inabox/deploy/config_types.go
@@ -7,7 +7,6 @@ import (
 	"path/filepath"
 	"strings"
 
-	"github.com/Layr-Labs/eigenda/core/payments/clientledger"
 	gethcommon "github.com/ethereum/go-ethereum/common"
 	"gopkg.in/yaml.v3"
 )
@@ -153,6 +152,8 @@ func (e Environment) IsLocal() bool {
 	return e.Type == "local"
 }
 
+// Config is used by devnet inabox, whereas inabox when spun up for tests uses InfrastructureConfig instead.
+// TODO: We should eventually find a way to consolidate them.
 type Config struct {
 	rootPath string
 
@@ -194,9 +195,7 @@ type Config struct {
 	// DisperserKMSKeyID is the KMS key ID used to encrypt disperser data
 	DisperserKMSKeyID string
 
-	UserReservationSymbolsPerSecond uint64
-	ClientLedgerMode                clientledger.ClientLedgerMode
-	UseControllerMediatedPayments   bool
+	UseControllerMediatedPayments bool
 }
 
 func (env *Config) IsEigenDADeployed() bool {
--- a/inabox/deploy/deploy.go
+++ b/inabox/deploy/deploy.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 	"os"
 	"path/filepath"
-	"strconv"
 	"strings"
 	"time"
 
@@ -73,15 +72,6 @@ func (env *Config) deployEigenDAContracts() error {
 		}
 	}
 
-	envVars := make(map[string]string)
-
-	// If UserReservationSymbolsPerSecond is 0, then the setup script uses the absence of the env var to know that a
-	// default value should be used. This is done for backwards compatibility with preexisting tests that rely on
-	// the default value in the setup script.
-	if env.UserReservationSymbolsPerSecond > 0 {
-		envVars["USER_RESERVATION_SYMBOLS_PER_SECOND"] = strconv.FormatUint(env.UserReservationSymbolsPerSecond, 10)
-	}
-
 	// Create deployment config for testbed
 	deployConfig := testbed.DeploymentConfig{
 		AnvilRPCURL:      deployer.RPC,
@@ -92,7 +82,6 @@ func (env *Config) deployEigenDAContracts() error {
 		MaxOperatorCount: env.Services.Counts.NumMaxOperatorCount,
 		PrivateKeys:      env.convertToTestbedPrivateKeys(),
 		Logger:           logger,
-		EnvVars:          envVars,
 	}
 
 	// Deploy contracts using testbed
--- a/inabox/tests/integration_suite_test.go
+++ b/inabox/tests/integration_suite_test.go
@@ -7,7 +7,6 @@ import (
 	"os"
 	"testing"
 
-	"github.com/Layr-Labs/eigenda/core/payments/clientledger"
 	integration "github.com/Layr-Labs/eigenda/inabox/tests"
 	"github.com/Layr-Labs/eigenda/test"
 	"github.com/Layr-Labs/eigensdk-go/logging"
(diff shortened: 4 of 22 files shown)


ejector/data_api_signing_rate_lookup.go
```diff
@@ -0,0 +1,316 @@
...
+		return nil, fmt.Errorf("error joining URL path with %s and %s: %w", srl.url, path, err)
+	}
+	url, err := url.Parse(urlStr)
+	if err != nil {
+		return nil, fmt.Errorf("error parsing URL: %w", err)
+	}
+	// add query parameters
+	q := url.Query()
+	// end: datetime formatted in "2006-01-02T15:04:05Z"
+	q.Set("end", now.UTC().Format("2006-01-02T15:04:05Z"))
```
litt3: I think you can do this instead of the literal:

`q.Set("end", now.UTC().Format(time.RFC3339))`

(also applies to the v2 method)

ejector/data_api_signing_rate_lookup.go
```diff
@@ -0,0 +1,316 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("error parsing response body: %w", err)
+	}
+
+	// Use a map to combine results from multiple quorums.
+	signingRateMap := make(map[core.OperatorID]*validator.ValidatorSigningRate)
+
+	for _, data := range response.Data {
+
+		if len(quorumSet) > 0 {
```
litt3, line 141: Should you be asserting that quorum set isn't empty at the very beginning of the method? Seems the whole thing is pointless if no quorums are requested

(any changes here should be mirrored for the v2 method as well)

ejector/data_api_signing_rate_lookup.go
```diff
@@ -0,0 +1,315 @@
...
+
+	signedBatches := data.TotalBatches - data.TotalUnsignedBatches
+	unsignedBatches := data.TotalUnsignedBatches
+
+	signingRate := &validator.ValidatorSigningRate{
+		ValidatorId:     validatorID[:],
+		SignedBatches:   uint64(signedBatches),
+		UnsignedBatches: uint64(unsignedBatches),
+		SignedBytes:     uint64(signedBatches),   // Not accurate, but we don't have byte info from DataAPI.
+		UnsignedBytes:   uint64(unsignedBatches), // Not accurate, but we don't have byte info from DataAPI.
```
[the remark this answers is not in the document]
litt3: I was gonna make the same suggestion, but not super confident what's the best strategy here

ejector/signing_rate_lookup.go
```diff
@@ -0,0 +1,34 @@
...
+)
+
+// A tool for looking up signing rates for validators.
+type SigningRateLookup interface {
+	// GetSigningRates returns signing rate information for all validators over the given time span. This method
+	// is not required to return data in any particular order.
+	GetSigningRates(
+		// The time span in the past over which to calculate signing rates.
+		timeSpan time.Duration,
+		// A list of quorums to include. If empty, all quorums are included. If more than one quorum is given,
```
litt3, line 25: >If more than one quorum is given, the results for each quorum is summed together.

Grammar issue with this sentence. But beyond that, I'm struggling to understand what exactly is meant by this

ejector/data_api_signing_rate_lookup.go
litt3: I see now looking at the interface file that empty quorum set has a special meaning. Consider noting that somewhere in this file, since it seems like a bug at first glance

ejector/signing_rate_lookup_test.go
```diff
@@ -0,0 +1,37 @@
...
+	"testing"
+	"time"
+
+	"github.com/Layr-Labs/eigenda/common"
+	"github.com/Layr-Labs/eigenda/core"
+	"github.com/Layr-Labs/eigenda/test"
+	"github.com/stretchr/testify/require"
+)
+
+func TestDataApiLookup(t *testing.T) {
```
litt3, line 14: Is this for manual sanity checking only?

ejector/utils.go
```diff
@@ -0,0 +1,60 @@
...
+
+	totalSignedBatches := rateA.GetSignedBatches() + rateB.GetSignedBatches()
+	var latency uint64
+	if totalSignedBatches > 0 {
+		latency = (rateA.GetSigningLatency()*rateA.GetSignedBatches() +
+			rateB.GetSigningLatency()*rateB.GetSignedBatches()) / totalSignedBatches
+	}
+
+	return &validator.ValidatorSigningRate{
+		ValidatorId:     rateA.GetValidatorId(),
```
litt3, line 40: Would it make sense to check that these are the same for the two inputs?

ejector/utils.go
```diff
@@ -0,0 +1,60 @@
...
+		SignedBatches:   rateA.GetSignedBatches() + rateB.GetSignedBatches(),
+		UnsignedBatches: rateA.GetUnsignedBatches() + rateB.GetUnsignedBatches(),
+		SignedBytes:     rateA.GetSignedBytes() + rateB.GetSignedBytes(),
```
litt3, line 43: nit: you could just use the value you already computed above, `totalSignedBatches`

ejector/data_api_signing_rate_lookup.go
```diff
@@ -0,0 +1,316 @@
...
+		}
+		return nil, fmt.Errorf(
+			"error response (%d) from dataapi: %s",
+			resp.StatusCode,
+			errResp.Error,
+		)
+	}
+
+	var response dataapi.OperatorsNonsigningPercentage
+	err = json.NewDecoder(strings.NewReader(string(respBody))).Decode(&response)
```
litt3: Claude says do this:

```
err = json.Unmarshal(respBody, &response)
```

(also for the v2 method)

litt3 (approved):

ejector/data_api_signing_rate_lookup.go
cody-littley: fixed both

ejector/data_api_signing_rate_lookup.go
cody-littley: added comment

ejector/data_api_signing_rate_lookup.go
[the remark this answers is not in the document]
cody-littley: The reason why I did it this way is so that I could write logic that will work when I switch over to using the controller to provide the actual information.

ejector/signing_rate_lookup.go
cody-littley: fixed

cody-littley: Updated godoc:

````
		// A list of quorums to include. If empty, all quorums are included. If more than one quorum is given,
		// the results for each quorum are "summed" together. That is to say, each validator will only be returned in
		// a single result, and its signing rate will be equal to the sum of its signing rates across the all 
		// given quorums.
		quorums []core.QuorumID,
```

ejector/signing_rate_lookup_test.go
cody-littley: Correct. I'm not sure how to set up inabox to provide dataAPI endpoints, and I'm sure it wouldn't be simple. This unit test allows me to point this at production environments and look at the signing rates from those environments.

ejector/utils.go
cody-littley: added check

ejector/utils.go
cody-littley: done

ejector/data_api_signing_rate_lookup.go
```diff
@@ -130,16 +128,16 @@ func (srl *dataApiSigningRateLookup) getV1SigningRates(
...
 
 	var response dataapi.OperatorsNonsigningPercentage
-	err = json.NewDecoder(strings.NewReader(string(respBody))).Decode(&response)
+	err = json.Unmarshal(respBody, &response)
 	if err != nil {
 		return nil, fmt.Errorf("error parsing response body: %w", err)
 	}
 
-	// Use a map to combine results from multiple quorums.
+	// Use a map to combine results from multiple quorums.©
```
litt3: ©

cody-littley: muscle memory still learning vs code, fixed

PR #2119, opened 2025-10-06
feat: cert verifier inherit signature checker

pakim249CAL:
### Current

The signature verifier is inherited by our Service Manager, and stale stakes forbidden is set to false on some chains. This presents a problem if we want to set this field to true without changing the behavior of verifying previous DA certs previously issued. Because upgrading the ServiceManager would propagate changes across all contracts that depend on it.

### This PR

Changes the cert verifier to inherit the BLSSignatureChecker. This cert verifier would reference itself instead of the ServiceManager, which decouples the Cert Verifier's signature verification logic away from the "centralized" signature verifier. And so this helps with our goal of changing stale stakes to be forbidden without changing behavior of past deployed cert verifiers, and also mitigates the need for a contract upgrade.

### TODO

Some tests seem to break. WHY!??? Suspect that stale stakes forbidden is the likely culprit, but need to investigate more.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/contracts/bindings/EigenDACertVerifier/binding.go
+++ b/contracts/bindings/EigenDACertVerifier/binding.go
@@ -103,9 +103,27 @@ type EigenDATypesV2BlobInclusionInfo struct {
 	InclusionProof  []byte
 }
 
+// IBLSSignatureCheckerNonSignerStakesAndSignature is an auto generated low-level Go binding around an user-defined struct.
+type IBLSSignatureCheckerNonSignerStakesAndSignature struct {
+	NonSignerQuorumBitmapIndices []uint32
+	NonSignerPubkeys             []BN254G1Point
+	QuorumApks                   []BN254G1Point
+	ApkG2                        BN254G2Point
+	Sigma                        BN254G1Point
+	QuorumApkIndices             []uint32
+	TotalStakeIndices            []uint32
+	NonSignerStakeIndices        [][]uint32
+}
+
+// IBLSSignatureCheckerQuorumStakeTotals is an auto generated low-level Go binding around an user-defined struct.
+type IBLSSignatureCheckerQuorumStakeTotals struct {
+	SignedStakeForQuorum []*big.Int
+	TotalStakeForQuorum  []*big.Int
+}
+
 // ContractEigenDACertVerifierMetaData contains all meta data concerning the ContractEigenDACertVerifier contract.
 var ContractEigenDACertVerifierMetaData = &bind.MetaData{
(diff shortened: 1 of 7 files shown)


contracts/src/integrations/cert/EigenDACertVerifier.sol
```diff
@@ -67,20 +68,20 @@ contract EigenDACertVerifier is
...
             revert InvalidSecurityThresholds();
         }
         if (initQuorumNumbersRequired.length == 0 || initQuorumNumbersRequired.length > 256) {
             revert InvalidQuorumNumbersRequired(initQuorumNumbersRequired.length);
         }
         _eigenDAThresholdRegistry = initEigenDAThresholdRegistry;
-        _eigenDASignatureVerifier = initEigenDASignatureVerifier;
         _securityThresholds = initSecurityThresholds;
         _quorumNumbersRequired = initQuorumNumbersRequired;
+        staleStakesForbidden = true;
```
ethenotethan, line 84: can we add a comment explaining that this field lives within the inherited `BLSSignatureChecker` and how setting it to true increases security posture?

ethenotethan (commented):
pretty straight forward and clean - just one comment around documentation

contracts/src/integrations/cert/EigenDACertVerifier.sol
samlaf: Something like
> Set to true by default to force the EigenDAStakeRegistry to sync its operator stakes' view with the Eigenlayer DelegationManager contract at least every `minWithdrawalDelayBlocks`. This syncing is currently done weekly by the EigenDA team running https://github.com/Layr-Labs/avs-sync.

contracts/src/integrations/cert/EigenDACertVerifier.sol
```diff
@@ -147,18 +148,21 @@ contract EigenDACertVerifier is
...
+            daCert,
+            _securityThresholds,
+            _quorumNumbersRequired
         );
     }
 
-    /// @inheritdoc IEigenDACertVerifier
-    function eigenDAThresholdRegistry() external view returns (IEigenDAThresholdRegistry) {
-        return _eigenDAThresholdRegistry;
+    function eigenDASignatureVerifier() external view returns (IEigenDASignatureVerifier) {
```
samlaf, line 159: add a comment to mark this deprecated

contracts/bindings/EigenDACertVerifier/binding.go
samlaf, line 1: hmm why are we still generated abigen v1 bindings for CertVerifier. Are we still using this anywhere? cc @ethenotethan 
If not let's get rid of this

PR #2120, opened 2025-10-06
chore: move operator setup docs into core repo

cody-littley:
Moves operator setup documentation from its old repo to the core repo. Old location: https://github.com/Layr-Labs/eigenda-operator-setup


cody-littley:
This is more complicated than I thought, and will require updates to the documentation repo

PR #2122, opened 2025-10-06
chore(inabox): Remove InMemoryBlobStore option

dmanc:

Removing this option for now. Will add back if we find a use case for running in memory blobstore on inabox.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/inabox/tests/setup_disperser_harness.go
+++ b/inabox/tests/setup_disperser_harness.go
@@ -147,7 +147,6 @@ func SetupDisperserHarness(
 	harness.DynamoDBTables.BlobMetaV2 = config.MetadataTableNameV2
 	harness.S3Buckets.BlobStore = config.S3BucketName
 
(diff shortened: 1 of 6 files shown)


inabox/tests/setup_disperser_harness.go
```diff
@@ -149,29 +148,24 @@ func SetupDisperserHarness(
 	harness.S3Buckets.BlobStore = config.S3BucketName
 
 	// Setup LocalStack if not using in-memory blob store
```
samlaf:
```suggestion
```

samlaf (approved):
LGTM. Where is InMemoryBlobStore even located? Is that an option on the disperser? Or do we have separate binary that we run for this? Should we also delete that code, or is it used somewhere?

PR #2123, opened 2025-10-06
chore(inabox): Remove docker compose file generation

dmanc:

This is currently not used by anyone. In the future we will just maintain a docker compose file by hand rather than generating it.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


inabox/grafana-agent/agent-config.yaml
samlaf, line 1: shouldn't we keep this around and spin it up in inabox-devnet?

samlaf (approved):
gud change

inabox/grafana-agent/agent-config.yaml
dmanc: I'll add it back in the future

PR #2126, opened 2025-10-07
fix(sigmap-EDA-02): Missing context cancellation

iquidus:
Closes DAINT-776

Addresses EDA-02 from the [Sigma Prime audit](https://linear.app/eigenlabs/issue/DAINT-774/address-audit-findings) report by modifying the chunk collection loop to use a select statement that respects context cancellation.

The chunk collection loop in `RetrieveBlobChunks()` performs unbuﬀered channel reads without checking context cancellation, causing goroutine leaks and preventing timely cleanup when blob retrieval operations are cancelled.

### DIFF
--- a/api/clients/retrieval_client.go
+++ b/api/clients/retrieval_client.go
@@ -199,6 +199,9 @@ func (r *retrievalClient) RetrieveBlobChunks(ctx context.Context,
 		case <-ctx.Done():
 			return nil, fmt.Errorf("context done: %w", ctx.Err())
 		case reply := <-chunksChan:
+			if ctx.Err() != nil {
+				return nil, fmt.Errorf("context done: %w", ctx.Err())
+			}
 			if reply.Err != nil {
 				r.logger.Warn("failed to get chunks from operator", "operator", reply.OperatorID.Hex(), "err", reply.Err)
 				continue


api/clients/retrieval_client.go
```diff
@@ -195,26 +195,30 @@ func (r *retrievalClient) RetrieveBlobChunks(ctx context.Context,
...
-		if reply.Err != nil {
-			r.logger.Warn("failed to get chunks from operator", "operator", reply.OperatorID.Hex(), "err", reply.Err)
-			continue
-		}
-		assignment, ok := assignments[reply.OperatorID]
-		if !ok {
-			return nil, fmt.Errorf("no assignment to operator %s", reply.OperatorID.Hex())
+		select {
+		case <-ctx.Done():
+			return nil, fmt.Errorf("context cancelled whilst retrieving chuncks: %w", ctx.Err())
```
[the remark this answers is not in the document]
samlaf: this is true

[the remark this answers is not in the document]
samlaf: also not necessarily cancelled, could just have timed out

api/clients/retrieval_client.go
```diff
@@ -195,26 +195,30 @@ func (r *retrievalClient) RetrieveBlobChunks(ctx context.Context,
...
+			err = r.verifier.VerifyFrames(reply.Chunks, assignment.GetIndices(), blobHeader.BlobCommitments, encodingParams)
+			if err != nil {
+				r.logger.Warn("failed to verify chunks from operator", "operator", reply.OperatorID.Hex(), "err", err)
+				continue
+			} else {
+				r.logger.Info("verified chunks from operator", "operator", reply.OperatorID.Hex())
+			}
+
+			chunks = append(chunks, reply.Chunks...)
+			indices = append(indices, assignment.GetIndices()...)
```
samlaf, line 223: I personally don't like code indentation like this for something that is run on every loop. Would prefer to just do
```
for ... {
        select {
        case <-ctx.Done():
            return ctx.Err()
        default:
        }
       // proceed with normal unindented computation
}
```
at the top

iquidus: updated in [c77e46e](https://github.com/Layr-Labs/eigenda/pull/2126/commits/c77e46ee95422c21c96f576f410f3ae0653251f7)

api/clients/retrieval_client.go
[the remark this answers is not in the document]
iquidus: updated in [c77e46e](https://github.com/Layr-Labs/eigenda/pull/2126/commits/c77e46ee95422c21c96f576f410f3ae0653251f7)

api/clients/retrieval_client.go
ethenotethan: @samlaf I think the idea here is to catch the context timeout when waiting on `reply := <-chunksChan` which is why it should happen under the `select` and not beneath. otherwise the context could timeout while `reply := <-chunksChan` never completes

samlaf: ehhhh good point idk how I missed that, my bad @iquidus

samlaf: @iquidus you'll hae to change it back to
```
select {
   case <-ctx.Done():
	return nil, fmt.Errorf("context done: %w", ctx.Err())
   case reply := <-chunksChan:
      // ...
}
```

samlaf: ok soo... thought about this more. It's just v1 code so I don't care, but it's actually been a good learning experience.
Turns out select will choose a branch uniformly at random if both are ready, so the select will take twice as long on average than it could/should. Asked claude and it proposed 2 solutions:
image
image
image

iquidus: Oh this is very interesting.

PR #2128, opened 2025-10-07
feat: config docs

cody-littley:

Create a tool for automatically generating documentation for config structs. Wire this tool up to work with the traffic generator config.

Things intentionally not done as part of this PR:
- wiring up other config structs
- actually using the new config framework for the traffic generator
- a CI action to enforce that config documentation is kept up to date


test/v2/load/config_docs/main.go
```diff
@@ -0,0 +1,41 @@
...
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+
+	"github.com/Layr-Labs/eigenda/common/config"
+	"github.com/Layr-Labs/eigenda/test/v2/load"
+)
+
+func main() {
```
litt3: Help me understand how this pattern will be applied across the codebase. Will every package that uses the config framework require this kind of main function, and a make command to generate the docs?

test/v2/Makefile
```diff
@@ -15,3 +15,7 @@ clean:
 
 test:
 	cd live && go test
+
+document-config:
```
litt3: Do we need this lower level make command?

I'm not sure what standard practice is here- is it normal to expose the same task at multiple tiers of the `make` command structure?

common/config/config_document_generator.go
```diff
@@ -0,0 +1,543 @@
...
+				escapeMarkdown(f.EnvVar),
+				escapeMarkdown(f.FieldType),
+				escapeMarkdown(reformatGodoc(f.Godoc))))
+		}
+		sb.WriteString("\n")
+	}
+
+	if len(optionalFields) > 0 {
+		sb.WriteString("## Optional Fields\n\n")
+		sb.WriteString("| Name | Info | Description |\n")
```
litt3:
```suggestion
		sb.WriteString("| Name | Type + Default | Description |\n")
```

Not sure how severely limited you are here on line width, but I think it would be nice to have the column label be a bit more precise

common/config/config_document_generator.go
```diff
@@ -0,0 +1,543 @@
...
+
+	// Write the markdown document.
+
+	sb.WriteString("<!-- Code generated by config_document_generator.go. DO NOT EDIT BY HAND. -->\n\n")
+
+	sb.WriteString(fmt.Sprintf("# %s Configuration\n\n", componentName))
+
+	if len(requiredFields) > 0 {
+		sb.WriteString("## Required Fields\n\n")
+		sb.WriteString("| Name | Info | Description |\n")
```
litt3:
```suggestion
		sb.WriteString("| Name | Type | Description |\n")
```

Any reason this has to be the generic label "Info"? Won't the cells in this column always be types?

common/config/config_document_generator.go
```diff
@@ -0,0 +1,543 @@
...
+			fields = append(fields, nestedFieldData...)
+		case reflect.Ptr:
+			// Handle pointer to struct
+			// nolint:nestif
+			if field.Type.Elem().Kind() == reflect.Struct {
+				fieldValue := targetValue.Field(i)
+
+				// Use the actual pointer value if it's not nil, otherwise use the actual field which may be nil
+				var nestedValue interface{}
+				if !fieldValue.IsNil() {
```
litt3: Not sure what you're trying to do here- both cases are identical

common/config/config_document_generator.go
```diff
@@ -0,0 +1,543 @@
...
+	target any,
+	envVarPrefix string,
+	tomlPrefix string,
+	packagePaths []string,
+) ([]*configFieldData, error) {
+
+	// Handle pointer to struct
+	targetValue := reflect.ValueOf(target)
+	if targetValue.Kind() == reflect.Ptr {
+		targetValue = targetValue.Elem()
```
litt3: claude claims this will break if the default value is `nil`

litt3 (approved):

common/config/config_document_generator.go
```diff
@@ -0,0 +1,543 @@
...
+// Only one tag value is allowed per field.
+func parseDocsTag(tag string) (required bool, deprecated bool, unsafe bool) {
+	if tag == "" {
+		return false, false, false
+	}
+
+	// Trim whitespace for flexibility
+	tag = strings.TrimSpace(tag)
+
+	switch tag {
```
litt3, line 242: Would it make sense to throw an error if an unexpected tag is encountered? Seems such a case would most likely indicate a typo in the tag

test/v2/load/config_docs/main.go
cody-littley: refactored this as discussed on slack call

test/v2/Makefile
cody-littley: There is now exactly one make command to build all config docs

common/config/config_document_generator.go
cody-littley: Updated this column name.

common/config/config_document_generator.go
cody-littley: updated this column name

common/config/config_document_generator.go
cody-littley: fixed

common/config/config_document_generator.go
cody-littley: fixed

common/config/config_document_generator.go
cody-littley: good idea, done

common/config/bootstrap_test/main.go
```diff
@@ -16,6 +16,21 @@ type TestConfig struct {
 	D time.Duration
 }
 
+// GetEnvVarPrefix implements config.DocumentedConfig.
+func (t *TestConfig) GetEnvVarPrefix() string {
+	panic("TEST")
```
litt3: I'm confused, why are you panicking here, instead of just returning the value to provide the example?

common/config/bootstrap.go
```diff
@@ -121,6 +110,20 @@ func buildHandler[T VerifiableConfig](
 			startPprofServer(logger, pprofPort)
 		}
 
+		// Pre-build the default config to get its env var prefix, then create a new constructor
+		// that returns this pre-built config.
+		defaultConfig := constructor()
+		prefix := defaultConfig.GetEnvVarPrefix()
+		constructor := func() T {
```
litt3: Would it make sense to just pass the default config instance you've already constructed into `ParseConfig`, instead of this redefinition of the `constructor` variable you're currently doing? I.e. the first param `ParseConfig` would just be the default struct

It seems a bit roundabout in the current form, but maybe I'm missing something.

common/config/config_document_generator.go
```diff
@@ -0,0 +1,547 @@
...
+	default:
+		return false, false, false, fmt.Errorf("invalid docs tag value %q", tag)
+	}
+	return required, deprecated, unsafe, nil
+}
+
+func gatherConfigFieldData(
+	target any,
+	envVarPrefix string,
+	tomlPrefix string,
```
litt3, line 258: Is this param necessary?

common/config/bootstrap_test/main.go
cody-littley: Oops, just forgot to fill these out. Fixed.

common/config/config_document_generator.go
cody-littley: Yes. Added the following comment to help clarify:

```
"", // toml prefix used for recursion, top-level has no prefix
```

common/config/bootstrap.go
cody-littley: done

PR #2129, opened 2025-10-07
Feat/config registry rework

pakim249CAL:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
```diff
@@ -51,34 +68,12 @@ library ConfigRegistryLib {
...
-            if (getKey(nameSet.nameList[i]) == key) {
-                nameSet.nameList[i] = nameSet.nameList[nameSet.nameList.length - 1];
-                nameSet.nameList.pop();
-                break;
-            }
+        if (bytes(nameSet.names[key]).length == 0) {
+            require(bytes(name).length > 0, "Name cannot be empty");
+            nameSet.names[key] = name;
+            nameSet.nameList.push(name);
         }
```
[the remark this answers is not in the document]
pakim249CAL: This is okay because configs are not removable anymore. So duplicate adds cannot happen as long as the name's length is > 0, and this check is enforced upon any configuration addition.

contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
```diff
@@ -2,20 +2,41 @@
...
+        uint256 activationKey;
+        bytes value;
+    }
+
+    /// @notice Struct to hold all bytes32 configuration checkpoints and associated names
+    /// @dev See docs for the structs for more information
     struct Bytes32Cfg {
-        mapping(bytes32 => bytes32) values;
-        mapping(bytes32 => string) extraInfo;
+        mapping(bytes32 => Bytes32Checkpoint[]) values;
```
samlaf, line 34: nit: prob good to add a comment explaining why we use an array instead of a mapping from activationKey->value, which is that we want to be able to see the whole array, and that we typically care about last elements only, which means we can walk back and find the value of interest cheaply in most cases.

contracts/test/unit/ConfigRegistryUnit.t.sol
```diff
@@ -6,59 +6,4 @@ import {Test} from "lib/forge-std/src/Test.sol";
...
-        for (uint256 i; i < layout.bytesConfig.nameSet.nameList.length; i++) {
-            if (keccak256(bytes(layout.bytesConfig.nameSet.nameList[i])) == keccak256(bytes(key))) {
-                found = true;
-                break;
-            }
-        }
-        assertFalse(found, "Key was not deregistered");
-    }
-}
+contract ConfigRegistryUnit is Test {}
```
samlaf, line 9: are you planning to add tests for the checkpointed types? Or does this just not apply anymore?

contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
```diff
@@ -2,20 +2,41 @@
...
         NameSet nameSet;
     }
 
+    /// @notice Struct to hold all bytes configuration checkpoints and associated names
+    /// @dev See docs for the structs for more information
     struct BytesCfg {
-        mapping(bytes32 => bytes) values;
-        mapping(bytes32 => string) extraInfo;
```
samlaf, line 18: why aren't we keeping the extraInfo around? Did you change your mind about the usefulness of this?

contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
```diff
@@ -2,20 +2,41 @@
...
+        mapping(bytes32 => BytesCheckpoint[]) values;
```
samlaf, line 41: we already have values inserted in this mapping right? How will the upgrade process work?

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
```diff
@@ -51,34 +64,12 @@ library ConfigRegistryLib {
...
-        // Here we utilize a simple swap and pop to remove the name from the list.
-        // There is no guarantee of preservation of ordering.
-        for (uint256 i; i < nameSet.nameList.length; i++) {
-            if (getKey(nameSet.nameList[i]) == key) {
-                nameSet.nameList[i] = nameSet.nameList[nameSet.nameList.length - 1];
-                nameSet.nameList.pop();
-                break;
-            }
+        if (bytes(nameSet.names[key]).length == 0) {
+            require(bytes(name).length > 0, "Name cannot be empty");
```
samlaf, line 76: shouldnt this check be done as the very first thing even before calling `getKey(name)`?

contracts/src/core/interfaces/IEigenDADirectory.sol
```diff
@@ -48,72 +50,39 @@ interface IEigenDAConfigRegistry {
     error ConfigAlreadyExists(string name);
     error ConfigDoesNotExist(string name);
 
-    event ConfigBytes32Added(string name, bytes32 indexed key, bytes32 value, string extraInfo);
-    event ConfigBytesAdded(string name, bytes32 indexed key, bytes value, string extraInfo);
-    event ConfigBytes32Replaced(string name, bytes32 indexed key, bytes32 oldValue, bytes32 newValue, string extraInfo);
-    event ConfigBytesReplaced(string name, bytes32 indexed key, bytes oldValue, bytes newValue, string extraInfo);
-    event ConfigBytes32Removed(string name, bytes32 indexed key);
-    event ConfigBytesRemoved(string name, bytes32 indexed key);
```
samlaf, line 56: getting rid of events b/c everything is onchain now? Might still be useful to keep in case someone wants to write a subgraph to index the values? Not sure..

contracts/src/core/interfaces/IEigenDADirectory.sol
```diff
@@ -48,72 +50,39 @@ interface IEigenDAConfigRegistry {
...
-    function getConfigBytesExtraInfo(string memory name) external view returns (string memory);
+    function getConfigBytes(bytes32 key, uint256 index) external view returns (bytes memory);
 
-    /// @notice Gets the bytes32 configuration entry by keccak256 hash of the name.
-    function getConfigBytes32(bytes32 key) external view returns (bytes32);
+    function getActivationKeyBytes32(bytes32 key, uint256 index) external view returns (uint256);
 
-    /// @notice Gets the bytes configuration entry by keccak256 hash of the name.
-    function getConfigBytes(bytes32 key) external view returns (bytes memory);
+    function getActivationKeyBytes(bytes32 key, uint256 index) external view returns (uint256);
```
samlaf: What's the plan for how to use these offchain?
Afaiu the getter function we would want to use is something like the router, where instead of passing an index, we pass an RBN (or timestamp), and the getter function walks the list and find the active value at that "referenceKey".

samlaf: This also randomly made me wonder.. do we really need to checkpoint by block numbers at all? Or could we only use timestamps like cafe does and then any offchain user just needs to convert the RBN to a timestamp before calling the configRegistry? This would standardize the config and then we could use "activation/reference time" instead of "activationKey". Thoughts? cc @litt3

contracts/test/unit/ConfigRegistryUnit.t.sol
pakim249CAL: I will add tests today. But as far as implementation goes, I think it is ready for a review.

I deleted the old tests because it is probably better to start from scratch and write the tests again from first principles with the scope of changes that happened.

contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
pakim249CAL: @litt3 i think proposed removing this. But I also am in favor of not really using this anymore, because it is a poorly defined field that expands the API and code surface. My proposal is that we can consider adding this as a separate feature in the future.

contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
pakim249CAL: There is no upgrade because this contract was never deployed yet.

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
pakim249CAL: Yes, that makes sense logically. However, I don't think it's ideal to be increasing the gas cost for any view getters to this call, and I don't think there's any real security concerns as long as this check is enforced upon name registration. Can you think of a security issue with doing this check here instead of the more load bearing function?

contracts/src/core/interfaces/IEigenDADirectory.sol
pakim249CAL: No, the events are just a lot less now because there is only additions to the configs allowed. They are defined in the library (since this allows events to be tested from the library in isolation which is a TODO)

contracts/src/core/interfaces/IEigenDADirectory.sol
pakim249CAL: See the MockConfigRetriever for examples on how we can implement a more complex API on top of this simple one.

contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
litt3: I don't recall discussing this- but I agree it doesn't really seem necessary

contracts/src/core/interfaces/IEigenDADirectory.sol
```diff
@@ -42,78 +44,95 @@ interface IEigenDAAddressDirectory {
 
 /// @title IEigenDAConfigRegistry
 /// @notice Interface for a configuration registry that allows adding, replacing, removing, and retrieving configuration entries by name.
```
litt3: interface doc is no longer accurate, given the redesign

contracts/src/core/interfaces/IEigenDADirectory.sol
```diff
@@ -42,78 +44,95 @@ interface IEigenDAAddressDirectory {
...
+    /// @param name The name of the configuration entry.
+    /// @param activationKey The activation key for the configuration entry.
+    ///                      This is an arbitrary key defined by the caller to indicate when the configuration should become active.
+    /// @param value The variable length byte configuration value.
+    function addConfigBytes(string memory name, uint256 activationKey, bytes memory value) external;
+
+    /// @notice Gets the number of checkpoints for a 32 byte configuration entry.
+    /// @param key The key of the configuration entry.
+    /// @return The number of checkpoints for the configuration entry.
+    function getNumCheckpointsBytes32(bytes32 key) external view returns (uint256);
```
litt3: I think you mentioned calling the parameter `nameDigest` instead of `key`, or something like that? I do think it would be a helpful change, since there are multiple keys within the registry structure

If these parameters are renamed, there are also some method names that should probably be changed to match

contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
```diff
@@ -2,20 +2,41 @@
...
 library ConfigRegistryTypes {
+    /// @notice Struct to keep track of names associated with config keys
+    /// @param names Mapping from config key to name
+    /// @param nameList List of all config names
     struct NameSet {
         mapping(bytes32 => string) names;
         string[] nameList;
     }
 
+    /// @notice Structs to represent checkpoints for fixed-size byte32 configurations
```
litt3:
```suggestion
    /// @notice Struct to represent checkpoints for fixed-size byte32 configurations
```

contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
```diff
@@ -2,20 +2,41 @@
...
+    /// @param activationKey The activation key (e.g., block number or timestamp) for the checkpoint
+    /// @param value The bytes32 configuration value at this checkpoint
+    struct Bytes32Checkpoint {
+        uint256 activationKey;
+        bytes32 value;
+    }
+
+    /// @notice Structs to represent checkpoints for variable-size bytes configurations
```
litt3:
```suggestion
    /// @notice Struct to represent checkpoints for variable-size bytes configurations
```

pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/2129/commits/acf5c318e7fbc72a7486c3cfd3603628b3f8a46e

contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/2129/commits/acf5c318e7fbc72a7486c3cfd3603628b3f8a46e

contracts/src/core/interfaces/IEigenDADirectory.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/2129/commits/acf5c318e7fbc72a7486c3cfd3603628b3f8a46e

contracts/src/core/interfaces/IEigenDADirectory.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/2129/commits/acf5c318e7fbc72a7486c3cfd3603628b3f8a46e

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
```diff
@@ -65,24 +69,24 @@ library ConfigRegistryLib {
     }
 
     function registerKey(T.NameSet storage nameSet, string memory name) internal {
```
litt3:
```suggestion
    function registerName(T.NameSet storage nameSet, string memory name) internal {
```

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
```diff
@@ -65,24 +69,24 @@ library ConfigRegistryLib {
...
             require(bytes(name).length > 0, "Name cannot be empty");
-            nameSet.names[key] = name;
+            nameSet.names[nameDigest] = name;
             nameSet.nameList.push(name);
         }
     }
 
-    function isKeyRegistered(T.NameSet storage nameSet, bytes32 key) internal view returns (bool) {
-        return bytes(nameSet.names[key]).length > 0;
+    function isKeyRegistered(T.NameSet storage nameSet, bytes32 nameDigest) internal view returns (bool) {
```
litt3:
```suggestion
    function isNameDigestRegistered(T.NameSet storage nameSet, bytes32 nameDigest) internal view returns (bool) {
```

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
```diff
@@ -65,24 +69,24 @@ library ConfigRegistryLib {
...
+        return bytes(nameSet.names[nameDigest]).length > 0;
     }
 
-    function isKeyRegisteredBytes32(bytes32 key) internal view returns (bool) {
-        return isKeyRegistered(S.layout().bytes32Config.nameSet, key);
+    function isKeyRegisteredBytes32(bytes32 nameDigest) internal view returns (bool) {
```
litt3:
```suggestion
    function isNameDigestRegisteredBytes32(bytes32 nameDigest) internal view returns (bool) {
```

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
```diff
@@ -65,24 +69,24 @@ library ConfigRegistryLib {
...
+        return isKeyRegistered(S.layout().bytes32Config.nameSet, nameDigest);
     }
 
-    function isKeyRegisteredBytes(bytes32 key) internal view returns (bool) {
-        return isKeyRegistered(S.layout().bytesConfig.nameSet, key);
+    function isKeyRegisteredBytes(bytes32 nameDigest) internal view returns (bool) {
```
litt3:
```suggestion
    function isNameDigestRegisteredBytes(bytes32 nameDigest) internal view returns (bool) {
```

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
```diff
@@ -65,24 +69,24 @@ library ConfigRegistryLib {
...
+        return isKeyRegistered(S.layout().bytesConfig.nameSet, nameDigest);
     }
 
     function getNumRegisteredKeysBytes32() internal view returns (uint256) {
```
litt3:
```suggestion
    function getNumRegisteredNamesBytes32() internal view returns (uint256) {
```

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
litt3:
```suggestion
    function getNumRegisteredNamesBytes() internal view returns (uint256) {
```

litt3:
```suggestion
    function getRegisteredNameBytes32(uint256 index) internal view returns (string memory) {
```

litt3:
```suggestion
    function getRegisteredNameBytes(uint256 index) internal view returns (string memory) {
```

contracts/src/core/EigenDADirectory.sol
```diff
@@ -130,7 +130,7 @@ contract EigenDADirectory is IEigenDADirectory {
 
     /// @inheritdoc IEigenDAConfigRegistry
     function addConfigBytes(string memory name, uint256 activationKey, bytes memory value) external onlyOwner {
-        bytes32 key = ConfigRegistryLib.getKey(name);
+        bytes32 key = ConfigRegistryLib.getNameDigest(name);
```
litt3:
```suggestion
        bytes32 nameDigest = ConfigRegistryLib.getNameDigest(name);
```

pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/2129/commits/517950f0b0bf535ab0eb42dffbd02d097f9db563

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/2129/commits/d5a535e52958008a44661350e85c491c4df77d48

pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/2129/commits/d5a535e52958008a44661350e85c491c4df77d48

pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/2129/commits/d5a535e52958008a44661350e85c491c4df77d48

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/2129/commits/d5a535e52958008a44661350e85c491c4df77d48

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/2129/commits/d5a535e52958008a44661350e85c491c4df77d48

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/2129/commits/d5a535e52958008a44661350e85c491c4df77d48

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/2129/commits/d5a535e52958008a44661350e85c491c4df77d48

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/2129/commits/d5a535e52958008a44661350e85c491c4df77d48

contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
```diff
@@ -2,20 +2,41 @@
 pragma solidity ^0.8.9;
 
 library ConfigRegistryTypes {
+    /// @notice Struct to keep track of names associated with config keys
```
litt3:
```suggestion
    /// @notice Struct to keep track of names associated with name digests
```

I assume that's how it should be worded? Or is something else meant here?

contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
```diff
@@ -2,20 +2,41 @@
...
+    /// @param names Mapping from config key to name
```
litt3:
```suggestion
    /// @param names Mapping from name digest to name
```

contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
pakim249CAL: > which is that we want to be able to see the whole array,

This comment makes sense to me but

> that we typically care about last elements only, which means we can walk back and find the value of interest cheaply in most cases.

This might not necessarily be true, right? For example, in cert verification, it is likely for older certs that we will be using a fixed activation key or index to map to a config value. Seems a bit too prescriptive to me.

contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/2129/commits/4a219ede129609775cb5f9a9545d0925d08192af

contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/2129/commits/4a219ede129609775cb5f9a9545d0925d08192af

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
```diff
@@ -5,108 +5,129 @@ import {ConfigRegistryStorage as S} from "src/core/libraries/v3/config-registry/
...
-                nameSet.nameList.pop();
-                break;
-            }
+    function registerName(T.NameSet storage nameSet, string memory name) internal {
+        bytes32 nameDigest = getNameDigest(name);
+        if (bytes(nameSet.names[nameDigest]).length == 0) {
+            require(bytes(name).length > 0, "Name cannot be empty");
+            nameSet.names[nameDigest] = name;
+            nameSet.nameList.push(name);
         }
```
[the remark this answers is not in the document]
pakim249CAL: The desired behavior is for adding configs to try to register a name if the config is new, and otherwise do nothing. This way of writing the code accomplishes this, but maybe we can consider alternatives, but I don't think it's necessary to explore this right now because this can be accomplished in an upgrade without affecting the storage model.

PR #2130, opened 2025-10-07
feat(inabox): Spin up controller as a goroutine in inabox integration tests 

dmanc:

Spins up controller service as a goroutine inside inabox integration tests.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/inabox/tests/setup_disperser_harness.go
+++ b/inabox/tests/setup_disperser_harness.go
@@ -43,11 +43,17 @@ type DisperserHarnessConfig struct {
 	LocalStackPort string
 
 	// LocalStack resources for blobstore and metadata store
-	MetadataTableName   string
-	BucketTableName     string
-	S3BucketName        string // S3 bucket name for blob storage
+	MetadataTableName string
+	BucketTableName   string
+
+	// S3 bucket name for blob storage
+	S3BucketName string
+
+	// V2 metadata table name
 	MetadataTableNameV2 string
-	OnDemandTableName   string // DynamoDB table name for on-demand payments
+
+	// DynamoDB table name for on-demand payments, currently used by the controller.
+	OnDemandTableName string
 
 	// Number of relay instances to start, if not specified, no relays will be started.
 	RelayCount int
@@ -56,17 +62,25 @@ type DisperserHarnessConfig struct {
 	OperatorStateSubgraphURL string
 }
 
-// TODO: Add encoder, api server, batcher
+// DisperserHarness is the harness for spinning up the disperser infrastructure as goroutines.
+// It will only support V2 components of the disperser.
+// TODO: Add encoder, api server
 type DisperserHarness struct {
+	// LocalStack infrastructure for blobstore and metadata store
 	LocalStack     *testbed.LocalStackContainer
 	DynamoDBTables struct {
 		BlobMetadataV1 string
-		BlobMetaV2     string
+		BlobMetadataV2 string
 	}
 	S3Buckets struct {
 		BlobStore string
 	}
-	RelayServers     []*relay.Server
+
+	// Relay
+	RelayServers []*relay.Server
+
+	// Controller components
+	// TODO: Refactor into a single struct for controller components
 	EncodingManager  *controller.EncodingManager
 	Dispatcher       *controller.Dispatcher
 	ControllerServer *server.Server
@@ -170,7 +184,7 @@ func SetupDisperserHarness(
 
 	// Populate the harness tables and buckets metadata
 	harness.DynamoDBTables.BlobMetadataV1 = config.MetadataTableName
-	harness.DynamoDBTables.BlobMetaV2 = config.MetadataTableNameV2
+	harness.DynamoDBTables.BlobMetadataV2 = config.MetadataTableNameV2
 	harness.S3Buckets.BlobStore = config.S3BucketName
 
 	localstack, err := setupLocalStackResources(ctx, logger, config)


inabox/tests/setup_disperser_harness.go
```diff
@@ -392,3 +422,145 @@ func stopAllRelays(servers []*relay.Server, logger logging.Logger) {
...
+		operatorStateSubgraphURL,
+		encodingManagerConfig,
+		dispatcherConfig,
+		chainStateConfig,
+		// Optional components (not needed for tests)
+		nil,                                  // metricsServer
+		"",                                   // readinessProbePath
+		healthcheck.HeartbeatMonitorConfig{}, // No heartbeat monitor for tests
+		// Server config
+		// TODO(dmanc): Add server config
```
dmanc: Not sure about the below config. I ran `make run-payments-tests` and it seemed to pass without the below stuff?

api/clients/v2/dispersal_request_signer.go
```diff
@@ -20,6 +20,18 @@ type DispersalRequestSigner interface {
 	SignStoreChunksRequest(ctx context.Context, request *grpc.StoreChunksRequest) ([]byte, error)
 }
 
+type DispersalRequestSignerConfig struct {
+	Region   string
+	Endpoint string
+	KeyID    string
```
samlaf: can you add comments to describe what these are? Are they secrets? If not, should we also have defaults for them in the Default config below?

samlaf: also from reading the code below looks like endpoint is optional..?

api/clients/v2/dispersal_request_signer.go
```diff
@@ -31,36 +43,34 @@ type requestSigner struct {
...
+		awsconfig.WithRegion(config.Region),
 	)
 	if err != nil {
 		return nil, fmt.Errorf("failed to load AWS config: %w", err)
 	}
 
 	var keyManager *kms.Client
-	if endpoint != "" {
+	if config.Endpoint != "" {
 		keyManager = kms.New(kms.Options{
```
samlaf: nit: can we rename `keyManager` to `kmsClient`?

api/clients/v2/dispersal_request_signer.go
```diff
@@ -31,36 +43,34 @@ type requestSigner struct {
...
-			Region:       region,
-			BaseEndpoint: aws.String(endpoint),
+			Region:       config.Region,
+			BaseEndpoint: aws.String(config.Endpoint),
 		})
 	} else {
 		keyManager = kms.NewFromConfig(cfg)
```
samlaf: how does this behave when there's no endpoint passed? how does it identify which AZ the keyID is in?

api/clients/v2/dispersal_request_signer.go
```diff
@@ -20,6 +20,18 @@ type DispersalRequestSigner interface {
...
+	Region   string
+	Endpoint string
+	KeyID    string
+}
+
+func DefaultDispersalRequestSignerConfig() DispersalRequestSignerConfig {
+	return DispersalRequestSignerConfig{
+		Region: "us-east-1",
+	}
+}
```
samlaf, line 42: should start thinking in terms of @cody-littley's config framework:
1. define fields that are required (I'm guessing KeyID is required)
2. if a required field is not part of the DefaultConfig, then we definitely need a validate() function defined that checks that KeyID is non empty
3. constructor should call validate()

Right now afaiu if you pass in an empty keyID the constructor will let you build the DispersalRequestSigner and only crash when you try to use it

common/healthcheck/heartbeat.go
```diff
@@ -9,18 +9,27 @@ import (
 	"github.com/Layr-Labs/eigensdk-go/logging"
 )
 
+type HeartbeatMonitorConfig struct {
+	FilePath         string
+	MaxStallDuration time.Duration
+}
```
samlaf, line 19: just like with dispersal signer, if you're going to refactor into a config you should add comments as to what each field is, whether they are optional or required, and probably define a validate() function

inabox/tests/setup_infra.go
```diff
@@ -24,6 +24,9 @@ type InfrastructureConfig struct {
 	// Number of relay instances to start, if not specified, no relays will be started.
 	RelayCount int
 
+	// DisableController disables the controller deployment when set to true.
+	DisableController bool
```
samlaf: When/why would someone want to disable the controller? Comments should explain intent/usage, not just document the obvious.

core/thegraph/state_integration_test.go
```diff
@@ -43,10 +43,11 @@ func setupTest(t *testing.T) *inaboxtests.InfrastructureHarness {
...
 	config := &inaboxtests.InfrastructureConfig{
-		TemplateName: templateName,
-		TestName:     testName,
-		Logger:       logger,
-		RootPath:     "../../",
+		TemplateName:      templateName,
+		TestName:          testName,
+		Logger:            logger,
+		RootPath:          "../../",
+		DisableController: true,
```
samlaf: does the graph need a disperser deployed at all? Or just validators for registering? If so, think we should just have a boolean DisableDisperser entirely as opposed to having to disable each component. Also unclear to a user the service dependencies, so is it safe to disable the controller if I want to enable the relay for eg? etc

disperser/controller/start_controller.go
```diff
@@ -0,0 +1,374 @@
...
+	EncodingManager *EncodingManager
+	Dispatcher      *Dispatcher
+	LivenessChan    chan healthcheck.HeartbeatMessage
+
+	Server *server.Server
+}
+
+// StartController creates and starts the encoding manager and dispatcher.
+func StartController(
+	ctx context.Context,
```
samlaf: imo better to separate into a New() function which builds the controller, and then another Start(ctx) function which starts the server.

disperser/cmd/controller/main.go
```diff
@@ -142,288 +115,64 @@ func RunController(ctx *cli.Context) error {
...
 		requestSigner, err = clients.NewDispersalRequestSigner(
 			context.Background(),
-			config.AwsClientConfig.Region,
-			config.AwsClientConfig.EndpointURL,
-			config.DisperserKMSKeyID)
+			clients.DispersalRequestSignerConfig{
+				Region:   config.AwsClientConfig.Region,
+				Endpoint: config.AwsClientConfig.EndpointURL,
+				KeyID:    config.DisperserKMSKeyID,
+			})
```
samlaf: how did you decide what to build inside StartController() and whta to build outside and inject? Distinction seems a bit arbitrary right now

inabox/tests/setup_disperser_harness.go
```diff
@@ -119,12 +129,17 @@ func SetupDisperserHarness(
 	ctx context.Context,
 	logger logging.Logger,
 	ethClient common.EthClient,
+	operatorStateSubgraphURL string,
 	config DisperserHarnessConfig,
```
samlaf: why isn't operatorStateSubgraphURL part of the config?

api/clients/v2/dispersal_request_signer.go
dmanc: KMS is region specific so just specifying the region is enough. The endpoint param is only useful if you're testing with localstack or some other implementation.

api/clients/v2/dispersal_request_signer.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2130/commits/ab7b973d5032be02108edfdcf3d7ef72dbb083b1

api/clients/v2/dispersal_request_signer.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2130/commits/ab7b973d5032be02108edfdcf3d7ef72dbb083b1

api/clients/v2/dispersal_request_signer.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2130/commits/ab7b973d5032be02108edfdcf3d7ef72dbb083b1

common/healthcheck/heartbeat.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2130/commits/5887c3803b295ffb15811a1c978a4735b618d309

inabox/tests/setup_disperser_harness.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2130/commits/6c02f04d8b799accde366812cd3e418d3eb204e7

core/thegraph/state_integration_test.go
dmanc: Added disable disperser config instead: https://github.com/Layr-Labs/eigenda/pull/2130/commits/6c02f04d8b799accde366812cd3e418d3eb204e7. We can decide later if we want to support disabling specific components.

inabox/tests/setup_infra.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2130/commits/6c02f04d8b799accde366812cd3e418d3eb204e7, removed in favor of disabling the disperser.

disperser/controller/start_controller.go
dmanc: tried it here https://github.com/Layr-Labs/eigenda/pull/2130/commits/feb79f2dd715ed1c9314b77f20e3c0be40d1d850

disperser/cmd/controller/main.go
dmanc: Okay did a refactor in this commit https://github.com/Layr-Labs/eigenda/pull/2130/commits/feb79f2dd715ed1c9314b77f20e3c0be40d1d850.

The current pattern is we pass in dependencies  + config required to start the encoding manager, dispatcher, and grpc server.

inabox/tests/setup_disperser_harness.go
```diff
@@ -392,3 +419,199 @@ func stopAllRelays(servers []*relay.Server, logger logging.Logger) {
 		}
 	}
 }
+
+// startController starts the controller as a singleton goroutine
+func startController(
```
litt3, line 451: nit: I wonder if this functionality should actually be a method on the `DisperserHarness`? It operates on the `harness`, even going as far as setting member variables. I'd probably argue the same thing for most of the functions within this file.

Up to you, since it's a matter of preference

inabox/tests/setup_disperser_harness.go
```diff
@@ -392,3 +419,199 @@ func stopAllRelays(servers []*relay.Server, logger logging.Logger) {
...
+		encoderClient,
+		ics,
+		batchMetadataManager,
+		sigAgg,
+		nodeClientManager,
+		metricsRegistry,
+		encodingManagerConfig,
+		dispatcherConfig,
+		nil, // No server config for tests
+		nil, // No payment authorization handler for tests
```
litt3: Both the server config and payment authorization handler need to be conditionally built, depending on configuration. Otherwise the new payments system will never be activated

api/clients/v2/dispersal_request_signer.go
```diff
@@ -20,49 +22,79 @@ type DispersalRequestSigner interface {
...
+
+func DefaultDispersalRequestSignerConfig() DispersalRequestSignerConfig {
+	return DispersalRequestSignerConfig{
+		Region:   "us-east-1",
+		Endpoint: "",
+	}
+}
+
+// Verify checks that the configuration is valid, returning an error if it is not.
+func (c *DispersalRequestSignerConfig) Verify() error {
```
litt3, line 45: Should you also assert that region is not empty?

api/clients/v2/dispersal_request_signer_test.go
```diff
@@ -79,6 +79,21 @@ func createTestKMSKey(
...
+
+	// Try to create signer with empty KeyID - validation should catch it immediately
+	_, err := NewDispersalRequestSigner(ctx, DispersalRequestSignerConfig{
+		Region:   region,
+		Endpoint: localstackHost,
+		KeyID:    "", // Empty KeyID
+	})
+
+	require.Error(t, err, "should fail to create signer with empty KeyID")
+	require.Contains(t, err.Error(), "invalid configuration")
```
litt3: IMO we should avoid making assertions about error strings if at all possible. Tests with error string matches are quite fragile

disperser/controller/controller.go
```diff
@@ -0,0 +1,218 @@
...
+	// Create worker pools
+	encodingPool := workerpool.New(encodingManagerConfig.NumConcurrentRequests)
+	dispatcherPool := workerpool.New(dispatcherConfig.NumConcurrentRequests)
+
+	// Create blob sets
+	encodingManagerBlobSet := NewBlobSet()
+	dispatcherBlobSet := NewBlobSet()
+
+	// Create encoding manager
+	encodingManager, err := NewEncodingManager(
```
litt3: IMO it would be cleaner to pass in an `EncodingManager`, `Dispatcher`, and `GRPCServer` already built. That type of dependency injection makes testing a lot easier, and would greatly declutter the constructor signature.

inabox/tests/setup_infra.go
```diff
@@ -105,23 +110,34 @@ func SetupInfrastructure(ctx context.Context, config *InfrastructureConfig) (*In
...
+			infraCtx,
+			logger,
+			infra.ChainHarness.EthClient,
+			*disperserHarnessConfig,
+		)
+		if err != nil {
+			setupErr = fmt.Errorf("failed to setup disperser harness: %w", err)
+			return nil, setupErr
+		}
+		infra.DisperserHarness = *disperserHarness
```
samlaf, line 139: you're doing a lot of pointer dereferencing. should prob stick to either using pointers everywhere, or values everywhere.
My preference is to pass configs by value, but service structs/harnesses by pointer.
But it's still an open debate to reach consensus on which approach we want across the repo.

disperser/controller/controller.go
litt3, line 1: I've reviewed most of this PR, and I'm wondering if it might make sense to split into multiple smaller PRs.

- The changes to controller main and the new `controller` struct are large and important enough that they probably should be in their own PR
- There is some cleanup stuff that you did in various structs (dispersal_request_signer, heartbeat, etc.). IMO this sort of thing is good for a smaller separate PR(s), to avoid cluttering more important changes
- The test changes could then be a relatively small PR, with all prod changes already reviewed and merged

@samlaf Having also reviewed, do you have any thoughts here?

samlaf: 1. agree, I left the exact same comment below
2. agree. it seems innocuous but actually just adds to more stuff to review (had a lot of comments on those small diffs on my first review this morning)
3. prob not that small haha, but agree

dmanc: I'll work on splitting it into smaller PRs but thanks for the initial reviews. Sometimes it's useful for me to see how it all looks combined together but makes sense to split so that the reviewer can have more focus.

litt3: > Sometimes it's useful for me to see how it all looks combined together

💯 That's my pattern too, I like to do a draft of a set of changes and iteratively peel off change subsets into separate PRs

api/clients/v2/dispersal_request_signer.go
dmanc: Resolved in https://github.com/Layr-Labs/eigenda/pull/2138

api/clients/v2/dispersal_request_signer_test.go
dmanc: Resolved in https://github.com/Layr-Labs/eigenda/pull/2138

inabox/tests/setup_disperser_harness.go
```diff
@@ -35,12 +47,16 @@ type DisperserHarnessConfig struct {
 	BucketTableName     string
 	S3BucketName        string // S3 bucket name for blob storage
 	MetadataTableNameV2 string
+	OnDemandTableName   string // DynamoDB table name for on-demand payments
```
samlaf: mention that this is used by controller

inabox/tests/setup_disperser_harness.go
```diff
@@ -50,7 +66,10 @@ type DisperserHarness struct {
 	S3Buckets struct {
 		BlobStore string
 	}
-	RelayServers []*relay.Server
+	RelayServers     []*relay.Server
+	EncodingManager  *controller.EncodingManager
+	Dispatcher       *controller.Dispatcher
+	ControllerServer *server.Server
```
samlaf, line 86: how are these things related? are some of these v1 or v2 only?

samlaf: are these separate processes in prod?

inabox/tests/setup_disperser_harness.go
samlaf: I agree with austin here. Also its not obvious just from the function name that this fct is modifying the harness, so would be good to document.

samlaf: my preference is always to not do things this way but instead have calls that return subsections of the harness and then only at the end build up a full harness from them. But not sure how easy it is to refactor the code to work this way.

samlaf (dismissed):

inabox/tests/setup_disperser_harness.go
dmanc: I'm going to make this disperser harness only support V2 components. For the second point, I will refactor this in a separate PR to create a controller struct.

inabox/tests/setup_disperser_harness.go
dmanc: I'll prefer to do the refactor in a separate PR since startRelays also suffers similar problem

inabox/tests/setup_disperser_harness.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2130/commits/4c74da3be4fb6dd0c465c11fa3cfb6af284a8356

inabox/tests/setup_disperser_harness.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2130/commits/4c74da3be4fb6dd0c465c11fa3cfb6af284a8356

samlaf (approved):

PR #2132, opened 2025-10-07
docs: update readme to remove holesky, and update env verifier address

bxue-l2:

This PR update proxy readme to remove holesky and v1 setup.

Update certVerifier address in .env file, the current [contract](https://sepolia.etherscan.io/address/0x9620dC4B3564198554e4D2b06dEFB7A369D90257#readProxyContract). Note we didn't use router. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/proxy/README.md
+++ b/api/proxy/README.md
@@ -277,7 +277,7 @@ $ PRIVATE_KEY=$(jq -r '.[0].private_key' keypair.json | tail -c +3)
 ## If running with reservation, send us the ETH address for requesting a reservation: https://forms.gle/niMzQqj1JEzqHEny9
 
 ## Start the binary
-$ set -a; source ./.env set +a && ./bin/eigenda-proxy
+$ set -a; source ./.env; set +a; ./bin/eigenda-proxy
 ```
 
 ### Features and Configuration Options (flags/env vars)


api/proxy/README.md
```diff
@@ -273,15 +273,11 @@ $ cast wallet new --json > keypair.json
...
-## If running V1 against testnet or mainnet, register the keypair ETH address and wait for approval: https://forms.gle/niMzQqj1JEzqHEny9
-
-## Run EigenDA Proxy with EigenDA V1 backend
-$ ./bin/eigenda-proxy \
-    --port 3100 \
-    --eigenda.disperser-rpc disperser-holesky.eigenda.xyz:443 \
-    --eigenda.signer-private-key-hex $PRIVATE_KEY \
-    --eigenda.eth-rpc https://ethereum-holesky-rpc.publicnode.com \
-    --eigenda.svc-manager-addr 0xD4A7E1Bd8015057293f0D0A557088c286942e84b
+## If running with on-demand, follow the steps to deposit ETH on L1: https://docs.eigencloud.xyz/products/eigenda/integrations-guides/quick-start/v2/#on-demand-data-dispersal
```
ethenotethan: nit - L1 is a glossary term that we prolly shouldn't be using here: 

```suggestion
## If running with on-demand, follow the steps to deposit ETH: https://docs.eigencloud.xyz/products/eigenda/integrations-guides/quick-start/v2/#on-demand-data-dispersal
```

api/proxy/.env.example
```diff
@@ -52,7 +52,7 @@ EIGENDA_PROXY_EIGENDA_V2_MAX_BLOB_LENGTH=1MiB
 #
 # See https://layr-labs.github.io/eigenda/integration/spec/6-secure-integration.html#upgradable-quorums-and-thresholds-for-optimistic-verification
 # for more details.
-EIGENDA_PROXY_EIGENDA_V2_CERT_VERIFIER_ROUTER_OR_IMMUTABLE_VERIFIER_ADDR=0x58D2B844a894f00b7E6F9F492b9F43aD54Cd4429
+EIGENDA_PROXY_EIGENDA_V2_CERT_VERIFIER_ROUTER_OR_IMMUTABLE_VERIFIER_ADDR=0x19a469Ddb7199c7EB9E40455978b39894BB90974
```
ethenotethan: why not just use the router address instead so we don't have to update this env every-time a new immutable verifier gets deployed? https://sepolia.etherscan.io/address/0x17ec4112c4BbD540E2c1fE0A49D264a280176F0D

api/proxy/README.md
samlaf: hmmm why? EigenDA will always be on Ethereum L1. I think its fine and actually useful to specify it. Otherwise L3s might be confused and think they need to deposit on L2s perhaps?

api/proxy/README.md
```diff
@@ -273,15 +273,11 @@ $ cast wallet new --json > keypair.json
...
+## If running with reservation, register the keypair ETH address and wait for approval: https://forms.gle/niMzQqj1JEzqHEny9
```
samlaf: bit weird wording. filling a form is not "registering".
I would reword as "send us your address, requesting a reservation".

api/proxy/README.md
```diff
@@ -273,15 +273,11 @@ $ cast wallet new --json > keypair.json
...
+
+## Start the binary
+$ set -a; . ./.env && ./bin/eigenda-proxy
```
samlaf: we have a .env.rc file which already loads the env vars. Also .env is not correct since you never mentioned to copy the .env.example file.

api/proxy/.env.example
samlaf: agree with ethen

bxue-l2: can change that

api/proxy/README.md
bxue-l2: Right, I think we only have payment vault on L1

api/proxy/README.md
bxue-l2: It is mentioned immediately above the bash example,

`2. Populate your `.env` file with required values.`

Btw, this is the .envrc

```
# Default example values
dotenv .env.example
# Overrides and secrets (private_key) should go in .env
dotenv_if_exists .env
```

I think it is unnecessarily automated, and requires `dotenv`. `set -a; source ./.env` just works

api/proxy/.env.example
bxue-l2: cbb0942cb1983e2397d573b21f2690281ab72018

api/proxy/README.md
bxue-l2: cbb0942cb1983e2397d573b21f2690281ab72018

api/proxy/README.md
bxue-l2: I agree with Ethen now. User can literally go to the link and figure out where to deposit. Just in case we have option to deposit somewhere else. a1e9ee3716d1d39b61f20165a50508095a678790

api/proxy/README.md
samlaf: you should at least reset `set +a` after sourcing .env then

bxue-l2: e5663fa1da1d09326d923ecbc3397ac85066dc44

samlaf: you still have a typo (missing ; or &&)
be consistent, either:
```
set -a; source ./.env; set +a; ./bin/eigenda-proxy
```
or
```
set -a && source ./.env && set +a && ./bin/eigenda-proxy
```

bxue-l2: it worked

bxue-l2: 7918af858c8d430ca1fd384574281d65cd6a0445

PR #2133, opened 2025-10-07
feat(payments): Add on-demand only inabox tests

litt3:
- Write inabox tests using only on-demand payments

### DIFF
--- a/api/clients/v2/validator/internal/validator_grpc_manager.go
+++ b/api/clients/v2/validator/internal/validator_grpc_manager.go
@@ -75,15 +75,15 @@ func (m *validatorGRPCManager) DownloadChunks(
 		grpc.WithTransportCredentials(insecure.NewCredentials()),
 		grpc.WithDefaultCallOptions(grpc.MaxCallRecvMsgSize(maxMessageSize)),
 	)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create connection to operator %s: %w", operatorID.Hex(), err)
+	}
 	defer func() {
 		err := conn.Close()
 		if err != nil {
 			m.logger.Error("validator retriever failed to close connection", "err", err)
 		}
 	}()
-	if err != nil {
-		return nil, fmt.Errorf("failed to create connection to operator %s: %w", operatorID.Hex(), err)
-	}
 
 	client := grpcnode.NewRetrievalClient(conn)
 	request := &grpcnode.GetChunksRequest{
--- a/api/proxy/servers/rest/handlers_misc.go
+++ b/api/proxy/servers/rest/handlers_misc.go
@@ -114,7 +114,7 @@ func (svr *Server) writeJSON(w http.ResponseWriter, r *http.Request, response in
 		return
 	}
 
-	w.Header().Set(contentTypeJSON, "application/json")
+	w.Header().Set(headerContentType, contentTypeJSON)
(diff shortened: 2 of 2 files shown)


dmanc (dismissed):

inabox/tests/test_harness.go
```diff
@@ -173,6 +173,39 @@ func (tc *TestHarness) UpdateReservationOnChain(
 	return nil
 }
 
+// Makes an on-demand deposit for an account
+func (tc *TestHarness) DepositOnDemandOnChain(
+	t *testing.T,
+	accountID gethcommon.Address,
+	depositAmount *big.Int,
+) error {
```
samlaf, line 181: hmm interesting. I typically don't like having methods defined on god objects like TestHarness. I think (?) I would prefer we keep TestHarness as purely a holder for the test dependencies.

Having a god object with gazillion methods defined is what we have in common and core and its super super messy to understand the chain of dependencies (which of these 30 fields in TestHarness is actually used in here?), which makes understanding the codebase and making refactors harder imo.

But curious for your thoughts, perhaps you think for inabox its fine, and we won't have many such functions?

inabox/tests/payments/payments_test.go
```diff
@@ -232,14 +240,93 @@ func testReservationIncrease(
...
+
+	payloadDisperserConfig := integration.GetDefaultTestPayloadDisperserConfig()
+	payloadDisperserConfig.ClientLedgerMode = clientLedgerMode
+	payloadDisperserConfig.PrivateKey = privateKeyHex
+
+	payloadDisperser, err := testHarness.CreatePayloadDisperser(t.Context(), logger, payloadDisperserConfig)
+	require.NoError(t, err)
+
+	// will be billed as a minimum size blob
+	blobBytes := 1000
```
samlaf:
```suggestion
	payloadBytes := 1000
```

samlaf (approved):
LGTM except for some hygiene nits.

Larger question, as you're structuring these tests to test the upgrade from legacy to your new payment clients, is it going to be able to delete the legacy tests once we've completely migrated? Good to keep this in mind.

inabox/tests/payments/payments_test.go
litt3: b2300af507da8149b1dbfdb79cdba9eb63cadb3c

inabox/tests/test_harness.go
litt3: I definitely see the concern. What are your thoughts on this distinction for what belongs as a method:

1. If you just want to do something _with_ the `TestHarness`, it shouldn't be a method. For example, I _don't_ think it would be good to have a `DispersePayload` method on the `TestHarness` that we disperse through: instead, you should get the object you actually want to work with, and call the correct method on that object.
2. If you want to perform an action which _modifies_ the `TestHarness` at a more fundamental level, _that_ is permitted to be a method. Take this `DepositOnDemandOnChain` method I created: its purpose is to modify the on-chain state, which is a core part of the `TestHarness`. Essentially, you're calling this method to setup the correct conditions for a test. I think that's why it feels natural as a `TestHarness` method to me.

samlaf: Interesting I like this split. Although my brain still has a big preference for immutable testHarnesses. But I can live with it. Sold.

PR #2134, opened 2025-10-07
feat(proxy): add default max request size limit to arb custom da

ethenotethan:
Imposes an upper limit on incoming http requests to arb custom da server. given the existing constant defined for rest server is set proportionally to eigenda max blob size then it makes sense to globally share this across servers. furthermore user configurability for this value likely not matter since we don't expose a nob for this on the rest server. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proxy/common/common.go
```diff
@@ -7,6 +7,11 @@ import (
 	"strings"
 )
 
+const (
+	// limit requests to only 32 MiB to mitigate potential DoS attacks
+	MaxServerPOSTRequestBodySize int64 = 1024 * 1024 * 32
```
litt3: Would this belong in the `ServerConfig`, rather than being a global variable?

samlaf: Couldn't we just use maxBlobSize? Since payload < blob.

ethenotethan: @litt3 the idea is that its shared between the Arb Custom DA and REST ALT DA servers

PR #2136, opened 2025-10-08
feat: bootstrap CLI utility

cody-littley:

This PR introduces a microscopic CLI for bootstrapping proper config. This should generally be reusable for anything that wants to use the new config framwork to configure itself.

### DIFF
--- a/common/config/bootstrap.go
+++ b/common/config/bootstrap.go
@@ -78,11 +78,17 @@ func Bootstrap[T VerifiableConfig](
 		return zero, fmt.Errorf("error parsing command line arguments: %w", err)
 	}
 
-	// If the help flag was set, don't block waiting for cfgChan.
-	// TODO Claude implement this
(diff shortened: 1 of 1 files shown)


common/config/bootstrap.go
```diff
@@ -0,0 +1,149 @@
...
+		Action: action,
+	}
+
+	err = app.Run(os.Args)
+	if err != nil {
+		var zero T
+		return zero, fmt.Errorf("error parsing command line arguments: %w", err)
+	}
+
+	// If the help flag was set, the action never runs and cfgChan is never written to.
```
litt3, line 81: I don't see any actual implementation for `help`. Is that baked into urfave somehow?

litt3 (approved):

common/config/bootstrap.go
cody-littley: Yes, urfave does it automatically if the `--help` or `-h` flag is present. If help triggers then it will not call the action function.

PR #2138, opened 2025-10-08
refactor(controller): Refactor disperser request signer to use new config framework

dmanc:

Adds some extra validation to ensure we don't create a misconfigured request signer (empty keyid, empty region). 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/dispersal_request_signer.go
+++ b/api/clients/v2/dispersal_request_signer.go
@@ -66,6 +66,10 @@ func NewDispersalRequestSigner(
 	ctx context.Context,
 	config DispersalRequestSignerConfig,
 ) (DispersalRequestSigner, error) {
+	if err := config.Verify(); err != nil {
+		return nil, fmt.Errorf("invalid config: %w", err)
+	}
+
 	var kmsClient *kms.Client
 	if config.Endpoint != "" {
 		kmsClient = kms.New(kms.Options{


api/clients/v2/dispersal_request_signer.go
```diff
@@ -8,9 +8,17 @@ import (
...
-	"github.com/aws/aws-sdk-go-v2/config"
+	awsconfig "github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/kms"
+	"github.com/pkg/errors"
+)
+
+// Sentinel errors for configuration validation
+var (
+	ErrKeyIDRequired  = errors.New("KeyID is required")
+	ErrRegionRequired = errors.New("Region is required")
```
samlaf: I personally think sentinel errors should only be defined when we plan on pattern matching on them somewhere. Since config errors are bugs and not expected errors, I don't think we should define sentinel errors (literally no one in their right mind would pattern match these).

api/clients/v2/dispersal_request_signer.go
```diff
@@ -20,49 +28,82 @@ type DispersalRequestSigner interface {
...
+	// Region is the AWS region where the KMS key is located (e.g., "us-east-1"). Required.
+	Region string
+	// Endpoint is an optional custom AWS KMS endpoint URL. If empty, the standard AWS KMS endpoint is used.
+	// This is primarily useful for testing with LocalStack or other custom KMS implementations. Default is empty.
+	Endpoint string
+}
+
+var _ config.VerifiableConfig = &DispersalRequestSignerConfig{}
+
+func DefaultDispersalRequestSignerConfig() DispersalRequestSignerConfig {
```
samlaf, line 37: given that this function builds in invalid Config (no KeyID), I feel like it would be more appropriate for it to take a KeyID as argument? Not sure how we want to deal with these things, but imo a default config should be valid..

If there is no valid default config for KeyID then perhaps it shouldn't be in the config and should be passed separately? Not sure.. worth debating these choices.

api/clients/v2/dispersal_request_signer.go
```diff
@@ -20,49 +28,82 @@ type DispersalRequestSigner interface {
...
+) (DispersalRequestSigner, error) {
+	if err := config.Verify(); err != nil {
+		return nil, fmt.Errorf("invalid configuration: %w", err)
+	}
 
 	// Load the AWS SDK configuration, which will automatically detect credentials
 	// from environment variables, IAM roles, or AWS config files
-	cfg, err := config.LoadDefaultConfig(ctx,
-		config.WithRegion(region),
+	cfg, err := awsconfig.LoadDefaultConfig(ctx,
```
samlaf: looks like this cfg is only used in the else branch below? should prob just instantiate it there then?

api/clients/v2/dispersal_request_signer.go
dmanc: Right now it's used for the test to make sure the right validation occurred:
`require.ErrorIs(t, err, ErrRegionRequired, "error should be ErrRegionRequired")`.

If we removed it we would just assert that some error occured, is that fine?

api/clients/v2/dispersal_request_signer.go
dmanc: cc: @cody-littley 

I think it makes sense that every default config should verify successfully.

api/clients/v2/dispersal_request_signer.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2138/commits/db7df80b25297bbd88aa291f9f65494d2673452a

api/clients/v2/dispersal_request_signer.go
litt3: > I think it makes sense that every default config should verify successfully.

I don't think that's actually possible with what @cody-littley  is working on, since we require a 0 arg "Get default" function, and some fields required for a config to be valid may not have a reasonable default.

IMO it doesn't seem bad to have a "default" config that is invalid. This flow seems reasonable:
1. Get default config
2. Set fields that don't have a reasonable default

api/clients/v2/dispersal_request_signer.go
litt3: > If we removed it we would just assert that some error occurred

This is the pattern I've been following. It's not ideal that we cannot assert specific reasons for a failure... but I personally prefer it to the two alternatives:
1. fragile error string matching
2. sentinel errors created only for testing

api/clients/v2/dispersal_request_signer.go
```diff
@@ -20,49 +28,81 @@ type DispersalRequestSigner interface {
...
+
+func DefaultDispersalRequestSignerConfig() DispersalRequestSignerConfig {
+	return DispersalRequestSignerConfig{
+		Region:   "us-east-1",
+		Endpoint: "",
+	}
+}
+
+// Verify checks that the configuration is valid, returning an error if it is not.
+func (c *DispersalRequestSignerConfig) Verify() error {
```
litt3, line 45: Theoretically we could do some verification of the URL, since I assume any non-empty value must be a valid URL

May not be worth addressing, though, if it's not trivial to do

api/clients/v2/dispersal_request_signer.go
```diff
@@ -20,49 +28,81 @@ type DispersalRequestSigner interface {
...
-	cfg, err := config.LoadDefaultConfig(ctx,
-		config.WithRegion(region),
-	)
-	if err != nil {
-		return nil, fmt.Errorf("failed to load AWS config: %w", err)
+	config DispersalRequestSignerConfig,
+) (DispersalRequestSigner, error) {
+	if err := config.Verify(); err != nil {
+		return nil, fmt.Errorf("invalid configuration: %w", err)
 	}
```
cody-littley: This isn't wrong, per se. But the intention of the `Verify()` method was for it to be called by the configuration framework automatically at config load time.

Is your intention here to do a double check? This isn't necessarily a bad thing, just wanted to ensure it was intentional.

api/clients/v2/dispersal_request_signer.go
cody-littley: My intention for config constructors was for it to produce a struct that had meaningful default values for everything except required arguments. The intended user of these constructors is the config framework, not manual calls (e.g. in unit tests).

IMO, if we are going to be manually calling a constructor, it makes sense to have required arguments be parameters on the constructor. An exception to this rule is when a nested struct's constructor is called in a different struct's constructor.

Due to the nature of the new config framework, it simply isn't possible to use anything other than a zero arg constructor. The config framework sets fields directly, and the logic required for it to introspect and figure out which arguments need to be passed in as which parameters would extraordinarily expensive.

cody-littley: As a middle ground, perhaps we should always add godocs explicitly calling out that there are some required but unset values in constructors like this.

api/clients/v2/dispersal_request_signer.go
cody-littley: IMO, cost benefit ratio is probably not in our favor on this one. Lots of ways that syntactically valid URLs are still not proper URLs.

disperser/cmd/controller/main.go
```diff
@@ -198,9 +198,11 @@ func RunController(ctx *cli.Context) error {
...
 		requestSigner, err = clients.NewDispersalRequestSigner(
 			context.Background(),
-			config.AwsClientConfig.Region,
-			config.AwsClientConfig.EndpointURL,
-			config.DisperserKMSKeyID)
+			clients.DispersalRequestSignerConfig{
+				Region:   config.AwsClientConfig.Region,
+				Endpoint: config.AwsClientConfig.EndpointURL,
+				KeyID:    config.DisperserKMSKeyID,
+			})
```
cody-littley: Is there a reason why we don't just nest this inner config struct inside the parent config struct, as opposed to copying the config out of one struct and into another? Or is this PR just an incremental step towards doing that?

api/clients/v2/dispersal_request_signer.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2138/commits/1d21290208f11c10f07d87170fb15e20c5401658

api/clients/v2/dispersal_request_signer.go
dmanc: Going to skip this one if that's okay

disperser/cmd/controller/main.go
dmanc: No reason, tried doing that here https://github.com/Layr-Labs/eigenda/pull/2138/commits/1d25fd31c47cda31c513bf65e6a7699b21919200

api/clients/v2/dispersal_request_signer.go
dmanc: I moved it inside the existing NewConfig method https://github.com/Layr-Labs/eigenda/pull/2138/commits/1d25fd31c47cda31c513bf65e6a7699b21919200. This will get replaced with the new config framework.

api/clients/v2/dispersal_request_signer.go
litt3, line 37: Is this config struct actually specific to a `DispersalRequestSigner`, or could it function for any KMS signing tool?

dmanc: Seems like we can make it generic but right now it only has a single function (signing StoreChunks requests)

litt3: In general I'd say it makes sense to just call it a `RequestSignerConfig`, and the `DispersalRequestSigner` would use it. Then, any future request signers would be ready to use the config, too

Since we're probably not gonna be adding any more KMS signers, it's probably not too important.

api/clients/v2/dispersal_request_signer.go
samlaf: Yes I agree. The perfect world I would like is that we define some "400 error" type, and all errors that are 400s would implement this interface. Then we could check for that. Any validation error should not be a sentinel, so this is the thing tha tmakes the most sense to me. aws sdk actually does this well, but I never managed to pull teh trigger and achieve this across the repo (very tedious/hard refactor)

api/clients/v2/dispersal_request_signer.go
samlaf: I see.... so... is this small Config ever embedded into a Config that would get fed via cody's framework? If not then prob don't need to have the zero argument constructor. If yes, then I'm fine with Cody's last middle-ground proposal.

api/clients/v2/dispersal_request_signer.go
samlaf: IMO double checking is good and prob what we should be doing. My model for this comes from https://tigerbeetle.com/blog/2023-12-27-it-takes-two-to-contract/

Checking config is only done once at initialization anyways. And also a config could be populated manually for some tests or other reasons which would bypass the verify inside the NewConfig(cliCtx) function.

PR #2139, opened 2025-10-08
refactor(controller): Add configuration validation for heartbeat monitor

dmanc:

Adds configuration + validation to the heartbeat monitor used in the controller.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/common/healthcheck/heartbeat.go
+++ b/common/healthcheck/heartbeat.go
@@ -51,6 +51,10 @@ func NewHeartbeatMonitor(
 	livenessChan <-chan HeartbeatMessage,
 	config HeartbeatMonitorConfig,
 ) error {
+	if err := config.Verify(); err != nil {
+		return fmt.Errorf("invalid config: %w", err)
+	}
+
 	// Create the heartbeat file if it doesn't exist
 	if _, err := os.Create(config.FilePath); err != nil {
 		return fmt.Errorf("failed to create heartbeat file: %w", err)


common/healthcheck/heartbeat.go
```diff
@@ -6,21 +6,51 @@ import (
 	"strings"
 	"time"
 
+	"github.com/Layr-Labs/eigenda/common/config"
 	"github.com/Layr-Labs/eigensdk-go/logging"
 )
 
+// HeartbeatMonitorConfig configures the heartbeat monitoring system that tracks component health.
+type HeartbeatMonitorConfig struct {
```
litt3, line 14: Not required, but consider adding a `GetDefaultHeartbeatMonitorConfig` function. The config framework @cody-littley is working on requires such a function for automatic config doc generation, and otherwise it's generally convenient

I think it would make sense to have a default `MaxStallDuration`, and probably also a `FilePath` (assuming a relative path is supported?)

dmanc: https://github.com/Layr-Labs/eigenda/pull/2139/commits/0ef261adbbd68f9caeb06885d6c6c07bc981d30d

common/healthcheck/heartbeat.go
```diff
@@ -6,21 +6,60 @@ import (
...
+	// FilePath is the path to the file where heartbeat status will be written. Required.
+	FilePath string
```
samlaf, line 16: Whats the point of this actually? Who reads/consumes these files? Is it just for debugging or we use them programmatically somehow?

dmanc: An example would be that a kubernetes probe can check the information in the file (like how stale it is) and it can restart the pod.

PR #2140, opened 2025-10-08
refactor(controller): Add configuration validation for encoding manager and dispatcher

dmanc:

Adds configuration validation for encoding manager and dispatcher components within the controller. Also adds more documentation for the configuration fields.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/controller/dispatcher.go
+++ b/disperser/controller/dispatcher.go
@@ -214,14 +214,14 @@ func NewDispatcher(
 		return nil, errors.New("config is required")
 	}
 
+	if err := config.Verify(); err != nil {
+		return nil, fmt.Errorf("invalid config: %w", err)
+	}
+
 	// CLI library doesn't support float slices at current version, parsing must happen manually.
-	// Verify() has already validated that these parse correctly and are in range.
 	significantThresholds := make([]float64, 0, len(config.SignificantSigningMetricsThresholds))
 	for _, threshold := range config.SignificantSigningMetricsThresholds {
-		significantThreshold, err := strconv.ParseFloat(threshold, 64)
-		if err != nil {
-			return nil, fmt.Errorf("failed to parse significant signing metrics threshold: %w", err)
-		}
+		significantThreshold, _ := strconv.ParseFloat(threshold, 64)
 		significantThresholds = append(significantThresholds, significantThreshold)
 	}
 
--- a/disperser/controller/encoding_manager.go
+++ b/disperser/controller/encoding_manager.go
@@ -154,6 +154,10 @@ func NewEncodingManager(
 	blobSet BlobSet,
 	controllerLivenessChan chan<- healthcheck.HeartbeatMessage,
 ) (*EncodingManager, error) {
+	if err := config.Verify(); err != nil {
+		return nil, fmt.Errorf("invalid config: %w", err)
+	}
+
 	return &EncodingManager{
 		EncodingManagerConfig:  config,
 		blobMetadataStore:      blobMetadataStore,


disperser/controller/dispatcher.go
```diff
@@ -29,42 +30,131 @@ var errNoBlobsToDispatch = errors.New("no blobs to dispatch")
...
 	BatchMetadataUpdatePeriod time.Duration
 
-	// The maximum time permitted to wait for a node to provide a signature for a batch.
+	// AttestationTimeout is the maximum time to wait for a single node to provide a signature.
+	// Must be positive.
 	AttestationTimeout time.Duration
 
-	// The maximum time permitted to wait for all nodes to provide signatures for a batch.
+	// BatchAttestationTimeout is the maximum time to wait for all nodes to provide signatures for a batch.
+	// Must be positive and should be longer than AttestationTimeout.
```
litt3: > should be longer than AttestationTimeout

Recommend changing "should" to "must", and adding validation in `Verify`, since it doesn't seem sensical to ever break this relationship

disperser/controller/encoding_manager.go
```diff
@@ -25,23 +26,95 @@ import (
...
+	// Must be positive.
+	StoreTimeout time.Duration
+	// NumEncodingRetries is the number of times to retry encoding a blob after the initial attempt fails.
+	// A value of 0 means no retries (only the initial attempt).
+	// Must be non-negative.
 	NumEncodingRetries int
-	// NumRelayAssignment defines how many relays will be assigned to a blob
+	// NumRelayAssignment is the number of relays to assign to each blob.
+	// Must be at least 1 and cannot exceed the length of AvailableRelays.
 	NumRelayAssignment uint16
```
litt3, line 49: Would it make any sense to define a default here? The potential options that come to mind:
1. `1`
2. len(AvailableRelays)

disperser/controller/encoding_manager.go
```diff
@@ -25,23 +26,95 @@ import (
...
-	// AvailableRelays is a list of available relays
+	// AvailableRelays is the list of relay keys that can be assigned to blobs.
+	// Must not be empty.
 	AvailableRelays []corev2.RelayKey
-	// EncoderAddress is the address of the encoder
+	// EncoderAddress is the network address of the encoder service (e.g., "localhost:50051").
+	// Must not be empty.
 	EncoderAddress string
```
litt3, line 55: Would it make sense to assert that this is a valid URL if not empty? Not sure if we have a convenient utility available for that check

disperser/controller/encoding_manager.go
```diff
@@ -80,14 +153,10 @@ func NewEncodingManager(
...
-		len(config.AvailableRelays) == 0 ||
-		config.MaxNumBlobsPerIteration < 1 {
-		return nil, fmt.Errorf("invalid encoding manager config")
-	}
-	if int(config.NumRelayAssignment) > len(config.AvailableRelays) {
-		return nil, fmt.Errorf("NumRelayAssignment (%d) cannot be greater than NumRelays (%d)", config.NumRelayAssignment, len(config.AvailableRelays))
+	if err := config.Verify(); err != nil {
+		return nil, fmt.Errorf("invalid encoding manager config: %w", err)
 	}
+
```
cody-littley: Once we use the new tooling to to actually load this config, then we may want to delete verification from this location, and instead let the config framework call the `Verify()` method.

disperser/controller/encoding_manager.go
dmanc: Similar to https://github.com/Layr-Labs/eigenda/pull/2138#discussion_r2417361594, going to skip it.

disperser/controller/encoding_manager.go
dmanc: added a TODO https://github.com/Layr-Labs/eigenda/pull/2140/commits/ca867eae7ab09595513526569e3e63618cd40951

disperser/controller/encoding_manager.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2140/commits/ca867eae7ab09595513526569e3e63618cd40951, used 1

disperser/controller/dispatcher.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2140/commits/ca867eae7ab09595513526569e3e63618cd40951

disperser/controller/encoding_manager.go
dmanc: moved inside NewConfig https://github.com/Layr-Labs/eigenda/pull/2140/commits/90b245420b9fa884cfa45e77186e4ac64d0e2838, which will get replaced by config framework.

disperser/controller/dispatcher.go
```diff
@@ -106,6 +106,10 @@ func (c *DispatcherConfig) Verify() error {
 	if c.BatchAttestationTimeout <= 0 {
 		return fmt.Errorf("BatchAttestationTimeout must be positive, got %v", c.BatchAttestationTimeout)
 	}
+	if c.BatchAttestationTimeout <= c.AttestationTimeout {
```
litt3: Should the condition actually be `<`, instead of `<=`?

With single blob batches, it would make sense to have these be the same value, so it seems like we should at least permit that configuration

dmanc: makes sense https://github.com/Layr-Labs/eigenda/pull/2140/commits/2ec9811042139196fff722e462db696274b489f1

disperser/controller/encoding_manager.go
samlaf: I would favor having it in both places personally: https://github.com/Layr-Labs/eigenda/pull/2138#discussion_r2420566810
Think it makes sense for constructor to do its own validation, just in case. It shouldn't assume that validation has been done by some other component. Becuse refactors do happen and that could lead to bugs.

disperser/controller/dispatcher.go
```diff
@@ -119,20 +213,14 @@ func NewDispatcher(
...
-		config.AttestationTimeout == 0 ||
-		config.BatchAttestationTimeout == 0 ||
-		config.SignatureTickInterval == 0 ||
-		config.MaxBatchSize == 0 {
-		return nil, errors.New("invalid config")
-	}
 
-	// CLI library doesn't support float slices at current version, parsing must happen manually
+	// CLI library doesn't support float slices at current version, parsing must happen manually.
+	// Verify() has already validated that these parse correctly and are in range.
```
samlaf: I personally think we should just call Verify() again. If we have to write a docs to document it anyways, just verify it again.

samlaf: Think constructors should be self-contained personally, without external "global" assumptions on what was called before calling them

dmanc: https://github.com/Layr-Labs/eigenda/pull/2140/commits/45cfbc28069ffd3a791daa828b68d9f94643f5e6

PR #2142, opened 2025-10-09
feat: SCREAMING_SNAKE_CASE env vars

cody-littley:

The config framework now formats env vars using SCREAMING_SNAKE_CASE.

### DIFF
--- a/common/config/config_document_generator.go
+++ b/common/config/config_document_generator.go
@@ -475,6 +475,7 @@ func toScreamingSnakeCase(s string) string {
 			if isPrevPrevUpper {
 				// Multiple uppercase letters followed by lowercase
 				// Check if this is a single lowercase letter or if multiple lowercase letters follow
+				// nolint:staticcheck
 				isSingleLowercase := i == len(runes)-1 || !(runes[i+1] >= 'a' && runes[i+1] <= 'z')
 
 				if !isSingleLowercase {


common/config/config_document_generator.go
```diff
@@ -415,6 +415,83 @@ func gatherConfigFieldData(
...
+//
+//  4. Strings that are already all uppercase remain unchanged
+//     Examples: "FIELD" -> "FIELD", "HTTP" -> "HTTP"
+func toScreamingSnakeCase(s string) string {
+	if s == "" {
+		return ""
+	}
+
+	var result strings.Builder
+	result.Grow(len(s) + len(s)/2) // Pre-allocate with estimate for underscores
```
litt3: seems probably overcomplicated, but not wrong

common/config/config_document_generator.go
```diff
@@ -415,6 +415,83 @@ func gatherConfigFieldData(
...
+
+	runes := []rune(s)
+	for i := 0; i < len(runes); i++ {
+		r := runes[i]
+
+		// Check if we should insert an underscore before this character
+		if i > 0 {
```
litt3:
```
		if i == 0 {
			continue
		}
```

Consider reversing the condition here, to decrease the deep nesting by 1 level. Alternatively, you could specifically handle `r := runes[0]` outside the loop, and start iteration at `i := 1`

common/config/config_document_generator.go
cody-littley: removed

common/config/config_document_generator.go
cody-littley: good suggestion, done

PR #2143, opened 2025-10-09
feat: ejection manager

cody-littley:
Introduces the EjectionManager, a utility responsible for managing the ejection workflow lifecycle.

### DIFF
--- a/ejector/ejection_manager.go
+++ b/ejector/ejection_manager.go
@@ -34,8 +34,8 @@ var _ EjectionManager = (*ejectionManager)(nil)
 type inProgressEjection struct {
 	// The time when the ejection can be finalized.
 	ejectionFinalizationTime time.Time
-	// For each quorum the validator is a member of, the validator's stake in that quorum as a fraction of 1.0.
-	stake map[core.QuorumID]float64
+	// For each quorum the validator is a member of, the validator's stakeFraction in that quorum as a fraction of 1.0.
+	stakeFraction map[core.QuorumID]float64
 }
 
 // A utility that manages ejections and the ejection lifecycle. An ejection manager is responsible for executing
@@ -142,7 +142,7 @@ func (em *ejectionManager) BeginEjection(
 	}
 
 	// Check to see if we are already in the process of ejecting this validator.
-	if _, inProgress := em.ejectionsInProgress[validatorAddress]; inProgress {
+	if _, ejectionAlreadyBeingTracked := em.ejectionsInProgress[validatorAddress]; ejectionAlreadyBeingTracked {
 		em.logger.Infof("ejection already in progress for validator %s, will not begin ejection",
 			validatorAddress.Hex())
 		return
@@ -156,14 +156,14 @@ func (em *ejectionManager) BeginEjection(
 	}
 
 	// Check to see if there is already an ejection in progress on-chain for this validator.
-	inProgress, err := em.transactor.IsEjectionInProgress(em.ctx, validatorAddress)
+	ejectionStartedOnchain, err := em.transactor.IsEjectionInProgress(em.ctx, validatorAddress)
 	if err != nil {
 		em.logger.Errorf("failed to check ejection status for validator %s, will not begin ejection: %v",
 			validatorAddress.Hex(), err)
 		return
 	}
-	if inProgress {
-		// An ejection is already in progress. Record it, and we can try to finalize it later.
+	if ejectionStartedOnchain {
+		// An ejection is already in progress onchain. Record it, and we can try to finalize it later.
 		em.logger.Infof("ejection already in progress on-chain for validator %s, "+
 			"will not begin ejection but will attempt to finalize",
 			validatorAddress.Hex())
@@ -200,7 +200,7 @@ func (em *ejectionManager) scheduleFutureEjectionFinalization(
 	em.recentEjectionTimes[validatorAddress] = em.timeSource()
 	em.ejectionsInProgress[validatorAddress] = &inProgressEjection{
 		ejectionFinalizationTime: em.timeSource().Add(em.config.EjectionFinalizationDelay),
-		stake:                    stakeFractions,
+		stakeFraction:            stakeFractions,
 	}
 }
 
@@ -240,7 +240,7 @@ func (em *ejectionManager) FinalizeEjections() {
 			ejected := em.finalizeEjection(address)
 
 			if !ejected {
-				em.cleanUpFailedEjection(address, ejection.stake)
+				em.cleanUpFailedEjection(address, ejection.stakeFraction)
 			}
 		}
 	}
--- a/ejector/ejector_config.go
+++ b/ejector/ejector_config.go
@@ -104,7 +104,6 @@ func (e *RootEjectorConfig) GetPackagePaths() []string {
 	}
 }
 
-// Verify implements config.DocumentedConfig.
 func (e *RootEjectorConfig) Verify() error {
 	err := e.Config.Verify()
 	if err != nil {


ejector/mock_ejection_transactor.go
```diff
@@ -0,0 +1,115 @@
...
+
+	if _, ok := m.completedEjections[addressToEject]; ok {
+		return fmt.Errorf("ejection already completed for address %s", addressToEject.Hex())
+	}
+
+	delete(m.inProgressEjections, addressToEject)
+	m.completedEjections[addressToEject] = struct{}{}
+
+	// Once ejected, the validator should no longer be present in any quorum.
+	m.isValidatorPresentInAnyQuorumResponses[addressToEject] = true
```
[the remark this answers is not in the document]
litt3: I had the same comment, seems correct

ejector/ejector_config.go
```diff
@@ -0,0 +1,150 @@
...
+
+import (
+	"fmt"
+	"time"
+
+	"github.com/Layr-Labs/eigenda/common/config"
+)
+
+// The environment variable prefix to use for the ejector configuration.
+const EjectorConfigEnvPrefix = "EJECTOR"
```
litt3: unused

ejector/ejector_config.go
```diff
@@ -0,0 +1,150 @@
...
+type EjectorConfig struct {
+
+	// The address of the contract directory contract.
+	ContractDirectoryAddress string `docs:"required"`
+
+	// The Ethereum RPC URL(s) to use for connecting to the blockchain.
+	EthRpcUrls []string `docs:"required"`
+
+	// The private key to use for signing ejection transactions.
+	PrivateKey string `docs:"required"`
```
litt3: I thought we didn't want to put secrets in the normal config structs?

ejector/ejector_config.go
```diff
@@ -0,0 +1,150 @@
...
+// The environment variable prefix to use for the ejector configuration.
+const EjectorConfigEnvPrefix = "EJECTOR"
+
+var _ config.DocumentedConfig = (*EjectorConfig)(nil)
+
+// Configuration for the ejector.
+type EjectorConfig struct {
+
+	// The address of the contract directory contract.
+	ContractDirectoryAddress string `docs:"required"`
```
litt3, line 26: Do we actually need this here?

Hard to evaluate since the code that uses this isn't in this PR. But my prior is that it makes more sense to get addresses for specific contracts from the contract registry, and then pass them in when building.

ejector/ejection_manager.go
```diff
@@ -0,0 +1,408 @@
...
+	"time"
+
+	"github.com/Layr-Labs/eigenda/common/enforce"
+	"github.com/Layr-Labs/eigenda/common/ratelimit"
+	"github.com/Layr-Labs/eigenda/core"
+	"github.com/Layr-Labs/eigensdk-go/logging"
+	geth "github.com/ethereum/go-ethereum/common"
+)
+
+// TODO add metrics
```
litt3: TODO

ejector/ejector_config.go
```diff
@@ -0,0 +1,150 @@
...
+	EthRpcUrls []string `docs:"required"`
+
+	// The private key to use for signing ejection transactions.
+	PrivateKey string `docs:"required"`
+
+	// The URL of the Eigenda Data API to use for looking up signing rates.
+	DataApiUrl string `docs:"required"`
+
+	// The number of times to retry a failed Ethereum RPC call.
+	EthRpcRetryCount int
```
litt3, line 32: You have a lot of eth specific stuff baked into this config.

Hard to evaluate without being able to actually see how it's being used. But I would generally expect an already constructed eth client to be passed into the main component, which would mean none of this eth stuff would need to be here.

ejector/ejection_manager.go
```diff
@@ -0,0 +1,408 @@
...
+	// There are two ways a validator can end up in this blacklist:
+	// 1. specified in configuration
+	// 2. we've made many attempts to eject the validator, and each attempt has failed (i.e. the validator is
+	//    cancelling the ejection on-chain).
+	ejectionBlacklist map[geth.Address]struct{}
+
+	// The timestamps of recent ejection attempts, keyed by validator address.
+	recentEjectionTimes map[geth.Address]time.Time
+
+	// Ejections that have been started but not completed, keyed by validator address. The value is the
```
litt3: Stale doc. The value is `*inProgressEjection`

ejector/ejection_manager.go
```diff
@@ -0,0 +1,408 @@
...
+	em := &ejectionManager{
+		ctx:                                  ctx,
+		logger:                               logger,
+		timeSource:                           timeSource,
+		ejectionBlacklist:                    make(map[geth.Address]struct{}),
+		recentEjectionTimes:                  make(map[geth.Address]time.Time),
+		ejectionsInProgress:                  make(map[geth.Address]*inProgressEjection),
+		failedEjectionAttempts:               make(map[geth.Address]uint32),
+		transactor:                           transactor,
+		ejectionFinalizationDelay:            config.EjectionFinalizationDelay,
```
litt3: Any reason that you unpack all these config fields, instead of just saving the config struct itself? Is it because there's all the eth client values that seem weird to keep around?

ejector/ejection_manager.go
```diff
@@ -0,0 +1,408 @@
...
+
+	// Check if we are prevented from starting an ejection by rate limiting.
+	allowedByRateLimits := em.checkRateLimits(validatorAddress, stakes)
+	if !allowedByRateLimits {
+		// Rate limiting prevents us from starting an ejection at this time.
+		// checkRateLimits() will have logged the reason, since it has more context.
+		return
+	}
+
+	if inProgress {
```
litt3: Ordering is odd here. Why do you compute the value of `inProgress`, then `checkRateLimits`, _then_ use `inProgress`?

ejector/ejection_manager.go
```diff
@@ -0,0 +1,408 @@
...
+		// An ejection is already in progress. Record it, and we can try to finalize it later.
+		em.logger.Infof("ejection already in progress on-chain for validator %s, "+
+			"will not begin ejection but will attempt to finalize",
+			validatorAddress.Hex())
```
litt3: You don't return if already `inProgress`? Is that intentional?

ejector/ejection_manager.go
```diff
@@ -0,0 +1,408 @@
...
+// TODO add metrics
+
+// EjectionManager manages and executes validator ejections.
+type EjectionManager interface {
+
+	// Begin ejection proceedings against a validator. May not take action if it is not appropriate to do so.
+	BeginEjection(
+		validatorAddress geth.Address,
+		// For each quorum the validator is a member of, the validator's stake in that quorum as a fraction of 1.0.
+		stakes map[core.QuorumID]float64,
```
litt3: You call this concept `stakes` throughout, but I think `stakeFractions` might be a better name. Otherwise, my default interpretation is to consider this a literal `stake` value

ejector/ejection_manager.go
```diff
@@ -0,0 +1,408 @@
...
+// in all quorums, we return true and debit the leaky buckets for each quorum.
+func (em *ejectionManager) checkRateLimits(
+	validatorAddress geth.Address,
+	stakes map[core.QuorumID]float64,
+) bool {
+
+	now := em.timeSource()
+	permittedQuorums := make([]core.QuorumID, 0, len(stakes))
+	for qid, stake := range stakes {
+		if stake <= 0.0 {
```
litt3: If you're checking the lower bound, any reason not to check the upper bound too?

ejector/ejection_manager.go
```diff
@@ -0,0 +1,408 @@
...
+			em.logger.Errorf(
+				"validator %s has non-positive stake %.4f in quorum %d, skipping rate limit check",
+				validatorAddress.Hex(), stake, qid)
+			continue
```
litt3: What would it even mean if a `stakeFraction` is <= 0? Is `continue` the correct response?

ejector/ejection_manager.go
```diff
@@ -0,0 +1,408 @@
...
+		}
+
+		leakyBucket := em.getLeakyBucketForQuorum(qid)
+
+		allowed, err := leakyBucket.Fill(now, stake)
+
+		// The only way we can get an error here is if time moves backwards, or if stake <= 0
```
litt3: > or if stake <= 0

Stale, this is explicitly checked above

ejector/ejection_manager.go
```diff
@@ -0,0 +1,408 @@
...
+		enforce.NilError(err, "should be impossible")
```
litt3, line 265: You `continue` above on an error, but here you panic? Why the discrepancy?

ejector/ejection_manager.go
```diff
@@ -0,0 +1,408 @@
...
+
+// Refund the rate limit fills for each quorum. This should be called if we fail to finalize an ejection.
+// Also removes the ejection from ejectionsInProgress.
+func (em *ejectionManager) cleanUpFailedEjection(
+	validatorAddress geth.Address,
+	stakes map[core.QuorumID]float64,
+) {
+	now := em.timeSource()
+	for qid, stake := range stakes {
+		if stake <= 0.0 {
```
litt3: However the question above is resolved should be mirrored here

ejector/ejection_manager.go
```diff
@@ -0,0 +1,408 @@
...
+		// An ejection is already in progress. Record it, and we can try to finalize it later.
+		em.logger.Infof("ejection already in progress on-chain for validator %s, "+
+			"will not begin ejection but will attempt to finalize",
+			validatorAddress.Hex())
+	} else {
+		// Start a new ejection.
+		err = em.transactor.StartEjection(em.ctx, validatorAddress)
+		if err != nil {
+			em.logger.Errorf("failed to start ejection for validator %s: %v", validatorAddress.Hex(), err)
+			em.cleanUpFailedEjection(validatorAddress, stakes)
```
litt3: What if an error is thrown in `StartEjection` before the debit actually completes? `cleanUpFailedEjection` would then refund a debit that never happened

ejector/ejection_manager.go
```diff
@@ -0,0 +1,408 @@
...
+			em.ejectionBlacklist[address] = struct{}{}
+			delete(em.failedEjectionAttempts, address)
+		} else {
+			em.logger.Infof("validator %s has %d consecutive failed ejection attempts",
+				address.Hex(), em.failedEjectionAttempts[address])
+		}
+	} else {
+		// A different ejector finalized the ejection for us, or the validator was removed from all quorums by
+		// some other mechanism. Either way, we are done here.
+		em.logger.Infof("validator %s no longer present in any quorum, ejection complete", address.Hex())
```
litt3, line 405: So, following the logic through: if a different ejector finalizes the ejection, we nevertheless refund the ejection leaky bucket debit? Seems like the wrong behavior

ejector/ejection_manager.go
```diff
@@ -0,0 +1,408 @@
...
+	if !inProgress {
+		// Either the validator cancelled the ejection or another ejector finalized it for us.
+		em.handleAbortedEjection(address)
+		return false
+	}
+
+	// Complete the ejection.
+	err = em.transactor.CompleteEjection(em.ctx, address)
+	if err != nil {
+		// We failed to eject. Leave the ejection in progress so we can try again later.
```
litt3: This doesn't seem to be the case. If this method returns false, then `cleanUpFailedEjection` is called, and it won't ever be reattempted

ejector/mock_ejection_transactor.go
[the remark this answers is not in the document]
cody-littley: fixed

ejector/ejection_manager_test.go
```diff
@@ -0,0 +1,1932 @@
...
+			require.Equal(t, 0.01, fill)
+			fill, err = em.getLeakyBucketForQuorum(2).GetFillLevel(currentTime)
+			require.NoError(t, err)
+			require.Equal(t, 0.33, fill)
+		}
+		if isTriggerTime(currentTime, previousTime, ejectionTimeC) {
+			// This should be allowed, since overfill is allowed and the bucket should not be completely full.
+			_, started := ejectionTransactor.inProgressEjections[validatorC]
+			require.False(t, started)
+			manager.BeginEjection(validatorC, stakes[validatorB])
```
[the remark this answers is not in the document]
cody-littley: fixed

ejector/ejection_manager_test.go
```diff
@@ -0,0 +1,1932 @@
...
```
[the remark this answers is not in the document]
cody-littley: fixed

ejector/ejection_manager_test.go
```diff
@@ -0,0 +1,1932 @@
...
```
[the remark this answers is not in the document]
cody-littley: fixed

ejector/ejector_config.go
cody-littley: removed

ejector/ejector_config.go
cody-littley: Discussed this offline, working towards augmenting the framework to more elegantly handle multiple structs, in order to support a secrets-only struct

ejector/ejector_config.go
cody-littley: This is needed. The user needs to provide the contract directory address before we can look up things using the contract directory. ;) Passing it in through configuration is the logical strategy, IMO.

ejector/ejection_manager.go
cody-littley: Converted to `// TODO(cody.littley) add metrics`. I will consider this a prereq to calling the ejector feature done, but let's do it as a follow up PR. This PR is already massive.

ejector/ejector_config.go
cody-littley: Agree this is worth considering, but it might be premature to start making those kinds of changes without doing some deeper research on how other parts of the system need similar information.

ejector/ejection_manager.go
cody-littley: Fixed.

ejector/ejection_manager.go
cody-littley: No particular reason. I guess it's less error prone to do less copying, since it means you can't forget to copy over a value. I've converted these to use the config struct.

ejector/ejection_manager.go
cody-littley: Unintentional and a bug. Fixed.

ejector/ejection_manager.go
cody-littley: renamed

ejector/ejection_manager.go
cody-littley: Added an upper bound check for sanity, and moved it upwards to the place where we ingest the data.

```
// Check that the stake fractions are all valid (i.e. in the range (0.0, 1.0]), returning true if they are valid,
// and false otherwise.
func (em *ejectionManager) areStakeFractionsValid(
	validatorAddress geth.Address,
	stakeFractions map[core.QuorumID]float64,
) bool {
	for qid, stake := range stakeFractions {
		if stake < 0.0 {
			em.logger.Errorf(
				"validator %s has non-positive stake %.4f in quorum %d, will not begin ejection",
				validatorAddress.Hex(), stake, qid)
			return false
		}
		if stake > 1.0 {
			em.logger.Errorf(
				"validator %s has stake %.4f > 1.0 in quorum %d, will not begin ejection",
				validatorAddress.Hex(), stake, qid)
			return false
		}
	}
	return true
}
```

ejector/ejection_manager.go
cody-littley: It can only be the result of invalid data coming from the eth RPCs. The reason why I just log it is that I don't want to `panic()`. With the way the code has been refactored, I now do this check as the first thing when receiving an ejection request, meaning that we never bail out of ejection half way through.

ejector/ejection_manager.go
cody-littley: `enforce.NilError` signals that I don't think it should ever be possible for the code to return a non-nil error here, but if my assumptions are wrong then we catch and panic. In the check that continues above, I was not sure it was impossible (i.e. if we were given bad input), so I put thought into the handling mechanism.

I've moved the stake fraction checking to the intake point, so we should now be able to assume that the stake factions are valid elsewhere in this code.

ejector/ejection_manager.go
cody-littley: Removed this check, we now check stake fractions at ingestion time

ejector/ejection_manager.go
cody-littley: Reordered.

ejector/ejection_manager.go
cody-littley: This class has no threads, so it's impossible to get to `StartEjection()` without the debit completing. The debit always happens first, and if successful then moves to the next phase in the process.

ejector/ejection_manager.go
cody-littley: The intention of this throttle is only to throttle ejections carried out by this ejector.

Even of we pessimistically debit the throttle in a situation like this, we have no guarantee that we will observe the ejections that other ejections are enacting, and so this mechanism can't be used to enforce a global throttle on ejection rates, even if we assume that every ejector will respect the limit (which we can't assume).

Perhaps worth a synchronous discussion.

ejector/ejection_manager.go
cody-littley: Updated comment. 

```
// We failed to eject, give up for now.
```

ejector/ejection_manager.go
cody-littley: updated comment

ejector/ejector_config.go
```diff
@@ -70,6 +71,52 @@ type EjectorConfig struct {
...
+	return "Ejector"
+}
+
+func (e *RootEjectorConfig) GetPackagePaths() []string {
+	return []string{
+		"github.com/Layr-Labs/eigenda/ejector",
+	}
+}
+
+// Verify implements config.DocumentedConfig.
```
litt3: This doc is a bit odd. It's technically correct, since `VerifiableConfig` is nested inside `DocumentedConfig`, but seems a bit misleading. Also, is there a reason to explicitly say this for `Verify`, but not for the other interface methods?

ejector/ejection_manager.go
```diff
@@ -0,0 +1,409 @@
...
+}
+
+var _ EjectionManager = (*ejectionManager)(nil)
+
+// Information tracked for each in-progress ejection.
+type inProgressEjection struct {
+	// The time when the ejection can be finalized.
+	ejectionFinalizationTime time.Time
+	// For each quorum the validator is a member of, the validator's stake in that quorum as a fraction of 1.0.
+	stake map[core.QuorumID]float64
```
litt3:
```suggestion
	stakeFraction map[core.QuorumID]float64
```

ejector/ejection_manager.go
litt3: It's confusing to reuse `inProgress` for a different manner of an ejection being "in progress"

The usage above means "*we* are currently ejecting this validator". And the usage here means "there is an on-chain transaction showing ejection of this validator is currently in progress."

ejector/ejector_config.go
cody-littley: This is the comment that gets auto-generated when you ask vs code to add the missing interface methods. I normally delete them, but missed this one.

ejector/ejection_manager.go
cody-littley: renamed

ejector/ejection_manager.go
cody-littley: Unfortunate name collision. There are two ways an ejection can be in progress, but I agree the variable reuse makes it very confusing. I've renamed the first variable to `ejectionAlreadyBeingTracked` and the second one to `ejectionStartedOnchain`.

PR #2146, opened 2025-10-09
feat(proxy): incorporate custom da interface changes

ethenotethan:
Updates JSON RPC methods and arguments to reflect latest changes made by Offchain Labs to the Custom DA interface. This required updating tests as well. 

Big functionality changes are:
1. changing `isValidHeaderByte` to `fetchValidHeaderBytes` which now takes 0 args and returns an activation slice of bytes
2. the addition of a `CollectPreimages` function that's used as part of the validator derivation pipeline (either defense or offensive). before preimage collection happened under the GET function but has now been segmented into a separate method. 

 In true OCL fashion - they've given us no spec and just code references. I've updated the spec [here](https://hackmd.io/@epociask/SkxP2Pa8eg) to showcase high level API. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proxy/servers/arbitrum_altda/handlers.go
```diff
@@ -27,13 +32,52 @@ const (
 			2. introduce missing key security checks that could impact the integration's L2 Beat assessment
 
 	TODO: Method implementations:
-		[X] IsValidHeaderByte // trusted integration
+		[X] GetSupportedHeaderBytes // trusted integration
 		[-] Store // trusted integration
 		[-] RecoverPayloadFromBatch // trusted integration
-		[ ] GenerateProof // secure integration
-		[ ] GenerateCertificateValidityProof // secure integration
+		[ ] GenerateProof // trustless AND secure integration
```
samlaf, line 38: trustless and secure? isn't trustless = secure?

api/proxy/servers/arbitrum_altda/handlers.go
```diff
@@ -13,9 +13,14 @@ import (
...
-	MethodIsValidHeaderByte       = "daprovider_isValidHeaderByte"
+	// trusted integration
+	MethodGetSupportedHeaderBytes = "daprovider_getSupportedHeaderBytes"
 	MethodStore                   = "daprovider_store"
-	MethodRecoverBatchFromPayload = "daprovider_recoverPayloadFromBatch"
+	MethodRecoverPayload          = "daprovider_recoverPayload"
+	MethodCollectPreimages        = "daprovider_collectPreimages"
+	// trustless integration
+	MethodGenerateReadPreimageProof = "daprovider_generateReadPreimageProof"
+	MethodGenerateCertValidityProof = "daprovider_generateCertificateValidityProof"
```
samlaf, line 23: took me a while to understand how these were exposed on proxy.
Find it very confusing that we are defining the methods here but then only ever using these in tests.
Since afaiu the methods are actually registered implicitly by the rpc framework, with the daprovider prefix injected somewhere else.

Would be good to document how these methods are ultimately exposed by our rpc framework. Also if these are only used in tests shouldnt they be defined there? Or again are you using this as documentation the same way you did for the interface?

api/proxy/servers/arbitrum_altda/handlers.go
```diff
@@ -27,13 +32,52 @@ const (
...
+		ctx context.Context,
+		batchNum hexutil.Uint64,
+		batchBlockHash common.Hash,
+		sequencerMsg hexutil.Bytes,
+	) (*PreimagesResult, error)
+
+	Store(
+		ctx context.Context,
+		message hexutil.Bytes,
+		timeout hexutil.Uint64,
```
samlaf, line 64: I hate interfaces, I'd delete this. If there is a single implementation imo don't use an interface. No need for the indirection.

Or is the idea that you're only exposing these functions to make sure no other public methods are exposed by mistake to the rpc server framework? If this is the case, this might actually be a good reason for having this... but would need to be documented.

I really don't like geth's rpc framework now that I look at it. Would be nice if we could just provide a mapping of handlers that we want to expose ourselves, instead of having it implicitly defined by public functions on the handler object we pass...

api/proxy/servers/arbitrum_altda/handlers.go
```diff
@@ -184,7 +214,26 @@ func (h *Handlers) Store(
 	return result, nil
 }
 
-// Generate proof is used to prove a 32 byte CustomDA preimage type for READPREIMAGE
+// CollectPreimages fetches the "polynomial evaluation form" of the dispersed rollup payload
```
samlaf, line 217: "polynomial evaluation form"
should we just call this encoded_payload here? Think that's what @bxue-l2 has been using in hokulea

api/proxy/servers/arbitrum_altda/handlers.go
samlaf: Thinking more I think I actually like having the interface now for the purpose listed above. Wondering if we shouldn't move all method documentation which are currently on the struct to be here instead. If this is meant as documentation (+ make sure we don't expose other methods by mistake), then shouldnt method docs also be here?

api/proxy/servers/arbitrum_altda/handlers.go
```diff
@@ -224,13 +273,12 @@ current encoding proposal:
...
-func (h *Handlers) GenerateProof(
+func (h *Handlers) GenerateReadPreimageProof(
 	ctx context.Context,
-	preimageType hexutil.Uint,
 	certHash common.Hash,
 	offset hexutil.Uint64,
 	certificate hexutil.Bytes,
-) (*GenerateProofResult, error) {
+) (*GenerateReadPreimageProofResult, error) {
 	panic("GenerateProof method is unimplemented")
```
shyam-patel-kira, line 282: ~Is this supposed to be unimplemented?~

api/proxy/servers/arbitrum_altda/handlers.go
bxue-l2: yes

PR #2147, opened 2025-10-10
refactor: decouple verifier-v2 and encoder

samlaf:
V2 verifier struct held an encoder instance and exposed a Decode() function.
Thought there is no reason for this, and that verifier and decoder should remain decoupled.
For example validators afaiu only need to validate the chunks they receive, but not decode them.
Decoding is really only done by retrieval clients to reconstruct blobs from sufficient number of chunks.

In this process, had to do some yak-shaving of converting encoding.ChunkNumber to be a uint64 instead of uint (many functions were using uint, others uint64). Also there is another definition of core.ChunkNumber which I also changed to uint64. All the types in core and encoding merit a larger cleanup (we should never use uint or int, and instead favor fixed size types like uint64 or uint32), but that should be its own separate PR.

### DIFF
--- a/disperser/batcher/batcher.go
+++ b/disperser/batcher/batcher.go
@@ -65,7 +65,7 @@ type Config struct {
 
 	FinalizationBlockDelay uint
 
-	TargetNumChunks          core.ChunkNumber
+	TargetNumChunks          uint64
 	MaxBlobsToFetchFromStore int
 }
 


api/clients/v2/validator/internal/blob_decoder.go
```diff
@@ -13,44 +15,54 @@ type BlobDecoder interface {
...
+		indices []encoding.ChunkNumber,
 		encodingParams *encoding.EncodingParams,
 		blobCommitments *encoding.BlobCommitments,
 	) ([]byte, error)
 }
 
 // BlobDecoderFactory is a function that creates a new BlobDecoder instance.
 type BlobDecoderFactory func(
-	verifier *verifier.Verifier,
+	verifier *rs.Encoder,
```
dmanc: nit: no longer a verifier right?

encoding/rs/encoder.go
```diff
@@ -103,14 +103,19 @@ func (g *Encoder) Encode(inputFr []fr.Element, params encoding.EncodingParams) (
...
 // the Frames and indices don't encode the length of the original data. If maxInputSize
 // is smaller than the original input size, decoded data will be trimmed to fit the maxInputSize.
+//
+// TODO(samlaf): Many call sites have frames and need to convert to FrameCoeffs.
+// Would be nice to figure out a Decode interface that doesn't require creating allocations.
+// Perhaps Decode could take an iterator that produces one FrameCoeffs at a time?
+// That way we could pass either chunks (frameCoeffs) or frames.
 func (e *Encoder) Decode(
-	frames []FrameCoeffs, indices []uint64, maxInputSize uint64, params encoding.EncodingParams,
+	frames []FrameCoeffs, indices []encoding.ChunkNumber, maxInputSize uint64, params encoding.EncodingParams,
```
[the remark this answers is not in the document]
samlaf: that's what we want.. both are using the same type

api/clients/v2/validator/internal/blob_decoder.go
samlaf: [44c729b](https://github.com/Layr-Labs/eigenda/pull/2147/commits/44c729b26b1df5f44f6687efc6b90073fdcfd126)

api/clients/v2/validator/internal/blob_decoder.go
```diff
@@ -13,44 +15,54 @@ type BlobDecoder interface {
...
 // NewBlobDecoder creates a new BlobDecoder instance.
-func NewBlobDecoder(verifier *verifier.Verifier) BlobDecoder {
+func NewBlobDecoder(encoder *rs.Encoder) BlobDecoder {
 	return &blobDecoder{
-		verifier: verifier,
+		encoder: encoder,
 	}
 }
 
 func (d *blobDecoder) DecodeBlob(
```
litt3, line 45: Seems like the concept of a `BlobDecoder` is just unnecessary.

Not saying it _necessarily_ should happen on this PR, if it expands the scope. But at first glance, seems like it might not really be that large of a diff?

disperser/batcher/batcher.go
```diff
@@ -65,7 +65,7 @@ type Config struct {
 
 	FinalizationBlockDelay uint
 
-	TargetNumChunks          uint
+	TargetNumChunks          core.ChunkNumber
```
litt3: You made this a `ChunkNumber`, but in `StreamerConfig` you made `TargetNumChunks` a `uint64`

tbh I don't see value in the `ChunkNumber` type, but this should probably be consistent

api/clients/v2/validator/internal/blob_decoder.go
samlaf: I agree, I don't understand this part of the code, and I didn't bother trying to. @cody-littley wrote it I think? Not sure why its abstracted behind these Factory methods that don't do much. Seems like there would have been a much simpler way to write it, but not sure..

disperser/batcher/batcher.go
samlaf: agree. I did this quickly.... it's really a huge mess. Was just trying to get the darn types to match.
I think the value in having the ChunkNumber type is just that we would keep things consistent... otherwise right now some places define it at uint32, others uint, others uint64. core.ChunkNumber and encoding.ChunkNumber are both redefined also in case you didn't know.... I would want to fix this but gosh it's so ugly I decided I don't care enough and will just make the minimal changes to make the stuff I want type-check and not force extra allocations just to make slice type conversions.

Moved it back to be a uint64: [f208162](https://github.com/Layr-Labs/eigenda/pull/2147/commits/f208162bf145e9a39272b716fddbaa3b5d7f4721)

PR #2150, opened 2025-10-10
fix(committer): make G2TrailingSRSPath optional

samlaf:
When I refactored committer into its own struct in https://github.com/Layr-Labs/eigenda/pull/2097, I was probably only thinking of the client/proxy usecase, where we always pass a G2TrailingSRS file path. However in the apiserver, which runs on aws, we point G2SRSPath to the full srs file living on EFS which contains all points. So G2TrailingSRS path should be optional, and only needed if G2SRSPath doesn't contain all points.

Updated constructor to implement this logic (copied from the prover-v2).

We should eventually refactor this into a srs loading utils that is only run once and injects points into the different structs.

### DIFF
--- a/encoding/kzg/committer/committer.go
+++ b/encoding/kzg/committer/committer.go
@@ -52,13 +52,15 @@ func New(g1SRS []bn254.G1Affine, g2SRS []bn254.G2Affine, g2TrailingSRS []bn254.G
 }
 
 type Config struct {
-	// Number of SRS points to load from all 3 SRS files. Must be a power of 2.
+	// Number of SRS points to load from SRS files. Must be a power of 2.
 	// Committer will only be able to compute commitments for blobs of size up to this number of field elements.
 	// e.g. if SRSNumberToLoad=2^19, then the committer can compute commitments for blobs of size up to
 	// 2^19 field elements = 2^19 * 32 bytes = 16 MiB.
-	SRSNumberToLoad   uint64
-	G1SRSPath         string
-	G2SRSPath         string
+	SRSNumberToLoad uint64
+	G1SRSPath       string
+	G2SRSPath       string
+	// G2TrailingSRSPath is optional and only needed if G2SRSPath does not
+	// point to the full SRS file (containing 2^28 points).
 	G2TrailingSRSPath string
 }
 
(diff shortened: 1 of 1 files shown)


encoding/kzg/committer/committer.go
```diff
@@ -84,20 +86,51 @@ func NewFromConfig(config Config) (*Committer, error) {
...
-	if err != nil {
-		return nil, fmt.Errorf("read G2 trailing points from %s: %w", config.G2TrailingSRSPath, err)
+
+	var g2TrailingSRS []bn254.G2Affine
+	hasG2TrailingFile := len(config.G2TrailingSRSPath) != 0
+	if hasG2TrailingFile {
+		// TODO(samlaf): this function/check should probably be done in ReadG2PointSection
+		numG2point, err := kzg.NumberOfPointsInSRSFile(config.G2TrailingSRSPath, kzg.G2PointBytes)
+		if err != nil {
+			return nil, fmt.Errorf("number of points in srs file %v: %w", config.G2TrailingSRSPath, err)
```
pschork, line 96:
```suggestion
			return nil, fmt.Errorf("failed to resolve number of points in srs file %v: %w", config.G2TrailingSRSPath, err)
```

encoding/kzg/committer/committer.go
```diff
@@ -84,20 +86,51 @@ func NewFromConfig(config Config) (*Committer, error) {
...
+		)
+		if err != nil {
+			return nil, fmt.Errorf("failed to read G2 trailing points (%v to %v) from file %v: %w",
+				numG2point-config.SRSNumberToLoad, numG2point, config.G2TrailingSRSPath, err)
+		}
+	} else {
+		// require entire G2SRSPath to contain all 2^28 points, from which we can read the trailing points
+		numG2point, err := kzg.NumberOfPointsInSRSFile(config.G2SRSPath, kzg.G2PointBytes)
+		if err != nil {
+			return nil, fmt.Errorf("number of points in srs file: %w", err)
```
pschork, line 119:
```suggestion
			return nil, fmt.Errorf("failed to resolve number of points in srs file: %w", err)
```

encoding/kzg/committer/committer.go
samlaf: I've been explicitly against these prefixes for a while, and trying to remove them all over the codebase. See https://github.com/uber-go/guide/blob/master/style.md#error-wrapping for rationale.

encoding/kzg/committer/committer.go
samlaf: see https://github.com/Layr-Labs/eigenda/pull/2150#discussion_r2421577418

PR #2151, opened 2025-10-10
feat(payments): Write expiring reservation test

litt3:
- Create integration tests to assert expected behavior when a reservation expires
- Does a little cleanup to the payment test utility

### DIFF
--- a/core/payments/clientledger/client_ledger_test.go
+++ b/core/payments/clientledger/client_ledger_test.go
@@ -277,7 +277,7 @@ func TestReservationOnly(t *testing.T) {
 		})
 	})
 
-	t.Run("time out of range panic", func(t *testing.T) {
+	t.Run("time out of range error", func(t *testing.T) {
 		clientLedger := NewClientLedger(
 			ctx,
 			test.GetLogger(),
@@ -292,9 +292,11 @@ func TestReservationOnly(t *testing.T) {
 		)
 		require.NotNil(t, clientLedger)
 
-		require.Panics(t, func() {
-			_, _ = clientLedger.Debit(ctx, 1, []core.QuorumID{0, 1})
-		}, "expired reservation should cause fatal panic")
+		paymentMetadata, err := clientLedger.Debit(ctx, 1, []core.QuorumID{0, 1})
+		require.Error(t, err, "time out of range should cause error")
+		require.Nil(t, paymentMetadata)
+		var timeOutOfRangeErr *reservation.TimeOutOfRangeError
+		require.ErrorAs(t, err, &timeOutOfRangeErr)
 	})
 }
 


inabox/tests/payments/payments_test.go
```diff
@@ -264,10 +266,83 @@ func testReservationIncrease(
...
-		t, testRandom, payloadDisperser, blobsPerSecond, payloadBytes, submissionDuration, 1.0, 0)
-	for range resultChan {
-	}
+	mustSubmitPayloads(t, testRandom, payloadDisperser, blobsPerSecond, payloadBytes, submissionDuration, 1.0, 0)
+}
+
+// - Create a reservation that expires soon
+// - Submit a blob and assert success
+// - Sleep until reservation expires
+// - Assert next blob submission panics (client ledger panics on expired reservation)
```
samlaf, line 276: interesting. why did we go with panic here? I guess its the best way to alert people of the expiration?

inabox/tests/payments/payload_submitter.go
```diff
@@ -25,82 +23,72 @@ func mustSubmitPayloads(
...
+			return
+		case <-ticker.C:
+			wg.Add(1)
+			go func() {
+				defer wg.Done()
+				currentBlob := blobCount.Add(1)
+				payload := coretypes.Payload(testRandom.Bytes(payloadSize))
+				timestamp := time.Since(startTime)
+
+				t.Logf("[%s] Dispersing blob #%d...", timestamp, currentBlob)
```
samlaf, line 80: nit: eventually we should create an slog testlogger that wraps t.Log to give us structured test logs that only appear when a test fails.

inabox/tests/payments/payments_test.go
litt3: This was originally put in place before I had the utility that detects on-chain updates.

I've tweaked the behavior of the `ClientLedger`:
- if in `ReservationOnly` mode, _don't_ `panic`, only error.
    - you can recover with a new reservation
- if in `ReservationAndOnDemand` mode, _do_ `panic` if a reservation expires
    - Theoretically you could recover with a new reservation, or simply continue by using on-demand funds. The problem is that if a user has on-demand funds deposited as a backup payment mode, they could have their reservation expire and not notice. That would burn through all on-demand funds, which doesn't seem right. 

See this commit for the details. What do you think of this?

249dfc472ef9a8cd01e7e7fe1e26d823d7297c70

inabox/tests/payments/payload_submitter.go
litt3: I have the `make` command using `gotestsum`, which does accomplish this for now. Do you think a test utility is necessary even with `gotestsum` available?

samlaf: ah I see. And gotestsum basically always runs with `-v` and then just hides the output of the tests that fails? That works actually... but I guess still 2 reasons for moving to slog based testlogger:
1. consistency throughout repo. Just better both for humans and LLMs to always log the same way imo
2. when running tests from IDE instead of from makefile

inabox/tests/payments/payments_test.go
samlaf: I see. Few questions:
1. is there even a use for `ReservationAndOnDemand` mode then if it just panics when the reservation expires? Isn't that equivalent to using `ReservationOnly`? Or are there still cases where it would use onDemand?
2. panic vs error probably ultimately behave the same given clients' main use case is inside proxy which catches panics and returns 500s. Main question is whether we really want this to be returning a 500 to customers... or if we would instead prefer to return a https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Status/402 or whatever

litt3: 1. The intended usage of `ReservationAndOnDemand` is to disperse more than is supported by a reservation. E.g. "I have a reservation for 1MiB/second, but I occasionally exceed that rate, and pay for the difference via on-demand". In this setup, some blobs will be paid for with reservation, some with on-demand.
2. But the panic will kill the clients, right? So the difference is that with an error dispersals will continue to be attempted, whereas with a `panic` it won't.

inabox/tests/payments/payload_submitter.go
litt3: The utility you're imagining would have the ability to _not_ suppress log output, right?

i.e., Sometimes when I'm running a test, I want to watch all logs fly by, whether the tests fail or not

samlaf: It would behave the exact same as t.Log, just with structured log instead of manually structured logs like what LLM is generating here. If you want to watch all logs, then you have to run the test with `-v`.

inabox/tests/payments/payments_test.go
samlaf: 1. gotcha
2. by "kill the clients" you mean "kill the entire goroutine immediately" I guess? Aka bypass all retry logic. Ultimately the "right" way would probably be to return a structured error which the retry loops would bypass, and the error middleware would turn into a 402 or whatever. But I don't think its that necessary...

litt3: > return a structured error

We're already returning a `TimeOutOfRangeError`. Do we just need to add handling of this error to the middleware?

samlaf: I guess so yes, although you'd need to know whether the client was in ondemand or reservation mode? Also proxy's retry loop would need to be aware of the same... and not even sure returning a 402 is actually what we want. lets chat live

litt3: As we discussed offline, we'll sync about this and fix in a followup

PR #2153, opened 2025-10-11
chore: bump icicle 3.4.0->3.9.0

samlaf:
This is needed to use icicle on macos (in fact afair 4.0.0 did not work, as they don't have a metal compatible release of 4.0.0 yet).

Have not tested that this works on CUDA. We should do that before merging this.

### DIFF
--- a/.github/workflows/docker-publish-encoder-icicle.yaml
+++ b/.github/workflows/docker-publish-encoder-icicle.yaml
@@ -1,0 +1,81 @@
+
+# NOTE: encoder-icicle is built in a separate workflow (instead of being included in the main
+# docker-publish workflow) because:
+# 1. It uses a different Dockerfile (icicle.Dockerfile) with GPU-specific dependencies (ICICLE library)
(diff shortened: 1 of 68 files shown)


go.mod
```diff
@@ -42,7 +42,7 @@ require (
 	github.com/grpc-ecosystem/go-grpc-middleware/providers/prometheus v1.0.1
 	github.com/grpc-ecosystem/go-grpc-middleware/v2 v2.1.0
 	github.com/hashicorp/go-multierror v1.1.1
-	github.com/ingonyama-zk/icicle/v3 v3.4.0
+	github.com/ingonyama-zk/icicle/v3 v3.9.0
```
dmanc: In the dockerfile we use 3.9.2 but here we use 3.9.0? Should we just use 3.9.2 for both?

samlaf: good catch [e57c302](https://github.com/Layr-Labs/eigenda/pull/2153/commits/e57c3023fdef04d299bb31153167603d4ebc0de2)

PR #2156, opened 2025-10-13
chore: cleanup prover-v2 srs config

samlaf:
Had left a lot of unneeded stuff after splitting off committer from the prover. No longer need to load any G2 SRS points.
Now configs look like:
```
# prover
type KzgConfig struct {
	SRSNumberToLoad uint64
	G1Path string
	PreloadEncoder bool
	CacheDir string
	NumWorker uint64
}
# verifier
type KzgConfig struct {
	SRSNumberToLoad uint64
	G1Path string
	NumWorker uint64
}
# committer
type Config struct {
  SRSNumberToLoad uint64
  G1SRSPath       string
  G2SRSPath         string
  G2TrailingSRSPath string
}
```
which makes it more obvious for example that G2 points are only needed to create commitments (on clients and apiserver only, not encoder).

### DIFF
--- a/encoding/bench/benchmark_eigenda_test.go
+++ b/encoding/bench/benchmark_eigenda_test.go
@@ -113,13 +113,11 @@ func BenchmarkBlobToChunksEncoding(b *testing.B) {
 // Where fft1 and fft2 are on G1, and preproc contains an FFT on Fr elements.
 func BenchmarkMultiproofFrameGeneration(b *testing.B) {
 	proverConfig := prover.KzgConfig{
-		G1Path:          "../../resources/srs/g1.point",
-		G2Path:          "../../resources/srs/g2.point",
-		G2TrailingPath:  "../../resources/srs/g2.trailing.point",
-		CacheDir:        "../../resources/srs/SRSTables",
 		SRSNumberToLoad: 1 << 19,
+		G1Path:          "../../resources/srs/g1.point",
+		PreloadEncoder:  true,
+		CacheDir:        "../../resources/srs/SRSTables",
 		NumWorker:       uint64(runtime.GOMAXPROCS(0)),
-		LoadG2Points:    true,
 	}
 	p, err := prover.NewProver(&proverConfig, nil)
 	require.NoError(b, err)


encoding/kzg/prover/v2/config.go
```diff
@@ -7,47 +7,27 @@ import (
 
 // KzgConfig holds configuration for the V2 KZG prover.
 type KzgConfig struct {
```
dmanc, line 9: Should we just call this Config at this point? Same comment for verifier.

encoding/backend.go
```diff
@@ -22,7 +22,6 @@ type Config struct {
 	NumWorker   uint64
 	BackendType BackendType
 	GPUEnable   bool
-	Verbose     bool
```
dmanc, line 25: No longer used?

encoding/kzg/verifier/v2/verifier_test.go
```diff
@@ -105,7 +105,7 @@ func TestUniversalVerifyWithPowerOf2G2(t *testing.T) {
 	require.NoError(t, err)
 
 	params := encoding.ParamsFromSysPar(harness.numSys, harness.numPar, uint64(len(harness.paddedGettysburgAddressBytes)))
-	enc, err := group.GetKzgEncoder(params)
+	enc, err := group.GetKzgProver(params)
```
dmanc: nit: enc -> prover?

encoding/kzg/prover/v2/prover.go
```diff
@@ -258,7 +194,7 @@ func (g *Prover) preloadAllEncoders() error {
 
 	for _, params := range paramsAll {
 		// get those encoders and store them
-		enc, err := g.GetKzgEncoder(params)
+		enc, err := g.GetKzgProver(params)
```
dmanc: nit: same comment enc-> prover

encoding/backend.go
samlaf: doesn't seem like it..

samlaf: I don't think they are used anywhere tbh..

encoding/kzg/prover/v2/config.go
samlaf: Had similar thought. Renamed the verifier one in [5645972](https://github.com/Layr-Labs/eigenda/pull/2156/commits/56459721748ca8858a843bbe12e12f9e36e96e1f)
But don't want to do it for prover b/c prover still holds an encoder, and is passed both a KZGConfig as well as an encoderConfig.
I guess we could still rename KzgConfig to just config... but I think it'll make more sense if/when we separate the encoder from the prover, and then prover only has a single config (the KzgConfig).

Honestly I think the proper refactor would be to further get rid of these kzgconfig, and instead just load SRSPoints once, and inject them into the prover/verifier/committer when needed, as opposed to each of their constructor having to reload the srs points.

encoding/kzg/verifier/v2/verifier_test.go
samlaf: fixed in a bunch more places: [d0f2277](https://github.com/Layr-Labs/eigenda/pull/2156/commits/d0f2277a4bfa73e49498e37752bb99fea3a18594)

encoding/kzg/prover/v2/prover.go
samlaf: fixed in a bunch more places: [d0f2277](https://github.com/Layr-Labs/eigenda/pull/2156/commits/d0f2277a4bfa73e49498e37752bb99fea3a18594)

encoding/kzg/prover/v2/prover.go
```diff
@@ -18,16 +18,14 @@ import (
 	_ "go.uber.org/automaxprocs"
 )
 
-// Prover is the main struct that is able to generate KZG commitments and proofs.
-// TODO(samlaf): we should split the kzg commitment functionality into its own struct/service.
-// In EigenDA V2, commitments are generated client side (or on the API server) and proofs are generated on the encoder.
-// It would make it a lot cleaner, as well as more explicit which exact SRS points are needed for which functionality.
+// Prover is the main struct that is able to generate frames (chunks and their proofs).
+// TODO(samlaf): should we refactor prover to only generate proofs and keep encoding separate?
```
bxue-l2, line 22: for distributed retrieval, it only needs to use RS and proof verifier

encoding/kzg/prover/v2/prover.go
```diff
@@ -143,12 +79,12 @@ func (e *Prover) GetFrames(data []byte, params encoding.EncodingParams) ([]*enco
...
 
-	enc, err := e.GetKzgEncoder(params)
+	prover, err := e.GetKzgProver(params)
 	if err != nil {
-		return nil, fmt.Errorf("get kzg encoder: %w", err)
+		return nil, fmt.Errorf("get kzg prover: %w", err)
 	}
 
-	kzgFrames, _, err := enc.GetFrames(symbols)
+	kzgFrames, _, err := prover.GetFrames(symbols)
```
bxue-l2, line 87: why is it called kzg-frame? given it only has chunks

encoding/kzg/prover/v2/prover.go
samlaf: hmm, how is this related to the prover comment here?

encoding/kzg/prover/v2/prover.go
samlaf: not sure I'm following.. it's called kzg-frame because the function is called `GetFrames`. Furthermore it returns both coeffs (what you call chunks?) and proofs.

Is this the nomenclature you use(?):
1. frame = (coeffs, proofs)
2. chunk = frame.coeffs

encoding/kzg/prover/v2/prover.go
bxue-l2: oh, we already has the rs separted

encoding/kzg/prover/v2/prover.go
bxue-l2: it says kzg, and I always associate the name with proof. But here is just data. Just nit.

samlaf: wdym by just the data? it liteally returns the proofs.

bxue-l2: good

PR #2157, opened 2025-10-14
fix(sigmap-EDA-06): Division by zero vulnerability in payment system

iquidus:
Closes DAINT-780

Addresses EDA-06 from the [Sigma Prime audit](https://linear.app/eigenlabs/issue/DAINT-774/address-audit-findings) report by adding validation to the `SetPaymentState()` function to reject zero values for `reservationWindow`.

A division by zero vulnerability exists in the payment system that can cause immediate client crashes when processing payment operations. The vulnerability occurs when the smart contract's `reservationPeriodInterval` parameter is set to zero, which gets synchronized to the Go client as `reservationWindow`. The issue occurs in the `getOrRefreshRelativePeriodRecord()` function in `accountant.go`

### DIFF
--- a/api/clients/v2/accountant_test.go
+++ b/api/clients/v2/accountant_test.go
@@ -1179,6 +1179,41 @@ func TestSetPaymentState(t *testing.T) {
 				periodRecords:     []PeriodRecord{},
 			},
 		},
+		{
+			name: "error if ReservationWindow is set to 0",
+			state: &disperser_rpc.GetPaymentStateReply{
+				PaymentGlobalParams: &disperser_rpc.PaymentGlobalParams{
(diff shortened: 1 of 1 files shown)


api/clients/v2/accountant.go
samlaf, line 1: is the accountant still being used in your new payment stuff @litt3 ?

samlaf: if not this is very low priority but no issue with merging that I can see

litt3: > is the accountant still being used in your new payment stuff

No, `accountant` will be fully removed as soon as the migration is complete.

litt3 (approved):

PR #2162, opened 2025-10-15
fix(sigmap-EDA-08): Missing length validation for keccak commitment

iquidus:
Closes DAINT-782

Addresses EDA-07  from the [Sigma Prime audit](https://linear.app/eigenlabs/issue/DAINT-774/address-audit-findings) report by adding explicit length validation after hex decoding the keccak commitment. Returns a clear parsing error when the commitment length is not exactly 32 bytes.

The `handleGetOPKeccakCommitment()` function does not validate that the decoded Keccak commitment is exactly 32
bytes, allowing malformed requests to be processed until they fail at a later stage. The function decodes the hex-encoded Keccak commitment from the request path but does not verify its length before passing it to the storage layer. Keccak256 hashes are always 32 bytes, and the system expects this invariant to hold throughout the processing pipeline.

### DIFF
--- a/api/proxy/servers/rest/handlers_cert.go
+++ b/api/proxy/servers/rest/handlers_cert.go
@@ -145,10 +145,6 @@ func (svr *Server) handlePostOPKeccakCommitment(w http.ResponseWriter, r *http.R
 		return proxyerrors.NewParsingError(
 			fmt.Errorf("failed to decode hex keccak commitment %s: %w", keccakCommitmentHex, err))
 	}
-	if len(keccakCommitment) != 32 {
-		return proxyerrors.NewParsingError(
-			fmt.Errorf("keccak commitment must be 32 bytes, got %d bytes", len(keccakCommitment)))
-	}
 
 	payload, err := io.ReadAll(http.MaxBytesReader(w, r.Body, common.MaxServerPOSTRequestBodySize))
 	if err != nil {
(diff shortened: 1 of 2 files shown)


api/proxy/servers/rest/handlers_cert.go
```diff
@@ -37,6 +37,10 @@ func (svr *Server) handleGetOPKeccakCommitment(w http.ResponseWriter, r *http.Re
 		return proxyerrors.NewParsingError(
 			fmt.Errorf("failed to decode hex keccak commitment %s: %w", keccakCommitmentHex, err))
 	}
+	if len(keccakCommitment) != 32 {
+		return proxyerrors.NewParsingError(
+			fmt.Errorf("keccak commitment must be 32 bytes, got %d bytes", len(keccakCommitment)))
+	}
```
samlaf: prob better to do this check in the routing logic itself, otherwise we could end up decoding a gazillion-bytes-length string.

Think the right way is to use a regexp in https://github.com/Layr-Labs/eigenda/blob/c90d2f37d70db716f3b9e683e6a9d0b6f64a029c/api/proxy/servers/rest/routing.go#L37 to make sure the length is exactly 32 bytes. Not sure why I didn't do this the first time around since I did it for other fields.

samlaf (approved):
LGTM although prob wouldn't have hurt to keep the check in the handler itself, just in case.

PR #2166, opened 2025-10-15
feat(payments): Document proxy payment flag

litt3:
closes EGDA-2012

- Gives succinct description of how to interact with the payment mode configuration value

### DIFF
--- a/api/proxy/README.md
+++ b/api/proxy/README.md
@@ -317,10 +317,10 @@ active for blob dispersals. For detailed information about the payment system, s
 > these configurations.
 
 2. **`reservation-only`** - Uses pre-purchased bandwidth reservations that provide guaranteed throughput for a
-   specified time period. Bandwidth is managed using a leaky bucket algorithm. Dispersals will fail if a reservation
-   is temporarily exhausted.
+   specified time period. Reservations are tracked in the `PaymentVault` contract, and bandwidth is managed using a
+   leaky bucket algorithm. Dispersals will fail if a reservation is temporarily exhausted.
 
-3. **`on-demand-only`** - Uses pay-per-dispersal payments from funds deposited in the PaymentVault contract.
+3. **`on-demand-only`** - Uses pay-per-dispersal payments from funds deposited in the `PaymentVault` contract.
    Limited to quorums 0 (ETH) and 1 (EIGEN). Dispersals will fail if on-demand funds are exhausted.
 
 4. **`reservation-and-on-demand`** - Enables both reservation and on-demand payment methods with intelligent fallback.
@@ -328,7 +328,7 @@ active for blob dispersals. For detailed information about the payment system, s
    capacity is temporarily exhausted. If a reservation *expires*, this mode will prevent any dispersals from being made
    to avoid inadvertent draining of on-demand funds due to an expired reservation.
 
-> **Note**: The payment mode should match your account's setup in the PaymentVault contract. Ensure you have an active
+> **Note**: The payment mode should match your account's setup in the `PaymentVault` contract. Ensure you have an active
 > reservation (for `reservation-only` or `reservation-and-on-demand`) or sufficient deposits (for `on-demand-only` or
 > `reservation-and-on-demand`) before starting the proxy.
 


api/proxy/README.md
```diff
@@ -283,7 +283,40 @@ $ set -a; source ./.env; set +a; ./bin/eigenda-proxy
 
 ### Features and Configuration Options (flags/env vars)
 
-Below is a list of the main high-level features offered for configuring the eigenda-proxy. These features are controlled via flags and/or env vars. To view the extensive list of available flags/env-vars to configure a given version of eigenda-proxy, run `eigenda-proxy --help`.
+Below is a list of the main high-level features offered for configuring the eigenda-proxy. These features are
+controlled via flags and/or env vars. To view the extensive list of available flags/env-vars to configure a given
+version of eigenda-proxy, run `eigenda-proxy --help`.
+
+#### Payment Mode Configuration
```
samlaf: Under `Requirements` section below, there's a `#### Authn/Authz/Payments` section. Would be good to consolidate with this section here.

api/proxy/README.md
```diff
@@ -283,7 +283,40 @@ $ set -a; source ./.env; set +a; ./bin/eigenda-proxy
...
+active for blob dispersals. For detailed information about the payment system, see the
+[payment system documentation](../../docs/spec/src/protocol/payments/payment_system.md).
+
+**Available Payment Modes:**
+
+1. **`legacy` (default)** - Uses the legacy bin-based payment system that handles both reservation and on-demand
+   payments. This mode is in the process of being deprecated and will be removed in a future release.
+   
+> **IMPORTANT**: All clients should continue using this mode until the new payment system has officially shipped. The
+> other payment modes are documented below for awareness, but they aren't ready to be used in production environments.
```
samlaf: shouldn't we say that they aren't supported by our currently deployed dispersers?

"not ready to be used in prod environments" still has the impliction that "if you are risk tolerant, you might want to use this beta feature"

api/proxy/README.md
```diff
@@ -283,7 +283,40 @@ $ set -a; source ./.env; set +a; ./bin/eigenda-proxy
...
+
+2. **`reservation-only`** - Uses pre-purchased bandwidth reservations that provide guaranteed throughput for a
```
samlaf, line 319: wwould be good to mention the reservation is also tracked in the PaymentVault

api/proxy/README.md
```diff
@@ -283,7 +283,40 @@ $ set -a; source ./.env; set +a; ./bin/eigenda-proxy
...
+   specified time period. Bandwidth is managed using a leaky bucket algorithm. Dispersals will fail if a reservation
+   is temporarily exhausted.
+
+3. **`on-demand-only`** - Uses pay-per-dispersal payments from funds deposited in the PaymentVault contract.
+   Limited to quorums 0 (ETH) and 1 (EIGEN). Dispersals will fail if on-demand funds are exhausted.
+
+4. **`reservation-and-on-demand`** - Enables both reservation and on-demand payment methods with intelligent fallback.
+   Uses reservation bandwidth when available, and automatically switches to on-demand payments when reservation
+   capacity is temporarily exhausted. If a reservation *expires*, this mode will prevent any dispersals from being made
+   to avoid inadvertent draining of on-demand funds due to an expired reservation.
```
samlaf, line 329: nit: perhaps we should separate this into 2 modes? Something like "reservation-with-ondemand-excess" for this current mode and "reservation-and-on-demand" for literally allowing both.

Thinking there might be a team with a reservation expiring soon that wants to switch over to on-demand. Seems like don't currently provide a way to allow this smooth transition?

ethenotethan (approved):
LGTM

api/proxy/README.md
litt3: 8740f7b6b34edce979f23bcec31eaeb612e27d2f

api/proxy/README.md
litt3: > "if you are risk tolerant, you might want to use this beta feature"

Well, I think this is pretty much correct. But I agree we should more explicitly discourage usage of the new payments for now

3a5ad0f67911da7a95b57496d09c473504863717

api/proxy/README.md
litt3: 49226963c14790a5fafd4e5babd4e33dd0ac2cc6

api/proxy/README.md
litt3: I agree this makes sense to do.

I created a linear ticket to capture the idea, and will implement as a follow up at some point

api/proxy/README.md
samlaf: Wait explain how this is correct? Is it *possible* to use the new configs on clients because the rate limiting would be done differently, but would still work (like disperser would be using old bins stuff but there's nothing forcing client and disperser to use same algo)?

litt3: Correct, it's possible to use the new configuration on clients, while the disperser still uses the old logic. 

This is actually being tested in the payments inabox tests

PR #2168, opened 2025-10-15
fix(sigmap-EDA-05): Unlimited hex string length.

iquidus:
Closes DAINT-779

Addresses EDA-05  from the [Sigma Prime audit](https://linear.app/eigenlabs/issue/DAINT-774/address-audit-findings) report by adding length validation to the hex string parameters before processing with reasonable upper bounds.

The routing pattern for OP generic endpoints accepts unlimited-length hex strings, potentially allowing denial-of-
service attacks through memory exhaustion.

### DIFF
--- a/api/proxy/servers/rest/cli.go
+++ b/api/proxy/servers/rest/cli.go
@@ -18,7 +18,10 @@ const (
 var (
 	// MaxCertSizeBytes is used to set an upper bound for validating a payload hex's length
 	// to harden the rest server against memory exhaustion attacks. It is NOT enforced on
-	// the protocol level. By default there is no upper bound.
+	// the protocol level. By default there is no upper bound. If this value is set too low
+	// requests containing valid certificates that exceed this limit will return a 404. It is
+	// strongly recommended to leave this as the default unless the api is exposed publicly
+	// and is experiencing attacks in which this setting may help mitigate.
 	MaxCertSizeFlagName = withFlagPrefix("max-certificate-size")
 )
 


api/proxy/servers/rest/cli.go
```diff
@@ -13,6 +13,12 @@ const (
 
 	DeprecatedAPIsEnabledFlagName = "api-enabled"
 	DeprecatedAdminAPIType        = "admin"
+
+	// MaxCertSizeBytes is used as an upper bound for validating a payload hex's length
+	// to harden the rest server against memory exhaustion attacks. It is NOT enforced on
+	// the protocol level. If we are expecting an increase in cert size (e.g: a significant
+	// increase in operators) this value needs to be updated accordingly.
+	MaxCertSizeBytes int = 21000
```
samlaf: where is MaxBlobSize defined? feel like those 2 should be defined next to each other. Also is 21k enough? Would be good to add some comment explaining how 21k was chosen, and how many non-signers/quorums roughly it allows.

samlaf: Also want to get @litt3 's thoughts written down here, since he was against having this in proxy.

api/proxy/servers/rest/handlers_cert.go
```diff
@@ -88,6 +88,11 @@ func (svr *Server) handleGetShared(
 	if !ok {
 		return proxyerrors.NewParsingError(fmt.Errorf("serializedDACert not found in path: %s", r.URL.Path))
 	}
+	// Check the size of the hex payload (in bytes) does not exceed MaxCertSizeBytes
+	if len(serializedCertHex)/2 > MaxCertSizeBytes {
```
samlaf: just like in the other PR, the issue with doing this check here is that it doesn't prevent DDoS attacks so its a pretty moot check. Because the entire cert is parsed when reading `mux.Vars(r)[routingVarNamePayloadHex]` above.

api/proxy/servers/rest/cli.go
litt3: Yeah, essentially my stance is that we shouldn't build hard limits like this into the proxy.

1. Publicly exposing the proxy isn't a supported use case. If we want to officially support this, we should make the decision and stick to it. But that would mean a LOT of other changes are necessary, too.
2. This limit isn't defined at a protocol level, so if we ever managed to exceed this size for a valid cert, the behavior would be very bad

If we feel we absolutely _need_ to place this limit (which I don't think we do), IMO it would be better as a configurable value that proxy owners can configure. We could set the default to `unlimited`, and anyone who cares to change that default would be free to do so. But at that point, we're adding non-trivial complexity and engineering effort to solve a problem for a use-case we don't actually support...

api/proxy/servers/rest/handlers_cert.go
```diff
@@ -88,6 +88,11 @@ func (svr *Server) handleGetShared(
...
+		return proxyerrors.NewParsingError(
+			fmt.Errorf("%s can not exceed %d bytes", routingVarNamePayloadHex, MaxCertSizeBytes))
```
ethenotethan: what's to stop a rollup batch poster from constructing a `DA Cert` that exceeds this limit, passing `/put`, getting published to a sequencer inbox, and failing here? iiuc this check introduces a trust assumption and could be abused to cause liveness failures for a rollup derivation pipeline by a faulty sequencer

api/proxy/servers/rest/cli.go
ethenotethan: a bit curious why we aren't aligning this to calldata limits

samlaf: +1 to using unlimited, and documenting why we recommend not having a limit. ethen's point in https://github.com/Layr-Labs/eigenda/pull/2168#discussion_r2445834160 is very valid, and its hard to even know what the max calldata size is, because it could change per hardfork, which we'd need to track, etc.

api/proxy/servers/rest/handlers_cert.go
iquidus: In this case we can't handle it with gorilla mux's regexp. the Go standard regexp package has a restriction that sets a maximum limit of 1000 on bounded quantifiers like {min, max}

[0-9a-fA-F]{0,n} when n is <= 1000 works as expected
[0-9a-fA-F]{0,n} when n is > 1000 does not compile, all requests 404

api/proxy/servers/rest/cli.go
iquidus: MaxBlobSize is defined in the v2 client configs, which doesn't make sense in this case as MaxCertSizeBytes is only used by the rest server. I have moved it to the rest server config, with a default value of 0 (unlimited).

api/proxy/docs/help_out.txt
```diff
@@ -121,8 +121,9 @@ This check is optional and will be skipped when set to 0. (default: 0) [$EIGENDA
 
    Proxy REST API Server (compatible with OP Stack ALT DA and standard commitment clients)
 
-   --addr value  Server listening address (default: "0.0.0.0") [$EIGENDA_PROXY_ADDR]
-   --port value  Server listening port (default: 3100) [$EIGENDA_PROXY_PORT]
+   --addr value                       Server listening address (default: "0.0.0.0") [$EIGENDA_PROXY_ADDR]
+   --port value                       Server listening port (default: 3100) [$EIGENDA_PROXY_PORT]
+   --rest.max-certificate-size value  Maximum certificate size (bytes) to accept in API requests. Setting to (0) results in no maximum size. (default: 0) [$EIGENDA_PROXY_MAX_CERTIFICATE_SIZE]
```
samlaf, line 126: I don't think we should be exposing this personally.
1. rest feels like the wrong category. cert size is an eigenda concept
2. I think if we really want to be discarding large certs then rollups would need to be aware of this, so this should actually be returning a 418 to indicate wrong cert, and rollup would have to provably discard the cert from their derivation pipeline as well.

Overall I think this is a non-trivial issue, as ethen pointed out. Hmm... not even sure what to do with this anymore. Let's sync with the team during standup before making you do any back and forths on this PR... sorry.

api/proxy/servers/rest/server.go
```diff
@@ -20,9 +20,10 @@ import (
 
 // Config ... Config for the proxy HTTP server
 type Config struct {
-	Host        string
-	Port        int
-	APIsEnabled *enablement.RestApisEnabled
+	Host             string
+	Port             int
+	APIsEnabled      *enablement.RestApisEnabled
+	MaxCertSizeBytes int
```
samlaf, line 26: as I said in my other comment, dont think this belongs here anymore.

api/proxy/servers/rest/handlers_cert.go
samlaf: interesting, good to know.

samlaf: that probably means mux regexp is prob not the right place to be protecting from ddos then... I'd guess even in the other PR gorilla will prob read the whole thing in memory before checking the regexp :\

api/proxy/servers/rest/cli.go
samlaf: Don't think it belongs in the rest server config though. see https://github.com/Layr-Labs/eigenda/pull/2168#discussion_r2455687792
Lets discuss during standup tomorrow. Tempted to just leave a comment in our spec and call it a day. WDYT @bxue-l2

iquidus:
Closing PR.

Addressing this without introducing potential liveness problems, e.g a valid cert exceeding the set limit, requires a significant lift (enforcing the size limit everywhere), as the protocol has no direct limitations on cert size (however, there are indirect limits enforced by the L1 network/nodes). The proxy is not intended to be publicly exposed. Liveness issues for rollups etc are of a much greater concern than a potential memory exhaustion attack on a proxy instance that shouldn't be exposed publicly.

PR #2171, opened 2025-10-15
docs(committer): document how length proofs work

samlaf:
This consolidates all the information that we've amassed over the last few weeks of digging into this in a single place:
1. theory (why we need length proofs)
2. implementation (why the simple approach doesn't work because of our choice of SRS; and how we fixed it. Note that our [spec](https://layr-labs.github.io/eigenda/protocol/architecture/encoding.html#validation-via-kzg) only documented the simple broken approach)
3. also included @gpsanant 's simplified scheme that he proposed last week. Would require a protocol fork to implement though so we likely won't do that... possibly ever lol, but at least it's documented now.

### DIFF
--- a/encoding/v2/kzg/committer/doc.go
+++ b/encoding/v2/kzg/committer/doc.go
@@ -31,7 +31,10 @@
 // - s: secret SRS value
 // - p: polynomial represented by the blob
 // - blob: list of field elements, representing the coefficients of p(x)
-// - [x]_1: KZG commitment of polynomial x in G1
+// - [x]_1: KZG commitment of field element x in G1
+// - [x]_2: KZG commitment of field element x in G2
+// - [p]_1: KZG commitment of polynomial p in G1
+// - [p]_2: KZG commitment of polynomial p in G2
 // - See https://dankradfeist.de/ethereum/2020/06/16/kate-polynomial-commitments.html for math background.
 //
 // In theory, proving an upper bound on the actual blob length is very simple (assuming knowledge of pairings),
@@ -39,7 +42,7 @@
 // - G1 and G2: generators of the bn254 curve groups
 // - BL: blob length (power of 2)
 // - BC_G1: blob commitment; [p]_1 := p(s)G1 (this is the same as our [encoding.BlobCommitments].Commitment)
-// - LC_G1: len commitment; [q(x)]_1=q(s)*G1 where q(x) := x^(2^28-BL)*p(x)
+// - LC_G1: len commitment; [q]_1=q(s)*G1 where q(x) := x^(2^28-BL)*p(x)
 // Verification is simply e(BC_G1, s^(2^28-BL)*G2) = e(LC_G1, G2)
 //
 // Unfortunately, this simple strategy does not work, due to our (unfortunate) choice of SRS ceremony,


encoding/kzg/committer/doc.go
```diff
@@ -0,0 +1,78 @@
...
+// There are 2 facets to data availability:
+// 1. Local (chunks) availability: validator attests to having received and being able to serve its chunks
+// 2. Global (blob) availability: validator attests to the entire blob being available in the network.
+//
+// Because of the sharded nature of EigenDA, each validator only receives a subset of the blob's content.
+// In order to attest to global availability, it thus needs to know how many chunks there are in total,
+// and to make sure that the chunks it receives are actually proportional to its stake. This is why
+// BlobCommitments contains a length field, as well as a proof of this length (LengthCommitment + LengthProof).
+//
+// Here's an example scenario which shows that EigenDA could go wrong without this Length.
```
dmanc: nit: 
```suggestion
// Here's an example scenario which shows that EigenDA could go wrong without this length.
```

encoding/kzg/committer/doc.go
```diff
@@ -0,0 +1,78 @@
...
+// the original blob (need at least 128 for that).
+//
+// ----- Length Commitment + Length Proof Explanation -----
+//
+// In theory, proving an upper bound on the actual blob length is very simple (assuming knowledge of pairings),
+// and would require only a LengthCommitment (no LengthProof needed).
+// - G1 and G2: generators of the bn254 curve groups
+// - BL: blob length (power of 2)
+// - BC_G1: blob commitment; [p(x)]_1 := p(s)G1 (this is the same as our [encoding.BlobCommitments].Commitment)
+// - LC_G1: len commitment; q(x) = x^(2^28-BL)*p(x)
```
dmanc: Was this supposed to be an element of G1 so something like:
```
q(x) = x^(2^28-BL)*p(x) 
[q(x)]_1 = q(s)*G1
```
?

encoding/kzg/committer/doc.go
```diff
@@ -0,0 +1,78 @@
...
+// and sign over the blob’s batch. But now there are only 8 FEs in the system, which is not enough to reconstruct
+// the original blob (need at least 128 for that).
+//
+// ----- Length Commitment + Length Proof Explanation -----
+//
+// In theory, proving an upper bound on the actual blob length is very simple (assuming knowledge of pairings),
+// and would require only a LengthCommitment (no LengthProof needed).
+// - G1 and G2: generators of the bn254 curve groups
+// - BL: blob length (power of 2)
+// - BC_G1: blob commitment; [p(x)]_1 := p(s)G1 (this is the same as our [encoding.BlobCommitments].Commitment)
```
dmanc: Isn't it [p(s)]_1 := p(s)G1, assuming notation from dankrads article https://dankradfeist.de/ethereum/2020/06/16/kate-polynomial-commitments.html

Image

encoding/kzg/committer/doc.go
```diff
@@ -0,0 +1,78 @@
...
+// - BL: actual blob length, same as above
+// - BC_G1: same as above
+// - FBL: fake blob length = BL/2
+// - FLC_G1: fake length commitment to q'(x) = x^(2^28-BL/2)*p(x)
+// Note that if there were only 2^28 G1 points, then the malicious client/disperser would not be able to generate
+// the commitment FLC_G1, because it has degree 2^28-BL/2+BL = 2^28+BL/2 > 2^28
+// - Verification works: e(BC_G1, s^{2^28-BL/2}G2) = e(FLC_G1, G2)
+//
+// So our actual implementation is as follows:
+// - C1 (G1): blob commitment
```
dmanc: The notation changing is kind of confusing, why not stick to the orignal variables like BC_G1 etc.

encoding/kzg/committer/doc.go
```diff
@@ -0,0 +1,78 @@
...
+// Note that EigenDA blob commitments are not simply a single KZG commitment, but also
+// include the blob's length, as well as a proof of this length (LengthCommitment + LengthProof).
+// This complexity stems from the fact that EigenDA, unlike Ethereum which only allows 128KiB blobs,
+// allows blobs of any power-of-2 size between 32B and 16MiB (currently).
+//
+// There are 2 facets to data availability:
+// 1. Local (chunks) availability: validator attests to having received and being able to serve its chunks
+// 2. Global (blob) availability: validator attests to the entire blob being available in the network.
+//
+// Because of the sharded nature of EigenDA, each validator only receives a subset of the blob's content.
```
mmurrs: It would be useful to set the context in the section about the blob size range. Introducing our chunking logic given this adds to the complexity 1. Variable blob lengths 2. Variable chunks (given a time t=0 v time = n).

mmurrs: Do you see this differently?

encoding/v2/kzg/committer/doc.go
```diff
@@ -0,0 +1,78 @@
...
+// Because of the sharded nature of EigenDA, each validator only receives a subset of the blob's content.
+// In order to attest to global availability, it thus needs to know how many chunks there are in total,
+// and to make sure that the chunks it receives are actually proportional to its stake. This is why
+// BlobCommitments contains a length field, as well as a proof of this length (LengthCommitment + LengthProof).
+//
+// Here's an example scenario which shows that EigenDA could go wrong without this Length.
+// In the extreme case, a malicious disperser could just tell the validators that the blob size is 1,
+// and ask all validators except for one to sign off on the commitment. For a slightly more involved but
+// analogous scenario, assume a network of 8 DA nodes with uniform stake distribution, and coding ratio 1/2.
+// For a blob containing 128 field elements (FEs), each node gets 128*2/8=32 FEs, meaning that any 4 nodes can
```
mmurrs, line 21: Is the connection between blobs and field elements commonly understood or does it have to be introduced?

mmurrs: essentially what is the connection between - blob, chunk and field element?

encoding/v2/kzg/committer/doc.go
```diff
@@ -0,0 +1,78 @@
...
+//
+// In theory, proving an upper bound on the actual blob length is very simple (assuming knowledge of pairings),
+// and would require only a LengthCommitment (no LengthProof needed).
+// - G1 and G2: generators of the bn254 curve groups
+// - BL: blob length (power of 2)
+// - BC_G1: blob commitment; [p(x)]_1 := p(s)G1 (this is the same as our [encoding.BlobCommitments].Commitment)
+// - LC_G1: len commitment; q(x) = x^(2^28-BL)*p(x)
+// Verification is simply e(BC_G1, s^(2^28-BL)*G2) = e(LC_G1, G2)
+//
+// Unfortunately, this simple strategy does not work, due to our (unfortunate) choice of SRS ceremony,
```
mmurrs, line 48: are there SRS ceremonies which avoid this? What are the tradeoffs? Feels incomplete if there's not at least partially some explanation around this choice or a gap

mmurrs: It makes it sound this case is unique to our implementation, is that true?

encoding/kzg/committer/doc.go
```diff
@@ -0,0 +1,78 @@
...
+//
+// So our actual implementation is as follows:
+// - C1 (G1): blob commitment
+// - C2 (G2): len commitment to p(x)
+// - P2 (G2): len proof to q(x) = x^{2^28-bloblen}p(x)
+// - Verify e(s^{2^28-bloblen}, C2) = e(G1, P2)
+// Note there is no C1 in above pairing, which is why we verify a second pairing e(C1,G2) = e(G1,C2)
+// in [VerifyCommitEquivalenceBatch]!
+//
+// Note that we actually missed a simpler scheme when initially implementing this,
```
mmurrs: what should someone do with these bottom notes? It's just an acknowledgement we are aware of other elegant solutions?

encoding/v2/kzg/committer/doc.go
samlaf: yes, but they have much fewer points. @bxue made the choice for our SRS points at first so he'd be best placed to answer. My guess is this was the only ceremony that would not restrict us growing to larger and larger blobs.

encoding/kzg/committer/doc.go
samlaf: yes. eigenda might want to fork at some point to implement this better solution, so good to at least keep it written down somewhere.

encoding/v2/kzg/committer/doc.go
samlaf: blob is made up of FEs that are interpreted as coefficients of a polynomial. chunk is... not that easy to describe.
I don't think this is the right place to describe these terms though? Like we have the datatypes defines in other files, plus the spec contains these explanations. Not every doc needs to be completely 100% self-contained no?

encoding/kzg/committer/doc.go
samlaf: Not sure I'm following. Are you saying that there are 2 forms of sharding:
1. sharding through time (each validator holds a subset of blobs)
2. sharding thorugh space (each validator holds a subset of every blob)
? That's a good point that I can mention, but not sure I'm following what you mean by "blob size range" and different times.

encoding/kzg/committer/doc.go
samlaf: [ef79936](https://github.com/Layr-Labs/eigenda/pull/2171/commits/ef799364190db409ad7d8f840ec01b13ee373689)

encoding/kzg/committer/doc.go
samlaf: yes sir: [ef79936](https://github.com/Layr-Labs/eigenda/pull/2171/commits/ef799364190db409ad7d8f840ec01b13ee373689)

encoding/kzg/committer/doc.go
samlaf: I don't really like that notation, because what we are committing to is the polynomial, not p(s).
Basically [p]_1 := [p(s)]_1 = p(s)G

I could use [p] instead of [p(x)] to make it more obvious that we're committing to the full polynomial, as opposed to at some random point x, but I think (?) the notation p(x) makes it more obvious that p is a polynomial without needing another line of legend...? WDYT?

encoding/kzg/committer/doc.go
samlaf: Good point thanks for catching this I was copy-pasting stuff from our google docs which were created by different people at different moment. Changed these in [ef79936](https://github.com/Layr-Labs/eigenda/pull/2171/commits/ef799364190db409ad7d8f840ec01b13ee373689)

Kept the last section with different notation because it uses 2 G1 proofs and a shift notation... do you think that's fine or do you think we should also make those with similar notation like BC1_G1 and BC2_G2?

Btw this made me realize I hate that we use the notation "length commitment" and "length proof", added a comment to that effect. There's literally nothing about the length proof that is more proofy than the length commitment. They are both just commitments which together make up a proof of the length.

encoding/kzg/committer/doc.go
samlaf: rephrase from "subset of the blob" to "subset of every blob" in [ef79936](https://github.com/Layr-Labs/eigenda/pull/2171/commits/ef799364190db409ad7d8f840ec01b13ee373689)

Don't think mentioning the different chunks matters for this high-level discussion. Maybe I'm missing something though, why do you think it does?

encoding/kzg/committer/doc.go
samlaf: added a sentence to mention this in [ef79936](https://github.com/Layr-Labs/eigenda/pull/2171/commits/ef799364190db409ad7d8f840ec01b13ee373689)

encoding/kzg/committer/verify_length_proof.go
```diff
@@ -64,6 +67,11 @@ type CommitmentPair struct {
 	LengthCommitment bn254.G2Affine
 }
 
+// VerifyCommitEquivalenceBath is conceptually part of VerifyLengthProof.
```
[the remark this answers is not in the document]
samlaf: [8790f67](https://github.com/Layr-Labs/eigenda/pull/2171/commits/8790f67d3d74b86aceb2b31eff44523d5b58eea7)

encoding/kzg/committer/doc.go
dmanc: Maybe use capital X, so p(X) if you want to refer to the polynomial object? Also p was never defined earlier so unless you knew KZG commitments you would have to figure that out.

samlaf:
Merged master into here, which moved committer package under `encoding/v2`.

encoding/kzg/committer/doc.go
samlaf: added a notation section in [6388d07](https://github.com/Layr-Labs/eigenda/pull/2171/commits/6388d07d0c0a04ba8a2a01f421fd8bc0244c74d9) and [208ff7a](https://github.com/Layr-Labs/eigenda/pull/2171/commits/208ff7a43b56b7c6f8e560d8789a14488c8a83e5)

PR #2177, opened 2025-10-16
feat(payments): Disperser payment refunds

litt3:
- previous payment implementation on the disperser doesn't allow for payment refunds
- there are a few edgecases where we should refund payments, if an error occurs after debits have occurred, but before a dispersal is complete
- This PR adds the required endpoint on the controller grpc server, and calls it in these edge cases

### DIFF
--- a/api/proxy/README.md
+++ b/api/proxy/README.md
@@ -283,7 +283,54 @@ $ set -a; source ./.env; set +a; ./bin/eigenda-proxy
 
 ### Features and Configuration Options (flags/env vars)
 
-Below is a list of the main high-level features offered for configuring the eigenda-proxy. These features are controlled via flags and/or env vars. To view the extensive list of available flags/env-vars to configure a given version of eigenda-proxy, run `eigenda-proxy --help`.
+Below is a list of the main high-level features offered for configuring the eigenda-proxy. These features are
+controlled via flags and/or env vars. To view the extensive list of available flags/env-vars to configure a given
+version of eigenda-proxy, run `eigenda-proxy --help`.
+
+#### Payments
+
+> Note: Proxy only supports using a single authorization (v1) or payment (v2) key. For RaaS providers, we discourage
+sharing keys between rollups, and thus recommend running a single instance of the Proxy per Rollup.
+
+##### V1 Payments
+
+In order to disperse to the EigenDA V1 network in production, or at high throughput on testnet, please register your
+authentication ethereum address through [this form](https://forms.gle/3QRNTYhSMacVFNcU8). Your EigenDA authentication
+keypair address should not be associated with any funds anywhere. 
+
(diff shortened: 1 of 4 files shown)


api/proto/controller/controller_service.proto
```diff
@@ -24,6 +24,12 @@ service ControllerService {
   // able to waste user funds. They would only be able to attack the liveness of the Controller through high submission
   // volume, which would be a vulnerability regardless of whether we had auth between the API server and the Controller.
   rpc AuthorizePayment(AuthorizePaymentRequest) returns (AuthorizePaymentResponse) {}
+
+  // RefundPayment reverts a previous payment authorization
+  //
+  // This is intended to be called by API server instances when a blob dispersal fails after payment authorization
```
samlaf, line 30: can we define "fails" here? Is it only if it reaches a specific blob status? Or all that aren't success?

api/hashing/refund_payment_request_hashing.go
samlaf, line 1: this is purely for internal authn/authz between our services right? Can we put these under `api/authn` or something? Find it weird to have all these hashing things that have very different purposes under this package.

core/payments/ondemand/ondemandvalidation/on_demand_payment_validator.go
```diff
@@ -76,3 +76,23 @@ func (pv *OnDemandPaymentValidator) Debit(
 	pv.metrics.IncrementUnexpectedErrors()
 	return err
 }
+
+// RevertDebit reverts a previous debit operation for an on-demand payment
+// This is used when a blob dispersal fails after payment authorization
+func (pv *OnDemandPaymentValidator) RevertDebit(
```
samlaf, line 82: this "OnDemandPaymentValidator" is run on the disperser only right? and not on the validator?
I'm probably just dumb but "Validator" makes me think this is run on the validator itself. Maybe there's a way to remove this confusion, or maybe the confusion is only in my brain?

core/payments/reservation/reservationvalidation/reservation_payment_validator.go
```diff
@@ -105,3 +105,24 @@ func (pv *ReservationPaymentValidator) Debit(
...
+) error {
+	ledger, err := pv.ledgerCache.GetOrCreate(ctx, accountID)
+	if err != nil {
+		return fmt.Errorf("get or create ledger: %w", err)
+	}
+
+	now := pv.timeSource()
+	_, err = ledger.RevertDebit(now, symbolCount)
+	if err != nil {
+		return fmt.Errorf("revert debit: %w", err)
```
samlaf, line 124: could this fail because of a clock getting out of sync? Or is this only running on a single disperser and using monotone clock?

PR #2179, opened 2025-10-16
fix: Enhance eigenda-releaser workflow with source_branch

anupsv:
Added source_branch input validation and checkout step.

Adding option of a branch for release workflow.


anupsv:
> Is the idea,
> 
> 1. create a branch starting with a release
> 2. use git flow UI to run eigenda-releaser
> 
> Screenshot 2025-10-16 at 9 12 59 AM
> But if I can create the release, what uses does the release-bot do, push another release/?

You can't create a release branch at all. This workflow is the only one that can create branches starting with `release/`

PR #2184, opened 2025-10-17
fix(sigmap-EDA-10): Missing length validation in security parameter verification loop

iquidus:
Closes DAINT-784

Addresses EDA-10  from the [Sigma Prime audit](https://linear.app/eigenlabs/issue/DAINT-774/address-audit-findings) report by adding length validation before the for loop to prevent a potential out-of-bounds panic if the blobHeader contains more quorum params than there is available quorums.

Adds tests for the `verifySecurityParams` function.

A security parameter verification loop assumes equal array lengths without validation, potentially causing array bounds
violations if the arrays have mismatched sizes.


ethenotethan (approved):
LGTM but also +1 to @samlaf's comment about this being V1 specific. A bit curious why we recommended our auditors to look at this

iquidus:
> This is V1 code, which we don't maintain anymore other than for critical bugs affecting liveness.... why was this part of the audit report?
> 
> Can you make sure that we do make this check in the V2 codepaths (which would be on the CertVerifier contract I guess..?).

At first glance it appears we have a similar issue in the verification lib used by the CertVerifier contract:

https://github.com/Layr-Labs/eigenda/blob/c2e6e3dff44ceb245b8b3744623005cb402e127b/contracts/src/integrations/cert/libraries/EigenDACertVerificationLib.sol#L208-L216

If length of `quorumStakeTotals.signedStakeForQuorum` or `quorumStakeTotals.totalStakeForQuorum` is less than length of `signedQuorumNumbers`, out of bounds. 

However, it's not possible for this scenario to actually happen. `quorumStakeTotals` is being initialized using the `checkSignatures` func from the BLSSignatureChecker here https://github.com/Layr-Labs/eigenlayer-middleware/blob/dev/src/BLSSignatureChecker.sol#L39, passing `signedQuorumNumbers` in for the `quorumNumbers` argument. 

Length validation is performed within the `checkSignatures` func, followed by both `signedStakeForQuorum` and `totalStakeForQuorum` being declared with a size matching `quorumNumbers`.

Conclusion. An out of bounds here is not possible.

samlaf (approved):

PR #2186, opened 2025-10-17
refactor: create kzgflags package

samlaf:
part of https://github.com/Layr-Labs/eigenda/pull/2182 split.

This isolates the kzg flags definitions into their own package, to make sure when we import them we are only importing the flags.

Goal is to separate kzg into v1 and v2, but keep sharing the flags between them, for backwards compatibility.

### DIFF
--- a/encoding/kzg/kzgconfig.go
+++ b/encoding/kzg/kzgconfig.go
@@ -8,9 +8,9 @@ import (
 	"github.com/urfave/cli"
 )
 
-// KzgConfig holds configuration for KZG prover and verifier.
+// KzgConfig holds configuration for KZG prover and verifier V1.
 // Some of the configurations only apply to the prover or verifier.
-// TODO(samlaf): split into separate Prover and Verifier configs.
+// V2 prover, verifier, and committer each have their own config structs.
 type KzgConfig struct {
 	// SRSOrder is the total size of SRS.
 	// TODO(samlaf): this should always be 2^28. Get rid of this field and replace with hardcoded constant.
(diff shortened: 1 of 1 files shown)


disperser/cmd/apiserver/lib/config.go
```diff
@@ -80,12 +79,7 @@ func NewConfig(ctx *cli.Context) (Config, error) {
 		return Config{}, err
 	}
 
-	kzgCommitterConfig := committer.Config{
-		SRSNumberToLoad:   ctx.GlobalUint64(kzg.SRSLoadingNumberFlagName),
-		G1SRSPath:         ctx.GlobalString(kzg.G1PathFlagName),
-		G2SRSPath:         ctx.GlobalString(kzg.G2PathFlagName),
-		G2TrailingSRSPath: ctx.GlobalString(kzg.G2TrailingPathFlagName),
-	}
+	kzgCommitterConfig := committer.ReadCLIConfig(ctx)
```
samlaf, line 82: I think ultimately we should change these names to be `ConfigFromFlags` or something, but opted to keep consistent with rest of codebase.

encoding/kzg/kzgconfig.go
```diff
@@ -53,3 +57,24 @@ type KzgConfig struct {
 	NumWorker uint64
 	Verbose   bool
 }
+
+func ReadCLIConfig(ctx *cli.Context) KzgConfig {
```
litt3, line 64: Consider making a note that this function doesn't load in any value for `LoadG2Points`.

That field is an odd duck... I know you were just moving this function, so feel free to ignore this comment if you prefer

samlaf: [27f7877](https://github.com/Layr-Labs/eigenda/pull/2186/commits/27f787788e6a79e7becfc6823c7c9bb9903860ec)

pschork (approved):

PR #2189, opened 2025-10-17
refactor(encoding): breakup kzg into v1/v2 separate pkgs

samlaf:
Part of https://github.com/Layr-Labs/eigenda/pull/2182 split.

### DIFF
--- a/encoding/v1/kzg/prover/prover_test.go
+++ b/encoding/v1/kzg/prover/prover_test.go
@@ -36,9 +36,9 @@ func setup() {
 	log.Println("Setting up suite")
 
 	kzgConfig = &kzg.KzgConfig{
-		G1Path:          "../../../resources/srs/g1.point",
-		G2Path:          "../../../resources/srs/g2.point",
-		CacheDir:        "../../../resources/srs/SRSTables",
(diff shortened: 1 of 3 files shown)


pschork (approved):

encoding/v2/kzg/verifier/config.go
```diff
@@ -32,7 +32,7 @@ func ConfigFromProverV2Config(v2Prover *prover.KzgConfig) *Config {
 // The V1 KzgConfig is used all over the place in multiple different structs,
 // making it very hard to update, optimize, change, or remove unused fields.
 // The V2 verifier has its own KzgConfig, which is a very small subset of the V1 KzgConfig.
-func ConfigFromV1KzgConfig(v1 *kzg.KzgConfig) *Config {
+func ConfigFromV1KzgConfig(v1 *kzgv1.KzgConfig) *Config {
```
bxue-l2, line 35: just for clarity, there is no kzgv2.KZGconfig

samlaf: yes and no. There isn't the same config no. But v2 prover/commiter/verifier have their own configs.

PR #2195, opened 2025-10-20
feat: add config retriever

pakim249CAL:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


contracts/src/periphery/cfg-getter/EigenDAConfigRetriever.sol
```diff
@@ -0,0 +1,76 @@
...
+        // Collect the checkpoints after the given activation key
+        ConfigRegistryTypes.Bytes32Checkpoint[] memory result = new ConfigRegistryTypes.Bytes32Checkpoint[](count);
+        uint256 index = 0;
+        for (uint256 i = 0; i < numCheckpoints; i++) {
+            uint256 checkpointActivationKey = configRegistry.getActivationKeyBytes32(nameDigest, i);
+            if (checkpointActivationKey > activationKey) {
+                result[index] = configRegistry.getCheckpointBytes32(nameDigest, i);
+                index++;
+            }
+        }
```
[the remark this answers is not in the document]
pakim249CAL: An increase in gas cost here is acceptable in return for more clarity, since this will only be used off-chain.

contracts/src/core/libraries/v3/address-directory/AddressDirectoryConstants.sol
```diff
@@ -34,6 +33,9 @@ library AddressDirectoryConstants {
     string internal constant CERT_VERIFIER_NAME = "CERT_VERIFIER";
     string internal constant CERT_VERIFIER_ROUTER_NAME = "CERT_VERIFIER_ROUTER";
 
+    /// @dev This is an optional contract that EigenDA may deploy to provide more complex getters for configurations.
+    string internal constant EIGEN_DA_CONFIG_REGISTRY_NAME = "EIGEN_DA_CONFIG_RETRIEVER";
```
samlaf:
```suggestion
    string internal constant EIGEN_DA_CONFIG_RETRIEVER_NAME = "EIGEN_DA_CONFIG_RETRIEVER";
```

contracts/src/periphery/cfg-getter/EigenDAConfigRetriever.sol
```diff
@@ -0,0 +1,76 @@
...
+    {
+        bytes32 nameDigest = ConfigRegistryLib.getNameDigest(name);
+        uint256 numCheckpoints = configRegistry.getNumCheckpointsBytes32(nameDigest);
+
+        // Count how many checkpoints are after the given activation key
+        uint256 count = 0;
+        for (uint256 i = 0; i < numCheckpoints; i++) {
+            if (configRegistry.getActivationKeyBytes32(nameDigest, i) > activationKey) {
+                count++;
+            }
```
samlaf: thought we wanted to walk from the back instead?

contracts/src/periphery/cfg-getter/EigenDAConfigRetriever.sol
```diff
@@ -0,0 +1,76 @@
...
+        }
+
+        // Collect the checkpoints after the given activation key
+        ConfigRegistryTypes.Bytes32Checkpoint[] memory result = new ConfigRegistryTypes.Bytes32Checkpoint[](count);
+        uint256 index = 0;
+        for (uint256 i = 0; i < numCheckpoints; i++) {
+            uint256 checkpointActivationKey = configRegistry.getActivationKeyBytes32(nameDigest, i);
+            if (checkpointActivationKey > activationKey) {
+                result[index] = configRegistry.getCheckpointBytes32(nameDigest, i);
+                index++;
```
samlaf: I guess the goal here was not to minimize gas costs because this is for looping twice over the same configuration. Might be worth documenting that this is only meant to be called offchain.

contracts/src/periphery/cfg-getter/EigenDAConfigRetriever.sol
pakim249CAL: https://github.com/Layr-Labs/eigenda/pull/2195/commits/9c9751692b077b3c1591cf593cb14e13f698690f

contracts/src/periphery/cfg-getter/EigenDAConfigRetriever.sol
```diff
@@ -0,0 +1,79 @@
...
+        // There are 3 cases to handle:
+        // 1. If no checkpoints have activation keys greater than the provided activation key, we return an empty array.
+        // 2. If all checkpoints have activation keys greater than the provided activation key, we return the entire array.
+        // 3. If some checkpoints have activation keys greater than the provided activation key, we return the relevant subset.
+
+        uint256 startIndex = numCheckpoints; // Default to numCheckpoints (case 1)
+        for (uint256 i = 0; i < numCheckpoints; i++) {
+            uint256 checkpointActivationKey = configRegistry.getActivationKeyBytes32(nameDigest, numCheckpoints - 1 - i);
+            if (checkpointActivationKey <= activationKey) {
+                startIndex = numCheckpoints - i; // Found the first checkpoint with activation key <= provided key
```
samlaf, line 35:
```suggestion
        for (uint256 i = numCheckpoints-1; i >= 0; i--) {
            uint256 checkpointActivationKey = configRegistry.getActivationKeyBytes32(nameDigest, i);
            if (checkpointActivationKey <= activationKey) {
                startIndex = i; // Found the first checkpoint with activation key <= provided key
```

contracts/src/periphery/cfg-getter/EigenDAConfigRetriever.sol
```diff
@@ -0,0 +1,79 @@
...
+                break;
+            }
+        }
+        // Collect the checkpoints from startIndex to the end
+        uint256 resultCount = numCheckpoints - startIndex;
+        ConfigRegistryTypes.Bytes32Checkpoint[] memory results =
+            new ConfigRegistryTypes.Bytes32Checkpoint[](resultCount);
+        for (uint256 i = 0; i < resultCount; i++) {
+            results[i] = configRegistry.getCheckpointBytes32(nameDigest, startIndex + i);
+        }
```
samlaf, line 45:
```suggestion
        for (uint256 i = startIndex; i < numCheckpoints; i++) {
            results[i] = configRegistry.getCheckpointBytes32(nameDigest, i);
        }
```

samlaf (commented):
Can you please simplify the for loops and indexing? its actually very hard to read in its current form, which means its also bug prone. We also don't have tests so hard to know whether there isnt an off-by-1 error in its current form.

contracts/src/periphery/cfg-getter/EigenDAConfigRetriever.sol
pakim249CAL: Personally, I do not consider this a simplification. I much prefer a standard for loop, starting at 0 and only incrementing i when it is possible to write it that way, because this is simply the standard for loop that programmers are used to.

This implementation you are proposing has a bug. It is because the decrement of i only happens after the loop body.

These kinds of issues are why I would not suggest modifying the for loop conditions.

samlaf: We chatted offline, Patrick mentioned that this is the idiomatic way to do things, because going backwards with a for loop iterating on a uint will underflow.
One other way to solve this is just to iterate using an `int`, whcih is what golang does for example. Not sure which is most solidity-idiomatic though so I'm fine with either.

0xClandestine: Not sure i'm following this explanation, we regularly use backwards loops in core. Can you explain further?

samlaf: The last loop (when i=0) will after executing decrement i with `i--`. This will cause an underflow.

contracts/src/periphery/cfg-getter/EigenDAConfigRetriever.sol
```diff
@@ -0,0 +1,79 @@
...
+/// @notice A stateless contract that defines getter functions using the underlying configuration registry's primitive API.
+contract EigenDAConfigRetriever {
+    IEigenDAConfigRegistry public immutable configRegistry;
+
+    constructor(address configRegistryAddress) {
+        configRegistry = IEigenDAConfigRegistry(configRegistryAddress);
+    }
+
+    /// @notice Retrieves all bytes32 configuration checkpoints for a given name that have an activation key greater than the provided activation key.
+    function getAllBytes32ConfigsAfterKey(string memory name, uint256 activationKey)
```
litt3, line 18: This is not what we actually need.

We need to get the currently active value, plus all future values. This is only getting all future values.

In english, what we're trying to accomplish is:

"get the currently active configuration value, and also tell me all upcoming updates so I can plan ahead and make the config switch at the right moment"

samlaf: Are you just talking about the comment? I think the code is actually doing the right thing though?
image

litt3: > I think the code is actually doing the right thing though?

Yeah, appears so 🤷 But also the function names are misleading

In any case, this code should probably be considered a draft for now. Once it's been reconsidered from first principles, and the docs + function names have been updated to be actually correct, then we can move forward with it.

For now, I'm converting it to `draft` mode

PR #2196, opened 2025-10-21
refactor(controller): Refactor payment authorization configuration

dmanc:

Adds verifiable config pattern to payment authorization handler and its dependencies.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/payments/ondemand/ondemandvalidation/on_demand_ledger_cache_config.go
+++ b/core/payments/ondemand/ondemandvalidation/on_demand_ledger_cache_config.go
@@ -2,6 +2,7 @@ package ondemandvalidation
 
 import (
 	"errors"
+	"fmt"
 	"time"
 )
 
@@ -45,7 +46,7 @@ func NewOnDemandLedgerCacheConfig(
 	}
 
 	if err := config.Verify(); err != nil {
-		return OnDemandLedgerCacheConfig{}, err
+		return OnDemandLedgerCacheConfig{}, fmt.Errorf("failed to verify on-demand ledger cache config: %w", err)
 	}
 
 	return config, nil
--- a/core/payments/reservation/reservationvalidation/reservation_ledger_cache_config.go
+++ b/core/payments/reservation/reservationvalidation/reservation_ledger_cache_config.go
@@ -2,6 +2,7 @@ package reservationvalidation
 
 import (
 	"errors"
+	"fmt"
 	"time"
 
 	"github.com/Layr-Labs/eigenda/common/ratelimit"
@@ -60,7 +61,7 @@ func NewReservationLedgerCacheConfig(
 	}
 
 	if err := config.Verify(); err != nil {
-		return ReservationLedgerCacheConfig{}, err
+		return ReservationLedgerCacheConfig{}, fmt.Errorf("failed to verify reservation ledger cache config: %w", err)
 	}
 
 	return config, nil


core/payments/ondemand/ondemandvalidation/on_demand_ledger_cache_config.go
```diff
@@ -15,27 +15,38 @@ type OnDemandLedgerCacheConfig struct {
 	UpdateInterval time.Duration
 }
 
+// Verify validates the OnDemandLedgerCacheConfig
+func (c *OnDemandLedgerCacheConfig) Verify() error {
+	if c.MaxLedgers <= 0 {
+		return errors.New("max ledgers must be > 0")
+	}
```
samlaf, line 23: nit: any reason not to make it a uint64 instead?

disperser/controller/payment_authorization.go
```diff
@@ -0,0 +1,170 @@
...
+
+// DefaultPaymentAuthorizationConfig returns a new PaymentAuthorizationConfig with default values
+func DefaultPaymentAuthorizationConfig() *PaymentAuthorizationConfig {
+	onDemandConfig, _ := ondemandvalidation.NewOnDemandLedgerCacheConfig(
+		1024,
+		"",
+		30*time.Second,
+	)
+
+	reservationConfig, _ := reservationvalidation.NewReservationLedgerCacheConfig(
```
samlaf: why discard error?

disperser/controller/payment_authorization.go
```diff
@@ -0,0 +1,170 @@
...
+	}
+	if err := c.ReservationConfig.Verify(); err != nil {
+		return fmt.Errorf("reservation config: %w", err)
+	}
+	return nil
+}
+
+// DefaultPaymentAuthorizationConfig returns a new PaymentAuthorizationConfig with default values
+func DefaultPaymentAuthorizationConfig() *PaymentAuthorizationConfig {
+	onDemandConfig, _ := ondemandvalidation.NewOnDemandLedgerCacheConfig(
```
samlaf: why discard error?

samlaf: better to panic imo, cause the code assumptions could change.
Otherwise just build it explicitly, don't call the constructor that calls verify

core/payments/ondemand/ondemandvalidation/on_demand_ledger_cache_config.go
dmanc: We would end up casting it into an int anyway since the LRU library takes an int. If we had uint64 we would still get a weird error if we used a huge uint64.
```
	ledgerCache.cache, err = lru.NewWithEvict(
		config.MaxLedgers,
		func(accountAddress gethcommon.Address, _ *ondemand.OnDemandLedger) {
			ledgerCache.metrics.IncrementEvictions()
			logger.Infof("evicted account %s from LRU on-demand ledger cache", accountAddress.Hex())
		},
	)
```

Actually I wonder why there's no maximum ondemand ledgers, I noticed that there is a maximum for reservation ledger.
```
const (
	// maxReservationLRUCacheSize is the maximum number of reservation ledgers that can be stored in the cache.
	// Set to 2^16 = 65,536 entries.
	//
	// To do some napkin math: each cache entry is <500 bytes in size, so 65k cache entries would have a memory
	// footprint <33MiB. This isn't a catastrophic amount of memory, and 65k active reservation users is absurdly high.
	maxReservationLRUCacheSize = 65536
)
```

disperser/controller/payment_authorization.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2196/commits/b505830f2bb83c4361155647d01400310dfc33cc

disperser/controller/payment_authorization.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2196/commits/b505830f2bb83c4361155647d01400310dfc33cc

core/payments/ondemand/ondemandvalidation/on_demand_ledger_cache_config.go
```diff
@@ -15,27 +15,38 @@ type OnDemandLedgerCacheConfig struct {
...
+	config := OnDemandLedgerCacheConfig{
+		MaxLedgers:        maxLedgers,
+		OnDemandTableName: onDemandTableName,
+		UpdateInterval:    updateInterval,
 	}
 
-	if updateInterval <= 0 {
-		return OnDemandLedgerCacheConfig{}, errors.New("update interval must be > 0")
+	if err := config.Verify(); err != nil {
+		return OnDemandLedgerCacheConfig{}, err
```
cody-littley: Please wrap the error, it can make it a lot harder to trace the origin of an error if we return errors without wrapping them.

core/payments/reservation/reservationvalidation/reservation_ledger_cache_config.go
```diff
@@ -20,37 +20,48 @@ type ReservationLedgerCacheConfig struct {
...
+	config := ReservationLedgerCacheConfig{
 		MaxLedgers:           maxLedgers,
 		BucketCapacityPeriod: bucketCapacityPeriod,
 		OverfillBehavior:     overfillBehavior,
 		UpdateInterval:       updateInterval,
-	}, nil
+	}
+
+	if err := config.Verify(); err != nil {
+		return ReservationLedgerCacheConfig{}, err
```
cody-littley: wrap error

core/payments/ondemand/ondemandvalidation/on_demand_ledger_cache_config.go
samlaf: guess we'll have to wait for @litt3 to come back.

core/payments/reservation/reservationvalidation/reservation_ledger_cache_config.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2196/commits/2ebf4c8cea08d0acdd539cdc891bb724f8971fb8

core/payments/ondemand/ondemandvalidation/on_demand_ledger_cache_config.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2196/commits/2ebf4c8cea08d0acdd539cdc891bb724f8971fb8

core/payments/ondemand/ondemandvalidation/on_demand_ledger_cache_config.go
litt3: @dmanc @samlaf Just saw this

> Actually I wonder why there's no maximum ondemand ledgers, I noticed that there is a maximum for reservation ledger.

The reason is that the reservation cache will automatically resize if premature evictions are detected. We configure a sane limit of `maxReservationLRUCacheSize` to prevent auto-resizing from going crazy, just in case.

Since on-demand doesn't have automatic cache resizing (since there's no such thing as a premature eviction from the on-demand cache), we don't need to have a separate "max value" for on-demand: the "max value" is just whatever you configure the cache with in the first place. 

In hindsight, I could have found better names for these concepts, to make this more clear.

PR #2198, opened 2025-10-21
refactor(prover-v2): pull encoder out of parametrizedProver

samlaf:
There are a few reasons for doing this:
1. First is that bowen found an optimization where encoding and proving will now require using different parameters, so it doesn't make sense for them to both be parametrized the same way. 
2. Second ParametrizedProver just held too many things, and is hard to follow the dependency of things. Namely, this made me realize that the encoder held by the parametrizedProver was a normal encoder (which itself holds a map of parametrizedEncoder), so it feels very weird to have it held inside the parametrizedProver. 
3. Third, I'd eventually like to completely pull the encoder out of the prover, and create a small external wrapper struct to combine them to generate a frame (don't think this should be the functionality of the prover, which should focus on generating proofs only).

Side note, I realized that we're being very inconsistent with which functions return indices vs which don't. The parametrizedProver.GetFrames() was returning indices, but the prover.GetFrames() was not. When I moved the encoding process to GetFrames, I made it also return indices just to keep the tests that were relying on such behavior intact. But ultimately I think indices should never be returned, and we should instead just document the way the proofs/chunks are ordered, similarly to how FFT encodes whether inputs/outputs are returned in normal or butterflied order. WDYT @bxue-l2 ?

### DIFF
--- a/encoding/v2/kzg/prover/prover.go
+++ b/encoding/v2/kzg/prover/prover.go
@@ -134,10 +134,12 @@ func (e *Prover) GetFrames(data []byte, params encoding.EncodingParams) ([]*enco
 	)
 
 	frames := make([]*encoding.Frame, len(proofs))
-	for ind, frame := range proofs {
-		frames[ind] = &encoding.Frame{
-			Coeffs: encodeResult.chunks[ind],
-			Proof:  frame,
+	for i, index := range encodeResult.indices {
+		frames[i] = &encoding.Frame{
+			Coeffs: encodeResult.chunks[i],
+			// Coeffs are returned according to indices order, but proofs are not
+			// TODO(samlaf): we should be consistent about this.
+			Proof: proofs[index],
 		}
 	}
 	return frames, encodeResult.indices, nil


encoding/v2/kzg/prover/prover.go
```diff
@@ -79,31 +80,67 @@ func NewProver(logger logging.Logger, kzgConfig *KzgConfig, encoderConfig *encod
...
 	prover, err := e.GetKzgProver(params)
 	if err != nil {
-		return nil, fmt.Errorf("get kzg prover: %w", err)
+		return nil, nil, fmt.Errorf("get kzg prover: %w", err)
 	}
 
-	kzgFrames, _, err := prover.GetFrames(symbols)
-	if err != nil {
-		return nil, fmt.Errorf("get frames: %w", err)
+	type encodeCResult struct {
```
dmanc: what is `encodeC` Result mean?

encoding/v2/kzg/prover/prover.go
```diff
@@ -79,31 +80,67 @@ func NewProver(logger logging.Logger, kzgConfig *KzgConfig, encoderConfig *encod
...
+			err:      err,
 		}
+	}()
+
+	getProofsStart := time.Now()
+	proofs, err := prover.GetProofs(symbols)
+	getProofsDuration := time.Since(getProofsStart)
+
+	// Wait for both chunks and frames to have finished generating
+	encodeRes := <-encodeChan
```
dmanc: nit: Prefer it to be consistent to above variable `encodeCResult` so use `encodeResult`

samlaf: [9003873](https://github.com/Layr-Labs/eigenda/pull/2198/commits/900387381db1d08b4d02ccbb6bd2d5d6403a6c03)

encoding/v2/kzg/prover/prover.go
samlaf: C is typically used as a `Chan` suffix. like intC, or msgC, or whatever. But I renamed to encodeChanResult for consistency everywhere, along with your other comment: [9003873](https://github.com/Layr-Labs/eigenda/pull/2198/commits/900387381db1d08b4d02ccbb6bd2d5d6403a6c03)

encoding/v2/kzg/verifier/verifier_test.go
```diff
@@ -82,7 +80,7 @@ func TestUniversalVerify(t *testing.T) {
 
 			sample := encoding.Sample{
 				Commitment:      (*encoding.G1Commitment)(commit),
-				Chunk:           &f,
+				Chunk:           f,
```
bxue-l2, line 83: nit pick, prover.GetFrame returns []*encoding.Frame
parametriedProver.GetFrame returns []encoding.Frame

encoding/v2/kzg/prover/parametrized_prover.go
```diff
@@ -2,119 +2,39 @@ package prover
...
-}
-
-type proofsResult struct {
-	Proofs   []bn254.G1Affine
-	Duration time.Duration
-	Err      error
-}
-
-func (g *ParametrizedProver) GetFrames(inputFr []fr.Element) ([]encoding.Frame, []uint32, error) {
+func (g *ParametrizedProver) GetProofs(inputFr []fr.Element) ([]encoding.Proof, error) {
```
bxue-l2, line 23: Not sure how easy it is, can we remove the `encodingParams` from the struct (Maybe you want to remove it altogether), instead only pass the encoding params in

encoding/v2/kzg/verifier/verifier_test.go
bxue-l2: but no need to fix

encoding/v2/kzg/prover/parametrized_prover.go
bxue-l2: git comment does not show. I mean here couple of lines above

```
type ParametrizedProver struct {
	srsNumberToLoad uint64

	encodingParams encoding.EncodingParams

	computeMultiproofNumWorker uint64
	kzgMultiProofBackend       backend.KzgMultiProofsBackendV2
}
```

encoding/v2/kzg/verifier/verifier_test.go
samlaf: oh still lol? ok will fix in future PR

encoding/v2/kzg/prover/parametrized_prover.go
samlaf: ah good call! I actually do want to move in this direction.
I was actually thinking of just completely getting rid of the parametrizedProver and in Prover instead of having
```
ParametrizedProvers map[encoding.EncodingParams]*ParametrizedProver
```
replace it with
```
ParametrizedBackends map[encoding.EncodingParams]*backend.KzgMultiProofsBackendV2
```

but I need to play with the code more to see if it makes sense, and I also didn't want to make this PR too large.
I can do in a followup PR though if that would help you with your prover refactor you're trying to do.

bxue-l2: in the optimized code the map key would be provingParams

PR #2199, opened 2025-10-21
chore(dataapi): fixes flaky TestFetchOperatorDispersalFeed test

pschork:
TestFetchOperatorDispersalFeed test is flaky due to the test expecting HTTP 200 responses but getting 500s when localstack DynamoDB connections drop temporarily.

```
[GIN] 2025/10/20 - 17:34:08 | 500 |   24.909607ms |       192.0.2.1 | GET      "/v2/operators/.../dispersals?limit=0&direction=backward"
Error #01: failed to fetch dispersals from blob metadata store: ... read tcp 127.0.0.1:44208->127.0.0.1:4574: use of closed network connection
```
Adds retry logic to the executeRequest test helper that retries specifically on the "use of closed network connection" error pattern. Other errors continue to fail immediately.

Unblocks yak shaving PR's that are failing merge queue

### DIFF
--- a/disperser/dataapi/v2/server_v2_test.go
+++ b/disperser/dataapi/v2/server_v2_test.go
@@ -275,20 +275,20 @@ const (
 
 func executeRequest(t *testing.T, router *gin.Engine, method, url string) *httptest.ResponseRecorder {
 	t.Helper()
-	
+
 	var lastResponse *httptest.ResponseRecorder
-	
+
 	for attempt := 0; attempt < maxRetries; attempt++ {
 		w := httptest.NewRecorder()
 		req := httptest.NewRequest(method, url, nil)
 		router.ServeHTTP(w, req)
-		
+
 		if w.Code == http.StatusOK {
 			return w
 		}
-		
+
 		lastResponse = w
-		
+
 		// Retry only on specific network-related 500 errors from localstack
 		if w.Code == http.StatusInternalServerError && isLocalstackNetworkError(w) {
 			if attempt < maxRetries-1 {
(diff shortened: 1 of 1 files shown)


disperser/dataapi/v2/server_v2_test.go
```diff
@@ -267,13 +268,48 @@ func setUpRouter() *gin.Engine {
...
+	}
+
+	require.Equal(t, http.StatusOK, lastResponse.Code,
+		"Request failed after %d attempts. Response: %s", maxRetries, lastResponse.Body.String())
+	return lastResponse
+}
+
+func isLocalstackNetworkError(w *httptest.ResponseRecorder) bool {
+	body := w.Body.String()
+	return strings.Contains(body, "use of closed network connection")
```
samlaf, line 312: is this error coming from the http client? or from the localstack server? are we sure its still going to be around if we bump our localstack version? Not that I have any better idea, but maybe at least good to document this

samlaf (approved):
Nice ty. Do you know why "localstack DynamoDB connections drop temporarily"? Guessing this is purely a github infra related issue and never happens locally? Very curious how github runners do networking; I've always assumed job runs were all running on a single box

disperser/dataapi/v2/server_v2_test.go
pschork: Excellent point! You're absolutely right to question the longevity of this error string.

The error "use of closed network connection" is actually from Go's standard library, specifically from the net package when the underlying TCP connection gets closed. It's not localstack-specific.

PR #2203, opened 2025-10-21
chore: commit encoding benchmarks for ec2 g6.xlarge (CPU)

samlaf:
These are CPU results (hence why they are very slow).

Also added a makefile `download_srs_tables` target to download the srs tables, which speeds up the first run of the eigenda benchmarks.


samlaf:
> lgtm, should we add benchmark for multiproof size 2^24 though?

Will do maybe as a future PR. We'd have to fix CI to download the SRS tables, which with the current setup would mean it would have to download 510MB of data everytime CI runs... unless we cache is somehow. Do you know how to set that up?

PR #2204, opened 2025-10-21
feat(proxy): Add tracing to proxy starting with dispersal client and payload disperser

dmanc:

Adds open telemetry based tracing to the V2 dispersal client / payload disperser. Enabled tracing on the proxy so that we can start getting insights into client side performance. 

Disabled by default.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/verification/block_number_monitor.go
+++ b/api/clients/v2/verification/block_number_monitor.go
@@ -14,7 +14,8 @@ import (
 	"go.opentelemetry.io/otel/trace"
 )
 
-var blockNumberMonitorTracer = otel.Tracer("github.com/Layr-Labs/eigenda/api/clients/v2/verification/block_number_monitor")
+var blockNumberMonitorTracer = otel.Tracer(
+	"github.com/Layr-Labs/eigenda/api/clients/v2/verification/block_number_monitor")
 
 // BlockNumberMonitor is a utility for waiting for a certain ethereum block number
 //


api/proxy/telemetry/README.md
```diff
@@ -0,0 +1,52 @@
...
+| `--otel.trace.sample-rate` | `EIGENDA_PROXY_OTEL_TRACE_SAMPLE_RATE` | `1.0` | Trace sampling rate (0.0 to 1.0) |
+
+## Quick Start with Docker Compose
+
+The easiest way to get started is using the included docker-compose setup:
+
+### 1. Start all services (Proxy + Tempo + Grafana)
+
+```bash
+cd api/proxy
```
litt3, line 24: Since this doc essentially just describes an additional feature that can be used during normal proxy operation, maybe it should be added to the main proxy readme? Currently that's where we describe all the different ways someone could use the proxy, as opposed to documenting features deeper down in different packages

api/proxy/telemetry/config.go
```diff
@@ -0,0 +1,103 @@
...
+	OTelExporterInsecureFlagName = "otel.exporter.otlp.insecure"
+	OTelTraceSampleRateFlagName  = "otel.trace.sample-rate"
+)
+
+// Config holds the configuration for OpenTelemetry tracing
+type Config struct {
+	// Enabled determines if OpenTelemetry tracing is enabled
+	Enabled bool
+	// ServiceName is the name of the service for tracing
+	ServiceName string
```
litt3, line 23: Does this need to be configurable? Would it ever be anything other than `eigenda-proxy`?

api/proxy/telemetry/config.go
```diff
@@ -0,0 +1,103 @@
...
+	// ExporterEndpoint is the OTLP exporter endpoint (e.g., "localhost:4317" for gRPC)
+	ExporterEndpoint string
+	// ExporterInsecure determines if the exporter should use an insecure connection
+	ExporterInsecure bool
```
litt3, line 27: Does this need to be configurable? Could we just always use `secure`, and avoid the additional configuration field?

If we do need this field, is `true` actually the right value? Seems that unless there is a large cost to `secure`, that should be the default option

api/proxy/telemetry/tracer.go
```diff
@@ -0,0 +1,73 @@
...
+	}
+
+	// Create tracer provider with sampling configuration
+	var sampler sdktrace.Sampler
+	if cfg.TraceSampleRate >= 1.0 {
+		sampler = sdktrace.AlwaysSample()
+	} else if cfg.TraceSampleRate <= 0.0 {
+		sampler = sdktrace.NeverSample()
+	} else {
+		sampler = sdktrace.TraceIDRatioBased(cfg.TraceSampleRate)
```
litt3: Looking at the doc for `TraceIDRatioBased`, it looks like it already handles the `TraceSampleRate >= 1.0` and `TraceSampleRate <= 0.0` cases

api/proxy/cmd/server/entrypoint.go
```diff
@@ -43,6 +44,34 @@ func StartProxyService(cliCtx *cli.Context) error {
 	if err := cfg.Check(); err != nil {
 		return err
 	}
+
+	// Initialize OpenTelemetry tracing
+	telemetryCfg := telemetry.ReadConfig(cliCtx)
```
litt3: The pre-existing pattern has been to nest all configs within the "app config". Is there a specific reason to change strategies with the telemetry config?

api/proxy/cmd/server/entrypoint.go
```diff
@@ -43,6 +44,34 @@ func StartProxyService(cliCtx *cli.Context) error {
...
+
+	ctx, ctxCancel := context.WithCancel(cliCtx.Context)
+	defer ctxCancel()
+
+	shutdownTracer, err := telemetry.InitTracer(ctx, telemetryCfg)
+	if err != nil {
+		return fmt.Errorf("initialize tracer: %w", err)
+	}
+	defer func() {
+		if err := shutdownTracer(context.Background()); err != nil {
```
litt3: Any reason to build a new context object here?

api/clients/v2/payloaddispersal/payload_disperser.go
```diff
@@ -89,6 +95,14 @@ func (pd *PayloadDisperser) SendPayload(
 	// payload is the raw data to be stored on eigenDA
 	payload coretypes.Payload,
 ) (coretypes.EigenDACert, error) {
+	ctx, span := tracer.Start(ctx, "PayloadDisperser.SendPayload",
```
litt3, line 98: Why is tracing optional for the `DisperserClient`, but always enabled for the `PayloadDisperser`?

api/clients/v2/payloaddispersal/payload_disperser.go
```diff
@@ -19,8 +19,14 @@ import (
 	corev2 "github.com/Layr-Labs/eigenda/core/v2"
 	"github.com/Layr-Labs/eigensdk-go/logging"
 	"github.com/prometheus/client_golang/prometheus"
+	"go.opentelemetry.io/otel"
+	"go.opentelemetry.io/otel/attribute"
+	"go.opentelemetry.io/otel/codes"
+	"go.opentelemetry.io/otel/trace"
 )
 
+var tracer = otel.Tracer("github.com/Layr-Labs/eigenda/api/clients/v2/payloaddispersal")
```
litt3, line 28: Does this need to be a global variable?

api/clients/v2/payloaddispersal/payload_disperser.go
litt3, line 1: Meta question about this PR:

Is there a specific reason we need this right now? I'm not asserting that we don't need it, just trying to understand motivations

api/clients/v2/payloaddispersal/payload_disperser.go
```diff
@@ -283,6 +367,12 @@ func (pd *PayloadDisperser) pollBlobStatusUntilSigned(
 	initialStatus dispgrpc.BlobStatus,
 	probe *common.SequenceProbe,
 ) (*dispgrpc.BlobStatusReply, error) {
+	ctx, span := tracer.Start(ctx, "PayloadDisperser.pollBlobStatusUntilSigned",
```
litt3, line 370: You're creating new spans for each of these utility methods. Does that actually make sense? Why not just pass the span through as a parameter?

api/clients/v2/payloaddispersal/payload_disperser.go
```diff
@@ -89,6 +95,14 @@ func (pd *PayloadDisperser) SendPayload(
 	// payload is the raw data to be stored on eigenDA
 	payload coretypes.Payload,
 ) (coretypes.EigenDACert, error) {
+	ctx, span := tracer.Start(ctx, "PayloadDisperser.SendPayload",
+		trace.WithAttributes(
```
litt3, line 99: I think it would make sense to encapsulate the tracing logic for the `PayloadDisperser`, to avoid cluttering the business logic with it.

api/clients/v2/disperser_client.go
```diff
@@ -163,18 +181,34 @@ func (c *DisperserClient) DisperseBlob(
 	// if this is nil, that indicates we will use the legacy payment system to create the paymentMetadata
 	paymentMetadata *core.PaymentMetadata,
 ) (*corev2.BlobHeader, *disperser_rpc.DisperseBlobReply, error) {
+	ctx, span := tracer.Start(ctx, "DisperserClient.DisperseBlob",
```
litt3, line 184: Probably also makes sense to encapsulate tracing logic for the `DisperserClient`

api/clients/v2/verification/cert_verifier.go
```diff
@@ -57,6 +61,9 @@ func (cv *CertVerifier) CheckDACert(
 	ctx context.Context,
 	cert coretypes.EigenDACert,
 ) error {
+	ctx, span := tracer.Start(ctx, "CertVerifier.CheckDACert")
```
litt3, line 65: Looks like the tracing in this struct doesn't add any metadata, e.g. what cert is being verified. With that in mind, does the tracing really add much value?

api/proxy/telemetry/config.go
dmanc: This is kind of hinting that this package should be part of common so that other services could also support tracing.

api/clients/v2/payloaddispersal/payload_disperser.go
dmanc: There's a [long thread in slack](https://eigenlabs.slack.com/archives/C07D9JLDHR8/p1761005937841579) but to summarize: We don't have great visibility into the performance of an individual blob and tracing will give us that. 

Once all components have tracing enabled we can get a picture like this for the entiire lifecycle so we can pin point areas of the code that are slow and why.
image

api/proxy/telemetry/tracer.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2204/commits/c574f87f8e1d4ba4a8578714a4b24b64e5fd1d4d

api/clients/v2/payloaddispersal/payload_disperser.go
dmanc: This is the opentelemetry suggested pattern https://opentelemetry.io/docs/languages/go/instrumentation/#create-nested-spans

dmanc: As for whether or not we should be creating a span for each utility method it really depends if it's worth tracking the timing / attributes associated with it and it's something we can adjust depending on what we want to observe. Here is an example span:
image

api/proxy/cmd/server/entrypoint.go
dmanc: No specific reason I just didn't notice https://github.com/Layr-Labs/eigenda/pull/2204/commits/b7ab7cb8ec661b0c3cefc32d4fb4b482984ee4c3

api/proxy/cmd/server/entrypoint.go
dmanc: I don't think so https://github.com/Layr-Labs/eigenda/pull/2204/commits/b7ab7cb8ec661b0c3cefc32d4fb4b482984ee4c3

PR #2207, opened 2025-10-22
feat: grpc logging from relays

ian-shim:

Adding grpc logging on relay servers similar to https://github.com/Layr-Labs/eigenda/pull/1998

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


relay/server.go
litt3, line 1: meta PR comment:

It difficult to review when there are logic changes _and_ the code is moved to a different file in the same PR. It breaks the diff, so it looks like all the moved code is totally new and you can't see what specifically has changed.

I think it would probably have been better to do the move and the logic changes in separate PRs. Moving stuff is always gonna result in an ugly diff, but at least if there are no logic changes included, it's easy to have claude help verify that there aren't any substantive changes

relay/get_blob.go
```diff
@@ -0,0 +1,93 @@
...
+		return nil, status.Newf(codes.ResourceExhausted, "bandwidth limit exceeded: %v", err)
+	}
+
+	data, err := s.blobProvider.GetBlob(ctx, key)
+	if err != nil {
+		if strings.Contains(err.Error(), blobstore.ErrBlobNotFound.Error()) {
+			return nil, status.Newf(codes.NotFound, "blob %s not found", key.Hex())
+		} else {
+			s.logger.Errorf("error fetching blob %s: %v", key.Hex(), err)
+			return nil, status.Newf(codes.NotFound,
```
litt3, line 80: What's the reason for changing from `Internal` to `NotFound`?

cody-littley: `Internal` errors trigger alerts in grafana, and should generally be used for unexpected failures. `NotFound` errors do not trigger grafana alerts, and just mean "you asked for data that we don't have, sorry".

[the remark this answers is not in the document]
cody-littley: Perhaps we should move all logging to the layer above? I think we might actually end up logging this error multiple times with the code as it currently is.

relay/get_blob.go
```diff
@@ -0,0 +1,93 @@
...
+	}
+
+	s.metrics.ReportBlobBandwidthUsage(len(data))
+	s.metrics.ReportBlobDataLatency(time.Since(finishedFetchingMetadata))
+	s.metrics.ReportBlobLatency(time.Since(start))
+
+	reply := &pb.GetBlobReply{
+		Blob: data,
+	}
+	return reply, status.New(codes.OK, "")
```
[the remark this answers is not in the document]
cody-littley: Yeah, feels odd to return a non-nil error in the success case. People who call this function are going to accidentally use `if err != nil` blocks, and will be surprised by the behavior.

relay/get_chunks.go
```diff
@@ -0,0 +1,335 @@
...
+			"too many chunk requests provided, max is %d", s.config.MaxKeysPerGetChunksRequest)
+	}
+
+	for _, chunkRequest := range request.GetChunkRequests() {
+		if chunkRequest.GetByIndex() == nil && chunkRequest.GetByRange() == nil {
+			return status.New(codes.InvalidArgument, "chunk request must be either by index or by range")
+		}
+	}
+
+	return status.New(codes.OK, "")
```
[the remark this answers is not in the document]
cody-littley: same comment as above

relay/get_chunks.go
```diff
@@ -0,0 +1,335 @@
...
+}
+
+// GetChunks retrieves chunks from blobs stored by the relay.
+func (s *Server) GetChunks(ctx context.Context, request *pb.GetChunksRequest) (*pb.GetChunksReply, error) {
+	reply, st := s.getChunks(ctx, request)
+	api.LogResponseStatus(s.logger, st)
```
cody-littley, line 43: Similar to comment in `GetBlobs()`, we should do logging here at the upper level, and delete the log statements for errors in the lower layers. Currently we log the same error multiple times.

relay/get_blob.go
litt3: Looking at the code, the block _does_ seem like an unexpected error. The first case `strings.Contains(err.Error(), blobstore.ErrBlobNotFound.Error())` sounds like something we would expect, and the second case seems like something that is unexpected

Are you implying that this is the error case which is triggering the nuisance alerts?

PR #2208, opened 2025-10-22
chore: Add precommit hook for make lint and make fmt-check

dmanc:

Adds a precommit hook that runs `make lint` and `make fmt-check`. I always forget and end up making a commit to fix the linter -> wait another 10 minutes for test to pass.

Try it using `mise run install-hooks`.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/scripts/hooks/pre-commit
+++ b/scripts/hooks/pre-commit
@@ -21,16 +21,22 @@ REPO_ROOT=$(git rev-parse --show-toplevel)
 # Change to the repository root
 cd "$REPO_ROOT"
 
-# Ensure mise tools are available
-if ! command -v mise &> /dev/null; then
-    echo -e "${RED}Error: mise is not installed or not in PATH${NC}"
-    echo "Please install mise to ensure all development tools are available."
+# Check that required tools are available
+if ! command -v golangci-lint &> /dev/null; then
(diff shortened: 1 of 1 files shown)


pschork (dismissed):

scripts/hooks/pre-commit
```diff
@@ -0,0 +1,55 @@
...
+# Ensure mise tools are available
+if ! command -v mise &> /dev/null; then
+    echo -e "${RED}Error: mise is not installed or not in PATH${NC}"
+    echo "Please install mise to ensure all development tools are available."
+    echo "See: https://mise.jdx.dev/"
+    exit 1
+fi
+
+# Activate mise environment to ensure tools like golangci-lint are available
+eval "$(mise env -s bash)"
```
samlaf: not a fan of forcing export overrides like this on people's envs. Also isn't mise supposed to manage this for you? If mise is misconfigured, then the person should just go and fix their env.

samlaf (dismissed):

scripts/hooks/pre-commit
dmanc: https://github.com/Layr-Labs/eigenda/pull/2208/commits/580bb2764dbaa365a9ee8a980cab37a30b08db00

samlaf (approved):

PR #2210, opened 2025-10-22
feat: new and improved ejector

cody-littley:

I had to recreate [this](https://github.com/Layr-Labs/eigenda/pull/2075) PR due to a merge gone wrong. Massive changes in master caused problems with the merge process.


ejector/ejector_config.go
```diff
@@ -69,21 +70,31 @@ type EjectorConfig struct {
 
 	// The period at which to periodically attempt to finalize ejections that have been started.
 	EjectionFinalizationPeriod time.Duration
-}
 
-var _ config.VerifiableConfig = (*EjectorSecretConfig)(nil)
+	// The number of blocks to wait before using a reference block number for quorum. That is to say, do not always
```
litt3:
```suggestion
	// The number of blocks to wait before using a reference block number. That is to say, do not always
```

Feel like I raised this elsewhere already, but not sure where? I find the doc confusing as-is: what role does quorum play here?

ejector/ejector_config.go
```diff
@@ -116,6 +127,28 @@ func (e *RootEjectorConfig) Verify() error {
...
 
+var _ config.VerifiableConfig = (*EjectorSecretConfig)(nil)
+
+// Configuration for secrets used by the ejector.
+type EjectorSecretConfig struct {
+	// The Ethereum RPC URL(s) to use for connecting to the blockchain.
+	EthRpcUrls []string `docs:"required"`
+
+	// The private key to use for signing ejection transactions, in hex.
+	// Do not include the '0x' prefix.
```
litt3, line 137: Does `0x` actually break stuff? Normally hex parsing utils are `0x` tolerant

ejector/ejector_config.go
```diff
@@ -131,6 +164,12 @@ func DefaultEjectorConfig() *EjectorConfig {
 		DataApiTimeout:                       60 * time.Second,
 		EthRpcRetryCount:                     3,
 		EthBlockConfirmations:                0,
+		ReferenceBlockNumberOffset:           10,
```
litt3: 64 blocks may make more sense, because that's how long it takes for eth finality

ejector/ejector_config.go
```diff
@@ -131,6 +164,12 @@ func DefaultEjectorConfig() *EjectorConfig {
...
+		ReferenceBlockNumberPollInterval:     10 * time.Second,
+		ChainDataCacheSize:                   1024,
+		LogOutputType:                        string(common.JSONLogFormat),
+		LogColor:                             false,
+		MaxGasOverride:                       10_000_000,
```
litt3, line 171: Should we actually have an override defined? I don't understand the implications behind overriding this vs using gas estimation

ejector/ejection_manager.go
```diff
@@ -301,7 +301,7 @@ func (em *ejectionManager) cleanUpFailedEjection(
 }
 
 // Get the leaky bucket for a specific quorum, creating it if it doesn't already exist.
-func (em *ejectionManager) getLeakyBucketForQuorum(qid core.QuorumID) *ratelimit.LeakyBucket {
+func (em *ejectionManager) getLeakyBucketForQuorum(now time.Time, qid core.QuorumID) *ratelimit.LeakyBucket {
```
litt3, line 307: any reason you changed the signature to accept `now`, instead of using `em.timeSource`?

I'm assuming it was for the sake of unit tests, but one option would be to just use a test time source

ejector/ejector.go
```diff
@@ -0,0 +1,219 @@
...
+
+// evaluateValidator evaluates a single validator's signing rate and decides whether to eject it.
+func (e *Ejector) evaluateValidator(signingRate *validator.ValidatorSigningRate) error {
+	isEjectable := signingRate.GetSignedBatches() == 0 && signingRate.GetUnsignedBatches() > 0
+	if !isEjectable {
+		return nil
+	}
+
+	if len(signingRate.GetValidatorId()) != 32 {
+		return fmt.Errorf("invalid validator ID length: %d", len(signingRate.GetValidatorId()))
```
litt3: consider also logging the invalid ID, to help trace a misconfiguration if this actually occurs

ejector/ejector_config.go
cody-littley: I removed the "for quorum" part, as it's not necessary for understanding how we use reference block numbers. Also, updated the wording:

```
// The number of blocks to wait before using a reference block number. That is to say, do not always
	// read data from the latest block  we know about, but rather read from a block that is sufficiently old as to make
	// choosing the wrong fork unlikely.
	ReferenceBlockNumberOffset uint64
```

ejector/ejector_config.go
cody-littley: I believe it does... some of the libraries and utilities we use require the `0x`, and others require it to be absent. It's really quite confusing.

ejector/ejector_config.go
cody-littley: changed

ejector/ejector_config.go
cody-littley: During testing, my transactions would always fail unless I increased the gas

ejector/ejection_manager.go
cody-littley: Yes, actually! There was this tricky bug where time kept moving backwards. As it turns out, I need to use the same timesetamp through entire for loop. So I capture the time at the top of the loop, and pass it into this method.

Added the following to the godoc:

```
// Note: this method must accept an external time instead of using em.timeSource() directly. The external
// context needs to use a specific time between multiple function calls, and so we have to pass it in.
```

ejector/ejector.go
cody-littley: added

PR #2214, opened 2025-10-23
perf(rs-icicle): 20x speedup on GPU ntt (used for chunk generation)

samlaf:
This speeds up the gpu RS FFT by 20x for a single blob.

Note sure if the times in this drawing are not accurate, but they give an idea of what we were previously doing.
image
We somehow thought that icicle can only deal with Fr in Barrett representation on the GPU, and so spent most of our time converting gnark-crypto's Frs (which are in Montgomery form) to Barrett form. After talking with ingonyama, turns out we don't need to do this, as the GPU can deal with Montgomery form directly (not sure if its actually computing over this representation or converting it on the gpu before computing over Barrett form though..).

New results:
```
$ go test -benchmem -run=^$ -tags icicle -bench ^BenchmarkRSBackendIcicle$ github.com/Layr-Labs/eigenda/encoding/v2/bench -v

2025/10/24 21:19:54 maxprocs: Leaving GOMAXPROCS=4: CPU quota undefined
goos: linux
goarch: amd64
pkg: github.com/Layr-Labs/eigenda/encoding/v2/bench
cpu: AMD EPYC 7R13 Processor
BenchmarkRSBackendIcicle
[WARNING] Defaulting to Ingonyama icicle-cuda-license-server at `5053@license.icicle.ingonyama.com`. For more information about icicle-cuda-license, please contact support@ingonyama.com.
[INFO] ICICLE backend loaded from /opt/icicle/lib/backend
BenchmarkRSBackendIcicle/2^15_Frs
BenchmarkRSBackendIcicle/2^15_Frs-4                 2606            462804 ns/op         1048901 B/op          9 allocs/op
BenchmarkRSBackendIcicle/2^18_Frs
BenchmarkRSBackendIcicle/2^18_Frs-4                  507           2204087 ns/op         8388925 B/op          9 allocs/op
BenchmarkRSBackendIcicle/2^19_Frs
BenchmarkRSBackendIcicle/2^19_Frs-4                  261           4490175 ns/op        16777538 B/op          9 allocs/op
BenchmarkRSBackendIcicle/2^22_Frs
BenchmarkRSBackendIcicle/2^22_Frs-4                   27          44845142 ns/op        134218080 B/op        10 allocs/op
PASS
ok      github.com/Layr-Labs/eigenda/encoding/v2/bench  10.017s
```

Hard to run benchmarks prior to this PR because the benchmark was added and improved throughout this PR and I'm too lazy to clean commit history, but if the times in the diagram above were correct then 2^22_Frs encoding was taking ~800ms,  and we are down to 44ms, which explains the 20x improvement.

## Note on gpuLock and achieving 30x throughput

There is actually a 30x total throughput speedup (instead of just 20x) possible when encoding multiple blobs in parallel, which requires removing the gpuLock that we currently take on each ExtendPolyEval.

If we remove the gpuLock and run the test, we get this result when running the test for 5s. Notice running for only 1s is not enough; my guess is that somehow the gpu pipeline gets "warmed up" (whatever this means...) and achieves full pipelined throughput only after some amount of time. 
```
$ go test -benchmem -run=^$ -tags icicle -bench ^BenchmarkRSBackendIcicle$ github.com/Layr-Labs/eigenda/encoding/v2/bench -v -benchtime=5s
2025/10/23 21:47:11 maxprocs: Leaving GOMAXPROCS=4: CPU quota undefined
goos: linux
goarch: amd64
pkg: github.com/Layr-Labs/eigenda/encoding/v2/bench
cpu: AMD EPYC 7R13 Processor
BenchmarkRSBackendIcicle
[WARNING] Defaulting to Ingonyama icicle-cuda-license-server at `5053@license.icicle.ingonyama.com`. For more information about icicle-cuda-license, please contact support@ingonyama.com.
[INFO] ICICLE backend loaded from /opt/icicle/lib/backend
BenchmarkRSBackendIcicle/2^17_Frs
BenchmarkRSBackendIcicle/2^17_Frs-4                 6703            891234 ns/op         4194990 B/op          9 allocs/op
BenchmarkRSBackendIcicle/2^20_Frs
BenchmarkRSBackendIcicle/2^20_Frs-4                  804           6630871 ns/op        33557824 B/op          9 allocs/op
BenchmarkRSBackendIcicle/2^21_Frs
BenchmarkRSBackendIcicle/2^21_Frs-4                  394          13874771 ns/op        67115453 B/op         10 allocs/op
BenchmarkRSBackendIcicle/2^24_Frs
BenchmarkRSBackendIcicle/2^24_Frs-4                   39         152882016 ns/op        536934431 B/op        23 allocs/op
PASS
ok      github.com/Layr-Labs/eigenda/encoding/v2/bench  43.608s
```

Here is Ingonyama's Jeremy Felder's explanation for why the gpuLock is not needed.
> I see, I believe the GPU lock is not necessary and the two functions ComputeMultiFrameProofV2 and ExtendPolyEval seem to be using ICICLE correctly.
>
> I want to point out that calling runtime.SetDevice (https://github.com/Layr-Labs/eigenda/blob/master/encoding/v2/rs/icicle/extend_poly.go#L35-L39) should not be used and runtime.RunOnDevice should be used instead (it looks like the code uses both which is fine but not necessary). Otherwise, when calling runtime.SetDevice it is possible that the goroutine it is being called in will change host threads after it has been called which can result in the "Attempt to use DeviceSlice on a different device" error you have seen.
> - runtime.SetDevice sets the device on the current thread goroutines can change host threads at any point so this method does not guarantee that future operations will be performed on the same thread and therefore the same device
> - runtime.RunOnDevice uses SetDevice internally but locks the OS/host thread for the duration of the function guaranteeing that any operations in the function passed to RunOnDevice are on the same device that SetDevice set

However, the issue with removing the gpuLock, and why we can't do it right now, is that icicle does not do backpressure, so if we send more parallel requests to the GPU than it has RAM available, it just panics and fails. There are many solutions to this problem. Maybe just exponential backoff could work... but I think it's safer to think through this issue and come up with the correct solution in a separate PR.

### DIFF
--- a/encoding/v2/bench/benchmark_eigenda_test.go
+++ b/encoding/v2/bench/benchmark_eigenda_test.go
@@ -83,10 +83,6 @@ func BenchmarkCommittmentGeneration(b *testing.B) {
 
 // TODO(samlaf): maybe move this to benchmark_icicle_test.go file?
 // That file is currently metal only, we should generalize it.
-// Not super sure how RunParallel behaves with icicle, but noticed
-// that on g6.xlarge, when running benchmark for the default 1 second,
-// we get ~250ms/op, whereas when running for even slightly longer (5seconds),
-// we get an average of ~150ms/op.
 func BenchmarkRSBackendIcicle(b *testing.B) {
 	if !icicle.IsAvailable {
 		b.Skip("code compiled without the icicle build tag")
@@ -104,10 +100,13 @@ func BenchmarkRSBackendGnark(b *testing.B) {
 
 func benchmarkRSBackend(b *testing.B, rsBackend backend.RSEncoderBackend) {
 	rand := random.NewTestRandomNoPrint(1337)
-	blobCoeffs := rand.FrElements(1 << 24)
-	for _, logNumFrs := range []uint8{17, 20, 21, 24} {
-		b.Run("2^"+fmt.Sprint(logNumFrs)+"_Frs", func(b *testing.B) {
-			numFrs := uint64(1) << logNumFrs
+	blobCoeffs := rand.FrElements(1 << 22) // max size we benchmark below: 24+3-5=22
+	for _, blobPowerBytes := range []uint8{17, 20, 21, 24} {
+		// Reed-Solomon encoding with 8x redundancy: 2^3 = 8
+		rsExtendedBlobPowerBytes := blobPowerBytes + 3
+		rsExtendedBlobPowerFrs := rsExtendedBlobPowerBytes - 5 // 32 bytes per Fr element
+		b.Run("2^"+fmt.Sprint(rsExtendedBlobPowerFrs)+"_Frs", func(b *testing.B) {
+			numFrs := uint64(1) << rsExtendedBlobPowerFrs
 			// run multiple goroutines in parallel to better utilize the GPU
 			b.RunParallel(func(pb *testing.PB) {
 				for pb.Next() {


dmanc (dismissed):

dmanc (dismissed):

encoding/v2/bench/benchmark_eigenda_test.go
```diff
@@ -76,6 +81,44 @@ func BenchmarkCommittmentGeneration(b *testing.B) {
 	}
 }
 
+// TODO(samlaf): maybe move this to benchmark_icicle_test.go file?
+// That file is currently metal only, we should generalize it.
+// Not super sure how RunParallel behaves with icicle, but noticed
+// that on g6.xlarge, when running benchmark for the default 1 second,
```
dmanc: this is pretty interesting!

encoding/v2/rs/backend/icicle/extend_poly.go
```diff
@@ -43,44 +43,33 @@ func BuildRSBackend(logger logging.Logger, enableGPU bool) (*RSBackend, error) {
...
 		defer wg.Done()
 		defer func() {
 			if r := recover(); r != nil {
 				icicleErr = fmt.Errorf("GPU operation panic: %v", r)
 			}
 		}()
 
-		ntt.Ntt(scalars, core.KForward, &g.NttCfg, outputDevice)
+		cfg := g.NttCfg
+		// We just perform the NTT synchronously here; we have nothing to do while waiting.
```
dmanc, line 73: I don't understand this, maybe we should try playing around with the nvidia profiling tools to confirm.

dmanc: Also what about in the context of `GetFrames`, wouldn't we want this to be async along with the multiproof operations?

encoding/v2/rs/backend/icicle/extend_poly.go
```diff
@@ -43,44 +43,33 @@ func BuildRSBackend(logger logging.Logger, enableGPU bool) (*RSBackend, error) {
 
 // Encoding Reed Solomon using FFT
 func (g *RSBackend) ExtendPolyEval(coeffs []fr.Element) ([]fr.Element, error) {
-	// Lock the GPU for operations
-	g.GpuLock.Lock()
```
dmanc, line 47: Only concern is if it's gonna crash at high throughputs so I would like this build to be part of a load test.

dmanc (approved):
:shipit:

encoding/v2/rs/backend/icicle/extend_poly.go
samlaf: Afaiu async should be thought similarly to tokio async/goroutines. It's about not blocking the CPU, not about the GPU. See https://dev.ingonyama.com/3.6.0/icicle/rust-bindings/ntt
image

So anything else happening in GetFrames is irrelevant, especially since we run chunk generation and proof generation in separate goroutines (aka they are already async!). What we could have done is to run them in the same goroutines, while using async streams to let the GPU do some computation while we do some CPU stuff (for example doing the conversion work or loading srs points or whatever).

dmanc: Ah got it

encoding/v2/rs/backend/icicle/extend_poly.go
samlaf: Confirmed that it does panic... 
image

Asked ingonyama what's the correct way to do backpressure without panic'ing.. for now might just re-add the gpuLock.

samlaf: added back gpuLock in [9dff501](https://github.com/Layr-Labs/eigenda/pull/2214/commits/9dff501be28ef6d88fc430cd44a9b317283e5f62), added a TODO, and updated README description here to describe issue and possible solutions.

dmanc (dismissed):

dmanc (approved):

encoding/v2/rs/backend/icicle/extend_poly.go
bxue-l2: what is the behavior when two goroutine trying to access the same GPU? can it handle it without the lock? Does it have something to do with SetDevice?

samlaf: I added back the gpuLock (see [here](https://github.com/Layr-Labs/eigenda/pull/2214#discussion_r2461887186)) so if 2 goroutines try to ExtendPoly only one of them holds the lock and can hence do computations on GPU.

Icicle can definitely handle multiple requests in parallel, the only issue is they don't have backpressure, so if we end up sending too much data and exhaust the RAM, then icicle panics and we need to reset the device... which is really bad. We need to figure out ways to utilize the GPU more efficiently, but there doesn't seem to be any very satisfying answer (semaphore with hardcoded ram-utilization values seems the best we got...)

This all has nothing to do with SetDevice no, afaiu. We only have 1 device anyways.

bxue-l2 (approved):

PR #2215, opened 2025-10-23
docs: chunk assignment

Lulu-Zhou-EigenLabs:

update the document for chunk assignment in v2

### DIFF
--- a/api/proxy/store/generated_key/v2/eigenda.go
+++ b/api/proxy/store/generated_key/v2/eigenda.go
@@ -316,19 +316,6 @@ func verifyCertRBNRecencyCheck(certRBN uint64, certL1IBN uint64, rbnRecencyWindo
 	if certL1IBN == 0 || rbnRecencyWindowSize == 0 {
 		return nil
 	}
-	if certRBN == 0 {
-		return fmt.Errorf("certRBN should never be 0, this is likely a bug")
-	}
-	if certL1IBN <= certRBN {
-		return fmt.Errorf(
-			"cert's l1 inclusion block number (%d) <= cert reference block number (%d), but this is physically impossible "+
-				"since the cert has to be signed by all eigenda validators before being submitted to the batcher inbox, "+
-				"and validators will only sign a batchRoot (contained in certs) if the RBN is in the past. "+
-				"This is a serious bug, please report it",
-			certRBN,
-			certL1IBN,
-		)
-	}
 
 	// Actual Recency Check
 	if !(certL1IBN <= certRBN+rbnRecencyWindowSize) { //nolint:staticcheck // inequality is clearer as is
--- a/api/proxy/store/generated_key/v2/verify_test.go
+++ b/api/proxy/store/generated_key/v2/verify_test.go
@@ -18,14 +18,6 @@ func TestVerifyCertRBNRecencyCheck(t *testing.T) {
 		expectedErrorContains string
 	}{
 		{
-			name:                  "input sanization: certRBN should always be > 0",
-			certRBN:               0,
-			certL1IBN:             100,
-			rbnRecencyWindowSize:  100,
-			expectError:           true,
-			expectedErrorContains: "bug",
-		},
-		{
 			name:                 "input sanization: certL1IBN=0 should skip the test (return nil)",
 			certRBN:              100,
 			certL1IBN:            0,
@@ -38,14 +30,6 @@ func TestVerifyCertRBNRecencyCheck(t *testing.T) {
 			certL1IBN:            101,
 			rbnRecencyWindowSize: 0,
 			expectError:          false,
-		},
-		{
-			name:                  "input sanization: certL1IBN should always be > certRBN (when != 0)",
-			certRBN:               100,
-			certL1IBN:             100,
-			rbnRecencyWindowSize:  100,
-			expectError:           true,
-			expectedErrorContains: "bug",
 		},
 		{
 			name:                 "ok: certL1IBN = certRBN + rbnRecencyWindowSize",
--- a/core/payments/reservation/reservationvalidation/reservation_validator_metrics.go
+++ b/core/payments/reservation/reservationvalidation/reservation_validator_metrics.go
@@ -1,13 +1,18 @@
 package reservationvalidation
 
 import (
+	"github.com/Layr-Labs/eigenda/encoding"
+	"github.com/docker/go-units"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 )
 
 // Tracks metrics for the [ReservationPaymentValidator]
 type ReservationValidatorMetrics struct {
-	reservationSymbols               prometheus.Histogram
+	// Although payments internally tracks things in symbols, the consumer of metrics wants to see things in bytes.
+	// For a histogram, it's actually not possible to automatically rename bucket labels in grafana, so using
+	// symbols here causes dashboards to be less intuitive.
+	reservationBytes                 prometheus.Histogram
 	reservationSymbolsTotal          prometheus.Counter
 	reservationInsufficientBandwidth prometheus.Counter
 	reservationQuorumNotPermitted    prometheus.Counter
@@ -25,15 +30,15 @@ func NewReservationValidatorMetrics(
 		return nil
 	}
 
-	symbols := promauto.With(registry).NewHistogram(
+	bytes := promauto.With(registry).NewHistogram(
 		prometheus.HistogramOpts{
 			Namespace: namespace,
-			Name:      "reservation_symbols",
+			Name:      "reservation_bytes",
 			Subsystem: subsystem,
-			Help: "Distribution of symbol counts for successful reservation payments. " +
-				"Counts reflect actual dispersed symbols, not billed symbols (which may be higher due to min size).",
+			Help: "Distribution of byte counts for successful reservation payments. " +
+				"Counts reflect actual dispersed bytes, not billed bytes (which may be higher due to min size).",
 			// Buckets chosen to go from min to max blob sizes (128KiB -> 16MiB)
-			Buckets: prometheus.ExponentialBuckets(4096, 2, 8),
+			Buckets: prometheus.ExponentialBuckets(128*units.KiB, 2, 8),
 		},
 	)
 
@@ -93,7 +98,7 @@ func NewReservationValidatorMetrics(
 	)
 
 	return &ReservationValidatorMetrics{
-		reservationSymbols:               symbols,
+		reservationBytes:                 bytes,
 		reservationSymbolsTotal:          symbolsTotal,
 		reservationInsufficientBandwidth: insufficientBandwidth,
 		reservationQuorumNotPermitted:    quorumNotPermitted,
@@ -108,7 +113,7 @@ func (m *ReservationValidatorMetrics) RecordSuccess(symbolCount uint32) {
 	if m == nil {
 		return
 	}
-	m.reservationSymbols.Observe(float64(symbolCount))
+	m.reservationBytes.Observe(float64(symbolCount) * encoding.BYTES_PER_SYMBOL)
 	m.reservationSymbolsTotal.Add(float64(symbolCount))
 }
 
--- a/docs/spec/src/integration/spec/6-secure-integration.md
+++ b/docs/spec/src/integration/spec/6-secure-integration.md
@@ -145,7 +145,9 @@ For the ~7-day challenge window overlaps EigenDA availability, we assume there i
 
 In the diagram, the top row shows L1 blocks every 12 s; the smaller squares are L2 blocks every 2 s. Yellow labels mark key artifacts across the batching pipeline: batches → channel → EigenDA blob. Dispersal completes between t=12 s and t=24 s. The resulting certificate has RBN equal to the L1 block at t=0 (two L1 blocks earlier). The cert is then submitted to L1 at t=24 s. Green annotations show the generalized L2→L1 submission, with batches posted to the adjacent L1 block.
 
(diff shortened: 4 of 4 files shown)


docs/spec/src/protocol/architecture/assignment.md
```diff
@@ -1,112 +1,206 @@
...
-The assignment module is essentially a rule which takes in the Ethereum chain state and outputs an allocation of chunks to DA operators. This can be generalized to a function that outputs a set of valid allocations.
-
-A chunk assignment has the following parameters: 
-1) **Indices**: the chunk indices that will be assigned to each DA node. Some DA nodes receive more than one chunk.
-2) **ChunkLength**: the length of each chunk (measured in number of symbols, as defined by the encoding module). We currently require all chunks to be of the same length, so this parameter is a scalar. 
-
-The assignment module is implemented by the `AssignmentCoordinator` interface. 
+The assignment module is implemented in `core/v2/assignment.go`. For blobs dispersed to multiple quorums, the algorithm employs overlap optimization to minimize storage requirements while maintaining security guarantees. 
 
 ![image](../../assets/assignment-module.png)
```
mmurrs, line 7: What does Assignment Consensus refer to? This could be an old image as it appears there were no changes in this PR. 

It would be useful to use a consistent the motivation of the chunk assignment, this could be a nit but including the "Logic to ensure..." from the image or the portion from the Chunk Assignment Algorithm for a Single Quorum which starts "...ensuring that any coalition of validators with sufficient combined stake..."

docs/spec/src/protocol/architecture/assignment.md
```diff
@@ -1,112 +1,206 @@
 ## Assignment Module
 
-> Warning: this page describes the assignment logic for EigenDA V1. We need to update it with Blazar assignment logic which is very different.
+The assignment module determines how encoded blob chunks are allocated to DA operators based on the Ethereum chain state, specifically operator stakes and quorum memberships. Given the operator state and blob parameters, it produces a deterministic mapping from operators to chunk indices.
```
mmurrs: Recommend being consistent on validators throughout this doc so there's one term used throughout

docs/spec/src/protocol/architecture/assignment.md
```diff
@@ -1,112 +1,206 @@
...
-2) **ChunkLength**: the length of each chunk (measured in number of symbols, as defined by the encoding module). We currently require all chunks to be of the same length, so this parameter is a scalar. 
-
-The assignment module is implemented by the `AssignmentCoordinator` interface. 
+The assignment module is implemented in `core/v2/assignment.go`. For blobs dispersed to multiple quorums, the algorithm employs overlap optimization to minimize storage requirements while maintaining security guarantees. 
 
 ![image](../../assets/assignment-module.png)
 
 
-### Chunk Assignment Scheme Overview
+### Chunk Assignment Algorithm for a Single Quorum
```
mmurrs: What's the intention of including the single quorum chunk assignment in addition to the multiple quorums given the default case of EigenDA is dispersing to multiple quorums?

docs/spec/src/protocol/architecture/assignment.md
```diff
@@ -1,112 +1,206 @@
...
 
-In the future, additional constraints on chunk length may be added; for instance, the chunk length may be set in order to maintain a fixed number of chunks per blob across all system states. Currently, the protocol does not mandate a specific value for the chunk length, but will accept the range satisfying the above constraint. The `CalculateChunkLength` function is provided as a convenience function that can be used to find a chunk length satisfying the protocol requirements. 
+This algorithm guarantees that validators participating in both quorums store only `max(chunks_in_quorum_1, chunks_in_quorum_2)` unique chunks rather than the sum.
 
-**Index Assignment**
+**3. MergeAssignmentsAndCap:** Merges assignments across all quorums and caps the total at the reconstruction threshold:
+$
+\text{max\_chunks} = c / r
+$,
+where $c$ is the total number of chunks and $r$ is the coding rate. This cap exists because once an operator has enough unique chunks to reconstruct the blob, additional chunks provide no incremental benefit.
```
mmurrs: maybe nit, "additional chunks provide no incremental benefit" ~= this algorithm either a. reduces unnecessary work or b. is maximally? (idk if true) efficient 

The distinction I'm drawing is essentially around efficiency and ensuring there's no unnecessary work done, you essentially allude to that already but believe worth including clear value of why this is useful

mmurrs: "The multi-quorum assignment algorithm minimizes storage requirements through overlap optimization while maintaining security guarantees." 

I now reread and see you included this minimization reference, my point it just making this clearer or connecting back to established claims around minimization or optimality

docs/spec/src/protocol/architecture/assignment.md
bxue-l2: I thought we stick with operators..

docs/spec/src/protocol/architecture/assignment.md
Lulu-Zhou-EigenLabs: Yes, I reused the old image. I'll remove it to avoid confusion, since chunk assignment is not a "consensus".

I also added a sentence to explain the chunk assignment and security gurantees.

docs/spec/src/protocol/architecture/assignment.md
Lulu-Zhou-EigenLabs: I changed the title of the section for clarity.

docs/spec/src/protocol/architecture/assignment.md
```diff
@@ -1,112 +1,204 @@
...
+The assignment algorithm uses two key strategies to minimize storage:
 
-### Assignment Logic
+1. **Chunk Overlap Maximization:** When a validator participates in multiple quorums for the same blob, the algorithm reuses the same chunk indices across quorums whenever possible.
 
-The standard assignment coordinator implements a very simple logic for determining the number of chunks per node and the chunk length, which we describe here.
+2. **Reconstruction Capping:** Each validator is assigned at most the number of chunks needed to independently reconstruct the blob.
 
-**Chunk Length**
+**Example:** Consider a validator with 5% stake in quorum 0 and 15% stake in quorum 1. Without optimization, the validator might receive two non-overlapping sets of chunks (one per quorum), totaling up to 20% of all chunks. With overlap optimization, the validator stores only `max(chunks_quorum_0, chunks_quorum_1)` unique chunks, which is 15% of the total chunks. With reconstruction capping, if the coding rate is $r = 8$, the validator only needs to store 1/8 of the total chunks.
```
bxue-l2: In coding theory, I think coding rate is always expressed as some number between 0 and 1. Say the systematic symbols is 4 and sum of systematic and parity symbols are 10, then the coding ratio is 0.4.

I think the doc has been consistently calling coding rate as 2.5.

docs/spec/src/protocol/architecture/assignment.md
```diff
@@ -1,112 +1,204 @@
...
+The chunk assignment scheme assigns encoded chunks to validators proportionally to their stake, ensuring that any coalition of validators with sufficient combined stake can reconstruct the blob.
 
+Given:
 - A set of $n$ validators with stakes $\eta_1, \eta_2, \ldots, \eta_n$, where $\sum_{i=1}^n \eta_i = 1$
 - A set of $c$ chunks to be assigned to the validators
 
-The assignment algorithm `GetAssignments` works as follows:
+Within a single quorum, the number of chunks assigned to validator $i$ is:
+```math
+c'_i = \lceil \eta_i(c - n) \rceil
```
bxue-l2: I think in the security params doc, you mention c'_i >= c_i, I wonder what c_i is, though you are using c'_i here

docs/spec/src/protocol/architecture/assignment.md
```diff
@@ -1,112 +1,204 @@
...
 
-In the future, additional constraints on chunk length may be added; for instance, the chunk length may be set in order to maintain a fixed number of chunks per blob across all system states. Currently, the protocol does not mandate a specific value for the chunk length, but will accept the range satisfying the above constraint. The `CalculateChunkLength` function is provided as a convenience function that can be used to find a chunk length satisfying the protocol requirements. 
+This algorithm guarantees that validators participating in both quorums store only `max(chunks_in_quorum_1, chunks_in_quorum_2)` unique chunks rather than the sum.
 
-**Index Assignment**
+**3. MergeAssignmentsAndCap:** Merges assignments across all quorums and caps the total at the reconstruction threshold:
+```math
+\text{max\_chunks} = c / r
+```
+ where $c$ is the total number of chunks and $r$ is the coding rate. This cap exists because once a validator has enough unique chunks to reconstruct the blob, additional chunks provide no incremental security benefit. Therefore, pruning the extra chunks improves performance and reduces storage and bandwidth requirements without affecting security.
```
bxue-l2: I thought you use gamma as the inverse coding ratio in the other doc. and r was used to described the reconstruction threshold for stake

docs/spec/src/protocol/architecture/assignment.md
```diff
@@ -1,112 +1,204 @@
...
-For each operator $i$, let $S_i$ signify the amount of stake held by that operator. We want for the number of chunks assigned to operator $i$ to satisfy
+**4. GetAssignmentsForBlob:** Coordinates the full multi-quorum assignment process:
+1. Generate the assignment for quorum 0 using `GetAssignmentsForQuorum`
+2. Generate assignments for all other quorums using `AddAssignmentsForQuorum` with quorum 0 as the baseline
+3. Merge all per-quorum assignments using `MergeAssignmentsAndCap` to produce the final assignment for each validator
 
-$$
-\frac{\gamma m_i C}{B} \ge \frac{S_i}{\sum_j S_j}
-$$
+**Note on Optimality:** The algorithm produces optimal storage assignments for two quorums. For three or more quorums, the assignment is not guaranteed to be globally optimal. Since quorums 0 and 1 are typically the largest (containing most validators), and other quorums are smaller custom quorums, the algorithm achieves near-optimal storage reduction for the majority of validators.
```
bxue-l2: What is the effective coding ratio here, let's its definition is number of chunks actually stored by all the operator dividing the original blob size that is power of 2. And let's ignore the capping optimization for now

docs/spec/src/protocol/architecture/assignment.md
Lulu-Zhou-EigenLabs: Good catch. I wrote the document for security parameters assuming that every chunk is assigned to a validator. Now I realize some chunks may remain unassigned. I'll change the security parms doc accordingly.

docs/spec/src/protocol/architecture/assignment.md
Lulu-Zhou-EigenLabs: Good catch. I will make the notation consistent.

docs/spec/src/protocol/architecture/assignment.md
Lulu-Zhou-EigenLabs: Aligned the notation for coding rate. Now the coding rate is $\gamma = 1/8$.

docs/spec/src/protocol/architecture/assignment.md
Lulu-Zhou-EigenLabs: The actual number stored by all the operators is $\sum_{i=1}^n \lceil \eta_i(c - n) \rceil $. Therefore, the effecitve coding rate is $\sum_{i=1}^n \lceil \eta_i(c - n) \rceil  / c$. It depends on the number of validators $n$.

docs/spec/src/protocol/architecture/assignment.md
bxue-l2: Does it affect other parts, like if the actual code is using the old name/

Lulu-Zhou-EigenLabs: Yes, the actual code is still calling the inverse of coding rate as `codingRate`. I've noted this in `./security-parameters.md#blob-parameters`.

docs/spec/src/protocol/architecture/assignment.md
```diff
@@ -1,112 +1,207 @@
...
```
cody-littley: nit
```suggestion
**Note on Optimality:** The algorithm produces optimal storage assignments for two quorums. For three or more quorums, the assignment is not guaranteed to be globally optimal. Since quorums 0 and 1 are the "default" quorums and are expected to be the larger than custom quorums (i.e. containing the most validators), the algorithm achieves near-optimal storage reduction for the majority of validators.
```

docs/spec/src/protocol/architecture/assignment.md
```diff
@@ -1,112 +1,207 @@
...
 
-Let
+### Code Walkthrough
+Notation note: In the code, we use `operator` to refer to `validator`.
```
cody-littley:
```suggestion
Notation note: In the code, we sometimes use the term `operator` to refer to a `validator`, although `validator` is now the preferred term.
```

docs/spec/src/protocol/architecture/assignment.md
```diff
@@ -1,112 +1,207 @@
...
+```go
+func GetAssignmentsForQuorum(
+    state *core.OperatorState,
+    blobParams *core.BlobVersionParameters,
+    quorum core.QuorumID,
+) (map[core.OperatorID]*Assignment, []core.OperatorID, error)
+```
 
+Algorithm:
+1. Sort operators by hex ID for determinism
```
cody-littley, line 93: Is it possible to cause us grief if a validator or validators intentionally chose an ID that shows up first or last in a sorted list? Probably not particularly hard, you just have to generate ~100 keys until you find one that appears first/last in the order.

I'm not aware of a specific attack, but it might be a good edge case to think about.

Lulu-Zhou-EigenLabs: After thinking more about this, I realize the sorting is not necessary. We can start with any order, and the order will not affect the assignment process.

cody-littley (approved):

PR #2218, opened 2025-10-23
feat(proxy): Arb Commitment Type, invalid cert signal, && logging

ethenotethan:
- Adds Arbitrum specific commitment type that's used for encoding DA Cert into Arbitrum DA Commitment. For OP alignment; I've chosen the following schema:
```
[CustomDAByte, DALayerByte, CertVersion, DACert]
        0          1            2        3-N
 
 where:
 CustomDAByte is the Arbitrum CustomDA message header identifier
 DALayerByte is used to distinguish EigenDA from other providers
 CertVersion refers to the Cert type version
 DA Cert is the RLP encoded cert (THIS SHOULD BE ABI, I've added a todo and will follow up with another to introduce the adjustment)
```

- Adds translation from `DerivationError` to `nil` return to indicate a status 418 "drop cert"
- Updates methods to be reflective of latest DA client [interface](https://github.com/OffchainLabs/nitro/blob/021fb7ec40dbc998437090ebead98218ad3c5435/daprovider/daclient/daclient.go#L188)
- Introduces info logging for method calls

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/meterer/on_demand_meterer.go
+++ b/core/meterer/on_demand_meterer.go
@@ -50,6 +50,7 @@ func (m *OnDemandMeterer) MeterDispersal(symbolCount uint32) (*rate.Reservation,
 		return nil, fmt.Errorf("global rate limit exceeded: cannot reserve %d symbols", symbolCount)
 	}
 
+	m.metrics.RecordGlobalMeterThroughput(symbolCount)
 	return reservation, nil
 }
 
--- a/core/meterer/on_demand_meterer_metrics.go
+++ b/core/meterer/on_demand_meterer_metrics.go
@@ -7,8 +7,10 @@ import (
 
 // Tracks metrics for the [OnDemandMeterer]
 type OnDemandMetererMetrics struct {
-	onDemandGlobalMeterExhaustedRequests prometheus.Counter
-	onDemandGlobalMeterExhaustedSymbols  prometheus.Counter
+	onDemandGlobalMeterExhaustedRequests  prometheus.Counter
+	onDemandGlobalMeterExhaustedSymbols   prometheus.Counter
+	onDemandGlobalMeterThroughputRequests prometheus.Counter
+	onDemandGlobalMeterThroughputSymbols  prometheus.Counter
 }
 
 func NewOnDemandMetererMetrics(
@@ -38,9 +40,29 @@ func NewOnDemandMetererMetrics(
 		},
 	)
 
+	onDemandGlobalMeterThroughputRequests := promauto.With(registry).NewCounter(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "on_demand_global_meter_throughput_requests_count",
+			Subsystem: subsystem,
+			Help:      "Total number of requests successfully metered for on-demand dispersals",
+		},
+	)
+
+	onDemandGlobalMeterThroughputSymbols := promauto.With(registry).NewCounter(
+		prometheus.CounterOpts{
+			Namespace: namespace,
+			Name:      "on_demand_global_meter_throughput_symbols_count",
+			Subsystem: subsystem,
+			Help:      "Total number of symbols successfully metered for on-demand dispersals",
+		},
+	)
+
 	return &OnDemandMetererMetrics{
-		onDemandGlobalMeterExhaustedRequests: onDemandGlobalMeterExhaustedRequests,
-		onDemandGlobalMeterExhaustedSymbols:  onDemandGlobalMeterExhaustedSymbols,
+		onDemandGlobalMeterExhaustedRequests:  onDemandGlobalMeterExhaustedRequests,
+		onDemandGlobalMeterExhaustedSymbols:   onDemandGlobalMeterExhaustedSymbols,
+		onDemandGlobalMeterThroughputRequests: onDemandGlobalMeterThroughputRequests,
+		onDemandGlobalMeterThroughputSymbols:  onDemandGlobalMeterThroughputSymbols,
 	}
 }
 
@@ -51,4 +73,13 @@ func (m *OnDemandMetererMetrics) RecordGlobalMeterExhaustion(symbolCount uint32)
 	}
 	m.onDemandGlobalMeterExhaustedRequests.Inc()
 	m.onDemandGlobalMeterExhaustedSymbols.Add(float64(symbolCount))
+}
+
+// RecordGlobalMeterThroughput records successful metering for on-demand dispersals
+func (m *OnDemandMetererMetrics) RecordGlobalMeterThroughput(symbolCount uint32) {
+	if m == nil {
+		return
+	}
+	m.onDemandGlobalMeterThroughputRequests.Inc()
+	m.onDemandGlobalMeterThroughputSymbols.Add(float64(symbolCount))
 }
--- a/core/payments/ondemand/ondemandvalidation/on_demand_validator_metrics.go
+++ b/core/payments/ondemand/ondemandvalidation/on_demand_validator_metrics.go
@@ -7,7 +7,8 @@ import (
 
 // Tracks metrics for the [OnDemandPaymentValidator]
 type OnDemandValidatorMetrics struct {
-	onDemandSymbolCount        prometheus.Histogram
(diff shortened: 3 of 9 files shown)


api/proxy/servers/arbitrum_altda/handlers.go
```diff
@@ -114,13 +116,10 @@ func NewHandlers(m *store.EigenDAManager) IHandlers {
...
 
-// GetSupportedHeaderBytes determines whether or not the sequencer message header byte is an EigenDAV2 cert type.
-// Arbitrum Nitro does this check via a bitwise AND which can cause overlapping and requires careful future
-// management. while we could determine a byte value with bits that don't overlap - it's more maintainable
-// to do a literal comparison and assume OCL NOR our competitors would never introduce a conflicting byte value
+// GetSupportedHeaderBytes returns the supported DA Header bytes by the CustomDA server
 func (h *Handlers) GetSupportedHeaderBytes(ctx context.Context) (*SupportedHeaderBytesResult, error) {
 	return &SupportedHeaderBytesResult{
-		HeaderBytes: []byte{EigenDAV2MessageHeaderByte},
+		HeaderBytes: []byte{commitments.ArbCustomDAHeaderByte},
```
samlaf, line 132: what are other possible header bytes?

api/proxy/servers/arbitrum_altda/handlers.go
```diff
@@ -139,23 +138,30 @@ func (h *Handlers) RecoverPayload(
...
-	//       Given the lack of response from OCL, it's still unknown what the exact expected schema should.
-	//       Once their interface is more hardended we can make a safer deduction for how to introduce this.
-	//       Will likely require introducing a new ArbCustomDA commitment type
-	//
-	versionByte := sequencerMsg[0]
-	versionedCert := certs.NewVersionedCert([]byte(sequencerMsg[1:]), certs.VersionByte(versionByte))
+	daCommitByte := sequencerMsg[0]
+	if daCommitByte != commitments.ArbCustomDAHeaderByte {
+		return nil,
+			fmt.Errorf("expected %x for header byte, got %x", commitments.ArbCustomDAHeaderByte, daCommitByte)
```
samlaf: nit: feels like we could give a more meaningful error msg, something like "proxy can only deal with altda commitments" or something?

samlaf: instead of just seeing bytes and being confused as to what they mean

samlaf (commented):
LGTM, although would be nice if arb could handle 418 errors.

api/proxy/servers/arbitrum_altda/handlers.go
ethenotethan: there are none lol - this is Arbitrum's way to compensate for AnyTrust compatibility where there are legacy header types that need to be supported

api/proxy/servers/arbitrum_altda/handlers.go
ethenotethan: [4e288e5](https://github.com/Layr-Labs/eigenda/pull/2218/commits/4e288e56338d122693ad8c102fd63f736810b327)

api/proxy/servers/arbitrum_altda/handlers.go
samlaf: This is what I meant though, might be good to add a comment as to that effect, since people less familiar with arbitrum (like me) would come here and wonder why we even need this function. Good to list alternate headerBytes and mention why proxy doesn't support them.

samlaf (dismissed):
LGTM. See that you added an EigenDA header byte
image

Is this new? Or was it there before just less explicit and I missed it?

api/proxy/servers/arbitrum_altda/handlers.go
ethenotethan: [cb1b442](https://github.com/Layr-Labs/eigenda/pull/2218/commits/cb1b442721f276b06ddbba77c657b44e6418ae17)

samlaf (dismissed):

api/proxy/servers/arbitrum_altda/handlers.go
```diff
@@ -105,31 +109,75 @@ type Handlers struct {
...
 	return &SupportedHeaderBytesResult{
-		HeaderBytes: []byte{EigenDAV2MessageHeaderByte},
+		HeaderBytes: []byte{commitments.ArbCustomDAHeaderByte},
 	}, nil
 }
 
+// deserializeCertFromSequencerMsg reads the VersionedCert from the raw sequencer message provided
+// by the DA Client
+func (h *Handlers) deserializeCertFromSequencerMsg(sequencerMsg hexutil.Bytes) (certs.VersionedCert, error) {
+	if len(sequencerMsg) <= 42 {
```
shyam-patel-kira, line 139: nit: Can we comment or define these numbers somewhere? something like
```go 
const (
      // ArbSequencerHeaderSize is the size of the Arbitrum sequencer message header
      // that precedes the CustomDA commitment data
      ArbSequencerHeaderSize = 40
      // MinArbSequencerMsgSize is the minimum size for a valid Arbitrum sequencer message
      // (40 byte header + at least 2 bytes for CustomDA header)
      MinArbSequencerMsgSize = 42
  )
  ```

File: api/proxy/servers/arbitrum_altda/handlers.go [hunk 4]
```diff
@@ -105,31 +109,75 @@ type Handlers struct {
...
+		return certs.VersionedCert{},
+			fmt.Errorf("expected CutomDAHeader byte (%x) for 0th index byte of message, instead got: %x ",
+				commitments.ArbCustomDAHeaderByte, daHeaderByte)
+	}
+
+	daLayerByte := seqMessageWithoutHeader[1]
+	if daLayerByte != commitments.EigenDALayerByte {
+		return certs.VersionedCert{},
+			fmt.Errorf("expected EigenDALayer byte (%x) for 1st index byte of message, instead got: %x ",
+				commitments.EigenDALayerByte, daHeaderByte)
```
shyam-patel-kira: iiuc this error message should reference `daHeaderByte` instead of `daLayerByte` in the format string

api/proxy/servers/arbitrum_altda/handlers.go
```diff
@@ -105,31 +109,75 @@ type Handlers struct {
...
+		return certs.VersionedCert{},
+			fmt.Errorf("sequencer message expected to be >=42 bytes, got: %d", len(sequencerMsg))
+	}
+
+	seqMessageWithoutHeader := sequencerMsg[40:]
+
+	daHeaderByte := seqMessageWithoutHeader[0]
+	if daHeaderByte != commitments.ArbCustomDAHeaderByte {
+		return certs.VersionedCert{},
+			fmt.Errorf("expected CutomDAHeader byte (%x) for 0th index byte of message, instead got: %x ",
```
shyam-patel-kira:
```suggestion
			fmt.Errorf("expected CustomDAHeader byte (%x) for 0th index byte of message, instead got: %x ",
```

shyam-patel-kira: nit: typo

api/proxy/servers/arbitrum_altda/handlers.go
ethenotethan: eh I'd rather not yet given this schema is subject to change and I've asked OCL about removing the header entirely when calling the DA Server

api/proxy/servers/arbitrum_altda/handlers.go
ethenotethan: https://github.com/Layr-Labs/eigenda/pull/2218/commits/ed96763b0a2030d86be75ebac095851e786494b2

api/proxy/servers/arbitrum_altda/handlers.go
ethenotethan: https://github.com/Layr-Labs/eigenda/pull/2218/commits/ed96763b0a2030d86be75ebac095851e786494b2

shyam-patel-kira (approved):

PR #2219, opened 2025-10-24
feat(proxy): Add `CompatibilityConfig` endpoints to proxy servers

iquidus:
Closes DAINT-809

Adds a `/config` GET endpoint to the proxy rest server intended to return proxy version and various configuration values allowing for external services to check compatibility

Adds a `daprovider_compatibilityConfig` RPC method to the arb custom DA server.

Adding endpoints exposing proxy version and various configuration values allows external services (e.g hokulea) to check compatibility with proxy on startup and crash early if there is any misalignments. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/proxy/cmd/server/entrypoint.go
+++ b/api/proxy/cmd/server/entrypoint.go
@@ -37,7 +37,7 @@ func StartProxyService(cliCtx *cli.Context) error {
 
 	log.Info("Starting EigenDA Proxy Service", "version", Version, "date", Date, "commit", Commit)
 
-	cfg, err := config.ReadAppConfig(cliCtx, Version)
+	cfg, err := config.ReadAppConfig(cliCtx)
 	if err != nil {
 		return fmt.Errorf("read cli config: %w", err)
 	}
@@ -89,7 +89,7 @@ func StartProxyService(cliCtx *cli.Context) error {
 		return fmt.Errorf("build storage managers: %w", err)
 	}
 
-	// Construct and the compatibility config for the rest and arb servers. This could not be done while reading configs
+	// Construct the compatibility config for the rest and arb servers. This could not be done while reading configs
 	// as ChainID is fetched from the ethClient afterwards.
 	compatibilityCfg, err := common.NewCompatibilityConfig(
 		Version,
--- a/api/proxy/config/app_config.go
+++ b/api/proxy/config/app_config.go
@@ -51,7 +51,7 @@ func (c AppConfig) Check() error {
 	return nil
 }
 
-func ReadAppConfig(ctx *cli.Context, version string) (AppConfig, error) {
+func ReadAppConfig(ctx *cli.Context) (AppConfig, error) {
 	storeBuilderConfig, err := builder.ReadConfig(ctx)
 	if err != nil {
 		return AppConfig{}, fmt.Errorf("read proxy config: %w", err)
--- a/test/v2/client/proxy_wrapper.go
+++ b/test/v2/client/proxy_wrapper.go
@@ -5,7 +5,6 @@ import (
 	"fmt"
 
 	"github.com/Layr-Labs/eigenda/api/proxy/clients/standard_client"
-	"github.com/Layr-Labs/eigenda/api/proxy/common"
 	proxycommon "github.com/Layr-Labs/eigenda/api/proxy/common"
 	proxyconfig "github.com/Layr-Labs/eigenda/api/proxy/config"
 	proxymetrics "github.com/Layr-Labs/eigenda/api/proxy/metrics"
@@ -38,7 +37,7 @@ func NewProxyWrapper(
 
 	registry := prometheus.NewRegistry()
 	proxyMetrics := proxymetrics.NewMetrics(registry)
-	ethClient, _, err := common.BuildEthClient(
+	ethClient, _, err := proxycommon.BuildEthClient(
 		ctx,
 		logger,
 		proxyConfig.SecretConfig.EthRPCURL,


api/proxy/servers/rest/routing.go
```diff
@@ -128,6 +128,9 @@ func (svr *Server) RegisterRoutes(r *mux.Router) {
 		r.HandleFunc("/admin/eigenda-dispersal-backend", svr.handleGetEigenDADispersalBackend).Methods("GET")
 		r.HandleFunc("/admin/eigenda-dispersal-backend", svr.handleSetEigenDADispersalBackend).Methods("PUT")
 	}
+
+	// proxy config endpoint
+	r.HandleFunc("/config", svr.handleGetConfig).Methods("GET")
```
samlaf: I used /config, but we might want to discuss what we actually want to call this. config feels a bit weird since what we're exposing is a very very small subset of the actual config.
Maybe /status? or /health? or /info? or /rollup-config? /version? not sure..

api/proxy/servers/rest/server.go
```diff
@@ -18,11 +18,18 @@ import (
 	"github.com/gorilla/mux"
 )
 
+// ProxyConfig ... ProxyConfig is used by the /config endpoint
+type ProxyConfig struct {
+	Version           string `json:"version"`
+	RecencyWindowSize uint64 `json:"recencyWindowSize"`
```
samlaf:
```suggestion
	RecencyWindowSize uint64 `json:"recency_window_size"`
```

samlaf: nit: not sure about this one but json keys are typically snake cased.
We can discuss what we prefer. I think we are already using camel case for other json we are returning right? so might be btter to remain consistent then..

api/proxy/cmd/server/entrypoint.go
```diff
@@ -76,26 +76,24 @@ func StartProxyService(cliCtx *cli.Context) error {
...
-		}
-
-		log.Info("Started EigenDA Proxy REST ALT DA server",
-			string(enabled_apis.Admin), restEnabledCfg.Admin,
-			string(enabled_apis.StandardCommitment), restEnabledCfg.StandardCommitment,
-			string(enabled_apis.OpGenericCommitment), restEnabledCfg.OpGenericCommitment,
-			string(enabled_apis.OpKeccakCommitment), restEnabledCfg.OpKeccakCommitment)
+	if err := restServer.Start(router); err != nil {
+		return fmt.Errorf("start proxy rest server: %w", err)
+	}
```
samlaf: you'll prob want to change https://github.com/Layr-Labs/eigenda/blob/364bb2dbd49f1b43a16dcc0e7c959a016f8eff81/api/proxy/config/enablement/enabled_apis.go#L29 to say CertEndpointEnabled() or something. And comment that config endpoint is always enabled by default.

api/proxy/servers/rest/server.go
```diff
@@ -18,11 +18,18 @@ import (
 	"github.com/gorilla/mux"
 )
 
+// ProxyConfig ... ProxyConfig is used by the /config endpoint
+type ProxyConfig struct {
```
samlaf: not a fan of this name but not super sure what else to use. Maybe PubliclyExposedConfig or something?

samlaf (commented):
LGTM. simpler than I would have thought. left some nits/discussion points

api/proxy/servers/rest/server.go
iquidus: Yeah was also not a fan but had nothing better. I like PubliclyExposedConfig.

api/proxy/servers/rest/routing.go
iquidus: /health endpoint already exists and simply returns 200. We could combine these?
/status I feel is more suitable if the values change (e.g block height or some other counter)
/info works
/rollup-config is too specific imo (doesn't make sense for non-rollup use cases) 
/version doesn't feel right as we are exposing other values

api/proxy/servers/rest/server.go
iquidus: There's some inconsistencies across the monorepo. Snake case is mostly used, however, camel and pascal are also. Here I went with camel for consistency with the admin endpoint. https://github.com/Layr-Labs/eigenda/blob/f79446ec1c339270eb019ffea70f10e71573a704/api/proxy/servers/rest/handlers_misc.go#L40

api/proxy/servers/rest/server.go
iquidus: Updated to `PubliclyExposedInfo` in [71a623f](https://github.com/Layr-Labs/eigenda/pull/2219/commits/71a623f1d61d03bf9b5bf9777bde6c497013f182)

api/proxy/cmd/server/entrypoint.go
iquidus: Changed to `HasCertEndpointEnabled` in [71a623f](https://github.com/Layr-Labs/eigenda/pull/2219/commits/71a623f1d61d03bf9b5bf9777bde6c497013f182)

api/proxy/servers/rest/server.go
iquidus: Changed to snake case as part of [3cbf9](https://github.com/Layr-Labs/eigenda/pull/2219/commits/ce3cbf9171688678e83c66a5ce8d4b7b1924d272)

After more thought, it makes sense to add dispersal backend to the info endpoint and deprecate the get endpoint in the admin api. The above commit does the first part, deprecating the admin get endpoint however will require more discussion and planning (it's used by various tests/ci, including op-fork).

api/proxy/servers/rest/routing.go
iquidus: updated to /info in [71a623f](https://github.com/Layr-Labs/eigenda/pull/2219/commits/71a623f1d61d03bf9b5bf9777bde6c497013f182)

samlaf: Now that you've added a lot more fields to the exposed config I feel like /config might make more sense than info lol? What do others think? cc @ethenotethan

api/proxy/servers/rest/server.go
samlaf: cool will start a new thread wrt dispersal backend. agree with ethen it should remain in admin API only, as we plan to deprecate it (with V1) anyways.

api/proxy/cmd/server/entrypoint.go
samlaf: hmm realized OPKeccak is part of that checked routes, and it isnt a cert endpoint....
this is bike shedding at this point, but maybe we just want to call them `hasDataEndpointsEnabled` (data as opposed to config/health endpoints?) Don't have a strong opinion here, feel free to resolve.

api/proxy/config/app_config.go
```diff
@@ -51,21 +51,29 @@ func (c AppConfig) Check() error {
...
+func ReadAppConfig(ctx *cli.Context, version string) (AppConfig, error) {
 	storeBuilderConfig, err := builder.ReadConfig(ctx)
 	if err != nil {
 		return AppConfig{}, fmt.Errorf("read proxy config: %w", err)
 	}
 
 	enabledServersCfg := enablement.ReadEnabledServersCfg(ctx)
+	restPublicInfo := rest.PubliclyExposedInfo{
+		Version:             version,
+		Network:             storeBuilderConfig.ClientConfigV2.EigenDANetwork.String(),
```
samlaf: don't think we should expose network as its more of a syntactic sugar config that desugars to populate other config values. Think we should instead expose the chainID, which is the only other one missing afaiu.

api/proxy/servers/rest/handlers_misc.go
```diff
@@ -105,6 +105,17 @@ func (svr *Server) handleSetEigenDADispersalBackend(w http.ResponseWriter, r *ht
 	svr.writeJSON(w, r, response)
 }
 
+// handleGetInfo handles the GET request to check proxy info.
+// This endpoint returns the proxy version, and any info that may be valuable to
+// external services (e.g recency window size), to ensure correct configuration on both sides.
+func (svr *Server) handleGetInfo(w http.ResponseWriter, r *http.Request) {
```
samlaf: ngl I think GetInfo sounds weird now that I read it. How about we revert to GetConfig?

api/proxy/servers/rest/server.go
```diff
@@ -18,11 +18,23 @@ import (
 	"github.com/gorilla/mux"
 )
 
+// PubliclyExposedInfo ... PubliclyExposedInfo is used by the /info endpoint
+type PubliclyExposedInfo struct {
+	Version             string `json:"version"`
```
samlaf: document how this is currently injected. Good for people to be aware

api/proxy/servers/rest/server.go
```diff
@@ -18,11 +18,23 @@ import (
...
+	DispersalBackend    string `json:"dispersal_backend"`
```
samlaf: as discussed during standup with @ethenotethan , we should not expose this, as it'll be deprecated and removed when we get rid of V1 in a month or so.

docker-bake.hcl
```diff
@@ -284,6 +284,11 @@ target "proxy" {
   context    = "."
   dockerfile = "./Dockerfile"
   target     = "proxy"
+  args       = {
+    SEMVER    = "${SEMVER}"
+    GITCOMMIT = "${GIT_SHORT_SHA}"
```
samlaf, line 289: should we not just use GITCOMMIT everywhere and get rid of GIT_SHORT_SHA?

iquidus: This should be in it's own PR. 

The opposite direction would make more sense, we use both `GIT_SHA` and `GIT_SHORT_SHA`, 
https://github.com/Layr-Labs/eigenda/blob/758e9487d8da019b9edd3c794d76f0a8b53d21e6/docker-bake.hcl#L20-L28

https://github.com/Layr-Labs/eigenda/blob/758e9487d8da019b9edd3c794d76f0a8b53d21e6/Makefile#L11-L13

`GIT_SHA` and `GIT_COMMIT` is less clear.

api/proxy/servers/rest/handlers_misc_test.go
```diff
@@ -16,6 +16,137 @@ import (
...
+		require.NoError(t, err)
+
+		// Verify all fields
+		require.Equal(t, testPublicInfo.Version, response.Version)
+		require.Equal(t, testPublicInfo.Network, response.Network)
+		require.Equal(t, testPublicInfo.DirectoryAddress, response.DirectoryAddress)
+		require.Equal(t, testPublicInfo.CertVerifierAddress, response.CertVerifierAddress)
+		require.Equal(t, testPublicInfo.MaxBlobSizeBytes, response.MaxBlobSizeBytes)
+		require.Equal(t, testPublicInfo.RecencyWindowSize, response.RecencyWindowSize)
+		require.Equal(t, common.EigenDABackendToString(common.V1EigenDABackend), response.DispersalBackend)
```
ethenotethan: why not update the standard client with some `GetConfig` and use that for testing here? https://github.com/Layr-Labs/eigenda/blob/master/api/proxy/clients/standard_client/client.go

docker-bake.hcl
samlaf: ah I see so proxy is the one not following the correct convention. Can you do a quick follow-up or parallel PR to change this then please?

api/proxy/servers/rest/handlers_misc_test.go
ethenotethan: tangentially would it also make sense to add a structured `Config` type to the clients package with data accessors for reading things like `semver` in a simple way? iiuc we're identifying semver through:
```
git describe --tags --always --dirty
v2.4.0-43-g3b4f9f40
```

^ it might be useful to have common logic for extracting (major, minor, patch, ..etc)

api/proxy/servers/rest/server.go
```diff
@@ -18,11 +18,23 @@ import (
...
+	Network             string `json:"network"`
```
ethenotethan: NetworkID

api/proxy/servers/rest/server.go
```diff
@@ -18,11 +18,23 @@ import (
...
+	DirectoryAddress    string `json:"directory_address"`
+	CertVerifierAddress string `json:"cert_verifier_address"`
+	MaxBlobSizeBytes    uint64 `json:"max_blob_size_bytes"`
```
ethenotethan: can we document how this value would be useful to a rollup (e.g, limit alignment for batch poster)?

api/proxy/servers/rest/server.go
```diff
@@ -18,11 +18,23 @@ import (
...
+	RecencyWindowSize   uint64 `json:"recency_window_size"`
+}
+
 // Config ... Config for the proxy HTTP server
 type Config struct {
 	Host        string
 	Port        int
 	APIsEnabled *enablement.RestApisEnabled
+	PublicInfo  PubliclyExposedInfo
```
ethenotethan: nit - I don't like the name `Public` as it implies publicly accessible to the world when in actually this is just a viewable config that's intelligently read by a client for compatibility checks and for sub-config population used for client -> proxy interaction.

api/proxy/servers/rest/server.go
```diff
@@ -18,11 +18,23 @@ import (
 	"github.com/gorilla/mux"
 )
 
+// PubliclyExposedInfo ... PubliclyExposedInfo is used by the /info endpoint
+type PubliclyExposedInfo struct {
```
ethenotethan: should we also return the APIs enabled?

Makefile
```diff
@@ -90,12 +90,19 @@ fmt-check:
 # The images will be tagged with :dev, which is the default BUILD_TAG in docker-bake.hcl.
 # This can be changed by running for example `BUILD_TAG=master make docker-build`.
 docker-build:
+	SEMVER=${SEMVER} GITDATE=${GITDATE} GIT_SHA=${GITSHA} GIT_SHORT_SHA=${GITCOMMIT}  \
 	docker buildx bake all --load
```
ethenotethan, line 93: feel like someone from infra should take a look here for sanity
cc @dmanc

api/proxy/Makefile
```diff
@@ -15,7 +15,7 @@ clean:
 
 docker-build:
 	# we only use this to build the docker image locally, so we give it the dev tag as a reminder
-	cd ../.. && docker buildx bake proxy --load
+	cd ../.. && SEMVER=$(GIT_TAG) GIT_SHORT_SHA=$(GIT_COMMIT) GITDATE=$(BUILD_TIME) docker buildx bake proxy --load
```
ethenotethan, line 18: where does `GIT_SHORT_SHA` come from? looking at other services in the mono Dockerfile only showcases build context for `(SEMVER, GITCOMMIT, GITDATE)`

api/proxy/config/app_config.go
```diff
@@ -51,21 +51,29 @@ func (c AppConfig) Check() error {
...
+		DirectoryAddress:    storeBuilderConfig.ClientConfigV2.EigenDADirectory,
+		CertVerifierAddress: storeBuilderConfig.ClientConfigV2.EigenDACertVerifierOrRouterAddress,
+		MaxBlobSizeBytes:    storeBuilderConfig.ClientConfigV2.MaxBlobSizeBytes,
```
ethenotethan: Does a client actually care about the `MaxBlobSizeBytes` or the `MaxSupportedPayloadSizeBytes` where:
```
MaxSupportedPayloadSizeBytes --codec_encode --> blob with length of MaxBlobSizeBytes
```

api/proxy/servers/rest/server.go
ethenotethan: nit - maybe `DisperserNetworkID` and `CertVerifierNetworkID` to support the potential inclusion of CAFE @samlaf

api/proxy/servers/rest/server.go
```diff
@@ -18,11 +18,23 @@ import (
...
+	Version             string `json:"version"`
+	DispersalBackend    string `json:"dispersal_backend"`
+	Network             string `json:"network"`
+	DirectoryAddress    string `json:"directory_address"`
+	CertVerifierAddress string `json:"cert_verifier_address"`
```
ethenotethan: does a client actually care about these values?

api/proxy/servers/rest/server.go
ethenotethan: any value in adding `ReadOnlyMode` as a field here for a batch poster compatibility check?

api/proxy/cmd/server/entrypoint.go
ethenotethan: hmm with the addition of ArbCustomDA server I'm not actually a big fan of making the rest server enabled by default; ie if we integrate into our Nitro x CustomDA fork then this requires our users to have expose two separate ports on their proxy applications and utilize two separate clients (one at bootstrap, one for DA ops). 

@samlaf what do we expect client integration to look like on the rollup side? do we just import standard client and use that for the purposes of bootstrapping or do we extend the existing DA Client objects to support calling into this method. If the latter then we should keep the REST Server disabled by default and add a custom RPC method to the ARB Server for info extraction. If we do the former then the standard client would just exist ephemerally as part of a rollup node's initialization cycle

api/proxy/test/testutils/test_suite.go
```diff
@@ -84,7 +84,7 @@ func CreateTestSuite(
 		panic(fmt.Sprintf("build storage managers: %v", err.Error()))
 	}
 
-	if appConfig.EnabledServersConfig.RestAPIConfig.Enabled() {
+	if appConfig.EnabledServersConfig.RestAPIConfig.HasCertEndpointEnabled() {
```
ethenotethan: nit - what about `DAEndpointEnabled()` instead?

api/proxy/servers/rest/server.go
iquidus: ChainID is more useful here than NetworkID. NetworkID is only used for basic peer validation (should my node bother talking to your node), where as ChainID is used for replay protection. Although in the vast majority of cases these values match, there are exceptions. E.g when a chain forks and both forks survive, ETC for example has a NetworkID of 1 (same as Ethereum), but ChainID of 61. If the endpoint returns NetworkID: 1, is it connected to ETH or ETC? Unlikely a scenario we would ever encounter but an edge case to be aware of. Also, all the EigenDANetwork logic uses ChainID.

This required some refactoring of where we construct the ethClient, but done in [cce4ec6](https://github.com/Layr-Labs/eigenda/pull/2219/commits/cce4ec68d4bbc6142fea2ed3750cdcaf4395dd98)

api/proxy/servers/rest/server.go
iquidus: Updated to `CompatibilityConfig`

api/proxy/servers/rest/server.go
iquidus: They might, for validation/verification purposes. Can remove if no one sees value here.

api/proxy/servers/rest/handlers_misc_test.go
iquidus: Have removed the 'v' prefix from the version string for compatibility with the existing eigenda/common/version helper functions in [b3f27ea](https://github.com/Layr-Labs/eigenda/pull/2219/commits/b3f27ea2060e0dd781d925fbaf73d66799db5383)

api/proxy/config/app_config.go
iquidus: Changed to `MaxPayloadSizeBytes` in [d2317d6](https://github.com/Layr-Labs/eigenda/pull/2219/commits/d2317d6ff35d53a03bb8da1811190f6391961611)

api/proxy/cmd/server/entrypoint.go
samlaf: is the health endpoint not enabled by default anyways? If not then that feels like a bug?
Or are you envisioning that each server (rest, json-rpc) should have its own health and config endpoint?
That could work but def introduces some complexity, especially if they turn on both since we need to make sure they are consistent. @iquidus how much work is it to expose these endpoints on both servers?

iquidus: This much: https://github.com/Layr-Labs/eigenda/pull/2219/commits/a741d3cce8d4c5317519e60fb030d132e88a0b00

api/proxy/servers/rest/server.go
iquidus: Yeah, this is a good addition. Even just for debugging purposes to signal whether a signer key has been set.

api/proxy/cmd/server/entrypoint.go
```diff
@@ -56,27 +58,60 @@ func StartProxyService(cliCtx *cli.Context) error {
...
 	}
 
-	restServer := rest.NewServer(cfg.RestSvrCfg, certMgr, keccakMgr, log, metrics)
-	router := mux.NewRouter()
-	restServer.RegisterRoutes(router)
-	if cfg.StoreBuilderConfig.MemstoreEnabled {
-		memconfig.NewHandlerHTTP(log, cfg.StoreBuilderConfig.MemstoreConfig).RegisterMemstoreConfigHandlers(router)
+	// Construct the compatibility config for the rest and arb servers. This could not be done while reading configs
+	// as ChainID is fetched from the ethClient afterwards.
+	compatibilityCfg, err := common.NewCompatibilityConfig(
```
samlaf, line 94: We have a hardcoded mapping `EigenDANetworksFromChainID(chainID.String())`. Couldnt we have used a similar reverse mapping to get the chainID from the eigendaNetwork?

api/proxy/common/eigenda_network.go
```diff
@@ -104,3 +112,43 @@ func EigenDANetworkFromString(inputString string) (EigenDANetwork, error) {
 			inputString, strings.Join(allowedNetworks, ", "))
 	}
 }
+
+// BuildEthClient creates an Ethereum client using the provided RPC URL and, if set, validates that the chain ID
+// matches the expected EigenDA network. It returns an ethClient, it's ChainID, and an error.
+func BuildEthClient(ctx context.Context, log logging.Logger, ethRpcUrl string,
```
samlaf, line 118: this feels like the wrong file to host this function.
Guessing you couldnt keep it in the storage_manager_builder because of circular dependency issues?
Does the monorepo not have another ethClient builder we could reuse? And then we can add our few checks and query for chainID on top?

No need to make a huge refactor but at least I feel like this should be moved to some other file, perhaps under api/proxy/common/eth_client.go or whatever until we find it a better home.

api/proxy/cmd/server/entrypoint.go
iquidus: We could but then it would only work for the included networks and if the network flag is used. With this approach it works for any network, with or without the network flag.

PR #2223, opened 2025-10-25
feat(controller): adds local key signing to controller as alternative to KMS

pschork:

OCI KMS does not currently support secp256k curve. This change allows controller to sign dispersal requests using local signer.

Note: OCI has a Dedicated HSM solution with secp256k support but there are significant cost and operational tradeoffs that need further exploration before deciding to use it.

Notes:
- Either `CONTROLLER_DISPERSER_KMS_KEY_ID` or `CONTROLLER_DISPERSER_PRIVATE_KEY` must be specified
- Modified `SignStoreChunksRequest` to use KMS or local signing based on available credentials
- Uses `crypto.Sign()` for local signing, maintains existing KMS signing flow

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/dispersal_request_signer.go
+++ b/api/clients/v2/dispersal_request_signer.go
@@ -145,7 +145,7 @@ func NewLocalDispersalRequestSigner(
 }
 
 func (s *kmsRequestSigner) SignStoreChunksRequest(
-	ctx context.Context, 
+	ctx context.Context,
 	request *grpc.StoreChunksRequest,
 ) ([]byte, error) {
 	hash, err := hashing.HashStoreChunksRequest(request)
--- a/api/clients/v2/dispersal_request_signer_test.go
+++ b/api/clients/v2/dispersal_request_signer_test.go
@@ -290,6 +290,203 @@ func TestLocalSignerWithBothKMSAndPrivateKey(t *testing.T) {
 	require.Error(t, err, "should fail when both KeyID and PrivateKey are specified")
 }
 
+func TestNewKMSDispersalRequestSignerDirect(t *testing.T) {
+	ctx := t.Context()
+	_ = setupLocalStack(t)
+
+	keyManager := kms.New(kms.Options{
+		Region:       region,
+		BaseEndpoint: aws.String(localstackHost),
+	})
+
+	// Create a test KMS key
+	keyID, _ := createTestKMSKey(t, ctx, keyManager)
+
+	// Test direct KMS factory function
+	signer, err := NewKMSDispersalRequestSigner(ctx, DispersalRequestSignerConfig{
+		Region:   region,
+		Endpoint: localstackHost,
+		KeyID:    keyID,
+	})
+	require.NoError(t, err, "failed to create KMS signer directly")
+	require.NotNil(t, signer, "signer should not be nil")
+}
+
+func TestNewLocalDispersalRequestSignerDirect(t *testing.T) {
+	// Generate a private key for testing
+	privateKey, err := crypto.GenerateKey()
+	require.NoError(t, err, "failed to generate test private key")
+	privateKeyHex := fmt.Sprintf("%x", crypto.FromECDSA(privateKey))
+
+	// Test direct local factory function
+	signer, err := NewLocalDispersalRequestSigner(DispersalRequestSignerConfig{
+		PrivateKey: privateKeyHex,
+	})
+	require.NoError(t, err, "failed to create local signer directly")
+	require.NotNil(t, signer, "signer should not be nil")
+}
+
+func TestNewKMSDispersalRequestSignerErrors(t *testing.T) {
+	ctx := t.Context()
+
+	tests := []struct {
+		name        string
+		config      DispersalRequestSignerConfig
+		expectError bool
+		errorMsg    string
+	}{
+		{
+			name: "invalid_region_empty",
+			config: DispersalRequestSignerConfig{
+				KeyID:    "test-key",
+				Region:   "",
+				Endpoint: localstackHost,
+			},
+			expectError: true,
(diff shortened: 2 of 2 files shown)


api/clients/v2/dispersal_request_signer.go
```diff
@@ -23,9 +24,11 @@ type DispersalRequestSigner interface {
 }
 
 type DispersalRequestSignerConfig struct {
-	// KeyID is the AWS KMS key identifier used for signing requests. Required.
+	// KeyID is the AWS KMS key identifier used for signing requests. Optional if PrivateKey is provided.
 	KeyID string
-	// Region is the AWS region where the KMS key is located (e.g., "us-east-1"). Required.
+	// PrivateKey is a hex-encoded private key for local signing (without 0x prefix). Optional if KeyID is provided.
+	PrivateKey string
```
samlaf, line 30: imo we shouldnt do this. we should have an interface and 2 separzte implementations for it: one using privateKey, another one using KMS. This kind of code is super bug prone and hard to maintain.

api/clients/v2/dispersal_request_signer.go
```diff
@@ -106,10 +127,19 @@ func (s *requestSigner) SignStoreChunksRequest(ctx context.Context, request *grp
...
+		// Use KMS signing
+		signature, err := aws2.SignKMS(ctx, s.kmsClient, s.keyID, s.publicKey, hash)
+		if err != nil {
+			return nil, fmt.Errorf("failed to sign request: %w", err)
+		}
+		return signature, nil
+	}
+
+	// Use local private key signing
+	signature, err := crypto.Sign(hash, s.privateKey)
```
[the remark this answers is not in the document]
pschork: Verified

disperser/cmd/controller/flags/flags.go
```diff
@@ -203,6 +203,12 @@ var (
 		Required: false,
 		EnvVar:   common.PrefixEnvVar(envVarPrefix, "DISPERSER_KMS_KEY_ID"),
 	}
+	DisperserPrivateKeyFlag = cli.StringFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "disperser-private-key"),
+		Usage:    "Private key for signing disperser requests (hex format without 0x prefix, alternative to KMS)",
```
samlaf, line 208: would be good to update the DisperserKMSKeyIDFlag as well to mention that its optional and mutually exclusive with this PrivateKeyFlag

api/clients/v2/dispersal_request_signer.go
```diff
@@ -43,24 +46,36 @@ func DefaultDispersalRequestSignerConfig() DispersalRequestSignerConfig {
...
 	}
 
 	return nil
 }
 
-var _ DispersalRequestSigner = &requestSigner{}
-
-type requestSigner struct {
+// kmsRequestSigner implements DispersalRequestSigner using AWS KMS.
+type kmsRequestSigner struct {
```
samlaf, line 63:
```suggestion
type KMSRequestSigner struct {
```

samlaf: make public otherwise people can't embed this struct in other structs

api/clients/v2/dispersal_request_signer.go
```diff
@@ -43,24 +46,36 @@ func DefaultDispersalRequestSignerConfig() DispersalRequestSignerConfig {
...
 	keyID     string
 	publicKey *ecdsa.PublicKey
 	kmsClient *kms.Client
 }
 
+var _ DispersalRequestSigner = &kmsRequestSigner{}
+
+// localRequestSigner implements DispersalRequestSigner using a local private key.
+type localRequestSigner struct {
```
samlaf, line 72:
```suggestion
type LocalRequestSigner struct {
```

samlaf: make public otherwise people can't embed this struct in other structs

api/clients/v2/dispersal_request_signer.go
```diff
@@ -43,24 +46,36 @@ func DefaultDispersalRequestSignerConfig() DispersalRequestSignerConfig {
...
+	privateKey *ecdsa.PrivateKey
+	publicKey  *ecdsa.PublicKey
+}
+
+var _ DispersalRequestSigner = &localRequestSigner{}
+
 // NewDispersalRequestSigner creates a new DispersalRequestSigner.
 func NewDispersalRequestSigner(
 	ctx context.Context,
```
samlaf, line 81: My Personal preference is to not have functions like this. IMO constructors should never return an interface, but should instead return structs (see https://bryanftan.medium.com/accept-interfaces-return-structs-in-go-d4cab29a301b). This gives most information/flexibility to callers.

Also main thing is I don't like these polymorphic constructors. Easy config wrong and end up using kms signers inside tests for example. Prefer to be explicit at all call sites for which struct you are building.

api/clients/v2/dispersal_request_signer.go
```diff
@@ -93,14 +122,32 @@ func NewDispersalRequestSigner(
...
 		publicKey: key,
 		kmsClient: kmsClient,
 	}, nil
 }
 
-func (s *requestSigner) SignStoreChunksRequest(ctx context.Context, request *grpc.StoreChunksRequest) ([]byte, error) {
+// NewLocalDispersalRequestSigner creates a new local private key-based DispersalRequestSigner.
+func NewLocalDispersalRequestSigner(
+	config DispersalRequestSignerConfig,
+) (DispersalRequestSigner, error) {
```
samlaf, line 135: should return a *localRequestSigner, not interface

PR #2229, opened 2025-10-27
feat: add several utilities for interacting with onchain data

cody-littley:

Split from auto-ejector PR. Adds new micro-utilities for reading/caching onchain information.

### DIFF
--- a/common/oci/object_storage.go
+++ b/common/oci/object_storage.go
@@ -256,10 +256,10 @@ func (c *ociClient) FragmentedUploadObject(
 
 	for _, fragment := range fragments {
 		fragmentCapture := fragment
-		c.concurrencyLimiter <- struct{}{}
+		<-c.concurrencyLimiter
 		go func() {
 			defer func() {
-				<-c.concurrencyLimiter
+				c.concurrencyLimiter <- struct{}{}
 			}()
 			c.fragmentedWriteTask(ctx, resultChannel, fragmentCapture, bucket)
 		}()
@@ -320,10 +320,10 @@ func (c *ociClient) FragmentedDownloadObject(
 	for i, fragmentKey := range fragmentKeys {
 		boundFragmentKey := fragmentKey
 		boundI := i
-		c.concurrencyLimiter <- struct{}{}
+		<-c.concurrencyLimiter
 		go func() {
 			defer func() {
-				<-c.concurrencyLimiter
+				c.concurrencyLimiter <- struct{}{}
 			}()
 			c.readTask(ctx, resultChannel, bucket, boundFragmentKey, boundI)
 		}()
--- a/common/oci/object_storage_test.go
+++ b/common/oci/object_storage_test.go
@@ -1415,3 +1415,57 @@ func TestObjectStorageConfig_WorkerCalculations(t *testing.T) {
 		})
 	}
 }
+
+// This test verifies the worker semaphore pattern does not deadlock
+func TestOCIClient_ConcurrencyLimiter(t *testing.T) {
+	t.Run("Semaphore pattern works with correct token operations", func(t *testing.T) {
+		limiter := make(chan struct{}, 2)
+
+		// Pre-fill with tokens (semaphore pattern)
+		limiter <- struct{}{}
+		limiter <- struct{}{}
+
+		// Test that we can acquire tokens (receive operation)
+		<-limiter // Should not block
+		<-limiter // Should not block
+
+		// Channel should now be empty, so sending should work (release operation)
+		limiter <- struct{}{} // Should not block
+		limiter <- struct{}{} // Should not block
+
+		// Verify channel is full again
+		assert.Equal(t, 2, len(limiter))
+	})
+
+	t.Run("Check concurrency limiter initialization", func(t *testing.T) {
+		// We can't easily test the actual fragmented methods due to OCI SDK dependencies,
+		// but we can test the concurrency limiter initialization and basic pattern
+
+		cfg := ObjectStorageConfig{
+			FragmentParallelismConstant: 3,
+		}
+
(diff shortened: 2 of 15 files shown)


core/eth/validator_stake_lookup.go
```diff
@@ -0,0 +1,250 @@
...
+	if err != nil {
+		return 0.0, fmt.Errorf("failed to get total quorum stake: %w", err)
+	}
+
+	if totalStake.Cmp(big.NewInt(0)) == 0 {
+		return 0.0, nil // Avoid division by zero; if total stake is zero, return 0.0 fraction.
+	}
+
+	fraction := new(big.Rat).SetFrac(validatorStake, totalStake)
+	floatFraction, _ := fraction.Float64()
```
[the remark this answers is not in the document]
cody-littley: This is not an error return value. Go home copilot, you're drunk.

core/eth/validator_stake_lookup.go
```diff
@@ -0,0 +1,250 @@
...
```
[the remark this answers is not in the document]
cody-littley: no-bugs-bunny

core/eth/validator_quorum_lookup.go
```diff
@@ -0,0 +1,138 @@
...
+
+	// This method returns a bitmap as a big.Int.
+	bigIntBitmap, err := v.registryCoordinator.GetCurrentQuorumBitmap(opts, validatorID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get quorum bitmap: %w", err)
+	}
+
+	quorumIDs := make([]core.QuorumID, 0)
+
+	for i := 0; i <= 192; i++ {
```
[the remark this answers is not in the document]
cody-littley: Added explanation. Constant doesn't improve readability.

core/eth/validator_quorum_lookup.go
```diff
@@ -0,0 +1,140 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create registry coordinator contract instance: %w", err)
+	}
+
+	return &validatorQuorumLookup{
+		registryCoordinator: registryCoordinator,
+	}, nil
+}
+
+// TODO test this by hand to verify behavior.
```
litt3: TODO

core/eth/validator_quorum_lookup.go
```diff
@@ -0,0 +1,140 @@
...
+	bigIntBitmap, err := v.registryCoordinator.GetCurrentQuorumBitmap(opts, validatorID)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get quorum bitmap: %w", err)
+	}
+
+	quorumIDs := make([]core.QuorumID, 0)
+
+	// An implementation detail of the solidity: the number returned by the contract is a bitmap backed by a
+	// uint192, so we need to check each bit up to 192. If we check for higher bits, we will panic.
+	for i := 0; i <= 192; i++ {
```
litt3:
```suggestion
	for i := 0; i < 192; i++ {
```

Am I missing something?

core/eth/validator_quorum_lookup.go
```diff
@@ -0,0 +1,140 @@
...
+
+// TODO test this by hand to verify behavior.
+
+func (v *validatorQuorumLookup) GetQuorumsForValidator(
+	ctx context.Context,
+	validatorID core.OperatorID,
+	referenceBlockNumber uint64,
+) ([]core.QuorumID, error) {
+
+	blockNumber := big.NewInt(int64(referenceBlockNumber))
```
litt3: IIUC you can do something like this to avoid having to pass the uint through an int

```
big.NewInt(0).SetUint64(referenceBlockNumber)
```

Probably doesn't matter in practice

core/eth/validator_quorum_lookup.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	quorums, err := c.base.GetQuorumsForValidator(ctx, validatorAddress, referenceBlockNumber)
+	if err != nil {
+		return nil, fmt.Errorf("failed to get quorums for validator: %w", err)
+	}
+
+	c.cache.Add(key, quorums)
+
+	return quorums, nil
```
litt3: For all of your caches, you're returning mutable references to the actual values within the cache.

Not sure if it's worth making copies, or potentially mentioning that the caller shouldn't modify the value.

core/eth/quorum_scanner.go
```diff
@@ -66,3 +65,45 @@ func (q *quorumScanner) GetQuorums(ctx context.Context, referenceBlockNumber uin
 
 	return quorums, nil
```
litt3, line 66: It was preexisting, but the above loop will fail if we ever had 256 quorums, since the increment will wrap back around to 0. Not sure if we care to fix this

```
	for i := uint8(0); i < quorumCount; i++ {
		quorums[i] = i
	}
```

core/eth/validator_quorum_lookup.go
cody-littley: testing completed last week, removed TODO

core/eth/validator_quorum_lookup.go
cody-littley: I think you're right. Not sure why we didn't panic... but this API is so janky I'm willing to take the win and move on. Fixed.

core/eth/validator_quorum_lookup.go
cody-littley: done

core/eth/quorum_scanner.go
cody-littley: I'd prefer not to take any steps to rectify `uint8` quorum count problems for the time being. The whole architecture of how we handle quorums is probably going to change when we move to use operator sets.

PR #2231, opened 2025-10-27
feat: convert chunk requests to use range query

cody-littley:

Instead of requesting chunk data by index, instead request chunk data by range. Prereq for significant performance enchantments on the relay.

### DIFF
--- a/node/grpc/server_v2_test.go
+++ b/node/grpc/server_v2_test.go
@@ -229,14 +229,24 @@ func TestV2StoreChunksSuccess(t *testing.T) {
 	require.NoError(t, err)
 	bundles20Bytes, err := bundles[2][0].Serialize()
 	require.NoError(t, err)
-	c.relayClient.On("GetChunksByIndex", mock.Anything, v2.RelayKey(0), mock.Anything).Return([][]byte{bundles00Bytes, bundles20Bytes}, nil).Run(func(args mock.Arguments) {
-		requests := args.Get(2).([]*relay.ChunkRequestByIndex)
+	c.relayClient.On(
+		"GetChunksByRange",
+		mock.Anything,
+		v2.RelayKey(0),
+		mock.Anything,
+	).Return([][]byte{bundles00Bytes, bundles20Bytes}, nil).Run(func(args mock.Arguments) {
+		requests := args.Get(2).([]*relay.ChunkRequestByRange)
 		require.Len(t, requests, 2)
 		require.Equal(t, blobKeys[0], requests[0].BlobKey)
 		require.Equal(t, blobKeys[2], requests[1].BlobKey)
 	})
-	c.relayClient.On("GetChunksByIndex", mock.Anything, v2.RelayKey(1), mock.Anything).Return([][]byte{bundles10Bytes}, nil).Run(func(args mock.Arguments) {
-		requests := args.Get(2).([]*relay.ChunkRequestByIndex)
+	c.relayClient.On(
+		"GetChunksByRange",
+		mock.Anything,
+		v2.RelayKey(1),
+		mock.Anything,
+	).Return([][]byte{bundles10Bytes}, nil).Run(func(args mock.Arguments) {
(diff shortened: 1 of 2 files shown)


node/node_v2.go
```diff
@@ -104,6 +102,44 @@ func (n *Node) DetermineChunkLocations(
...
+		return requests
+	}
+
+	startIndex := indices[0]
+	for i := 1; i < len(indices); i++ {
+		if indices[i] != indices[i-1]+1 {
+			// break in continuity, create a request for the previous range
+			request := &relay.ChunkRequestByRange{
+				Start: startIndex,       // inclusive
+				End:   indices[i-1] + 1, // exclusive
```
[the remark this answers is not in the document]
cody-littley: fixed

node/node_v2.go
```diff
@@ -104,6 +102,44 @@ func (n *Node) DetermineChunkLocations(
 	return downloadSizeInBytes, relayRequests, nil
 }
 
+// TODO write tests for this method
+
+// Given a list of chunk indices we want to download, create a list of relay requests by range.
+// Although indices may not be contiguous, it is safe to assume that they will be "mostly contiguous".
+// In practice, we should expect to see at most one continuous range of indices per quorum.
+//
+// Eventually, the assignment logic aught to be refactored to return ranges of chunks instead of individual
```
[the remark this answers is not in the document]
cody-littley: fixed

node/node_v2.go
```diff
@@ -104,6 +102,43 @@ func (n *Node) DetermineChunkLocations(
 	return downloadSizeInBytes, relayRequests, nil
 }
 
+// Given a list of chunk indices we want to download, create a list of relay requests by range.
+// Although indices may not be contiguous, it is safe to assume that they will be "mostly contiguous".
+// In practice, we should expect to see at most one continuous range of indices per quorum.
+//
+// Eventually, the assignment logic ought to be refactored to return ranges of chunks instead of individual
```
litt3, line 116: "instead of individual of individual indices"

repeat words

node/node_v2.go
```diff
@@ -84,16 +84,14 @@ func (n *Node) DetermineChunkLocations(
...
 				metadata:      make([]*requestMetadata, 0),
 			}
 			relayRequests[relayKey] = req
 		}
 		// Chunks from one blob are requested to the same relay
-		req.chunkRequests = append(req.chunkRequests, &relay.ChunkRequestByIndex{
-			BlobKey: blobKey,
-			Indices: assgn.Indices,
-		})
+		rangeRequests := convertIndicesToRangeRequests(blobKey, assgn.Indices)
```
litt3, line 93: I'm assuming the only tests for non-contiguous indices are the unit tests? Claude identified this bug:

in `DownloadChunksFromRelays`, there is an assertion that the number of requests matches the number of metadatas. But it looks look non-contiguous indices would cause these counts to mismatch, since only a single metadata object is added here, even if there are multiple requests

```
		if len(resp.bundles) != len(resp.metadata) {
			return nil, nil,
				fmt.Errorf("number of bundles and metadata do not match (%d != %d)",
					len(resp.bundles), len(resp.metadata))
		}
```

I haven't dug in super deep to 100% confirm, but this looks like a legitimate problem to me

node/node_v2.go
```diff
@@ -104,6 +102,43 @@ func (n *Node) DetermineChunkLocations(
 	return downloadSizeInBytes, relayRequests, nil
 }
 
+// Given a list of chunk indices we want to download, create a list of relay requests by range.
```
litt3, line 108: Consider noting that though indices might not be contiguous, they must be sorted

node/node_v2.go
cody-littley: fixed

node/node_v2.go
cody-littley: fixed

node/node_v2.go
cody-littley: added comment

pschork (approved):

PR #2212, opened 2025-10-22
perf: optimize kzg multi-reveal by using smaller toeplitz matrix

bxue-l2:
## Why are these changes needed?

  Optimizes multi-reveal KZG proof generation by allowing the use of smaller Toeplitz matrices when the actual blob size is smaller than the encoding parameters would normally require.

  Key Changes:
  1. New ProvingParams struct - Separates Toeplitz matrix dimensions from encoding parameters, enabling more flexible proof generation based on actual blob size
  2. Optimized matrix multiplication - Uses ToeplitzMatrixLength = blobLength / ChunkLength instead of always using NumChunks from encoding params, reducing computation for smaller blobs
  3. SRS table caching - Refactored to key on ProvingParams instead of EncodingParams. With this optimization, we no longer can infer a parametrizedProver from a SRSTable

##  Performance Impact:

Screenshot 2025-10-22 at 5 16 41 PM
Both MSM and FFT1 shows 8 times speed up

I found there is no implementation for icicle for v2. But the idea is really similar. I can update if we want it

Close https://github.com/Layr-Labs/eigenda/pull/2160#issue-3516210449, but it is interesting to see, with FFT optimization how far we have got from 9 sec to 0.75 sec

## Theory behind the change
Screenshot 2025-10-26 at 11 28 04 PM

The previous implementation generates the `kzgProof` by taking FFT on the result of the matrix multiplication between A and b. i.e FFT(A*b), where `A` is size 8192 square matrix. 

But you can see, almost all of entries in A is 0 except A' where all the data from blob resides. The idea behind the optimization is to take matrix multiplication between A' and b', and then pad 0 at the end of the matrix. Then taking the FFT at the end. Mathematically they are equivalent. 

One nice observation from it is that we are erasure coding the KZG proof, such that KZG proofs themselves are recoverable from sufficient chunks.

## Note to infra
With this change, SRSTable is again 2 dimensional: ToeplitzMatrixLen and ChunkLen. 

ToeplitzMatrixLen: can be 2^{0,1,2,3,4,5,6,7,8,9,10}
ChunkLen: can be 2^{0,1,2,3,4,5,6,7,8,9}

More details see
Screenshot 2025-10-22 at 5 25 49 PM

## Checks

- [ ] I've made sure the tests are passing. Note that there might be a few flaky tests, in that case, please comment that they are not relevant.
- [ ] I've checked the new test coverage and the coverage percentage didn't drop.
- Testing Strategy
   - [ ] Unit tests
[...]

PR #2238, opened 2025-10-28
perf: optimize kzg multi-reveal by using smaller toeplitz matrix

bxue-l2:

Closing https://github.com/Layr-Labs/eigenda/pull/2212

  Optimizes multi-reveal KZG proof generation by allowing the use of smaller Toeplitz matrices when the actual blob size is smaller than the encoding parameters would normally require.

  Key Changes:
  1. New ProvingParams struct - Separates Toeplitz matrix dimensions from encoding parameters, enabling more flexible proof generation based on actual blob size
  2. Optimized matrix multiplication - Uses ToeplitzMatrixLength = blobLength / ChunkLength instead of always using NumChunks from encoding params, reducing computation for smaller blobs
  3. SRS table caching - Refactored to key on ProvingParams instead of EncodingParams. With this optimization, we no longer can infer a parametrizedProver from a SRSTable

##  Performance Impact:

Screenshot 2025-10-22 at 5 16 41 PM
Both MSM and FFT1 shows 8 times speed up

I found there is no implementation for icicle for v2. But the idea is really similar. I can update if we want it

Close https://github.com/Layr-Labs/eigenda/pull/2160#issue-3516210449, but it is interesting to see, with FFT optimization how far we have got from 9 sec to 0.75 sec

## Theory behind the change
Screenshot 2025-10-26 at 11 28 04 PM

The previous implementation generates the `kzgProof` by taking FFT on the result of the matrix multiplication between A and b. i.e FFT(A*b), where `A` is size 8192 square matrix. 

But you can see, almost all of entries in A is 0 except A' where all the data from blob resides. The idea behind the optimization is to take matrix multiplication between A' and b', and then pad 0 at the end of the matrix. Then taking the FFT at the end. Mathematically they are equivalent. 

One nice observation from it is that we are erasure coding the KZG proof, such that KZG proofs themselves are recoverable from sufficient chunks.

## Note to infra
With this change, SRSTable is again 2 dimensional: ToeplitzMatrixLen and ChunkLen. 

ToeplitzMatrixLen: can be 2^{0,1,2,3,4,5,6,7,8,9,10}
ChunkLen: can be 2^{0,1,2,3,4,5,6,7,8,9}

More details see
Screenshot 2025-10-22 at 5 25 49 PM

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
+++ b/encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
@@ -276,44 +276,63 @@ func (c *KzgMultiProofBackend) twoEcnttOnDevice(
 	toeplitzMatrixLen int,
 ) ([]bn254.G1Affine, time.Time, error) {
 	var p iciclebn254.Projective
-	var numChunksProjectivePointsOnDevice core.DeviceSlice
+	// we only allocate one large gpu memory for all operation, so it has to be large enough to cover all cases
+	// including the first and the second ECNTT
+	var bufferProjectivePointsOnDevice core.DeviceSlice
+	var secondECNTTDeviceSlice core.DeviceSlice
+	numPointsOnDevice := numChunks
 
-	_, err := numChunksProjectivePointsOnDevice.Malloc(p.Size(), numChunks)
+	// the size is twice because of the FFT trick on toeplitz matrix
+	firstECNTTLen := toeplitzMatrixLen * 2
+
+	// when first ecntt is larger than numChunk, we must allocate enough memory
+	// it happens if numChunks is equal of less than toeplitzMatrixLen
+	if numChunks < firstECNTTLen {
+		numPointsOnDevice = firstECNTTLen
+	}
+	_, err := bufferProjectivePointsOnDevice.Malloc(p.Size(), numPointsOnDevice)
 	if err != runtime.Success {
 		return nil, time.Time{}, fmt.Errorf("allocating bytes on device failed: %v", err.AsString())
 	}
 
-	// the size is twice because of the FFT trick on toeplitz matrix
-	firstECNTTLen := toeplitzMatrixLen * 2
-	// Device memory for first ecntt
-	firstECNTTDeviceSlice := numChunksProjectivePointsOnDevice.RangeTo(firstECNTTLen, false)
+	// specify device memory sluce for first ecntt
+	firstECNTTDeviceSlice := bufferProjectivePointsOnDevice.RangeTo(firstECNTTLen, false)
 	err = ecntt.ECNtt(batchPoints, core.KInverse, &c.NttCfg, firstECNTTDeviceSlice)
 	if err != runtime.Success {
 		return nil, time.Time{}, fmt.Errorf("inverse ecntt failed: %v", err.AsString())
 	}
 
-	// now only keep the toeplitzMatrixLen elements as they are, set the rest to zero.
-	// Zeros are the infinity points for G1Projective points
-	// unit in the Range function is measured by element size
-	infinityPointsOnDevice := numChunksProjectivePointsOnDevice.Range(toeplitzMatrixLen, numChunks, false)
-	infinityProjectivePoints := make([]iciclebn254.Projective, numChunks-toeplitzMatrixLen)
-	infinityPointsHost := core.HostSliceFromElements[iciclebn254.Projective](
-		infinityProjectivePoints[:numChunks-toeplitzMatrixLen],
-	)
-	// copy to device, but don't allocate memory
-	infinityPointsHost.CopyToDevice(&infinityPointsOnDevice, false)
-
-	// create output buffer
 	proofsBatchHost := make(core.HostSlice[iciclebn254.Projective], numChunks)
 
+	// if numChunk is smaller or equal to toeplitzMatrixLen, there is no need to set points to infinity
+	// otherwise set all points to infinity
+	if numChunks > toeplitzMatrixLen {
+		// now only keep the toeplitzMatrixLen elements as they are, set the rest to zero.
+		// Zeros are the infinity points for G1Projective points
+		// unit in the Range function is measured by element size
+		infinityPointsOnDevice := bufferProjectivePointsOnDevice.Range(toeplitzMatrixLen, numChunks, false)
+		infinityProjectivePoints := make([]iciclebn254.Projective, numChunks-toeplitzMatrixLen)
+		// explicitly sets all value to zero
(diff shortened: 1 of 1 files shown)


encoding/v2/bench/benchmark_eigenda_test.go
```diff
@@ -212,11 +212,15 @@ func benchmarkMultiproofGeneration(b *testing.B, encodingConfig encoding.Config)
 				NumChunks:   8192,                           // blob_version=0
 				ChunkLength: max(1, rsExtendedBlobFrs/8192), // chosen such that numChunks*ChunkLength=rsExtendedBlobFrs
 			}
-			parametrizedProver, err := p.GetKzgProver(params)
+			provingParams := prover.ProvingParams{
+				BlobLength:  rsExtendedBlobFrs,
```
samlaf: probably wanted to use `1 << blobPowerBytes - 5` here afaiu? because rsExtendedBlob is not blobLength. its blobLength * 8

encoding/v2/kzg/prover/parametrized_prover.go
```diff
@@ -20,9 +20,11 @@ type ParametrizedProver struct {
...
 }
 
-func (g *ParametrizedProver) GetProofs(inputFr []fr.Element) ([]encoding.Proof, error) {
-	// pad inputFr to NumEvaluations(), which encodes the RS redundancy
-	paddedCoeffs := make([]fr.Element, g.encodingParams.NumEvaluations())
+// The inputFr has not been padded to the next power of 2 field of elements. But ComputeMultiFrameProofV2
+// requires it.
+func (g *ParametrizedProver) GetProofs(inputFr []fr.Element, provingParams ProvingParams) ([]encoding.Proof, error) {
+	// pad inputFr to BlobLength if it is not power of 2, which encodes the RS redundancy
+	paddedCoeffs := make([]fr.Element, provingParams.BlobLength)
```
samlaf: you don't need provingParams here. You only use it to get the BlobLength, which is just `nextPowerOf2(len(inputFr))`.

Seems like provingParams are really only used to index the SRSTables? How about we rename that and only use them for that purpose?

encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
```diff
@@ -288,26 +262,60 @@ func (c *KzgMultiProofBackend) msmBatchOnDevice(frs []fr.Element, g1Points []ici
...
+	numChunks int,
+	toeplitzMatrixLen int,
+) ([]bn254.G1Affine, time.Time, error) {
 	var p iciclebn254.Projective
-	var out core.DeviceSlice
+	var deviceWithNumChunkElement core.DeviceSlice
+	var output core.DeviceSlice
 
-	output, err := out.Malloc(p.Size(), totalSize)
+	_, err := deviceWithNumChunkElement.Malloc(p.Size(), numChunks)
```
samlaf: can we rename this to `numChunkProjectiveFrsOnDevice` or something? This is not a device, its elements ON the device

encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
```diff
@@ -288,26 +262,60 @@ func (c *KzgMultiProofBackend) msmBatchOnDevice(frs []fr.Element, g1Points []ici
...
+		return nil, time.Time{}, fmt.Errorf("forward ecntt failed: %v", err.AsString())
+	}
+
+	// free intermediate GPU memory
+	deviceWithNumChunkElement.Free()
+
+	proofsBatchHost := make(core.HostSlice[iciclebn254.Projective], numChunks)
+	proofsBatchHost.CopyFromDevice(&output)
+	// free GPU memory for proof
+	output.Free()
```
samlaf: isnt there a config in NttCfg to specify that we want the output on the host instead of device?

encoding/v2/kzg/prover/parametrized_prover.go
samlaf: As a more general comment, I really don't like these parametrizedProver. Feel like we should only have parametrized resources instead (SRSTables, fft domain). This is conflating code with config/data, just like we do in CertVerifier.

Anyways we can do this refactor in a future PR, might be not so trivial.

bxue-l2: that is right

encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
bxue-l2: device on device

bxue-l2: but its not Fr

encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
bxue-l2: so you can specify the output on host, nice

bxue-l2: what is not something you can specify

encoding/v2/kzg/prover/parametrized_prover.go
bxue-l2: 49bc94e9d513a69f1925f4634db2553666a02bfb

encoding/v2/bench/benchmark_eigenda_test.go
bxue-l2: 49bc94e9d513a69f1925f4634db2553666a02bfb

encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
bxue-l2: 49bc94e9d513a69f1925f4634db2553666a02bfb

encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
bxue-l2: 49bc94e9d513a69f1925f4634db2553666a02bfb

samlaf: you don't seem to have changed the output config? or is it output on host by default?

bxue-l2: the icicle interface accept HostOrDevice, so they must be able to figure out
https://github.com/ingonyama-zk/icicle/blob/5ef9ed1634c679d8a23d3b5cd338e36354998d16/wrappers/golang/curves/bn254/ecntt/ecntt.go#L12

PR #2239, opened 2025-10-28
chore: bump foundry

0xClandestine:
## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Integration tests

### DIFF
--- a/contracts/test/unit/EigenDABlobUtilsV1Unit.t.sol
+++ b/contracts/test/unit/EigenDABlobUtilsV1Unit.t.sol
@@ -59,9 +59,10 @@ contract EigenDABlobUtilsV1Unit is MockEigenDADeployer {
         blobVerificationProof.inclusionProof = abi.encodePacked(keccak256(firstBlobHash));
         blobVerificationProof.blobIndex = 1;
         blobVerificationProof.quorumIndices = new bytes(batchHeader.quorumNumbers.length);
+
+        if (batchHeader.quorumNumbers.length > type(uint8).max) revert(); // Sanity check.
+        // forge-lint: disable-next-item(unsafe-typecast)
         for (uint256 i = 0; i < batchHeader.quorumNumbers.length; i++) {
-            // forge-lint: disable-next-item(unsafe-typecast)
-            // If batchHeader.quorumNumbers.length is greater than 255, overflow will occur.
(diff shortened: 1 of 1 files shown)


mise.toml
```diff
@@ -38,9 +38,9 @@ protoc-gen-go-grpc = "v1.3.0"
 "go:github.com/segmentio/golines" = "0.12.0"
 
 # Forge Dependencies
-forge = "v1.0.0"
-cast = "v1.0.0"
-anvil = "v1.0.0"
```
0xClandestine, line 43: Main change

pakim249CAL (dismissed):

contracts/test/unit/EigenDABlobUtilsV1Unit.t.sol
```diff
@@ -60,6 +60,8 @@ contract EigenDABlobUtilsV1Unit is MockEigenDADeployer {
         blobVerificationProof.blobIndex = 1;
         blobVerificationProof.quorumIndices = new bytes(batchHeader.quorumNumbers.length);
         for (uint256 i = 0; i < batchHeader.quorumNumbers.length; i++) {
+            // forge-lint: disable-next-item(unsafe-typecast)
+            // If batchHeader.quorumNumbers.length is greater than 255, overflow will occur.
```
samlaf: shouldnt we explicitly check this? offchain is supposed to check this, but what if they dont. Oh nvm this is V1 code we dgaf about this, but if there is the equivalent in V2 then we should be careful.

samlaf (dismissed):
Afaiu the commits are:
1. bump the version to 1.4.3
2. fix a few actual lint issues
3. just run forge fmt

contracts/test/unit/EigenDABlobUtilsV1Unit.t.sol
0xClandestine: can add a check, think i opted not to since it was a unit test

0xClandestine: [06de28e](https://github.com/Layr-Labs/eigenda/pull/2239/commits/06de28e8c2975b7b05988cec91f6d28ffb19074b)

samlaf (approved):

pakim249CAL (approved):

PR #2220, opened 2025-10-24
perf(proofs-gpu): 400ms speedup for multiproof generation

samlaf:
Applies the same trick as in https://github.com/Layr-Labs/eigenda/pull/2214, but for proof generation instead of chunk generation. The improvements are much smaller because we were only needlessly converting 1 slice here. That PR should merge first. Features are orthogonal but have reused some of the benchmark code improvements in here.

Before fix:
```
go test -benchmem -run=^$ -tags icicle -bench ^BenchmarkMultiproofGenerationIcicle$ github.com/Layr-Labs/eigenda/encoding/v2/bench -v

2025/10/24 22:57:14 maxprocs: Leaving GOMAXPROCS=4: CPU quota undefined
goos: linux
goarch: amd64
pkg: github.com/Layr-Labs/eigenda/encoding/v2/bench
cpu: AMD EPYC 7R13 Processor
BenchmarkMultiproofGenerationIcicle
Oct 24 22:59:53.159 INF icicle/device_setup.go:118 CUDA device available, setting device
BenchmarkMultiproofGenerationIcicle/Multiproof_size_2^17_bytes
Oct 24 22:59:56.365 INF icicle/multiframe_proof.go:167 Multiproof Time Decomp total=106.80322ms preproc=8.765276ms msm=40.773015ms fft1=32.145063ms fft2=16.506604ms
BenchmarkMultiproofGenerationIcicle/Multiproof_size_2^17_bytes-4         	      10	 100186766 ns/op	13634608 B/op	   24648 allocs/op
BenchmarkMultiproofGenerationIcicle/Multiproof_size_2^20_bytes
Oct 24 22:59:57.528 INF icicle/multiframe_proof.go:167 Multiproof Time Decomp total=265.627822ms preproc=70.324366ms msm=138.10585ms fft1=31.626728ms fft2=17.203653ms
BenchmarkMultiproofGenerationIcicle/Multiproof_size_2^20_bytes-4         	       4	 265274404 ns/op	94375214 B/op	   24733 allocs/op
BenchmarkMultiproofGenerationIcicle/Multiproof_size_2^21_bytes
Oct 24 22:59:58.724 INF icicle/multiframe_proof.go:167 Multiproof Time Decomp total=397.99549ms preproc=146.924987ms msm=194.050231ms fft1=31.563197ms fft2=16.982827ms
BenchmarkMultiproofGenerationIcicle/Multiproof_size_2^21_bytes-4
[...]

PR #2240, opened 2025-10-28
fix(icicle-prover): multiframe msm bug

samlaf:
Bug was introduced in https://github.com/Layr-Labs/eigenda/pull/2220. We passed montgomery fr points to the msm, but forgot to set the config to indicate this.

Threw in an improvement to the encoding Config to allow running the unit tests using icicle. WIthout this I wasn't able to easily run our test suite to confirm that my changes did actually fix the bug.

### DIFF
--- a/encoding/backend.go
+++ b/encoding/backend.go
@@ -26,7 +26,7 @@ type Config struct {
 }
 
 // DefaultConfig returns a Config struct with default values.
-// If icicle is availeble (binary built with icicle tag), it sets the backend to icicle and enables GPU.
+// If icicle is available (binary built with icicle tag), it sets the backend to icicle and enables GPU.
 // Make sure to set GPUEnable to false if you want to run on CPU only.
 // If icicle is not available (build without icicle tag), it sets the backend to gnark.
 func DefaultConfig() *Config {


encoding/backend.go
```diff
@@ -24,8 +25,18 @@ type Config struct {
 	GPUEnable   bool
 }
 
-// DefaultConfig returns a Config struct with default values
+// DefaultConfig returns a Config struct with default values.
+// If icicle is availeble (binary built with icicle tag), it sets the backend to icicle and enables GPU.
```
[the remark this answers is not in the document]
samlaf: [20a0e20](https://github.com/Layr-Labs/eigenda/pull/2240/commits/20a0e201e0558e8e559a73fa017b63456250a54b)

dmanc (approved):

encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
```diff
@@ -276,11 +274,16 @@ func (c *KzgMultiProofBackend) msmBatchOnDevice(frs []fr.Element, g1Points []ici
 		return out, fmt.Errorf("allocating bytes on device failed: %v", err.AsString())
 	}
 
-	frsHostSlice := core.HostSliceFromElements(frs)
+	// The msm is computed synchronously (default config value is async=false).
```
bxue-l2, line 277: why did we set c.MsmCfg.Async to true?

samlaf: No idea... it might have been a bug: it feels to me like it creates a race condition by passing the async output to the next stream (ntt's) without synchronizing the stream.

PR #2242, opened 2025-10-29
feat: ejection transactor

cody-littley:
This PR introduces the ejection transactor, which encapsulates business logic for communicating with onchain ejection contracts.

### DIFF
--- a/ejector/ejection_transactor.go
+++ b/ejector/ejection_transactor.go
@@ -94,6 +94,9 @@ func NewEjectionTransactor(
 	if registryCoordinatorAddress == zeroAddress {
 		return nil, fmt.Errorf("registryCoordinatorAddress must be non-zero")
 	}
+	if chainID.Sign() <= 0 {
+		return nil, fmt.Errorf("invalid chainID: %s", chainID.String())
+	}
 
 	caller, err := contractEigenDAEjectionManager.NewContractIEigenDAEjectionManagerCaller(
 		ejectionContractAddress, client)
@@ -254,7 +257,8 @@ func (e *ejectionTransactor) IsValidatorPresentInAnyQuorum(
 
 func (e *ejectionTransactor) StartEjection(
 	ctx context.Context,
-	addressToEject gethcommon.Address) error {
+	addressToEject gethcommon.Address,
+) error {
 
 	rbn, err := e.referenceBlockProvider.GetReferenceBlockNumber(ctx)
 	if err != nil {


disperser/controller/metadata/reference_block_provider.go
cody-littley, line 1: This file got duplicated in my other PRs, due to the way I've been splitting smaller PRs off of my feature branch.

ejector/ejection_transactor.go
```diff
@@ -22,3 +31,257 @@ type EjectionTransactor interface {
 	// Complete the ejection proceedings against the operator with the given address.
 	CompleteEjection(ctx context.Context, addressToEject gethcommon.Address) error
 }
+
+var _ EjectionTransactor = &ejectionTransactor{}
+
+// ejectionTransactor is the production implementation of the EjectionTransactor interface.
+//
+// The ejection transactor is thread safe, although parallel calls may result in duplicate work (i.e. two calls might
```
litt3, line 39: Here you claim that the ejection transactor _is_ threadsafe, but it uses the periodic rbn provider, which claims to _not_ be threadsafe.

IIRC when reviewing the periodic provider, there was actually something that made it not threadsafe, though I don't recall the specifics.

ejector/ejection_transactor.go
```diff
@@ -22,3 +31,257 @@ type EjectionTransactor interface {
...
+	idToEject, err := e.validatorIDToAddressConverter.ValidatorAddressToID(ctx, addressToEject)
+	if err != nil {
+		return fmt.Errorf("failed to get validator ID from address: %w", err)
+	}
+
+	quorums, err := e.validatorQuorumLookup.GetQuorumsForValidator(ctx, idToEject, rbn)
+	if err != nil {
+		return fmt.Errorf("failed to get quorums for validator: %w", err)
+	}
+	quorumBytes := eth.QuorumListToBytes(quorums)
```
litt3, line 184: would it make sense to check if this is empty before submitting the ejection transaction?

ejector/ejection_transactor.go
```diff
@@ -22,3 +31,257 @@ type EjectionTransactor interface {
...
+		Context: ctx,
+		From:    e.selfAddress,
+		Signer:  e.signer,
+	}
+
+	if e.maxGasOverride != 0 {
+		opts.GasLimit = e.maxGasOverride
+	}
+
+	txn, err := e.transactor.CompleteEjection(opts, addressToEject, quorumBytes)
```
litt3, line 196: Is it a problem that the transaction could fail to be mined onchain, and this code would never know? What would be the implications of this transaction failing to get mined?

ejector/ejection_transactor.go
```diff
@@ -22,3 +31,257 @@ type EjectionTransactor interface {
...
+
+	// The maximum gas limit to use for transactions.
+	maxGasOverride uint64
+}
+
+// Create a new EjectionTransactor.
+func NewEjectionTransactor(
+	logger logging.Logger,
+	client bind.ContractBackend,
+	ejectionContractAddress gethcommon.Address,
```
litt3, line 73: Should you also validate that these aren't 0x0 contract addresses?

ejector/ejection_transactor.go
```diff
@@ -22,3 +31,257 @@ type EjectionTransactor interface {
...
+	registryCoordinatorAddress gethcommon.Address,
+	selfAddress gethcommon.Address,
+	privateKey *ecdsa.PrivateKey,
+	chainID *big.Int,
```
litt3, line 77: claude alleges `NewKeyedTransactorWithChainID` will panic if `chainID` is 0

ejector/ejection_transactor.go
cody-littley: Added a mutex into the reference block provider to make it thread safe. It's not something I expect to see much contention in, given how we use it. So I think a mutex should be performant enough for now.

ejector/ejection_transactor.go
cody-littley:
```
	if len(quorums) == 0 {
		return fmt.Errorf("cannot complete ejection: validator %s is not present in any quorum at RBN %d",
			idToEject.Hex(), rbn)
	}
```

ejector/ejection_transactor.go
cody-littley: If the transaction is not mined, then eventually the ejector will attempt to eject the same validator again. As long as the ejector continues retrying, the ejection workflow will complete once the ejection transaction finally gets mined.

Note that the ejection manager is capable of handling ejections that are "mid-phase". That is to say, failure to mine the ejection transaction will not cause the ejector to enter a broken state. As long as transactions eventually start getting mined, the ejector will converge on a good state.

ejector/ejection_transactor.go
cody-littley: Added checks. I don't mind defensive checks like this, but I don't always think they are necessary. Specifically for contract addresses, the offchain contract directory code will return an error if the address is a zero address.

ejector/ejection_transactor.go
cody-littley: It's not obvious to me when reading the code that this will happen, and if it does happen then that's dumb API. Can you ask claude to explain its reasoning? It's currently tricky for me to test this edge case.

litt3:
```
func newModernSigner(chainID *big.Int, fork forks.Fork) Signer {
	if chainID == nil || chainID.Sign() <= 0 {
		panic(fmt.Sprintf("invalid chainID %v", chainID))
	}
```

I dug down through the call, and found the same thing. This will panic if == 0

I don't know, probably fine to just ignore. Dumb API, but also I'm not sure how a chainId of 0 is ever gonna arrive here

litt3 (dismissed):

ejector/ejection_transactor.go
```diff
@@ -22,3 +31,257 @@ type EjectionTransactor interface {
...
+	if err != nil {
+		return false, fmt.Errorf("failed to get quorums for validator: %w", err)
+	}
+
+	return len(quorums) > 0, nil
+}
+
+func (e *ejectionTransactor) StartEjection(
+	ctx context.Context,
+	addressToEject gethcommon.Address) error {
```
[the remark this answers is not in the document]
cody-littley: fixed

ejector/ejection_transactor.go
cody-littley: added check

```
	if chainID.Sign() <= 0 {
		return nil, fmt.Errorf("invalid chainID: %s", chainID.String())
	}
```

cody-littley: So frustrating. `NewKeyedTransactorWithChainID()` returns an error already, whoever write this was just lazy.

PR #2243, opened 2025-10-29
refactor(prover-v2): GetFrames takes []fr.Element instead of []byte

samlaf:
We shouldn't be passing byte arrays around anywhere, its bug prone. 
We shouldnt be passing []fr.Element either but that's for a future larger scope refactor.

Also added a new benchmark for GetFrames() (to benchmark chunk + proof generation at the same time) which prompted this change in the first place. Need this to benchmark and understand the gpuSemaphore Im working on.


encoding/v2/bench/benchmark_eigenda_test.go
```diff
@@ -226,3 +227,53 @@ func benchmarkMultiproofGeneration(b *testing.B, encodingConfig encoding.Config)
 		})
 	}
 }
+
+// This does both chunk and proof generation, in separate goroutines.
+// In a sense it combines both benchmarks above.
+func BenchmarkFrameGeneration(b *testing.B) {
+	proverConfig := prover.KzgConfig{
+		// The loaded G1 point is not used because we require the SRSTables to be preloaded for the benchmark.
+		// We don't have enough SRS points in resourcs/srs/g1.point to compute the largest SRSTables anyways.
```
dmanc, line 236: Is this comment still true after Bowen's change?

dmanc (approved):

encoding/v2/bench/benchmark_eigenda_test.go
samlaf: Oh crap good point.......... need to update the srs-util also then :\

PR #2245, opened 2025-10-29
feat(proxy)!: bye bye `holesky`

ethenotethan:
holesky is being sunset 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/dispersal_request_signer.go
+++ b/api/clients/v2/dispersal_request_signer.go
@@ -12,6 +12,7 @@ import (
 	"github.com/aws/aws-sdk-go-v2/aws"
 	awsconfig "github.com/aws/aws-sdk-go-v2/config"
 	"github.com/aws/aws-sdk-go-v2/service/kms"
+	"github.com/ethereum/go-ethereum/crypto"
 	"github.com/pkg/errors"
 )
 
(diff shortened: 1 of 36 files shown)


api/proxy/Makefile
```diff
@@ -27,20 +27,12 @@ disperse-test-blob:
...
 	BACKEND=memstore gotestsum --format testname -- -v -timeout 10m ./test/e2e -parallel 8
 
-# E2E tests using holesky testnet backend, leveraging op-e2e framework. Also tests the standard client against the proxy.
-# If holesky tests are failing, consider checking https://dora.holesky.ethpandaops.io/epochs for block production status.
-test-e2e-testnet:
-	BACKEND=testnet gotestsum --format testname -- -v -timeout 20m ./test/e2e -parallel 32
-
-## Equivalent to `test-e2e-testnet`, but against preprod instead of testnet
-test-e2e-preprod:
-	BACKEND=preprod gotestsum --format testname -- -v -timeout 20m ./test/e2e -parallel 32
```
samlaf, line 41: shouldnt we keep these and instead migrate them to use the new hoodi networks instead?

ethenotethan: we should but I feel like that's a follow up PR given we've introduced `hoodi` already into proxy and this PR is specific to holesky deprecation. can add a todo

ethenotethan: [05ac760](https://github.com/Layr-Labs/eigenda/pull/2245/commits/05ac760c75622aa16a8536bac311af7bffce14c8)

api/proxy/test/testutils/setup.go
```diff
@@ -242,16 +226,6 @@ func BuildTestSuiteConfig(testCfg TestConfig) config.AppConfig {
...
-	case PreprodBackend:
-		disperserHostname = disperserPreprodHostname
-		certVerifierAddress = preprodCertVerifierAddress
-		svcManagerAddress = preprodSvcManagerAddress
-		eigenDADirectory = preprodEigenDADirectory
-	case TestnetBackend:
-		disperserHostname = disperserTestnetHostname
-		certVerifierAddress = testnetCertVerifierAddress
-		svcManagerAddress = testnetSvcManagerAddress
-		eigenDADirectory = testnetEigenDADirectory
```
samlaf, line 254: why not replace these with hoodi equivalent?

api/proxy/Makefile
samlaf: oh wait we already have hoodi support? Then why not keep this test-e2e-testnet/preprod around and just use the hoodi backend then?

samlaf (approved):
LGTM but if adding hoodi support is just a matter of wiring it and doesn't require any large amount of new code, then I'd vote for making the move in this PR. Looks wrong that we are removing all preprod stuff from my pov, but I might be missing some complications. From outsider's perspective though, we should instead move "laterally" if we can and instead and move preprod to run against hoodi, not just get rid of preprod support/make/tests.

PR #2246, opened 2025-10-29
feat(proxy): Support multi-serialization (ABI, RLP)

ethenotethan:
## Changes
- Adds dynamic encoding through a `serializationType` enum which is passed down the application callstack through each server router implementation
- RLP for REST && ABI for ARB Custom DA
- 50% of this PR is just tests

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/coretypes/eigenda_cert.go
+++ b/api/clients/v2/coretypes/eigenda_cert.go
@@ -1,6 +1,7 @@
 package coretypes
 
 import (
+	"encoding/json"
 	"fmt"
 
 	disperser "github.com/Layr-Labs/eigenda/api/grpc/disperser/v2"
@@ -71,6 +72,19 @@ type EigenDACert interface {
 	// For the theoretical reasoning behind this choice, see
 	// https://www.tedinski.com/2018/02/27/the-expression-problem.html
 	isEigenDACert()
+}
+
+// DeserializeEigenDACert deserializes raw bytes into an EigenDACert
+// based on the provided version and serialization type
+func DeserializeEigenDACert(data []byte, version CertificateVersion, ct CertSerializationType) (EigenDACert, error) {
+	switch version {
+	case VersionTwoCert:
+		return DeserializeEigenDACertV2(data, ct)
+	case VersionThreeCert:
+		return DeserializeEigenDACertV3(data, ct)
+	default:
+		return nil, fmt.Errorf("unsupported certificate version: %d", version)
+	}
 }
 
 var _ EigenDACert = &EigenDACertV2{}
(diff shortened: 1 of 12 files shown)


api/clients/v2/coretypes/eigenda_cert.go
```diff
@@ -73,6 +74,19 @@ type EigenDACert interface {
 	isEigenDACert()
 }
 
+// DeserializeEigenDACert deserializes raw bytes into an EigenDACert
+// based on the provided version and serialization type
+func DeserializeEigenDACert(data []byte, version CertificateVersion, ct CertSerializationType) (EigenDACert, error) {
```
samlaf, line 79: do we need this function? Can't we just create 2 separate `DeserializeABIEncodedCert` and `DeserializeRLPEncodedCert`?

samlaf: especially if we know at compile time that those are ever only going to be used in independent places (abi for arb, rlp for op), then I don't think we should be merging them into a function. Always best to have info available at compile time made obvious to programmer imo.

api/clients/v2/coretypes/eigenda_cert.go
```diff
@@ -185,6 +199,63 @@ func (c *EigenDACertV3) Serialize(ct CertSerializationType) ([]byte, error) {
 
 }
 
+// DeserializeEigenDACertV2 deserializes raw bytes into an EigenDACertV2
+func DeserializeEigenDACertV2(data []byte, ct CertSerializationType) (*EigenDACertV3, error) {
```
samlaf, line 203: I think these as well should prob be kept separate? aka separate fct for abi and rlp decoding. Unless we really plan on offering more than those 2 serialization.

api/proxy/store/eigenda_manager.go
```diff
@@ -7,21 +7,36 @@ import (
...
 )
 
+/*
+	TODO: right now, the serialization type is passed through the application call chain from
+	handlers -> eigenda_manager -> underlying clients where a DA Cert is either serialized/deserialized.
+
+	This incurs the additive cost of an additional param being passed through. The intersection of
+	V1 x V2 within this construction makes it challenging to modularize. Once V1 code paths are nuked, the
+	serialization call chain should be reworked to support a smaller overhead implementation.
+*/
```
samlaf, line 25: I see. So we can't have the store manager return a cert because the cert struct is not used in V1. But after we nuke V2 then we can have the store return a cert which we can serialize in the handlers?

api/clients/v2/coretypes/eigenda_cert.go
samlaf: EDIT: hmm looking at the code in https://github.com/Layr-Labs/eigenda/pull/2246/files realized this might be harder than I thought. Because the store IS actually multiplexed and already doesn't know whether its in an arb or op context...

Gosh I really hate our store abstraction. I guess we're stuck with it for now but we really need to rework it after we nuke V1. Prob worth pondering a little how we got into this state, to make sure we don't ever end up in similar state in other projects. My main issue with it is this dynamic multiplexing. You should basically as much as possible never have such a god class in teh middle of the code. Better to wire things independently and keep information as local as possible.

samlaf (approved):

api/proxy/store/eigenda_manager.go
ethenotethan: yup that is the thinking here

api/clients/v2/coretypes/eigenda_cert.go
ethenotethan: yeah the store V1/B2 backend multiplexing makes it nontrivial to abstract this properly given this bloated intermediary construction between the server handlers and the underlying clients. definitely agree that we should fix this asap once we nuke all this shitty legacy V1 code

api/clients/v2/coretypes/eigenda_cert.go
```diff
@@ -185,6 +199,63 @@ func (c *EigenDACertV3) Serialize(ct CertSerializationType) ([]byte, error) {
 
 }
 
+// DeserializeEigenDACertV2 deserializes raw bytes into an EigenDACertV2
```
litt3, line 202:
```suggestion
// DeserializeEigenDACertV2 deserializes raw bytes into an EigenDACertV3
```

PR #2248, opened 2025-10-29
feat(payments): Make sure only authorized dispersers can you on-demand payments

litt3:
closes EGDA-2235

- This is in preparation for permissionless dispersal
- Validator nodes must only permit on-demand dispersal from authorized dispersers
- Currently it's hardcoded to just the eigenlabs disperser. In the future, this should be configured via the configuration registry

### DIFF
--- a/node/auth/authenticator.go
+++ b/node/auth/authenticator.go
@@ -25,11 +25,11 @@ type RequestAuthenticator interface {
 		request *grpc.StoreChunksRequest,
 		now time.Time) ([]byte, error)
 
-	// CheckOnDemandPaymentAuthorization returns true if the disperser is authorized to disperse the given batch.
+	// IsDisperserAuthorized returns true if the disperser is authorized to disperse the given batch.
 	// Returns true if the batch contains only reservation payments, or if the batch contains on-demand payments
 	// and the disperser is authorized to handle them. Returns false if the batch contains on-demand payments
 	// and the disperser is not authorized.
-	CheckOnDemandPaymentAuthorization(disperserID uint32, batch *corev2.Batch) bool
+	IsDisperserAuthorized(disperserID uint32, batch *corev2.Batch) bool
 }
 
 // keyWithTimeout is a key with that key's expiration time. After a key "expires", it should be reloaded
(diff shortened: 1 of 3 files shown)


node/auth/authenticator.go
```diff
@@ -58,18 +68,26 @@ func NewRequestAuthenticator(
 	keyCacheSize int,
 	keyTimeoutDuration time.Duration,
 	disperserIDFilter func(uint32) bool,
-	now time.Time) (RequestAuthenticator, error) {
+	authorizedOnDemandDispersers []uint32,
```
cody-littley, line 71: As discussed offline, this should eventually use onchain data. Ok that it doesn't for the current PR.

node/auth/authenticator.go
cody-littley, line 1: Ok merging this for now, given that the architecture changes we discussed offline require smart contract changes.

node/auth/authenticator.go
```diff
@@ -23,6 +24,12 @@ type RequestAuthenticator interface {
 		ctx context.Context,
 		request *grpc.StoreChunksRequest,
 		now time.Time) ([]byte, error)
+
+	// CheckOnDemandPaymentAuthorization returns true if the disperser is authorized to disperse the given batch.
+	// Returns true if the batch contains only reservation payments, or if the batch contains on-demand payments
+	// and the disperser is authorized to handle them. Returns false if the batch contains on-demand payments
+	// and the disperser is not authorized.
+	CheckOnDemandPaymentAuthorization(disperserID uint32, batch *corev2.Batch) bool
```
cody-littley:
```suggestion
	IsAuthorizedForOnDemand(disperserID uint32, batch *corev2.Batch) bool
```

Nit, I like it when function names cause expressions to read like understandable phrases.

litt3: I changed the name in 52053f081929d0c8e367917b0641e74d39289b95 to `IsDisperserAuthorized`

`IsAuthorizedForOnDemand` is misleading, since it implies we are only checking if the disperser is authorized for on demand. In reality, we're actually checking:
1. Is the disperser authorized for on demand?
2. OR does the batch contain only reservation payments?

LMK if you're still unhappy with the naming, or if you think this would be better split into two methods (e.g. `doesBatchContainOnDemand` combined with `IsDisperserAuthorizedForOnDemand`)

cody-littley (approved):

PR #2249, opened 2025-10-29
perf(icicle): gpu semaphore to allow >1 concurrent gpu task

samlaf:
This will allow using the GPU more efficiently by submitting more than 1 request at a time.

Copying comment I left in the code's config field for visibility:
```
// Increase this value to allow more concurrent GPU frame (chunk+proof) tasks.
// Only used by V2 when Backend=icicle and GPUEnable=true.
// Note Chunk generation (encoding/v2/rs) and multiproofs generation (encoding/v2/kzg/prover)
// each have their own separate semaphore which is weighted using this same value.
//
// This protects against out-of-memory errors on the GPU, not GPU time usage.
// WARNING: setting this value too high may lead to out-of-memory errors on the GPU.
// If this ever happens, the GPU device needs to be rebooted as it can be left in a bad state.
//
// For now we use this very coarse-grained approach, instead of using a RAM-usage based semaphore,
// because that would feel brittle and require approximations of RAM usage per MSM/NTT operation.
// We can rethink this abstraction later if needed.
```

Default value for semahore is still set to 1, so as is this PR will not change the behavior of the code.
However it will allow experimenting with submitting concurrent tasks to GPU.
I've tested a benchmark run submitting 20 requests in parallels and never saw > 12GiB/24GiB being used on the g6.xlarge's L4 GPU.


encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
```diff
@@ -3,16 +3,19 @@
 package icicle
 
 import (
+	"context"
 	"fmt"
 	"slices"
 	"sync"
 	"time"
 
+	_ "github.com/Layr-Labs/eigenda/encoding"
```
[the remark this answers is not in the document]
samlaf: its used in some godoc comment.

encoding/backend.go
```diff
@@ -23,6 +23,19 @@ type Config struct {
...
+	// each have their own separate semaphore which is weighted using this same value.
+	//
+	// This protects against out-of-memory errors on the GPU, not GPU time usage.
+	// WARNING: setting this value too high may lead to out-of-memory errors on the GPU.
+	// If this ever happens, the GPU device needs to be rebooted as it can be left in a bad state.
+	//
+	// For now we use this very coarse-grained approach, instead of using a RAM-usage based semaphore,
+	// because that would feel brittle and require approximations of RAM usage per MSM/NTT operation.
+	// We can rethink this abstraction later if needed.
+	GPUConcurrentFrameGenerationDangerous int64
```
dmanc, line 42: why int64 instead of uint64?

encoding/v2/rs/backend/rs_backend.go
```diff
@@ -1,14 +1,15 @@
...
+	"context"
 	"github.com/Layr-Labs/eigenda/encoding/v2/rs/backend/gnark"
 	"github.com/Layr-Labs/eigenda/encoding/v2/rs/backend/icicle"
 	"github.com/consensys/gnark-crypto/ecc/bn254/fr"
 )
 
 // Proof device represents a device capable of computing reed-solomon operations.
 type RSEncoderBackend interface {
-	ExtendPolyEval(coeffs []fr.Element) ([]fr.Element, error)
+	ExtendPolyEvalV2(ctx context.Context, coeffs []fr.Element) ([]fr.Element, error)
```
dmanc, line 12: why did we rename it to V2?

encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
```diff
@@ -92,9 +97,15 @@ func (p *KzgMultiProofBackend) ComputeMultiFrameProofV2(polyFr []fr.Element, num
...
 	var icicleErr error
 
-	// GPU operations
-	p.GpuLock.Lock()
-	defer p.GpuLock.Unlock()
+	// We acquire a semaphore here to avoid too many concurrent GPU requests,
+	// each of which does 1 MSM + 2 NTTs. This is a very unideal and coarse grain solution, but unfortunately
+	// icicle doesn't have nice backpressure, and the GPU kernel just panics if RAM is exhausted.
+	// We could use a finer-grained semaphore that calculates the RAM usage per request,
+	// but we'd have to hardcode some approximation of the RAM usage per MSM/NTT, which feels
```
dmanc, line 104: So if I understand correctly it's coarse-grained because it ignores blob size right? I feel we may want to limit the value based on our largest size for our deployments.

encoding/v2/bench/benchmark_eigenda_test.go
```diff
@@ -243,9 +245,15 @@ func BenchmarkFrameGeneration(b *testing.B) {
 		CacheDir:       "../../../resources/srs/SRSTables",
 		NumWorker:      uint64(runtime.GOMAXPROCS(0)),
 	}
+	encodingConfig := encoding.Config{
+		NumWorker:                             uint64(runtime.NumCPU()),
```
pschork: why not `uint64(runtime.GOMAXPROCS(0))`?

disperser/cmd/encoder/flags/flags.go
```diff
@@ -113,6 +113,18 @@ var (
...
+	GPUConcurrentFrameGenerationDangerousFlag = cli.IntFlag{
+		Name: common.PrefixFlag(FlagPrefix, "gpu-concurrent-frame-generation-dangerous"),
+		Usage: "Weight of the GPU semaphore. Increase this value to allow more concurrent GPU tasks. " +
+			"This is only used by V2 when GPU is enabled (Backend=icicle and GPUEnable=true). " +
+			"Chunk generation (encoding/v2/rs) and multiproofs generation (encoding/v2/kzg/prover) " +
+			"each have their own separate semaphore which is weighted using this value. " +
+			"WARNING: setting this value too high may lead to out-of-memory errors on the GPU. " +
+			"If this ever happens, the GPU device needs to be rebooted as it can be left in a bad state.",
+		Required: false,
+		Value:    1,
```
pschork: What happens if icicle is enabled but GPUConcurrentFrameGenerationDangerousFlag is set to 0?
The benchmark sets this to 20. What is a reasonably safe range for a single L40/A10?

encoding/backend.go
samlaf: that's what the semaphore takes

encoding/v2/rs/backend/rs_backend.go
samlaf: Just a convenience since otherwise the IDE shows the encoding/v1 implementations as well. I did the same thing on the prover interface already.

encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
samlaf: correct. thats a good point actually

encoding/v2/bench/benchmark_eigenda_test.go
samlaf: good question. I started using NumCPU() recently unknowingly. I guess you're right I should be using GOMAXPROCS because that can be changed dynamically, although in practice probably always behaves the same?

disperser/cmd/encoder/flags/flags.go
samlaf: 1. you get a panic. Added config verification to encoder's constructor: [e77b650](https://github.com/Layr-Labs/eigenda/pull/2249/commits/e77b65096ee9b4d95a932b7f4dfbfaa646e23585)

encoding/v2/bench/benchmark_eigenda_test.go
samlaf: [b8e4a4a](https://github.com/Layr-Labs/eigenda/pull/2249/commits/b8e4a4a6909eb3dccebff0528a28bfd6ee978712)

samlaf:
@dmanc @pschork addressed most comments and cleaned up notation. The PR bloated quite a bit because of making the rs.Encoder validate its config, which changed the API to return an err, and had to update a crap ton of call sites. :\

But otherwise meaningful changes should be simpler now: got rid of new flag and reusing `max-concurrent-requests` instead.

> > I've tested a benchmark run submitting 20 requests in parallels and never saw > 12GiB/24GiB being used on the g6.xlarge's L4 GPU.
> 
> What size blobs?

16MiB blobs. My bad, but whenver I quote benchmarks itll always be with max sized blobs.

> Can we make consistent usage of `gpuConcurrentEncodings`, `gpuConcurrentTasks` and `maxConcurrentRequests`? Are these all referencing the same thing of is there a distinction between encoding and frame concurrency?

There's a lot of different names because frameGeneration = chunkGeneration + proofGeneration, and the same weight is used in both semaphores. So basically we have 2 separate goroutines shipping stuff to GPU (chunk + proofs). They are guarded by their own semaphore that share the same weight, so overall the semantics of `maxConcurrentRequests` remains "max amount of blob that can be encoded+proven in parallel on GPU". Let me know if things are still unclear I can update docs or naming.

disperser/cmd/encoder/flags/flags.go
samlaf: as for your second question... we'll really have to experiment here, I'm not sure. For a L40/A10 we have 24GiB of GPU RAM. After setup we send afair <2GiB of srs points and ntt domain points, so we have 22GiB left to play with. FFT uses 2*blob_size of RAM, so at most 256MiB (16MiB * 8codingRatio * 2). MSM I honestly am not sure, would need to benchmark and play around with it more. Ive added https://dev.ingonyama.com/api/cpp/msm#memory-usage-estimation to the docs. Could do some more research if you're interested.

disperser/cmd/encoder/flags/flags.go
```diff
@@ -74,8 +74,14 @@ var (
...
-		Name:     common.PrefixFlag(FlagPrefix, "max-concurrent-requests"),
-		Usage:    "maximum number of concurrent requests",
+		Name: common.PrefixFlag(FlagPrefix, "max-concurrent-requests"),
+		Usage: "maximum number of concurrent requests. " +
+			"This also sets the weight of the GPU semaphore when using EigenDA V2 with GPU is enabled " +
+			"(Backend=icicle and GPUEnable=true). " +
+			"Chunk generation (encoding/v2/rs) and multiproofs generation (encoding/v2/kzg/prover) " +
+			"each have their own separate semaphore which is weighted using this value. " +
+			"WARNING: setting this value too high may lead to out-of-memory errors on the GPU. " +
+			"If this ever happens, the GPU device needs to be rebooted as it can be left in a bad state.",
```
dmanc, line 84: This seems disturbing lol. Are you saying the machine will hang?

samlaf: This is what the Jeremy from Ingonyama said:
image

api/clients/v2/validator/validator_non_mock_test.go
```diff
@@ -103,7 +104,7 @@ func TestNonMockedValidatorClientWorkflow(t *testing.T) {
 	// Create the actual blob frames using the prover
 	dataFr, err := rs.ToFrArray(data)
 	require.NoError(t, err)
-	frames, _, err := p.GetFrames(dataFr, encodingParams)
+	frames, _, err := p.GetFrames(t.Context(), dataFr, encodingParams)
```
litt3: nit: you can use the `ctx` already created

core/v2/core_test.go
```diff
@@ -171,7 +171,7 @@ func prepareBlobs(
 		require.NoError(t, err)
 		blobFr, err := rs.ToFrArray(blob)
 		require.NoError(t, err)
-		frames, _, err := p.GetFrames(blobFr, params)
+		frames, _, err := p.GetFrames(t.Context(), blobFr, params)
```
litt3: here too, `ctx` already exists

disperser/cmd/encoder/flags/flags.go
```diff
@@ -74,8 +74,14 @@ var (
 		EnvVar:   common.PrefixEnvVar(envVarPrefix, "ENABLE_METRICS"),
 	}
 	MaxConcurrentRequestsFlag = cli.IntFlag{
-		Name:     common.PrefixFlag(FlagPrefix, "max-concurrent-requests"),
-		Usage:    "maximum number of concurrent requests",
+		Name: common.PrefixFlag(FlagPrefix, "max-concurrent-requests"),
+		Usage: "maximum number of concurrent requests. " +
+			"This also sets the weight of the GPU semaphore when using EigenDA V2 with GPU is enabled " +
```
litt3:
```suggestion
			"This also sets the weight of the GPU semaphore when using EigenDA V2 with GPU enabled " +
```

litt3 (approved):

disperser/cmd/encoder/flags/flags.go
samlaf: [feb9a44](https://github.com/Layr-Labs/eigenda/pull/2249/commits/feb9a441d02156c7ece025deac28f6e1af74aaa4)

api/clients/v2/validator/validator_non_mock_test.go
samlaf: [feb9a44](https://github.com/Layr-Labs/eigenda/pull/2249/commits/feb9a441d02156c7ece025deac28f6e1af74aaa4)
Changed but personally not a fan of doing `ctx := t.Context()` at the top. I prefer things to be explicit, and when I see a `ctx` I expect it to not be a ctx with additions (timeout, cancel, etc).

core/v2/core_test.go
samlaf: [feb9a44](https://github.com/Layr-Labs/eigenda/pull/2249/commits/feb9a441d02156c7ece025deac28f6e1af74aaa4)
Changed but personally not a fan of doing `ctx := t.Context()` at the top. I prefer things to be explicit, and when I see a `ctx` I expect it to not be a ctx with additions (timeout, cancel, etc).

PR #2250, opened 2025-10-30
fix: enforce all padding in encoded payload be 0

bxue-l2:
# PR update

I ended up renaming the function again, from `RemoveInternalPadding` to `CheckAndRemoveInternalPadding` then to `CheckAndRemoveInternalFieldElementPadding`. Then create a new function with the name `CheckAndRemoveInternalPadding` with a new function signature `CheckAndRemoveInternalPadding(paddedData []byte, payloadLen uint32)`. It is driven by the part that old naming isn't correct. It is going to break the users who rely on this library, which I think there are even a few or none. For them, I think it is better that it breaks.

This PR also addressed the padding issue for the encoded payload header.

The scope increased a bit. Changed the PR title from
Before: `fix: enforce integration spec the first byte of every 32 bytes must be 0`
to: `fix: enforce all padding in encoded payload be 0"

In the eigenda integration [spec](https://layr-labs.github.io/eigenda/integration/spec/3-data-structs.html#encoding-payload-version-0x0), when padding the payload, we requires the padding to be 0. However, it is not enforced in the decoding process.

Technically we can allow non zero padding as long as it is a valid field element. After discussing with the team, we decide to enforce it. 

Note it is a breaking change for rollups that have padding scheme other than the one specified in the [spec](https://layr-labs.github.io/eigenda/integration/spec/3-data-structs.html#encoding-payload-version-0x0), i.e pad zero byte in front of every 31 bytes of rollup payload.

## Spec consideration

The spec is also updated to reflect the change. I am debating myself if we should just enforce all the padding must be zero. As it is really simple to think about. In contrast, currently we only check if the first byte is 0x00 for every 32 bytes data chunk, and we don't care what is in the remaining 31 bytes. (Maybe this is simpler to think about after all?)

Consider this. If the rollup payload is 32 bytes => encoded payload of size 64 bytes, with the second chunk containing only 1 byte of payload information, and we don't care for the remaining 30 bytes of padding.

Changing to enforce all 0 padding introduces additional steps at decoding side, i.e. we need to see if all the non-useful bytes (after `payloadLen`) in the `decoded_body` are zero, by looping through an array. The number of zero-check at worst is the entire blob.

## equivalence

Here is the rust implementation for the same check. Two must have identical behavior. https://github.com/Layr-Labs/hokulea/pull/236

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/coretypes/encoded_payload.go
+++ b/api/clients/v2/coretypes/encoded_payload.go
@@ -131,10 +131,13 @@ func (ep *EncodedPayload) decodeHeader() (uint32, error) {
 	return payloadLength, nil
 }
 
-// decodePayload decodes the body by checking and removing internal 0 byte padding, that include both the
-// padding added to make every multiple of 32 bytes a valid Field elements and
-// the padding added to make encoded payload contain a power of 2 number of Field elements.
-// it returns error if any padding is non-zero, or the body contains insufficient number of bytes for payload length.
+// decodePayload decodes the body by checking for and removing internal zero-byte padding,
+// including both:
+//   - padding added to make each 32-byte chunk a valid field element, and
+//   - padding added to make the encoded payload contain a power-of-two number of field elements.
+//
+// It returns an error if any padding bytes are non-zero, or if the body contains insufficient amount of
+// data required for the payload length.
 func (ep *EncodedPayload) decodePayload(payloadLen uint32) ([]byte, error) {
 	body := ep.bytes[codec.EncodedPayloadHeaderLenBytes:]
 	// Decode the body by removing 0x00 initial padding byte for every 32 byte chunk


encoding/codec/codec.go
```diff
@@ -147,6 +150,12 @@ func RemoveInternalPadding(paddedData []byte) ([]byte, error) {
 		dstIndex := i * bytesPerChunk
 		srcIndex := i*encoding.BYTES_PER_SYMBOL + 1
 
+		if paddedData[dstIndex] != 0x0 {
```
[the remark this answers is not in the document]
shyam-patel-kira: @bxue-l2 This seems like an actual bug

encoding/codec/test/codec_test.go
```diff
@@ -260,3 +260,23 @@ func TestPadUnpad(t *testing.T) {
...
+	testRandom := random.NewTestRandom()
+	testIterations := 1000
+
+	for i := 0; i < testIterations; i++ {
+		originalBytes := testRandom.Bytes(1 + testRandom.Intn(1023))
+
+		paddedBytes := codec.PadPayload(originalBytes)
+
+		// first byte be non-zero violation
+		paddedBytes[0] = 1
```
shyam-patel-kira: why are we corrupting only this byte?

encoding/codec/codec.go
```diff
@@ -120,15 +120,18 @@ func PadPayload(inputData []byte) []byte {
...
 //
 // This function assumes that the input aligns to 32 bytes. Since it is removing 1 byte for every 31 bytes kept, the
 // output from this function is not guaranteed to align to 32 bytes.
 //
 // NOTE: this method is a reimplementation of RemoveEmptyByteFromPaddedBytes, with one meaningful difference: this
 // function relies on the assumption that the input is aligned to encoding.BYTES_PER_SYMBOL, which makes the padding
 // removal logic simpler.
-func RemoveInternalPadding(paddedData []byte) ([]byte, error) {
+//
+// In addition, this function rquires the first byte in every multiple of 32 bytes to be 0x00.
```
shyam-patel-kira:
```suggestion
// In addition, this function requires the first byte in every multiple of 32 bytes to be 0x00.
```
nit: typo

encoding/codec/test/codec_test.go
bxue-l2: because it is enough, though you can also correct other byts

samlaf (approved):
LGTM, but what about the other function RemoveEmptyByteFromPaddedBytes? Seems like its still used in at least 1 V2 place on the retriever server (whatever that is).

encoding/codec/test/codec_test.go
ethenotethan: @bxue-l2 I think for brevity we should be corrupting bytes at a non-zero index as well for this test

encoding/codec/codec.go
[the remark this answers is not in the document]
ethenotethan: iiuc we're calculating the first symbol index by taking:
```
bytesPerChunk := encoding.BYTES_PER_SYMBOL - 1

where:
   encoding.BYTES_PER_SYMBOL = 32
   bytesPerChunk = 31
```

and multiplying `bytesPerChunk * i` to get `dstIndex` which would be at 31 index offsets:
```
[0, 31, 62, ...]
```

so I'm inclined to agree with @shyam-patel-kira here. think we should extend the test coverage to also capture non-zero index assertions

encoding/codec/test/codec_test.go
ethenotethan: given we're already sourcing entropy, what if we just computed an arbitrary multiple of 32 index; something like:
```
func TestDetectInvalidPad(t *testing.T) {
	testRandom := random.NewTestRandom()
	testIterations := 1000

	for i := 0; i < testIterations; i++ {
		originalBytes := testRandom.Bytes(64 + testRandom.Intn(1023))

		paddedBytes := codec.PadPayload(originalBytes)

		corruptionIndex := testRandom.Int32Range(0, int32(len(paddedBytes)/32)) * 32

		// first byte of some field element be non-zero violation
		paddedBytes[corruptionIndex] = 1
		require.Equal(t, len(paddedBytes)%32, 0)

		_, err := codec.CheckAndRemoveInternalPadding(paddedBytes)
		require.Error(t, err)
	}
}
```

encoding/codec/codec.go
```diff
@@ -147,6 +150,12 @@ func RemoveInternalPadding(paddedData []byte) ([]byte, error) {
 		dstIndex := i * bytesPerChunk
 		srcIndex := i*encoding.BYTES_PER_SYMBOL + 1
 
+		if paddedData[dstIndex] != 0x0 {
+			return nil, fmt.Errorf(
+				"the first byte of every encoding.BYTES_PER_SYMBOL in the padded data must be 0x00",
```
ethenotethan: nit - any value in returning the field element index in the error?

encoding/codec/test/codec_test.go
litt3: Agreed- testing only the 0 case is liable to miss index bugs (like [this one](https://github.com/Layr-Labs/eigenda/pull/2250#discussion_r2476180674) )

encoding/codec/codec.go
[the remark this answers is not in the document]
bxue-l2: gosh, I was checking the wrong bytse, ehhh. My bad

encoding/codec/test/codec_test.go
bxue-l2: The old implementation was plainly wrong. Didn't spend enough time on it, before getting everyone agreeing on this.

But yes, my fault proved it is a really good idea.

encoding/codec/codec.go
bxue-l2: can do

encoding/codec/codec.go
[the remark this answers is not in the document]
bxue-l2: https://github.com/Layr-Labs/eigenda/pull/2250/commits/abd1fcc906dcaf5b18b9147a1f15533a59d0c368#diff-7383d1980f651b87f4e2909d25097860fa43c9812d292154fef099a525d3b3a9R153

encoding/codec/test/codec_test.go
bxue-l2: https://github.com/Layr-Labs/eigenda/pull/2250/commits/abd1fcc906dcaf5b18b9147a1f15533a59d0c368#diff-5bc3be1f676b43f3810f7c36059bbcfb00969dd1e08219bfb826c1303431b2f2R272

encoding/codec/codec.go
bxue-l2: https://github.com/Layr-Labs/eigenda/pull/2250/commits/abd1fcc906dcaf5b18b9147a1f15533a59d0c368#diff-7383d1980f651b87f4e2909d25097860fa43c9812d292154fef099a525d3b3a9R133

encoding/codec/codec.go
bxue-l2: https://github.com/Layr-Labs/eigenda/pull/2250/commits/abd1fcc906dcaf5b18b9147a1f15533a59d0c368#diff-7383d1980f651b87f4e2909d25097860fa43c9812d292154fef099a525d3b3a9R155

retriever/v2/server.go
```diff
@@ -84,8 +84,11 @@ func (s *Server) RetrieveBlob(ctx context.Context, req *pb.BlobRequest) (*pb.Blo
 	if err != nil {
 		return nil, err
 	}
-	restored := bytes.TrimRight(data, "\x00")
-	restored = codec.RemoveEmptyByteFromPaddedBytes(restored)
+
+	restored, err := codec.CheckAndRemoveInternalPadding(data)
```
litt3: The old code was wrong, and so is this new code.

This method allegedly retrieves a blob, but this method is partially (and incorrectly) doing blob -> payload related conversions.

IMO we should just delete the whole `v2` package within `retriever`, if we cannot find a compelling reason it needs to exist

bxue-l2: I don't think anyone is using the retriever, so I am for it. But it is good to get consensus on that, but not as a part of this PR.

encoding/codec/codec.go
```diff
@@ -120,15 +120,18 @@ func PadPayload(inputData []byte) []byte {
...
 // This function assumes that the input aligns to 32 bytes. Since it is removing 1 byte for every 31 bytes kept, the
 // output from this function is not guaranteed to align to 32 bytes.
 //
 // NOTE: this method is a reimplementation of RemoveEmptyByteFromPaddedBytes, with one meaningful difference: this
 // function relies on the assumption that the input is aligned to encoding.BYTES_PER_SYMBOL, which makes the padding
 // removal logic simpler.
-func RemoveInternalPadding(paddedData []byte) ([]byte, error) {
+//
+// In addition, this function requires the first byte in every multiple of 32 bytes to be 0x00.
+func CheckAndRemoveInternalFieldElementPadding(paddedData []byte) ([]byte, error) {
```
samlaf, line 134: do we really need to expose this function publicly? We have a lot of similar functions in this package already, and they are very confusing to read. adding more to the public API is going to add to the confusion imo.

encoding/codec/codec.go
```diff
@@ -147,12 +150,52 @@ func RemoveInternalPadding(paddedData []byte) ([]byte, error) {
...
 
 	return outputData, nil
 }
 
+// CheckAndRemoveInternalPadding accepts an array of padded data, then checks and removes all the padding that was
+// added to the payload in order to make every 32 bytes a valid field element in the final output, and to make
+// the length of the output to hold a power of 2 field elements i.e (32, 64, 128,...) bytes.
+//
+// This function checks all the padding must be 0.
+func CheckAndRemoveInternalPadding(paddedData []byte, payloadLen uint32) ([]byte, error) {
```
samlaf: Looks like this function doesnt belong here now, because imo codec.go should not have to know about `payloadLen`. Maybe keep the other one public and move this function to coretypes where its used? Not sure..

encoding/codec/codec.go
bxue-l2: I don't think this function is required by public, it is too low level.

encoding/codec/codec.go
bxue-l2: that sounds good

encoding/codec/codec.go
bxue-l2: tried it, I think we have to create some `internal` directory name to only restrict access within the repo. Simply changing the function name to lower case won't work. So not going to include it

encoding/codec/codec.go
bxue-l2: 840753c8256d80bdeb33ac33800e9fb5fefde35e

samlaf (approved):
Overall looks fine. Still think all the serialize/deserialize/padding/codec functions/packages need a huge refactor, but I think this works for now. Would be good to wait for others to check though.

retriever/v2/server.go
litt3: If you don't want to delete in this PR, then you probably shouldn't make the change here.

It's incorrect, so making it just muddies the water IMO

api/clients/v2/coretypes/encoded_payload.go
```diff
@@ -121,27 +121,26 @@ func (ep *EncodedPayload) decodeHeader() (uint32, error) {
...
-	decodedBody, err := codec.CheckAndRemoveInternalPadding(body)
+	decodedPayload, err := checkAndRemoveInternalPadding(body, payloadLen)
 	if err != nil {
 		return nil, fmt.Errorf("remove internal padding: %w", err)
 	}
 
-	// data length is checked when constructing an encoded payload. If this error is encountered, that means there
-	// must be a flaw in the logic at construction time (or someone was bad and didn't use the proper construction methods)
-	if uint32(len(decodedBody)) < payloadLen {
-		return nil, fmt.Errorf(
```
litt3, line 139: If you don't want to return an error here, I think it still makes sense to log this occurrence as a sanity check

api/clients/v2/coretypes/encoded_payload_test.go
```diff
@@ -100,6 +100,18 @@ func TestDecodePayloadErrors(t *testing.T) {
...
 		},
+		{
+			name: "padding in encoded payload must be 0x00",
+			encodedPayloadHex: "0000000000200000000000000000000000000000000000000000000000000000" +
+				"0001020304050607080910111213141516171819202122232425262728293031" +
+				"0032000000000000000000000000000000000000000000000000000000000000" +
+				"0000000000000000000000010000000000100000000000000000000000000111",
+		},
+		{
+			name: "padding in encoded payload must be 0x00",
```
litt3: You have the same name for these two test cases. If they're both necessary, the names should explain the difference

encoding/codec/test/codec_test.go
```diff
@@ -246,17 +246,40 @@ func TestPadUnpad(t *testing.T) {
...
-		unpaddedBytes, err := codec.RemoveInternalPadding(paddedBytes)
+		unpaddedBytes, err := codec.CheckAndRemoveInternalFieldElementPadding(paddedBytes)
 		require.Nil(t, err)
 
 		expectedUnpaddedLength, err := codec.GetUnpaddedDataLength(uint32(len(paddedBytes)))
 		require.Nil(t, err)
 		require.Equal(t, expectedUnpaddedLength, uint32(len(unpaddedBytes)))
 
-		// unpadded payload may have up to 31 extra trailing zeros, since RemoveInternalPadding doesn't consider these
+		// unpadded payload may have up to 31 extra trailing zeros, since CheckAndRemoveInternalPadding
```
litt3: This comment about `CheckAndRemoveInternalPadding` no longer matches the actual function name

encoding/codec/codec.go
```diff
@@ -199,7 +209,7 @@ func GetPaddedDataLength(inputLen uint32) uint32 {
 
 // GetUnpaddedDataLength accepts the length of an array that has been padded with [PadPayload]
 //
-// It returns what the length of the output array would be if you called [RemoveInternalPadding] on it,
+// It returns what the length of the output array would be if you called [CheckAndRemoveInternalPadding] on it,
```
litt3: CheckAndRemoveInternalPadding doesn't match the actual name anymore

I suggest searching for other instances of stale naming, too

api/clients/v2/coretypes/encoded_payload.go
```diff
@@ -198,3 +197,36 @@ func fftSettingsFromBlobLengthSymbols(blobLengthSymbols uint32) *fft.FFTSettings
 	maxScale := uint8(gomath.Log2(float64(blobLengthSymbols)))
 	return fft.NewFFTSettings(maxScale)
 }
+
+// checkAndRemoveInternalPadding accepts an array of padded data, then checks and removes all the padding that was
+// added to the payload in order to make every 32 bytes a valid field element in the final output, and to make
+// the length of the output to hold a power of 2 field elements i.e (32, 64, 128,...) bytes.
+//
+// This function checks all the padding must be 0.
+func checkAndRemoveInternalPadding(paddedData []byte, payloadLen uint32) ([]byte, error) {
```
litt3: Why did you duplicate this logic here?

api/clients/v2/coretypes/encoded_payload.go
```diff
@@ -198,3 +197,36 @@ func fftSettingsFromBlobLengthSymbols(blobLengthSymbols uint32) *fft.FFTSettings
...
+		return nil, fmt.Errorf("padding check failed for ensuring every 32 bytes is a valid field element: %w", err)
+	}
+
+	// data length is checked when constructing an encoded payload. If this error is encountered, that means there
+	// must be a flaw in the logic at construction time (or someone was bad and didn't use the proper construction
+	// methods)
+	if uint32(len(decodedPayloadWithPadding)) < payloadLen {
+		return nil, fmt.Errorf(
+			"length of unpadded data %d is less than length claimed in encoded payload header %d."+
+				"this should neverhappen", uint32(len(decodedPayloadWithPadding)), payloadLen)
```
iquidus:
```suggestion
				"this should never happen", uint32(len(decodedPayloadWithPadding)), payloadLen)
```

api/clients/v2/coretypes/encoded_payload.go
bxue-l2: I moved this function under `checkAndRemoveInternalPadding`, the error is still returned.

api/clients/v2/coretypes/encoded_payload_test.go
bxue-l2: good catch 58cf717fbb25cbbe15dbfb68c50b0db10a9f1506

api/clients/v2/coretypes/encoded_payload.go
```diff
@@ -198,3 +197,36 @@ func fftSettingsFromBlobLengthSymbols(blobLengthSymbols uint32) *fft.FFTSettings
...
+	}
+
+	// ensure all the padding bytes in the encoded payload are zero. This is accomplished by first checking that the
+	// first byte of every multiple of 32 bytes is zero, and then verifying that all remaining bytes beyond the useful
+	// payload length are also zero with the following loop.
+	for _, b := range decodedPayloadWithPadding[payloadLen:] {
```
shyam-patel-kira: nit: using an index loop is more idiomatic for byte-by-byte checks imo

encoding/codec/test/codec_test.go
bxue-l2: nice catch, I forgot to adjust the comment, and there is another place, will do both

bxue-l2: 10689c79f8bc1d7bd9fa783e833fea636f7921cc

encoding/codec/codec.go
bxue-l2: reference are updated, 08f7bf76e4ca904f15c06a94af07331e795d0c0a

api/clients/v2/coretypes/encoded_payload.go
bxue-l2: This is not duplicated, I moved the padding removal into this function

api/clients/v2/coretypes/encoded_payload.go
bxue-l2: It is hard to decide, I asked chatgpt about it, range check is also recommended. Here is gpt's reason
```
Reads like intent: for _, v := range s says “for each value in this slice.” An index loop says “count from 0 to n.” Go style favors the former when you don’t need i.

Fewer footguns: no off-by-one mistakes, no accidental i++/i+=2, no out-of-bounds risk. Range handles length and bounds for you.

Short-circuit naturally: returning as soon as you see a non-zero is obvious and clean in a range loop.

Compiler-friendly: the compiler can eliminate bounds checks in both styles, but range already gives you elements without repeated indexing. For []byte, the copy per element is trivial.

Common practice: the standard library and effective Go examples use range when the index isn’t needed; reviewers expect it.

Avoids array pitfalls: ranging over a slice doesn’t copy the underlying data; ranging directly over a large array would copy the array. Slicing first and ranging is the idiom.
```

encoding/codec/codec.go
bxue-l2: I resolved it, because it has been long fixed

retriever/v2/server.go
bxue-l2: reverted https://github.com/Layr-Labs/eigenda/pull/2250.

For some reason, I thought I had to modify this code. But it is not required.

api/clients/v2/coretypes/encoded_payload.go
bxue-l2: It is also ok not to move it. I was originally thinking to expose it. But decided not to, because it bakes assumption on encoded payload. 

Now I think it is too many function, so I will move it back

api/clients/v2/coretypes/encoded_payload.go
bxue-l2: now, I removed the function `checkAndRemoveInternalPadding`, and the logics is back. The error is still returned. 1778a81a152ae1858c45422eb40d7236685d27fc

api/clients/v2/coretypes/encoded_payload.go
bxue-l2: 1778a81a152ae1858c45422eb40d7236685d27fc

shyam-patel-kira (approved):
LGTM

PR #2252, opened 2025-10-30
fix: chunk deduplication

cody-littley:

Fixes how the validator client de-duplicates chunks which are assigned to multiple validators. Adds a unit test that is sensitive to this bug (the test fails in master but passes in this branch).

### DIFF
--- a/api/clients/v2/validator/retrieval_worker.go
+++ b/api/clients/v2/validator/retrieval_worker.go
@@ -373,14 +373,14 @@ func (w *retrievalWorker) updateValidatorStatus(validatorId core.OperatorID, val
 		oldStatus, chunkHasStatus := w.chunkStatusMap[chunkIndex]
 		enforce.True(chunkHasStatus, "chunk %d has no status in chunkStatusMap", chunkIndex)
 
-		chunkOwner, hasOwner := w.chunkOwnerMap[chunkIndex]
+		currentChunkOwner, hasOwner := w.chunkOwnerMap[chunkIndex]
 		if !hasOwner {
 			// If this chunk has no owner, take ownership
 			w.chunkOwnerMap[chunkIndex] = validatorId
-			chunkOwner = validatorId
+			currentChunkOwner = validatorId
 		}
 
-		if validatorStatus.isBetterThan(oldStatus) || chunkOwner == validatorId {
+		if validatorStatus.isBetterThan(oldStatus) || currentChunkOwner == validatorId {
 			// There are two conditions where we update the chunk status:
 			// 1. The validator reporting the status change owns the chunk
 			// 2. The validator reporting the status change has a better status than the current chunk status


api/clients/v2/validator/validator_client_test.go
```diff
@@ -774,3 +777,211 @@ func MockCommitment(t *testing.T) encoding.BlobCommitments {
...
+	mockDeserializer.DeserializeAndVerifyFunction = func(
+		blobKey v2.BlobKey,
+		operatorID core.OperatorID,
+		getChunksReply *grpcnode.GetChunksReply,
+		blobCommitments *encoding.BlobCommitments,
+		encodingParams *encoding.EncodingParams,
+	) ([]*encoding.Frame, error) {
+
+		// verify we have the expected blob key
+		require.Equal(t, blobKey, blobKey)
```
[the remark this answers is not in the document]
cody-littley: fixed

api/clients/v2/validator/retrieval_worker.go
```diff
@@ -176,21 +195,26 @@ type retrievalWorker struct {
 	// Contains chunks that have been verified.
 	verifiedChunksQueue queues.Queue
 
-	// The status of the chunks from each validator. The key is the operator ID, and the value is the
-	// status of the chunks.
-	chunkStatusMap map[core.OperatorID]chunkStatus
+	// The status of the chunks from each validator. The key is the validator ID, and the value is the
+	// status for all chunks assigned to that validator.
+	validatorStatusMap map[core.OperatorID]chunkStatus
```
samlaf, line 201:
```suggestion
	// The status of the chunks from each validator.
	validatorStatusMap map[core.OperatorID]chunkStatus
```

samlaf: imo no need for this redundant info that is already encoded in the types.

api/clients/v2/validator/retrieval_worker.go
```diff
@@ -73,9 +74,15 @@ chunks to enter this state.
...
+// Chunk statuses are ordered from worst to best. Do not change the order of these statuses without understanding
+// the consequences.
 const (
+	// chunks that have failed to be downloaded or verified
+	failed chunkStatus = iota
+	// chunks that have timed of the pessimistic download timeout, but still may be downloaded successfully
+	pessimisticTimeout
 	// chunks that can be downloaded
-	available chunkStatus = iota
+	available
```
samlaf, line 86: what does it mean for pessimisticTimeout to be worse than available?

samlaf: aka should we really be modeling this as a total order given that its not? Guessing this is only to simplify the algorithm? Is it that much more complex to encode the proper state transition (according to ascii diagram above)? For example you mention below that a state can go backwards, but we know that a state can never go back to "available" state.

api/clients/v2/validator/retrieval_worker.go
```diff
@@ -317,40 +358,67 @@ func newRetrievalWorker(
...
+		)
 	}
-	w.chunkStatusCounts[status] += numChunks
 }
 
-// getChunkStatus returns the status of a chunk from a given operator.
+// getChunkStatus returns the status a validator's chunks.
 func (w *retrievalWorker) getChunkStatus(operatorID core.OperatorID) chunkStatus {
-	return w.chunkStatusMap[operatorID]
+	return w.validatorStatusMap[operatorID]
```
samlaf: why not change the name of the function to `getValidatorStatus`?

api/clients/v2/validator/retrieval_worker.go
```diff
@@ -317,40 +358,67 @@ func newRetrievalWorker(
...
 	}
 
-	if ok {
-		w.chunkStatusCounts[oldStatus] -= numChunks
+	if w.config.DetailedLogging {
+		w.logger.Debug("updating chunk status counts",
+			"validatorId", validatorId.Hex(),
+			"validatorStatus", chunkStatusStrings[validatorStatus],
+			"failed", w.chunkStatusCounts[failed],
+			"pessimisticTimeout", w.chunkStatusCounts[pessimisticTimeout],
```
samlaf, line 404: why do we need this DetailedLogging boolean? Isn't that the point of log levels? Is the point that slog doesn't have a trace and you would like to emit this log at the trace level instead? Perhaps we could extend our logger interface to add a trace level then?

api/clients/v2/validator/retrieval_worker.go
```diff
@@ -317,40 +358,67 @@ func newRetrievalWorker(
...
-		if !ok || (status == verified && oldStatus != verified) {
-			w.chunkOwner[index] = operatorID
+func (w *retrievalWorker) updateChunkStatus(validatorId core.OperatorID, validatorStatus chunkStatus) {
+	w.validatorStatusMap[validatorId] = validatorStatus
+
+	assignments, ok := w.assignments[validatorId]
+	if !ok {
+		// This validator has no assigned chunks
+		w.logger.Warnf("validator %s has no assigned chunks", validatorId.Hex())
+		return
```
samlaf, line 369: 1. this is possible because we chose an assignment where not necessarily all validators receive chunks right?
2. should we panic here instead? or are there reasons where the algorithm would actually call this on a validator that has no chunks..?

api/clients/v2/validator/retrieval_worker.go
```diff
@@ -317,40 +358,67 @@ func newRetrievalWorker(
...
-	numChunks := 0
-	for _, index := range w.assignments[operatorID].Indices {
-		_, ok := w.chunkOwner[index]
-		// If the new status is verified and the oldStatus is anything else, we move the chunk to the current operator
-		// as a new owner.
-		// TODO(@cody-littley): When setting a chunk to failed, if there is any other validator assigned to the chunk with a non-failed status,
-		// we should shift ownership to that validator.
-		if !ok || (status == verified && oldStatus != verified) {
-			w.chunkOwner[index] = operatorID
+func (w *retrievalWorker) updateChunkStatus(validatorId core.OperatorID, validatorStatus chunkStatus) {
```
samlaf: nit: should this be renamed `updateValidatorStatus`, or `updateValidatorChunkStatus`?

api/clients/v2/validator/retrieval_worker.go
```diff
@@ -317,40 +358,67 @@ func newRetrievalWorker(
...
+	if !ok {
+		// This validator has no assigned chunks
+		w.logger.Warnf("validator %s has no assigned chunks", validatorId.Hex())
+		return
+	}
+
+	for _, chunkIndex := range assignments.Indices {
+		oldStatus, chunkHasStatus := w.chunkStatusMap[chunkIndex]
+		if !chunkHasStatus {
+			oldStatus = available
```
samlaf: why? should we check that the `validatorStatus` argument is `available` to prevent bugs? Or why is it ok to just give it the available status like this?

samlaf: oh I see this chunk could already be in downloading state wrt another validator, so we just put this validator in "available" state in case the other validator's download times out or something?

Could you describe this in a comment?

api/clients/v2/validator/retrieval_worker.go
```diff
@@ -317,40 +358,67 @@ func newRetrievalWorker(
...
+			// If this chunk has no owner, take ownership
+			w.chunkOwnerMap[chunkIndex] = validatorId
+			chunkOwner = validatorId
+		}
+
+		if validatorStatus.isBetterThan(oldStatus) || chunkOwner == validatorId {
+			// There are two conditions where we update the chunk status:
+			// 1. The validator reporting the status change owns the chunk
+			// 2. The validator reporting the status change has a better status than the current chunk status
+			//    (it will "steal" ownership of the chunk in this case)
```
samlaf: what happens if validator owns the chunk but fct is called with worse status? then this for loop will just silently terminate. Would we want to panic if this branch was not taken? Or are there reasons for this branch not being taken?

samlaf (commented):
Leaving a bunch of nits/questions. Most of the code makes sense when reading linearly, but have def not understood how the global algorithm works. Also not sure what the bug was and what the fix is tbh... could you describe it in the PR description or give a TLDR? The new `updateChunkStatus` code seems quite different from the previous code.

api/clients/v2/validator/retrieval_worker.go
cody-littley: We need this information when checking for pessimistic timeouts. 

```
		if w.validatorStatusMap[operatorID] != downloading {
			// The operator has finished downloading, we can remove it from the queue.
			w.downloadsInProgressQueue.Dequeue()
			continue
		}
```

api/clients/v2/validator/retrieval_worker.go
cody-littley: We want to allow a long time period to download the data, in case there is network congestion. That maximum amount of time could be very long, perhaps even longer than a minute.

We also don't want to download all chunks, since downloading all chunks results in ~8x data redundancy.

Suppose a validator is slow to provide chunks. It hasn't failed yet or fully timed out, but it's starting to be a suspicious amount of time. Perhaps something like 10+ seconds. At that point in time, we change the blob status to `pessimisticTimeout`. We don't cancel the download when this happens, but we may schedule additional downloads in the hope that the additional downloads will produce data faster.

The complicated part about modeling this is not the state machine visualized above. It's actually the fact that we have N state machines layered on top of each other, one for each validator. Since a chunk may be assigned to multiple validators, that chunk will actually have multiple simultaneous states. The simplification is to only keep track of the state of each chunk for whichever validator has produced the "best" state so far. Sometimes the state we are tracking can go backwards (for example, if a connection gets broken). But if later, a different validator that is assigned the same chunk can "revive" the chunk with a non-failure status.

api/clients/v2/validator/retrieval_worker.go
cody-littley: I've removed this method, since referencing the map inline is easy and clean. I agree this function wasn't named well.

api/clients/v2/validator/retrieval_worker.go
cody-littley: The core problem is that all of our loggers in all environments seem to have debug mode turned on by default. These logs are super spammy, and so we really only want to turn them on when we are truly debugging a problem. If we want to avoid hacks like this, we either need a trace level like you suggest, or we need to default to only showing info level logs for most things.

api/clients/v2/validator/retrieval_worker.go
cody-littley: I hit this edge case in a unit test when I was manually creating unrealistic assignments to trigger a bug. I'm not sure if this can actually happen with the current assignments logic (since that code is pretty hard to follow). If it does happen to some low stake validator in practice, I don't think it harms anything to just log something. If chunk assignment is truly broken, we won't be able to reconstruct the blob. No need to panic prematurely.

api/clients/v2/validator/retrieval_worker.go
cody-littley: changed to `updateValidatorStatus()`

api/clients/v2/validator/retrieval_worker.go
cody-littley: This should be impossible to trigger, since we set the status of each chunk to `available` in the constructor. 

I added an assertion that will cause us to panic if we don't find a status:
```
		enforce.True(chunkHasStatus, "chunk %d has no status in chunkStatusMap", chunkIndex)
```

api/clients/v2/validator/retrieval_worker.go
cody-littley: It is completely legimate for this branch not to be taken. Consider the following scenario:

- Chunk `C` is assigned to validator `V1` and `V2`
- Validator `V1` is chosen first, and chunk `C` advances from `available` -> `downloading` -> `downloaded`
    - at this point in time, `V1` "owns" chunk `C` 
- Some time later, validator `V2` transitions from `available` -> `downloading`. Since it doesn't own the chunk and since these statuses are "worse" than `downloaded`, the chunk status doesn't change.
- Suppose there is a network disruption, and the connection with `V2` breaks. `V2` owned chunks should transition to status `failed`. But because validator `V2` doesn't currently own chunk `C`, the status of chunk `C` remains unchanged and `V1` continues to own it.

cody-littley: Funnily enough, the bug this PR is patching was due to this branch always being taken when the status transitions to `validated`, due to the wrong thing being compared. This resulted in chunks assigned to multiple validators to be double counted.

api/clients/v2/validator/retrieval_worker.go
samlaf: was just talking about the part of the comment that I removed in the suggested change!

api/clients/v2/validator/retrieval_worker.go
samlaf: cc @dmanc

dmanc: I probably would prefer adding trace level to our logger. I feel in principle we should be running in `info` or `warn` modes in production but a lot of our debug logs feel useful to have enabled during an incident is my current feeling.

dmanc: I don't think it should block this PR though

api/clients/v2/validator/retrieval_worker.go
```diff
@@ -315,42 +357,62 @@ func newRetrievalWorker(
...
+		chunkOwner, hasOwner := w.chunkOwnerMap[chunkIndex]
+		if !hasOwner {
+			// If this chunk has no owner, take ownership
+			w.chunkOwnerMap[chunkIndex] = validatorId
+			chunkOwner = validatorId
 		}
-		if w.chunkOwner[index] == operatorID {
-			numChunks++
+
+		if validatorStatus.isBetterThan(oldStatus) || chunkOwner == validatorId {
```
samlaf: nit: would be easier to read if the nomenclature here was more consistent wrt status and owner, ie either:
```
if validatorStatus.isBetterThan(chunkStatus) || chunkOwner == validatorId
```
or
```
if validatorStatus.isBetterThan(oldChunkStatus) || validatorId == oldChunkOwner
```

samlaf: aka chunkStatus/chunkOwner or oldChunkStatus/oldChunkOwner, but not chunkOwner/oldStatus

samlaf (approved):
LGTM. Thanks for running me through the code, made it a lot easier to understand this second time around. Left one last nit but otherwise LGTM!

api/clients/v2/validator/retrieval_worker.go
cody-littley: renamed to `currentChunkOwner`. I think "current" fits better than "old", since we may not end up replacing the old owner.

PR #2254, opened 2025-10-30
fix(sigmap-timeout): Add timeout to `CheckDACert` call

iquidus:
Adds missing timeout to CheckDACert call 

* Adds `contractCallTimeout` to `Store` type
* Updates `NewStore` constructor to set `contractCallTimeout`
* Calls `CheckDACert` with a context with timeout
* Uses timeout value set by `--eigenda.v2.contract-call-timeout` flag

### DIFF
--- a/api/proxy/store/generated_key/v2/eigenda.go
+++ b/api/proxy/store/generated_key/v2/eigenda.go
@@ -45,6 +45,9 @@ type Store struct {
 
 	// Retrieval related fields.
 	retrievers []clients.PayloadRetriever
+
+	// Timeout used for contract calls
+	contractCallTimeout time.Duration
 }
 
 var _ common.EigenDAV2Store = (*Store)(nil)
@@ -56,6 +59,7 @@ func NewStore(
 	certVerifier *verification.CertVerifier,
 	rbnRecencyWindowSize uint64,
 	retrievers []clients.PayloadRetriever,
+	contractCallTimeout time.Duration,
 ) (*Store, error) {
 	if putTries == 0 {
 		return nil, fmt.Errorf(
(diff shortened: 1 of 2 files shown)


api/proxy/store/generated_key/v2/eigenda.go
```diff
@@ -264,8 +269,11 @@ func (e Store) VerifyCert(ctx context.Context, versionedCert certs.VersionedCert
 		return err
 	}
 
+	timeoutCtx, cancel := context.WithTimeout(ctx, e.contractCallTimeout)
+	defer cancel()
+
 	// verify cert via simulation call to verifier contract
-	err = e.certVerifier.CheckDACert(ctx, sumDACert)
+	err = e.certVerifier.CheckDACert(timeoutCtx, sumDACert)
```
samlaf, line 276: interesting. Don't really like passing a `contractCallTimeout` to checkDACert here because checkDACert is actually making 2 eth Calls, so the 5sec default timeout here is actually used for the entirety of CheckDACert whereas it should probably be used on a per-call basis. In either case given that the payload disperser is doing the exact same thing its prob fine to merge.

I think an ideal solution would prob have the network call timeouts on the eth-client itself instead of the higher-level structs, but that would be a much larger refactor.

samlaf (approved):
LGTM but would like to get @litt3 to take a look as well before we merge.

Also other random thought while reading the CertVerifier implementation in the context of our recencyWindow discussion. Realized that it would have been nice to breakup the CertVerifier struct into one used in the GET path and another in the POST path, to make the distinction clearer.

api/proxy/store/generated_key/v2/eigenda.go
litt3: The timeout logic in the `PayloadDisperser` ought to have been revisited when the router was put in place. This is another thing that would be good to clean up, whenever the `CertVerifier` is cleaned up

PR #2255, opened 2025-10-30
chore(payments): Enable validator payment validation in inabox

litt3:
- This PR enables validator payment verification in inabox payment tests

### DIFF
--- a/inabox/tests/setup_operator_harness.go
+++ b/inabox/tests/setup_operator_harness.go
@@ -19,6 +19,7 @@ import (
 	"github.com/Layr-Labs/eigenda/core"
 	coreeth "github.com/Layr-Labs/eigenda/core/eth"
 	"github.com/Layr-Labs/eigenda/core/eth/directory"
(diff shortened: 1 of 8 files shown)


inabox/tests/setup_operator_harness.go
```diff
@@ -19,6 +19,7 @@ import (
 	"github.com/Layr-Labs/eigenda/core"
 	coreeth "github.com/Layr-Labs/eigenda/core/eth"
 	"github.com/Layr-Labs/eigenda/core/eth/directory"
+	"github.com/Layr-Labs/eigenda/core/payments/reservation/reservationvalidation"
```
dmanc, line 22: nit: Not related to this PR but it seems weird to see reservation twice here? Would `github.com/Layr-Labs/eigenda/core/payments/reservation/validation` also work?

dmanc (approved):

inabox/tests/setup_operator_harness.go
litt3: Yeah, that would probably be better. I might do a pass at some point and clean up package names within `payments`, since some of them are iffy

PR #2256, opened 2025-10-30
refactor(apiserver): Inject listener into API server constructor

dmanc:

Allows us to inject listeners that are already created into the API server when constructing it. Also updates the payments test harness to support the OS-allocated ports exposed by the disperser harness.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/apiserver/server_v2.go
+++ b/disperser/apiserver/server_v2.go
@@ -450,18 +450,13 @@ func (s *DispersalServerV2) getPaymentState(
 // Gracefully shuts down the server and closes any open connections
 func (s *DispersalServerV2) Stop() error {
 	if s.grpcServer != nil {
+		// GracefulStop will close the listener that was passed to Serve()
 		s.grpcServer.GracefulStop()
-	}
-
-	if s.listener != nil {
-		if err := s.listener.Close(); err != nil {
-			s.logger.Error("failed to close listener", "error", err)
-		}
 	}
 
 	if s.controllerConnection != nil {
 		if err := s.controllerConnection.Close(); err != nil {
-			s.logger.Error("failed to close controller connection", "error", err)
+			return fmt.Errorf("failed to close controller connection: %w", err)
 		}
 	}
 	return nil


disperser/cmd/apiserver/lib/apiserver.go
```diff
@@ -148,6 +149,13 @@ func RunDisperserServer(ctx *cli.Context) error {
 		})
 		blobStore := blobstorev2.NewBlobStore(bucketName, objectStorageClient, logger)
 
+		// Create listener for the gRPC server
+		addr := fmt.Sprintf("%s:%s", disperser.Localhost, config.ServerConfig.GrpcPort)
```
samlaf:
```suggestion
		addr := fmt.Sprintf("0.0.0.0:%s", config.ServerConfig.GrpcPort)
```

samlaf: yuck can we get rid of this `Localhost = "0.0.0.0"` variable? Its not needed and a disturbingly bad name.

inabox/tests/setup_disperser_harness.go
```diff
@@ -956,16 +959,28 @@ func startController(
...
-			30000, // TODO(dmanc): Extract port from listener instead when implementing dynamic allocation
+			uint16(assignedPort),
 			1024*1024,
 			5*time.Minute,
 			5*time.Minute,
 			3*time.Minute,
 		)
 		if err != nil {
+			_ = listener.Close()
 			_ = logFile.Close()
```
samlaf: one cleaner pattern is to define the error variable at the top, and then have a defer statement that checks if the err was assigned to, and cleans up if so

disperser/apiserver/server_v2.go
```diff
@@ -78,6 +78,10 @@ type DispersalServerV2 struct {
 	// Controller gRPC client connection and service client
 	controllerConnection *grpc.ClientConn
 	controllerClient     controller.ControllerServiceClient
+
+	// Pre-created listener for the gRPC server
+	listener   net.Listener
+	grpcServer *grpc.Server
```
samlaf, line 84: personally don't like having a struct field that is populated by Start(), meaning that constructor is not fully populating a struct. But don't see a way out of this one given you've exposed the Stop() function. At some point we need to rethink and unify these abstractions but ok for now I guess

samlaf (approved):

disperser/apiserver/server_v2.go
```diff
@@ -446,6 +449,16 @@ func (s *DispersalServerV2) getPaymentState(
 
 // Gracefully shuts down the server and closes any open connections
 func (s *DispersalServerV2) Stop() error {
```
litt3, line 451: Seems odd to have an `error` return value, but this method just logs errors and always returns `nil`

inabox/tests/test_harness.go
```diff
@@ -108,6 +107,9 @@ type TestHarness struct {
 
 	// Chain ID for this test context
 	ChainID *big.Int
+
+	// API Server V2 address for the disperser
+	APIServerV2Address string
```
litt3: Does this need the `V2` suffix?

In general, I think we should shoot to remove it entirely from the codebase once V1 is gone. Since that's imminent, I've personally been trying to avoid it unless absolutely necessary

disperser/apiserver/server_v2_test.go
```diff
@@ -569,6 +569,10 @@ func newTestServerV2(t *testing.T) *testComponents {
 		},
 	}, nil)
 
+	// Create listener for test server
+	listener, err := net.Listen("tcp", "0.0.0.0:0")
+	require.NoError(t, err)
+
 	s, err := apiserver.NewDispersalServerV2(
 		disperser.ServerConfig{
 			GrpcPort:    "51002",
```
litt3, line 578: This port number isn't correct. It doesn't impact the test, but for clarity, it's probably best to get whatever port was automatically assigned by the OS

This exposes a bit of a smell in the config flow, IMO. `NewDispersalServerV2` accepts a `ServerConfig` instance, and even validates that `GrpcPort` is valid. But that field isn't actually used anywhere within `NewDispersalServerV2`

This is in the same vein as [this comment](https://github.com/Layr-Labs/eigenda/pull/2256#discussion_r2479830632) from @samlaf. The smell could be improved with some work on the abstractions in this system, but no immediate action required

litt3 (approved):

disperser/cmd/apiserver/lib/apiserver.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2256/commits/1ba6ce3346a06e10eacbdbd276232c96f1c13c31

inabox/tests/test_harness.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2256/commits/1b741a3097fe39abff465ea10f0f37a9f8be7fea reduced the amount of V2 comments/naming.

inabox/tests/setup_disperser_harness.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2256/commits/91f8f15f176e6cad63c797f172ce7d4888ae93d0

disperser/apiserver/server_v2.go
dmanc: Fixed https://github.com/Layr-Labs/eigenda/pull/2256/commits/981fcb638876df1a03fadfdd5c5ef6dbb97e37d5

PR #2262, opened 2025-10-31
feat(payments): Make new payments the default

litt3:
- New payments are slated for the next release, so it's time to turn the system on by default
- In addition to setting new payments as the default, this PR makes some changes to how the API server is instantiated: these changes were necessary due to how the unit tests are written


inabox/tests/setup_disperser_harness.go
```diff
@@ -1211,6 +1214,26 @@ func startAPIServer(
 	// Onchain state refresh interval
 	onchainStateRefreshInterval := 1 * time.Second
 
+	// Create controller client if using new payments
+	var controllerConnection *grpc.ClientConn
+	var controllerClient grpccontroller.ControllerServiceClient
+	if config.TestConfig.UseNewPayments {
+		if controllerAddress == "" {
+			_ = logFile.Close()
```
samlaf: dont think this is needed anymore, since there's a defer statement that closes it already.

inabox/tests/setup_disperser_harness.go
```diff
@@ -1211,6 +1214,26 @@ func startAPIServer(
...
+			return nil, fmt.Errorf("controller address is empty but UseNewPayments is true")
+		}
+		connection, err := grpc.NewClient(
+			controllerAddress,
+			grpc.WithTransportCredentials(insecure.NewCredentials()),
+		)
+		if err != nil {
+			_ = logFile.Close()
```
samlaf: also not needed afaiu

samlaf (approved):

inabox/tests/setup_disperser_harness.go
litt3: Bad merge. When I implemented this, that `defer` didn't exist yet. Will fix.

litt3: c8ae300801af9b21f7b8aa6234b92eb618d08aec

inabox/tests/setup_disperser_harness.go
litt3: c8ae300801af9b21f7b8aa6234b92eb618d08aec

api/proxy/common/client_config_v2.go
```diff
@@ -56,6 +56,11 @@ type ClientConfigV2 struct {
 	EigenDANetwork EigenDANetwork
 
 	// Determines which payment mechanism to use
+	//
+	// There is no default value for this field: users must explicitly state the type of payments they wish to use.
+	// Though it would be potentially possible to make some inferences about what the user likely wants to use based on
+	// PaymentVault state, this is risky since a misconfiguration could, for example, cause a user to burn through
+	// backup on-demand funds when they actually wanted to be using a reservation.
 	ClientLedgerMode clientledger.ClientLedgerMode
```
cody-littley: Would it make sense to default to reservation only?

- Most existing customers are currently using reservations, right?
- If a customer has a reservation that's available, they probably always want to use it before using on-demand funds. So I think the consequences of accidentally having this value when you wanted something else aren't extreme... you just use your reservation, which goes away anyways if you don't use it.
- Would allow this to be backwards compatible, and not require existing users to modify their config.

cody-littley (dismissed):

api/proxy/common/client_config_v2.go
litt3: That's a good point... I'll go ahead and make this change

litt3: Done in a2799212511a52f57b60c0c51986a7588d0b23bd

samlaf (approved):

PR #2264, opened 2025-11-01
perf: remove unneeded equivalence check and use universal verification

bxue-l2:

The validator retrieval is the bottleneck for read amplification. CPU consumption might be the reason.

This PR improves the performance by two changes
1. remove the unnecessary equivalence check. It is safe because it only does it to check if G2 commitment is equal to the G1 commitment, which is the job for EigenDA operator. For validator client, it does not need G2 commitment, so there is no need to verify it.
2. replace the VerifyFrame function to `UniversalVerifySubBatch` function that has far superior speed. Intuitively, for a 2MB blob, with coding rate 1/8 and 2^13=8192 chunks. Each chunk has size 2^24/2^5/2^13 = 2^6=64field elements. The reconstruction requires 1024 chunks. Almost all the verification cost is MSM (i,e taking inner product of G1 points and field elements). In the old methods, there 1024*64= 65536 unit of work. But with universal verification, it only takes about 1024+64=1088 units of works.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/validator/internal/chunk_deserializer.go
+++ b/api/clients/v2/validator/internal/chunk_deserializer.go
@@ -51,7 +51,7 @@ func NewChunkDeserializer(
 }
 
 // assume all the chunks come from one blob. In theory, universal verification
-// work as long as alll chunk lengths are equal, we can find the right root of
+// works as long as all chunk lengths are equal, and we can find the right root of
 // unities.
 func (d *chunkDeserializer) DeserializeAndVerify(
 	_ v2.BlobKey, // used for unit tests
@@ -89,12 +89,12 @@ func (d *chunkDeserializer) DeserializeAndVerify(
 	}
 
 	// verify all chunks for operator using universal verification, it reduces the complexity from
-	// n*m to n + m, where n is the number of chunk, and m is the length of chunk in field elements
+	// n*m to n + m, where n is the number of chunks, and m is the length of each chunk in field elements
 	// For theory, see https://ethresear.ch/t/a-universal-verification-equation-for-data-availability-sampling/13240
 	err := d.verifier.UniversalVerifySubBatch(
 		*encodingParams,
 		samples,
-		1, // only verify for one blob
+		1, // only verify one blob
 	)
 	if err != nil {
 		return nil, fmt.Errorf("failed to verify chunks from operator %s: %w", operatorID.Hex(), err)


api/clients/v2/validator/internal/chunk_deserializer.go
```diff
@@ -75,7 +78,24 @@ func (d *chunkDeserializer) DeserializeAndVerify(
...
+		samples[ind] = encoding.Sample{
+			Commitment:      blobCommitments.Commitment,
+			Chunk:           chunks[ind],
+			AssignmentIndex: assignmentIndices[ind],
+			BlobIndex:       0, // there is only 1 blob
+		}
+	}
+
+	// verify all chunks for operator using universal verification, it reduces the complexity from
+	// n*m to n + m, where n is the number of chunk, and m is the length of chunk in field elements
```
cody-littley:
```suggestion
	// n*m to n + m, where n is the number of chunks, and m is the length of each chunk in field elements
```

api/clients/v2/validator/internal/chunk_deserializer.go
```diff
@@ -50,6 +50,9 @@ func NewChunkDeserializer(
 	}
 }
 
+// assume all the chunks come from one blob. In theory, universal verification
+// work as long as alll chunk lengths are equal, we can find the right root of
```
cody-littley:
```suggestion
// works as long as all chunk lengths are equal, and we can find the right root of
```

api/clients/v2/validator/internal/chunk_deserializer.go
```diff
@@ -75,7 +78,24 @@ func (d *chunkDeserializer) DeserializeAndVerify(
...
+		}
+	}
+
+	// verify all chunks for operator using universal verification, it reduces the complexity from
+	// n*m to n + m, where n is the number of chunk, and m is the length of chunk in field elements
+	// For theory, see https://ethresear.ch/t/a-universal-verification-equation-for-data-availability-sampling/13240
+	err := d.verifier.UniversalVerifySubBatch(
+		*encodingParams,
+		samples,
+		1, // only verify for one blob
```
cody-littley:
```suggestion
		1, // only verify one blob
```

litt3 (approved):
LGTM from a code perspective

samlaf (approved):

PR #2265, opened 2025-11-02
fix: gpu memory leak by freeing them

bxue-l2:

This PR fixes a bug that does not free gpu memory. This PR also moves the msm computation into the main function, as it is cumbersome to return a device pointer with memory allocated to it, but not freed in the same function

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
+++ b/encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
@@ -165,12 +165,12 @@ func (p *KzgMultiProofBackend) ComputeMultiFrameProofV2(ctx context.Context, pol
 
 	end := time.Now()
 
-	p.Logger.Info("Multiproof Time Decomp (ms)",
+	p.Logger.Info("Multiproof Time Decomp (microseconds)",
 		"total", end.Sub(begin).Milliseconds(),
-		"preproc", preprocessDone.Sub(begin).Milliseconds(),
-		"msm", msmDone.Sub(preprocessDone).Milliseconds(),
-		"fft1", firstECNttDone.Sub(msmDone).Milliseconds(),
-		"fft2", secondECNttDone.Sub(firstECNttDone).Milliseconds(),
+		"preproc", preprocessDone.Sub(begin).Microseconds(),
+		"msm", msmDone.Sub(preprocessDone).Microseconds(),
+		"fft1", firstECNttDone.Sub(msmDone).Microseconds(),
+		"fft2", secondECNttDone.Sub(firstECNttDone).Microseconds(),
 	)
 
 	return proofs, nil
--- a/encoding/v2/kzg/prover/prover.go
+++ b/encoding/v2/kzg/prover/prover.go
@@ -197,12 +197,12 @@ func (e *Prover) GetFrames(
 			len(encodeResult.chunks), len(proofs))
 	}
 
-	e.logger.Info("Frame process details (milliseconds)",
+	e.logger.Info("Frame process details (microseconds)",
 		"input_size_bytes", len(inputFr)*encoding.BYTES_PER_SYMBOL,
 		"num_chunks", params.NumChunks,
 		"chunk_length", params.ChunkLength,
-		"rs_encode_duration", encodeResult.duration.Milliseconds(),
-		"multi_proof_duration", getProofsDuration.Milliseconds(),
+		"rs_encode_duration", encodeResult.duration.Microseconds(),
+		"multi_proof_duration", getProofsDuration.Microseconds(),
 	)
 
 	frames := make([]*encoding.Frame, len(proofs))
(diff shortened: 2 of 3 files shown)


encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
```diff
@@ -290,7 +283,9 @@ func (c *KzgMultiProofBackend) twoEcnttOnDevice(
 	// we only allocate one large gpu memory for all operation, so it has to be large enough to cover all cases
 	// including the first and the second ECNTT
 	var bufferProjectivePointsOnDevice core.DeviceSlice
-	var secondECNTTDeviceSlice core.DeviceSlice
+	// free intermediate GPU memory
+	defer bufferProjectivePointsOnDevice.Free()
```
dmanc: Shouldn't we defer after malloc succeeds, otherwise we will call free on unitialized mem right

dmanc (approved):

encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
bxue-l2: f1ec18a258e23dea4cddf526b90581116c0f810e

encoding/v2/kzg/prover/prover.go
```diff
@@ -197,12 +197,12 @@ func (e *Prover) GetFrames(
 			len(encodeResult.chunks), len(proofs))
 	}
 
-	e.logger.Info("Frame process details",
+	e.logger.Info("Frame process details (milliseconds)",
```
dmanc: nit: kinda prefer nanoseconds

dmanc (approved):

encoding/v2/kzg/prover/prover.go
bxue-l2: How about microsecond, I literally cannot count digits

samlaf: Why convert the duration at all? Just let golang stringify with units no? https://goplay.tools/snippet/WQuNq5foh6R
image

encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
```diff
@@ -119,9 +119,28 @@ func (p *KzgMultiProofBackend) ComputeMultiFrameProofV2(ctx context.Context, pol
...
-			icicleErr = fmt.Errorf("msm error: %w", err)
+		var projectivePoint iciclebn254.Projective
+		var sumVec core.DeviceSlice
+
+		_, mallocErr := sumVec.Malloc(projectivePoint.Size(), int(toeplitzMatrixLen)*2)
+		if mallocErr != runtime.Success {
+			icicleErr = fmt.Errorf("allocating bytes on device failed: %v", mallocErr.AsString())
+			return
+		}
+		defer sumVec.Free()
```
samlaf, line 130: did you just inline the msmBatchOnDevice function just to call defer here? You could have kept the function and called defer on the returned value no?

encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
```diff
@@ -141,12 +165,12 @@ func (p *KzgMultiProofBackend) ComputeMultiFrameProofV2(ctx context.Context, pol
...
-		"preproc", preprocessDone.Sub(begin),
-		"msm", msmDone.Sub(preprocessDone),
-		"fft1", firstECNttDone.Sub(msmDone),
-		"fft2", secondECNttDone.Sub(firstECNttDone),
+	p.Logger.Info("Multiproof Time Decomp (ms)",
+		"total", end.Sub(begin).Milliseconds(),
+		"preproc", preprocessDone.Sub(begin).Milliseconds(),
+		"msm", msmDone.Sub(preprocessDone).Milliseconds(),
+		"fft1", firstECNttDone.Sub(msmDone).Milliseconds(),
+		"fft2", secondECNttDone.Sub(firstECNttDone).Milliseconds(),
```
samlaf: dont call Milliseconds()

samlaf (dismissed):
Think we should be more explicit about ownership of pointers instead of using defer() statements. defer statements for memory allocation works but it's not very "precise" imo.

Also according to Ingonyama guy we might still have another memory leak somewhere else (check tg). Also iiuc what he's saying we need to do all malloc/free operations async on the same stream as we do for the rest, or else add stream synchronization barriers, otherwise we have races.

If this code is working I guess we can merge but we'll need to fix these other issues.

encoding/v2/kzg/prover/prover.go
bxue-l2: the old code is printing the duration, but the log shows a 8 digit numbers, very hard to tell how fast. I think it has something to do with the Logger we usedd

bxue-l2: going to stick with microsecond

PR #2266, opened 2025-11-03
feat(proxy): Arb CustomDA Recency Check

ethenotethan:
embeds the recency check into the ARB CustomDA Server by overloading the existing recency check logic present in the `EigenDAV2Store`. This check is targeted via the optional `GetParams` object passed through the retrieval code path and targeted via the CustomDA Server `daprovider_retrieve` RPC method. 

the E2E tests ignore the recency check since the L1 inclusion block number cannot be reasonably calculated given lack of rollup interaction with a bypass value of `0`. 

**Follow ups**
- Test recency correctness E2E by spinning up a testnet deployment of ARB CustomDA x EigenDA 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


api/proxy/test/mocks/eth_client.go
```diff
@@ -0,0 +1,58 @@
...
+// EXPECT returns an object that allows the caller to indicate expected use.
+func (m *MockIEthClient) EXPECT() *MockIEthClientMockRecorder {
+	return m.recorder
+}
+
+// BlockByHash mocks base method.
+func (m *MockIEthClient) BlockByHash(ctx context.Context, hash common.Hash) (*types.Block, error) {
+	m.ctrl.T.Helper()
+	ret := m.ctrl.Call(m, "BlockByHash", ctx, hash)
+	ret0, _ := ret[0].(*types.Block)
```
bxue-l2, line 49: It has been a while, so not familar with gomock logics now. How does it mock module work? It seems like the file containing `NewMockEthClient` below set the mock behavior, but I didn't see where it is used.

bxue-l2 (approved):

api/proxy/test/mocks/eth_client.go
ethenotethan: > How does it mock module work?

it's being used in `handlers_test.go` to stub rpc calls. this file is autogenerated from the `IEthClient` interface's signature to create a stub type that allows for setting arbitrary return behaviors

PR #2268, opened 2025-11-04
fix(encoding): Fix encoding benchmarks and add the benchmarks to CI

dmanc:

Fixes an error introduced in https://github.com/Layr-Labs/eigenda/commit/085b9cbbebf430322d134f0228c923a1636a556a that caused the benchmark tests to fail. This PR also adds the benchmark tests to CI so we can catch failures in CI.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/encoding/v2/bench/benchmark_eigenda_test.go
+++ b/encoding/v2/bench/benchmark_eigenda_test.go
@@ -239,18 +239,20 @@ func BenchmarkFrameGenerationIcicle(b *testing.B) {
 		b.Skip("code compiled without the icicle build tag")
 	}
 	encodingConfig := encoding.Config{
-		NumWorker:   uint64(runtime.GOMAXPROCS(0)),
-		BackendType: encoding.IcicleBackend,
-		GPUEnable:   true,
+		NumWorker:                             uint64(runtime.GOMAXPROCS(0)),
+		BackendType:                           encoding.IcicleBackend,
+		GPUEnable:                             true,
+		GPUConcurrentFrameGenerationDangerous: 20,
 	}
 	benchmarkFrameGeneration(b, encodingConfig)
 }
 
 func BenchmarkFrameGenerationGnark(b *testing.B) {
 	encodingConfig := encoding.Config{
-		NumWorker:   uint64(runtime.GOMAXPROCS(0)),
-		BackendType: encoding.GnarkBackend,
-		GPUEnable:   false,
+		NumWorker:                             uint64(runtime.GOMAXPROCS(0)),
+		BackendType:                           encoding.GnarkBackend,
+		GPUEnable:                             false,
+		GPUConcurrentFrameGenerationDangerous: 20,
 	}
 	benchmarkFrameGeneration(b, encodingConfig)
 }


.github/workflows/benchmark-tests.yml
```diff
@@ -0,0 +1,76 @@
+name: benchmark-tests
+
+# TODO: Implement benchstat comparison workflow to catch performance regressions
+# This would involve:
+# 1. Running benchmarks on base branch
+# 2. Running benchmarks on PR branch
+# 3. Using benchstat to compare results
```
samlaf, line 7: we could also just commit the benchmark results to a file in repo and compare ran benchmarks to those. that was my original intention. But honestly I think these ideas all suck, because even for benchstat we'd need to parse the result table using some ad-hoc shell script, then do some random math to accept some error bound (like we dont want to call a regression if its only 1ms slower), etc.

Not sure building this is worth the pain, nor the maintenance burden, unless we find some better way to manage this automatically.

.github/workflows/benchmark-tests.yml
```diff
@@ -0,0 +1,76 @@
...
+
+# TODO: Add icicle benchmarks to this workflow (requires GPU runners).
+
+on:
+  pull_request:
+    paths:
+      - 'encoding/v2/**'
+      - 'common/**'
+      - 'api/clients/codecs/**'
+      - 'api/clients/v2/coretypes/**'
```
samlaf, line 18: llm generated? worth adding a comment that this is kinda arbitrary dependency list.

encoding/v2/bench/Makefile
```diff
@@ -5,7 +5,7 @@ bench_all: bench_primitives bench_eigenda
 # Our benchmarks currently only use up to coset32, however, so if download is slow feel free to ctrl-c early.
 # TODO(samlaf): now that we have this we can increase encoded_blob sizes benchmarked against up to 128MiB (8*16MiB).
 download_srs_tables:
-	cd ../../../tools/srs-utils && go run ./cmd/main.go download-tables --dimension dimE8192 --output-dir ../../resources/srs/SRSTables
+	cd ../../../tools/srs-utils && go run ./cmd/main.go download-tables --dimension dimE1024 --output-dir ../../resources/srs/SRSTables
```
samlaf, line 8: nice. do we need all the cosets that it downloads as well?

samlaf: I guess all those are small enough that it doesn't really matter?

encoding/v2/bench/benchmark_eigenda_test.go
```diff
@@ -212,28 +212,52 @@ func benchmarkMultiproofGeneration(b *testing.B, encodingConfig encoding.Config)
...
+// TODO(samlaf): maybe move this to benchmark_icicle_test.go file?
+// That file is currently metal only, we should generalize it.
+func BenchmarkFrameGenerationIcicle(b *testing.B) {
+	if !icicle.IsAvailable {
+		b.Skip("code compiled without the icicle build tag")
+	}
+	encodingConfig := encoding.Config{
+		NumWorker:   uint64(runtime.GOMAXPROCS(0)),
+		BackendType: encoding.IcicleBackend,
+		GPUEnable:   true,
```
samlaf: you deleted the `GPUConcurrentFrameGenerationDangerous: 20` field. prob want to readd to at least show users that they can change this. Maybe commented if you dont want to instantiate it.

encoding/v2/bench/benchmark_eigenda_test.go
```diff
@@ -212,28 +212,52 @@ func benchmarkMultiproofGeneration(b *testing.B, encodingConfig encoding.Config)
...
-				BlobLength:  1 << (blobPowerBytes - 5),
+				BlobLength:  blobFrs,
 				ChunkLength: max(1, rsExtendedBlobFrs/8192), // chosen such that numChunks*ChunkLength=rsExtendedBlobFrs
 			}
 			parametrizedProver, err := p.GetKzgProver(params, provingParams)
 			require.NoError(b, err)
 
 			for b.Loop() {
-				_, err = parametrizedProver.GetProofs(b.Context(), maxSizeBlobCoeffs[:rsExtendedBlobFrs])
+				_, err = parametrizedProver.GetProofs(b.Context(), maxSizeBlobCoeffs[:blobFrs])
```
samlaf, line 228: hmm why this change? We want to be generating proofs for the rs extended blob lengths. Otherwise I have no idea how the code above works. why is chunkLength being set to `rsExtendedBlobFrs/8192` for example.

Feel like you changed this place but its probably the proving params that we wanted to change instead?

encoding/v2/bench/benchmark_eigenda_test.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2268/commits/0337bf55960af45339e835e7dc363eb56e484b82

encoding/v2/bench/benchmark_eigenda_test.go
bxue-l2: getProof expect input of original data, not the erasure coded data, such that they can run in parallel
Screenshot 2025-11-03 at 5 12 17 PM

.github/workflows/benchmark-tests.yml
dmanc: It's pretty much just the dependencies of the benchmark files, but it seems pretty annoying to maintain if we move stuff around. Thinking about just running it on every commit, doesn't seem that bad.

encoding/v2/bench/Makefile
bxue-l2: we need 2^0 to 2^9. If our blob becomes 32MiB, then we need 2^10

tools/srs-utils/table_downloader/tables_downloader.go
```diff
@@ -11,7 +11,7 @@ import (
 const (
 	defaultTablesBaseURL   = "https://srs-mainnet.s3.amazonaws.com/kzg/SRSTables"
 	defaultTablesOutputDir = "resources/srs/SRSTables"
-	defaultDimension       = "dimE8192"
+	defaultDimension       = "dimE1024"
```
bxue-l2, line 14: why not other 2^0 to 2^9, those are useful for blob that is less than 32KiB, toeplitz merge PR

bxue-l2: https://github.com/Layr-Labs/eigenda/pull/2238

encoding/v2/bench/Makefile
dmanc: Yeah they generate pretty quickly now

encoding/v2/bench/benchmark_eigenda_test.go
samlaf: ah! My bad then... I guess it was always like this.. I hallucinated some stuff here. Oh probably because before it was calling GetFrames and forgot to change after I refactored to only call GetProofs?

Anyways sgtm. Wondering if theres a cleaner way to parametrize `chunkLength` though that doesn't require defining all those extendedBlobLen variables

PR #2269, opened 2025-11-04
refactor(encoding): Refactor icicle encoding to properly use streams

dmanc:

Previously we were creating a global stream per encoding operation (MultiFrameProofs / ExtendPolyEval) so if we increased the concurrency we could have multiple blobs operating under the same stream. The correct way to utilize streams is to have each blob operate on its own stream per operation.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
+++ b/encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
@@ -356,6 +356,12 @@ func (c *KzgMultiProofBackend) twoEcnttOnDevice(
 		return nil, time.Time{}, fmt.Errorf("forward ecntt failed: %v", err.AsString())
 	}
 
+	// Synchronize stream to ensure async ECNTT completes before converting results
+	syncErr := runtime.SynchronizeStream(stream)
+	if syncErr != runtime.Success {
+		return nil, time.Time{}, fmt.Errorf("stream synchronization failed: %v", syncErr.AsString())
+	}
+
 	proofs := icicle.HostSliceIcicleProjectiveToGnarkAffine(proofsBatchHost, int(c.NumWorker))
 
 	return proofs, firstECNTTDone, nil


encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
```diff
@@ -119,21 +118,36 @@ func (p *KzgMultiProofBackend) ComputeMultiFrameProofV2(ctx context.Context, pol
...
+			icicleErr = fmt.Errorf("failed to create stream: %v", streamErr.AsString())
+			return
+		}
+		defer func() {
+			// Synchronize stream to ensure all GPU operations complete before cleanup
+			syncErr := runtime.SynchronizeStream(stream)
+			if syncErr != runtime.Success && icicleErr == nil {
+				icicleErr = fmt.Errorf("stream synchronization failed: %v", syncErr.AsString())
+			}
+			runtime.DestroyStream(stream)
```
samlaf, line 134: is it safe to destroy if we couldn't synchronize it? what does destroyStream do?

encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
```diff
@@ -119,21 +118,36 @@ func (p *KzgMultiProofBackend) ComputeMultiFrameProofV2(ctx context.Context, pol
...
+		stream, streamErr := runtime.CreateStream()
+		if streamErr != runtime.Success {
+			icicleErr = fmt.Errorf("failed to create stream: %v", streamErr.AsString())
+			return
+		}
+		defer func() {
+			// Synchronize stream to ensure all GPU operations complete before cleanup
+			syncErr := runtime.SynchronizeStream(stream)
+			if syncErr != runtime.Success && icicleErr == nil {
+				icicleErr = fmt.Errorf("stream synchronization failed: %v", syncErr.AsString())
```
samlaf: do we really want to return an error if we couldnt synchronize the stream? Does it matter?

encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
```diff
@@ -146,10 +160,14 @@ func (p *KzgMultiProofBackend) ComputeMultiFrameProofV2(ctx context.Context, pol
...
 		msmDone = time.Now()
 
-		// Compute the first ecntt, and set new batch size for ntt
-		p.NttCfg.BatchSize = int32(1)
+		// Create NTT config for ECNTT operations
+		nttCfg := ntt.GetDefaultNttConfig()
+		nttCfg.IsAsync = true
+		nttCfg.Ordering = core.KNN
+		nttCfg.StreamHandle = stream
+		nttCfg.BatchSize = int32(1)
```
samlaf:
```suggestion
		nttCfg.StreamHandle = stream
```

samlaf: nit: dont need to set default values.

encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
```diff
@@ -146,10 +160,14 @@ func (p *KzgMultiProofBackend) ComputeMultiFrameProofV2(ctx context.Context, pol
...
 		// run two ecntt in one function, the first and second ecntt operates on the same device slice
-		proofs, firstECNttDone, err = p.twoEcnttOnDevice(sumVec, int(numChunks), int(toeplitzMatrixLen))
+		proofs, firstECNttDone, err = p.twoEcnttOnDevice(sumVec, int(numChunks), int(toeplitzMatrixLen), nttCfg)
```
samlaf: nit: pass the stream instead and create the nttCfg inside the fct

encoding/v2/rs/backend/icicle/extend_poly.go
```diff
@@ -76,11 +73,33 @@ func (g *RSBackend) ExtendPolyEvalV2(ctx context.Context, coeffs []fr.Element) (
...
+				icicleErr = fmt.Errorf("stream synchronization failed: %v", syncErr.AsString())
+			}
+			icicle_runtime.DestroyStream(stream)
+		}()
+
+		// Create NTT config for this operation
+		cfg := ntt.GetDefaultNttConfig()
+		cfg.IsAsync = true
+		cfg.Ordering = core.KNN
+		cfg.StreamHandle = stream
```
samlaf, line 95: don't think we need all of this complexity here, we literally are only doing NTT. No point in creating a stream for this afaiu

samlaf: there's nothing to run concurrently or even pipeline

samlaf (approved):

encoding/v2/rs/backend/icicle/extend_poly.go
dmanc: Wouldn't it run concurrently against the multiframeproof since they're on separate streams?

encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2269/commits/7d82771087a33cf42b38bafc9f78f2dff1201201

encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2269/commits/7d82771087a33cf42b38bafc9f78f2dff1201201

encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
dmanc: https://github.com/Layr-Labs/eigenda/pull/2269/commits/7d82771087a33cf42b38bafc9f78f2dff1201201 I downgraded it to a warning log

encoding/v2/kzg/prover/backend/icicle/multiframe_proof.go
dmanc: https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__STREAM.html#group__CUDART__STREAM_1gfda584f1788ca983cb21c5f4d2033a62

Seems like from the docs that it will wait anyway for all operations to complete before destroying.

encoding/v2/rs/backend/icicle/extend_poly.go
samlaf: So Jeremy answered that if you don't specify a stream it uses the default device stream. But then different extend_blobs would all be on the same default stream. So agree this is better.

samlaf (approved):

PR #2270, opened 2025-11-04
feat(contracts): Add recency window to cert verifier

iquidus:
Closes DAINT-825, DAINT-826

* Adds recency window to `IEigenDACertVerifier` interface
* Adds recency window to `EigenDACertVerifier` contract
* Bumps cert verifier semver to 3.2.0
* Updates deployment scripts
* Updates Go bindings for cert verifier contract

### DIFF
--- a/contracts/script/deploy/eigenda/testnet.hoodi.config.toml
+++ b/contracts/script/deploy/eigenda/testnet.hoodi.config.toml
@@ -75,7 +75,7 @@ batchConfirmers = ["0x3aaabc5361fa6fcd4ac4623253a709a2e476577b"]
 confirmationThreshold = 55
 adversaryThreshold = 33
 quorumNumbersRequired = [0, 1]
-recencyWindow = 0
+recencyWindow = 14400
 
 [initParams.eigenDA.disperser]
 dispersers = ["0x34200f3326bfa13df47fdbe39e51a0df5512aa22"]


contracts/script/deploy/certverifier/config/v2/holesky.testnet.config.json
```diff
@@ -4,5 +4,6 @@
         "0_confirmationThreshold": 55,
         "1_adversaryThreshold": 33
     },
-    "quorumNumbersRequired": "0x0001"
+    "quorumNumbersRequired": "0x0001",
+    "recencyWindow": 0
```
ethenotethan: why are we recommending a default of 0? shouldn't this be set proportional to some time delay which should be set a particular block production rate based on the chain env?

iquidus: With the current client logic `0` = no recency check. Yes, this value should be set (assuming a recency check is desired), however this value will change from env to env, and given changing it would be a consensus breaking change, thought should be put into setting an appropriate value. 0 seems like a sane default, but very open to suggestions.

iquidus: Updated in [a3f00e0](https://github.com/Layr-Labs/eigenda/pull/2270/commits/a3f00e0ec11d35257a69128d49f9d42f5d0b1060) 

14400 = 48 hours (with 12 second blocks).

contracts/script/deploy/eigenda/mainnet.beta.config.toml
samlaf, line 1: do we still need this file? mainnet beta has never actually materialized right?

contracts/script/deploy/eigenda/preprod.hoodi.config.toml
```diff
@@ -75,7 +75,7 @@ batchConfirmers = ["0xf7fd61910d1b5d705c25e4a55a67d577a650bf2e"]
 confirmationThreshold = 55
 adversaryThreshold = 33
 quorumNumbersRequired = [0, 1]
-
+recencyWindow = 14400
```
samlaf, line 78: Afaiu this is used when making a full initial deployment to a new chain right? including (maybe?) inabox?

IMO we should remove all certVerifier values from this config file, and the deploy script should read the certVerifier config values from deploy/certverifier/config/v2/xyz.testnet.config.json. Otherwise we have too many config values duplicated all over the place, as evidenced by this PR.

Anyway prob more of a point for @ethenotethan and greg, nothing to do as part of this PR.

contracts/script/deploy/eigenda/testnet.hoodi.config.toml
```diff
@@ -75,6 +75,7 @@ batchConfirmers = ["0x3aaabc5361fa6fcd4ac4623253a709a2e476577b"]
 confirmationThreshold = 55
 adversaryThreshold = 33
 quorumNumbersRequired = [0, 1]
+recencyWindow = 0
```
samlaf: should this be 14400?

contracts/script/deploy/eigenda/mainnet.beta.config.toml
iquidus: I assume no, but don't want to remove it in this PR.

contracts/script/deploy/eigenda/testnet.hoodi.config.toml
iquidus: Yes. Fixed in [3a9635e](https://github.com/Layr-Labs/eigenda/pull/2270/commits/3a9635e97a1cc52463f9b24bdb0e9057c1fa92ed)

PR #2271, opened 2025-11-04
feat: better signing metrics

cody-littley:
Improves metrics we capture for signing rate

### DIFF
--- a/core/aggregation_test.go
+++ b/core/aggregation_test.go
@@ -276,9 +276,9 @@ func TestFilterQuorums(t *testing.T) {
 		update)
 	assert.NoError(t, err)
 	assert.Len(t, aq.SignerMap, numOpr-advCount)
-	assert.Equal(t, aq.SignerMap, map[core.OperatorID]bool{
-		mock.MakeOperatorId(0): true,
-		mock.MakeOperatorId(1): true,
+	assert.Equal(t, aq.SignerMap, map[core.OperatorID]struct{}{
+		mock.MakeOperatorId(0): struct{}{},
+		mock.MakeOperatorId(1): struct{}{},
 	})
 	assert.Contains(t, aq.AggSignature, core.QuorumID(0))
 	assert.Contains(t, aq.AggSignature, core.QuorumID(1))


disperser/controller/signature_receiver.go
```diff
@@ -478,12 +502,27 @@ func getSignedPercentage(signedStake *big.Int, totalStake *big.Int) uint8 {
...
 }
 
+// getFraction returns a fraction (as a float64) representing part / whole.
+func getFraction(part *big.Int, whole *big.Int) float64 {
+	if whole.Cmp(big.NewInt(0)) == 0 {
+		// avoid dividing by 0
+		return 0.0
+	}
+
+	signedStakeFloat := new(big.Float).SetInt(part)
```
litt3: nit: the utility method parameters are generic (`part`, `whole`), but the local variables here are specific

disperser/controller/signature_receiver.go
```diff
@@ -177,7 +182,42 @@ forLoop:
 				break forLoop
 			}
 
-			sr.handleNextSignature(signingMessage, attestationChan)
+			isValid := sr.handleNextSignature(signingMessage, attestationChan)
+
+			for quorumId, opInfo := range sr.indexedOperatorState.Operators {
```
litt3: Claude identified this issue, which looks right to me:


> This loop iterates over sr.indexedOperatorState.Operators, which is `map[core.QuorumID]map[core.OperatorID]*core.OperatorInfo`. For each quorum, `opInfo` is the inner map containing all operators in THAT quorum.
> However, the operator who sent `signingMessage` may not be a member of every quorum. When you do `opInfo[signingMessage.Operator]`, if that operator is not in the current quorum being iterated, this will return `nil`. Then accessing `.Stake` on `nil` will panic.

The indexed operator state is so hard to reason about... I really don't like that structure

disperser/controller/signature_receiver.go
```diff
@@ -177,7 +182,42 @@ forLoop:
 				break forLoop
 			}
 
-			sr.handleNextSignature(signingMessage, attestationChan)
+			isValid := sr.handleNextSignature(signingMessage, attestationChan)
```
litt3: I don't think you can use `isValid` the way you're using it, since it's not nuanced enough:

- `handleNextSignature` returns `false` if the operator isn't in the state. But if the operator isn't in the state, should its stake really be reported as "unsigned"?
- `handleNextSignature` returns `false` if the message is a duplicate. So in that case, you would be counting the stake as both signed AND unsigned

disperser/controller/controller_metrics.go
```diff
@@ -502,46 +489,27 @@ func (m *dispatcherMetrics) newSendToValidatorProbe() *common.SequenceProbe {
...
+	label := prometheus.Labels{"id": id.Hex(), "quorum": fmt.Sprintf("%d", quorum)}
 
-		m.signedBatchCount.With(label).Add(1)
-		m.signedByteCount.With(label).Add(float64(batchSize))
-		m.signingLatency.With(label).Observe(signingLatency.Seconds())
+	if success {
+		m.validatorSignedBatchCount.With(label).Add(1)
+		m.validatorSignedByteCount.With(label).Add(float64(batchSize) * stakeFraction)
+	} else {
+		m.validatorUnsignedBatchCount.With(label).Add(1)
```
litt3, line 514: I'm a bit confused here.

You are updating `validatorUnsignedByteCount` within `ReportValidatorSigningResult`. But `ReportValidatorSigningResult` is only called when a signing message is _received_. So, it seems like `validatorUnsignedByteCount` will only be update when a validator DOES send a signing message, but verification fails for some reason.

I think calling this `validatorUnsignedByteCount` is misleading, since it won't include validators that send no message at all

disperser/controller/controller_metrics.go
```diff
@@ -264,142 +247,154 @@ func newDispatcherMetrics(
...
-	if m == nil {
-		return
-	}
-	m.sendChunksRetryCount.WithLabelValues().Set(retries)
+	globalSigningFractionHistogram := promauto.With(registry).NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: controllerNamespace,
+			Name:      "global_signing_fraction_histogram",
+			Help:      "Histogram of the fraction of validators that signed each batch",
+			Buckets:   prometheus.LinearBuckets(0.0, 0.05, 21),
```
litt3, line 295: Are linear buckets actually the best choice here? Would something like this make more sense?

`[]float64{0.0, 0.30, 0.55, 0.60, 0.65, 0.70, 0.75, 0.80, 0.85, 0.90, 0.95, 0.99, 1.0}`

It seems additional granularity around the top of the range could be helpful.

disperser/controller/controller_metrics.go
```diff
@@ -32,31 +32,43 @@ type dispatcherMetrics struct {
...
-// A count of batches meeting each specified threshold is reported as a metric.
-func newDispatcherMetrics(
+// Sets up metrics for the controller.
+func newControllerMetrics(
 	registry *prometheus.Registry,
-	importantSigningThresholds []float64,
-) (*dispatcherMetrics, error) {
+	// The minimum fraction of signers for a batch to be considered properly signed. Any fraction greater
+	// than or equal to this value is considered a successful signing.
+	minimumSigningThreshold float64,
```
litt3, line 58: Consider verifying that `minimumSigningThreshold` is between `0.0` and `1.0`

disperser/controller/dispatcher.go
```diff
@@ -544,7 +526,16 @@ func (d *Dispatcher) HandleSignatures(
 		return fmt.Errorf("update batch status: %w", err)
 	}
 
-	// Track attestation metrics
+	// Update global signing metrics.
+	for quorumID := range batchData.OperatorState.Operators {
+		signingFraction := float64(finalAttestation.QuorumResults[quorumID].PercentSigned) / 100.0
```
litt3: `QuorumResults[quorumID]` could return `nil`

disperser/controller/signature_receiver.go
litt3: I see that you pushed some changes here after I started my review. Your changes still don't seem to address the whole problem:

```
				validatorInfo, ok := opInfo[signingMessage.Operator]
				if !ok {
					// This triggers in some unit tests
					sr.logger.Error("could not find validator stake for operator in quorum",
						"quorumID", quorumId,
						"operatorID", signingMessage.Operator.Hex(),
						"batchHeaderHash", hex.EncodeToString(sr.batchHeaderHash[:]))
					continue
				}
```

You're logging this as an error, but won't this just happen whenever the signer isn't in all quorums?

disperser/controller/signature_receiver.go
```diff
@@ -177,7 +182,23 @@ forLoop:
...
+				stakeFraction := getFraction(validatorStake, totalStake)
+
+				sr.metrics.ReportValidatorSigningResult(
+					signingMessage.Operator,
+					stakeFraction,
+					sr.batchSizeBytes,
+					quorumId,
+					isValid,
+				)
+			}
```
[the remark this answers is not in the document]
cody-littley: fixed

disperser/controller/dispatcher.go
```diff
@@ -818,13 +809,26 @@ func (d *Dispatcher) NewBatch(
...
 	}
 
+	batchSizeBytes := uint64(0)
+	for _, blobKey := range keys {
+		blobMetadata, ok := metadataMap[blobKey]
+		if !ok {
+			d.logger.Warn("missing blob metadata for blob key when updating signing metrics",
+				"blobKey", blobKey.Hex(),
+				"batchHeaderHash", batchHeaderHash)
+			continue
```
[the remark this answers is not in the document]
cody-littley: The error log is just defensive programming, I'd prefer not to add new places where this can return an error when I'm just adding logs.

disperser/controller/signature_receiver.go
cody-littley: fixed

disperser/controller/signature_receiver.go
cody-littley: agree, this is now just a `continue`

disperser/controller/controller_metrics.go
cody-littley: We should discuss this. My grug instincts tingle when I think about trying to be fancy about buckets.

disperser/controller/controller_metrics.go
cody-littley: You make a good point. I've refactored the code structure to fix this.

disperser/controller/signature_receiver.go
cody-littley: You are correct, I've refactored to improve this.

disperser/controller/controller_metrics.go
cody-littley: added

disperser/controller/dispatcher.go
cody-littley: this has been fixed

disperser/controller/controller_metrics.go
litt3: I'm fine leaving this as linear buckets, just wanted to bring it up 👍

disperser/controller/signature_receiver.go
```diff
@@ -245,6 +252,8 @@ func (sr *signatureReceiver) handleNextSignature(
 		// Delay the next tick since we just submitted an attestation.
 		sr.ticker.Reset(sr.tickInterval)
 	}
+
+	return
```
litt3: redundant return statement

disperser/controller/signature_receiver.go
```diff
@@ -569,3 +595,98 @@ func (sr *signatureReceiver) debugEquivalenceError(
...
+// This method should be called when the controller is finished collecting signatures for a batch.
+func (sr *signatureReceiver) reportBatchMetrics() {
+	if sr.metrics == nil {
+		return
+	}
+
+	sr.reportThresholdSignedToDoneLatency()
+	sr.metrics.reportAttestationUpdateCount(float64(sr.attestationUpdateCount))
+
+	finalAttestation := sr.latestAttestation
```
litt3: `finalAttestation := sr.latestAttestation`

Is this actually necessary?

disperser/controller/signature_receiver.go
```diff
@@ -569,3 +595,98 @@ func (sr *signatureReceiver) debugEquivalenceError(
...
+	// Update per-validator metrics
+	for quorumID, validatorsInQuorum := range sr.indexedOperatorState.Operators {
+
+		quorumTotals, ok := sr.indexedOperatorState.Totals[quorumID]
+		if !ok {
+			sr.logger.Errorf("missing quorum totals for quorum %d in final attestation for batch %s",
+				quorumID, batchHeaderHashString)
+			continue
+		}
+		quorumStake := quorumTotals.Stake
```
litt3: Is local variable `quorumStake` necessary?

disperser/controller/signature_receiver.go
```diff
@@ -569,3 +595,98 @@ func (sr *signatureReceiver) debugEquivalenceError(
...
+
+		for operatorID := range validatorsInQuorum {
+			_, signed := sr.validSignerMap[operatorID]
+
+			quorumInfo, ok := sr.indexedOperatorState.Operators[quorumID]
```
litt3: I'm confused here... you're getting `sr.indexedOperatorState.Operators[quorumID]`, and assigning the value to `quorumInfo`

But isn't the value of `sr.indexedOperatorState.Operators[quorumID]` already stored in `validatorsInQuorum`?

disperser/controller/signature_receiver.go
```diff
@@ -569,3 +595,98 @@ func (sr *signatureReceiver) debugEquivalenceError(
...
+
+		quorumTotals, ok := sr.indexedOperatorState.Totals[quorumID]
+		if !ok {
+			sr.logger.Errorf("missing quorum totals for quorum %d in final attestation for batch %s",
+				quorumID, batchHeaderHashString)
+			continue
+		}
+		quorumStake := quorumTotals.Stake
+
+		for operatorID := range validatorsInQuorum {
```
litt3: outdated terminology "operator" (also elsewhere in your new code)

disperser/controller/signature_receiver.go
```diff
@@ -569,3 +595,98 @@ func (sr *signatureReceiver) debugEquivalenceError(
...
+
+			validatorInfo, ok := quorumInfo[operatorID]
+			if !ok {
+				sr.logger.Errorf(
+					"missing validator info for operator %s in quorum %d in final attestation for batch %s",
+					operatorID.Hex(), quorumID, batchHeaderHashString)
+				continue
+			}
+
+			validatorStake := validatorInfo.Stake
```
litt3: is this local variable necessary?

disperser/controller/signature_receiver.go
cody-littley: removed

disperser/controller/signature_receiver.go
cody-littley: removed variable

disperser/controller/signature_receiver.go
cody-littley: removed

disperser/controller/signature_receiver.go
cody-littley: Good point, simplified

disperser/controller/signature_receiver.go
cody-littley: removed

disperser/controller/signature_receiver.go
cody-littley: fixed the places I could find

disperser/controller/signature_receiver.go
```diff
@@ -569,3 +593,87 @@ func (sr *signatureReceiver) debugEquivalenceError(
...
+
+		quorumTotals, ok := sr.indexedOperatorState.Totals[quorumID]
+		if !ok {
+			sr.logger.Errorf("missing quorum totals for quorum %d in final attestation for batch %s",
+				quorumID, batchHeaderHashString)
+			continue
+		}
+
+		for validatorID := range validatorsInQuorum {
+			_, signed := sr.validSignerMap[validatorID]
```
litt3, line 642: When I originally implemented this, I used a map of bools. In hindsight, it should have been implemented as a set `map[core.OperatorID]struct{}`

With that in mind, the way you're doing `_, signed := sr.validSignerMap[validatorID]` is a bit counterintuitive. The ignored first return value is the `bool`, which intuitively you'd think _could_ be either `true` or `false` (even though in practice that isn't the case)

Recommend changing the type of `validSignerMap` to be an idiomatic set

cody-littley: changed

litt3 (approved):

litt3 (approved):

PR #2276, opened 2025-11-05
docs: add write and read workflow documentation

Lulu-Zhou-EigenLabs:
Add documentation about the workflow of writing and reading in EigenDA.


docs/spec/src/protocol/architecture/write-and-read-workflow.md
```diff
@@ -0,0 +1,57 @@
...
+
+
+### Write
+
+When a user writes data to EigenDA (in the form of a blob), the blob is encoded into chunks and distributed to the validators in accordance with the [Chunk Assignment Logic](./assignment.md). After enough validators have acknowledged receipt of their chunks and returned their signatures to the disperser, the disperser aggregates the signatures into a data availability (DA) certificate and sends it to the user upon request.
+
+The write process follows the sequence below. The labels in parentheses (e.g., W1, W2) correspond to the steps shown in the diagram above.
+
+1. **Disperser Receives Blob (W1, W2, W3).**
+   The disperser receives a blob consisting of a `BlobHeader` and `BlobData`. As a precaution, the disperser can validate the `PaymentMetadata` contained in the `BlobHeader` to ensure that the blob is properly funded, and that the KZG commitments in the `BlobHeader` are correct.
```
cody-littley: Note that the disperser verifying the payment data doesn't necessarily mean the payment data is valid, and the dispersal may fail at a later stage due to payments. This becomes possible with distributed dispersal, since no one disperser has a global view of payment state. A disperser does validation to weed out obvious fraud, and it's up to the validators to prevent bad behavior the disperser misses.

Not sure how to modify this part of the document. Perhaps you could add something like this?

```
(Note that validators may still reject payment data as invalid even if approved by a disperser, since a disperser lacks knowledge of global payment state.)
```

docs/spec/src/protocol/architecture/write-and-read-workflow.md
```diff
@@ -0,0 +1,57 @@
...
+
+2. **Disperser Encodes Blob (W6, W7).**
+   The disperser references the Chunk Assignment Logic to translate the `BlobHeader` into a set of `EncodingParams`. The disperser then encodes the blob according to the [Encoding Module](./encoding.md) and the `EncodingParams` to produce a collection of encoded `Chunk`s.
+
+3. **Disperser Serves Chunks.**
+   The disperser makes the encoded chunks available via the relay's `GetChunks` interface. This is an authenticated and rate-limited interface where each validator can only request its allocated amount of data.
```
cody-littley, line 25: > can only request its allocated amount of data.

Technically not true today, but it will eventually be true. IMO, ok to leave sentence it the way it is.

docs/spec/src/protocol/architecture/write-and-read-workflow.md
```diff
@@ -0,0 +1,57 @@
...
+
+### Read
+
+To read a blob, a client follows the sequence below. The labels in parentheses (e.g., R1, R2) correspond to the steps shown in the diagram above.
+
+1. **Read from Relay (R1).**
+   The client attempts to retrieve the blob from the `GetBlob` interface of the relay(s) identified in the `BlobHeader`. This is the primary and most efficient retrieval method, as the relay stores complete blobs.
+
+2. **Read from Validators (R2).**
+   If the blob is not available from the relay(s), the client falls back to retrieving individual chunks directly from the validators and reconstructing the blob. The client iterates through the `QuorumNumbers` in the `BlobHeader`. For each quorum (until successful recovery), the client requests chunks from each validator registered with that quorum, validates the integrity of the chunks, and then reconstructs the blob using the erasure coding scheme.
```
cody-littley: We don't do a `for each quorum` when reading from the validators. Instead, the client reconstructs the blob assignments for all quorums, and downloads from validators in a random order until it has enough chunks.

cody-littley (approved):

PR #2279, opened 2025-11-05
feat(controller): Drop stale blobs in encoding manager

litt3:
Part of EGDA-2273 ([link](https://linear.app/eigenlabs/issue/EGDA-2273/drop-old-dispersals-throughout-the-system))


disperser/cmd/controller/config.go
```diff
@@ -152,6 +152,10 @@ func NewConfig(ctx *cli.Context) (Config, error) {
 			MaxNumBlobsPerIteration:     int32(ctx.GlobalInt(flags.MaxNumBlobsPerIterationFlag.Name)),
 			OnchainStateRefreshInterval: ctx.GlobalDuration(flags.OnchainStateRefreshIntervalFlag.Name),
 			NumConcurrentRequests:       ctx.GlobalInt(flags.NumConcurrentEncodingRequestsFlag.Name),
+			// TODO(litt3): once the checkpointed onchain config registry is ready, that should be used
+			// instead of hardcoding. At that point, this field should be removed from here
+			// entirely, and the value will be fetched dynamically at runtime.
+			MaxDispersalAge: 45 * time.Second,
```
dmanc, line 158: How did we determine 45s?

disperser/controller/encoding_manager.go
```diff
@@ -220,19 +238,72 @@ func (e *EncodingManager) Start(ctx context.Context) error {
...
+	e.logger.Warnf(
+		"discarding stale dispersal: blobKey=%s dispersalAge=%s maxAge=%s dispersalTime=%s",
+		blobKey.Hex(),
+		dispersalAge.String(),
+		e.MaxDispersalAge.String(),
+		dispersalTime.Format(time.RFC3339),
+	)
+
+	storeCtx, cancel := context.WithTimeout(ctx, e.StoreTimeout)
+	err := e.blobMetadataStore.UpdateBlobStatus(storeCtx, blobKey, v2.Failed)
```
dmanc, line 299: If this fails do we eventually retry?

dmanc (approved):

disperser/cmd/controller/config.go
cody-littley: I'm guessing its chosen to loosely correspond to our 50s dispersal timeout? Although probably important to point out that the dispersal timeout starts when we start sending data to validators, and the timeout above is limiting the time we spend before sending data to dispersers.

disperser/controller/encoding_manager.go
```diff
@@ -62,6 +62,17 @@ type EncodingManagerConfig struct {
 	// NumConcurrentRequests is the size of the worker pool for processing encoding requests concurrently.
 	// Must be at least 1.
 	NumConcurrentRequests int
+
+	// MaxDispersalAge is the maximum age a dispersal request can be before it is discarded.
+	// Dispersals older than this duration are marked as Failed and not processed.
+	//
+	// Age is determined by the BlobHeader.PaymentMetadata.Timestamp field, which is set by the
+	// client at dispersal request creation time (in nanoseconds since Unix epoch).
```
cody-littley, line 69: Do we reject timestamps too far in the future? IMO, ideally this timeout would be one assigned by the disperser's clock, since it can't be gamed. But it's probably ok to use the timestamp in the payment data as a temporary thing (since I assume piping a new timestamp through our system is nontrivial).

Perhaps worthwhile adding a `TODO(litt3)` discussing this topic (assuming you agree with the ideal way this might work).

disperser/controller/encoding_manager.go
cody-littley: How do we handle dynamo write failures elsewhere?

cody-littley (approved):

disperser/cmd/controller/config.go
litt3: It has to do with what we determined our "max expected latency" to be. In my [current proposal](https://linear.app/eigenlabs/issue/EGDA-1801/determine-what-leaky-bucket-sizes-to-use-for-reservations), that's `1 minute`. So I selected 45 seconds, to be some amount < the max expected latency

disperser/controller/encoding_manager.go
litt3: I've looked through all of the status logic, and AFAICT there are _never_ any retries if this update fails. I agree this is a problem, but it's not one that is unique to this status update.

I've discussed this with @cody-littley in the past- but I think the correct solution would be to create a `BlobStatusStateMachine` component on the `Controller`, which is is solely responsible for updating blob status. The state machine would ingest events that impact blob status, and expose methods to query current blob statuses.

With a central point of control, it would be a lot easier to understand the blob status flow, and implement stuff like retry logic when writes fail

disperser/controller/encoding_manager.go
litt3: > Do we reject timestamps too far in the future?

Not in the current code. Do you think I should implement that in this PR? It would only be a minor change

> ideally this timeout would be one assigned by the disperser's clock, since it can't be gamed.

If we were to additionally reject timestamps that are too far in the future, I think options for gaming the system are pretty minimal. You can lie about your dispersal time, but only up to a given point in either direction. And it's not clear to me what you'd have to gain from doing so

disperser/controller/encoding_manager.go
cody-littley: In any case, this PR is probably not the place to address such issues.

disperser/controller/encoding_manager.go
cody-littley: The attack is that you could cause the disperser to enqueue work for a while and not discard it. I guess not too horrific of an attack, since you'd also need to respect your reservation. I have no problem merging this as-is, just something to think about for the future.

cody-littley (approved):

disperser/controller/encoding_manager.go
litt3: To capture our offline discussion:

- We'll go ahead and merge the current PR as-is.
- As a fast followup, I'll put up a PR for the API server, which checks that blob dispersal time is within a valid range, in both directions (past and future)
- The new API server check will protect us against time traveler events, and also provide better feedback to users who have clock drift

PR #2280, opened 2025-11-05
test(rust): add first integration test

samlaf:
This was a first exploration/playing around with the code that eiger wrote. Pretty happy with the result. There's a few API  kinks that I noticed and would like to iron out. Left some TODO in the integration test.

Currently only supports Sepolia. Plan to add support for inabox eventually so that we can run this in CI.

### DIFF
--- a/rust/crates/eigenda-tests/tests/common/proxy.rs
+++ b/rust/crates/eigenda-tests/tests/common/proxy.rs
@@ -61,7 +61,9 @@ impl EigenDaProxy {
                 cmd_args.push(
                     "--eigenda.v2.cert-verifier-router-or-immutable-verifier-addr".to_string(),
                 );
-                cmd_args.push("0x58D2B844a894f00b7E6F9F492b9F43aD54Cd4429".to_string());
+                // Latest CertVerifier on the Router: https://sepolia.etherscan.io/address/0x17ec4112c4BbD540E2c1fE0A49D264a280176F0D#readProxyContract
+                // TODO(samlaf): make this lib support router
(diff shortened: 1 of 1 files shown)


rust/crates/eigenda-tests/tests/common/proxy.rs
```diff
@@ -0,0 +1,94 @@
...
+        ];
+
+        match mode {
+            ProxyNetwork::Sepolia => {
+                cmd_args.push("--eigenda.v2.network".to_string());
+                cmd_args.push("sepolia_testnet".to_string());
+                cmd_args.push(
+                    "--eigenda.v2.cert-verifier-router-or-immutable-verifier-addr".to_string(),
+                );
+                cmd_args.push("0x58D2B844a894f00b7E6F9F492b9F43aD54Cd4429".to_string());
```
bxue-l2: I think the latest router address is 0x17ec4112c4BbD540E2c1fE0A49D264a280176F0D

bxue-l2 (dismissed):
looks nice, but one thing I noticed. It is still not checking the zero padding in the encoded payload

rust/crates/eigenda-tests/tests/common/proxy.rs
samlaf: this rust lib doesn't support router yet. updated to latest certVerifier address though: [1135e31](https://github.com/Layr-Labs/eigenda/pull/2280/commits/1135e31d7a2a6c489122c3877287a4ef74579d08)

bxue-l2 (approved):

PR #2282, opened 2025-11-05
fix(errors): Fix incorrect error comparisons

dmanc:

During the load test we observed a lot of 
```
  "msg": "failed to put blob certificate",
  "component": "EncodingManager",
  "err": "record already exists"
```

I think this could be happening if:
1. `encodeBlob` succeeds
2. `PutBlobCertificate` succeeds
3. `UpdateBlobStatus` fails, retries the blob
4. `PutBlobCertificate` fails and returns `blobstore.ErrAlreadyExists`
3. errors.Is(err, dispcommon.ErrAlreadyExists) returns false
5. Retry loop 

This PR just updates errors.Is(err, dispcommon.ErrAlreadyExists) to errors.Is(err, blobstore.ErrAlreadyExists). Not 100% sure this is the fix though. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/apiserver/get_blob_status_v2.go
+++ b/disperser/apiserver/get_blob_status_v2.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"errors"
 	"fmt"
-	"strings"
 	"time"
 
 	"github.com/Layr-Labs/eigenda/api"
@@ -49,7 +48,7 @@ func (s *DispersalServerV2) getBlobStatus(
 
 	metadata, err := s.blobMetadataStore.GetBlobMetadata(ctx, blobKey)
 	if err != nil {
-		if strings.Contains(err.Error(), "metadata not found") {
+		if errors.Is(err, blobstore.ErrMetadataNotFound) {
 			s.logger.Info("blob metadata not found", "err", err, "blobKey", blobKey.Hex())
 			return nil, status.New(codes.NotFound, "no such blob found")
 		}


disperser/controller/encoding_manager.go
```diff
@@ -321,7 +320,7 @@ func (e *EncodingManager) HandleBatch(ctx context.Context) error {
 				storeCtx, cancel := context.WithTimeout(ctx, e.StoreTimeout)
 				err = e.blobMetadataStore.PutBlobCertificate(storeCtx, cert, fragmentInfo)
 				cancel()
-				if err != nil && !errors.Is(err, dispcommon.ErrAlreadyExists) {
+				if err != nil && !errors.Is(err, blobstore.ErrAlreadyExists) {
```
samlaf, line 323: whats the fix here? Why do we have 2 errors that look exactly the same? And why does changing from one to the other matter?

samlaf: image

Oh I see you're making sure this error condition hits? If this was a real mistake then it should be a listen to be cleaner with our error handling and to stop duplicating Errors all over the place. This PR should instead cleanup this mess and delete one of the 2 duplicate ErrAlreadyExists conditions imo.

samlaf: also why are we only seeing this error now though? git blame shows this was wrong for at least 5 months now it looks like

dmanc: My hypothesis is that the scenario I described in the description is unlikely to occur unless we're at high throughput. Here is another weird one https://github.com/Layr-Labs/eigenda/pull/2283

dmanc: I fixed some of the other error comparisons https://github.com/Layr-Labs/eigenda/pull/2282/commits/c149a3deb73ecece19c7cde97dc32205af02aff9.

cody-littley: This sort of fiddly error handling is exactly why I'm not a fan of passing anything other than debug messages as golang errors. If a function returns data that is actionable, then that data aught to be explicitly returned.

(Not suggesting we make that change here, just ranting. 😜 )

cody-littley (approved):

disperser/apiserver/get_blob_status_v2.go
```diff
@@ -67,7 +66,7 @@ func (s *DispersalServerV2) getBlobStatus(
 
 	cert, _, err := s.blobMetadataStore.GetBlobCertificate(ctx, blobKey)
 	if err != nil {
-		if errors.Is(err, dispcommon.ErrMetadataNotFound) {
+		if errors.Is(err, blobstore.ErrMetadataNotFound) {
```
litt3, line 68: Would it make sense to also update line 52 to use the correct error as well?

Currently it's using string matching. Yikes!

dmanc: yup https://github.com/Layr-Labs/eigenda/pull/2282/commits/824c68e8813381f41cf8a1d91e1ac25b105ec711

litt3 (approved):

PR #2284, opened 2025-11-05
fix(inabox): Fix inabox tests 

dmanc:

Inabox tests have been failing recently, not sure which PR introduced it or why it wasn't caught earlier.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests


inabox/deploy/config.go
```diff
@@ -251,6 +251,7 @@ func (env *Config) generateBatcherVars(ind int, key, graphUrl, logPath string) B
 		BATCHER_LOG_FORMAT:                    "text",
 		BATCHER_S3_BUCKET_NAME:                "test-eigenda-blobstore",
 		BATCHER_DYNAMODB_TABLE_NAME:           "test-BlobMetadata",
+		BATCHER_OBJECT_STORAGE_BACKEND:        "s3",
```
pschork, line 254: Shouldn't be required, but support this being here for sanity.

pschork:
Breakage seems to have been caused by [this](https://github.com/Layr-Labs/eigenda/pull/2261/files#diff-5421e47771fda80b444c5bd6a84098fb5a87cd398262dfcba5dd235a3d9e63bbR86) change to batcher where it was updated to use the object storage client factory which expects a blobStoreConfig with Backend field defined to determine which client to create. Previous batcher logic was creating a native AWS S3 client directly.

PR #2289, opened 2025-11-07
fix(batcher): add nil G1Point check to prevent panic in signature aggregation

pschork:
Panic observed on testnet-sepolia where batcher attempted to call Clone() on a nil G1Point (PubkeyG1)

This change makes the assertion that is it better to drop a non-signer from the batch confirmation rather than panic, drop the batch confirmation and trigger a short liveness outage as batcher restarts.

Flow:
 1. An operator had nil PubkeyG1 in IndexedOperatorState (Root cause unknown)
 3. Operator didn't sign the batch
 4. Code attempted: signersAggKey.Sub(nsk) where nsk was nil
 5. Sub() method internally called nsk.Clone() on nil pointer
 6. Clone() method dereferenced nil pointer → SIGSEGV

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0xc1f57f]

goroutine X [running]:
github.com/Layr-Labs/eigenda/core.(*G1Point).Clone(...)
      /workspace/core/attestation.go:70
github.com/Layr-Labs/eigenda/core.(*StdSignatureAggregator).ReceiveSignatures(...)
      /workspace/core/aggregation.go:305
github.com/Layr-Labs/eigenda/disperser/batcher.(*Batcher).HandleSingleBatch(...)
      /workspace/disperser/batcher/batcher.go:550
```

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/core/aggregation_test.go
+++ b/core/aggregation_test.go
@@ -285,10 +285,10 @@ func TestNilPubkeyG1Handling(t *testing.T) {
 	}()
 
 	// This should not panic even with nil PubkeyG1 in non-signers
-	// Before the fix, this would panic when trying to process non-signers with nil PubkeyG1
+	attestationCtx := ctx
 	aq, _ := agg.ReceiveSignatures(
 		ctx,
-		ctx,
+		attestationCtx,
 		state.IndexedOperatorState,
 		message,
 		update)


core/aggregation.go
```diff
@@ -375,7 +378,10 @@ func (a *StdSignatureAggregator) AggregateSignatures(
 	for id, op := range indexedOperatorState.IndexedOperators {
 		_, found := quorumAttestation.SignerMap[id]
 		if !found {
-			nonSignerKeys = append(nonSignerKeys, op.PubkeyG1)
+			// Only add non-signers with valid G1 public keys to prevent nil pointer dereference
+			if op.PubkeyG1 != nil {
+				nonSignerKeys = append(nonSignerKeys, op.PubkeyG1)
```
litt3, line 383: I'm trying to decide if it makes sense to just drop the non-signer in this case. The question is whether the non-signer's pubkey is part of the quorum aggregate pubkey... if so, then _not_ adding it to the non-signers will prevent successful aggregation anyway, since we won't be able to derive the pubkey of signers via subtraction.

I guess in any case, dropping the non-signer is better than panicking. And by dropping the non-signer instead of returning an error, there's at least the _possibility_ of succeeding, if the non-signer isn't included in the quorum aggregate pubkey

Alright, I'm on board with this solution. Just wanted to capture this train of thought

PR #2290, opened 2025-11-07
feat(validator): Implements dispersal request auth for multiple registered disperser keys

pschork:
Controller
- Add required `CONTROLLER_DISPERSER_ID` flag
- Pass disperser ID through dispatcher → node client manager → node client chain
- Remove hardcoded `EigenLabsDisperserID` constant usage

Validator
- Remove key cache preload
- ~~Updates key cache preloading to load all registered disperser keys from registry up to the specified limit `DispersalAuthenticationKeyCacheSize` or stops when resolving a default address 0x0~~
- ~~Preload will **return** error if the key at ID 0 resolves to the default address 0x0~~
- ~~Preload will **return** error if no keys were loaded~~
- ~~Preload will **log** error and continue if `chainReader.GetDisperserAddress` returns an error. This is meant to allow for partial preload of the key cache in the event that one of many keys is unable to be resolved.~~
- Authenticates requests using disperser ID from request header
- Maintains default of disperser ID 0 for on-demand payment authorization

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -1,10 +1,2 @@
-# Clients
-# See https://github.com/orgs/Layr-Labs/teams/eigenda-intg
-/api/clients @Layr-Labs/eigenda-intg
-/api/proxy @Layr-Labs/eigenda-intg
-
-# Contracts
-/contracts @pakim249CAL
-
 # Security docs
 /docs/audits @anupsv
--- a/Dockerfile
+++ b/Dockerfile
@@ -82,6 +82,13 @@ RUN --mount=type=cache,target=/go/pkg/mod \
     --mount=type=cache,target=/root/.cache/go-build \
     go build -o ./bin/controller ./cmd/controller
 
+# Ejector build stage
+FROM common-builder AS ejector-builder
+WORKDIR /app/ejector
+RUN --mount=type=cache,target=/go/pkg/mod \
+    --mount=type=cache,target=/root/.cache/go-build \
+    go build -o ./bin/ejector ./main
+
 # Relay build stage
 FROM common-builder AS relay-builder
 WORKDIR /app/relay
@@ -165,6 +172,10 @@ ENTRYPOINT ["nodeplugin"]
 FROM alpine:3.22 AS controller
 COPY --from=controller-builder /app/disperser/bin/controller /usr/local/bin
 ENTRYPOINT ["controller"]
+
+FROM alpine:3.22 AS ejector
+COPY --from=ejector-builder /app/ejector/bin/ejector /usr/local/bin
+ENTRYPOINT ["ejector"]
 
 FROM alpine:3.22 AS relay
 COPY --from=relay-builder /app/relay/bin/relay /usr/local/bin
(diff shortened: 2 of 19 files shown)


disperser/cmd/controller/flags/flags.go
```diff
@@ -322,6 +322,12 @@ var (
 		Value:    30 * time.Second,
 		EnvVar:   common.PrefixEnvVar(envVarPrefix, "PAYMENT_VAULT_UPDATE_INTERVAL"),
 	}
+	DisperserIDFlag = cli.Uint64Flag{
```
litt3, line 336: Does this actually need to be a CLI flag?

Since the disperser will have its own pubkey, it seems like it should be able to just fetch its `DisperserID` from the `DisperserRegistry`. That way, we avoid adding an additional config field that can get out of sync

api/clients/v2/node_client.go
```diff
@@ -76,7 +76,7 @@ func (c *nodeClient) StoreChunks(ctx context.Context, batch *corev2.Batch) (*cor
 			},
 			BlobCertificates: blobCerts,
 		},
-		DisperserID: api.EigenLabsDisperserID, // this will need to be updated when dispersers are decentralized
+		DisperserID: c.config.DisperserID,
```
litt3, line 79: Do we even need to include the `DisperserID` in this request?

Validators could just get the disperser pubkey from the signature included on the request, and then use that pubkey to look the `DisperserID` up on-chain, right?

disperser/cmd/controller/flags/flags.go
pschork: `DisperserRegistry` does not currently provide an `Address->ID` method.

api/clients/v2/node_client.go
pschork: > Do we even need to include the DisperserID in this request?

Technically we don't. But removing it will add 20–30% computational overhead on the validator if it has to `ecrecover` every dispersal request as opposed the just verifying the sig against a known pubkey.

> Validators could just get the disperser pubkey from the signature included on the request, and then use that pubkey to look the DisperserID up on-chain, right?

As per previous comment, there is no `Address->DisperserID` method on disperser registry currently.

disperser/cmd/controller/flags/flags.go
litt3: I forgot we are missing that- it's actually part of my [draft](https://linear.app/eigenlabs/issue/EGDA-2236/modify-disperserregistry-to-support-permissionless-dispersal) of changes we should make for permissionless dispersal.

I agree the strategy you used makes sense in the short term- we can just remove this config requirement once the contract has been updated with the necessary functions

api/clients/v2/node_client.go
litt3: My guess would be that this is a trivial cost increase. 20-30%, sure, but of a very cheap operation.

In any case, this isn't something we need to consider on this PR, since the `DisperserID` field already exists in the message. And, as you pointed out above, we don't actually have the necessary contract functions to support the alternate flow anyway

litt3 (approved):

PR #2293, opened 2025-11-07
chore: refactor metric to use bytes instead of symbols

cody-littley:

Refactor payments histogram to use bytes for bucket sizes instead of symbols

### DIFF
--- a/core/payments/reservation/reservationvalidation/reservation_validator_metrics.go
+++ b/core/payments/reservation/reservationvalidation/reservation_validator_metrics.go
@@ -9,6 +9,9 @@ import (
 
 // Tracks metrics for the [ReservationPaymentValidator]
 type ReservationValidatorMetrics struct {
+	// Although payments internally tracks things in symbols, the consumer of metrics wants to see things in bytes.
+	// For a histogram, it's actually not possible to automatically rename bucket labels in grafana, so using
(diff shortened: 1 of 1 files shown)


core/payments/reservation/reservationvalidation/reservation_validator_metrics.go
```diff
@@ -1,13 +1,15 @@
...
+	"github.com/Layr-Labs/eigenda/encoding"
+	"github.com/docker/go-units"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promauto"
 )
 
 // Tracks metrics for the [ReservationPaymentValidator]
 type ReservationValidatorMetrics struct {
-	reservationSymbols               prometheus.Histogram
+	reservationBytes                 prometheus.Histogram
```
litt3, line 15: maybe leave a comment about why this is bytes, but everything else payment related is symbols

cody-littley:
```
// Although payments internally tracks things in symbols, the consumer of metrics wants to see things in bytes.
	// For a histogram, it's actually not possible to automatically rename bucket labels in grafana, so using
	// symbols here causes dashboards to be less intuitive.
	reservationBytes                 prometheus.Histogram
```

PR #2303, opened 2025-11-10
feat(payments): Update reservation bucket sizes with verdict

litt3:
- Closes EGDA-1801 ([link](https://linear.app/eigenlabs/issue/EGDA-1801/determine-what-leaky-bucket-sizes-to-use-for-reservations))
- The verdict was to make client-side bucket 1 minute in duration
- To accommodate 1 minute e2e latency, that yields a validator side bucket of 2 minutes 
- Disperser bucket is in the middle, at 1.5 minutes


cody-littley (approved):
I'm really hoping we never have to change these values. Going to be a huge pain since we don't have the config registry up and ready to go.

PR #2307, opened 2025-11-12
feat(proxy): Arb CustomDA Failover Signal

ethenotethan:
Converts our failover signal returned on dispersal attempts into an a special type that's understand by the nitro batch poster. Changes based on reference commit: https://github.com/OffchainLabs/nitro/commit/d298d2b62e033e3195f33740c55e5396ff76a478 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/common/s3/aws/aws_s3_client.go
+++ b/common/s3/aws/aws_s3_client.go
@@ -143,6 +143,50 @@ func (s *awsS3Client) DownloadObject(ctx context.Context, bucket string, key str
 		return nil, s3common.ErrObjectNotFound
 	}
 
+	if len(buffer.Bytes()) != objectSize {
+		return nil, fmt.Errorf("downloaded object size (%d) does not match expected size (%d)",
+			len(buffer.Bytes()), objectSize)
+	}
+
+	return buffer.Bytes(), nil
+}
+
+func (s *awsS3Client) DownloadPartialObject(
+	ctx context.Context,
+	bucket string,
+	key string,
+	startIndex int64,
+	endIndex int64) ([]byte, error) {
+
+	if startIndex < 0 || endIndex <= startIndex {
+		return nil, fmt.Errorf("invalid startIndex (%d) or endIndex (%d)", startIndex, endIndex)
+	}
+
+	rangeHeader := fmt.Sprintf("bytes=%d-%d", startIndex, endIndex-1)
+
+	buffer := manager.NewWriteAtBuffer(make([]byte, 0, endIndex-startIndex))
+
+	var partMiBs int64 = 10
+	downloader := manager.NewDownloader(s.s3Client, func(d *manager.Downloader) {
+		// 10MB per part
+		d.PartSize = partMiBs * 1024 * 1024
+		// The number of goroutines to spin up in parallel per call to download when sending parts
+		d.Concurrency = 3
+	})
+
+	_, err := downloader.Download(ctx, buffer, &s3.GetObjectInput{
+		Bucket: aws.String(bucket),
+		Key:    aws.String(key),
+		Range:  aws.String(rangeHeader),
+	})
+	if err != nil {
+		return nil, fmt.Errorf("failed to download partial object: %w", err)
+	}
+
+	if buffer == nil || len(buffer.Bytes()) == 0 {
+		return nil, s3common.ErrObjectNotFound
+	}
+
 	return buffer.Bytes(), nil
 }
 
--- a/common/s3/mock_s3_client.go
+++ b/common/s3/mock_s3_client.go
@@ -25,6 +25,7 @@ func NewMockS3Client() *MockS3Client {
 			"CreateBucket":             0,
 			"FragmentedUploadObject":   0,
 			"FragmentedDownloadObject": 0,
+			"DownloadPartialObject":    0,
 		},
 	}
 }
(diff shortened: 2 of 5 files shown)


api/proxy/servers/arbitrum_altda/handlers.go
```diff
@@ -221,10 +222,8 @@ func (h *Handlers) RecoverPayload(
 //	@return bytes: Arbitrum Custom DA commitment bytes
 //	@return error: a structured error message (if applicable)
 //
-// TODO: Determine the encoding standard to use for the returned DA Commitment. It's assumed that an EigenDAV2 message
-// header byte will be prefixed. We can likely reuse the Standard Commitment mode but will require some analysis.
-//
-// TODO: Add processing for client provided timeout value
+// TODO: Add processing for client provided timeout value.
+// do we actually need this?
```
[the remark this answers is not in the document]
litt3: Agreed- I'm not sure what "this" is referring to here

api/proxy/servers/arbitrum_altda/handlers.go
```diff
@@ -244,6 +243,12 @@ func (h *Handlers) Store(
 
 	certBytes, err := h.eigenDAManager.Put(ctx, message, coretypes.CertSerializationABI)
 	if err != nil {
+		// translate a "failover" error into the FallbackRequested type error
+		// that arbitrum nitro understands to be the same
+		if errors.Is(err, &api.ErrorFailover{}) {
+			return nil, ErrFallbackRequested
```
litt3: You're losing the `err` context that's normally included in `ErrorFailover`. Is it not possible to include the same context in `ErrFallbackRequested`?

api/proxy/servers/arbitrum_altda/types.go
```diff
@@ -1,11 +1,28 @@
...
 
+var (
+	// Vendored from:
+	// https://github.com/OffchainLabs/nitro/blob/d298d2b62e033e3195f33740c55e5396ff76a478/daprovider/writer.go
+	//
+	// ErrFallbackRequested is returned by a DA provider to explicitly signal that
+	// the batch poster should fall back to the next available DA writer.
+	// Without this explicit signal, any error will cause batch posting to fail
+	// rather than automatically falling back, preventing expensive surprise costs
+	// from fixable infrastructure issues. Although the rpcclient will retry a certain
```
litt3:
```
Without this explicit signal, any error will cause batch posting to fail
	// rather than automatically falling back, preventing expensive surprise costs
	// from fixable infrastructure issues.
```

This is phrased in a confusing way. The sentence starts by saying what happens if you _don't_ send this explicit signal, but then goes on to imply that _not_ sending the signal will "prevent surprise costs". I assume that's the opposite of what's actually intended

api/proxy/servers/arbitrum_altda/types.go
```diff
@@ -1,11 +1,28 @@
...
+	// number of times on transient errors, there could be other issues like
+	// misconfigurations or temporary outages that are better fixed by operator
+	// intervention than automatically falling back.
```
litt3:
```
there could be other issues like
	// misconfigurations or temporary outages that are better fixed by operator
	// intervention than automatically falling back.
```

I guess I'm just not grasping what's being said here. This sentence seems to imply that using this error type will _not_ cause automatic fallback, and instead require manual intervention.

ethenotethan: hmm fwiw this is copy pasta'd from arbitrum's docstring

api/proxy/servers/arbitrum_altda/handlers.go
ethenotethan: actually your correct - we could a wrapping here to preserve the underlying the failure context for visibility in the batch poster since it uses `errors.Is`:
```
if errors.Is(err, daprovider.ErrFallbackRequested) {
	log.Warn("DA writer explicitly requested fallback", "writerIndex", i, "error", err, "duration", storeDuration)
	fallbackRequested = true
	continue // Try next writer
}
```

api/proxy/servers/arbitrum_altda/types.go
ethenotethan: also copy pasta from OCL's docstring on the error definition - I can refactor to use language more friendly to us

api/proxy/servers/arbitrum_altda/types.go
ethenotethan: [628847f](https://github.com/Layr-Labs/eigenda/pull/2307/commits/628847f2dabe3099e4b0297eb0b758ba60a3bafd)

api/proxy/servers/arbitrum_altda/types.go
ethenotethan: [628847f](https://github.com/Layr-Labs/eigenda/pull/2307/commits/628847f2dabe3099e4b0297eb0b758ba60a3bafd)

api/proxy/servers/arbitrum_altda/handlers.go
ethenotethan: [fd8bf8c](https://github.com/Layr-Labs/eigenda/pull/2307/commits/fd8bf8cbe2f30fd0dea06bf7801f2ceab80c4062)

api/proxy/servers/arbitrum_altda/handlers.go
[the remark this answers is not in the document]
litt3: @ethenotethan I assume you resolved this while my review was still pending, so didn't see my response? Unresolved for now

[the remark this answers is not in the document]
ethenotethan: just referring to the todo comment - would rather just leave as is given how experimental the server implementation still is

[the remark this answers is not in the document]
litt3: My point is just that I have no idea what this TODO is trying to say, but fair enough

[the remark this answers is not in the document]
ethenotethan: ah roger - it's just referring to the third `daprovider_Store` method argument:
```
func (h *Handlers) Store(
	ctx context.Context,
	message hexutil.Bytes,
	timeout hexutil.Uint64,
)
```

PR #2308, opened 2025-11-12
fix(proxy): Dynamic V2 protocol version byte encoding

ethenotethan:
We were doing a switch statement in the ARB Custom DA and REST ALT DA server handlers to manually set the `DA Commitment Version` based on the dispersal backend type. For the EigenDA V2 protocol, the `EigenDACertVerifier` is the canonical source of truth for the DA Cert version being used via the contract's `version()` view call. 

This updates the "version determination" logic to instead be based on what the type returned by the contract vs the dispersal backend. This ensures better long-term extensibility for EigenDA V2 integrations with the medium-term introduction of new cert versions and rollup derivation pipeline hardforks. 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/common/s3/aws/aws_s3_client.go
+++ b/common/s3/aws/aws_s3_client.go
@@ -143,6 +143,50 @@ func (s *awsS3Client) DownloadObject(ctx context.Context, bucket string, key str
 		return nil, s3common.ErrObjectNotFound
 	}
 
+	if len(buffer.Bytes()) != objectSize {
+		return nil, fmt.Errorf("downloaded object size (%d) does not match expected size (%d)",
+			len(buffer.Bytes()), objectSize)
+	}
+
+	return buffer.Bytes(), nil
+}
+
+func (s *awsS3Client) DownloadPartialObject(
+	ctx context.Context,
+	bucket string,
+	key string,
+	startIndex int64,
+	endIndex int64) ([]byte, error) {
+
+	if startIndex < 0 || endIndex <= startIndex {
+		return nil, fmt.Errorf("invalid startIndex (%d) or endIndex (%d)", startIndex, endIndex)
+	}
+
+	rangeHeader := fmt.Sprintf("bytes=%d-%d", startIndex, endIndex-1)
+
+	buffer := manager.NewWriteAtBuffer(make([]byte, 0, endIndex-startIndex))
+
+	var partMiBs int64 = 10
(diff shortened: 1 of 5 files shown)


api/proxy/common/types/certs/eigenda.go
```diff
@@ -6,7 +6,12 @@ import (
 	"github.com/Layr-Labs/eigenda/api/clients/v2/coretypes"
 )
 
-// Version byte that prefixes serialized EigenDACert to identify their type.
+// DA Commitment version byte that prefixes serialized EigenDACert to identify their type.
+// This is off by one with the Cert Version persisted in the EigenDACertVerifier
+// e.g. if CertVerifier.CertVersion() = 3 then DACommit.Version() = 2
```
litt3, line 11: This off-by-one never stops being confusing. Could we just bump to a "new" DACommit.Version without making any changes, so that these line up? If we did that, would there be anything that could happen moving forward to cause these to de-sync again?

If these were synced, could be collapse these two concepts into one?

api/proxy/servers/arbitrum_altda/handlers.go
```diff
@@ -154,7 +154,7 @@ func (h *Handlers) deserializeCertFromSequencerMsg(sequencerMsg hexutil.Bytes) (
 
 	certVersionByte := daCommit[2]
 	versionedCert := certs.NewVersionedCert([]byte(daCommit[DACommitPrefixBytes+1:]), certs.VersionByte(certVersionByte))
-	return versionedCert, nil
+	return *versionedCert, nil
```
litt3: Is there a reason to be switching back and forth between pointers and not? Could we just make all functions accept the `versionedCert` as a pointer?

litt3 (dismissed):

api/proxy/servers/arbitrum_altda/handlers.go
ethenotethan: [99ade09](https://github.com/Layr-Labs/eigenda/pull/2308/commits/99ade09bf7d088a421167f4993fe8f5201031830)

api/proxy/common/types/certs/eigenda.go
ethenotethan: in theory yes but I think we'd still have to maintain the legacy mapping - need to think on this a bit

PR #2312, opened 2025-11-12
fix(proxy): Update errors to return 418

iquidus:
Closes EGDA-2297

Updates the cert conversion and deserialization errors to return http 418.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/api/clients/v2/coretypes/derivation_errors.go
+++ b/api/clients/v2/coretypes/derivation_errors.go
@@ -15,7 +15,8 @@ import (
 // Note: we purposefully don't use StatusCode 0 here, to prevent default value bugs in case people
 // create a DerivationError by hand without using the constructors or sentinel errors defined here.
 var (
-	// Signifies that the input can't be parsed into a versioned cert.
+	// Signifies that the input can't be parsed into a versioned cert,
+	// meaning either the cert has an invalid version byte, or failed to get rlp.decoded from the given hex string
 	ErrCertParsingFailedDerivationError = DerivationError{StatusCode: 1}
 	// Signifies that the cert is invalid due to a recency check failure,
 	// meaning that `cert.L1InclusionBlock > batch.RBN + rbnRecencyWindowSize`.


api/proxy/store/generated_key/v2/eigenda.go
```diff
@@ -90,7 +90,10 @@ func (e Store) Get(
 ) ([]byte, error) {
 	certTypeVersion, err := versionedCert.Version.IntoCertVersion()
 	if err != nil {
-		return nil, fmt.Errorf("casting into cert version: %w", err)
+		return nil, coretypes.NewCertParsingFailedError(
```
bxue-l2, line 93: Not sure if `NewCertParsingFailedError ` is the best name. It is more about unsupported version type.  Though ethen and I argued in the meeting that cert_Version is probably a wrong name, but we better handle it in another PR. So for now, we can either try to do something like [hokulea](https://github.com/Layr-Labs/hokulea/blob/b20a765e3b1cba03e0bfaaca47b474a3ed9dad4f/crates/eigenda-cert/src/altda_commitment.rs#L37); or use this error function, but put comments to explain why the name is inaccurate.

api/proxy/store/generated_key/v2/eigenda.go
```diff
@@ -99,7 +102,10 @@ func (e Store) Get(
 		serializationType,
 	)
 	if err != nil {
-		return nil, fmt.Errorf("deserialize cert: %w", err)
+		return nil, coretypes.NewCertParsingFailedError(
+			hex.EncodeToString(versionedCert.SerializedCert),
```
ethenotethan, line 106: @bxue-l2 this was my doing - curious if there's any LLM magic we can use to try capturing nuances like this at review time

api/proxy/store/generated_key/v2/eigenda.go
ethenotethan: @bxue-l2 hmmm so thinking about this a bit - what happens in the event of a rollup upgrade where the DA Commit version is bumped but is passed to an older version of proxy that doesn't have this conception; e.g. the batch poster's proxy supports V5 but the nodes running derivation pipeline are on V4.

bxue-l2: The proxy should discard the certs. This is what has been done in hokulea, see [here](https://github.com/Layr-Labs/hokulea/blob/ca100c76285bce38f91dfb8a391c32af8d1fc741/crates/eigenda/src/eigenda.rs#L101) and [here](https://github.com/Layr-Labs/hokulea/blob/ca100c76285bce38f91dfb8a391c32af8d1fc741/crates/eigenda-cert/src/altda_commitment.rs#L83), though it should really renamed to unsupported derivation version, if we finalize to work on this.

Imagine a malicious batcher sends an unrecognized cert/derivation version to L1 inbox, the best L2 consensus to do is to drop it. 

That is also why in the new derivation upgrade design, we need the source of activation number (either provided by router or offchain config), to make it such that it is impossible for a malicious sequencer to submit a cert with older derivation.

bxue-l2: A rollup governace should organize an upgrade, at the timestamp. In this design, we can't let batcher to choose arbitrary derivation version.

iquidus: `NewCertParsingFailedError` is what we are using elsewhere for this case. e.g: https://github.com/Layr-Labs/eigenda/blob/c264fe9eaa115abd04d9ba7e33736886e9ab28dc/api/proxy/store/generated_key/v2/eigenda.go#L225-L241

bxue-l2: Then it feels like that needs a cleanup too? But it can be done in another PR.

PR #2313, opened 2025-11-12
feat: add config retriever

0xClandestine:

https://github.com/Layr-Labs/eigenda/pull/2195

## Checks

- Testing Strategy
   - [x] Unit tests
   - [ ] Integration tests

### DIFF
--- a/contracts/src/periphery/cfg-getter/EigenDAConfigRetriever.sol
+++ b/contracts/src/periphery/cfg-getter/EigenDAConfigRetriever.sol
@@ -1,0 +1,85 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+import {IEigenDAConfigRegistry} from "src/core/interfaces/IEigenDADirectory.sol";
+import {ConfigRegistryTypes} from "src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol";
+import {ConfigRegistryLib} from "src/core/libraries/v3/config-registry/ConfigRegistryLib.sol";
+
+/// @title EigenDAConfigRetriever
+/// @notice A stateless contract that defines getter functions using the underlying configuration registry's primitive API.
+contract EigenDAConfigRetriever {
+    IEigenDAConfigRegistry public immutable configRegistry;
+
+    constructor(address configRegistryAddress) {
+        configRegistry = IEigenDAConfigRegistry(configRegistryAddress);
+    }
+
+    /// @notice Retrieves the currently active bytes32 configuration checkpoint and all future checkpoints for a given name.
+    /// @dev Returns the checkpoint with the highest activation key that is less than or equal to the provided activation key,
+    ///      plus all checkpoints with activation keys greater than the provided activation key.
+    ///      This allows clients to know the current configuration value and plan ahead for upcoming updates.
+    function getActiveAndFutureBytes32Configs(string memory name, uint256 activationKey)
+        external
+        view
+        returns (ConfigRegistryTypes.Bytes32Checkpoint[] memory)
+    {
+        bytes32 nameDigest = ConfigRegistryLib.getNameDigest(name);
+        uint256 numCheckpoints = configRegistry.getNumCheckpointsBytes32(nameDigest);
+
+        // There are 3 cases to handle:
+        // 1. If no checkpoints have activation keys less than or equal to the provided activation key, we return an empty array.
+        // 2. If all checkpoints have activation keys less than or equal to the provided activation key, we return the last checkpoint only.
+        // 3. If some checkpoints have activation keys less than or equal to the provided activation key, we return the currently active checkpoint and all future ones.
+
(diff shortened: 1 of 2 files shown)


contracts/src/periphery/cfg-getter/EigenDAConfigRetriever.sol
```diff
@@ -0,0 +1,85 @@
...
+        // There are 3 cases to handle:
+        // 1. If no checkpoints have activation keys less than or equal to the provided activation key, we return an empty array.
+        // 2. If all checkpoints have activation keys less than or equal to the provided activation key, we return the last checkpoint only.
+        // 3. If some checkpoints have activation keys less than or equal to the provided activation key, we return the currently active checkpoint and all future ones.
+
+        uint256 startIndex = numCheckpoints; // Default to numCheckpoints (case 1)
+        for (uint256 i = 0; i < numCheckpoints; i++) {
+            uint256 checkpointActivationKey = configRegistry.getActivationKeyBytes32(nameDigest, numCheckpoints - 1 - i);
+            if (checkpointActivationKey <= activationKey) {
+                startIndex = numCheckpoints - 1 - i; // Found the currently active checkpoint (include it)
```
litt3, line 38: Nice find on the off-by-one that existed in the previous PR 🚀

contracts/src/periphery/cfg-getter/EigenDAConfigRetriever.sol
```diff
@@ -0,0 +1,85 @@
...
+    function getActiveAndFutureBytes32Configs(string memory name, uint256 activationKey)
+        external
+        view
+        returns (ConfigRegistryTypes.Bytes32Checkpoint[] memory)
+    {
+        bytes32 nameDigest = ConfigRegistryLib.getNameDigest(name);
+        uint256 numCheckpoints = configRegistry.getNumCheckpointsBytes32(nameDigest);
+
+        // There are 3 cases to handle:
+        // 1. If no checkpoints have activation keys less than or equal to the provided activation key, we return an empty array.
```
litt3, line 30: I know you just carried this logic forward, but IMO this isn't the behavior we actually want
```suggestion
        // 1. If no checkpoints have activation keys less than or equal to the provided activation key, we return all checkpoints as future values.
```

I don't know why we'd want it to return empty if there isn't any active value. The idea is that we should see future updates coming in advance, so it makes sense you'd just return all future values even if there isn't any active value

contracts/src/periphery/cfg-getter/EigenDAConfigRetriever.sol
```diff
@@ -0,0 +1,85 @@
...
+        // 2. If all checkpoints have activation keys less than or equal to the provided activation key, we return the last checkpoint only.
+        // 3. If some checkpoints have activation keys less than or equal to the provided activation key, we return the currently active checkpoint and all future ones.
+
+        uint256 startIndex = numCheckpoints; // Default to numCheckpoints (case 1)
```
litt3, line 34:
```suggestion
        uint256 startIndex = 0; // Default to 0 (case 1: all checkpoints are future)
```

And here is the corresponding logic change. (of course, the duplicate logic and docs below should be changed too)

contracts/src/periphery/cfg-getter/EigenDAConfigRetriever.sol
```diff
@@ -0,0 +1,85 @@
...
+contract EigenDAConfigRetriever {
+    IEigenDAConfigRegistry public immutable configRegistry;
+
+    constructor(address configRegistryAddress) {
+        configRegistry = IEigenDAConfigRegistry(configRegistryAddress);
+    }
+
+    /// @notice Retrieves the currently active bytes32 configuration checkpoint and all future checkpoints for a given name.
+    /// @dev Returns the checkpoint with the highest activation key that is less than or equal to the provided activation key,
+    ///      plus all checkpoints with activation keys greater than the provided activation key.
```
litt3, line 19:
```suggestion
    ///      If no checkpoint is active (all activation keys are greater than the provided key), returns all checkpoints as future values.
```

contracts/src/periphery/cfg-getter/EigenDAConfigRetriever.sol
```diff
@@ -0,0 +1,85 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+import {IEigenDAConfigRegistry} from "src/core/interfaces/IEigenDADirectory.sol";
+import {ConfigRegistryTypes} from "src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol";
+import {ConfigRegistryLib} from "src/core/libraries/v3/config-registry/ConfigRegistryLib.sol";
+
+/// @title EigenDAConfigRetriever
+/// @notice A stateless contract that defines getter functions using the underlying configuration registry's primitive API.
```
ethenotethan, line 9: can we add a comment specifying this is only intended for offchain usage?

PR #2317, opened 2025-11-13
feat: efficient range query

cody-littley:
Use a more efficient access pattern for downloading chunk ranges.


relay/server.go
```diff
@@ -408,8 +439,144 @@ func getKeysFromChunkRequest(request *pb.GetChunksRequest) ([]v2.BlobKey, error)
...
+	results := make(chan downloadResult, requestCount*2)
+
+	// Fan out and make requests in parallel
+	for i, chunkRequest := range request.GetChunkRequests() {
+		blobKey := v2.BlobKey(chunkRequest.GetByRange().GetBlobKey())
+		metadata := metadataMap[blobKey]
+
+		// Download proofs
+		go func() {
+			data, found, err := s.chunkReader.GetBinaryChunkProofsRange(
```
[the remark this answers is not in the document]
cody-littley: Not a real bug, this golang language level problem is fixed in go 1.22+. We are on 1.24

relay/chunkstore/chunk_reader.go
```diff
@@ -71,30 +91,97 @@ func (r *chunkReader) GetBinaryChunkProofs(ctx context.Context, blobKey corev2.B
...
+// TODO handle not found error properly!!!
+
+func (r *chunkReader) GetBinaryChunkProofsRange(
+	ctx context.Context,
+	blobKey corev2.BlobKey,
+	startIndex uint32,
+	endIndex uint32,
+) ([][]byte, bool, error) {
+
+	firstByteIndex := startIndex * encoding.SerializedProofLength
```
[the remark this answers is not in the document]
cody-littley: Not a real bug. Attacker already has the ability to read any chunk they want from this s3 object, or even the whole object. Playing games with indices isn't necessary, and doesn't harm us any more than requesting redundant data would.

common/s3/oci/oci_s3_client.go
```diff
@@ -116,7 +113,10 @@ func (c *ociS3Client) DownloadObject(ctx context.Context, bucket string, key str
 
 	response, err := c.objectStorageClient.GetObject(ctx, getObjectRequest)
 	if err != nil {
-		return nil, fmt.Errorf("failed to get object from OCI: %w", err)
+		if response.RawResponse != nil && response.RawResponse.StatusCode == 404 {
```
litt3, line 116: Will `err` actually be non-nil in this case? Or should this check be outside the `err != nil` block?

(whatever the result of this comment is ought to be applied below in `DownloadPartialObject` as well)

common/s3/aws/aws_s3_client.go
```diff
@@ -136,30 +137,34 @@ func (s *awsS3Client) DownloadObject(ctx context.Context, bucket string, key str
 		Key:    aws.String(key),
 	})
 	if err != nil {
-		return nil, err
+		errString := err.Error()
+		if strings.Contains(errString, "StatusCode: 404") {
```
litt3, line 141: Did you test that this magic string is actually what's contained in the 404 case?

disperser/common/blobstore/shared_storage.go
```diff
@@ -146,11 +146,15 @@ func (s *SharedBlobStore) StoreBlob(ctx context.Context, blob *core.Blob, reques
 
 // GetBlobContent retrieves blob content by the blob key.
 func (s *SharedBlobStore) GetBlobContent(ctx context.Context, blobHash disperser.BlobHash) ([]byte, error) {
-	return s.s3Client.DownloadObject(ctx, s.bucketName, blobObjectKey(blobHash))
+	data, _, err := s.s3Client.DownloadObject(ctx, s.bucketName, blobObjectKey(blobHash))
```
litt3: Is this actually what you want to do? Silently return empty data if the object wasn't available?

encoding/v2/rs/frame_coeffs.go
```diff
@@ -106,3 +106,24 @@ func DeserializeSplitFrameCoeffs(elementCount uint32, binaryFrameCoeffs [][]byte
...
 	return coeffs
 }
+
+// SplitSerializedFrameCoeffsWithElementCount splits serialized frame coefficients data into a slice of byte slices,
+// each containing the serialized data for a single FrameCoeffs object.
+func SplitSerializedFrameCoeffsWithElementCount(serializedData []byte, symbolsPerFrame uint32) ([][]byte, error) {
+	index := uint32(0)
+
+	bytesPerFrameCoeffs := encoding.BYTES_PER_SYMBOL * symbolsPerFrame
+	remainingBytes := uint32(len(serializedData[index:]))
```
litt3:
```suggestion
	remainingBytes := uint32(len(serializedData))
```

Alternatively, just get the length once and store in a local variable, since you're using it in multiple places

relay/server.go
```diff
@@ -408,8 +439,144 @@ func getKeysFromChunkRequest(request *pb.GetChunksRequest) ([]v2.BlobKey, error)
...
+				chunkRequest.GetByRange().GetStartIndex(),
+				chunkRequest.GetByRange().GetEndIndex(),
+			)
+
+			proofs[i] = data
+			if err != nil {
+				err = fmt.Errorf("failed to download proofs: %w", err)
+			}
+			results <- downloadResult{key: blobKey, found: found, err: err}
+		}()
```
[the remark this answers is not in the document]
cody-littley: not a bug

relay/chunkstore/chunk_reader.go
```diff
@@ -71,30 +91,109 @@ func (r *chunkReader) GetBinaryChunkProofs(ctx context.Context, blobKey corev2.B
...
+) (uint32, [][]byte, error) {
 
-	bytes, err := r.client.FragmentedDownloadObject(
-		ctx,
-		r.bucket,
-		s3.ScopedChunkKey(blobKey),
-		int(fragmentInfo.TotalChunkSizeBytes),
-		int(fragmentInfo.FragmentSizeBytes))
+	bytes, found, err := r.client.DownloadObject(ctx, r.bucket, s3.ScopedChunkKey(blobKey))
+	if !found {
```
litt3: You should check if `err!=nil` before checking `found`. `found` is false in the case where `err!=nil`

relay/chunkstore/chunk_writer.go
```diff
@@ -104,11 +93,11 @@ func (c *chunkWriter) ProofExists(ctx context.Context, blobKey corev2.BlobKey) b
...
-func (c *chunkWriter) CoefficientsExists(ctx context.Context, blobKey corev2.BlobKey) (bool, *encoding.FragmentInfo) {
+func (c *chunkWriter) CoefficientsExists(ctx context.Context, blobKey corev2.BlobKey) bool {
 	// TODO(ian-shim): check latency
 	objs, err := c.s3Client.ListObjects(ctx, c.bucketName, s3.ScopedChunkKey(blobKey))
 	if err != nil {
-		return false, nil
+		return false
 	}
 
 	keys := make([]string, len(objs))
```
litt3: this loop is vestigial

common/s3/s3_client.go
```diff
@@ -42,35 +43,6 @@ type S3Client interface {
...
-	// smaller parts and uploaded in parallel. The file will be reassembled on download.
-	//
-	// Note: if a file is uploaded with this method, only the FragmentedDownloadObject method should be used to
-	// download the file. It is not advised to use DeleteObject on files uploaded with this method (if such
-	// functionality is required, a new method to do so should be added to this interface).
-	//
-	// Note: if this operation fails partway through, some file fragments may have made it to S3 and others may not.
-	// In order to prevent long term accumulation of fragments, it is suggested to use this method in conjunction with
-	// a bucket configured to have a TTL.
-	FragmentedUploadObject(
```
litt3, line 57: `FragmentedUploadObject` is gone, but it's still mentioned by name in the doc for `ListObjects`

relay/chunkstore/chunk_writer.go
```diff
@@ -104,11 +93,11 @@ func (c *chunkWriter) ProofExists(ctx context.Context, blobKey corev2.BlobKey) b
 	return false
 }
 
-func (c *chunkWriter) CoefficientsExists(ctx context.Context, blobKey corev2.BlobKey) (bool, *encoding.FragmentInfo) {
+func (c *chunkWriter) CoefficientsExists(ctx context.Context, blobKey corev2.BlobKey) bool {
```
litt3, line 96: Are you actually accomplishing anything with this method anymore?

It used to list objects, and then do some kind of checking. Now, it just lists whatever is in the bucket, and returns `true` if there was no error. It's not clear to me that this method would ever return `false`

relay/metadata_provider.go
```diff
@@ -212,16 +210,12 @@ func (m *metadataProvider) fetchMetadata(key v2.BlobKey) (*blobMetadata, error)
 	// TODO(cody-littley): blob size is not correct https://github.com/Layr-Labs/eigenda/pull/906#discussion_r1847396530
 	blobSize := uint32(cert.BlobHeader.BlobCommitments.Length) * encoding.BYTES_PER_SYMBOL
 
-	chunkSize, err := m.computeChunkSize(cert.BlobHeader, fragmentInfo.TotalChunkSizeBytes)
```
litt3, line 215: `computeChunkSize` is now unused

relay/chunkstore/chunk_reader.go
```diff
@@ -71,30 +91,109 @@ func (r *chunkReader) GetBinaryChunkProofs(ctx context.Context, blobKey corev2.B
...
+	size := count * encoding.SerializedProofLength
+
+	s3Key := s3.ScopedProofKey(blobKey)
+
+	data, found, err := r.client.DownloadPartialObject(
+		ctx,
+		r.bucket,
+		s3Key,
+		int64(firstByteIndex),
+		int64(firstByteIndex+size))
```
litt3, line 135: seems like you're doing a lot of gymnastics to compute this value, instead of just using `endIndex` passed in as a parameter

same comment applies for `GetBinaryChunkCoefficientRange` method below

relay/server.go
```diff
@@ -408,8 +439,144 @@ func getKeysFromChunkRequest(request *pb.GetChunksRequest) ([]v2.BlobKey, error)
...
-func gatherChunkDataToSend(
+// Used to pass status of downloads from goroutines up to controlling function.
+type downloadResult struct {
+	key   v2.BlobKey
+	found bool
+	err   error
+}
+
+// Download and compile the chunk data to send back to the client.
+func (s *Server) gatherChunkDataToSend(
```
litt3, line 450: this new method is very long and hard to read. Could it be broken up into more manageable parts?

relay/metadata_provider.go
```diff
@@ -17,15 +17,13 @@ import (
 
 // Metadata about a blob. The relay only needs a small subset of a blob's metadata.
 // This struct adds caching and threading on top of blobstore.BlobMetadataStore.
-type blobMetadata struct {
+type blobMetadata struct { // TODO what is actually needed here?
```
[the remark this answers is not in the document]
litt3: @cody-littley You resolved this, but I agree that the TODO is vague and could be improved

relay/chunkstore/chunk_reader.go
```diff
@@ -71,30 +91,109 @@ func (r *chunkReader) GetBinaryChunkProofs(ctx context.Context, blobKey corev2.B
...
 	elementCount, frames, err := rs.SplitSerializedFrameCoeffs(bytes)
 	if err != nil {
-		r.logger.Error("failed to split coefficient frames", "blob", blobKey.Hex(), "error", err)
 		return 0, nil, fmt.Errorf("failed to split coefficient frames for blob %s: %w", blobKey.Hex(), err)
 	}
 
 	return elementCount, frames, nil
 }
+
+// TODO handle not found error properly!!!
```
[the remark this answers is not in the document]
cody-littley: removed

relay/metadata_provider.go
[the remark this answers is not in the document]
cody-littley: Can't remember why I resolved the comment. The TODO was stale though, so I removed it.

common/s3/oci/oci_s3_client.go
cody-littley: I'm actually not sure how this will work. This API is undocumented, and I don't have any way to test it. The current iteration is the way claude guessed it would work... not super confident in the result but it's all I currently have to go on.

If I've guessed wrong, the net effect will be that we get alert-triggering errors if we have a not-found error when we deploy to OCI. Not the end of the world, and we can fix the issue if it arises.

common/s3/aws/aws_s3_client.go
cody-littley: Yup, we've got unit tests that test this workflow via localstack (which is supposed to mimic actual dynamo behavior). That's actually how I derived the proper magic string.

This is one of the primary reasons why I strongly oppose passing non-debug data as error messages. It can be a nightmare for end users to determine API of a method, short of trial and error.

disperser/common/blobstore/shared_storage.go
cody-littley: Good point, fixed. This method should really return a `bool`, but I'd like to avoid increasing the blast radius of this PR even further. So I'll just return an error for now.

encoding/v2/rs/frame_coeffs.go
cody-littley: done

relay/chunkstore/chunk_reader.go
cody-littley: Good point. Fixed.

relay/chunkstore/chunk_writer.go
cody-littley: fixed

common/s3/s3_client.go
cody-littley: fixed

relay/chunkstore/chunk_writer.go
cody-littley: Good point, made this use the `HeadObject()` pathway.

relay/metadata_provider.go
cody-littley: removed

relay/chunkstore/chunk_reader.go
cody-littley: The end index passed into the method is the chunk index. The indices passed into `DownloadPartialObject()` are the byte indices in the serialized data, so the computation is needed. Or am I misunderstanding your suggestion?

relay/server.go
cody-littley: Split into several helper methods.

disperser/common/blobstore/shared_storage.go
```diff
@@ -163,7 +166,7 @@ func (s *SharedBlobStore) getBlobContentParallel(ctx context.Context, blobKey di
 }
 
 func (s *SharedBlobStore) MarkBlobConfirmed(ctx context.Context, existingMetadata *disperser.BlobMetadata, confirmationInfo *disperser.ConfirmationInfo) (*disperser.BlobMetadata, error) {
-	// TODO (ian-shim): remove this check once we are sure that the metadata is never overwritten
+	// frame (ian-shim): remove this check once we are sure that the metadata is never overwritten
```
litt3: Was this change intentional? "TODO" -> "frame"?

disperser/common/blobstore/shared_storage.go
```diff
@@ -146,11 +146,18 @@ func (s *SharedBlobStore) StoreBlob(ctx context.Context, blob *core.Blob, reques
...
+	}
+	if !found {
+		return nil, fmt.Errorf("blob not found for blob hash: %s", blobHash)
+	}
+	return data, nil
 }
 
 func (s *SharedBlobStore) getBlobContentParallel(ctx context.Context, blobKey disperser.BlobKey, blobRequestHeader core.BlobRequestHeader, resultChan chan<- blobResultOrError) {
-	blob, err := s.s3Client.DownloadObject(ctx, s.bucketName, blobObjectKey(blobKey.BlobHash))
+	blob, _, err := s.s3Client.DownloadObject(ctx, s.bucketName, blobObjectKey(blobKey.BlobHash))
```
litt3: Here's another place where you're discarding the boolean, which seems wrong

encoding/v2/rs/frame_coeffs.go
```diff
@@ -106,3 +106,25 @@ func DeserializeSplitFrameCoeffs(elementCount uint32, binaryFrameCoeffs [][]byte
...
+
+// SplitSerializedFrameCoeffsWithElementCount splits serialized frame coefficients data into a slice of byte slices,
+// each containing the serialized data for a single FrameCoeffs object.
+func SplitSerializedFrameCoeffsWithElementCount(serializedData []byte, symbolsPerFrame uint32) ([][]byte, error) {
+	index := uint32(0)
+	remainingBytes := uint32(len(serializedData))
+	bytesPerFrameCoeffs := encoding.BYTES_PER_SYMBOL * symbolsPerFrame
+
+	if remainingBytes%bytesPerFrameCoeffs != 0 {
+		return nil, fmt.Errorf("invalid data size: %d", len(serializedData))
```
litt3:
```suggestion
		return nil, fmt.Errorf("invalid data size: %d", remainingBytes)
```

relay/chunkstore/chunk_reader.go
litt3: I was looking incorrectly, and interpreted `startIndex` and `endIndex` as _bytes_ instead of _chunks_

One potential change would be to qualify both index types that exist in this method, e.g. `firstChunkIndex` as a corollary  to `firstByteIndex`. But that's not required

relay/server.go
```diff
@@ -408,8 +439,191 @@ func getKeysFromChunkRequest(request *pb.GetChunksRequest) ([]v2.BlobKey, error)
...
+	coefficients = make([][][]byte, requestCount)
+	proofs = make([][][]byte, requestCount)
+
+	ctx, cancel := context.WithCancel(ctx)
+	defer cancel()
+
+	results := make(chan downloadResult, requestCount*2)
+
+	// Fan out and make requests in parallel
+	for i, chunkRequest := range request.GetChunkRequests() {
```
litt3, line 497: I think this might be a case where using `errgroup` would make sense. You're manually doing what `errgroup` does: it performs a set of tasks, and cancels if any individual task fails

Not a required change, though.

relay/server.go
```diff
@@ -408,8 +439,191 @@ func getKeysFromChunkRequest(request *pb.GetChunksRequest) ([]v2.BlobKey, error)
...
+
+	requestCount := len(request.GetChunkRequests())
+
+	chunkDataObjects := make([]*core.ChunksData, requestCount)
+	for i := 0; i < requestCount; i++ {
+		blobKey := v2.BlobKey(request.GetChunkRequests()[i].GetByRange().GetBlobKey())
+		metadata := metadataMap[blobKey]
+		chunkData, err := buildChunksData(proofs[i], int(metadata.symbolsPerFrame), coefficients[i])
+		if err != nil {
+			return nil, fmt.Errorf("error building frame data: %w", err)
```
litt3:
```suggestion
			return nil, fmt.Errorf("error building chunk data: %w", err)
```

relay/server.go
```diff
@@ -408,8 +439,191 @@ func getKeysFromChunkRequest(request *pb.GetChunksRequest) ([]v2.BlobKey, error)
...
+			followingChunkData := chunkDataObjects[i]
+			chunkDataToSend.Chunks = append(chunkDataToSend.Chunks, followingChunkData.Chunks...)
+
+			// Bump the counter for the outer loop since this iteration handles it
+			requestIndex++
+		}
+
+		bundleBytes, err := chunkDataToSend.FlattenToBundle()
+		if err != nil {
+			return nil, fmt.Errorf("error serializing frame subset: %w", err)
```
litt3: error message doesn't really match new naming

relay/chunkstore/chunk_reader.go
cody-littley: renamed

disperser/common/blobstore/shared_storage.go
cody-littley: Odd... I don't remember doing this, and this was too complex to let claude do it for me. Reverted.

disperser/common/blobstore/shared_storage.go
cody-littley: This is v1 code I think, so it will never be deployed (we are still deploying v1 from 0.9.0). But fixed anyways.

relay/server.go
cody-littley: converted to `errgroup`

relay/server.go
cody-littley: fixed

relay/server.go
cody-littley: fixed

encoding/v2/rs/frame_coeffs.go
litt3: @cody-littley Looks like you missed this one

cody-littley: oops, fixed

PR #2321, opened 2025-11-14
chore(payments): Remove legacy accountant

litt3:
- This removes the option for clients to use the old accounting logic
- It doesn't break backwards compatibility for anyone still using old clients, i.e. the disperser is still compatible with the old logic

### DIFF
--- a/test/v2/live/live_network_test.go
+++ b/test/v2/live/live_network_test.go
@@ -2,7 +2,9 @@ package live
 
 import (
 	"context"
+	"errors"
 	"fmt"
+	"os"
 	"strings"
 	"testing"
 	"time"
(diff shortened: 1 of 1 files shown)


api/clients/v2/disperser_client.go
```diff
@@ -117,30 +110,10 @@ func NewDisperserClient(
 		signer:     signer,
 		clientPool: clientPool,
 		committer:  committer,
-		accountant: accountant,
 		metrics:    metrics,
 	}, nil
 }
 
-// PopulateAccountant populates the accountant with the payment state from the disperser.
-func (c *DisperserClient) PopulateAccountant(ctx context.Context) error {
```
samlaf, line 126: bye bvye

litt3: 👋

PR #2323, opened 2025-11-14
feat: qol improvements

0xClandestine:

We're using a significantly outdated version of Foundry, and it shows.

- Full foundry.toml added.
- Solidity bumped -> 0.8.29
- Fixed all typos
- Removed all unused imports

## Checks

- Testing Strategy
   - [x] Unit tests
   - [x] Integration tests


contracts/script/EjectionManagerDeployer.s.sol
```diff
@@ -1,5 +1,5 @@
 // SPDX-License-Identifier: BUSL-1.1
-pragma solidity =0.8.12;
+pragma solidity ^0.8.12;
```
0xClandestine, line 2: Some pragmas were hardcoded for some reason...

PR #2326, opened 2025-11-17
feat(proxy-arb): Add invalid cert handling toggle

ethenotethan:
- Adds handler tests to assert correctness of handler logic
- Adds a `PROCESS_INVALID_CERT` flag which when set to true results in the return of a `CertificateValidationError` which indicates "drop cert" to the arbitrum nitro software 

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/contracts/script/EigenDADeployer.s.sol
+++ b/contracts/script/EigenDADeployer.s.sol
@@ -385,7 +385,6 @@ contract EigenDADeployer is DeployOpenEigenLayer {
             defaultSecurityThresholds,
             hex"0001"
         );
-        eigenDADirectory.addAddress(AddressDirectoryConstants.CERT_VERIFIER_NAME, address(eigenDACertVerifier));
 
         eigenDACertVerifierRouterImplementation = new EigenDACertVerifierRouter();
 
--- a/contracts/script/deploy/eigenda/DeployEigenDA.s.sol
+++ b/contracts/script/deploy/eigenda/DeployEigenDA.s.sol
@@ -302,22 +302,19 @@ contract DeployEigenDA is Script {
             AddressDirectoryConstants.OPERATOR_STATE_RETRIEVER_NAME, address(new OperatorStateRetriever())
         );
 
-        directory.addAddress(
-            AddressDirectoryConstants.CERT_VERIFIER_NAME,
-            address(
-                new EigenDACertVerifier(
-                    IEigenDAThresholdRegistry(directory.getAddress(AddressDirectoryConstants.THRESHOLD_REGISTRY_NAME)),
-                    IEigenDASignatureVerifier(directory.getAddress(AddressDirectoryConstants.STAKE_REGISTRY_NAME)),
-                    cfg.certVerifierSecurityThresholds(),
-                    cfg.certVerifierQuorumNumbersRequired()
-                )
+        address certVerifier = address(
+            new EigenDACertVerifier(
+                IEigenDAThresholdRegistry(directory.getAddress(AddressDirectoryConstants.THRESHOLD_REGISTRY_NAME)),
+                IEigenDASignatureVerifier(directory.getAddress(AddressDirectoryConstants.STAKE_REGISTRY_NAME)),
+                cfg.certVerifierSecurityThresholds(),
+                cfg.certVerifierQuorumNumbersRequired()
             )
         );
 
         address routerImpl = address(new EigenDACertVerifierRouter());
         address[] memory certVerifiers = new address[](1);
 
-        certVerifiers[0] = directory.getAddress(AddressDirectoryConstants.CERT_VERIFIER_NAME);
+        certVerifiers[0] = certVerifier;
 
         directory.addAddress(
             AddressDirectoryConstants.CERT_VERIFIER_ROUTER_NAME,
--- a/contracts/src/core/libraries/v3/address-directory/AddressDirectoryConstants.sol
+++ b/contracts/src/core/libraries/v3/address-directory/AddressDirectoryConstants.sol
@@ -30,7 +30,6 @@ library AddressDirectoryConstants {
     /// @dev This name is prefixed with EIGEN_DA to differentiate it from the previous ejection manager which was vendored from eigenlayer-middleware.
     string internal constant EIGEN_DA_EJECTION_MANAGER_NAME = "EIGEN_DA_EJECTION_MANAGER";
 
-    string internal constant CERT_VERIFIER_NAME = "CERT_VERIFIER";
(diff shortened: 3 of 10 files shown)


api/proxy/docs/help_out.txt
```diff
@@ -25,6 +25,13 @@ GLOBAL OPTIONS:
    
     --arbitrum-da.port value            (default: 3101)                    ($EIGENDA_PROXY_ARB_DA_PORT)
           Server listening port
+   
+    --arbitrum-da.process-invalid-cert  (default: false)                   ($EIGENDA_PROXY_ARB_DA_PROCESS_INVALID_CERT)
+          Whether or not the CustomDA server should return a `CertificateValidationError`
+          to the arbitrum nitro derivation pipeline which "drops" the DA Cert by treating
```
samlaf: is `process-invalid-cert` the right wording here? what is returned when this is false? are both returning 400s just with a different error in the body?

api/proxy/servers/arbitrum_altda/handlers.go
```diff
@@ -102,17 +102,21 @@ type Handlers struct {
...
+	processInvalidCert bool
+	log                logging.Logger
+	eigenDAManager     store.IEigenDAManager
+	compatibilityCfg   proxy_common.CompatibilityConfig
 }
 
 // NewHandlers is a constructor
-func NewHandlers(m *store.EigenDAManager, l logging.Logger, compatCfg proxy_common.CompatibilityConfig) IHandlers {
+func NewHandlers(
+	m store.IEigenDAManager, l logging.Logger, pic bool, compatCfg proxy_common.CompatibilityConfig,
```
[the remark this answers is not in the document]
samlaf: agree

api/proxy/servers/arbitrum_altda/types.go
```diff
@@ -15,6 +15,10 @@ var (
 	// ErrFallbackRequested is returned by a CustomDA provider to explicitly signal that
 	// the batch poster should fall back to the next available DA writer (e.g, AnyTrust).
 	ErrFallbackRequested = errors.New("DA provider requests fallback to next writer")
+
+	// CertificateValidationError is returned by a CustomDA provider to signal an "invalid DA Cert"
+	// condition.
+	CertificateValidationError = errors.New("certificate validation failed")
```
samlaf: whats the point of this error? I don't see it used anywhere. Who should be matching for this?
I'm guessing the string it returned as text and then you'll parse it on the rollup side? Why isnt it used by the proxy std client instead? I'm prob missing something..

samlaf (commented):
Not following how this error is returned. Left some questions.
Also before you were returning nil, nil. Did the arb spec change? Or you just added new logic on the orllup side to process non-nil? Or is nil still meaningful to drop cert? Bit confused.

api/proxy/servers/arbitrum_altda/types.go
ethenotethan: This is an arbitrum specific signal: https://github.com/OffchainLabs/nitro/blob/2b3cf2138b17af6411e7e391eac346267fec121a/daprovider/reader.go#L19-L31

api/proxy/docs/help_out.txt
ethenotethan: technically since it's JSON RPC there isn't a notion of 400 return. both would be returning 200s with a different error body where (`process-invalid-cert=true`) would return a `CertificateValidationError` that the nitro derivation pipeline understands to discard the DA Cert and continue

api/proxy/servers/arbitrum_altda/types.go
ethenotethan: I've added a code link and changed the error name to adhere to linting format: [43b56f9](https://github.com/Layr-Labs/eigenda/pull/2326/commits/43b56f9ede54b7158b75c4c12bf11bb02de915af)

ethenotethan: also yes - OCL updated the "invalid cert" handling with this explicit `CertificateValidationError ` error with a recent PR last week

api/proxy/docs/help_out.txt
samlaf: Interesting I hadnt even realized that the spec doesnt use the underlying http codes. Though Claude seems to say a lot of implementations actually do.
image

samlaf: In any case the real point of my question is I don't quite understand what `process-invalid-cert` means even after your explanation. It seems like in both cases the invalid cert is processed, and its more about what is returned (nil or `CertificateValidationError`)?

api/proxy/servers/arbitrum_altda/types.go
samlaf: I see. I guess my main point is.. is this error string ("certificate validation failed") serialized into the response body and parsed on the nitro side? Hopefully they don't change the string yet another time lol.

Seems like it would have been better to follow the json-rpc spec and use a [application reserved error code](https://www.jsonrpc.org/specification#error_object) in the range [-32000,-32099]? Maybe you can ask them why they didn't go this route.

api/proxy/servers/arbitrum_altda/types.go
```diff
@@ -15,6 +15,13 @@ var (
 	// ErrFallbackRequested is returned by a CustomDA provider to explicitly signal that
 	// the batch poster should fall back to the next available DA writer (e.g, AnyTrust).
 	ErrFallbackRequested = errors.New("DA provider requests fallback to next writer")
+
+	// Vendored from:
+	// https://github.com/OffchainLabs/nitro/blob/2b3cf2138b17af6411e7e391eac346267fec121a/daprovider/reader.go#L19-L31
+	//
+	// ErCertValidationError is returned by a CustomDA provider to signal an "invalid DA Cert"
+	// condition to the Arbitrum derivation pipeline.
+	ErCertValidationError = errors.New("certificate validation failed")
```
samlaf:
```suggestion
	ErrCertValidationError = errors.New("certificate validation failed")
```

ethenotethan: [754e4e2](https://github.com/Layr-Labs/eigenda/pull/2326/commits/754e4e22b2eacbc7e8b6f7af5b4ede8d1fca1b67)

api/proxy/docs/help_out.txt
ethenotethan: yup - updated the user CLI flag: [754e4e2](https://github.com/Layr-Labs/eigenda/pull/2326/commits/754e4e22b2eacbc7e8b6f7af5b4ede8d1fca1b67)

PR #2329, opened 2025-11-17
feat: Implement `SafeDial` to avoid leaking RPC secrets in logs

litt3:
closes EGDA-2327 ([link](https://linear.app/eigenlabs/issue/EGDA-2327/protect-against-leaking-rpc-secrets))

- Implements utilities necessary to safely dial and log descriptive errors when connecting to RPCs
- Modifies all unsafe logging to use the new utils

### DIFF
--- a/common/geth/rpc_utils_test.go
+++ b/common/geth/rpc_utils_test.go
@@ -9,7 +9,7 @@ import (
 func TestSanitizeRpcUrl(t *testing.T) {
 	require.Equal(t, "https://rpc.example.com", SanitizeRpcUrl("https://user:password@rpc.example.com"))
 	require.Equal(t, "https://rpc.example.com", SanitizeRpcUrl("https://rpc.example.com/v2/SECRET_API_KEY"))
-	require.Equal(t, "https://rpc.example.com", SanitizeRpcUrl("https://rpc.example.com?apikey=SECRET"))
+	require.Equal(t, "https://rpc.example.com", SanitizeRpcUrl("https://rpc.example.com/eth_network?apikey=SECRET"))
 	require.Equal(t, "https://rpc.example.com", SanitizeRpcUrl("https://SECRET_KEY@rpc.example.com"))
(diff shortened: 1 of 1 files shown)


common/geth/rpc_utils_test.go
```diff
@@ -0,0 +1,17 @@
+package geth
+
+import (
+	"testing"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestSanitizeRpcUrl(t *testing.T) {
+	require.Equal(t, "https://rpc.example.com", SanitizeRpcUrl("https://user:password@rpc.example.com"))
```
pschork, line 10: can you add a quicknode style check for url like 
`https://foo-bar-baz.ethereum-hoodi.quiknode.pro/1234512345`
where account id = `1234512345` is stripped?

common/geth/rpc_utils_test.go
```diff
@@ -0,0 +1,17 @@
...
+	require.Equal(t, "https://rpc.example.com", SanitizeRpcUrl("https://rpc.example.com/v2/SECRET_API_KEY"))
+	require.Equal(t, "https://rpc.example.com", SanitizeRpcUrl("https://rpc.example.com?apikey=SECRET"))
```
pschork: can you include network ie `https://rpc.example.com/eth_hoodi?apikey=SECRET"`

common/geth/rpc_utils_test.go
litt3: Is that different from this test?

```
	require.Equal(t, "https://rpc.example.com", SanitizeRpcUrl("https://rpc.example.com/v2/SECRET_API_KEY"))
```

common/geth/rpc_utils_test.go
litt3: b03cc0777d0475c4bae847f7b5e4bcb4ca214316

pschork (approved):

PR #2330, opened 2025-11-17
feat: add permissionless disperser registry

0xClandestine:
## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/contracts/src/core/EigenDADisperserRegistryStorageV2.sol
+++ b/contracts/src/core/EigenDADisperserRegistryStorageV2.sol
@@ -2,12 +2,10 @@
 pragma solidity ^0.8.9;
 
 import {EigenDATypesV2} from "src/core/libraries/v2/EigenDATypesV2.sol";
-import {
-    EnumerableSetUpgradeable
-} from "lib/openzeppelin-contracts-upgradeable/contracts/utils/structs/EnumerableSetUpgradeable.sol";
+import {EnumerableSet} from "lib/openzeppelin-contracts/contracts/utils/structs/EnumerableSet.sol";
 
 abstract contract EigenDADisperserRegistryStorageV2 {
-    using EnumerableSetUpgradeable for EnumerableSetUpgradeable.UintSet;
+    using EnumerableSet for EnumerableSet.UintSet;
 
     /// -----------------------------------------------------------------------
     /// Constants
@@ -30,27 +28,34 @@ abstract contract EigenDADisperserRegistryStorageV2 {
     /// Mutable Storage
     /// -----------------------------------------------------------------------
 
-    /// @notice Returns the total number of registrations.
-    uint32 public totalRegistrations;
+    /// @dev Returns the ERC-7201 namespace for this contract.
+    ///      https://eips.ethereum.org/EIPS/eip-7201
+    bytes32 private constant DISPERSER_REGISTRY_STORAGE_LOCATION =
+        keccak256(abi.encode(uint256(keccak256("eigenda.disperser.registry.storage")) - 1)) & ~bytes32(uint256(0xff));
 
-    /// @notice Returns the nonce for a given disperser address.
-    mapping(address disperser => uint256 nonce) public nonces;
+    /// @dev Struct containing the storage layout for this contract (without dependencies).
+    struct Layout {
+        /// @dev Returns the total number of registrations.
+        uint32 totalRegistrations;
+        /// @dev Returns the nonce for a given disperser address.
+        mapping(address disperser => uint256 nonce) nonces;
+        /// @dev Set of disperser IDs for default dispersers.
+        /// Validators should default to accepting dispersals from dispersers in this set.
+        EnumerableSet.UintSet defaultDispersers;
+        /// @dev Set of disperser IDs for on-demand dispersers.
+        /// Dispersers in this set are authorized to use on-demand (pay-per-use) payments.
+        EnumerableSet.UintSet onDemandDispersers;
+        /// @dev Mapping from disperser ID to disperser info.
+        mapping(uint32 disperserId => EigenDATypesV2.DisperserInfoV2 disperserInfo) disperserInfo;
+    }
 
-    /// @dev Set of disperser IDs for default dispersers.
-    /// Validators should default to accepting dispersals from dispersers in this set.
-    EnumerableSetUpgradeable.UintSet internal _defaultDispersers;
-
-    /// @dev Set of disperser IDs for on-demand dispersers.
-    /// Dispersers in this set are authorized to use on-demand (pay-per-use) payments.
-    EnumerableSetUpgradeable.UintSet internal _onDemandDispersers;
-
-    /// @dev Mapping from disperser ID to disperser info.
-    mapping(uint32 disperserId => EigenDATypesV2.DisperserInfoV2 disperserInfo) internal _disperserInfo;
-
-    /// -----------------------------------------------------------------------
-    /// Storage Gap
-    /// -----------------------------------------------------------------------
-
-    // slither-disable-next-line shadowing-state
-    uint256[41] private __GAP;
+    /// @dev Returns the storage layout for this contract.
+    /// Usage: `Layout storage $ = getDisperserRegistryStorage();`.
+    function getDisperserRegistryStorage() internal pure returns (Layout storage $) {
+        bytes32 ptr = DISPERSER_REGISTRY_STORAGE_LOCATION;
+        /// @solidity memory-safe-assembly
+        assembly {
+            $.slot := ptr
+        }
+    }
 }
--- a/contracts/src/core/EigenDADisperserRegistryV2.sol
+++ b/contracts/src/core/EigenDADisperserRegistryV2.sol
@@ -1,10 +1,10 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.9;
 
+// TODO: We MUST use OZ v5.5.0 in order to use EIP712Upgradeable.
+
 import {OwnableUpgradeable} from "lib/openzeppelin-contracts-upgradeable/contracts/access/OwnableUpgradeable.sol";
-import {
-    EnumerableSetUpgradeable
-} from "lib/openzeppelin-contracts-upgradeable/contracts/utils/structs/EnumerableSetUpgradeable.sol";
+import {EnumerableSet} from "lib/openzeppelin-contracts/contracts/utils/structs/EnumerableSet.sol";
 import {
     EIP712Upgradeable
 } from "lib/openzeppelin-contracts-upgradeable/contracts/utils/cryptography/draft-EIP712Upgradeable.sol";
@@ -29,7 +29,7 @@ contract EigenDADisperserRegistryV2 is
     IEigenDADisperserRegistryV2
 {
     using SignatureCheckerUpgradeable for address;
-    using EnumerableSetUpgradeable for *;
+    using EnumerableSet for *;
 
     /// -----------------------------------------------------------------------
     /// Initialization
@@ -55,21 +55,25 @@ contract EigenDADisperserRegistryV2 is
         virtual
         returns (uint32 disperserId)
     {
-        bytes32 digest = keccak256(abi.encode(REGISTRATION_TYPEHASH, disperser, relayURL, nonces[disperser]++));
+        // Create a reference to the contract's storage.
+        Layout storage $ = getDisperserRegistryStorage();
+        // Create a reference to the disperser info.
+        EigenDATypesV2.DisperserInfoV2 storage disperserInfo = $.disperserInfo[disperserId];
+
+        // Increment and assign the next available disperserId, starting at 1 (not 0).
+        disperserId = $.totalRegistrations + 1; // Monotonically increasing.
+
+        bytes32 digest = keccak256(abi.encode(REGISTRATION_TYPEHASH, disperser, relayURL, $.nonces[disperser]++));
 
         // Assert that the signature is valid (supports EIP-1271).
         _checkSignature(disperser, digest, signature);
         // Assert that the disperser address is non-zero.
         if (disperser == address(0)) revert InputAddressZero();
-
-        // Increment and assign the next available disperserId, starting at 1 (not 0).
-        disperserId = ++totalRegistrations; // Monotonically increasing.
-
-        EigenDATypesV2.DisperserInfoV2 storage disperserInfo = _disperserInfo[disperserId];
-
         // Assert that the disperser is not already registered.
         if (disperserInfo.disperser != address(0)) revert DisperserIsRegistered();
 
+        // Increment total registrations.
+        $.totalRegistrations = disperserId;
         // Set the disperser info.
         disperserInfo.disperser = disperser;
         disperserInfo.relayURL = relayURL;
@@ -79,19 +83,27 @@ contract EigenDADisperserRegistryV2 is
 
     /// @inheritdoc IEigenDADisperserRegistryV2
     function deregisterDisperser(uint32 disperserId, bytes memory signature) external virtual {
-        EigenDATypesV2.DisperserInfoV2 storage disperserInfo = _disperserInfo[disperserId];
+        // Create a reference to the contract's storage.
+        Layout storage $ = getDisperserRegistryStorage();
+        // Create a reference to the disperser info.
+        EigenDATypesV2.DisperserInfoV2 storage disperserInfo = $.disperserInfo[disperserId];
 
-        bytes32 digest = keccak256(abi.encode(DEREGISTRATION_TYPEHASH, disperserId, nonces[disperserInfo.disperser]++));
+        bytes32 digest =
+            keccak256(abi.encode(DEREGISTRATION_TYPEHASH, disperserId, $.nonces[disperserInfo.disperser]++));
+
+        // TODO: Instead we should simply remove them from the sets.
 
         // Assert that the disperser is not in the default or on-demand dispersers sets.
         // This means owner can only deregister dispersers after revoking their default or on-demand status.
-        if (_defaultDispersers.contains(disperserId) || _onDemandDispersers.contains(disperserId)) {
+        if ($.defaultDispersers.contains(disperserId) || $.onDemandDispersers.contains(disperserId)) {
             revert DisperserInSet();
         }
 
         _checkSignature(disperserInfo.disperser, digest, signature);
         // Assert that the disperser is registered.
         if (disperserInfo.disperser == address(0)) revert DisperserIsNotRegistered();
+
+        // Total registrations are monotonically increasing, no need to decrement.
 
         // Delete the disperser info.
         delete disperserInfo.disperser;
@@ -102,10 +114,13 @@ contract EigenDADisperserRegistryV2 is
 
     /// @inheritdoc IEigenDADisperserRegistryV2
     function updateRelayURL(uint32 disperserId, string memory relayURL, bytes memory signature) external {
-        EigenDATypesV2.DisperserInfoV2 storage disperserInfo = _disperserInfo[disperserId];
+        Layout storage $ = getDisperserRegistryStorage();
 
-        bytes32 digest =
-            keccak256(abi.encode(UPDATE_RELAY_URL_TYPEHASH, disperserId, relayURL, nonces[disperserInfo.disperser]++));
+        EigenDATypesV2.DisperserInfoV2 storage disperserInfo = $.disperserInfo[disperserId];
+
+        bytes32 digest = keccak256(
+            abi.encode(UPDATE_RELAY_URL_TYPEHASH, disperserId, relayURL, $.nonces[disperserInfo.disperser]++)
+        );
 
         // Assert that the signature is valid (supports EIP-1271).
         _checkSignature(disperserInfo.disperser, digest, signature);
@@ -120,7 +135,10 @@ contract EigenDADisperserRegistryV2 is
 
     /// @inheritdoc IEigenDADisperserRegistryV2
     function revokeNonce() external {
-        ++nonces[msg.sender];
+        Layout storage $ = getDisperserRegistryStorage();
+        unchecked {
+            ++$.nonces[msg.sender];
+        }
     }
 
     /// @dev Verifies that the signature is valid if caller is not the disperser.
@@ -140,27 +158,31 @@ contract EigenDADisperserRegistryV2 is
 
     /// @inheritdoc IEigenDADisperserRegistryV2
     function addDefaultDisperser(uint32 disperserId) external onlyOwner {
-        if (_disperserInfo[disperserId].disperser == address(0)) revert DisperserIsNotRegistered();
-        if (!_defaultDispersers.add(disperserId)) revert DisperserInSet();
+        Layout storage $ = getDisperserRegistryStorage();
+        if ($.disperserInfo[disperserId].disperser == address(0)) revert DisperserIsNotRegistered();
+        if (!$.defaultDispersers.add(disperserId)) revert DisperserInSet();
         emit DefaultDisperserAdded(disperserId);
     }
 
     /// @inheritdoc IEigenDADisperserRegistryV2
     function addOnDemandDisperser(uint32 disperserId) external onlyOwner {
-        if (_disperserInfo[disperserId].disperser == address(0)) revert DisperserIsNotRegistered();
-        if (!_onDemandDispersers.add(disperserId)) revert DisperserInSet();
+        Layout storage $ = getDisperserRegistryStorage();
+        if ($.disperserInfo[disperserId].disperser == address(0)) revert DisperserIsNotRegistered();
+        if (!$.onDemandDispersers.add(disperserId)) revert DisperserInSet();
         emit OnDemandDisperserAdded(disperserId);
     }
 
     /// @inheritdoc IEigenDADisperserRegistryV2
     function removeDefaultDisperser(uint32 disperserId) external onlyOwner {
-        if (!_defaultDispersers.remove(disperserId)) revert DisperserNotInSet();
+        Layout storage $ = getDisperserRegistryStorage();
+        if (!$.defaultDispersers.remove(disperserId)) revert DisperserNotInSet();
         emit DefaultDisperserRemoved(disperserId);
     }
 
     /// @inheritdoc IEigenDADisperserRegistryV2
     function removeOnDemandDisperser(uint32 disperserId) external onlyOwner {
-        if (!_onDemandDispersers.remove(disperserId)) revert DisperserNotInSet();
+        Layout storage $ = getDisperserRegistryStorage();
+        if (!$.onDemandDispersers.remove(disperserId)) revert DisperserNotInSet();
         emit OnDemandDisperserRemoved(disperserId);
     }
 
@@ -168,23 +190,37 @@ contract EigenDADisperserRegistryV2 is
     /// View Logic
     /// -----------------------------------------------------------------------
 
+    // /// @inheritdoc IEigenDADisperserRegistryV2
+    // function getTotalRegistrations() external view returns (uint32) {
+    //     Layout storage $ = getDisperserRegistryStorage();
+    //     return $.totalRegistrations;
+    // }
+
+    // /// @inheritdoc IEigenDADisperserRegistryV2
+    // function getNonce(address disperser) external view returns (uint256) {
+    //     Layout storage $ = getDisperserRegistryStorage();
+    //     return $.nonces[disperser];
+    // }
+
     /// @inheritdoc IEigenDADisperserRegistryV2
     function getDisperserInfo(uint32[] memory ids)
         external
         view
         returns (EigenDATypesV2.DisperserInfoV2[] memory info)
     {
+        Layout storage $ = getDisperserRegistryStorage();
         uint256 len = ids.length;
         info = new EigenDATypesV2.DisperserInfoV2[](len);
         for (uint256 i = 0; i < len; ++i) {
-            info[i] = _disperserInfo[ids[i]];
+            info[i] = $.disperserInfo[ids[i]];
         }
         return info;
     }
 
     /// @inheritdoc IEigenDADisperserRegistryV2
     function getDefaultDisperserIds() external view returns (uint32[] memory ids) {
-        uint256[] memory keys = _defaultDispersers.values();
+        Layout storage $ = getDisperserRegistryStorage();
+        uint256[] memory keys = $.defaultDispersers.values();
         /// @solidity memory-safe-assembly
         assembly {
             ids := keys
@@ -194,7 +230,8 @@ contract EigenDADisperserRegistryV2 is
 
     /// @inheritdoc IEigenDADisperserRegistryV2
     function getOnDemandDisperserIds() external view returns (uint32[] memory ids) {
-        uint256[] memory keys = _onDemandDispersers.values();
+        Layout storage $ = getDisperserRegistryStorage();
+        uint256[] memory keys = $.onDemandDispersers.values();
         /// @solidity memory-safe-assembly
         assembly {
             ids := keys
@@ -204,11 +241,13 @@ contract EigenDADisperserRegistryV2 is
 
     /// @inheritdoc IEigenDADisperserRegistryV2
     function isDefaultDisperserId(uint32 disperserId) external view returns (bool) {
-        return _defaultDispersers.contains(disperserId);
+        Layout storage $ = getDisperserRegistryStorage();
+        return $.defaultDispersers.contains(disperserId);
     }
 
     /// @inheritdoc IEigenDADisperserRegistryV2
     function isOnDemandDisperserId(uint32 disperserId) external view returns (bool) {
-        return _onDemandDispersers.contains(disperserId);
+        Layout storage $ = getDisperserRegistryStorage();
+        return $.onDemandDispersers.contains(disperserId);
     }
 }
--- a/.gitmodules
+++ b/.gitmodules
@@ -11,3 +11,6 @@
 	path = contracts/lib/eigenlayer-middleware
 	url = https://github.com/Layr-Labs/eigenlayer-middleware
 	branch = m2-mainnet-fixes
+[submodule "contracts/lib/openzeppelin-contracts-upgradeable-v5.5.0"]
+	path = contracts/lib/openzeppelin-contracts-upgradeable-v5.5.0
+	url = https://github.com/openzeppelin/openzeppelin-contracts-upgradeable


contracts/src/core/EigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,189 @@
...
+        disperserInfo.relayURL = relayURL;
+
+        emit DisperserRegistered(disperserId, disperser);
+    }
+
+    /// @inheritdoc IEigenDADisperserRegistryV2
+    function deregisterDisperser(uint32 disperserId, bytes memory signature) external virtual {
+        EigenDATypesV2.DisperserInfoV2 storage disperserInfo = _disperserInfo[disperserId];
+
+        bytes32 digest = keccak256(abi.encode(DEREGISTRATION_TYPEHASH, disperserId));
```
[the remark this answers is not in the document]
0xClandestine: 9321f40

contracts/src/core/EigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,205 @@
...
+        if (msg.sender != disperser) {
+            if (!disperser.isValidSignatureNow(digest, signature)) revert InvalidSignature();
+        }
+    }
+
+    /// -----------------------------------------------------------------------
+    /// Owner-only Logic
+    /// -----------------------------------------------------------------------
+
+    // TODO: Extra checks?
```
0xClandestine: Reviewers note any additional checks we might want here.

contracts/src/core/libraries/v2/EigenDATypesV2.sol
```diff
@@ -13,6 +13,11 @@ library EigenDATypesV2 {
         address disperserAddress;
     }
 
+    struct DisperserInfoV2 {
```
0xClandestine, line 16: Don't think it's safe for us to modify the old struct.

contracts/src/core/EigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    /// @inheritdoc IEigenDADisperserRegistryV2
+    function getDefaultDisperserIds() external view returns (uint32[] memory ids) {
+        uint256[] memory keys = _defaultDispersers.values();
+        /// @solidity memory-safe-assembly
+        assembly {
+            ids := keys
+        }
+        return ids;
```
0xClandestine: Assembly is used to typecast uint256[] -> uint32[] array.

contracts/src/core/libraries/v2/EigenDATypesV2.sol
```diff
@@ -13,6 +13,11 @@ library EigenDATypesV2 {
...
+        address disperser;
```
litt3, line 17: Any specific reason you went with `disperser` instead of `disperserAddress`?

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,145 @@
+// SPDX-License-Identifier: MIT
+pragma solidity ^0.8.9;
+
+import {EigenDATypesV2} from "src/core/libraries/v2/EigenDATypesV2.sol";
+
+/// @title IEigenDADisperserRegistryV2
+/// @author Layr Labs, Inc.
+/// @notice This interface defines the contract for managing EigenDA disperser registration and configuration.
+/// @dev This registry manages disperser registration, deregistration, and relay URL updates.
+interface IEigenDADisperserRegistryV2 {
```
litt3, line 17:
```suggestion
interface IDisperserRegistryV2 {
```

As we discussed, personally I'd be in favor of removing clutter like this from our naming. Curious other peoples' opinions

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,145 @@
...
+    /// -----------------------------------------------------------------------
+    /// Events
+    /// -----------------------------------------------------------------------
+    /// @notice This event is emitted when a new disperser is successfully registered with the registry.
+    /// @param disperserId The unique ID that has been assigned to the newly registered disperser.
+    /// @param disperser The address of the disperser that has been registered.
+    event DisperserRegistered(uint32 indexed disperserId, address indexed disperser);
```
litt3, line 24: Would it make more sense to emit the entire `DisperserInfo`, instead of just the address?

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,145 @@
...
+    error AlreadyRegistered();
+
+    /// @notice This error is thrown when an invalid signature is provided for verification.
+    error InvalidSignature();
+
+    /// @notice This error is thrown when attempting an operation on a disperser that is not registered.
+    error NotRegistered();
+
+    /// @notice This error is thrown when attempting to add a disperser that already exists in the specified set.
+    error DisperserAlreadyExists();
```
litt3:
```suggestion
    error DisperserAlreadyInSet();
```

What do you think about this name instead?

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,145 @@
...
+    /// @param signature The signature from the disperser that authorizes this deregistration.
+    function deregisterDisperser(uint32 disperserId, bytes memory signature) external;
+
+    /// @notice This function updates the relay URL for a disperser using a valid signature.
+    /// @dev The `signature` parameter can be empty if `msg.sender` is the registered disperser.
+    /// Reverts if the signature is invalid and the caller is not the disperser.
+    /// @param disperserId The ID of the disperser whose relay URL should be updated.
+    /// @param relayURL The new relay URL that should be set for this disperser.
+    /// @param signature The signature from the disperser that authorizes this update.
+    function updateRelayURL(uint32 disperserId, string memory relayURL, bytes memory signature) external;
```
litt3, line 100: @pschork has brought up potentially supporting disperser key rotation.

@0xClandestine Would it be possible to add a similar function for modifying disperser key, if we decide as a group that we want to support it? If we decide we don't want to add this now, how hard would it be to add in the future?

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,145 @@
...
+
+    /// @notice This function revokes the nonce for the caller.
+    /// @dev This function is used to revoke the caller's current nonce.
+    function revokeNonce() external;
+
+    /// -----------------------------------------------------------------------
+    /// View Functions
```
litt3: These don't appear to be view functions

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,145 @@
...
+    /// -----------------------------------------------------------------------
+
+    /// @notice This function adds a disperser to the default dispersers set (only callable by owner).
+    /// @param disperserId The ID of the disperser that should be added to the default set.
+    function addDefaultDisperser(uint32 disperserId) external;
```
litt3, line 113: Now is probably the time to come to consensus on naming

@pschork I know you weren't a fan of "DefaultDispersers". What was the alternative name you suggested?

Are there any other names people would prefer here?

contracts/src/core/EigenDADisperserRegistryStorageV2.sol
```diff
@@ -0,0 +1,47 @@
...
+    /// -----------------------------------------------------------------------
+    /// Mutable Storage
+    /// -----------------------------------------------------------------------
+
+    /// @notice Returns the total number of registrations.
+    uint32 public totalRegistrations;
+    /// @notice Returns the nonce for a given disperser address.
+    mapping(address disperser => uint256 nonce) public nonces;
+
+    /// @dev Mapping from disperser ID to disperser address for default dispersers.
```
litt3: stale doc

contracts/src/core/EigenDADisperserRegistryStorageV2.sol
```diff
@@ -0,0 +1,47 @@
...
+    EnumerableSetUpgradeable.UintSet internal _defaultDispersers;
+    /// @dev Mapping from disperser ID to disperser address for on-demand dispersers.
```
litt3: stale doc

contracts/src/core/EigenDADisperserRegistryStorageV2.sol
```diff
@@ -0,0 +1,47 @@
...
+
+abstract contract EigenDADisperserRegistryStorageV2 {
+    using EnumerableSetUpgradeable for EnumerableSetUpgradeable.UintSet;
+
+    /// -----------------------------------------------------------------------
+    /// Constants
+    /// -----------------------------------------------------------------------
+
+    /// @notice The EIP-712 typehash signed by a disperser that signals intent to deregister.
+    /// @dev Owner has the ability to censor deregistrations by not processing the signature.
```
litt3, line 19: I'm confused by this comment about censorship. Is this doc up-to-date?

contracts/src/core/EigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,205 @@
...
+        __EIP712_init("EigenDADisperserRegistry", "2");
+        _transferOwnership(_initialOwner);
+    }
+
+    /// -----------------------------------------------------------------------
+    /// External Logic
+    /// -----------------------------------------------------------------------
+
+    /// @inheritdoc IEigenDADisperserRegistryV2
+    function registerDisperser(address disperser, string memory relayURL)
```
litt3: There is an interesting mismatch here, in that the disperser itself must call `registerDisperser`, but anyone with a valid signature can call `deregisterDisperser`.

Would it make sense to add a signature parameter to `registerDisperser` as well, so that the same indirect approach can be used during the registration phase, as is used in the deregistration phase?

contracts/src/core/EigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,205 @@
...
+    /// @inheritdoc IEigenDADisperserRegistryV2
+    function deregisterDisperser(uint32 disperserId, bytes memory signature) external virtual {
+        EigenDATypesV2.DisperserInfoV2 storage disperserInfo = _disperserInfo[disperserId];
+
+        bytes32 digest = keccak256(abi.encode(DEREGISTRATION_TYPEHASH, disperserId, nonces[disperserInfo.disperser]++));
+
+        // Assert that the disperser is not in the default or on-demand dispersers sets.
+        // This means owner can only deregister dispersers after revoking their default or on-demand status.
+        if (_defaultDispersers.contains(disperserId) || _onDemandDispersers.contains(disperserId)) {
+            revert DisperserInSet();
```
litt3, line 99: I think might make more sense to just remove the disperser from any special set it is a member of, if it is deregistered.

Only eigenlabs has permission to _add_ to these sets, but if a member of these sets wants to leave the network, there's not really any way we can force them to stick around

contracts/src/core/EigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,205 @@
...
+        }
+
+        _checkSignature(disperserInfo.disperser, digest, signature);
+        // Assert that the disperser is registered.
+        if (disperserInfo.disperser == address(0)) revert NotRegistered();
```
litt3: Since it's impossible to create a valid signature from the zero address, I think this is technically unnecessary?

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
samlaf: so this registry subsumes the RelayRegistry?
image

Should we delete it and/or deprecate it?

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,145 @@
...
+    event DisperserRegistered(uint32 indexed disperserId, address indexed disperser);
+
+    /// @notice This event is emitted when a disperser is successfully deregistered from the registry.
+    /// @param disperserId The ID of the disperser that has been deregistered.
+    event DisperserDeregistered(uint32 indexed disperserId);
+
+    /// @notice This event is emitted when a disperser successfully updates their relay URL.
+    /// @param disperserId The ID of the disperser that updated their relay URL.
+    /// @param relayURL The new relay URL that has been set for the disperser.
+    event RelayURLUpdated(uint32 indexed disperserId, string relayURL);
```
samlaf, line 33: do we really need these? do we plan on using subgraph to index this? I thought we were trying to move away from subgraph

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
samlaf: might be good to mention what the default and on-demand register sets are here?

samlaf: unclear to me from looking at the events below

contracts/src/core/EigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,205 @@
...
+    }
+
+    /// @inheritdoc IEigenDADisperserRegistryV2
+    function revokeNonce() external {
+        ++nonces[msg.sender];
+    }
+
+    /// @dev Verifies that the signature is valid if caller is not the disperser.
+    /// @dev Reverts if the signature is invalid and the caller is not the disperser.
+    /// @param disperser The address of the disperser that is updating their relay URL.
```
litt3: param doc specifically mentions "disperser that is updating their relay URL", but this method is also used in other contexts

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
samlaf: how does this differ from AlreadyRegistered above?

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,145 @@
...
+    error InvalidSignature();
+
+    /// @notice This error is thrown when attempting an operation on a disperser that is not registered.
+    error NotRegistered();
+
+    /// @notice This error is thrown when attempting to add a disperser that already exists in the specified set.
+    error DisperserAlreadyExists();
+
+    /// @notice This error is thrown when attempting to remove a disperser that does not exist in the specified set.
+    error DisperserNotFound();
```
samlaf: what does "specified set" mean? Can we add that specified set as an argument to the error? or do cuwtom errors not take arguments?

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,145 @@
...
+
+    /// @notice This error is thrown when attempting to perform an operation on a disperser that is in the default or on-demand dispersers sets.
+    error DisperserInSet();
```
samlaf: are default and on-demand the only 2 disperser sets? you mention "specified set" above but here list these 2, so it sounds like there are more sets, but unclear

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,145 @@
...
+
+    /// -----------------------------------------------------------------------
+    /// External Functions
+    /// -----------------------------------------------------------------------
+
+    /// @notice This function registers a new disperser with the registry and assigns it a unique ID.
+    /// @param disperser The address of the disperser that should be registered.
+    /// @param relayURL The relay URL that will be associated with this disperser.
+    /// @return disperserId The unique ID that has been assigned to the newly registered disperser.
+    function registerDisperser(address disperser, string memory relayURL) external returns (uint32 disperserId);
```
samlaf: are we moving to single relayURL per disperser now? Up to now our disperser was associated to many relays which each had their own URL

samlaf: also why does registerDisperser not allow passing a signature like deregister and updateRelayURL?

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,145 @@
...
+    /// @param disperserId The ID of the disperser that should be added to the on-demand set.
+    function addOnDemandDisperser(uint32 disperserId) external;
+
+    /// @notice This function removes a disperser from the default dispersers set (only callable by owner).
+    /// @param disperserId The ID of the disperser that should be removed from the default set.
+    function removeDefaultDisperser(uint32 disperserId) external;
+
+    /// @notice This function removes a disperser from the on-demand dispersers set (only callable by owner).
+    /// @param disperserId The ID of the disperser that should be removed from the on-demand set.
+    function removeOnDemandDisperser(uint32 disperserId) external;
```
samlaf, line 128: these are not view functions

contracts/src/core/EigenDADisperserRegistryStorageV2.sol
samlaf, line 1: So Patrick was very against these Storage contracts with storage gaps and inheritance, and favored the other more composable/modular approach (see contracts under libraries/v3).

I feel like we should try to be consistent. Are you planning to move everything back to using this storage model? My understanding was that the other approach is better overall (even OZ is moving to it in its v5 contracts afaiu)

contracts/src/core/EigenDADisperserRegistryV2.sol
samlaf: how does this work...? its basically typecasting + copying over to the other array?

contracts/src/core/EigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,205 @@
...
+        disperserInfo.relayURL = relayURL;
+
+        emit DisperserRegistered(disperserId, disperser);
+    }
+
+    /// @inheritdoc IEigenDADisperserRegistryV2
+    function deregisterDisperser(uint32 disperserId, bytes memory signature) external virtual {
+        EigenDATypesV2.DisperserInfoV2 storage disperserInfo = _disperserInfo[disperserId];
+
+        bytes32 digest = keccak256(abi.encode(DEREGISTRATION_TYPEHASH, disperserId, nonces[disperserInfo.disperser]++));
```
litt3: Looks like EIP-712 calls for including the domain separator in this digest. Claude says you can use `_hashTypedDataV4` from `EIP712Upgradeable` to add it

Same comment applies the other places where you're doing sig verification

contracts/src/core/EigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,205 @@
...
+
+    /// -----------------------------------------------------------------------
+    /// Owner-only Logic
+    /// -----------------------------------------------------------------------
+
+    // TODO: Extra checks?
+
+    /// @inheritdoc IEigenDADisperserRegistryV2
+    function addDefaultDisperser(uint32 disperserId) external onlyOwner {
+        if (!_defaultDispersers.add(disperserId)) revert DisperserAlreadyExists();
```
litt3: We should probably only add to these sets after checking that the disperser is registered

contracts/src/core/EigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,205 @@
...
+    {
+        // Increment and assign the next available disperserId, starting at 1 (not 0).
+        disperserId = ++totalRegistrations; // Monotonic increasing.
+
+        EigenDATypesV2.DisperserInfoV2 storage disperserInfo = _disperserInfo[disperserId];
+
+        // Assert that the disperser address is non-zero.
+        if (disperser == address(0)) revert InputAddressZero();
+        // Assert that the disperser is not already registered.
+        if (disperserInfo.disperser != address(0)) revert AlreadyRegistered();
```
litt3: theoretically, this can only happen if there is a bug in the contract, right? Since we always increment `totalRegistrations` upon successful registration, it doesn't seem possible to ever arrive here with an ID that's already been registered

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
0xClandestine: Best practice is to always include an event for state modifying methods, additional cost in gas is well worth it here imo. Never know how other people will want to integrate.

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
0xClandestine: Still a bit unclear to me as well, will sync with @litt3 and clarify.

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
0xClandestine: Likely a duplicate/unused, need to lint.

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
0xClandestine: On demand vs default sets of dispersers. I'd prefer to avoid using parameters in errors, while they are supported though.

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
0xClandestine: Yes, clarified above.

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
0xClandestine: 1) I'm not sure, likely needs discussion with team.
2) Nice catch, will add.

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
0xClandestine: Oops, ty

contracts/src/core/EigenDADisperserRegistryStorageV2.sol
0xClandestine: I actually prefer namespaced storage, however to properly and safely implement it with OZ dependencies we need to upgrade OZ from 0.4.6 -> +0.5.x.

0xClandestine: Honestly would like to make a PR that properly converts this entire repo to namespacing, to be backwards compatible we would need to define constants for every compiler assigned storage slot.

0xClandestine: > how does this work...? its basically typecasting + copying over to the other array? @samlaf 

Can't comment below, it's a solidity workaround for typecasting arrays. No memory is copied/duplicated, just tricks solidity into thinking it was originally that type (compiler ignores truncated/dirty bits if any).

contracts/src/core/libraries/v2/EigenDATypesV2.sol
0xClandestine: Best practice naming conventions, if type is address we know it's an address.

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
0xClandestine: Totally agree we should remove `EigenDA` prefix from all of our contract names. Curious what others think though.

0xClandestine: cc @samlaf @ethenotethan @ian-shim

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
0xClandestine: Native types are always preferred for event parameters, however I will admit we're pretty loose with this in Protocol.

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
0xClandestine: Would need to use a different identifier than address for dispersers, could be easily implemented now though.

contracts/src/core/EigenDADisperserRegistryV2.sol
0xClandestine: Not without the private key to address zero no, one could in theory create signatures whose private keys are unknown. This was actually abused early on for permissionless cross-chain deterministic deployments (handy for proxy factories).

https://github.com/noot/eth-one-time-sig/blob/master/ots.go

contracts/src/core/EigenDADisperserRegistryV2.sol
0xClandestine: Can you elaborate and I'll update.

contracts/src/core/EigenDADisperserRegistryV2.sol
0xClandestine: Rip had it originally, think autocomplete removed it when wrapping in internal method.

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
0xClandestine: Need team input.

contracts/src/core/EigenDADisperserRegistryV2.sol
0xClandestine: Can a disperser be in both sets? Otherwise we need a check that uses .contains on the opposite set.

contracts/src/core/EigenDADisperserRegistryV2.sol
0xClandestine: If storage isn't already populated for this disperser, without this check one could overwrite their info by calling this twice. You can read from the mapping with any key without revert, even it nothing has ever been stored.

contracts/src/core/EigenDADisperserRegistryV2.sol
0xClandestine: See above.

contracts/src/core/EigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,205 @@
...
+    /// @inheritdoc IEigenDADisperserRegistryV2
+    function getDisperserInfo(uint32[] memory ids)
+        external
+        view
+        returns (EigenDATypesV2.DisperserInfoV2[] memory info)
+    {
+        uint256 len = ids.length;
+        info = new EigenDATypesV2.DisperserInfoV2[](len);
+        for (uint256 i = 0; i < len; i++) {
+            info[i] = _disperserInfo[ids[i]];
```
pschork: what is returned if a undefined/deregistered disperser id is specified?

contracts/src/core/EigenDADisperserRegistryV2.sol
```diff
@@ -0,0 +1,205 @@
...
+    /// @inheritdoc IEigenDADisperserRegistryV2
+    function getOnDemandDisperserIds() external view returns (uint32[] memory ids) {
+        uint256[] memory keys = _onDemandDispersers.values();
+        /// @solidity memory-safe-assembly
+        assembly {
+            ids := keys
+        }
+        return ids;
+    }
+
```
pschork, line 241: can we get a `getDisperserIds()` view that would return all registered disperser ids - including dispersers that are not members of default or ondemand sets?

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
pschork: > are we moving to single relayURL per disperser now?

yes. relay sharding is not needed. recent throughput achievement was accomplished with single relay.

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
pschork: these improve readability on tenderly/etherscan/etc

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
pschork: I'm good with "DefaultDispersers". I think it is the best neutral option.

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
pschork: supporting seamless disperser address rotation requires a time period where both old and new addresses are simultaneously valid.
i think we can achieve zero downtime rotation using the current implementation with the trade off that rotation will change the disperser id.
1. register new disperser address + add to default/ondemand set
2. give operators time to sync (min-hrs)
3. restart controller with new disperser id + updated eoa/kms configured
4. deregister old disperser id

alternatively if we allow disperser address to be updated - you will have guaranteed downtime during period where validators sync and new controller is deployed. i don't think the effort/complexity of supporting multiple addresses in disperser info is worth it since rotation should be rare event. 

as long as we don't get tied to eigenda == disperser 0 we should be good.

contracts/src/core/EigenDADisperserRegistryV2.sol
litt3: > The address of the disperser that is updating their relay URL.

This doc suggests that the function is only for updating relay URL. But `_checkSignature` is also used for other purposes, so the relayURL-update-specific language is misleading

contracts/src/core/EigenDADisperserRegistryV2.sol
litt3: Yes, a disperser can (and will) be in both sets.

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
litt3: Makes sense- so no changes needed.

key rotation == register a new disperser with a different key

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
samlaf: cc @litt3

contracts/src/core/EigenDADisperserRegistryStorageV2.sol
samlaf: Why do you say the memory is not copied? It's moving from keys to ids, which are 2 separate memory regions iiuc.
image

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
samlaf: How would people on etherscan find our contracts though?

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
samlaf: how are clients supposed to discover disperser rotations though? Would they have to index the contract? Or would our disperser play nice and tell them "hey I'm about to rotate so you better use this new URL now"?

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
litt3: > so this registry subsumes the RelayRegistry?

Correct.

TLDR is that we've moved away from the original idea that a single disperser needs to have multiple relays. We, for example, are able to handle 1gb of throughput using only a single disperser that has a single relay address.

litt3: Descriptions for these are included in the [linear ticket](https://linear.app/eigenlabs/issue/EGDA-2236/modify-disperserregistry-to-support-permissionless-dispersal) I wrote up for this task. I think the description there is sufficient, but let me know if anything is still unclear

litt3: Are you implying that people on etherscan search for contracts by keyword?

Seems the proper way is _always_ to use the `ContractRegistry`. And to find the `ContractRegistry`, you just need to acquire the address from a trusted source

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
litt3: Clients will periodically call `getDefaultDisperserIds` to find out which dispersers are available by default. For non-default dispersers, there would need to be some sort of offchain communication.

But you bring up a good point concerning _URL_, specifically. I'll follow up in slack

contracts/src/core/EigenDADisperserRegistryStorageV2.sol
0xClandestine: Assembly handles this differently, `ids := keys` simply tells the compiler that `ids` reference is the same as `keys`. Using either value would return the same array in memory. We do this a lot on protocol if it helps.

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
0xClandestine: They would use the directory, also we really need to have an organization etherscan account to label all EigenLayer contracts.

0xClandestine: Oh I'll also be working on a proposal to integrate Etherscan across the organization.

0xClandestine: So good to remove? @litt3

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
0xClandestine: d3d966b

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
0xClandestine: [7a87397](https://github.com/Layr-Labs/eigenda/pull/2330/commits/7a873974550f753fac4089ebb11e124203d689fc)

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
0xClandestine: [d3d966b](https://github.com/Layr-Labs/eigenda/pull/2330/commits/d3d966bae1bf49ae03247fb56371a76d66a4cbff)

contracts/src/core/EigenDADisperserRegistryV2.sol
0xClandestine: 6316aad635d4617489e21ce49a97239dd7406cb6

contracts/src/core/EigenDADisperserRegistryV2.sol
0xClandestine: Null data, can optionally revert on unregistered dispersers if wanted.

contracts/src/core/EigenDADisperserRegistryV2.sol
0xClandestine: [e2be8df](https://github.com/Layr-Labs/eigenda/pull/2330/commits/e2be8df61e34860e495d341ea896af5b31528c65)

contracts/src/core/EigenDADisperserRegistryV2.sol
0xClandestine: [9e0eff0](https://github.com/Layr-Labs/eigenda/pull/2330/commits/9e0eff0f06029f79245c1e1e286106a35eefdc8e)

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
0xClandestine: [3f77c53](https://github.com/Layr-Labs/eigenda/pull/2330/commits/3f77c5364e1a5c25632ee957eaa524a2c5c6c587)

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
0xClandestine: f3291c6

contracts/src/core/EigenDADisperserRegistryV2.sol
0xClandestine: How needed is this? Would likely require refactoring our storage.

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
0xClandestine: Any changes needed here? @litt3

litt3: Yes, we need to add a `DisperserUrl` field (string) to the `DisperserInfo` struct

It should have all the same access patterns as the `RelayUrl` field

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
litt3: I'm not sure if deleting it yet is the way to go- this PR creates the new contract, but migration timelines still aren't set in stone.

IMO it would make sense to leave the `RelayRegistry` untouched for now, and we can delete it later. But I don't feel super strongly, and could be persuaded otherwise

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
pschork: hmm, was only considering validator wrt rotation. in my original proposal it assumes that the disperser url does not change - controller config gets updated and redeployed with new disperser id. also validator sync is not a problem since the validator will fetch the disperser info if it does not find disperser id in cache.

if we want to register disperser urls, i think we should allow them to specify the same disperser url for new disperser. if we don't allow this it complicates the transition as we will need to have 2 controllers (with distinct ingress mapped to different urls) running at the same time. doable but its add a bunch of complexity and requires new tooling we don't have in order to make it work.

contracts/src/core/EigenDADisperserRegistryV2.sol
pschork: null data is probably better than revert from validator perspective.

contracts/src/core/EigenDADisperserRegistryV2.sol
pschork: dropped

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
litt3: >  i think we should allow them to specify the same disperser url for new disperser.

Agreed, IMO the disperser URL should just be a string. We shouldn't try to make it unique or anything

In fact, I don't think there should be anything that prevents a disperser from re-registering with the same address, either

litt3: > the validator will fetch the disperser info if it does not find disperser id in cache.

I don't think we can do that. The problem is that an attacker could make up a bogus disperser ID, and force validators to make an RPC call each time to "check" if that disperser ID exists.

I think it's actually fine if there is a delay between registering a disperser, and validators finding out about the disperser during a periodic fetch. It just means if you are registering a disperser, or doing a rotation of some kind, you will have some amount of downtime, where not everyone will know about the disperser yet

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
0xClandestine: Will leave as-is for now.

contracts/src/core/EigenDADisperserRegistryStorageV2.sol
samlaf: It doesn't make sense to me though. a uint256[] is clearly different from a uint32[]... so you can't return the same memory otherwise things will be completely broken.

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
samlaf: fine with making the change.

contracts/src/core/EigenDADisperserRegistryStorageV2.sol
0xClandestine: I know intuitively it doesn't make sense, just trust me here.

contracts/src/core/interfaces/IEigenDADisperserRegistryV2.sol
litt3: If we don't want to emit the whole `DisperserInfo`, I think we probably just shouldn't emit anything at all beyond the `disperserId`.

The `address` alone isn't enough to interact with that disperser, so emitting it doesn't really help much

PR #2331, opened 2025-11-17
test(inabox): Add proxy harness to inabox

iquidus:
* Adds proxy harness to inabox

Part of the migration of proxy e2e tests to inabox

### DIFF
--- a/inabox/templates/testconfig-anvil.yaml
+++ b/inabox/templates/testconfig-anvil.yaml
@@ -36,6 +36,16 @@ services:
     operators: 4
     maxOperatorCount: 3
     relays: 4
+  proxy:
+    enabled: true
+    rest_port: 3100
+    arb_port: 3101
+    max_blob_size: "1mib"
+    enabled_apis:
+      standard_commitment: true
+      op_generic_commitment: true
+      op_keccak_commitment: true
+    enable_arb_custom_da: true
   stakes:
     - total: 100e18
       distribution: [1, 4, 6, 10]
--- a/inabox/tests/setup_infra.go
+++ b/inabox/tests/setup_infra.go
@@ -34,6 +34,10 @@ type InfrastructureConfig struct {
 	// once legacy payments are removed. When true, the controller, API server, and validator nodes will all
 	// use the new payment system.
 	UseNewPayments bool
+
+	// DisableProxy disables the proxy deployment when set to true. This is useful for
+	// tests that do not require the proxy infrastructure to be deployed.
+	DisableProxy bool
 }
 
 // SetupInfrastructure creates the shared infrastructure that persists across all tests.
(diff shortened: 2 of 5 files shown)


inabox/tests/setup_infra.go
```diff
@@ -34,6 +34,10 @@ type InfrastructureConfig struct {
 	// once legacy payments are removed. When true, the controller, API server, and validator nodes will all
 	// use the new payment system.
 	UseNewPayments bool
+
+	// DisableProxy disables the proxy deployment when set to true. This is useful for
+	// tests that do not require the proxy infrastructure to be deployed.
+	DisableProxy bool
```
samlaf, line 40: mention that proxy will be auto disabled when DisableDisperser=true

inabox/tests/setup_proxy_harness.go
```diff
@@ -0,0 +1,352 @@
...
+	// Server addresses
+	RestAddress string
+	ArbAddress  string
+
+	// Metrics
+	Metrics *proxy_metrics.EmulatedMetricer
+
+	// Configuration used
+	Config ProxyHarnessConfig
+}
```
samlaf, line 80: havent thought about this for very long but do we really need this harness? Proxy is already a struct. This whole file feels like a lot of duplicate code from our proxy builder. Am I missing something?

inabox/templates/testconfig-anvil.yaml
```diff
@@ -36,6 +36,16 @@ services:
...
+  proxy:
+    enabled: true
+    rest_port: 3100
+    arb_port: 3101
+    max_blob_size: "1mib"
+    enabled_apis:
+      standard_commitment: true
+      op_generic_commitment: true
+      op_keccak_commitment: true
+    enable_arb_custom_da: true
```
samlaf, line 48: @dmanc wdyt? does this fit with what you had in mind? other services are started with ENV_VARS below. should we also use that, or...? whats the long-term plan?

samlaf (commented):
quick first pass

inabox/templates/testconfig-anvil.yaml
dmanc: This seems like how I envisioned configuring each component. Long-term plan is to deprecate the env var based deployment below.

inabox/tests/setup_infra.go
```diff
@@ -153,6 +157,35 @@ func SetupInfrastructure(ctx context.Context, config *InfrastructureConfig) (*In
 	}
 	infra.OperatorHarness = *operatorHarness
 
+	// Setup Proxy Harness fourth (requires chain and disperser to be ready)
+	if !config.DisableProxy && !config.DisableDisperser {
+		// Get contract addresses from test config
+		proxyHarnessConfig := ProxyHarnessConfig{
+			DisperserHostname:     "localhost",
+			DisperserPort:         "32005", // V2 API server port from testconfig
```
dmanc, line 165: I think we should be able to get the hostname and port from the disperser harness struct so it's not hardcoded.

inabox/tests/setup_infra.go
```diff
@@ -153,6 +157,35 @@ func SetupInfrastructure(ctx context.Context, config *InfrastructureConfig) (*In
...
+			EthRPCURL:             "http://localhost:8545",
```
dmanc, line 166: Same comment for the RPC url

inabox/tests/setup_infra.go
```diff
@@ -153,6 +157,35 @@ func SetupInfrastructure(ctx context.Context, config *InfrastructureConfig) (*In
...
+			S3BucketName:          "test-eigenda-blobstore", // Same bucket used by disperser
```
dmanc, line 167: Same comment for the bucket, we should expose it in some object and pass it in here to avoid hardcoding the name

inabox/tests/setup_proxy_harness.go
```diff
@@ -0,0 +1,352 @@
...
+	}
+
+	if config.MaxBlobSizeBytes == 0 {
+		config.MaxBlobSizeBytes = 1 * 1024 * 1024 // 1 MiB default
+	}
+
+	if len(config.BackendsToEnable) == 0 {
+		// Enable both V1 and V2 by default
+		config.BackendsToEnable = []common.EigenDABackend{
+			common.V1EigenDABackend,
```
dmanc, line 107: Do we still care about testing V1 related stuff?

inabox/templates/testconfig-anvil.yaml
ethenotethan, line 48: why is `enable_arb_custom_da` a separate depth from the other nested `enabled_apis` boolean fields?

inabox/templates/testconfig-anvil.yaml
```diff
@@ -36,6 +36,16 @@ services:
     operators: 4
     maxOperatorCount: 3
     relays: 4
+  proxy:
+    enabled: true
+    rest_port: 3100
+    arb_port: 3101
+    max_blob_size: "1mib"
```
ethenotethan, line 43: why are we defaulting to `1mib` and not aligning to our protocol limit of `16mib`?

iquidus:
Replaced by https://github.com/Layr-Labs/eigenda/pull/2336

PR #2332, opened 2025-11-17
fix(rust): decode_payload checks that padding is 0s

samlaf:
Spec was recently changed to require padding to be 0s. This aligns our rust implementation with the golang implementation and spec.

Implementation this made me really question our choice of enforcing this. What do we get from enforcing this other than more expensive (especially when proven in zkvm), brittle, and harder to read code? Unclear to me at this point..

EDIT: as suggested by bowen below, I updated all of the decode_payload code to use the [hokulea](https://github.com/Layr-Labs/hokulea/blob/ca100c76285bce38f91dfb8a391c32af8d1fc741/crates/eigenda/src/eigenda_data.rs#L79) implementation which was recently audited. Copy pasted the code for now given that I don't want to be importing hokulea from this crate. Ultimately we should refactor hokulea to import from this crate instead, but given that the implementations are still slightly different (hokulea uses methods on structs instead of functions like we do here) then it'll have to wait for a larger refactor.

### DIFF
--- a/rust/crates/eigenda-verification/src/verification/blob/codec.rs
+++ b/rust/crates/eigenda-verification/src/verification/blob/codec.rs
@@ -230,7 +230,9 @@ fn check_and_remove_zero_padding_for_field_elements(
 /// only available when the `test-utils` feature is enabled or during testing.
 pub mod tests_utils {
     use crate::verification::blob::BlobVerificationError::{self, *};
-    use crate::verification::blob::codec::{BYTES_PER_CHUNK, BYTES_PER_SYMBOL, HEADER_BYTES_LEN};
+    use crate::verification::blob::codec::{
+        BYTES_PER_CHUNK, BYTES_PER_SYMBOL, HEADER_BYTES_LEN, PAYLOAD_ENCODING_VERSION_0,
+    };
 
     /// Guard byte value used to prefix field elements in the EigenDA encoding.
     ///
@@ -239,11 +241,6 @@ pub mod tests_utils {
     /// operations. The value 0 ensures that the resulting 32-byte value is always
     /// less than the BN254 field modulus.
     pub const FIELD_ELEMENT_GUARD_BYTE: u8 = 0;
-
-    /// Version byte used in the encoded payload header.
-    ///
-    /// This indicates the encoding format version. Currently, only version 0 is defined.
-    pub const VERSION: u8 = 0;
 
     /// Encodes a raw payload into an EigenDA-compatible encoded payload format.
     ///
@@ -351,7 +348,7 @@ pub mod tests_utils {
     ) -> Result<[u8; HEADER_BYTES_LEN], BlobVerificationError> {
         let mut header = [0; HEADER_BYTES_LEN];
         header[0] = FIELD_ELEMENT_GUARD_BYTE;
-        header[1] = VERSION;
+        header[1] = PAYLOAD_ENCODING_VERSION_0;
         let raw_payload_len: u32 = raw_payload.len().try_into()?;
         header[2..6].copy_from_slice(&raw_payload_len.to_be_bytes());
         Ok(header)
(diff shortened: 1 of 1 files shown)


rust/crates/eigenda-verification/src/verification/blob/codec.rs
```diff
@@ -114,19 +144,54 @@ pub fn decode_encoded_payload(encoded_payload: &[u8]) -> Result<Vec<u8>, BlobVer
...
         })?;
 
+    let raw_payload_bytes_len = raw_payload_chunks_len
+        .checked_mul(BYTES_PER_CHUNK)
+        .ok_or(Overflow)?;
     let mut raw_payload = Vec::with_capacity(raw_payload_bytes_len);
 
-    for symbol in padded_payload.chunks_exact(BYTES_PER_SYMBOL) {
+    // Check 7: Validate guard bytes and decode symbols
+    for (symbol_idx, symbol) in padded_payload.chunks_exact(BYTES_PER_SYMBOL).enumerate() {
```
bxue-l2: why not take the code from [hokulea](https://github.com/Layr-Labs/hokulea/blob/ca100c76285bce38f91dfb8a391c32af8d1fc741/crates/eigenda/src/eigenda_data.rs#L79), we have it audited already. Now we have to maintain two implementation of rust

bxue-l2:
> Implementation this made me really question our choice of enforcing this. What do we get from enforcing this other than more expensive (especially when proven in zkvm), brittle, and harder to read code? Unclear to me at this point..

The motivation of enforcing it is to make sure every first byte of 32 bytes is a valid field element. Then I raised the question if we shall enforce them all, given we can literally put random bytes in other places. It is true that the code becomes a bit more complex, but we at the time in favor for conceptually simplicity

rust/crates/eigenda-verification/src/verification/blob/codec.rs
samlaf: [940a32b](https://github.com/Layr-Labs/eigenda/pull/2332/commits/940a32bfdac8f071b9868415b64ab23d1e892d42)

rust/crates/eigenda-verification/src/verification/blob/codec.rs
```diff
@@ -388,28 +474,46 @@ mod tests {
...
     fn claimed_encoded_payload_len_too_small_for_actual_payload() {
-        let mut encoded_payload = vec![0u8; HEADER_BYTES_LEN];
+        use crate::verification::blob::error::{
+            BlobVerificationError, EncodedPayloadDecodingError,
+        };
+
+        // Create a properly sized (power of two) encoded payload, but with a header claiming
+        // more payload data than is actually present
+        let mut encoded_payload = vec![0u8; 128]; // Power of two
         encoded_payload[0] = FIELD_ELEMENT_GUARD_BYTE;
```
bxue-l2: This is InvalidHeaderFirstByte

bxue-l2: think a lot of constants in the test can be replaced by the ones from hokulea, and the existing constant can be removed

samlaf: > This is InvalidHeaderFirstByte

Think you misread. This FIELD_ELEMENT_GUARD_BYTE = 0, its just a name for the first byte that the spec forces to be 0. I think its actually good to keep this constant.

But refactored the tests and imported all your hokulea tests: [7129d9d](https://github.com/Layr-Labs/eigenda/pull/2332/commits/7129d9de16df466de44a498266aa1de60c9180f5)

bxue-l2: right

rust/crates/eigenda-verification/src/verification/blob/codec.rs
```diff
@@ -331,33 +420,26 @@ pub mod tests_utils {
...
+    fn construct_header_format() {
+        for (payload, expected_len) in [
+            (vec![], 0u32),
+            (vec![1, 2, 3, 4, 5], 5u32),
+            (vec![0u8; 1000], 1000u32),
+        ] {
+            let header = construct_header(&payload).unwrap();
 
+            assert_eq!(header[0], FIELD_ELEMENT_GUARD_BYTE);
+            assert_eq!(header[1], VERSION);
```
bxue-l2: use `PAYLOAD_ENCODING_VERSION_0` defined above?

samlaf: good catch: [ebe6915](https://github.com/Layr-Labs/eigenda/pull/2332/commits/ebe69151f225145f4d49e87740f629502e1b4e63)

PR #2336, opened 2025-11-20
test(inabox): Add proxy e2e tests to inabox

iquidus:
Closes DAINT-798

Begins migrating proxy e2e tests to inabox

### DIFF
--- a/inabox/tests/proxy_setup.go
+++ b/inabox/tests/proxy_setup.go
@@ -288,7 +288,7 @@ func StartProxyServer(
 	}
 
 	if appConfig.EnabledServersConfig.ArbCustomDA {
-		arbHandlers := arbitrum_altda.NewHandlers(certMgr, logger, compatibilityCfg)
+		arbHandlers := arbitrum_altda.NewHandlers(certMgr, logger, false, compatibilityCfg)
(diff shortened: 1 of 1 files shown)


inabox/tests/proxy_setup.go
```diff
@@ -0,0 +1,323 @@
...
+	UseKeccak256ModeS3            bool
+	UseS3Caching                  bool
+	UseS3Fallback                 bool
+	ErrorOnSecondaryInsertFailure bool
+
+	ClientLedgerMode     clientledger.ClientLedgerMode
+	VaultMonitorInterval time.Duration
+
+	GlobalInfra *InfrastructureHarness
+}
```
samlaf, line 52: this is all copy-pasted from https://github.com/Layr-Labs/eigenda/blob/8217a8f9fd72d49f34ee3323918c29f7146c312f/api/proxy/test/testutils/setup.go#L153 right?
Are there any diffs? is this really needed? can't you just import the code from proxy testutils instead?

iquidus: I have a local branch taking this approach (adding inabox as a backend to existing e2e tests). However there is one issue. - Inabox SRS points are insufficient for payloads used in the tests (e.g 1_000_000 bytes)

PR #2338, opened 2025-11-20
feat(contracts): Update Config Registry to use type specific activation keys

ethenotethan:
**Changes**
- Bump directory semver to `2.0.0`
- Change storage namespaced position to deprecate legacy mappings on deployed envs
- Get rid of 32byte and bytes distinctive mappings in-favor of of just bytes
- Maintain two mappings based on activation key type (i.e, `timestamp`, `block_number`)
   - adds a time relational check when adding new entries to ensure that time based activation key is set in the future wrt block based time context 

**Consideration(s)**
- each config storage mapping requires duplicated accessor functions (e.g, `getConfigTimeStamp`, `getConfigBlockNumber`). One thing we could do is encapsulate the checkpoint type (i.e, `timestamp`, `block_number`) into an enum that's provided by a user when **introducing** a new registry value (would introduce context switching though). since EVM is 32 byte word aligned, both `timestamp` and `block_number` are 32 byte values in Solidity. 

- prior storage namespace has been subsumed by a new name-spaced key directive. this assumes that the prior (`bytes`, `bytes32`) mappings will never be referenced where they're currently deployed (i.e, `hoodi-testnet`, `hoodi-preprod`, `sepolia`). This is an ok assumption because the registry was deployed prematurely with no offchain code having been developed yet.


contracts/src/core/libraries/v3/config-registry/ConfigRegistryStorage.sol
```diff
@@ -6,13 +6,13 @@ import {ConfigRegistryTypes as T} from "src/core/libraries/v3/config-registry/Co
 /// @notice Defines the storage layout for a config registry based on ERC-7201
 ///         https://eips.ethereum.org/EIPS/eip-7201
 library ConfigRegistryStorage {
-    /// @custom: storage-location erc7201:config.registry.storage
+    /// @custom: storage-location erc7201:contract-directory.config-registry.storage
```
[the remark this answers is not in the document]
0xClandestine: we should define standard naming practice throughout the org if we're going to use namespacing

contracts/src/core/EigenDADirectory.sol
```diff
@@ -118,91 +118,91 @@ contract EigenDADirectory is IEigenDADirectory, IEigenDASemVer {
     /// CONFIG REGISTRY FUNCTIONS ///
 
     /// @inheritdoc IEigenDAConfigRegistry
-    function addConfigBytes32(string memory name, uint256 activationKey, bytes32 value) external onlyOwner {
+    function addConfigBlockNumber(string memory name, uint256 activationKey, bytes memory value) external onlyOwner {
```
samlaf: how about `addConfigAtBlockNumber/Timestamp`, or `activateConfigAtBlock/Time`?

samlaf: oh I see its a bit more awkward for functions like `getNumCheckpointsTimeStamp`... hmm

samlaf: or how about not using suffixes at all? solidity allows polymorphism afaik

contracts/src/core/EigenDADirectory.sol
```diff
@@ -118,91 +118,91 @@ contract EigenDADirectory is IEigenDADirectory, IEigenDASemVer {
...
-        return ConfigRegistryLib.getNameListBytes();
+    function getAllConfigNamesTimeStamp() external view returns (string[] memory) {
+        return ConfigRegistryLib.getNameListTimeStamp();
     }
 
     /// @inheritdoc IEigenDASemVer
     function semver() external pure returns (uint8 major, uint8 minor, uint8 patch) {
         major = 1;
-        minor = 1;
+        minor = 2;
```
samlaf: shouldnt it be 2.0.0? this is a big breaking change, all function names are changing.

contracts/src/core/libraries/v3/config-registry/ConfigRegistryStorage.sol
```diff
@@ -6,13 +6,13 @@ import {ConfigRegistryTypes as T} from "src/core/libraries/v3/config-registry/Co
...
     struct Layout {
-        T.Bytes32Cfg bytes32Config;
-        T.BytesCfg bytesConfig;
+        T.BlockNumberCfg BlockNumberCfg;
+        T.TimeStampCfg TimeStampCfg;
     }
 
-    string internal constant STORAGE_ID = "config.registry.storage";
+    string internal constant STORAGE_ID = "contract-directory.config-registry.storage";
```
samlaf: why use contract-directory? this is just a lib that should be standalone no? for example we are even considering using it in the router

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
```diff
@@ -5,84 +5,168 @@ import {ConfigRegistryStorage as S} from "src/core/libraries/v3/config-registry/
...
         if (cfg.values[nameDigest].length > 0) {
-            uint256 lastActivationKey = cfg.values[nameDigest][cfg.values[nameDigest].length - 1].activationKey;
-            if (activationKey <= lastActivationKey) {
-                revert NotIncreasingActivationKey(lastActivationKey, activationKey);
+            uint256 lastActivationKey = cfg.values[nameDigest][cfg.values[nameDigest].length - 1].activationBlock;
+            if (abn <= lastActivationKey) {
+                revert NotIncreasingActivationKey(lastActivationKey, abn);
+            }
+        } else {
+            if (block.number > abn) {
```
samlaf:
```suggestion
            if (abn < block.number) {
```

samlaf: just a nit but would be more consistent with if check above it

contracts/test/mock/MockConfigRetriever.sol
```diff
@@ -54,27 +80,32 @@ contract MockConfigRetriever {
...
         external
         view
         returns (uint256[] memory, bytes[] memory)
     {
-        uint256 numCheckpoints = directory.getNumCheckpointsBytes(key);
+        uint256 numCheckpoints = directory.getNumCheckpointsTimeStamp(key);
         uint256 count = 0;
         for (uint256 i = 0; i < numCheckpoints; i++) {
-            if (directory.getActivationKeyBytes(key, i) >= activationKey) {
+            if (directory.getActivationKeyTimeStamp(key, i) >= activationKey) {
```
samlaf: think we should get rid of the word "activationKey" everywhere. Its either a timestamp or a blocknumber. should just be called activationBlock/Time imo

samlaf: Wait I'm confused why is this in the MockConfigRetriever? Thought we wanted this view function in the ConfigRegistry itself?

ethenotethan: this is an example class that patrick left us - I vote we just nuke it going forward since the retrieval logic will deprecate its necessity
cc @0xClandestine

contracts/src/core/libraries/v3/config-registry/ConfigRegistryStorage.sol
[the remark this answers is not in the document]
ethenotethan: attempting to do that here: https://github.com/Layr-Labs/eigenda/pull/2343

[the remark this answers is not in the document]
ethenotethan: currently however we've defined namespaced storage for other contracts with inconsistent name conventions

contracts/src/core/EigenDADirectory.sol
ethenotethan: yeah makes sense - was bumping minor since this was deployed prematurely on testnets

contracts/src/core/libraries/v3/config-registry/ConfigRegistryStorage.sol
ethenotethan: was trying to change the namespace to migrate old storage - will adjust to just use `v3` suffix with code comment

contracts/src/core/interfaces/IEigenDADirectory.sol
```diff
@@ -44,94 +44,94 @@ interface IEigenDAAddressDirectory {
...
-    /// @notice Adds a 32 byte configuration value to the configuration registry.
+    /// @notice Adds a variable length byte configuration value to the configuration registry using block number as activation key.
     /// @param name The name of the configuration entry.
     /// @param activationKey The activation key for the configuration entry.
     ///                      This is an arbitrary key defined by the caller to indicate when the configuration should become active.
-    /// @param value The 32 byte configuration value.
+    /// @param value The variable length byte configuration value.
     /// @dev The activationKey must be strictly greater than the last activationKey for the same name.
-    function addConfigBytes32(string memory name, uint256 activationKey, bytes32 value) external;
+    function addConfigBlockNumber(string memory name, uint256 activationKey, bytes memory value) external;
```
litt3: Now that we are differentiating between block number and timestamp, consider renaming parameters to the specific type, e.g. `activationBlock`, and `activationTimestamp`. 

The term `activationKey` was created as a more general term to encompass both concepts, but now it's not necessary to be general. Personally, I'd search for and nuke the term entirely

contracts/src/core/interfaces/IEigenDADirectory.sol
```diff
@@ -44,94 +44,94 @@ interface IEigenDAAddressDirectory {
...
 
-    /// @notice Adds a variable length byte configuration value to the configuration registry.
+    /// @notice Adds a variable length byte configuration value to the configuration registry using timestamp as activation key.
     /// @param name The name of the configuration entry.
     /// @param activationKey The activation key for the configuration entry.
     ///                      This is an arbitrary key defined by the caller to indicate when the configuration should become active.
```
litt3: Some of the docs still mention `arbitrary key`, which is no longer the case.

Consider doing a pass over these docs, to strip out stale language like this, that has smells from the old design

contracts/src/core/interfaces/IEigenDADirectory.sol
```diff
@@ -44,94 +44,94 @@ interface IEigenDAAddressDirectory {
...
 ///         by an arbitrary activation key.
 interface IEigenDAConfigRegistry {
-    /// @notice Adds a 32 byte configuration value to the configuration registry.
+    /// @notice Adds a variable length byte configuration value to the configuration registry using block number as activation key.
     /// @param name The name of the configuration entry.
     /// @param activationKey The activation key for the configuration entry.
     ///                      This is an arbitrary key defined by the caller to indicate when the configuration should become active.
-    /// @param value The 32 byte configuration value.
+    /// @param value The variable length byte configuration value.
     /// @dev The activationKey must be strictly greater than the last activationKey for the same name.
```
litt3: Since for `activationBlockNumbers` we enforce that the block must be in the future, that condition probably ought to be mentioned here alongside the "strictly greater" condition

contracts/src/core/interfaces/IEigenDADirectory.sol
```diff
@@ -44,94 +44,94 @@ interface IEigenDAAddressDirectory {
...
-    function getActivationKeyBytes32(bytes32 nameDigest, uint256 index) external view returns (uint256);
+    function getActivationKeyBlockNumber(bytes32 nameDigest, uint256 index) external view returns (uint256);
 
-    /// @notice Gets the activation key for a variable length byte configuration entry at a specific index.
+    /// @notice Gets the activation key for a timestamp configuration entry at a specific index.
     /// @param nameDigest The hash of the name of the configuration entry.
     /// @param index The index of the configuration value to retrieve the activation key for.
     /// @return The activation key at the specified index.
-    function getActivationKeyBytes(bytes32 nameDigest, uint256 index) external view returns (uint256);
+    function getActivationKeyTimeStamp(bytes32 nameDigest, uint256 index) external view returns (uint256);
```
litt3: Piggybacking off what I said above about `activationKey`:

for example, I think `getActivationKeyTimeStamp` could be renamed to `getActivationTimeStamp`, to be more direct

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
```diff
@@ -5,84 +5,168 @@ import {ConfigRegistryStorage as S} from "src/core/libraries/v3/config-registry/
...
 
+    /// @notice Thrown when adding the first block number configuration with an activation block in the past
+    /// @param currBlock The current block number
+    /// @param abn The activation block number being added (must be >= currBlock)
+    error BlockNumberActivationKeyInPast(uint256 currBlock, uint256 abn);
+
+    /// @notice Thrown when adding the first timestamp configuration with an activation timestamp in the past
+    /// @param currTS The current timestamp
+    /// @param ats The activation timestamp being added (must be >= currTS)
+    error TimeStampActivationKeyInPast(uint256 currTS, uint256 ats);
```
litt3: How closely do the block times actually match real-world times? Is it technically possible for block timestamps to drift from real time?

The block-number assertion clearly makes sense, but I lean toward saying we _shouldn't_ be making this timestamp assertion.

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
```diff
@@ -5,84 +5,168 @@ import {ConfigRegistryStorage as S} from "src/core/libraries/v3/config-registry/
...
-    function addConfigBytes32(bytes32 nameDigest, uint256 activationKey, bytes32 value) internal {
-        T.Bytes32Cfg storage cfg = S.layout().bytes32Config;
+    /// @notice Adds a new timestamp-based configuration checkpoint
+    /// @param nameDigest The hash of the configuration name
+    /// @param activationTS The activation timestamp (must be > last activation timestamp for this config)
+    /// @param value The bytes configuration value
+    /// @dev For the first checkpoint, activationTS must be >= block.timestamp
+    /// @dev Subsequent checkpoints must have strictly increasing activation timestamps
+    function addConfigTimeStamp(bytes32 nameDigest, uint256 activationTS, bytes memory value) internal {
+        T.TimeStampCfg storage cfg = S.layout().TimeStampCfg;
```
litt3: nit: throughout the codebase, we write `timestamp` as a single word (even in your docs for these contracts). Therefore it seems odd to camelCase it as `TimeStamp`, as if it were two words: "time stamp"

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
```diff
@@ -5,84 +5,168 @@ import {ConfigRegistryStorage as S} from "src/core/libraries/v3/config-registry/
...
-            uint256 lastActivationKey = cfg.values[nameDigest][cfg.values[nameDigest].length - 1].activationKey;
-            if (activationKey <= lastActivationKey) {
-                revert NotIncreasingActivationKey(lastActivationKey, activationKey);
+            uint256 lastActivationKey = cfg.values[nameDigest][cfg.values[nameDigest].length - 1].activationBlock;
+            if (abn <= lastActivationKey) {
+                revert NotIncreasingActivationKey(lastActivationKey, abn);
+            }
+        } else {
+            if (block.number > abn) {
+                revert BlockNumberActivationKeyInPast(block.number, abn);
```
litt3: Shouldn't we be asserting that the new block number is > `lastActivationKey` AND not in the past?

Looks like the way you have it coded, the only time you assert that the ABN isn't in the past is when adding the first config value

contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
```diff
@@ -11,34 +11,34 @@ library ConfigRegistryTypes {
...
+    /// @param activationBlock The activation block number for the checkpoint
     /// @param value The bytes configuration value at this checkpoint
-    struct BytesCheckpoint {
-        uint256 activationKey;
+    struct BlockNumberCheckpoint {
+        uint256 activationBlock;
         bytes value;
     }
 
     /// @notice Struct to hold all bytes32 configuration checkpoints and associated names
```
litt3: Here and elsewhere, you have stale docs referencing `bytes32 configuration`.

contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
```diff
@@ -11,34 +11,34 @@ library ConfigRegistryTypes {
...
     /// @param nameSet The NameSet struct to manage names associated with the configuration entries
     /// @dev See docs for the structs for more information
-    struct Bytes32Cfg {
-        mapping(bytes32 => Bytes32Checkpoint[]) values;
+    struct TimeStampCfg {
+        mapping(bytes32 => TimeStampCheckpoint[]) values;
         NameSet nameSet;
     }
 
     /// @notice Struct to hold all bytes configuration checkpoints and associated names
```
litt3: This doc is also stale, referencing `bytes configuration` instead of `BlockNumber`, which is actually the notable characteristic of this struct

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
ethenotethan: hmm - iiuc there's only around 15 seconds of allowed drift that a block miner is allowed wrt current time: https://metaschool.so/articles/what-is-block-timestamp-in-solidity

ethenotethan: from where I got 15 seconds from: https://owasp.org/www-project-smart-contract-top-10/2023/en/src/SC03-timestamp-dependence.html

ethenotethan: tried looking at go-ethereum implementation to see if I could source the block mining logic directly but got lazy

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
ethenotethan: hmm yeah that is a good point

contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
ethenotethan: [109580e](https://github.com/Layr-Labs/eigenda/pull/2338/commits/109580e4a0486b63a45b6a6bdaa62f1449be04ef)

contracts/src/core/libraries/v3/config-registry/ConfigRegistryTypes.sol
ethenotethan: [109580e](https://github.com/Layr-Labs/eigenda/pull/2338/commits/109580e4a0486b63a45b6a6bdaa62f1449be04ef)

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
ethenotethan: [109580e](https://github.com/Layr-Labs/eigenda/pull/2338/commits/109580e4a0486b63a45b6a6bdaa62f1449be04ef)

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
ethenotethan: [109580e](https://github.com/Layr-Labs/eigenda/pull/2338/commits/109580e4a0486b63a45b6a6bdaa62f1449be04ef)

ethenotethan: ^ updated for [06815e7](https://github.com/Layr-Labs/eigenda/pull/2338/commits/06815e7a4a7fad52db63ba8532f3f6bc0699f46f)

contracts/src/core/interfaces/IEigenDADirectory.sol
ethenotethan: [109580e](https://github.com/Layr-Labs/eigenda/pull/2338/commits/109580e4a0486b63a45b6a6bdaa62f1449be04ef)

contracts/src/core/interfaces/IEigenDADirectory.sol
ethenotethan: [109580e](https://github.com/Layr-Labs/eigenda/pull/2338/commits/109580e4a0486b63a45b6a6bdaa62f1449be04ef)

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
ethenotethan: [ab17bda](https://github.com/Layr-Labs/eigenda/pull/2338/commits/ab17bdaeb7bcba3e056576fbb4767306b91b90d6)

contracts/src/core/libraries/v3/config-registry/ConfigRegistryStorage.sol
ethenotethan: [ab17bda](https://github.com/Layr-Labs/eigenda/pull/2338/commits/ab17bdaeb7bcba3e056576fbb4767306b91b90d6)

contracts/src/core/EigenDADirectory.sol
ethenotethan: [ab17bda](https://github.com/Layr-Labs/eigenda/pull/2338/commits/ab17bdaeb7bcba3e056576fbb4767306b91b90d6)

contracts/src/core/interfaces/IEigenDADirectory.sol
ethenotethan: https://github.com/Layr-Labs/eigenda/pull/2338/commits/109580e4a0486b63a45b6a6bdaa62f1449be04ef

ethenotethan: https://github.com/Layr-Labs/eigenda/pull/2338/commits/cefc88a636ca12e0e43f4c6805e732a1d403cbe1

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
```diff
@@ -5,26 +5,32 @@ import {ConfigRegistryStorage as S} from "src/core/libraries/v3/config-registry/
...
+
+    /// @notice Thrown when trying to add a configuration with a block number that is not strictly increasing
+    /// @param prevABN The last activation block number for this configuration
+    /// @param newABN The activation block number being added (must be > prevABN)
+    error NotIncreasingBlockNumber(uint256 prevABN, uint256 newABN);
 
     /// @notice Thrown when adding the first block number configuration with an activation block in the past
-    /// @param currBlock The current block number
+    /// @param currBlock The current block number (sourced via block.number)
     /// @param abn The activation block number being added (must be >= currBlock)
```
litt3, line 27: > (must be >= currBlock)

This statement doesn't agree with the doc above:

> must be greater than the current block number.

contracts/src/core/libraries/v3/config-registry/ConfigRegistryStorage.sol
samlaf: we prob should keep a comment mentioning the previous namespaces we've used in the past and that are polluted in the proxy right? In case someone comes later and tries to reuse another name. There's less chance of someone removing v2 for sure, but still, just good habit?

contracts/src/core/libraries/v3/config-registry/ConfigRegistryLib.sol
```diff
@@ -133,34 +140,35 @@ library ConfigRegistryLib {
...
-        } else {
-            if (block.number > abn) {
-                revert BlockNumberActivationKeyInPast(block.number, abn);
+            uint256 lastABN = cfg.values[nameDigest][cfg.values[nameDigest].length - 1].activationBlock;
+            if (abn <= lastABN) {
+                revert NotIncreasingBlockNumber(lastABN, abn);
             }
         }
 
+        /// @dev abn being provided must always be at a future block
```
litt3, line 151: > must always be at a future block

This doesn't agree with the impl, which permits the block number being the _current_ block, in addition to future blocks. The same doc mismatch exists above for the timestamp version of this

contracts/src/core/EigenDADirectory.sol
0xClandestine: Changing method identifiers isn't problematic here assuming nobody is integrating the da directory right?

PR #2340, opened 2025-11-20
chore: Flag to disable GetBlobCommitment()

crypt0fairy:
## Summary

This PR adds a configuration flag to deprecate the `GetBlobCommitment()` gRPC endpoint, allowing operators to disable it gracefully before eventual removal.

## Motivation

The `GetBlobCommitment()` endpoint is being deprecated for the following reasons:

- **Major DOS vector**: The endpoint is free, unpermissioned, and requires massive server-side computational load per call
- **Minimal usage**: Only used by one customer (calling once every 15 minutes on mainnet)
- **Unnecessary**: Clients can compute blob commitments locally using the SRS points

Rather than immediately disabling the endpoint (which would break the existing customer), this PR provides a deprecation mechanism that:
1. Allows operators to disable the endpoint with a clear error message
2. Gives customers time to migrate to local computation
3. Enables a graceful deprecation path before eventual removal

### Configuration
- Added `--disperser-server-disable-get-blob-commitment` CLI flag
- Added `DISPERSER_SERVER_DISABLE_GET_BLOB_COMMITMENT` environment variable
- Flag defaults to `false` (endpoint enabled by default)

### Implementation
- When the flag is enabled, `GetBlobCommitment()` returns:
  - Status code: `Unimplemented`
  - Error message: "GetBlobCommitment is deprecated and has been disabled. This service will be removed in a future release. Please compute blob commitments locally."

### Testing
- Added unit test `TestV2GetBlobCommitment_Disabled` to verify deprecation behavior
- Existing test `TestV2GetBlobCommitment` verifies normal operation when flag is disabled
- Both tests pass successfully

### Enable Deprecation (Disable Endpoint)
```bash
# Via CLI flag
--disperser-server-disable-get-blob-commitment

# Via environment variable
DISPERSER_SERVER_DISABLE_GET_BLOB_COMMITMENT=true
```

### Expected Behavior

**When flag is disabled (default):**
```bash
$ grpcurl -plaintext -d '{"blob": "..."}' localhost:32005 disperser.v2.Disperser/GetBlobCommitment
{
  "blobCommitment": {
    "commitment": "...",
    "lengthCommitment": "...",
    "lengthProof": "...",
    "length": 1
  }
}
```

**When flag is enabled:**
```bash
$ grpcurl -plaintext -d '{"blob": "..."}' localhost:32005 disperser.v2.Disperser/GetBlobCommitment
ERROR:
  Code: Unimplemented
  Message: GetBlobCommitment is deprecated and has been disabled. This service will be removed in a future release. Please compute blob commitments locally.
```

## Migration Path

1. **Current state**: Flag defaults to `false` - endpoint remains enabled
2. **Deprecation phase**: After public announcement, operators can enable the flag to disable the endpoint
3. **Removal phase**: After 1-2 months, the endpoint will be removed entirely (separate PR)

## Files Changed

- `disperser/server_config.go` - Added `DisableGetBlobCommitment` field to `ServerConfig`
- `disperser/cmd/apiserver/flags/flags.go` - Added CLI flag definition
- `disperser/cmd/apiserver/lib/config.go` - Read flag and set in config
- `disperser/apiserver/server_v2.go` - Added deprecation check in `getBlobCommitment()`
- `disperser/apiserver/server_v2_test.go` - Added test for deprecation behavior

## Testing

- ✅ Unit tests pass (both enabled and disabled scenarios)
- ✅ gRPC endpoint tested with grpcurl
- ✅ Error message and status code verified
- ✅ Backward compatible (defaults to enabled)

### DIFF
--- a/disperser/apiserver/server_v2_test.go
+++ b/disperser/apiserver/server_v2_test.go
@@ -447,7 +447,28 @@ func TestV2GetBlobCommitment(t *testing.T) {
 	require.Equal(t, uint32(commit.Length), reply.GetBlobCommitment().GetLength())
 }
 
+func TestV2GetBlobCommitment_Disabled(t *testing.T) {
+	ctx := t.Context()
+	c := newTestServerV2WithDeprecationFlag(t, true)
+	data := make([]byte, 50)
+	_, err := rand.Read(data)
+	require.NoError(t, err)
+
+	data = codec.ConvertByPaddingEmptyByte(data)
+	reply, err := c.DispersalServerV2.GetBlobCommitment(ctx, &pbv2.BlobCommitmentRequest{
+		Blob: data,
+	})
+	require.Error(t, err)
+	require.Nil(t, reply)
+	require.ErrorContains(t, err, "GetBlobCommitment is deprecated and has been disabled")
+	require.ErrorContains(t, err, "This service will be removed in a future release")
+}
+
 func newTestServerV2(t *testing.T) *testComponents {
+	return newTestServerV2WithDeprecationFlag(t, false)
+}
+
+func newTestServerV2WithDeprecationFlag(t *testing.T, disableGetBlobCommitment bool) *testComponents {
 	t.Helper()
 
 	ctx := t.Context()
@@ -550,8 +571,9 @@ func newTestServerV2(t *testing.T) *testComponents {
 
 	s, err := apiserver.NewDispersalServerV2(
 		disperser.ServerConfig{
-			GrpcPort:    "51002",
-			GrpcTimeout: 1 * time.Second,
+			GrpcPort:                 "51002",
+			GrpcTimeout:              1 * time.Second,
+			DisableGetBlobCommitment: disableGetBlobCommitment,
 		},
 		time.Now,
 		blobStore,


cody-littley:
Please mark yourself as the assignee when you open a PR. Not a huge deal if you forget, but it helps keeps PRs organized.

Screenshot 2025-11-20 at 12 08 50 PM

disperser/apiserver/server_v2_test.go
```diff
@@ -447,7 +447,28 @@ func TestV2GetBlobCommitment(t *testing.T) {
...
+	require.NoError(t, err)
+
+	data = codec.ConvertByPaddingEmptyByte(data)
+	reply, err := c.DispersalServerV2.GetBlobCommitment(ctx, &pbv2.BlobCommitmentRequest{
+		Blob: data,
+	})
+	require.Error(t, err)
+	require.Nil(t, reply)
+	require.ErrorContains(t, err, "GetBlobCommitment is deprecated and has been disabled")
+	require.ErrorContains(t, err, "This service will be removed in a future release")
```
cody-littley, line 464: When I'm writing unit tests, I personally prefer to avoid doing string matching on error messages, since the test ends up being more fragile. Up to you whether you'd like to keep this assertion or not, since this request is more of a personal preference thing.

PR #2348, opened 2025-11-21
feat(clients): Implement simple reputation system

litt3:
- part of EGDA-2299 ([link](https://linear.app/eigenlabs/issue/EGDA-2299/client-disperser-multiplexing))
- implements a simple reputation system for choosing between different dispersers

### DIFF
--- a/common/reputation/reputation.go
+++ b/common/reputation/reputation.go
@@ -55,15 +55,17 @@ func (r *Reputation) Score(now time.Time) float64 {
 }
 
 // Applies time-based drift toward the neutral forgiveness target.
-// Only increases scores that are below the target - scores above the target are unchanged.
+// Only increases scores that are below the target: scores >= the target are unchanged.
+//
+// The score approaches the target exponentially. After one half-life period, the score will have recovered halfway
+// from its starting value to the target.
+//
+// Forgiveness applies only while the score is below the target. Within such periods, the forgiveness curve is
+// continuous and time-invariant: the final score depends only on the total time spent below the target, not on
+// how frequently forgiveness is applied.
 func (r *Reputation) forgive(now time.Time) {
 	if r.previousForgivenessTime.IsZero() {
 		r.previousForgivenessTime = now
-		return
-	}
-
-	// Only forgive if score is below the forgiveness target
-	if r.score >= r.config.ForgivenessTarget {
 		return
 	}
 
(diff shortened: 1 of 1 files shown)


common/reputation/reputation.go
```diff
@@ -0,0 +1,74 @@
...
+}
+
+// Returns the current reputation score.
+func (r *Reputation) Score() float64 {
+	return r.score
+}
+
+// Forgive applies time-based drift toward the neutral forgiveness target.
+// Only increases scores that are below the target - scores above the target are unchanged.
+func (r *Reputation) Forgive(now time.Time) {
```
cody-littley: Would it make sense to do the following?

- make `Forgive()` a non-exported function
- add a `now time.Time` argument to `Sucess()`, `Failure()`, and `Score()`
- Before modifying the score as a result of a success or a failure, always call `forgive()`

This would simplify the API of this class a little, as there is now one fewer public method to call. 

It would also make this utility simulate continual forgiveness. As it is now, if you wait long periods between calling `Forgive()` with continuous calls to `Success()` and `Failure()`, the reputation you will have at the end will not be the same as the reputation you would have if forgiveness was continuous (since reputation is clamped between 0.0 and 1.0). IMO, continual forgiveness is how my mind intuitively thought a utility like this would work, and I didn't realize the difference until I read the implementation.

common/reputation/reputation.go
```diff
@@ -0,0 +1,74 @@
...
+	return &Reputation{
+		config:                  config,
+		score:                   config.ForgivenessTarget,
+		previousForgivenessTime: now,
+	}
+}
+
+// Updates the reputation after a successful interaction.
+// Moves the score toward 1.0 based on the configured success update rate.
+func (r *Reputation) Success() {
```
cody-littley: nit, feel free to ignore if you think otherwise: `Success()` -> `ReportSuccess()`, `Failure()` -> `ReportFailure()`.

common/reputation/reputation.go
```diff
@@ -0,0 +1,74 @@
...
+		return
+	}
+
+	elapsed := now.Sub(r.previousForgivenessTime).Seconds()
+	if elapsed <= 0 {
+		return
+	}
+
+	forgivenessRate := math.Log(2) / r.config.ForgivenessHalfLife.Seconds()
+	forgivenessFraction := 1 - math.Exp(-forgivenessRate*elapsed)
```
cody-littley, line 85: Can you add documentation for the curve that this produces? Assuming we ignore floating point rounding errors, does this produce different results depending on the frequency that `Forgive()` is called? For example, if I call `Forgive()` once per second vs. once per minute, will I get different results?

common/reputation/reputation.go
litt3: Good callout, that's much better 08312b84a3f27b202cc35e6ed034cf93ad33d628

common/reputation/reputation.go
litt3: 7737625c672eb04c3c08cb19505df84bf62812e2

common/reputation/reputation.go
litt3: Improved the doc.

f38fc381f62df316a267b50e6d7ae84200a2ff9a

I also reorganized the `forgiveness` method, since there was a problem with the old ordering that would cause a forgiveness windfall upon dropping below the target.

PR #2351, opened 2025-11-23
refactor: migrate encoder to use config.Bootstrap pattern like ejector

crypt0fairy:
# Refactor: Encoder Config to Match Ejector Pattern

## Summary

This PR refactors the encoder configuration to follow the same pattern as the ejector configuration, improving consistency across the codebase and enabling automatic documentation generation.

### 1. Encoder Config Refactoring
- **Moved encoder config** from `disperser/cmd/encoder/config.go` (main package) to `disperser/encoder/encoder_config.go` (encoder package)
  - This allows the config to be imported and used by other packages (e.g., for documentation generation)
  - Follows the same pattern as `ejector/ejector_config.go`
- **Added `RootEncoderConfig`** with `Config` and `Secret` separation for better security
- **Implemented `DocumentedConfig` interface** to enable automatic documentation generation
- **Added comprehensive GoDoc comments** for all config fields

### 2. Configuration Documentation
- **Added encoder config documentation generation** to `common/config/doc_generator/main.go`
- **Generated `docs/config/Encoder.md`** with complete configuration reference
- Updated `GetPackagePaths()` to include all nested config package paths (blobstore, chunkstore, kzg, aws)

### 3. Logging Improvements
- **Added `StringToLogLevel()` utility function** in `common/logger_config.go`
  - Converts string values to `slog.Level` with case-insensitive parsing
  - Returns descriptive errors for invalid log levels
- **Added GoDoc comments** for encoder log configuration fields (LogFormat, LogColor, LogLevel)
- **Set default log level** to "info" in encoder config

### 4. Inabox Deployment Updates
- Updated `inabox/deploy/config.go` and `inabox/deploy/env_vars.go` to work with new encoder config structure
- Updated `inabox/templates/testconfig-anvil.yaml` with encoder configuration
- Fixed encoder service startup in `inabox/bin.sh` and test harness

### 5. Other Improvements
- Updated AWS CLI configuration handling
- Fixed controller service proto definitions
- Updated Rust integration tests and contract bindings
- Various bug fixes for e2e tests

## Benefits

1. **Consistency**: Encoder config now follows the same pattern as ejector, making the codebase more maintainable
2. **Documentation**: Automatic generation of encoder config documentation ensures it stays up-to-date
3. **Reusability**: Config can now be imported by other packages (e.g., for testing, documentation)
4. **Better Developer Experience**: Clear GoDoc comments and utility functions make configuration easier to understand and use

## Testing

- ✅ All e2e tests pass
- ✅ Inabox deployment works correctly
- ✅ Config documentation generates successfully
- ✅ Rust integration tests updated and passing

## Files Changed

- Key files:
  - `disperser/encoder/encoder_config.go` (new, 197 lines)
  - `disperser/cmd/encoder/config.go` (deleted)
  - `common/logger_config.go` (added StringToLogLevel)
  - `common/config/doc_generator/main.go` (added encoder doc generation)
  - `docs/config/Encoder.md` (new, auto-generated)
  - `inabox/deploy/*` (updated for new config structure)

## Migration Notes

For users deploying encoder services:
- Environment variable prefix remains `ENCODER`
- Config structure is now separated into `Config` and `Secret` sections
- Default log level is now "info" (was previously unset)
- All existing configuration options remain supported

### DIFF
--- a/disperser/encoder/encoder_config.go
+++ b/disperser/encoder/encoder_config.go
@@ -83,7 +83,7 @@ func DefaultEncoderConfig() *EncoderConfig {
 		},
 		Kzg: kzg.KzgConfig{
 			SRSOrder:        268435456,
-			SRSNumberToLoad: 2097152,
+			SRSNumberToLoad: 524288,
 			NumWorker:       uint64(runtime.GOMAXPROCS(0)),
 			PreloadEncoder:  false,
 			Verbose:         false,


common/aws/cli.go
```diff
@@ -23,11 +23,11 @@ type ClientConfig struct {
 	// Region is the region to use when interacting with S3. Default is "us-east-2".
 	Region string
 	// AccessKey to use when interacting with S3.
-	AccessKey string
+	AccessKey string `docs:"required"`
 	// SecretAccessKey to use when interacting with S3.
-	SecretAccessKey string
+	SecretAccessKey string `docs:"required"`
```
crypt0fairy: @cody-littley should we use SecretString here?

cody-littley: Yup! This is the exact sort of thing the `secret.Secret` is intended for. Using a normal string here won't necessarily result in us leaking a secret, but it's good defensive programming to guard against.

common/config/doc_generator/main.go
```diff
@@ -16,4 +17,8 @@ func main() {
 
 	err = config.DocumentConfig(ejector.DefaultRootEjectorConfig, configDocsDir, true)
 	enforce.NilError(err, "failed to generate docs for the ejector config")
+
+	// requireDocs is false because nested configs (blobstore, chunkstore, etc.) may not have complete GoDoc comments
+	err = config.DocumentConfig(encoder.DefaultRootEncoderConfig, configDocsDir, false)
+	enforce.NilError(err, "failed to generate docs for the encoder config")
```
cody-littley: We'll want to circle back at some point in time to fix this, but ok to do that later and in a separate PR.

disperser/cmd/encoder/main.go
```diff
@@ -30,69 +28,74 @@ var (
...
+// Run the encoder. This method is split from main() so we only have to use log.Fatalf() once.
+func run(_ context.Context) error {
+	rootCfg, err := config.Bootstrap(encoder.DefaultRootEncoderConfig)
 	if err != nil {
-		return err
+		return fmt.Errorf("failed to bootstrap config: %w", err)
+	}
+	encoderConfig := rootCfg.Config
+	// Ensure we don't accidentally use rootCfg after this point.
+	rootCfg = nil
```
cody-littley: Instead of having a special config for secrets, we can now collapse this into a single struct and store the secret values as `secret.Secret` types. The only reason why the encoder config was split this way was an attempt to avoid accidentally leaking secrets if the config struct ever gets printed.

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,197 @@
...
+
+var _ config.DocumentedConfig = (*RootEncoderConfig)(nil)
+
+// The root configuration for the encoder service. This config should be discarded after parsing
+// and only the sub-configs should be used. This is a safety mechanism to make it harder to
+// accidentally print/log the secret config.
+type RootEncoderConfig struct {
+	Config *EncoderConfig
+	Secret *EncoderSecretConfig
+}
```
cody-littley: See above comment, these encoder configs can be collapsed into a single struct.

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,197 @@
...
+// Currently empty as AWS credentials are handled through aws.ClientConfig,
+// but this structure is kept for consistency with the ejector pattern
+// and potential future secret fields.
+type EncoderSecretConfig struct {
+}
+
+func (c *EncoderSecretConfig) Verify() error {
+	// No secrets to verify currently
+	return nil
+}
```
cody-littley: This can be deleted, and the root config just replaced with `EncoderConfig`. If we ever have a secret type that can't represented by a string, we can always create utilities like `secret.Secret` that can store those types.

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,197 @@
...
+			HTTPPort: "9100",
+		},
+		LogFormat: string(common.JSONLogFormat),
+		LogColor:  false,
+		LogLevel:  "info",
+		Aws: aws.ClientConfig{
+			Region: "us-east-1",
+		},
+	}
+}
```
cody-littley, line 108: @dmanc can you sanity check these default values?

disperser/cmd/encoder/flags/flags.go
```diff
@@ -10,7 +10,7 @@ import (
 
 const (
 	FlagPrefix   = "disperser-encoder"
-	envVarPrefix = "DISPERSER_ENCODER"
+	envVarPrefix = "ENCODER"
```
litt3, line 13: Won't this cause all the existing env vars to change, and break current configurations?

Seems like we need @cody-littley's env var aliasing before we make breaking changes like that, but it's possible I'm missing something here

common/config/bootstrap.go
```diff
@@ -97,9 +97,10 @@ func Bootstrap[T DocumentedConfig](
 		return cfg, nil
 	default:
 		// Help was shown, return zero value
-		var zero T
-		return zero, nil
+		os.Exit(0)
 	}
+	var zero T
+	return zero, nil
```
litt3, line 115:
```
	var zero T
	return zero, nil
```

These lines are unreachable

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,197 @@
...
+	Config *EncoderConfig
+	Secret *EncoderSecretConfig
+}
+
+var _ config.VerifiableConfig = (*EncoderConfig)(nil)
+
+// Configuration for the encoder.
+type EncoderConfig struct {
+	// Encoder version (1 or 2)
+	EncoderVersion uint `docs:"required"`
```
litt3: Is there a specific reason to remove the enum for this field?

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,197 @@
...
+
+	// Port at which encoder listens for gRPC calls
+	GrpcPort string `docs:"required"`
+
+	// Object storage configuration
+	BlobStore  blobstore.Config
+	ChunkStore chunkstore.Config
+
+	// KZG configuration
```
litt3:
```
// KZG configuration
Kzg kzg.KzgConfig
```

nit: IMO these sort of comments don't add anything. My personal strategy is just to leave no comment, if there's nothing to add beyond what's stated in the field + type name

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,197 @@
...
+		EncoderVersion: 1,
+		GrpcPort:       "34000",
+		BlobStore: blobstore.Config{
+			Backend: blobstore.S3Backend,
+		},
+		ChunkStore: chunkstore.Config{
+			Backend: string(blobstore.S3Backend),
+		},
+		Kzg: kzg.KzgConfig{
+			SRSOrder:        10000,
```
litt3: Proper default is `268435456`, see [this comment](https://github.com/Layr-Labs/eigenda/blob/master/encoding/v1/kzg/kzgconfig.go#L16)

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,197 @@
...
+		if c.BlobStore.BucketName == "" {
+			return fmt.Errorf("blob store bucket name is required for encoder v2")
+		}
+		if c.ChunkStore.BucketName == "" {
+			return fmt.Errorf("chunk store bucket name is required for encoder v2")
+		}
+	}
+
+	// Verify KZG config
+	if c.Kzg.G1Path == "" {
```
litt3, line 130: You're doing checks for the `Kzg` and `Server` configs here, but would it make more sense to make those config types implement the correct config interface?

Maybe it would make sense to do that for those nested types first? curious your thoughts @cody-littley

common/aws/cli.go
crypt0fairy: This change might be a separate PR, as AWS Config is used in other services. Too many changes outside the encoder's scope.

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,197 @@
...
+		GrpcPort:       "34000",
+		BlobStore: blobstore.Config{
+			Backend: blobstore.S3Backend,
+		},
+		ChunkStore: chunkstore.Config{
+			Backend: string(blobstore.S3Backend),
+		},
+		Kzg: kzg.KzgConfig{
+			SRSOrder:        10000,
+			SRSNumberToLoad: 10000,
```
litt3: IMO we shouldn't have a default value here, since the `SRSNumberToLoad` is dependent on the configured max blob size.

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,197 @@
...
+			NumWorker:       12,
```
litt3: I'm also not sure if we should have a default here. AFAICT this is normally configured to be `runtime.GOMAXPROCS(0)`, and defining a default seems likely to cause that default to be used, instead of properly checking `runtime.GOMAXPROCS(0)`

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,197 @@
...
+
+func (c *EncoderSecretConfig) Verify() error {
+	// No secrets to verify currently
+	return nil
+}
+
+// DefaultEncoderConfig returns a default configuration for the encoder.
+func DefaultEncoderConfig() *EncoderConfig {
+	return &EncoderConfig{
+		EncoderVersion: 1,
```
litt3: A more sensible default would probably be the newest version.

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,197 @@
...
+	Kzg kzg.KzgConfig
+
+	// Server configuration
+	Server ServerConfig
+
+	// Metrics configuration
+	Metrics MetricsConfig
+
+	// LogFormat is the format of the logs: json or text
+	LogFormat string
```
litt3: This isn't marked as `required`, but if it's undefined then creating the encoder will cause an error.

disperser/cmd/encoder/flags/flags.go
cody-littley: This will change environment variable names, but since the encoder is 100% internally manged by us, we can just rename the variables in the devops repo (it will be important to merge the devops repo PR in tandem with this one).

common/config/bootstrap.go
cody-littley: Yup, but the compiler is a dum dum and isn't aware that `os.Exit()` never returns. Might be worth a one line comment saying as much though.

disperser/encoder/encoder_config.go
cody-littley: https://github.com/Layr-Labs/eigenda-devops/blob/master/charts/encoder/values/eigenda-mainnet/us-east-1/ethereum/values-v2-icicle.yaml is probably a good guide for default values

disperser/encoder/encoder_config.go
cody-littley: I wonder if it makes sense to have defaults that work with max blob size == 16mb, since I doubt we are going to change the max blob size any time soon.

disperser/encoder/encoder_config.go
cody-littley: Could we just make sure it has a good default value?

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,197 @@
...
+	// LogColor is a flag to enable color in the logs
+	LogColor bool
+	// LogLevel is the level of the logs: debug, info, warn, error
+	LogLevel string
```
cody-littley, line 51: Could we just make sure it has a good default value?

disperser/encoder/encoder_config.go
cody-littley: The configuration engine doesn't care if nested structs implement any interfaces, it only requires that the top level config implements the interface. 

In several other places, I've made sub-structs implement `VerifiableConfig`. This is mostly just to make the code look nice, rather than a functional requirement.

- It allows you to verify sub-config in other places, e.g. unit tests, or if you reuse a sub-struct as a member of a different subsystem's configuration.
- It co-locates the verification logic with the struct definition, which is nice.
- It signifies the intent that the struct is a config struct and that it will be verified at startup time.
- It can reduce the line count for files containing very large config structs.

litt3: > The configuration engine doesn't care if nested structs implement any interfaces, it only requires that the top level config implements the interface.

We should look through these sub-structs, and consider each field to decide whether it ought to be marked `required`

> This is mostly just to make the code look nice, rather than a functional requirement.

Makes sense. From a maintainability perspective, IMO we should always put this logic where the structs are defined.

Can we split the setting of default values and the verification functions into separate functions on each sub-struct, where possible? I think this is particularly important for the `KzgConfig`, which is used throughout to code base

disperser/encoder/encoder_config.go
litt3: I'd be on board with that. 

Note, however, that there are times when _clients_, who also use this structure, may want to select a smaller value. If a given client is only used to submit small blobs, it can be configured to load in fewer points

disperser/encoder/encoder_config.go
litt3: Let's make sure we're on the same page regarding the semantics of a config field being marked `required`:

When doing my review pass, I was thinking that a field should be marked `required` when that field _must_ be defined, in order for the config struct to function correctly in the system.

But based on your comment, I'm inferring that you understand the `required` tag to mean something else: specifically, "required" in the CLI flag sense: "the user must specify a value, since there is no reasonable default".

Could you confirm which understanding is the correct one?

disperser/encoder/encoder_config.go
cody-littley: @crypt0fairy if you just leave the godocs blank for stuff you don't have context on, I'm willing to circle back and generate godocs for the config structs that are missing them.

disperser/encoder/encoder_config.go
cody-littley: > required" in the CLI flag sense: "the user must specify a value, since there is no reasonable default

☝️ This is how I've been using the term.

- Required: no sensical default value, user must provide a config value or else the program will crash at config loading time
- Not required: there is a sensical default value, and in theory the user can run the program without setting it.

disperser/encoder/encoder_config.go
cody-littley: With respect to figuring out which flags need to be required, I suggest the following general approach:

- Check to see if the flag is in defined for the mainnet config
    - If not defined, it must have a sensical default and is not required.
    - If defined, then check to see if it has the same value in all environments (mainnet/testnets/preprod)
        - If it has the same value everywhere then use that value as the default value and remove it from the devops config
         - If it has a different value everywhere, then it is probably a required variable

disperser/encoder/encoder_config.go
litt3: Ok, my understanding during my initial pass was wrong. My initial comment above may be ignored

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,197 @@
...
+
+var _ config.VerifiableConfig = (*EncoderConfig)(nil)
+
+// Configuration for the encoder.
+type EncoderConfig struct {
+	// Encoder version (1 or 2)
+	EncoderVersion uint `docs:"required"`
+
+	// Port at which encoder listens for gRPC calls
+	GrpcPort string `docs:"required"`
```
litt3: I don't think that `EncoderVersion` and `GrpcPort` should be marked as `required`. 

As I discussed with @cody-littley below: `required` can be understood to mean "the user must set this value, since there is no sensible default to be chosen". 

Since these both have sensible default values, they shouldn't be marked as `required`.

disperser/encoder/encoder_config.go
crypt0fairy: @cody-littley @litt3 it doesn't make sense to change it here. as There are more than 20 places across other services where this value is used.

common/aws/cli.go
crypt0fairy: AWS dependency is in more than 20 files.

disperser/encoder/encoder_config.go
litt3: `10000` is simply the wrong value. If it's wrong elsewhere, those should be changed at some point, too. 

We definitely don't want to define a "default" value which is incorrect.

disperser/encoder/encoder_config.go
crypt0fairy: I returned enum back.

disperser/encoder/encoder_config.go
crypt0fairy: now this field is defaulted to `json`

disperser/encoder/encoder_config.go
crypt0fairy: default value: `debug`

common/config/bootstrap.go
crypt0fairy: comment added

crypt0fairy (commented):
Files are updated

common/aws/cli.go
```diff
@@ -21,13 +21,13 @@ var (
 
 type ClientConfig struct {
 	// Region is the region to use when interacting with S3. Default is "us-east-2".
```
litt3, line 23: > Default is "us-east-2"

Since this is now a `required` field, it seems this doc about a default value is stale

common/aws/cli.go
```diff
@@ -21,13 +21,13 @@ var (
...
-	Region string
+	Region string `docs:"required"`
 	// AccessKey to use when interacting with S3.
-	AccessKey string
+	AccessKey string `docs:"required"`
 	// SecretAccessKey to use when interacting with S3.
-	SecretAccessKey string
+	SecretAccessKey string `docs:"required"`
 	// EndpointURL of the S3 endpoint to use. If this is not set then the default AWS S3 endpoint will be used.
```
litt3, line 29: This doc also mentions "default" behavior, but is marked `required`.

More generally, I'm curious if these fields actually _should_ be marked required. The docs state that they are for S3. Since it's possible to use OCI instead of S3, wouldn't it be ok to have these fields be undefined in OCI mode? I'm not super familiar with how these different modes work, it's possible I'm missing something

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,157 @@
...
+		return fmt.Errorf("G1 path is required")
+	}
+
+	if c.Kzg.SRSNumberToLoad == 0 {
+		return fmt.Errorf("SRS number to load must be greater than 0")
+	}
+
+	if c.Kzg.NumWorker == 0 {
+		return fmt.Errorf("number of workers must be greater than 0")
+	}
```
litt3, line 140: There are some fields in the `Kzg` struct that are now marked `required`, but values aren't being checked here.

This is one reason why I think it would make sense to put the `Kzg` related verification into the `Kzg` config file directly. It's easy to miss checks like this when the verification logic is far away from the struct definition

encoding/v1/kzg/kzgconfig.go
```diff
@@ -37,7 +37,13 @@ type KzgConfig struct {
 	//    contains the trailing end of the G2 SRS file.
 	// TODO(samlaf): to prevent misconfigurations and simplify the code, we should probably not multiplex G2Path like this,
 	// and instead use a G2PrefixPath config. Then EITHER G2Path is used, OR both G2PrefixPath and G2TrailingPath are used.
-	G2Path         string
+	G2Path string `docs:"required"`
```
litt3: > // G2Path and G2TrailingPath are only needed if LoadG2Points is true.

This sentence in the doc above seems to indicate that the `required` tag here is incorrect

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,157 @@
...
+type EncoderConfig struct {
+	// Encoder version (1 or 2)
+	Version EncoderVersion
+
+	// Port at which encoder listens for gRPC calls (default: 34000)
+	GrpcPort string
+
+	Aws        aws.ClientConfig
+	BlobStore  blobstore.Config
+	ChunkStore chunkstore.Config
```
litt3, line 37: There are fields in `BlobStore` and `ChunkStore` marked `required`, that aren't being checked within `Verify` below. Presumably if they are `required`, we should throw an error if they're missing

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,157 @@
...
+	// MetricsPort is the port that the encoder metrics server listens on.
+	MetricsPort string
+	// EnableMetrics enables the metrics HTTP server for prometheus metrics collection
+	EnableMetrics bool
+
+	// LogFormat is the format of the logs: json or text
+	LogFormat string
+	// LogColor is a flag to enable color in the logs
+	LogColor bool
+	// LogLevel is the level of the logs: debug, info, warn, error
```
litt3, line 50: If I'm not mistaken, empty log level would result in a runtime error. It should probably be checked within `Verify`

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,157 @@
...
+		Version:  V2,
+		GrpcPort: "34000",
+		Aws: aws.ClientConfig{
+			Region: "",
+		},
+		BlobStore: blobstore.Config{
+			Backend: blobstore.S3Backend,
+		},
+		ChunkStore: chunkstore.Config{
+			Backend: string(blobstore.S3Backend),
```
litt3, line 82: Would there be a runtime error if `Backend` for `ChunkStore` and/or `BlobStore` was undefined? If so, they should probably be checked in `Verify`.

Or, alternatively, the checks for these structs should be moved to the config structs themselves. IMO this would be the ideal solution

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,157 @@
...
+			EnableGnarkChunkEncoding:       false,
+			PreventReencoding:              true,
+			Backend:                        string(encoding.GnarkBackend),
+			GPUEnable:                      false,
+			PprofHttpPort:                  "6060",
+			EnablePprof:                    false,
+		},
+		MetricsPort:   "9094",
+		EnableMetrics: true,
+		LogFormat:     string(common.JSONLogFormat),
```
litt3: `LogFormat` isn't being checked in `Verify`

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,157 @@
...
+			PreloadEncoder:  false,
+			Verbose:         false,
+		},
+		Server: ServerConfig{
+			MaxConcurrentRequestsDangerous: 16,
+			RequestPoolSize:                32,
+			RequestQueueSize:               32,
+			EnableGnarkChunkEncoding:       false,
+			PreventReencoding:              true,
+			Backend:                        string(encoding.GnarkBackend),
```
litt3: Here's another example of something that _must_ be defined, so it probably ought to be checked, either in `Verify` below, or in a new `Verify` method on the `ServerConfig` itself

I'll stop commenting these, since the gist is the same. 

Are you planning on doing a followup to properly handle all the fields in these nested configs? I want to avoid a situation where these nested structs are forgotten about, since they must each individually be correct for the `EncoderConfig` as a whole to be exhaustively checked for correctness

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,157 @@
...
+		"github.com/Layr-Labs/eigenda/common/aws",
+	}
+}
+
+// DefaultEncoderConfig returns a default configuration for the encoder.
+func DefaultEncoderConfig() *EncoderConfig {
+	return &EncoderConfig{
+		Version:  V2,
+		GrpcPort: "34000",
+		Aws: aws.ClientConfig{
```
litt3: There exists `DefaultClientConfig` function. That should probably be used instead of redefining defaults here, though it might need modification, if you don't find the values there to be suitable

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,157 @@
...
+			Backend:                        string(encoding.GnarkBackend),
+			GPUEnable:                      false,
+			PprofHttpPort:                  "6060",
+			EnablePprof:                    false,
+		},
+		MetricsPort:   "9094",
+		EnableMetrics: true,
+		LogFormat:     string(common.JSONLogFormat),
+		LogColor:      false,
+		LogLevel:      "debug",
```
litt3, line 106: Your PR description says you set the default log level to `info`, but here it's `debug`

IMO `info` is a much better default, since `debug` logs are intended only for `debug` situations

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,157 @@
...
+// DefaultEncoderConfig returns a default configuration for the encoder.
+func DefaultEncoderConfig() *EncoderConfig {
+	return &EncoderConfig{
+		Version:  V2,
+		GrpcPort: "34000",
+		Aws: aws.ClientConfig{
+			Region: "",
+		},
+		BlobStore: blobstore.Config{
+			Backend: blobstore.S3Backend,
```
litt3, line 79: Would it be permissible for `BlobStore.Backend` and `ChunkStore.Backend` to be different?

If not, consider checking that they are equal inside `Verify`

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,157 @@
...
+		},
+		ChunkStore: chunkstore.Config{
+			Backend: string(blobstore.S3Backend),
+		},
+		Kzg: kzg.KzgConfig{
+			SRSOrder:        268435456,
+			SRSNumberToLoad: 2097152,
```
litt3: I don't think `2097152` is correct. That translates to 64MiB, right? `16MiB` would be `524288 symbols`, I believe. @dmanc could you verify what makes sense here?

See [this thread for additional context](https://github.com/Layr-Labs/eigenda/pull/2351#discussion_r2577740649)

dmanc: That sounds right to me but in the config repo we're using 2097152 🤔. We should test 524288 in preprod and confirm it's still functional.

litt3: Hmm, weird.

~~In that case, I'd be fine leaving it like this. No big rush to fix it if it's already working.~~

@bxue-l2 Says it should be `524288`. I think it probably makes sense to fix it here now, and we can update other places later.

encoding/v1/kzg/kzgconfig.go
crypt0fairy: if I am getting it right, G2_PATH in devops repo is redundant:
https://github.com/Layr-Labs/eigenda-devops/pull/2183/files#diff-c23e5811452bd8e61fc48efc0b2b46b73d63e7a53b338876493b24c6a41bb5deR8

> ENCODER_KZG_CACHE_DIR: /data/kzg/SRSTables
    ENCODER_KZG_G1_PATH: /data/kzg/g1.point
    ENCODER_KZG_G2_PATH: /data/kzg/g2.point #has to be removed

disperser/encoder/encoder_config.go
crypt0fairy: The biggest confusion for me right now is that we have two `KzgConfig`s:
* `encoding/v1/kzg`
* `encoding/v2/kzg/prover`
Based on comment:
```
// The V1 KzgConfig is used all over the place in multiple different structs,
```
if I am going to add `Verify` it will touch other places.

disperser/encoder/encoder_config.go
crypt0fairy: `2097152` is based on values in devops repo. `2097152` is the same across all deployments. 
Honestly I don't like this notation, I would prefer (1<<21) which is more intuitive

disperser/encoder/encoder_config.go
crypt0fairy: Current default values for encoder are based on values in devops branch and follows two principles:
* default config should be close to mainnet configuration. This keeps config small enough to the users/operators
* It appears `debug` is same value across all deployment

disperser/encoder/encoder_config.go
crypt0fairy: `Blobstore` is used in 7 different places and are outside encoder's scope.
`Chunstore` in 2 places.
If to move checks to config structs themselves, I would do it in a followup PR.
I hope that different services expect same validation rules for the same structure, worst casE scenario is when `Verify` in `relay` and in `batcher` require different rules.

disperser/encoder/encoder_config.go
crypt0fairy: I don't have context on that. ChunkStore is also used in relay.

disperser/encoder/encoder_config.go
```diff
@@ -0,0 +1,157 @@
...
+		},
+		Server: ServerConfig{
+			MaxConcurrentRequestsDangerous: 16,
+			RequestPoolSize:                32,
+			RequestQueueSize:               32,
+			EnableGnarkChunkEncoding:       false,
+			PreventReencoding:              true,
+			Backend:                        string(encoding.GnarkBackend),
+			GPUEnable:                      false,
+			PprofHttpPort:                  "6060",
```
crypt0fairy, line 99: @cody-littley @litt3 shouldn't port be a number instead of a string?

disperser/encoder/encoder_config.go
crypt0fairy: loglevel is verified in `main.go `

disperser/encoder/encoder_config.go
litt3: Ideally, yes. We've been trying to move port types to `uint16` wherever possible, but there are still lots of places where it's a string.

Up to you whether you think it makes sense to put that change into this PR

disperser/encoder/encoder_config.go
litt3: > Honestly I don't like this notation, I would prefer (1<<21) which is more intuitive

I'd be on board with that. Ideally with a comment saying what that evaluates to, for the sake of completeness

> 2097152 is based on values in devops repo. 2097152 is the same across all deployments.

Changing the default in this struct won't change the values in the `devops` repo, so nothing will immediately change for the deployments. If @bxue-l2 says it ought to be `524288`, then I think it makes sense to do it.

The values in the `devops` repo are a good starting point, but if we identify any problems with the values there, this PR is our chance to pay off some tech debt, and have good defaults moving forward.

disperser/encoder/encoder_config.go
litt3: Right, and if it is empty there will be an error. 

The point of `Verify` is that you can call it and know that your config struct is well formed.

disperser/encoder/encoder_config.go
cody-littley: We should be able to add a `Verify()` method without disrupting other parts of the code. We can just not call `Verify()` in the other parts of the code until we get around to refactoring the configuration for those things.

disperser/encoder/encoder_config.go
cody-littley: There is a config flag `--only-verify-config` that if set causes the binary to check config then exit with a message saying if the config is good or bad. It does this by calling the `Verify()` method. Although we don't have this utility tied into our current deployment pipelines, it's on our TODO list.

At the end of the day, it's impossible to verify all config in the `Verify()` method. For example, it's tricky to know if a provided private key has proper authorization to perform necessary onchain actions. But we should make an attempt to do all simple and straightforward verification in `Verify()`, where possible.

encoding/v1/kzg/kzgconfig.go
cody-littley: I think @samlaf's comment above may have actually been premature. Until we've tested it without the G2Path, I think we should assume that it's a required argument. And once we test and verify that we don't need a G2Path, we should delete the flag from the config entirely.

 Perhaps we should edit the comment as follows.

```
// The file path to the G2 file.
//
// TODO(samlaf): to prevent misconfigurations and simplify the code, we should probably not multiplex G2Path like this,
// and instead use a G2PrefixPath config. Then EITHER G2Path is used, OR both G2PrefixPath and G2TrailingPath are used.
// Note that code changes are possibly required to make this into an optional argument. And once we do make it
 into an optional argument, we should probably clean up this struct and remove the necessary fields.
G2Path string `docs:"required"`
```

disperser/encoder/encoder_config.go
litt3: Let's have this discussion as a wider group, we can leave it as `debug` for now

disperser/encoder/encoder_config.go
cody-littley: The blobstore/chunkstore backend will never be different (short of massive code changes). It's safe to assume for the purposes of this PR that they will never be different.

encoding/v1/kzg/kzgconfig.go
dmanc: Adding some context: G2Path is not needed for V2 encoder but it's needed for V1 encoder

common/aws/cli.go
crypt0fairy: You are rights, I double checked the code where it is used. It defaults to what AWS provides.

disperser/encoder/encoder_config.go
crypt0fairy: I found at least one example where nested struct `Verify` doesn't work. We have `TableName` in blob config which is not used in encoder but is used in apiserver. If I create Verify suited for `encoder` only, later apiservers migration will not be complete.

disperser/encoder/encoder_config.go
crypt0fairy: Thant is the point, first followup PR is for devops repo as we can't deploy this without it. The goal is to make devops configuration as short as possible (mainnet as a main config). if `2097152` is same across all environments (main net, pre-prod, test-nets), why it should not be default.

litt3: The devops values are wrong, and should eventually be changed to this new value (according to @bxue-l2, who I'm deferring to as the expert here).

A good place to start defining the proper value is here, where we are defining the default. Then, we can remove the `devops` values entirely, and rely on this _correct_ default value

bxue-l2: Yes 524288 is the correct value, here is our encoder workflow. We only need G1 to compute BLOB commitment which is at most 16MiB. I guess we changed it to 2,097,152 before some SRS table aren't pre-generated, but it was fixed.

Screenshot 2025-12-10 at 10 11 18 AM

common/aws/cli.go
cody-littley: Resolving this issue since we decided to do it as a follow up

disperser/encoder/encoder_config.go
crypt0fairy: checked. Too many places to update that are outside the scope of the encoder.

PR #2352, opened 2025-11-24
feat: config improvements

cody-littley:

This PR improves several things related to the new config framework:

- fix a godoc parsing bug where a `}` character in a godoc broke the parser
- Do config validation for the traffic generator
- Add a flag that causes the framework to exist after doing config verification

### DIFF
--- a/common/config/bootstrap.go
+++ b/common/config/bootstrap.go
@@ -46,8 +46,8 @@ var (
 		Aliases: []string{"c"},
 		Usage:   "Path to a configuration file. Can be specified multiple times to load multiple files.",
 	}
-	verifyConfigFlag = &cli.BoolFlag{
-		Name:    "verify-config",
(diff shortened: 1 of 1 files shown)


common/config/bootstrap.go
```diff
@@ -131,6 +140,11 @@ func buildHandler[T DocumentedConfig](
 			return fmt.Errorf("failed to load configuration: %w", err)
 		}
 
+		if verifyConfig {
```
litt3: I'm confused what you're doing here. If `verifyConfig` is `true`, you exit the program? I'm missing something

cody-littley: Renamed flag to `--only-verify-config`. The idea is that you use this flag to see if your config is valid or not. If valid, the program exits with code 0. If non-valid, it exits with an error message and a non-zero return code. We can use this to sanity check configuration before we deploy a build to an environment.

PR #2353, opened 2025-11-24
feat(disperser): adds dispersal client request timestamp drift metrics

pschork:

Allows for tracking client timestamp drift
~~Adds delay config to dataapi feed endpoints~~

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/disperser/apiserver/metrics_v2.go
+++ b/disperser/apiserver/metrics_v2.go
@@ -35,6 +35,8 @@ type metricsV2 struct {
 	dispersalTimestampConfigMaxAge    prometheus.Gauge
 	dispersalTimestampConfigMaxFuture prometheus.Gauge
 
+	enablePerAccountMetrics bool
+
 	registry *prometheus.Registry
 	httpPort string
 	logger   logging.Logger
@@ -176,6 +178,7 @@ func newAPIServerV2Metrics(registry *prometheus.Registry, metricsConfig disperse
 		dispersalTimestampDrift:           dispersalTimestampDrift,
 		dispersalTimestampConfigMaxAge:    dispersalTimestampConfigMaxAge,
 		dispersalTimestampConfigMaxFuture: dispersalTimestampConfigMaxFuture,
+		enablePerAccountMetrics:           !metricsConfig.DisablePerAccountMetrics,
 		registry:                          registry,
 		httpPort:                          metricsConfig.HTTPPort,
 		logger:                            logger.With("component", "DisperserV2Metrics"),
@@ -236,7 +239,12 @@ func (m *metricsV2) reportDispersalTimestampRejected(reason string) {
 }
 
 func (m *metricsV2) reportDispersalTimestampDrift(driftSeconds float64, status string, accountID string) {
-	m.dispersalTimestampDrift.WithLabelValues(status, accountID).Observe(driftSeconds)
+	// If per-account metrics are disabled, aggregate under "0x0"
+	labelValue := accountID
+	if !m.enablePerAccountMetrics {
+		labelValue = "0x0"
+	}
+	m.dispersalTimestampDrift.WithLabelValues(status, labelValue).Observe(driftSeconds)
 }
 
 func (m *metricsV2) setDispersalTimestampConfig(maxAgeSeconds, maxFutureSeconds float64) {
--- a/disperser/cmd/apiserver/flags/flags.go
+++ b/disperser/cmd/apiserver/flags/flags.go
@@ -267,6 +267,12 @@ var (
 		Required: false,
 		EnvVar:   common.PrefixEnvVar(envVarPrefix, "DISABLE_GET_BLOB_COMMITMENT"),
 	}
+	DisablePerAccountMetricsFlag = cli.BoolFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "disable-per-account-metrics"),
+		Usage:    "Disables account level metrics collection (default: false)",
+		Required: false,
+		EnvVar:   common.PrefixEnvVar(envVarPrefix, "DISABLE_PER_ACCOUNT_METRICS"),
+	}
 )
 
 // Flags needed for computing kzg commitments.
@@ -336,6 +342,7 @@ var optionalFlags = []cli.Flag{
 	ControllerAddressFlag,
 	UseControllerMediatedPayments,
 	DisableGetBlobCommitment,
+	DisablePerAccountMetricsFlag,
 	OperatorStateRetrieverFlag,
 	EigenDAServiceManagerFlag,
 	EigenDADirectoryFlag,
--- a/disperser/cmd/apiserver/lib/config.go
+++ b/disperser/cmd/apiserver/lib/config.go
@@ -111,8 +111,9 @@ func NewConfig(ctx *cli.Context) (Config, error) {
 		},
 		LoggerConfig: *loggerConfig,
 		MetricsConfig: disperser.MetricsConfig{
-			HTTPPort:      ctx.GlobalString(flags.MetricsHTTPPort.Name),
-			EnableMetrics: ctx.GlobalBool(flags.EnableMetrics.Name),
+			HTTPPort:                 ctx.GlobalString(flags.MetricsHTTPPort.Name),
+			EnableMetrics:            ctx.GlobalBool(flags.EnableMetrics.Name),
+			DisablePerAccountMetrics: ctx.GlobalBool(flags.DisablePerAccountMetricsFlag.Name),
 		},
 		RatelimiterConfig:             ratelimiterConfig,
 		RateConfig:                    rateConfig,
--- a/disperser/metrics.go
+++ b/disperser/metrics.go
@@ -14,8 +14,9 @@ import (
 )
 
 type MetricsConfig struct {
-	HTTPPort      string
-	EnableMetrics bool
+	HTTPPort                 string
+	EnableMetrics            bool
+	DisablePerAccountMetrics bool
 }
 
 type Metrics struct {


disperser/apiserver/metrics_v2_test.go
```diff
@@ -0,0 +1,219 @@
...
+
+	metrics := newAPIServerV2Metrics(registry, metricsConfig, logger)
+
+	// Set config
+	maxAgeSeconds := 45.0
+	maxFutureSeconds := 5.0
+	metrics.setDispersalTimestampConfig(maxAgeSeconds, maxFutureSeconds)
+
+	// Simulate various timestamp scenarios
+	now := time.Now()
```
litt3: why are you creating this variable, just to suppress the `unused` warning below?

disperser/apiserver/metrics_v2.go
```diff
@@ -22,15 +22,18 @@ const namespace = "eigenda_disperser_api"
...
+	getPaymentStateLatency            *prometheus.SummaryVec
+	disperseBlobLatency               *prometheus.SummaryVec
+	disperseBlobSize                  *prometheus.CounterVec
+	disperseBlobMeteredBytes          *prometheus.CounterVec
+	validateDispersalRequestLatency   *prometheus.SummaryVec
+	storeBlobLatency                  *prometheus.SummaryVec
+	getBlobStatusLatency              *prometheus.SummaryVec
+	dispersalTimestampRejected        *prometheus.CounterVec
+	dispersalTimestampDrift           *prometheus.HistogramVec
+	dispersalTimestampConfigMaxAge    prometheus.Gauge
```
litt3, line 35: Is there a particular reason you're adding metrics for these config values? This isn't a pattern that we seem to follow in general

disperser/dataapi/v2/feed_delay_test.go
```diff
@@ -0,0 +1,82 @@
...
+	if blobMetadataStore == nil {
+		t.Skip("Blob metadata store is not initialized")
+	}
+
+	now := time.Now()
+	feedDelay := 10 * time.Second
+	router := testDataApiServerV2
+
+	// Test that beforeTime is properly capped
+	_, err := http.NewRequest("GET", "/v2/blobs/feed", nil)
```
litt3: What is this `NewRequest` call testing?

disperser/dataapi/v2/feed_delay_test.go
```diff
@@ -0,0 +1,82 @@
...
+	// Verify response
+	require.Equal(t, http.StatusOK, rr.Code)
+
+	var response serverv2.BlobFeedResponse
+	err = json.Unmarshal(rr.Body.Bytes(), &response)
+	require.NoError(t, err)
+
+	// Verify that all returned blobs have RequestedAt timestamp older than (now - feedDelay)
+	minTime := now.Add(-feedDelay)
+	for _, blob := range response.Blobs {
```
litt3: Are there actually any blobs to iterative over here? I'm not seeing anywhere they are being created

disperser/apiserver/metrics_v2_test.go
litt3, line 1: I'm assuming these tests were claude generated?

IMO there's not a lot of value here. These tests seem to pretty much be asserting that the prometheus objects work correctly

disperser/apiserver/metrics_v2.go
pschork: I'm adding gauges that will represent the current config limits. I wish more config values were exposed as gauges for better dashboard observability.

disperser/apiserver/metrics_v2_test.go
pschork: Removed

disperser/dataapi/v2/feed_delay_test.go
pschork: Fixed

disperser/dataapi/v2/feed_delay_test.go
pschork: refactored

disperser/apiserver/metrics_v2.go
pschork: image

disperser/apiserver/metrics_v2.go
```diff
@@ -133,20 +136,49 @@ func newAPIServerV2Metrics(registry *prometheus.Registry, metricsConfig disperse
...
 	)
 
+	dispersalTimestampDrift := promauto.With(registry).NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: namespace,
+			Name:      "dispersal_timestamp_drift_seconds",
+			Help:      "Distribution of timestamp drift for dispersal requests. Negative values indicate timestamps in the future, positive values indicate timestamps in the past.",
+			Buckets:   []float64{-60, -30, -10, -5, -1, -0.5, 0, 0.5, 1, 2, 5, 10, 30, 60, 120, 300},
+		},
+		[]string{"status", "account_id"}, // "accepted" or "rejected", account address
```
dmanc, line 148: Account id going to be high cardinality assuming we're tracking arbitrary customers. Maybe it's fine for now though?

litt3: For some other high cardinality metrics, we've been adding a flag to optionally disable, just in case it ever becomes a problem that needs rapid resolution. Potentially could be done here as well

disperser/cmd/dataapi/flags/flags.go
```diff
@@ -153,6 +153,13 @@ var (
 		Value:    1,
 		EnvVar:   common.PrefixEnvVar(envVarPrefix, "DATA_API_VERSION"),
 	}
+	FeedDelayFlag = cli.DurationFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "feed-delay"),
+		Usage:    "Minimum age (duration) a blob/batch must have before appearing in feed endpoints. This prevents replay attacks by delaying public visibility of dispersal requests.",
+		Required: false,
+		Value:    0,
+		EnvVar:   common.PrefixEnvVar(envVarPrefix, "FEED_DELAY"),
+	}
```
cody-littley: A note to external observers: the attack discussed in this thread is only plausible with multiple dispersers, and since there is currently only one disperser, this attack is not currently possible.

It might be worth having a discussion as to whether or not adding this delay will be sufficient to mitigate the attack against multiple dispersers. Anybody running a validator can be used to leak this signature information. Since this doesn't require significant stake, and since the stake can be later withdrawn by the attacker, this is not a very significant barrier. IMO, this vulnerability would still require the "anchored dispersal" fix to be closed, meaning we'd still need to wait in order to deploy additional dispersers.

With this in mind, is it still worthwhile to add a feed delay? If this doesn't fully close the vulnerability, then extra latency added to the data feed pipeline doesn't really benefit us.

disperser/apiserver/metrics_v2.go
```diff
@@ -133,20 +136,49 @@ func newAPIServerV2Metrics(registry *prometheus.Registry, metricsConfig disperse
 		[]string{"reason"},
 	)
 
+	dispersalTimestampDrift := promauto.With(registry).NewHistogramVec(
+		prometheus.HistogramOpts{
+			Namespace: namespace,
+			Name:      "dispersal_timestamp_drift_seconds",
+			Help:      "Distribution of timestamp drift for dispersal requests. Negative values indicate timestamps in the future, positive values indicate timestamps in the past.",
+			Buckets:   []float64{-60, -30, -10, -5, -1, -0.5, 0, 0.5, 1, 2, 5, 10, 30, 60, 120, 300},
```
cody-littley, line 146: Our of curiosity, why are you reporting this data as a histogram instead of reporting it as a summary vec? Is the exact count of dispersals in each bucket important, or would average/min/max/p99/etc. be sufficient? In all other places that I am aware, we typically use summary vecs to report latencies.

disperser/apiserver/metrics_v2.go
```diff
@@ -22,15 +22,18 @@ const namespace = "eigenda_disperser_api"
...
+	disperseBlobLatency               *prometheus.SummaryVec
+	disperseBlobSize                  *prometheus.CounterVec
+	disperseBlobMeteredBytes          *prometheus.CounterVec
+	validateDispersalRequestLatency   *prometheus.SummaryVec
+	storeBlobLatency                  *prometheus.SummaryVec
+	getBlobStatusLatency              *prometheus.SummaryVec
+	dispersalTimestampRejected        *prometheus.CounterVec
+	dispersalTimestampDrift           *prometheus.HistogramVec
+	dispersalTimestampConfigMaxAge    prometheus.Gauge
+	dispersalTimestampConfigMaxFuture prometheus.Gauge
```
cody-littley, line 36:
```suggestion
	dispersalTimestampConfigMaxAge    *prometheus.Gauge
	dispersalTimestampConfigMaxFuture *prometheus.Gauge
```

disperser/apiserver/metrics_v2.go
pschork: HistogramVec gives you buckets (for heatmaps) and you can convert to quantiles in grafana query to get a latency time series.
image

disperser/apiserver/metrics_v2.go
pschork: Non-Vec types like Gauge are created by NewGauge(), which returns an interface (prometheus.Gauge), not a pointer to a concrete type. Interfaces in Go are reference types already, so they don't need an additional pointer indirection.

disperser/apiserver/metrics_v2.go
pschork: Good idea. I added a flag to disable per account metrics.

disperser/cmd/dataapi/flags/flags.go
pschork: Dropped feed delay

PR #2356, opened 2025-11-25
feat: config aliases

cody-littley:

Adds the ability to alias environment variables. Although not needed for internal disperser components, for public facing components like validators and the proxy it's super important. This will enable us to migrate to the new config framework without breaking existing user's config files.

### DIFF
--- a/common/config/config_parser.go
+++ b/common/config/config_parser.go
@@ -52,13 +52,13 @@ func ParseConfig[T VerifiableConfig](
 		}
 	}
 
-	err := aliasEnvVars(logger, aliasedEnvVars)
-	if err != nil {
-		var zero T
-		return zero, fmt.Errorf("failed to alias environment variables: %w", err)
-	}
-
 	if envPrefix != "" {
+		err := aliasEnvVars(logger, aliasedEnvVars)
+		if err != nil {
+			var zero T
+			return zero, fmt.Errorf("failed to alias environment variables: %w", err)
+		}
+
 		viperInstance.SetEnvPrefix(envPrefix)
 		viperInstance.SetEnvKeyReplacer(strings.NewReplacer(".", "_"))
 		viperInstance.AutomaticEnv()
@@ -288,6 +288,9 @@ func checkForInvalidEnvVars(
 	for _, v := range ignoredEnvVars {
 		ignoredSet[v] = struct{}{}
 	}
+	// The config parser will return an error if it discovers an environment variable that doesn't map to a struct
+	// value. Since the aliased environment variables indirectly map to struct values, we need to instruct the config
+	// parser to ignore them when it's checking for un-mapped environment variables.
 	for k := range aliasedEnvVars {
 		ignoredSet[k] = struct{}{}
 	}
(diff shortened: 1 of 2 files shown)


common/config/config_parser.go
```diff
@@ -90,6 +107,22 @@ func ParseConfig[T VerifiableConfig](
 	return cfg, nil
 }
 
+// Applies environment variable aliases by copying the value of each aliased variable to its target variable.
+// This function sets new environment variables using os.Setenv if old environment variables in need of
+// aliasing are set.
+func aliasEnvVars(aliasedEnvVars map[string]string) error {
+	for oldVar, newVar := range aliasedEnvVars {
+		value, exists := os.LookupEnv(oldVar)
+		if exists {
```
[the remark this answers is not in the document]
cody-littley: Good point. Fixed.

test/v2/load/main/load_main.go
```diff
@@ -15,8 +15,11 @@ func main() {
 
 	cfg, err := config.Bootstrap(
 		load.DefaultTrafficGeneratorConfig,
-		"TRAFFIC_GENERATOR_SIGNER_PRIVATE_KEY_HEX",
-		"TRAFFIC_GENERATOR_RPC_URLS",
+		nil,
+		[]string{
```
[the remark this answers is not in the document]
cody-littley: nah

common/config/config_parser.go
```diff
@@ -245,6 +288,9 @@ func checkForInvalidEnvVars(
 	for _, v := range ignoredEnvVars {
 		ignoredSet[v] = struct{}{}
 	}
+	for k := range aliasedEnvVars {
```
litt3, line 294: could you leave a comment about why it makes sense to ignore these?

I think I understand the logic, but I'm not sure the answer is immediately obvious

common/config/config_parser.go
```diff
@@ -41,6 +52,12 @@ func ParseConfig[T VerifiableConfig](
 		}
 	}
 
+	err := aliasEnvVars(logger, aliasedEnvVars)
```
litt3: Shouldn't `aliasEnvVars` be called within the block below?`if envPrefix != "" {`

common/config/config_test.go
```diff
@@ -401,7 +401,72 @@ func TestEnvironmentVariables(t *testing.T) {
...
+
+	// Baz field - the env vars use FOO_BAZ_PARTIAL_* which doesn't match foo.Baz,
+	// so these should come from config
+	require.NotNil(t, foo.Baz)
+	require.Equal(t, "baz X", foo.Baz.X) // from config
+	require.Equal(t, 27, foo.Baz.Y)      // from config
+	require.Equal(t, true, foo.Baz.Z)    // from config
+}
+
+func TestAliasedEnvironmentVariables(t *testing.T) {
```
litt3, line 443: Do you have a test anywhere for when both the old and new vars are set? I don't see one

crypt0fairy (commented):
What would be an example of how to use this feature.

common/config/config_parser.go
cody-littley:
```
	// The config parser will return an error if it discovers an environment variable that doesn't map to a struct
	// value. Since the aliased environment variables indirectly map to struct values, we need to instruct the config
	// parser to ignore them when it's checking for un-mapped environment variables.
	for k := range aliasedEnvVars {
		ignoredSet[k] = struct{}{}
	}
```

common/config/config_parser.go
cody-littley: Good point, fixed. No point in touching environment variables if we are not going to be reading them.

common/config/config_test.go
cody-littley: added a test

PR #2358, opened 2025-11-25
feat: periodically flush signing rate data to disk, load at startup

cody-littley:

Periodically flush signing rate data to dynamo, load it at startup time

### DIFF
--- a/disperser/cmd/controller/flags/flags.go
+++ b/disperser/cmd/controller/flags/flags.go
@@ -356,9 +356,8 @@ var (
 	SigningRateDynamoDbTableNameFlag = cli.StringFlag{
 		Name:     common.PrefixFlag(FlagPrefix, "signing-rate-dynamodb-table-name"),
 		Usage:    "The name of the DynamoDB table used to store signing rate data",
-		Required: false,
+		Required: true,
 		EnvVar:   common.PrefixEnvVar(envVarPrefix, "SIGNING_RATE_DYNAMODB_TABLE_NAME"),
-		Value:    "validator-signing-rates",
 	}
 )
 
@@ -368,11 +367,11 @@ var requiredFlags = []cli.Flag{
 	EncodingPullIntervalFlag,
 	AvailableRelaysFlag,
 	EncoderAddressFlag,
-
 	DispatcherPullIntervalFlag,
 	AttestationTimeoutFlag,
 	BatchAttestationTimeoutFlag,
 	DisperserIDFlag,
+	SigningRateDynamoDbTableNameFlag,
 }
 
 var optionalFlags = []cli.Flag{
@@ -416,7 +415,6 @@ var optionalFlags = []cli.Flag{
 	SigningRateRetentionPeriodFlag,
 	SigningRateBucketSpanFlag,
 	SigningRateFlushPeriodFlag,
-	SigningRateDynamoDbTableNameFlag,
 }
 
 var Flags []cli.Flag
--- a/inabox/tests/setup_disperser_harness.go
+++ b/inabox/tests/setup_disperser_harness.go
@@ -803,7 +803,7 @@ func startController(
 	dispatcherConfig := controller.DefaultDispatcherConfig()
 	dispatcherConfig.FinalizationBlockDelay = 5
 	dispatcherConfig.BatchMetadataUpdatePeriod = 100 * time.Millisecond
-	dispatcherConfig.SigningRateDynamoDbTableName = "signing-rate-tracker"
+	dispatcherConfig.SigningRateDynamoDbTableName = "validator-signing-rates"
 
 	// Chain state config
 	chainStateConfig := thegraph.Config{


disperser/cmd/controller/main.go
```diff
@@ -242,6 +238,36 @@ func RunController(cliCtx *cli.Context) error {
...
+	}
+
+	// Periodically flush signing rate data to persistent storage.
+	go signingrate.SigningRateStorageFlusher(
+		ctx,
+		logger,
+		signingRateTracker,
+		signingRateStorage,
+		config.DispatcherConfig.SigningRateFlushPeriod,
+	)
```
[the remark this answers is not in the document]
cody-littley: We have no mechanism for a clean shutdown, so it's pointless. As a part of this design, I decided that it was ok to have small amounts of signing rate data loss.

core/signingrate/signing_rate_flusher.go
```diff
@@ -0,0 +1,47 @@
...
+			if len(buckets) == 0 {
+				// nothing to flush
+				continue
+			}
+
+			err = storage.StoreBuckets(ctx, buckets)
+			if err != nil {
+				logger.Errorf("Error storing signing rate buckets: %v", err)
+				continue
+			}
```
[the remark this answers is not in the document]
cody-littley: `GetUnflushedBuckets()` already tracks this.

disperser/controller/controller_config.go
```diff
@@ -93,6 +99,8 @@ func DefaultDispatcherConfig() *ControllerConfig {
 		MaxDispersalAge:                     45 * time.Second,
 		SigningRateRetentionPeriod:          14 * 24 * time.Hour, // 2 weeks
 		SigningRateBucketSpan:               10 * time.Minute,
+		SigningRateFlushPeriod:              1 * time.Minute,
+		SigningRateDynamoDbTableName:        "validator-signing-rates",
```
litt3: All over the place, you're defining this default value. But in practice, this will never actually be the name of a dynamo table on any network. The real table name includes region specific stuff, etc.

With that in mind, I don't think we should be setting a default value, here or elsewhere. Any actual usage of that default value would be wrong

core/signingrate/signing_rate_flusher.go
```diff
@@ -0,0 +1,47 @@
...
+				logger.Errorf("Error getting unflushed buckets: %v", err)
+				continue
+			}
+
+			if len(buckets) == 0 {
+				// nothing to flush
+				continue
+			}
+
+			err = storage.StoreBuckets(ctx, buckets)
```
litt3, line 40: Is it a problem that if storage fails, all the buckets you were trying to flush are permanently lost?

core/signingrate/signing_rate_flusher.go
```diff
@@ -0,0 +1,47 @@
...
+	flushPeriod time.Duration,
+) {
+
+	ticker := time.NewTicker(flushPeriod)
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return
```
litt3, line 27: Would it make sense to do a final flush before shutdown?

disperser/controller/controller_config.go
cody-littley: Made this into a required config with no default, will update devops repo prior to merge (so don't resolve this comment please 😅)

core/signingrate/signing_rate_flusher.go
cody-littley: Not a problem, this was intentionally part of the design. This data is used for ejection decisions and for external looky loos monitoring EigenDA health, and so perfect data retention is not required. More important to be highly performant and simple.

core/signingrate/signing_rate_flusher.go
cody-littley: Possibly, but I'm not sure it's worth doing. On shutdown the context will be cancelled, so the flush will fail unless we use a special shutdown context (and we'd have to manage the lifecycle of that context so that it can't block shutdown indefinitely). Also, in production environments, there is no clean shutdown, we just power down the kubernetes pod, meaning the final flush would only ever get executed in tests.

litt3 (approved):

disperser/cmd/controller/flags/flags.go
```diff
@@ -346,6 +346,20 @@ var (
...
+		Name:     common.PrefixFlag(FlagPrefix, "signing-rate-flush-period"),
+		Usage:    "The period at which signing rate data is flushed to persistent storage",
+		Required: false,
+		EnvVar:   common.PrefixEnvVar(envVarPrefix, "SIGNING_RATE_FLUSH_PERIOD"),
+		Value:    1 * time.Minute,
+	}
+	SigningRateDynamoDbTableNameFlag = cli.StringFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "signing-rate-dynamodb-table-name"),
+		Usage:    "The name of the DynamoDB table used to store signing rate data",
+		Required: false,
```
pschork: seems like this should be required flag. If not specified controller will fail to start when it can't connect to default `validator-signing-rates` table

inabox/tests/setup_disperser_harness.go
```diff
@@ -803,6 +803,7 @@ func startController(
 	dispatcherConfig := controller.DefaultDispatcherConfig()
 	dispatcherConfig.FinalizationBlockDelay = 5
 	dispatcherConfig.BatchMetadataUpdatePeriod = 100 * time.Millisecond
+	dispatcherConfig.SigningRateDynamoDbTableName = "signing-rate-tracker"
```
pschork:
```suggestion
	dispatcherConfig.SigningRateDynamoDbTableName = "validator-signing-rates"
```

disperser/cmd/controller/flags/flags.go
cody-littley: made required

inabox/tests/setup_disperser_harness.go
cody-littley: renamed

litt3 (approved):

PR #2364, opened 2025-11-26
fix(specs): MD Book build error

ethenotethan:
- MD Book v0.5 was released and introduced breaking changes to legacy `toml` keys and is unsupported by the existing MD Book third party processors that we utilize. Fixed this by hard pinning the version dependency to [v0.4.52](http://github.com/rust-lang/mdBook/releases/tag/v0.4.52)
 - Builds have been [failing](http://github.com/Layr-Labs/eigenda/actions/workflows/mdbook.yaml) silently for over a week now. To avoid regressions like this happening in the future, I've broken the md book CI into two jobs (i.e, `publish`, `test`) where `test` is ran on every PR to ensure that builds are passing before being promoted to master where the `publish` action is ran.

## Checks

- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/.github/workflows/mdbook-test.yaml
+++ b/.github/workflows/mdbook-test.yaml
@@ -1,3 +1,5 @@
+name: Test Spec MdBook
+
 on:
   pull_request:
   merge_group:


.github/workflows/mdbook-test.yaml
samlaf, line 1: nit: personally prefer having both the build and publish jobs in the same yaml file.
Otherwise our CI files get out of hand and hard to keep them consistent.

You can gate the publish job as such:
```
  publish:
    runs-on: ubuntu-latest
    if: github.event_name == 'push' && github.ref == 'refs/heads/master'
```

ethenotethan: I had that thought but felt like we'd wanna add additional testing around the specs in-which case it'd no longer be a subset of the publish and job and would diverge

PR #2366, opened 2025-11-26
feat(clients): Create network address utility struct for `DisperserClient`

litt3:
- Creates a new `NetworkAddress` utility struct, and uses it in the `DisperserClientConfig`
- This is helpful cleanup in preparation for client disperser multiplexing
-  This new utility will be useful moving forward, to standardize how we handle network addresses


samlaf (approved):
LGTM as a cleanup/centralization of our current code (afaiu this is not introducing any external/config changes right?).

But overall as I mentioned in my comment on slack, think it would be even best to depart from having separate config fields for hostname and port, and instead start using URIs the way grpc is designed (https://grpc.io/docs/guides/custom-name-resolution/). This would give maximal future flexibility imo, to use grpc's current, or add new add custom name resolvers.

PR #2368, opened 2025-11-26
feat(clients): Simplify DisperserClient config to just use `NetworkAddress` instead of `hostname` and `port`

litt3:
- Originally I made [a util](https://github.com/Layr-Labs/eigenda/pull/2366) to handle hostname:port
- @samlaf suggested we just store the whole thing as a string, which I think makes sense. So I've done that in this PR

### DIFF
--- a/common/config/config_parser.go
+++ b/common/config/config_parser.go
@@ -7,6 +7,7 @@ import (
 	"slices"
 	"strings"
 
+	"github.com/Layr-Labs/eigenda/common/config/secret"
 	"github.com/Layr-Labs/eigenda/litt/util"
 	"github.com/Layr-Labs/eigensdk-go/logging"
 	"github.com/go-viper/mapstructure/v2"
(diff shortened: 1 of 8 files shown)


cody-littley (dismissed):
Reminder to update the devops repo for this flag change

litt3:
> Reminder to update the devops repo for this flag change

@cody-littley I'm not changing any flags here. What update are you saying is necessary in the `devops` repo?

api/clients/v2/dispersal/disperser_client.go
```diff
@@ -22,8 +22,7 @@ import (
 const maxNumberOfConnections = 32
 
 type DisperserClientConfig struct {
-	Hostname          string
-	Port              string
+	NetworkAddress    string
```
samlaf: my personal preference would be to explicitly name this `grpcServiceURI` or `grpcURI` which would make it explicit that it's going to get parsed following grpc's protocol semantics (and for example allows werd name schema/resolutions, but not `http://` in the beginning).

samlaf (dismissed):

api/clients/v2/dispersal/disperser_client.go
litt3: 4db42757c776c65027370bf597f84d96bf6db750

PR #2369, opened 2025-11-26
feat(disperser-registry): Create `DisperserRegistry` interface and initial implementation

litt3:
- We are in the process of creating a new `DisperserRegistry` contract
- The interface in this PR is what we'll use to interact with that contract
- Since the contract isn't ready yet, I created a `legacy` implementation, that works with the existing single-disperser networks. That way, I can move forward with the multiplexing work without waiting for the contract to be done

### DIFF
--- a/api/clients/v2/dispersal/disperser_client.go
+++ b/api/clients/v2/dispersal/disperser_client.go
@@ -22,8 +22,7 @@ import (
 const maxNumberOfConnections = 32
 
 type DisperserClientConfig struct {
-	Hostname          string
-	Port              string
+	GrpcUri           string
 	UseSecureGrpcFlag bool
 	// The number of grpc connections to the disperser server. A value of 0 is treated as 1.
 	DisperserConnectionCount uint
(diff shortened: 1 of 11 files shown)


common/disperser/disperser_registry.go
```diff
@@ -0,0 +1,13 @@
+package disperser
+
+import "context"
+
+// DisperserRegistry provides access to disperser information from the DisperserRegistry contract.
+type DisperserRegistry interface {
+	// Returns the set of dispersers that network participants should interact with by default.
+	GetDefaultDispersers(ctx context.Context) (map[uint32]struct{}, error)
+	// Returns the set of dispersers that support on-demand payments
+	GetOnDemandDispersers(ctx context.Context) (map[uint32]struct{}, error)
```
cody-littley: Any particular reason why you return sets instead of arrays? Are there queries where having an O(1) membership test is important?

litt3: As we discussed offline, slices do make sense. Fixed.

PR #2370, opened 2025-11-27
chore(rust)!: fix verify_and_extract_payload interface + docs

samlaf:
Simplify the interface, plus document it better. Had a lot of trouble when I tried to actually integrate it into the integration test.

Documenting the things that weren't obvious to me (like why cert_state and encoded_payload are optional), as well as why the function returns an Option<Result> instead of just Result.

Think just returning a Result with an error that would implement a `safe_to_discard_cert()` function might be cleaner, but won't break the API too too much just yet. :D

Also really thinking following bowen's push that we might want to rename the high-level entry function verify_and_extract_payload to just derive_payload() or something.

### DIFF
--- a/rust/crates/eigenda-verification/src/verification/mod.rs
+++ b/rust/crates/eigenda-verification/src/verification/mod.rs
@@ -113,7 +113,7 @@ pub fn extract_certificate(
 /// 1. Decoded payload as [`Bytes`] if verification succeeds
 /// 2. `None` to mean that this cert/encoded-payload should be discarded, since
 ///   one of the verification steps failed in a way that means "cert is invalid"
-/// 3. `Some(Err(_))` if some unexpected error occured that most likely means a bug in the
+/// 3. `Some(Err(_))` if some unexpected error occurred that most likely means a bug in the
 ///  state extraction or proof extraction logic. If this error occurs, the rollup's derivation
 ///  pipeline is NOT safe to continue, as the cert's validity could not be determined.
 ///


rust/crates/eigenda-verification/src/verification/mod.rs
```diff
@@ -85,28 +85,37 @@ pub fn extract_certificate(
...
+/// * `encoded_payload` - Optional encoded payload to verify
+/// Note that the cert_state and encoded_payload are only optional in case where the input
+/// can be safely rejected before needing them (e.g. recency check fails).
 ///
 /// # Returns
-/// Decoded blob data as [`Bytes`]
+/// 1. Decoded payload as [`Bytes`] if verification succeeds
+/// 2. `None` to mean that this cert/encoded-payload should be discarded, since
+///   one of the verification steps failed in a way that means "cert is invalid"
+/// 3. `Some(Err(_))` if some unexpected error occured that most likely means a bug in the
```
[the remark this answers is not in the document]
samlaf: [245670a](https://github.com/Layr-Labs/eigenda/pull/2370/commits/245670a615e62191554b66434e1ce65563291f7d)

rust/crates/eigenda-verification/src/verification/mod.rs
```diff
@@ -118,24 +127,24 @@ pub fn extract_certificate(
...
     inclusion_height: u64,
     referenced_height: u64,
     cert_recency_window: u64,
-    encoded_payload: &Option<Bytes>,
+    encoded_payload: Option<&[u8]>,
 ) -> Option<Result<Bytes, EigenDaVerificationError>> {
     use EigenDaVerificationError::*;
 
-    // if certificate recency verification fails: ignore
+    // if certificate recency verification fails: safe to discard
```
bxue-l2, line 139: In the future, proxy is going to switch the order of checking to have recency later

bxue-l2:
> Think just returning a Result with an error that would implement a safe_to_discard_cert() function might be cleaner

How about implementing something like hokulea, https://github.com/Layr-
Labs/hokulea/blob/a465144d9b3fc5202898b320ed7573c5e5922ac5/crates/eigenda/src/errors.rs#L8 . Then define `impl From<HokuleaStatelessError> for HokuleaErrorKind` ...

rust/crates/eigenda-verification/src/verification/mod.rs
samlaf: why? what would be the order then?

bxue-l2: because it is more natural to first know the derivation is correct, then use that to check recency. Though effectively they are the same

samlaf: but if cert is very not recent then you can't even download the blob so you won't be able to check the derivation is correct..

PR #2371, opened 2025-12-01
feat: api server signing rate

cody-littley:

Give the API server the ability to answer signing rate queries.

### DIFF
--- a/inabox/tests/setup_disperser_harness.go
+++ b/inabox/tests/setup_disperser_harness.go
@@ -904,8 +904,8 @@ func startController(
 
 	signingRateTracker, err := signingrate.NewSigningRateTracker(
 		controllerLogger,
-		10*time.Minute,
 		1*time.Minute,
+		1*time.Second,
 		time.Now)
 	signingRateTracker = signingrate.NewThreadsafeSigningRateTracker(ctx, signingRateTracker)
 
@@ -1273,8 +1273,8 @@ func startAPIServer(
 
 	signingRateTracker, err := signingrate.NewSigningRateTracker(
 		apiServerLogger,
-		10*time.Minute,
 		1*time.Minute,
+		1*time.Second,
 		time.Now)
 	signingRateTracker = signingrate.NewThreadsafeSigningRateTracker(ctx, signingRateTracker)
 


core/signingrate/signing_rate_mirroring.go
```diff
@@ -0,0 +1,82 @@
...
+
+	previousScrapeTime := time.Now()
+
+	for {
+		select {
+		case <-ctx.Done():
+			logger.Info("Stopping signing rate mirroring")
+			return
+		case <-ticker.C:
+			buckets, err := scraper(ctx, previousScrapeTime)
```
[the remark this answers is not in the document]
cody-littley: fixed

inabox/tests/setup_disperser_harness.go
```diff
@@ -1287,6 +1287,7 @@ func startAPIServer(
 		controllerConnection,
 		controllerClient,
 		listener,
+		signingrate.NewNoOpSigningRateTracker(),
```
litt3: Is there a specific reason you don't want to test this in inabox? Seems like it should be tested

core/signingrate/signing_rate_mirroring.go
```diff
@@ -0,0 +1,86 @@
...
+// A function that fetches signing rate data from some source starting from the given time.
+type SigningRateScraper func(ctx context.Context, startTime time.Time) ([]*validator.SigningRateBucket, error)
+
+// Do an initial scrape of signing rate data from a remote source and ingest it into the given tracker.
+// This makes it so that external callers never view an empty tracker at startup.
+func DoInitialScrape(
+	ctx context.Context,
+	logger logging.Logger,
+	// A function that can fetch signing rate data from some source.
+	scraper SigningRateScraper,
```
litt3, line 21: nit: Since this is a _function_ I think a better name for the parameter would be `scrape`. That way the usage reads like a function:

```
buckets, err := scrape(ctx, startTime)
```

disperser/cmd/apiserver/lib/apiserver.go
```diff
@@ -176,14 +183,62 @@ func RunDisperserServer(ctx *cli.Context) error {
...
+		// Clone signing rate data from controller. This is blocking, so that when we start the server we have
+		// data to serve right away.
+		err = signingrate.DoInitialScrape(
+			context.Background(),
+			logger,
+			scraper,
+			signingRateTracker,
+			config.ServerConfig.SigningRateRetentionPeriod)
+		if err != nil {
+			return fmt.Errorf("failed to do initial scrape of signing rate data from controller: %w", err)
```
litt3: bikeshed (no change necessary): Here is a case where the argument for omitting redundant wording like "failed to" is clear. As currently written, the error returned would read:

> failed to do initial scrape of signing rate data from controller: failed to do initial scrape of signing rate data: <...>

I think it's much better to keep error messages to a succinct minimum, since it reduces clutter. Personally, I'd do the following:

> do initial scrape: scrape: <...>

The "failed to" part is implicit, since this is an error. And each error substring only provides the context that's necessary for the final unpacked error to paint a clear picture

litt3 (approved):

inabox/tests/setup_disperser_harness.go
cody-littley: I was mostly just trying to avoid breaking things, but you make a good point. I've now added real implementations for the inabox suite.

core/signingrate/signing_rate_mirroring.go
cody-littley: renamed

disperser/cmd/apiserver/lib/apiserver.go
cody-littley: edited

I wish there was a compiler supported feature that would just resolve to the string `apiserver.go:219` (like a C macro).

inabox/tests/setup_disperser_harness.go
```diff
@@ -902,6 +902,13 @@ func startController(
 	}
 	dispatcherBlobSet := controller.NewBlobSet()
 
+	signingRateTracker, err := signingrate.NewSigningRateTracker(
+		controllerLogger,
+		10*time.Minute,
```
litt3: You might consider some shorter timespans for inabox tests, given the tests generally tend to be on the order of dozens of seconds, as opposed to minutes.

cody-littley: Changed to one minute and one second, respectively.

PR #2373, opened 2025-12-01
fix: relative path contract imports

8sunyuan:

The new stable version of foundry v1.5.0 is having build issues with absolute import definitions of contracts.
This refactors the DA contracts to use relative import paths and also the openzeppelin imports to use the @openzeppelin remapping format for consistency and to fix downstream submodule usage.

Ensured contracts still build and tests pass

## Checks

- [X] I've made sure the tests are passing. Note that there might be a few flaky tests, in that case, please comment that they are not relevant.
- [X] I've checked the new test coverage and the coverage percentage didn't drop.
- Testing Strategy
   - [ ] Unit tests
   - [ ] Integration tests

### DIFF
--- a/contracts/src/integrations/cert/legacy/v2/EigenDACertVerificationV2Lib.sol
+++ b/contracts/src/integrations/cert/legacy/v2/EigenDACertVerificationV2Lib.sol
@@ -1,15 +1,15 @@
 // SPDX-License-Identifier: MIT
 pragma solidity ^0.8.9;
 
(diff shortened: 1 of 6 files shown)


samlaf:
> @samlaf Wasn't there pushback around modifying imports? If so why?

From who? Not that I know..
I tend to be against relative imports just because they are more annoying to move files around while refactoring. Would prefer if we could use the @ syntax with a remappings file everywhere. But not strongly opinionated about contract stuff either. If this works for you guys then fine with me.

PR #2377, opened 2025-12-02
feat: handle slices of secrets in config

cody-littley:

Augment config parsing to handle `[]*secret.Secret` types. Needed to support configs that take a list of eth rpc urls.


common/config/secret/secret_parser.go
```diff
@@ -23,6 +24,33 @@ var DecodeHook mapstructure.DecodeHookFunc = func(from reflect.Type, to reflect.
...
+			default:
+				// If it's not a string or []byte, let mapstructure handle it normally
+				return data, nil
+			}
+
+			// Split the string by commas and create a slice of secrets
+			parts := strings.Split(sourceStr, ",")
+			secrets := make([]*Secret, len(parts))
+			for i, part := range parts {
+				secrets[i] = NewSecret(part)
```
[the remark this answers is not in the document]
cody-littley: good idea

common/config/secret/secret_parser.go
```diff
@@ -23,6 +24,33 @@ var DecodeHook mapstructure.DecodeHookFunc = func(from reflect.Type, to reflect.
...
+			switch v := data.(type) {
+			case string:
+				sourceStr = v
+			case []byte:
+				sourceStr = string(v)
+			default:
+				// If it's not a string or []byte, let mapstructure handle it normally
+				return data, nil
+			}
+
```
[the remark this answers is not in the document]
cody-littley: done

common/config/config_test.go
```diff
@@ -94,6 +95,7 @@ func TestTOMLParsing(t *testing.T) {
 	require.Equal(t,
 		"you're no stranger to love, you know the rules and so do I (so do I)",
 		foo.ThisIsASecretField.Get())
+	require.Equal(t, 0, len(foo.ThisIsASliceOfSecrets))
```
litt3, line 101: Is there a reason you have an empty slice of secrets here, instead of a slice that's been populated?

common/config/config_test.go
```diff
@@ -552,3 +554,34 @@ func TestScreamingSnakeCaseFlag(t *testing.T) {
...
+	expected := []string{
+		"Never gonna give you up",
+		"Never gonna let you down",
+		"Never gonna run around and desert you",
+		"Never gonna make you cry",
+		"Never gonna say goodbye",
+		"Never gonna tell a lie and hurt you",
+	}
+
+	fullString := ""
```
litt3: `fullString := strings.Join(expected, ", ")` seems better here

common/config/secret/secret_parser.go
```diff
@@ -23,6 +24,38 @@ var DecodeHook mapstructure.DecodeHookFunc = func(from reflect.Type, to reflect.
...
+			// Get the source data as a string
+			var sourceStr string
+			switch v := data.(type) {
+			case string:
+				sourceStr = v
+			case []byte:
+				sourceStr = string(v)
+			default:
+				// If it's not a string or []byte, let mapstructure handle it normally
+				return data, nil
```
litt3: If you got to this point, that means this _was_ meant to be a secret. But in this `default` case, you are just returning the `data` without hiding it within a `Secret` struct.

I don't fully understand what edge case would cause this to happen, but this doesn't quite seem right to me.

common/config/config_test.go
cody-littley: There was a bug pointed out by the copilot reviewer that an empty list of secrets resolved to `[""]` instead of `[]`. In this specific test, the string slice is unset, so the assertion is correct.

I added a comment to explain.

```
	// The slice of secrets is unset in this config, so we should expect an empty slice.
	// There used to be a bug where it would instead return [""].
	require.Equal(t, 0, len(foo.ThisIsASliceOfSecrets))
```

common/config/config_test.go
cody-littley: change made

common/config/secret/secret_parser.go
cody-littley: Made it return an error.

```
				// If it's not a string or []byte then we can't handle it here
				return nil, fmt.Errorf("cannot convert %v to slice of Secrets", from)
```

When parsing environment variables, we will only ever encounter strings. But if we are parsing things like json, higher order types like integers and floating point values might be present.

For now, I think it's simplest if we restrict support to strings and lists of strings.

PR #2382, opened 2025-12-03
chore: delete rogue printfs

cody-littley:
Delete printfs that are spamming traffic generator logs

### DIFF
--- a/encoding/v2/kzg/verifier/verifier.go
+++ b/encoding/v2/kzg/verifier/verifier.go
@@ -279,7 +279,9 @@ func (v *Verifier) universalVerify(params encoding.EncodingParams, samples []Sam
 	// precheck
 	for _, s := range samples {
 		if s.RowIndex >= numBlobs {
-			return errors.New("sample.RowIndex and numBlob are inconsistent")
+			return fmt.Errorf(
+				"sample.RowIndex and numBlob are inconsistent: sample has %d rows, but there are only %d blobs",
+				s.RowIndex, numBlobs)
 		}
 	}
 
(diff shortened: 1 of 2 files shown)


encoding/v2/kzg/verifier/verifier.go
```diff
@@ -277,9 +277,8 @@ func genRhsG1(
 // Each sample need not have unique row, it is possible that multiple chunks of the same blob are validated altogether
 func (v *Verifier) universalVerify(params encoding.EncodingParams, samples []Sample, numBlobs int) error {
 	// precheck
-	for i, s := range samples {
+	for _, s := range samples {
 		if s.RowIndex >= numBlobs {
-			fmt.Printf("sample %v has %v Row, but there are only %v blobs\n", i, s.RowIndex, numBlobs)
 			return errors.New("sample.RowIndex and numBlob are inconsistent")
```
samlaf: prob want to incorporate the print in the error? otherwise error will not have enough info right?

encoding/v2/kzg/verifier/verifier.go
```diff
@@ -296,7 +295,6 @@ func (v *Verifier) universalVerify(params encoding.EncodingParams, samples []Sam
 	}
 
 	n := len(samples)
-	fmt.Printf("Batch verify %v frames of %v symbols out of %v blobs \n", n, params.ChunkLength, numBlobs)
```
samlaf, line 299: guessing this is the one spamming that you don't like? fine with removing but if this is useful maybe we should make it a debug statement instead?

samlaf (approved):

encoding/v2/kzg/verifier/verifier.go
cody-littley: Augmented error message to contain this data

encoding/v2/kzg/verifier/verifier.go
cody-littley: This is quite spammy, so unless there is a specific reason to leave this in I'd prefer to remove it.

bxue-l2: This was added to get some data on verification workload. Since you are doing single blob batch, no that imporatnt

PR #2383, opened 2025-12-03
feat: Implement weighted selector utility

litt3:
- This PR splits out a subset of the changes made in https://github.com/Layr-Labs/eigenda/pull/2347
- Needed for EGDA-2299 ([link](https://linear.app/eigenlabs/issue/EGDA-2299/client-disperser-multiplexing))


common/selector/weighted_selector.go
```diff
@@ -0,0 +1,130 @@
...
+import (
+	"fmt"
+	"math"
+	"math/rand"
+	"slices"
+
+	"github.com/Layr-Labs/eigensdk-go/logging"
+)
+
+// Performs weighted random selection with configurable filtering of low performers.
```
cody-littley: If this is generic, "low performers" may not make sense. It's conceivable we'd want to use this utility in a scenario where weights don't correspond to performance.

PR #2385, opened 2025-12-04
feat(proxy): embed srs points into built binaries

samlaf:
This removes the need for loading SRS points from the filesystem, which makes it easier to distribute the proxy binary as a self-contained artifact. Previously the SRS points were packaged in a docker container, but the rust client may need to start a proxy binary without docker, hence this change to embed the SRS points directly into the binary.

### DIFF
--- a/api/proxy/resources/srs.go
+++ b/api/proxy/resources/srs.go
@@ -20,55 +20,34 @@ var serializedG2Points []byte
 var serializedG2TrailingPoints []byte
 
 var (
-	// 2^28 points: [1], [tau], [tau^2],..,[tau^(2^28-1)]
-	g1SRS     = make([]bn254.G1Affine, len(serializedG1Points)/kzg.G1PointBytes)
-	g1SRSOnce sync.Once
-)
-
-var (
-	// first 16MiB of G2 points: [1], [tau], [tau^2],..,[tau^(2^18-1)]
-	g2SRS     = make([]bn254.G2Affine, len(serializedG2Points)/kzg.G2PointBytes)
-	g2SRSOnce sync.Once
-)
-
-var (
-	// trailing 16MiB of G2 points
-	g2TrailingSRS     = make([]bn254.G2Affine, len(serializedG2TrailingPoints)/kzg.G2PointBytes)
-	g2TrailingSRSOnce sync.Once
-)
-
-// GetG1SRS returns the embedded G1 SRS points, deserializing them on the first call.
-// Subsequent calls return the already deserialized points.
-// This function is safe for concurrent use.
-func GetG1SRS() []bn254.G1Affine {
-	g1SRSOnce.Do(func() {
+	// Deserializes embedded G1 SRS points on first call. Safe for concurrent use.
+	// Points represent [1], [tau], [tau^2],...,[tau^(n-1)] where n is determined by the embedded file size.
+	GetG1SRS = sync.OnceValue(func() []bn254.G1Affine {
 		fmt.Println("deserializing embedded g1 srs points...")
-		deserializePoints(serializedG1Points, g1SRS, kzg.G1PointBytes)
+		points := make([]bn254.G1Affine, len(serializedG1Points)/kzg.G1PointBytes)
+		deserializePoints(serializedG1Points, points, kzg.G1PointBytes)
+		return points
 	})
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -198,7 +198,9 @@ ENTRYPOINT ["blobapi"]
 FROM alpine:3.22 AS proxy
 WORKDIR /app
 COPY --from=proxy-builder /app/api/proxy/bin/eigenda-proxy .
-COPY --from=proxy-builder /app/api/proxy/resources/ /app/resources/
+# All SRS points are now embedded into the binary, but we keep g1.point here
+# because it is needed for V1 codepaths that need to dynamically read single srs points from the file.
+COPY --from=proxy-builder /app/api/proxy/resources/g1.point /app/resources/g1.point
```
ethenotethan, line 203: is this file actually being embedded? looks like `proxy/resources/g1.point` was deleted in this PR

api/proxy/resources/srs.go
```diff
@@ -0,0 +1,95 @@
...
+//go:embed g1.point
+var serializedG1Points []byte
+
+//go:embed g2.point
+var serializedG2Points []byte
+
+//go:embed g2.trailing.point
+var serializedG2TrailingPoints []byte
+
+var (
```
litt3, line 22: Curious if you considered using `sync.OnceValue` as an alternative to `sync.Once`? IMO it seems like a much more natural fit, and would collapse down into:

```
var (
	// Deserializes G1 SRS on first call. Points represent [1], [tau], [tau^2],...,[tau^(n-1)]
	// where n is determined by the embedded file size.
	GetG1SRS = sync.OnceValue(func() []bn254.G1Affine {
		fmt.Println("deserializing embedded g1 srs points...")
		points := make([]bn254.G1Affine, len(serializedG1Points)/kzg.G1PointBytes)
		deserializePoints(serializedG1Points, points, kzg.G1PointBytes)
		return points
	})

	// Deserializes G2 SRS on first call. Points represent [1], [tau], [tau^2],...,[tau^(n-1)]
	// where n is determined by the embedded file size.
	GetG2SRS = sync.OnceValue(func() []bn254.G2Affine {
		fmt.Println("deserializing embedded g2 srs points...")
		points := make([]bn254.G2Affine, len(serializedG2Points)/kzg.G2PointBytes)
		deserializePoints(serializedG2Points, points, kzg.G2PointBytes)
		return points
	})

	// Deserializes G2 trailing SRS on first call.
	GetG2TrailingSRS = sync.OnceValue(func() []bn254.G2Affine {
		fmt.Println("deserializing embedded g2 srs trailing points...")
		points := make([]bn254.G2Affine, len(serializedG2TrailingPoints)/kzg.G2PointBytes)
		deserializePoints(serializedG2TrailingPoints, points, kzg.G2PointBytes)
		return points
	})
)
```

Dockerfile
samlaf: wasn't deleted, I just shrunk it from 32MiB to 16MiB.
It is being embedded in srs.go, but also still being copied into the docker container here for V1 only.

api/proxy/resources/srs.go
samlaf: nice, updated: [87e6f0f](https://github.com/Layr-Labs/eigenda/pull/2385/commits/87e6f0f436570852b870cbe380af99d3c35c3cb5)

api/proxy/resources/srs.go
```diff
@@ -0,0 +1,75 @@
...
+	// Points represent [1], [tau], [tau^2],...,[tau^(n-1)] where n is determined by the embedded file size.
+	GetG2SRS = sync.OnceValue(func() []bn254.G2Affine {
+		fmt.Println("deserializing embedded g2 srs points...")
+		points := make([]bn254.G2Affine, len(serializedG2Points)/kzg.G2PointBytes)
+		deserializePoints(serializedG2Points, points, kzg.G2PointBytes)
+		return points
+	})
+
+	// Deserializes embedded G2 trailing SRS points on first call. Safe for concurrent use.
+	// Points represent [tau^(2^28 - n)], [tau^(2^28 - n +1)],...,[tau^(2^28 -1)],
```
bxue-l2, line 42: looks right

PR #2388, opened 2025-12-04
feat(metrics): Add per account blob status reporting

litt3:
- The addition of the `accountId` labels will give us detailed insight into success + failures for individual users


cody-littley (approved):
LGTM. When this merges, we'll need to do some grafana dashboard iteration.

disperser/cmd/controller/flags/flags.go
```diff
@@ -183,6 +183,12 @@ var (
 		Required: false,
 		EnvVar:   common.PrefixEnvVar(envVarPrefix, "DETAILED_VALIDATOR_METRICS"),
 	}
+	EnablePerAccountBlobStatusMetricsFlag = cli.BoolTFlag{
+		Name:     common.PrefixFlag(FlagPrefix, "enable-per-account-blob-status-metrics"),
+		Usage:    "Whether to report per-account blob status metrics. If false, all metrics will be aggregated under account 0x0. (default: true)",
```
dmanc, line 188: What would be the alternative assuming we had a lot of Account IDs? I'm afraid that we would start relying on this metric for alerts and have to lose all the account labels in the future.

dmanc (dismissed):

disperser/cmd/controller/flags/flags.go
litt3: > What would be the alternative assuming we had a lot of Account IDs?

Alternatively, we could define a set of keys that we care about, and only report those accounts directly. IMO it's a bit of a premature optimization to do that right now.

> start relying on this metric for alerts

I'm not sure it would make sense to configure alerts on a per-account basis. Do you have a reason in mind that we would want to do that?

dmanc: Was thinking we probably want to be aware when a key customer stops posting blobs or has an increase in failed blobs and follow up with them.

litt3: I'm tinkering with a solution that would let us inject an account -> user mapping from the devops repo. This would allow us to pretty easily limit the cardinality to our major users. I'll follow up with this if/when I get that feature done

PR #2391, opened 2025-12-04
feat(metrics): Implement account ID -> account name metric remapping

litt3:
- Implements a simple name remapping system, using yaml files injected at deploy time
- This PR doesn't do all the name remapping we'll want to have eventually. For example, there are still payment metrics reported without the remapping
- Eventually, we'll be able to also do name remapping for validators and dispersers, to improve metric readability


common/nameremapping/name_remapping.go
cody-littley, line 1: Nit. This utility seems specific to the disperser, but `common` is a place for utilities that are also useful to the validators/clients. The wording in the godocs seem to make this utility seem fairly specific to the disperser's use case. Do you think these specific functions will be something we will end up using elsewhere?

Would it make sense to move this utility into `disperser/nameremapping`?

disperser/controller/controller_metrics.go
```diff
@@ -437,6 +438,32 @@ func (m *controllerMetrics) reportBlobSetSize(size int) {
...
+	} else {
+		// if no remapping is found, and per-account metrics are enabled, just return the account ID
+		if m.enablePerAccountMetrics {
+			return accountId
+		} else {
+			// otherwise, return the catch-all label to keep cardinality low
+			return "0x0"
+		}
+	}
+}
```
cody-littley: This logic is pretty similar to the logic in the encoder. Since you've already split out a utility file, would it make sense to capture this logic in it as well? A function signature might look something like this:

```
func GetAccountLabel(baseId string, remappedNames map[string]string, highCardinalityNames bool)
```

disperser/controller/controller_metrics.go
```diff
@@ -437,6 +438,32 @@ func (m *controllerMetrics) reportBlobSetSize(size int) {
 	m.blobSetSize.WithLabelValues().Set(float64(size))
 }
 
+// Gets the appropriate account label based on remapping and per-account settings.
+func (m *controllerMetrics) getAccountLabel(accountId string) string {
+	if m.userAccountRemapping == nil {
+		if m.enablePerAccountMetrics {
```
cody-littley: Would it make sense to move this up to the top of the function? Then, you could remove it from both places where you check it below.

```
if !m.enablePerAccountMetrics {
    return "0x0"
}
```

This comment also applies to the implementation in the controller (if you don't end up pulling this logic into the utility).

common/nameremapping/name_remapping.go
litt3: I actually don't think it's specific to the disperser.

For example, once we have multiple dispersers, we might want to use name remapping on the _client_, side to report metrics for different dispersers with user friendly names. It also might be possible that a validator would want to report metrics on a per-account basis.

Which parts of the docs do you think sound disperser specific?

disperser/controller/controller_metrics.go
```diff
@@ -437,6 +438,32 @@ func (m *controllerMetrics) reportBlobSetSize(size int) {
...
+			// if there aren't any remappings, and per-account metrics are enabled, just return the account ID
+			return accountId
+		} else {
+			// otherwise, return the catch-all label to keep cardinality low
+			return "0x0"
+		}
+	}
+
+	if remappedName, found := m.userAccountRemapping[accountId]; found && remappedName != "" {
```
pschork: Is the `accountId` guaranteed to be a check summed address? The devops remap definition is using check summed addresses.

disperser/controller/controller_metrics.go
litt3: Done ec68f42db1400442dbd8f2a00224b7e047e45a43

disperser/controller/controller_metrics.go
litt3:
```
if !m.enablePerAccountMetrics {
    return "0x0"
}
```

This isn't exactly what we want: the idea is that, if there is a remapping available, it's always used. And `enablePerAccountMetrics` only applies to accounts that don't have a remapping defined.

I simplified the logic in ec68f42db1400442dbd8f2a00224b7e047e45a43 to be less confusing. LMK what you think: is it a problem to no longer expose a lever to completely disable per-account metrics? My rationale is that if you really don't want any account-based metrics, you could remove all remappings

disperser/controller/controller_metrics.go
litt3: > Is the accountId guaranteed to be a check summed address? 

Currently, yes: the reporting methods use `gethcommon.Address.Hex()`, which yields the checksummed version.

IMO it's probably best to keep the remapping simple (i.e., don't add special considerations for checksummed addresses). And then we can just make sure to pass checksummed addresses when reporting the metrics. It'd be immediately obvious if we passed in a non-checksummed address, since the grafana metrics would show addresses instead of the remapped names

PR #2397, opened 2025-12-05
feat(clients): Add v4 cert to offchain

iquidus:
Adds v4 cert to offchain

* Adds `EigenDACertV4` type
* Adds funcs to `EigenDACertV4` to satisfy `EigenDACert` interface
* Adds `NewEigenDACertV4` constructor
* Includes some minor refactoring to reduce duplicate code
* Updates cert builder to support V4
* Updates offchain cert verifier to read and cache offchain derivation version from contract

### DIFF
--- a/api/proxy/test/e2e/server_rest_test.go
+++ b/api/proxy/test/e2e/server_rest_test.go
@@ -219,7 +219,7 @@ func testProxyClientServerIntegration(t *testing.T, dispersalBackend common.Eige
 					err.Error(),
 					"404") && !isNilPtrDerefPanic(err.Error()))
 
-			testCert = []byte{3}
+			testCert = []byte{4}
 			_, err = daClient.GetData(ts.Ctx, testCert)
 			require.Error(t, err)
 			assert.True(


api/clients/v2/coretypes/eigenda_cert.go
```diff
@@ -82,13 +91,154 @@ func DeserializeEigenDACert(data []byte, version CertificateVersion, ct CertSeri
...
 	default:
 		return nil, fmt.Errorf("unsupported certificate version: %d", version)
 	}
 }
 
 var _ EigenDACert = &EigenDACertV2{}
 var _ EigenDACert = &EigenDACertV3{}
+var _ EigenDACert = &EigenDACertV4{}
+
+// This struct represents the composition of a EigenDA V3 certificate, as it would exist in a rollup inbox.
```
ethenotethan: wrong version

api/clients/v2/coretypes/eigenda_cert.go
```diff
@@ -82,13 +91,154 @@ func DeserializeEigenDACert(data []byte, version CertificateVersion, ct CertSeri
...
+	blobStatusReply *disperser.BlobStatusReply,
+	nonSignerStakesAndSignature *certTypesBinding.EigenDATypesV1NonSignerStakesAndSignature,
+	offchainDerivationVersion uint16,
+) (*EigenDACertV4, error) {
+	certv3, err := NewEigenDACertV3(blobStatusReply, nonSignerStakesAndSignature)
+	if err != nil {
+		return nil, fmt.Errorf("create EigenDACertV3: %w", err)
+	}
+
+	return certv3.ToV4(offchainDerivationVersion), nil
```
ethenotethan: nit - I understand the rationale behind this approach but would personally prefer if we explicitly added an independent constructor for the V4 cert given there is some unnecessary overhead with this operation

ethenotethan: Screenshot 2025-12-05 at 10 57 32 PM

api/clients/v2/coretypes/eigenda_cert.go
```diff
@@ -82,13 +91,154 @@ func DeserializeEigenDACert(data []byte, version CertificateVersion, ct CertSeri
...
+		return nil, fmt.Errorf("unknown serialization type: %d", ct)
+	}
+}
+
+// Commitments returns the blob's cryptographic kzg commitments
+func (c *EigenDACertV4) Commitments() (*encoding.BlobCommitments, error) {
+	return commitments(&c.BlobInclusionInfo)
+}
+
+// Version returns the version of the EigenDA certificate
```
ethenotethan: bad doc string

api/clients/v2/dispersal/payload_disperser.go
```diff
@@ -203,10 +203,20 @@ func (pd *PayloadDisperser) buildEigenDACert(
...
+	// For cert versions >= V4, we need to get the offchain derivation version from the CertVerifier contract
+	var offchainDerivationVersion uint16
+	if certVersion >= coretypes.VersionFourCert {
+		offchainDerivationVersion, err = pd.certVerifier.GetOffchainDerivationVersion(
+			ctx, blobStatusReply.GetSignedBatch().GetHeader().GetReferenceBlockNumber())
+		if err != nil {
+			return nil, fmt.Errorf("get offchain derivation version: %w", err)
+		}
+	}
+
```
samlaf, line 215: isnt it better to do this inside certBuilder.BuildCert? We already have a switch statement in there. Almost contains any eth-calls to that function (similar to how we already get the operator state retriever stuff)

api/clients/v2/coretypes/eigenda_cert.go
```diff
@@ -82,13 +91,173 @@ func DeserializeEigenDACert(data []byte, version CertificateVersion, ct CertSeri
...
 		return nil, fmt.Errorf("unsupported certificate version: %d", version)
 	}
 }
 
 var _ EigenDACert = &EigenDACertV2{}
 var _ EigenDACert = &EigenDACertV3{}
+var _ EigenDACert = &EigenDACertV4{}
+
+// This struct represents the composition of a EigenDA V4 certificate, as it would exist in a rollup inbox.
+type EigenDACertV4 certTypesBinding.EigenDACertTypesEigenDACertV4
```
samlaf: what does composition mean here? id suggest removing that word if not needed.

samlaf: also I think the cleanest way forward for the codebase would be to define these types ourselves, and then just define conversion functions between the offchain defined types and the bindings ones. This way we could define the EigenDACertV4 to use a struct embedding of the V3 cert, and we would get all of the V3 cert methods for free (like commitments). Not sure how this behaves wrt (de)serialization though... prob would have to define that ourselves, and might not play well with rlp lib... not sure.

Anyways no need to change, but might be worth discussing a bit.

api/clients/v2/verification/cert_verifier.go
```diff
@@ -58,23 +60,48 @@ func (cv *CertVerifier) CheckDACert(
...
+
+		certBytes, err = certV4.Serialize(coretypes.CertSerializationABI)
+		if err != nil {
+			return &CertVerifierInternalError{Msg: "serialize cert", Err: err}
+		}
+	} else {
+		// Normalize cert to V3
+		certV3 := NormalizeCertV3(cert)
+		certVerifierAddr, err = cv.addressProvider.GetCertVerifierAddress(
+			ctx,
```
samlaf: do we really need this if/else? doesnt the EigenDACert interface already provide all of the methods that we need (ReferenceBlockNumber(), Serialize())

api/clients/v2/verification/cert_verifier.go
```diff
@@ -317,6 +365,44 @@ func (cv *CertVerifier) GetCertVersion(ctx context.Context, referenceBlockNumber
...
+		return castVersion, nil
+	}
+
+	// version wasn't cached, so proceed to fetch it
+	certVerifierCaller, err := cv.getVerifierCallerFromAddress(certVerifierAddress)
+	if err != nil {
+		return 0, fmt.Errorf("get verifier caller from address: %w", err)
+	}
+
+	offchainDerivationVersion, err := certVerifierCaller.OffchainDerivationVersion(&bind.CallOpts{Context: ctx})
```
samlaf, line 350: prob good to document somewhere that the offchainDerivationVersion function was introduced in CertVerifier semver 4.0.0, which coincides with certVersion 4.0. so this function should only be called with >= V4 certs.

api/clients/v2/coretypes/eigenda_cert.go
samlaf: personally dont think we should worry about 1KiB being copied over a few times if it makes the code more legible. Not saying this is not important, just that our most important metric atm imo is code maintainability going forward. Also related to my comment https://github.com/Layr-Labs/eigenda/pull/2397#discussion_r2599754062

api/clients/v2/verification/cert_verifier.go
ethenotethan: also if we need to do version aware switch then I'd rather we use the switch pattern already expressed in the EigenDA manager construct: https://github.com/Layr-Labs/eigenda/blob/151bf614652807707dc6bddfb6b76034e0e6cb93/api/proxy/store/generated_key/v2/eigenda.go#L204-L216 

this current approach results in unnecessary code traversal in the case of a V3 cert

api/proxy/store/generated_key/v2/eigenda.go
```diff
@@ -210,6 +210,12 @@ func (e Store) Put(
 			return nil, fmt.Errorf("serialize cert: %w", err)
 		}
 		return certs.NewVersionedCert(serializedCert, certs.V2VersionByte), nil
```
ethenotethan, line 212: how much longer will the proxy support V3 DA Cert? would we ever wanna update or modify the software to intentionally disallow the construction of a V3 cert?

inabox/tests/integration_v2_test.go
```diff
@@ -62,35 +62,35 @@ func TestEndToEndV2Scenario(t *testing.T) {
 	err = testHarness.RouterCertVerifier.CheckDACert(ctx, cert2)
 	require.NoError(t, err)
 
-	eigenDAV3Cert1, ok := cert1.(*coretypes.EigenDACertV3)
+	eigenDAV4Cert1, ok := cert1.(*coretypes.EigenDACertV4)
```
ethenotethan, line 65: if we keep the construction of V3 cert in the proxy then removing correctness tests here could be considered a coverage regression

api/clients/v2/verification/cert_verifier.go
```diff
@@ -296,7 +344,7 @@ func (cv *CertVerifier) GetCertVersion(ctx context.Context, referenceBlockNumber
 	if ok {
 		castVersion, ok := cachedVersion.(uint8)
 		if !ok {
-			return 0, fmt.Errorf("expected version to be uint64")
+			return 0, fmt.Errorf("expected version to be uint8")
```
ethenotethan, line 300: nit:
```suggestion
			return 0, fmt.Errorf("expected DA Cert version to be of type uint8")
```

api/clients/v2/dispersal/payload_disperser.go
```diff
@@ -203,10 +203,20 @@ func (pd *PayloadDisperser) buildEigenDACert(
 		return nil, fmt.Errorf("get certificate version: %w", err)
 	}
 
+	// For cert versions >= V4, we need to get the offchain derivation version from the CertVerifier contract
+	var offchainDerivationVersion uint16
+	if certVersion >= coretypes.VersionFourCert {
+		offchainDerivationVersion, err = pd.certVerifier.GetOffchainDerivationVersion(
+			ctx, blobStatusReply.GetSignedBatch().GetHeader().GetReferenceBlockNumber())
+		if err != nil {
+			return nil, fmt.Errorf("get offchain derivation version: %w", err)
```
bxue-l2, line 212: I think it worths to deliberate the migration path.

should we disperse a V3 cert is offchain derivation is not available.

Option 1:
For example, when celo wants to upgrade, they need to schedule an activation block number(ABN) for the new certVerifier. It would return 500s error, if the block number does not reach the ABN. It basically means the rollup devops teams must restart the batcher manually right after the activation number is reached. There are a period of unavailbility there.

Option 2:
If we failover to V3 cert, batcher can upgrade early and be ready to see the activation number when it happens. But for celo, since they are using V2 cert, there will be a period of time, celo disperser will use v3 cert (proxy after 2.0.0 release no longer supports dispersal with V2cert). But given we can only assume all Celo nodes are (V4 and V3) compatible after the activation number, and otherwise celo nodes are running 1.8.2 legacy version, we face a dangereous situation of having 1.8.2 proxy parsing V3 cert. Which I don't think is compatible, given RLP decoding is sensitive to order.

New idea,
I am thinking we should have a flag just for celo which just panic if offchain derivation version can't be found; for the rest of users, we fallback to V3 cert when dispersal. @iquidus @samlaf @ethenotethan wdyt.

But since this is only 1 time migration. Also fine if we just keep it as what it is

api/clients/v2/verification/cert_verifier.go
```diff
@@ -317,6 +365,44 @@ func (cv *CertVerifier) GetCertVersion(ctx context.Context, referenceBlockNumber
...
+// derivation version and cache the result for future use.
+func (cv *CertVerifier) GetOffchainDerivationVersion(ctx context.Context, referenceBlockNumber uint64) (uint16, error) {
+	certVerifierAddress, err := cv.addressProvider.GetCertVerifierAddress(ctx, referenceBlockNumber)
+	if err != nil {
+		return 0, fmt.Errorf("get cert verifier address: %w", err)
+	}
+
+	// if the offchain derivation version for the active cert verifier address has already been cached, return it
+	// immediately
+	cachedVersion, ok := cv.offchainDerivationVersions.Load(certVerifierAddress)
```
bxue-l2, line 335: Does that mean even if the onchain contract changes, the offchain code won't update if it already cached. 

I guess it is fine here because if the offchain derivation version ever changes, the node must upgrade anyway

api/clients/v2/dispersal/payload_disperser.go
bxue-l2: I think this problem is partially addressed, but I second this comment https://github.com/Layr-Labs/eigenda/pull/2397/files#r2599735709

api/clients/v2/verification/cert_verifier.go
iquidus: The interface provides the methods yes, however in the case of a V2 cert we need to convert it to a V3 before calling the methods. Will do some refactoring.

inabox/tests/integration_v2_test.go
iquidus: afaik inabox deploys the latest cert verifier contract. We'll need coverage for V3 via testnet e2e tests. Or maintain two versions of the contract (V3, and V4).

api/clients/v2/verification/cert_verifier.go
iquidus: The onchain contract can only change via a new deployment, with a new address. the cache is a mapping of certVerifierAddess->offchainDerivationVersion. The certVerifierAddress is retrieved using the RBN in the cert. If a new certVerifier is active at the certs RBN, the cache for that address will be empty, it will fetch the derivation version from the contract then cache it for that instance of the cert verifier contract.

api/proxy/store/generated_key/v2/eigenda.go
iquidus: This I am unsure on. For example Celo is still using V2 certs, in our last call it seems like they are happy to update straight to V4 (skipping V3) in Q1. I do not know the status/plans of other users, we will want to make sure everyone is on V4 before dropping V3.

iquidus: If we were to use offchainDerivationVersion: 1 as the first version for v4 (instead of 0), we could use 0 as pre recency-window (e.g V3). Then we could simply convert v3 certs to v4 much like we do v2->v3.

api/clients/v2/verification/cert_verifier.go
iquidus: See: [1feec18](https://github.com/Layr-Labs/eigenda/pull/2397/commits/1feec186dc09c8786a949343858d43d9af0724f2)

api/clients/v2/dispersal/payload_disperser.go
iquidus: import cycle makes this more complicated than at first glance.

samlaf (approved):
My comments have been addressed so stamping but seems like there's still some issues surrounding upgradeability that you guys are dealing with.

api/proxy/store/generated_key/v2/eigenda.go
bxue-l2: For celo, their migration plan is something like option 1 listed here, https://github.com/Layr-Labs/eigenda/pull/2397/changes#r2600401167

inabox/tests/integration_v2_test.go
bxue-l2: We cannot change the router to update V3 yet, so if we want to test V4 end2end on sepolia, we will have to deploy a new router

bxue-l2 (dismissed):

api/proxy/test/e2e/server_rest_test.go
```diff
@@ -219,7 +219,7 @@ func testProxyClientServerIntegration(t *testing.T, dispersalBackend common.Eige
 					err.Error(),
 					"404") && !isNilPtrDerefPanic(err.Error()))
 
-			testCert = []byte{3}
+			testCert = []byte{4}
```
bxue-l2, line 222: should we just add a new test

inabox/tests/integration_v2_test.go
bxue-l2: We need to maintain V3 integration for quite some time before all the customer migrating off our router, so good to keep the v3 test

api/proxy/test/e2e/server_rest_test.go
iquidus: This test was failing after the introduction of the V3VersionByte (keep in mind the offset between version bytes and cert versions). The seemingly arbitrary single byte was increased to 4 so the test behaves as it did prior to this PR. (it would fail if the byte was set to 1, or 2).

inabox/tests/integration_v2_test.go
iquidus: Inabox uses the contract deploy scripts from the branch its running in to deploy the contracts used for testing. As is, this branch only has V4 (ignoring legacy). the cert it's type checking here is v4, checking against v3 fails (as it should for a v4 cert). As mentioned earlier, we'll need both V3 and V4 versions of the cert verifier contract.

api/proxy/test/e2e/server_rest_test.go
bxue-l2: oh, that makes sense

inabox/tests/integration_v2_test.go
iquidus: Will add v3.2.0 to the legacy dir (as well as the lib), and update tests - in a separate PR

bxue-l2: I see, so the question is whether we can coverage for V3 anymore. Maybe it is something we can ponder on.

bxue-l2: It probably makes sense to merge this ticket and then add V3 cert test if we think it is necessary

PR #2398, opened 2025-12-05
feat(rust): validate correct cert-verifier is used via router proof-of-storage

samlaf:
Integration test passing against sepolia, but still WIP to clean this up.

Realized the library's logic for extracting keys from dynamic arrays was assuming no packing, which works for all the data structures extracted from our middleware contracts, but didn't work for the []uint32 abn array we have in the router (in which 8 abns get packed in every slot). So had to update that to fix it. Added more notes/warnings in the library extraction code.

### DIFF
--- a/rust/crates/eigenda-verification/src/extraction/extractor.rs
+++ b/rust/crates/eigenda-verification/src/extraction/extractor.rs
@@ -651,7 +651,11 @@ impl CertVerifierABNsExtractor {
 
 impl StorageKeyProvider for CertVerifierABNsExtractor {
     fn storage_keys(&self) -> Vec<StorageKey> {
-        storage_key_helpers::dynamic_array_keys(CERT_VERIFIER_ABNS_ARRAY_SLOT, self.num_abns, U32::BITS)
+        storage_key_helpers::dynamic_array_keys(
+            CERT_VERIFIER_ABNS_ARRAY_SLOT,
+            self.num_abns,
+            U32::BITS,
+        )
     }
 }
 


rust/crates/eigenda-verification/src/extraction/extractor.rs
```diff
@@ -653,6 +665,46 @@ impl StorageKeyProvider for CertVerifierABNsExtractor {
...
+            )?;
+            for i in 0..8 {
+                if out.len() >= self.num_abns {
+                    break;
+                }
+                let start = 32 - (i + 1) * 4;
+                let end = start + 4;
+                let abn_bytes: [u8; 4] = proof.value.to_be_bytes::<32>()[start..end]
+                    .try_into()
+                    .unwrap();
```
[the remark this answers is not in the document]
samlaf: there's already a bunch of unwrap()s in this file, so keeping as is for now.

rust/crates/eigenda-verification/src/extraction/mod.rs
```diff
@@ -159,6 +181,32 @@ impl CertStateData {
...
+            .iter()
+            .rev()
+            .find(|abn| **abn as u64 <= rbn)
+            .ok_or_else(|| CertExtractionError::NoActiveCertVerifierAtRBN { rbn: rbn })?;
+        let cert_verifier = cert_verifiers
+            .get(abn)
+            .ok_or_else(|| CertExtractionError::NoActiveCertVerifierAtRBN { rbn: rbn })?;
+        if !cert_verifier.eq(&self.cert_verifier.address) {
+            return Err(CertExtractionError::WrongActiveCertVerifier {
+                rbn: rbn,
```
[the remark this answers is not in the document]
samlaf: [7ee2acb](https://github.com/Layr-Labs/eigenda/pull/2398/commits/7ee2acb693da147588862f415d68c2f41b76887b)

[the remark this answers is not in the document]
samlaf: [7ee2acb](https://github.com/Layr-Labs/eigenda/pull/2398/commits/7ee2acb693da147588862f415d68c2f41b76887b)

rust/crates/eigenda-verification/src/extraction/mod.rs
```diff
@@ -159,6 +181,32 @@ impl CertStateData {
...
+
+        let rbn = cert.reference_block();
+        let abn = abns
+            .iter()
+            .rev()
+            .find(|abn| **abn as u64 <= rbn)
+            .ok_or_else(|| CertExtractionError::NoActiveCertVerifierAtRBN { rbn: rbn })?;
+        let cert_verifier = cert_verifiers
+            .get(abn)
+            .ok_or_else(|| CertExtractionError::NoActiveCertVerifierAtRBN { rbn: rbn })?;
```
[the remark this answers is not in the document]
samlaf: [7ee2acb](https://github.com/Layr-Labs/eigenda/pull/2398/commits/7ee2acb693da147588862f415d68c2f41b76887b)

rust/crates/eigenda-verification/src/extraction/extractor.rs
```diff
@@ -111,7 +110,7 @@ impl StorageKeyProvider for QuorumCountExtractor {
 impl DataDecoder for QuorumCountExtractor {
     type Output = u8;
 
-    #[instrument(skip_all, fields(component = std::any::type_name::<Self>()))]
+    #[instrument(skip_all, fields(component = std::any::type_name::<Self>().split("::").last().unwrap_or("Unknown")), ret)]
```
bxue-l2, line 113: what does it print like

rust/crates/eigenda-verification/src/extraction/extractor.rs
```diff
@@ -610,14 +601,31 @@ impl Default for CertVerifierABNsLenExtractor {
...
 
+impl DataDecoder for CertVerifierABNsLenExtractor {
+    type Output = u32;
+
+    #[instrument(skip_all, fields(component = std::any::type_name::<Self>().split("::").last().unwrap_or("Unknown")), ret)]
+    fn decode_data(
+        &self,
+        storage_proofs: &[StorageProof],
+    ) -> Result<Self::Output, CertExtractionError> {
+        let storage_key = &self.storage_keys()[0];
```
bxue-l2, line 620: why you always take the first one? why not just pass a single element

rust/crates/eigenda-verification/src/extraction/extractor.rs
```diff
@@ -649,6 +661,45 @@ impl StorageKeyProvider for CertVerifierABNsExtractor {
...
+            let proof = decode_helpers::find_required_proof(
+                storage_proofs,
+                storage_key,
+                "certVerifierABNs",
+            )?;
+            for i in 0..8 {
+                if out.len() >= self.num_abns {
+                    break;
+                }
+                let start = 32 - (i + 1) * 4;
```
bxue-l2: better document the funciton. Why you reverse the abn parsing? If assuming the nature order, abn array would look like 10,20,30. After parsing it becomes 30, 20, 10. Then the window func below won't pass.

rust/crates/eigenda-verification/src/extraction/extractor.rs
```diff
@@ -688,6 +738,29 @@ impl<'a> StorageKeyProvider for CertVerifiersExtractor<'a> {
...
+    type Output = HashMap<u32, Address>;
+
+    #[instrument(skip_all, fields(component = std::any::type_name::<Self>().split("::").last().unwrap_or("Unknown")), ret)]
+    fn decode_data(
+        &self,
+        storage_proofs: &[StorageProof],
+    ) -> Result<Self::Output, CertExtractionError> {
+        let mut out: HashMap<u32, Address> = HashMap::new();
+
+        for (&abn, storage_key) in self.abns.iter().zip(self.storage_keys().iter()) {
```
bxue-l2, line 757: how do they ensure those have the same length

rust/crates/eigenda-verification/src/extraction/extractor.rs
```diff
@@ -688,6 +738,29 @@ impl<'a> StorageKeyProvider for CertVerifiersExtractor<'a> {
...
+            let proof =
+                decode_helpers::find_required_proof(storage_proofs, storage_key, "certVerifiers")?;
+            let address_bytes = proof.value.to_be_bytes::<32>();
+            let address: Address = Address::from_slice(&address_bytes[12..32]);
```
bxue-l2: what are the rest of the bytes, 0?

or is address_bytes the ecdsa public keys?

rust/crates/eigenda-verification/src/extraction/storage_key_helpers.rs
```diff
@@ -53,6 +53,9 @@ pub fn mapping_key(key: U256, slot: u64) -> StorageKey {
 /// Implements the Ethereum dynamic array storage rule:
 /// `storage_key = keccak256(slot) + index`
 ///
+/// SAFETY CAVEAT: This function assumes that the values in the array have size >= 16 bytes.
```
bxue-l2: Nit. so basically key (in the dynamic_array_key) must be u32 or higher? Do you want to put inside the func name?

rust/crates/eigenda-verification/src/extraction/extractor.rs
samlaf: You can run https://github.com/Layr-Labs/eigenda/blob/75b3f907cb6d28965d89ff13d70f8185adddf571/rust/crates/eigenda-tests/tests/integration.rs#L20 and you'll se the output.

image

The change I made is to only keep the name of the functions, removing the crate/modules prefix. Makes the output cleaner, but thinking now might be confusing if 3rd parties integrate this library into their binaries..

rust/crates/eigenda-verification/src/extraction/extractor.rs
samlaf: because storage_keys() is a trait method used for all the other types. Added an assert that length is 1: [b47bf18](https://github.com/Layr-Labs/eigenda/pull/2398/commits/b47bf18eddd6871b918ad1d83addc43fa775da21)

Tbh I've also been wondering why they made it a trait... doesn't seem needed at all, and I agree that storage_keys() could prob have slightly different signatures based on the type... but maybe we can do in future PR if we really need it?

rust/crates/eigenda-verification/src/extraction/storage_key_helpers.rs
samlaf: no must have >128 bits so that they cant get packed. so u136 and above.

Refactored the function btw to make it more legible: [3e82512](https://github.com/Layr-Labs/eigenda/pull/2398/commits/3e825126ea84c6eeff9e9ad84af8b1d243e4659a)

rust/crates/eigenda-verification/src/extraction/extractor.rs
samlaf: Actually not sure what solidity does... an address is keccak(public_key)[12..]. But anyway the address here is the address of a contract, so there's no public key associated. so for a contract I'm guessing they put 0s there...?

Changed it to use from_word instead of from_slice: [4bfb202](https://github.com/Layr-Labs/eigenda/pull/2398/commits/4bfb2024ff40777f528d903107682a2e10564674)

rust/crates/eigenda-verification/src/extraction/extractor.rs
samlaf: so surprisingly solidity packs in the reverse direction... I'm not sure why but thats the way it does it.
Cleaned up the function greatly and added documentation (pointing to solidity docs): [2e93581](https://github.com/Layr-Labs/eigenda/pull/2398/commits/2e935819dc49d49335fd02922043e328dcac9114)
Let me know what you think

rust/crates/eigenda-verification/src/extraction/extractor.rs
samlaf: good catch.. added a check: [78ec0d1](https://github.com/Layr-Labs/eigenda/pull/2398/commits/78ec0d1ac8c4ff786e071c61d84b8c5f05cdd225)

rust/crates/eigenda-verification/src/extraction/extractor.rs
bxue-l2: My guess is they internally treats uint8 array as big endian data. so 30 is the most significant bits so it appears in the front, but just guessing

